PR #1797, opened 2024-11-13
Extract view-selector component and use in cards-grid

burieberry:
- Extracted ViewSelector boxel-ui component
- Currently used in cards-grid. It will be used in Blog App etc.


richardhjtan (approved):
LGTM

PR #1798, opened 2024-11-14
Use matrix event to notify host to fetch the updated credit info

FadhlanR:
In this PR, I implemented a mechanism that allows the realm server to notify the host app to re-fetch the user's credit info. The idea is that when a user authenticates to the realm server, a Matrix room shared by both the realm server and the host app is used. This room can be reused for the notification. When the realm server successfully handles the webhook from Stripe, it sends a message with the following structure: `{msgtype: 'org.boxel.realm-server-event', body: 'billing-notification'}` to the session room. The host app will listen for this message and react by fetching the user's credit info.


packages/host/app/components/operator-mode/profile-info-popover.gts
```diff
@@ -200,35 +200,40 @@ export default class ProfileInfoPopover extends Component<ProfileInfoPopoverSign
...
 
+  constructor(...args: [any, any]) {
+    super(...args);
+    this.fetchCreditInfo.perform();
+  }
+
   @service private declare realmServer: RealmServerService;
   @service declare matrixService: MatrixService;
 
+  private fetchCreditInfo = task(async () => {
```
jurgenwerk, line 211: I think a better name would be `fetchSubscriptionData` since general billing and subscription info is included in the response, not only credits

packages/host/app/services/realm-server.ts
```diff
@@ -173,7 +181,86 @@ export default class RealmServerService extends Service {
...
+        event.getRoomId() !== sessionRoomId ||
+        event.getType() !== 'm.room.message' ||
+        content?.msgtype !== 'org.boxel.realm-server-event' ||
+        content?.body !== 'billing-notification'
+      ) {
+        return;
+      }
+
+      await this.fetchCreditInfoTask.perform();
+    });
```
habdelra: this is a weird place to do this binding, all the other matrix event bindings happen in `MatrixService.bindEventListeners()`. I would think we want to consolidate all this.

habdelra: Also, anytime you create a binding you should also be pairing it with an unbind--otherwise you can get into situations where you have multiple bindings to the same handler, like logging out and logging back in. I don't see any cleanup for this event binding

FadhlanR: @habdelra I’ve updated the code. Instead of registering the refresher in `RealmServerService`, I added a call to `realmServerService.handleRealmServerEvent` within the `processDecryptedEvent` function in `MatrixService`. Since that function is already a listener for the `RoomEvent.Timeline` event, I thought it made sense. What do you think? If this approach looks fine, I’ll update the tests accordingly.

habdelra:
Thanks, I think this approach is better and easier to reason about. I imagine that `RealmServer.handleRealmServerEvent()` is probably going to expand into a much more rich set of capabilities, for which we'll probably want to do further refactoring. But i think we can cross that bridge after we have some more realm server events that we want to listen for.

habdelra:
I think also in this PR we'll have to expand how we mock the creation of DM's in our host tests. It looks like the host tests are trying to actually create real DM rooms which i don't think will work based on how we are mocking.

packages/realm-server/tests/realm-server-test.ts
```diff
@@ -2809,6 +2809,7 @@ module('Realm Server', function (hooks) {
 
     hooks.beforeEach(async function () {
       shimExternals(virtualNetwork);
+      process.env.STRIPE_WEBHOOK_SECRET = 'stripe-webhook-secret';
```
jurgenwerk: Would it be more appropriate to put the env vars in the `pnpm test` command?

packages/realm-server/tests/realm-server-test.ts
```diff
@@ -4248,6 +4250,101 @@ module('Realm Server', function (hooks) {
...
+      let done = assert.async();
+      setTimeout(async () => {
+        let messages = await matrixClient.roomMessages(json.room);
+        assert.strictEqual(
+          messages[0].content.msgtype,
+          'org.boxel.realm-server-event',
+        );
+        assert.strictEqual(messages[0].content.body, 'billing-notification');
+        done();
+      }, 50);
```
jurgenwerk: We can do better than magic number waiting - maybe something like `waitUntil` util function that waits in very short periods until the relevant message appears.

packages/billing/stripe-webhook-handlers/index.ts
```diff
@@ -149,5 +150,6 @@ export default async function stripeWebhookHandler(
       break;
   }
 
+  sendBillingNotification(event.data.object.customer);
```
jurgenwerk: Stripe could be sending many different webhooks which we won't care about (we only care about the ones that are handled by our 3 handlers). So I think this will send a billing notification for stripe webhooks that we do not handle. I think that's redundant.

packages/realm-server/tests/realm-server-test.ts
```diff
@@ -4248,6 +4250,101 @@ module('Realm Server', function (hooks) {
...
+        .set('Accept', 'application/json')
+        .set('Content-Type', 'application/json');
+
+      const secret = process.env.STRIPE_WEBHOOK_SECRET;
+      await insertUser(dbAdapter, userId!, 'cus_123');
+      await insertPlan(dbAdapter, 'Free plan', 0, 100, 'prod_free');
+      if (!secret) {
+        throw new Error('STRIPE_WEBHOOK_SECRET is not set');
+      }
+      let event = {
```
jurgenwerk, line 4341: Here you are testing the `invoice.payment_succeeded` webhook only, while we are handling several other handlers too. 

To avoid repeating other events here I am wondering if there should also be a general test that makes sure `stripeWebhookHandler` will notify the user after any of its available handlers are called.

packages/realm-server/tests/realm-server-test.ts
FadhlanR: ~~Instead, I use `Sinon` to assert the arguments when `matrixClient.sendEvent` is called.~~

I created the `waitForBillingNotification` function.

packages/host/app/services/matrix-service.ts
```diff
@@ -1087,7 +1087,10 @@ export default class MatrixService extends Service {
           } while (currentFragmentId);
         }
       }
+    } else {
```
lukemelia: Consider making this conditional based on some quality of the event (msgtype perhaps?)

packages/host/app/services/realm-server.ts
```diff
@@ -173,9 +176,72 @@ export default class RealmServerService extends Service {
...
+    let sessionRoomId = await this.getSessionRoomId();
+    if (
+      maybeRealmServerEvent.room_id !== sessionRoomId ||
+      maybeRealmServerEvent.type !== 'm.room.message' ||
+      maybeRealmServerEvent.content?.msgtype !==
+        'org.boxel.realm-server-event' ||
+      maybeRealmServerEvent.content?.body !== 'billing-notification'
+    ) {
+      return;
+    }
```
lukemelia: Consider moving billing related logic into a billing service, and delegating to it from here or from the matrix service

packages/host/app/services/realm-server.ts
```diff
@@ -204,45 +270,47 @@ export default class RealmServerService extends Service {
...
-      return;
-    }
-    let response = await this.network.authedFetch(
-      `${this.url.origin}/_catalog-realms`,
-    );
-    if (response.status !== 200) {
-      throw new Error(
-        `Failed to fetch public realms for realm server ${this.url.origin}: ${response.status}`,
-      );
+  private async getSessionRoomId(): Promise<string> {
```
lukemelia: Instead of this approach, let's store the session room ID in the JWT that we receive at login. This way, we can read it from there whenever we need to, without an HTTP request, and we avoid confusing 401 requests in network log.

packages/host/tests/helpers/index.gts
```diff
@@ -632,3 +632,93 @@ export async function elementIsVisible(element: Element) {
...
+    },
+  ];
+
+  let handleRealmServerRequest = async (req: Request) => {
+    let endpoint = endpoints?.find((e) => req.url.includes(e.route));
+    if (endpoint) {
+      return await endpoint.getResponse(req);
+    }
+
+    let defaultEndpoint = defaultEndpoints.find((e) =>
```
lukemelia:
```suggestion
    endpoint = defaultEndpoints.find((e) =>
```

packages/host/tests/helpers/index.gts
```diff
@@ -632,3 +632,93 @@ export async function elementIsVisible(element: Element) {
...
+      req.url.includes(e.route),
+    );
+    if (defaultEndpoint) {
```
lukemelia:
```suggestion
    if (endpoint) {
```

packages/host/tests/helpers/index.gts
```diff
@@ -632,3 +632,93 @@ export async function elementIsVisible(element: Element) {
...
+      return await defaultEndpoint.getResponse(req);
```
lukemelia:
```suggestion
      return await endpoint.getResponse(req);
```

packages/host/tests/helpers/index.gts
```diff
@@ -632,3 +632,93 @@ export async function elementIsVisible(element: Element) {
...
+
+    return null;
+  };
+
+  hooks.beforeEach(function () {
+    lookupNetworkService().mount(handleRealmServerRequest, { prepend: true });
+  });
+
+  hooks.afterEach(function () {
+    lookupNetworkService().virtualNetwork.unmount(handleRealmServerRequest);
```
lukemelia: I think we can do without this, because the virtual network will be re-created every test

packages/host/app/services/realm-server.ts
lukemelia: Also, re-order the conditional checks here so that we don't get the room session ID unless we actually have something to process. This should make it so that most tests don't need the extra setupRealmServerEndpoints hook.

packages/realm-server/server.ts
```diff
@@ -423,6 +424,24 @@ export class RealmServer {
...
+      )) ?? {};
+    let roomId = dmRooms[user];
+    if (!roomId) {
+      console.error(
+        `Failed to send event: ${eventName}, cannot find session room for user: ${user}`,
+      );
+    }
+
+    await this.matrixClient.sendEvent(roomId, 'm.room.message', {
+      body: `${eventName}`,
```
lukemelia:
```suggestion
      body: `{ eventType: ${eventName} }`,
```

This would let us add additional payload in the future without breaking existing handler code

packages/realm-server/tests/realm-server-test.ts
```diff
@@ -4085,6 +4100,16 @@ module('Realm Server', function (hooks) {
...
+      sendEvent.callsFake((_roomId, type, content) => {
+        assert.strictEqual(_roomId, roomId);
+        assert.strictEqual(type, 'm.room.message');
+        assert.deepEqual(content, {
+          body: 'billing-notification',
+          msgtype: 'org.boxel.realm-server-event',
+        });
+
+        return 'test-event-id';
+      });
```
lukemelia: Instead of using sinon, let's exercise the matrix behavior more fully and assert on the room messages from the end user point of view, and not on the internals of how the realm server used sendEvent

packages/billing/stripe-webhook-handlers/index.ts
FadhlanR: I moved this line to after calling each handler.

packages/realm-server/tests/realm-server-test.ts
FadhlanR: Since we already have a test for payment deleted event, I added the assertion there for billing notification and I also created a test for checkout session completed event.

FadhlanR:
I made a few updates based on feedback from @lukemelia during our 1-on-1 session. Here are the updates:

1. I included the session room ID in the JWT, so the host no longer needs to fetch the `_server-session` API to retrieve the room ID for listening to realm server events.

2. I updated the message body for `billing-notification` to a JSON string: `{ eventType: 'billing-notification' }`. This structure is intended to be extensible for other events in the future.

3. I created a `billing-service` in the host and moved all subscription-related code from `realm-server-service` to the new service.

lukemelia (approved):

jurgenwerk (approved):

PR #1799, opened 2024-11-14
Use open router for ai bot

jurgenwerk:
We are replacing the AI "backend" from OpenAI to OpenRouter so that we can track $ cost for our AI generations (needed for tracking user credit spending), and be more flexible about which AI model to use in case we'll want to change it later.

This depends on https://github.com/cardstack/infra/pull/561 - need to deploy that one first.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -100,7 +100,7 @@ To build the entire repository and run the application, follow these steps:
 
     ```zsh
     cd ./packages/ai-bot
-    OPENAI_API_KEY=*** pnpm start
+    OPENROUTER_API_KEY=*** pnpm start
     ```
 
 ## Cleanup command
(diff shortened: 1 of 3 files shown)


richardhjtan (approved):

IanCal:
Does the model name need to be updated too? I think the model name should be `openai/chatgpt-4o-latest` https://openrouter.ai/openai/chatgpt-4o-latest unless openrouter recognises the openai names as aliases

jurgenwerk:
@IanCal yeah you were right, needed to update the model (see https://github.com/cardstack/boxel/pull/1799/commits/e843e47d65a228c85dcaf895ed87a892536af0d6), if I use `model: 'openai/gpt-4o'` then I am able to see this model is actually used for generating responses:

image

However I am having difficulties using the `latest` version of gpt-4o. Here's what happens if I use `openai/chatgpt-4o-latest` (which is the codename from your link):

image

IanCal:
@jurgenwerk I think the one you've found is the correct one, ignore the one I got 

edit - iiuc, the one you have there should always point to the latest production model of 4o, which is what we want

PR #1800, opened 2024-11-14
Using atom format for team-member in task app

burieberry:
@lucaslyl @tintinthong Here is how you can use the atom format without sacrificing on the `scoped` attribute. You can set `@displayContainer` to false to hide the borders and add a classname on the field component to style it from where you consume it.

I noticed that the overlay design doesn't work so well on these. I think any improvements needed for the overlays should be handled by core team so that it works for all apps they're used in.

### DIFF
--- a/packages/experiments-realm/productivity/task.gts
+++ b/packages/experiments-realm/productivity/task.gts
@@ -232,17 +232,19 @@ export class TeamMember extends User {
 
   static atom = class Atom extends Component<typeof this> {
     <template>
-      <div class='assignee-display'>
-        <Avatar
-          class='avatar'
-          @userId={{@model.id}}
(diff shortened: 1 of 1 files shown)


lucaslyl:
> @lucaslyl @tintinthong Here is how you can use the atom format without sacrificing on the `scoped` attribute. You can set `@displayContainer` to false to hide the borders and add a classname on the field component to style it from where you consume it.
> 
> I noticed that the overlay design doesn't work so well on these. I think any improvements needed for the overlays should be handled by core team so that it works for all apps they're used in.

@burieberry - thank you for pointing out that we have the displayContainer feature for the atom view. @tintinthong ,We will raise a ticket for this issue.

PR #1803, opened 2024-11-15
Set browser tab title based on stack item

habdelra:
This PR sets the browser tab title and history item based on the title of the top most card of the right stack. As well as, it also sets the browser tab title based on the file displayed in code mode.

title


packages/host/app/templates/host-freestyle.gts
```diff
@@ -48,8 +47,6 @@ class HostFreestyleComponent extends Component<HostFreestyleSignature> {
   }
 
   <template>
-    {{pageTitle 'Host Components'}}
```
habdelra, line 51: this collides with our dynamic <title> element, and `pageTitle` complains loudly that there can't 2 title elements.

jurgenwerk (approved):
To me it's a bit odd to see URLs in the titles

Screenshot from your video:
image

I'd prefer the following nomenclature for titles

Hello World (Code mode)
Experiments Workspace 
Argentina
Argentina (Code mode)
...

Not sure if this was already discussed, just saying my opinion, otherwise LGTM!

habdelra:
> To me it's a bit odd to see URLs in the titles
> 
> Screenshot from your video: image
> 
> I'd prefer the following nomenclature for titles
> 
> Hello World (Code mode) Experiments Workspace Argentina Argentina (Code mode) ...
> 
> Not sure if this was already discussed, just saying my opinion, otherwise LGTM!

well so the challenge is that in code mode specifically you don't always have a title. consider a card that has a bug in its card def. neither it nor any of its instances will be runnable--what would the title of the window be? what about files that are not instances nor card field field--like an arbitrary .TS file? i think there are many circumstances where is just not clear what to use for a title in code mode aside from the code path of the file you are editing

habdelra:
> > To me it's a bit odd to see URLs in the titles
> > Screenshot from your video: image
> > I'd prefer the following nomenclature for titles
> > Hello World (Code mode) Experiments Workspace Argentina Argentina (Code mode) ...
> > Not sure if this was already discussed, just saying my opinion, otherwise LGTM!
> 
> well so the challenge is that in code mode specifically you don't always have a title. consider a card that has a bug in its card def. neither it nor any of its instances will be runnable--what would the title of the window be. what about files that are not instances nor card field field--like an arbitrary .TS file? i think there are many circumstances where is just not clear what to use for a title in code mode aside from the code path of the file you are editing

lukemelia:
> well so the challenge is that in code mode specifically you don't always have a title. consider a card that has a bug in its card def. neither it nor any of its instances will be runnable--what would the title of the window be? what about files that are not instances nor card field field--like an arbitrary .TS file? i think there are many circumstances where is just not clear what to use for a title in code mode aside from the code path of the file you are editing

It might be better to use "[path] in [Workspace Name]". i.e. "AIAppGenerator/commands.ts in Experiments" Less URL-like, but communicates effectively.

lukemelia (approved):

habdelra:
> > well so the challenge is that in code mode specifically you don't always have a title. consider a card that has a bug in its card def. neither it nor any of its instances will be runnable--what would the title of the window be? what about files that are not instances nor card field field--like an arbitrary .TS file? i think there are many circumstances where is just not clear what to use for a title in code mode aside from the code path of the file you are editing
> 
> It might be better to use "[path] in [Workspace Name]". i.e. "AIAppGenerator/commands.ts in Experiments" Less URL-like, but communicates effectively.

fixed in https://github.com/cardstack/boxel/pull/1803/commits/d0d1f81c28e1f7fdc359bbab9f1dc9177093d8dd

PR #1805, opened 2024-11-18
Implement progressVariant API

lucaslyl:
Implement progressVariant API: When 'horizontal' is selected, it displays a traditional linear progress bar, while 'circle' renders a circular progress indicator

Uploading Screen Recording 2024-11-18 at 14.23.32.mov…

### DIFF
--- a/packages/boxel-ui/addon/src/components/progress-bar/index.gts
+++ b/packages/boxel-ui/addon/src/components/progress-bar/index.gts
@@ -11,8 +11,8 @@ interface Signature {
     label?: string;
     max: number;
     position?: BoxelProgressBarPosition;
-    progressVariant?: 'horizontal' | 'circle';
     value: number;
+    variant?: 'horizontal' | 'circular';
   };
   Element: HTMLDivElement;
 }
@@ -25,12 +25,7 @@ export default class ProgressBar extends Component<Signature> {
   }
 
   get progressWidth(): ReturnType<typeof htmlSafe> {
-    const max = this.args.max ?? 100;
-    const value = this.args.value ?? 0;
-    const percentage =
-      Math.round(Math.min(Math.max((value / max) * 100, 0), 100)) + '%';
-
-    return htmlSafe(`width: ${percentage};`);
+    return htmlSafe(`width: ${this.progressPercentage};`);
   }
 
   get progressBarPosition() {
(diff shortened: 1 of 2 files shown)


packages/boxel-ui/addon/src/components/progress-bar/index.gts
```diff
@@ -1,24 +1,36 @@
...
+import { cssVar } from '../../helpers.ts';
+
 export type BoxelProgressBarPosition = 'start' | 'center' | 'end';
 
 interface Signature {
   Args: {
     label?: string;
     max: number;
     position?: BoxelProgressBarPosition;
+    progressVariant?: 'horizontal' | 'circle';
```
tintinthong: can u name this variant

packages/boxel-ui/addon/src/components/progress-bar/index.gts
```diff
@@ -1,24 +1,36 @@
...
+  }
+
   get progressWidth(): ReturnType<typeof htmlSafe> {
     const max = this.args.max ?? 100;
     const value = this.args.value ?? 0;
-    const percentage = Math.min(Math.max((value / max) * 100, 0), 100);
-    return htmlSafe(`width: ${percentage}%`);
+    const percentage =
+      Math.round(Math.min(Math.max((value / max) * 100, 0), 100)) + '%';
+
```
tintinthong: progressPercentage already gives you the percentage

packages/boxel-ui/addon/src/components/progress-bar/index.gts
```diff
@@ -30,26 +42,37 @@ export default class ProgressBar extends Component<Signature> {
...
-      ...attributes
-    >
-      <div class='progress-bar'>
-        <div class='progress-bar-value' style={{this.progressWidth}}>
-          <div class='progress-bar-info {{this.progressBarPosition}}'>
-            <div class='progress-bar-label'>
-              {{#if @label}}
-                {{@label}}
-              {{/if}}
+    {{#if (eq @progressVariant 'circle')}}
```
tintinthong: maybe have an outer class that has aria-label that is shared between the 2 variants

packages/boxel-ui/addon/src/components/progress-bar/index.gts
```diff
@@ -118,6 +141,44 @@ export default class ProgressBar extends Component<Signature> {
...
+          );
+          --progress-bar-fill-color: var(
+            --boxel-progress-bar-fill-color,
+            var(--boxel-highlight)
+          );
+          --progress-bar-font-color: var(
+            --boxel-progress-bar-font-color,
+            var(--boxel-light)
+          );
+          --progress-bar-size: 80px;
```
tintinthong: these are called progress-bar not progess-circle. For vars only related to circle then call it progress-circle. For other vars that may be shared, then you can move it to the outer-class I was talking about above

lucaslyl: Thank for advise!

richardhjtan (approved):

tintinthong (approved):

lukemelia (commented):
Why choose to make this one component rather than two separate components?

lucaslyl:
> Why choose to make this one component rather than two separate components?

I chose to create a single progress bar component with variants (horizontal and circular) because it allows us to share core functionality, like handling the percentage progress, shared CSSvar for color and etc while simply triggering the variant to change the UI's appearance. This approach makes the code more maintainable, as updates and improvements to the underlying logic only need to be made once, benefiting both variants. It's also more efficient for future expansions, as we can easily add new variants without duplicating Logic. 


However, I agree if we decide to switch to two separate components, as both approaches have their own pros. we can refactor it into two separate components if needed.

PR #1807, opened 2024-11-18
Handle Stripe subscription expired webhook

FadhlanR:
I added an extra step in the subscription deletion handler to subscribe the user back to the free plan if the subscription status is `expired`. To achieve this, we need to call two Stripe endpoints: one to fetch the price by product code and another to create the subscription. In the test, I mock these endpoints using `sinon`. However, I have also called these endpoints manually to verify the mock responses and ensure the flow works correctly.


packages/billing/stripe-webhook-handlers/stripe.ts
```diff
@@ -0,0 +1,14 @@
+import Stripe from 'stripe';
+
+let stripe: Stripe;
+export function getStripe() {
```
FadhlanR, line 4: I need to create this function for test purpose, so I can mock the stripe instance in the test.

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
```diff
@@ -71,9 +73,43 @@ export async function handleSubscriptionDeleted(
...
     if (newStatus === 'expired') {
-      // TODO: Put the user back on the free plan (by calling Stripe API). Will be handled in CS-7466
+      await subcribeUserToFreePlan(dbAdapter, event.data.object.customer);
     }
 
     await markStripeEventAsProcessed(dbAdapter, event.id);
   });
 }
+
+async function subcribeUserToFreePlan(
```
jurgenwerk:
```suggestion
async function subscribeUserToFreePlan(
```

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
```diff
@@ -71,9 +73,43 @@ export async function handleSubscriptionDeleted(
...
+  dbAdapter: DBAdapter,
+  stripeCustomerId: string,
+) {
+  try {
+    let stripe = getStripe();
+    let freePlan = await getPlanByMonthlyPrice(dbAdapter, 0);
+    if (!freePlan) {
+      console.error('Free plan is not found');
```
jurgenwerk: Throw an error, it's a critical error, free plan must be present in the plans table.

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
```diff
@@ -71,9 +73,43 @@ export async function handleSubscriptionDeleted(
...
+    let prices = await stripe.prices.list({
+      product: freePlan.stripePlanId,
+      active: true,
+    });
+    if (!prices.data[0]) {
+      console.error('No price found for free plan');
+      return;
+    }
+
+    await stripe.subscriptions.create({
```
jurgenwerk, line 104: Does this charge the user (with 0 amount) immediately? What happens if the user's payment method is invalid? What are the possible errors here, returned by Stripe?

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
```diff
@@ -71,9 +73,43 @@ export async function handleSubscriptionDeleted(
...
+    if (!freePlan) {
+      console.error('Free plan is not found');
+      return;
+    }
+    let prices = await stripe.prices.list({
+      product: freePlan.stripePlanId,
+      active: true,
+    });
+    if (!prices.data[0]) {
+      console.error('No price found for free plan');
```
jurgenwerk: Throw an error too, Stripe is misconfigured when the price for a plan cannot be found

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
```diff
@@ -71,9 +73,43 @@ export async function handleSubscriptionDeleted(
...
+    }
+
+    await stripe.subscriptions.create({
+      customer: stripeCustomerId,
+      items: [
+        {
+          price: prices.data[0].id,
+        },
+      ],
+      payment_behavior: 'error_if_incomplete',
```
jurgenwerk, line 111: Add a comment why this behaviour is needed. Please also add more comments on how this works after a subscription is created, i.e. describing that the the invoice.payment_succeeded webhook will take care of making a new subscription, add plan allowance etc.

jurgenwerk: Have you tested this locally using Stripe's sandbox? Does the subscription get created successfully?

packages/billing/stripe-webhook-handlers/stripe.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (!process.env.STRIPE_WEBHOOK_SECRET) {
```
jurgenwerk: Stripe webhook secret is intended for verifying that the webhook came from Stripe. For talking to Stripe's API you need the api key not the webhook secret. Use STRIPE_API_KEY

jurgenwerk (changes requested):

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
FadhlanR: Yes, this charge user immediately.

Screenshot 2024-11-18 at 20 10 14

jurgenwerk: Did you confirm the subsequent invoice.payment_succeeded webhook activated the free plan?

FadhlanR: > Does this charge the user (with 0 amount) immediately? What happens if the user's payment method is invalid? What are the possible errors here, returned by Stripe?

I tested using my stripe project with this scenario, I created a customer without payment method and I triggered subscription create endpoint for this user. It's fine to subscribe to free plan even the customer haven't configured the payment method. However, for creator plan user needs to have payment method configured.

FadhlanR: > Did you confirm the subsequent invoice.payment_succeeded webhook activated the free plan?

Yes it activated the free plan.
Screenshot 2024-11-18 at 20 19 03

Screenshot 2024-11-18 at 20 33 45

packages/billing/stripe-webhook-handlers/stripe.ts
FadhlanR: Updated

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
jurgenwerk: I mean in the boxel app - did the `invoide.payment_succeeded` webhook create a new free plan `subscription`, `subscription_cycle`, added free plan allowance?

FadhlanR: I just tested it, and it worked. One question: don’t we need to ensure the user is always subscribed to a plan? If so, why don’t we subscribe the user to a free plan when their subscription is canceled, not just when it expires?

jurgenwerk: > I just tested it, and it worked

👍 

> why don’t we subscribe the user to a free plan when their subscription is canceled

Because even on the stripe free plan, users will keep receiving the monthly $0 invoice. So if they cancel we must respect their decision to not be subscribed anymore.

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
FadhlanR: Added comments [b4a1015](https://github.com/cardstack/boxel/pull/1807/commits/b4a1015620dc7040bdee17f5c6d1b51fee89ca06)

packages/billing/stripe-webhook-handlers/subscription-deleted.ts
```diff
@@ -71,9 +73,44 @@ export async function handleSubscriptionDeleted(
...
+    });
+  } catch (err) {
+    console.error(`Failed to subscribe user back to free plan, error:`, err);
```
jurgenwerk: Is there a reason this is an error log as opposed to throwing an error? Which are the cases where `stripe.subscriptions.create` can fail, apart from network/server errors?

FadhlanR: I think we should allow the subscription expiration webhook to succeed, even if we fail to subscribe the user to the free plan.

jurgenwerk: Yeah I'm fine with that logic. Just make sure this gets reported to sentry.

jurgenwerk (approved):

lukemelia (approved):

PR #1809, opened 2024-11-18
Track credit consumption in AI bot

jurgenwerk:
This PR adds a mechanism to AI bot which tracks the AI-related cost, where it fetches the dollar cost from OpenRouter after ai bot streaming is done, converts to credits and writes it to the credit ledger. 

Below is an example how the credits ledger looks like after a couple of interactions with the AI bot. You can see that it first tries to spend the plan's allowance credits, and when that's gone it spends the extra credits:

image


jurgenwerk:
As an interesting observation, when users simply say "hello" to the bot, the response will cost them around 1-3 credits, sometimes up to 5. 

For editing cards, and for longer responses, this cost can go up to 10-15 or more. So it looks like free plan users (100 credits) will be able to only have a couple interactions with the free plan's allowance.

packages/ai-bot/lib/ai-cost.ts
```diff
@@ -0,0 +1,75 @@
...
+  generationId: string,
+) {
+  try {
+    // Generation data is sometimes not immediately available, so we retry a couple of times until we are able to get the cost
+    let costInUsd = await retry(() => fetchGenerationCost(generationId), {
+      retries: 10,
+      delayMs: 500,
+    });
+
+    let creditsConsumed = Math.round(costInUsd / 0.001);
```
lukemelia: Make 0.001 a named constant in module scope. CREDITS_PER_USD or something like that?

packages/billing/billing-queries.ts
```diff
@@ -472,3 +472,67 @@ export async function expireRemainingPlanAllowanceInSubscriptionCycle(
...
+
+    if (extraCreditsToSpend > 0) {
+      await addToCreditsLedger(dbAdapter, {
+        userId,
+        creditAmount: -extraCreditsToSpend,
+        creditType: 'extra_credit_used',
+        subscriptionCycleId: subscriptionCycle.id,
+      });
+    }
+  }
```
lukemelia, line 537: I wonder if it would be worth tracking the generation ID.

lukemelia (approved):

packages/billing/billing-queries.ts
jurgenwerk: I think now with this simple variant this may not be necessary but later on when we introduce jobs and queues, we could save the generation id there (in the serialized job) and have it available for inspection if needed.

packages/ai-bot/lib/ai-cost.ts
jurgenwerk: Thanks, done

PR #1815, opened 2024-11-19
Implement filter display UI of work tracker based on chris design 

lucaslyl:
linear: eco-93-display-filter-of-work-tracker-onselect

Screenshot 2024-11-19 at 17 29 41

### DIFF
--- a/packages/experiments-realm/app-task-card.gts
+++ b/packages/experiments-realm/app-task-card.gts
@@ -30,9 +30,9 @@ import { FilterTrigger } from './productivity/filter-trigger';
 import getTaskCardsResource from './productivity/task-cards-resource';
 import { FilterDisplay } from './productivity/filter-display';
 import Checklist from '@cardstack/boxel-icons/checklist';
-import Subtask from '@cardstack/boxel-icons/subtask';
-import UserRound from '@cardstack/boxel-icons/user-round';
-import FileCog from '@cardstack/boxel-icons/file-cog';
+import RectangleEllipsis from '@cardstack/boxel-icons/rectangle-ellipsis';
+import User from '@cardstack/boxel-icons/user';
+import FolderGit from '@cardstack/boxel-icons/folder-git';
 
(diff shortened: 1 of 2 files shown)


packages/experiments-realm/app-task-card.gts
```diff
@@ -349,59 +353,78 @@ class AppTaskCardIsolated extends Component<typeof AppCard> {
...
-          <div class='filter-display'>
+                {{item}}
+              </BoxelSelect>
+            {{/if}}
+          </div>
+
+          <div class='filter-display-sec'>
             {{#each this.filterTypes as |filterType|}}
+              {{log filterType}}
               {{#let (this.selectedFilterItems filterType) as |items|}}
```
tintinthong: Remove log

packages/experiments-realm/app-task-card.gts
```diff
@@ -349,59 +353,78 @@ class AppTaskCardIsolated extends Component<typeof AppCard> {
     return this.selectedItems.get(filterType) ?? [];
   }
 
+  @action getFilterIcon(filterType: FilterType) {
+    switch (filterType) {
```
tintinthong, line 357: there is already a static configuration for all the filters. Let's keep all data in one place

tintinthong:
Can u put the value for the filter value like Current Sprint slightly lighter

lucaslyl:
> Can u put the value for the filter value like Current Sprint slightly lighter

Sure thank for the advise, but I had to tweak it a bit. I made the filter value more prominent so it stands out more than the key in the UI.

tintinthong:
We have to make a decision here either to follow chris design and introduce those components in boxel-ui or perhaps these icons which I think match what he previously shared

```
import RectangleEllipsis from '@cardstack/boxel-icons/rectangle-ellipsis';
import User from '@cardstack/boxel-icons/user';
import FolderGit from '@cardstack/boxel-icons/folder-git';
```

I think lets use this for now

PR #1816, opened 2024-11-19
Capture last known good state in index when creating error doc

habdelra:
This PR captures the last known good state of an index entry within the index when we create an error doc for it.


packages/runtime-common/index-writer.ts
```diff
@@ -142,62 +143,72 @@ export class Batch {
...
+      // we do this because pg automatically casts big ints into strings, so
+      // we unwind that to accurately type the structure that we want to pass
+      // _in_ to the DB
+      last_modified: number;
+      indexed_at: number;
+    };
+
+    if (entry.type === 'error') {
+      // merge the last known good deps with the error deps so we can invalidate
+      // when upstream issue is repaired
```
lukemelia, line 198: 👍

packages/runtime-common/index-writer.ts
```diff
@@ -222,6 +233,40 @@ export class Batch {
...
+        ]),
+        realmVersionExpression(),
+      ]),
+    ] as Expression)) as unknown as BoxelIndexTable[];
+    if (!entry) {
+      return undefined;
+    }
+
+    let {
+      indexed_at: _remove1,
```
lukemelia, line 255: What is _removeN and where do they come from?

lukemelia (approved):

packages/runtime-common/index-writer.ts
habdelra: it's a simple way to use object destructuring to delete properties from an object. This way it we don't mutate a type by deleting a property. TS can derive the type of `productVersion` without me having to define an intermediate type because I deleted properties off of it.

PR #1819, opened 2024-11-20
Blog App cards grid fitted views

burieberry:
- Currently card view is displaying 'Publish Date' and 'Status' fields along with the fitted card

card-view
strip-view
grid-view


lukemelia:
Can you share a screenshot or video?

PR #1825, opened 2024-11-20
Hide sort options and admin data for author and category filters

burieberry:
Since we have decided that Author and Category types won't have PublishDate and Status fields, hiding the Sort Menu and admin data for these types on Blog App. They will be displayed in alphabetical order of card title.

Author type:
author

Blog Post type (sort and info not hidden):
blog-post

### DIFF
--- a/packages/experiments-realm/blog-app.gts
+++ b/packages/experiments-realm/blog-app.gts
@@ -43,11 +43,22 @@ import AuthorIcon from '@cardstack/boxel-icons/square-user';
 import type { BlogPost } from './blog-post';
 
 type ViewOption = 'card' | 'strip' | 'grid';
+type Sort = Query['sort'];
 
 interface SortOption {
   displayName: string;
-  sort: Query['sort'];
+  sort: Sort;
 }
+const sortByCardTitle: Sort = [
+  {
+    on: {
+      module: `${baseRealm.url}card-api`,
+      name: 'CardDef',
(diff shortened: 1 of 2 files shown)


packages/experiments-realm/blog-app.gts
```diff
@@ -90,26 +92,31 @@ interface SidebarFilter {
...
     cardTypeName: 'Blog Post',
     createNewButtonText: 'Post',
+    showAdminData: true,
   },
   {
     displayName: 'Author Bios',
     icon: AuthorIcon,
     cardTypeName: 'Author Bio',
     createNewButtonText: 'Author',
+    isSortDisabled: true,
```
lukemelia: Instead of `isSortDisabled`, maybe we should have a `sortOptions` prop here. If provided, sort is rendered with those options. If not, no sort UI is rendered.

lukemelia (approved):

packages/experiments-realm/blog-app.gts
burieberry: That's what I initially did, but that implies sortOptions might be different per filter and we have to reset or change selectedSort as the filter changes so as not to do a halfway-done implementation. I can change back to it though.

lukemelia: Your call -- just seemed duplicative to me, but I understand your point.

PR #1777, opened 2024-11-08
Add payment setup flow upon registration

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7396/send-user-to-stripe-checkout-for-free-plan-during-registration-flow

### What is changing
- Add payment setup flow right after the user signs up and verifies it successfully.
- If the user skips this step, prompt them to set up a payment method until it's completed.

Note:
- I have tweaked the overall basic font size (16px to 13px) as per design when translating into the app. I found that if we apply the font size to 16px, users will need to scroll more when they have a smaller screen size, e.g., a 13-inch MacBook.
- There is a discussion to apply this similar flow after the user registered but has not subscribed yet (https://github.com/cardstack/boxel/pull/1756#discussion_r1832966855), which is included in this PR.
- Account popover is shown at payment setup, so user able see their account and option to logout

### How to test 
There is some pre-setup needed to enable free plan on development account:

1. Use stripe cli to listen for the webhooks that Stripe sends to the realm server

```
stripe listen --forward-to localhost:4201/_stripe-webhook --api-key sk_test_api_key_from_the_sandbox_account
```

2. You will get webhook signing secret from stripe cli after Step 1 is done

```
> Ready! You are using Stripe API Version [x]. Your webhook signing secret is whsec_xxxxx
```

3. Go to `packages/realm-server`, pass STRIPE_WEBHOOK_SECRET & STRIPE_API_KEY environment value and start the realm server

```
STRIPE_WEBHOOK_SECRET=... STRIPE_API_KEY=... pnpm start:all
```

4. Go to `packages/realm-server` and run the following command to sync the stripe products to the database, make sure you have the stripe api key set

```
STRIPE_API_KEY=... pnpm sync-stripe-products
```

5. Perform "Setup up Secure Payment Method" flow. Subscribe with valid test card [here](https://docs.stripe.com/testing#cards)

You should be able to subscribe successfully after you perform the steps above.

### Screenshots
**Video**

https://github.com/user-attachments/assets/9317790b-0b2c-4b42-9b8a-8b5ec6d35513

**Monitor size**

Screenshot 2024-11-21 at 5 36 37 PM

**Laptop Size**

Screenshot 2024-11-21 at 5 38 17 PM

PR #1826, opened 2024-11-21
Check if user has enough credits before responding with AI generated …

jurgenwerk:
When user only has 10 credits left and tries to use the AI bot, an error message is shown:

image

For now, this error uses the standard error format, but we are planning to add a special treatment for "out of credit" error message in the bot, with included buttons to upgrade or add credits (this will be a separate task). 

Important: this should be merged after https://github.com/cardstack/boxel/pull/1777, which is a PR that makes sure users have a subscription.


packages/ai-bot/lib/ai-billing.ts
```diff
@@ -58,6 +59,25 @@ export async function saveUsageCost(
...
+
+  if (!user) {
+    throw new Error(
+      `should not happen: user with matrix id ${matrixUserId} not found in the users table`,
+    );
+  }
+
+  let availableCredits = await sumUpCreditsLedger(pgAdapter, {
+    userId: user.id,
+  });
```
lukemelia, line 78: In the interest of efficiency, should we consider collapsing the above queries into a single query with a join?

packages/ai-bot/main.ts
```diff
@@ -214,11 +220,17 @@ Common issues are:
...
 
-        if (historyError) {
-          responder.finalize(
-            'There was an error processing chat history. Please open another session.',
+        let availableCredits = await getAvailableCredits(
+          assistant.pgAdapter,
+          senderMatrixUserId,
+        );
+
+        let minimumCredits = 10;
```
lukemelia: Make this a constant in module scope.

lukemelia (approved):

FadhlanR (approved):

habdelra (approved):

PR #1828, opened 2024-11-21
Avoid second instance of RoomResource for same room

lukemelia:


lukemelia:
Breaking this up into smaller PRs

PR #1833, opened 2024-11-22
Issues with updating of cards

tintinthong:
After some research I have narrowed it down to 2 issues tht both can confound each other but should be treated separately

1. when u load a card resource (eg open a stack item), then you somehow mutate that card and then load the card resource again (eg opening a stack item again), the card resource will show the stale value of the card.

Question: do we expect loading a card resource (eg opening a stack item) to always show the newest value from server?

2. when u send a save (not via auto-save), the card resource which exists does not update. Here is the [code](https://github.com/cardstack/boxel/blob/32e2a594893997fc54bde302930466a361084b33/packages/host/app/resources/card-resource.ts#L230) which breaks this 

Question: do we expect card resources to always capture changes from sse events? (eg seeing a stack item or title in recent cards). Presumably not while were typing


tintinthong:
Made this change to describe the test for 2. better https://github.com/cardstack/boxel/pull/1833/commits/6a3cdc9c7a8db43a32339f200a4cfa56dd524242

I notice ive been constantly saying "sse is not working" but that is not true. Sse is very much working, just the subscription of the card-resource is ignoring it. Excuse ma english

ef4:
These changes are enough to fix the specific case in @tintinthong's test. 

They ensure that cards loaded via relationships are in the shared liveCards identity map.

But there are still other cases where a card could be loaded not-via-card-resource, mutated, and then invalidations will be ignored by another copy of the same card instance that *is* in the card-resource's liveCards identity map:

 - getCards
 - getCard with isLive===false

PR #1834, opened 2024-11-22
Add last known good rendering of card in error state to stack item

habdelra:
This PR adds the last known good rendering of a card in an error state for a stack item. In the case there is no last known good HTML, we'll show a generic message.

Note that this does not yet support live updates into and out of an error state. that will come in a subsequent PR.

error

This is the view for a card without a last known good state:

Screenshot from 2024-11-27 13-14-37


packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -341,6 +341,11 @@ export default class InteractSubmode extends Component<Signature> {
...
 
   private close = task(async (item: StackItem) => {
+    if (item.cardError) {
+      this.operatorModeStateService.trimItemsFromStack(item);
+      return;
+    }
+
     let { card, request } = item;
     // close the item first so user doesn't have to wait for the save to complete
     this.operatorModeStateService.trimItemsFromStack(item);
```
lukemelia:
```suggestion
    // close the item first so user doesn't have to wait for the save to complete
    this.operatorModeStateService.trimItemsFromStack(item);
    if (item.cardError) {
      return;
    }

    let { card, request } = item;
```

packages/host/app/resources/stack-backgrounds.ts
```diff
@@ -46,6 +46,9 @@ export class StackBackgroundsResource extends Resource<Args> {
         }
         let bottomMostStackItem = stack[0];
         await bottomMostStackItem.ready;
+        if (bottomMostStackItem.cardError) {
+          return undefined;
```
lukemelia: Is it impossible to know what realm an error doc is from?

lukemelia (approved):

packages/host/app/resources/stack-backgrounds.ts
habdelra: Hmm, I’ll look into it. Presumably some realm is returning the error response. Perhaps it’s a header? If so we’ll have to refactor how the stack item calculates the realm, as right now I does so from a running card.

lukemelia: Maybe it makes sense for the error doc to carry the realm?

habdelra: Good idea

habdelra: addressed in https://github.com/cardstack/boxel/pull/1834/commits/64e950c562812baca7c245484cb49a506f2f7fc0

lukemelia (approved):

PR #1839, opened 2024-11-25
Isolated BlogPost template and add markdown content styles

burieberry:
- Added available markdown field styles
- Styled isolated blog-post template with custom fonts
- Added `blog` linksTo field to blog-post card
- Added `website` field to blog-app and its fitted template as displayed on blog post (note: will need other fitted sizes)
- Added catalog-entry for BlogApp
- **Removed forced font styles on atom template. I haven't noticed any failures because we usually use the Pill component and it already has that styling.**

A couple of screenshots:

blog-post-1
blog-post-2
blog-post-3


packages/experiments-realm/blog-post.gts
```diff
@@ -588,36 +595,165 @@ export class BlogPost extends CardDef {
...
+                  {{this.formattedDatePublished}}
+                </time>
+              </li>
+            {{/if}}
+          </ul>
         </header>
         <@fields.body />
       </article>
       <style scoped>
+        @import url('https://fonts.googleapis.com/css2?family=Lora:ital,wght@0,400..700;1,400..700&family=Playfair+Display:ital,wght@0,400..900;1,400..900&display=swap');
```
backspace, line 668: It’s fine of course but this being repeated across the different views does make me wish we could share things more easily.

packages/base/markdown.gts
```diff
@@ -14,13 +14,203 @@ class View extends Component<typeof MarkdownField> {
...
+      }
+
+      /* Definition List */
+      /* Must use <dl> <dt> <dd> tags -- default browser styling */
+
+      /* Footnote */
+      /* Not available */
+
+      /* Emoji */
+      /* Must copy/paste emoji */
```
backspace, line 201: What are these empty sections?

backspace (approved):

packages/experiments-realm/blog-post.gts
burieberry: Yeah, I was wondering about that

packages/base/markdown.gts
burieberry: Some markdown editors allow more syntax (https://www.markdownguide.org/cheat-sheet/) for stuff like definition lists, subscript/superscript, emojis, etc. Ours doesn't, but for most it's possible to use html inside the markdown field for it so I made notes.

PR #1842, opened 2024-11-26
Skip animation on edit and finish editing to isolated

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7452/should-not-play-stack-animation-when-transitioning-between-edit-mode

### What is changing
Skip the animation in edit mode and return to isolated mode after editing.

### Screenshot
https://github.com/user-attachments/assets/c154f99c-3cd2-4a1c-962e-f9699ec2103d


lukemelia (commented):
Is there a way we could keep animation logic/flags out of the model layer? Ideally, the StackItem model would contain enough semantic state to allow the component layer (css and/or js) to make animation decisions as needed.

richardhjtan:
> Is there a way we could keep animation logic/flags out of the model layer? Ideally, the StackItem model would contain enough semantic state to allow the component layer (css and/or js) to make animation decisions as needed.

Yes @lukemelia, there is a way to do that. I have simplified the implementation with the latest changes, thanks.

lukemelia (commented):
Seems better. Anyway to have test coverage for this? (There may not be, just wondering.)

burieberry (approved):

richardhjtan:
I added a integration test to ensure that the opening animation's class name does not exist, since the animation is triggered by class name

PR #1843, opened 2024-11-26
Get Apply button state working for autoExecute commands

lukemelia:
Accomplished via a big refactor of how we load matrix events in the RoomResource


packages/host/app/services/command-service.ts
```diff
@@ -100,11 +117,13 @@ export default class CommandService extends Service {
   }
 
   //TODO: Convert to non-EC async method after fixing CS-6987
-  run = task(async (command: CommandCard, roomId: string) => {
+  run = task(async (command: MessageCommand, roomId: string) => {
```
habdelra, line 120: this being a task really doesnt buy us anything since the service lives forever (the async never needs to be cancelled)

packages/host/app/services/command-service.ts
```diff
@@ -100,11 +117,13 @@ export default class CommandService extends Service {
...
     let { payload, eventId } = command;
     let res: any;
     try {
       this.matrixService.failedCommandState.delete(eventId);
+      this.currentlyExecutingCommandEventIds.add(eventId);
+      await timeout(DELAY_FOR_APPLYING_UI); // leave a beat for the "applying" state of the UI to be shown
```
habdelra: probably a better way to do this is not to serially add to the async, but to have the timeout happen concurrently with the rest of the async in a Promise.all (or Promise.settled). that way if the rest of the async already takes as long as `DELAY_FOR_APPLYING_UI`, there is no need to wait any more.

habdelra (approved):
looks good just a few very minor comments

packages/host/app/services/command-service.ts
lukemelia: Does the caveat about CS-6987 still apply?

lukemelia: Actually, looks like we are using derived state from the task in room.gts.

PR #1844, opened 2024-11-26
tools: Fix creation of JS, JSON, TS files

backspace:
The current release errors when you try to create various types of file:

Untitled (Workspace) 2024-12-02 13-43-56

Now it works:

https://github.com/user-attachments/assets/31acafb0-8339-403f-bda6-cc398614991d


packages/vscode-boxel-tools/src/file-system-provider.ts
```diff
@@ -380,15 +378,7 @@ export class RealmFS implements vscode.FileSystemProvider {
       }
       console.log('Response:', response);
 
-      const contentType = response.headers.get('Content-Type');
-      let content: Uint8Array;
-
-      if (contentType && contentType.includes('application/json')) {
-        const jsonData = await response.json();
-        content = new TextEncoder().encode(JSON.stringify(jsonData, null, 2));
```
backspace, line 388: I removed this because it was pretty-printing JSON files instead of displaying their extra contents, so a file with `[1,2,3]` rendered as

```
[
  1,
  2,
  3
]
```

packages/runtime-common/realm.ts
```diff
@@ -1108,9 +1108,6 @@ export class Realm {
     requestContext: RequestContext,
   ): Response {
     try {
-      if (content.match(/^\s*$/)) {
-        throw new Error('File is empty');
-      }
```
backspace, line 1113: An empty file for `.ts` and the like isn’t invalid, so empty file detection is moved to the client.

packages/runtime-common/index-query-engine.ts
```diff
@@ -210,7 +210,7 @@ export class IndexQueryEngine {
       last_modified: lastModified,
       resource_created_at: resourceCreatedAt,
     } = moduleEntry;
-    if (!executableCode) {
+    if (executableCode === null) {
```
backspace, line 213: `''` shouldn’t produce an error.

lukemelia (approved):
I'd like to make sure @habdelra sees this, but LGTM

packages/runtime-common/realm.ts
habdelra: hmmm, i think an empty file is not technically a TS file. in order to be a TS at the very least it should export nothing. like: 
```
export {};
```
perhaps for executable files we always add at least this on file create? this can then be override as the user modifies the file

packages/vscode-boxel-tools/src/file-system-provider.ts
```diff
@@ -359,15 +360,12 @@ export class RealmFS implements vscode.FileSystemProvider {
...
         Authorization: `${await this.realmAuth.getJWT(apiUrl)}`,
       };
 
-      // Add special header for .gts files
-      if (uri.path.endsWith('.gts')) {
-        headers['Accept'] = 'application/vnd.card+source';
-      } else if (uri.path.endsWith('.json')) {
-        headers['Accept'] = 'application/vnd.card+json';
+      if (hasExecutableExtension(uri.path)) {
+        headers.Accept = SupportedMimeType.CardSource;
```
habdelra: yes!

packages/vscode-boxel-tools/src/file-system-provider.ts
```diff
@@ -359,15 +360,12 @@ export class RealmFS implements vscode.FileSystemProvider {
 
     try {
       let headers = {
-        Accept: 'application/vnd.api+json',
+        Accept: SupportedMimeType.JSONAPI,
```
habdelra: I don't understand why we set this here--shouldn't we always be using card-source header regardless--just like the monaco editor does in the host app?

packages/vscode-boxel-tools/src/file-system-provider.ts
```diff
@@ -359,15 +360,12 @@ export class RealmFS implements vscode.FileSystemProvider {
...
         Authorization: `${await this.realmAuth.getJWT(apiUrl)}`,
       };
 
-      // Add special header for .gts files
-      if (uri.path.endsWith('.gts')) {
-        headers['Accept'] = 'application/vnd.card+source';
-      } else if (uri.path.endsWith('.json')) {
-        headers['Accept'] = 'application/vnd.card+json';
+      if (hasExecutableExtension(uri.path)) {
```
habdelra: shouldn't card-source accept header be used for all file types, like the monaco editor in the host app?

packages/runtime-common/index-query-engine.ts
habdelra: really? because that is not actually executable js. one item I learned from @ef4 if that if you want a no-op ES6 module, then have the contents of the file be:
```js
export {};
```

ef4: I may have overstated that. In a purely ES module situation, an empty file is actually OK. But whenever there is CJS interoperability going on it can cause problems.

habdelra: would the babel transpilation step that we are doing in our realm server work on an empty file? I guess if it doesn't, then we can just make it work as well as part of this PR...

packages/vscode-boxel-tools/src/file-system-provider.ts
backspace: I don’t know why it was like this, I was just changing what worked to the extracted variable. But I changed to `CardSource` and it worked, thanks.

packages/runtime-common/index-query-engine.ts
backspace: This is the row from an empty file in `boxel_index`:

```
	{
		"url": "http://localhost:4201/2024-11-29-11-57/personal/none.js",
		"file_alias": "http://localhost:4201/2024-11-29-11-57/personal/none",
		"type": "module",
		"realm_version": 42,
		"realm_url": "http://localhost:4201/2024-11-29-11-57/personal/",
		"pristine_doc": null,
		"search_doc": null,
		"error_doc": null,
		"deps": [
		],
		"types": null,
		"isolated_html": null,
		"indexed_at": 1733266647053,
		"is_deleted": false,
		"source": "",
		"transpiled_code": "",
		"last_modified": 1733266646,
		"embedded_html": null,
		"atom_html": null,
		"fitted_html": null,
		"display_names": null,
		"resource_created_at": 1733266646
	}
```

`transpiled_code` being `''` seems correct to me, agreed? When I worked on this I saw that [all-whitespace files are skipped](https://github.com/cardstack/boxel/blob/751e2daefe7317d3936e32726c7d7ccf87b74905/packages/runtime-common/transpile.ts#L29), so I made use of that.

habdelra: Sounds good, then. Let’s use the blank modules then. 👍🏽

PR #1845, opened 2024-11-26
Add support for tracking errors in indexing identity map

habdelra:
This PR addresses a cycle that was discovered in staging. The issue is that the identity map that we use during indexing is blind to error documents--it only recognizes running cards. Because of this, it is possible to introduce a cycle in our indexer using a strategically placed error document. In this PR we provide visibility into the instances that are in an error state as part of our identity map so that we can skip over trying to visit them when we know the instance is in an error state.

I had difficulty isolating the specific logic that triggered the cycle in our test realms that spin up within the each test--but i could absolutely see it in our actual realm. so i just used the actual files that caused the cycle in staging in our node-test realm. When the fix is not working, the realm server will hang from the cycle. the fact that the realm server starts up proves that the issue is fixed.

### DIFF
--- a/packages/host/app/services/render-service.ts
+++ b/packages/host/app/services/render-service.ts
@@ -137,7 +137,7 @@ export default class RenderService extends Service {
         if (isCardError(err) && err.status !== 500 && notLoaded) {
           let links =
             typeof notLoaded.reference === 'string'
-              ? notLoaded.reference
+              ? [notLoaded.reference]
               : notLoaded.reference;
           for (let link of links) {
             if (identityContext.errors.has(link)) {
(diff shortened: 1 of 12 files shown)


packages/host/app/services/render-service.ts
```diff
@@ -135,13 +135,22 @@ export default class RenderService extends Service {
           isNotLoadedError(e),
         ) as NotLoaded | undefined;
         if (isCardError(err) && err.status !== 500 && notLoaded) {
-          let linkURL = new URL(`${notLoaded.reference}.json`);
```
habdelra, line 138: this code was actually hiding a type issue--the unloaded references could be a string or an array of strings for links-to-many relationships. we want to handle both.

lukemelia (approved):
👍

PR #1847, opened 2024-11-27
Cs-7580 contact card embedded format

lucaslyl:
linear: cs-7580-contact-card-embedded-format


tintinthong:
Screenshot 2024-11-28 at 10 26 16
Screenshot 2024-11-28 at 10 25 45
Screenshot 2024-11-28 at 10 22 16

packages/experiments-realm/crm/contact.gts
```diff
@@ -91,6 +108,10 @@ class ViewCompanyCardTemplate extends Component<typeof CompanyCard> {
         font-weight: 300;
         text-decoration: underline;
       }
+      .no-buidling {
```
burieberry: typo in class name

packages/experiments-realm/crm/contact.gts
```diff
@@ -258,18 +272,22 @@ export class PhoneField extends FieldDef {
   };
 }
 
-class FittedTemplate extends Component<typeof Contact> {
+class EmbeddedTemplate extends Component<typeof Contact> {
   <template>
-    <article class='fitted-contact-card'>
+    <article class='embedded-contact-card'>
       <div class='avatar-container'>
```
burieberry: Let's make this a component in `experiments-realm/components` from this avatar group because it is used in multiple places.

packages/experiments-realm/crm/contact.gts
```diff
@@ -352,64 +370,270 @@ class FittedTemplate extends Component<typeof Contact> {
...
+        word-break: break-all;
+        gap: var(--boxel-sp-xxs);
+      }
+      .icon {
+        width: var(--boxel-icon-sm);
+        height: var(--boxel-icon-sm);
+        flex-shrink: 0;
+      }
+      .icon.gray {
+        color: var(--boxel-300);
```
burieberry: the latest color for these icons seems closer to `--boxel-400`

packages/experiments-realm/crm/contact.gts
```diff
@@ -294,7 +312,7 @@ class FittedTemplate extends Component<typeof Contact> {
           </div>
         {{/if}}
 
-        {{#if @model.phoneMobile}}
+        {{#if (formatPhone @model.phoneMobile)}}
           <div class='row primary-phone'>
             <PhoneIcon class='icon gray' />
             <span>{{formatPhone @model.phoneMobile}}</span>
```
burieberry: The formatting and styling for this should be in the view or atom template of this Phone field def.

packages/experiments-realm/crm/contact.gts
```diff
@@ -143,12 +164,14 @@ class ViewSocialLinksTemplate extends Component<typeof SocialLinksField> {
...
+        --boxel-icon-button-width: var(--boxel-icon-lg);
+        --boxel-icon-button-height: var(--boxel-icon-lg);
         border: 1px solid var(--boxel-300);
+        border-radius: 5px;
       }
       .social-link:hover {
-        border-color: var(--boxel-400);
+        border-color: var(--boxel-500);
+        background-color: var(--boxel-100);
         cursor: pointer;
```
burieberry: The styling for social media links should be in URLField field def. The version with just the icon might be the atom template. Confirm it with design team. 
I think this should be:`@field socialLinks = **containsMany**(URLField);` based on the Linear ticket description.

burieberry (commented):
So this looks good, however, I think we should implement some other tickets first. There are linear tickets for `PhoneField`, `EmailField`, `AccountCard` etc etc... I don't know why some of these cards are in Next Sprint column on Linear. The reason is that we're using embedded or atom views of these fields and cards on this card. The formatting and styling goes inside those templates. This is so that we can implement a composable system.

packages/experiments-realm/crm/contact.gts
lucaslyl: Thanks! I handled this in another PR: https://github.com/cardstack/boxel/pull/1870

packages/experiments-realm/crm/contact.gts
lucaslyl: Sure, I will implement both the email field and the phone field in atom format after the pull requests for those two fields are ready

PR #1848, opened 2024-11-27
Fix flashing on loading screen

richardhjtan:
### What is changing
This PR is to fix 
- flashing loading screen when the user has logged in and refresh

**Current**

https://github.com/user-attachments/assets/bea4a4e4-9fe2-4b61-b4b4-36b29575a9da

**After**

https://github.com/user-attachments/assets/fa2e2aef-697a-44ae-a070-ce6bc4c86e07

### DIFF
--- a/packages/host/app/services/billing-service.ts
+++ b/packages/host/app/services/billing-service.ts
@@ -1,7 +1,7 @@
 import Owner from '@ember/owner';
 import Service from '@ember/service';
 import { service } from '@ember/service';
-import { tracked } from '@glimmer/tracking';
+import { tracked, cached } from '@glimmer/tracking';
 
 import { dropTask } from 'ember-concurrency';
 
@@ -60,6 +60,7 @@ export default class BillingService extends Service {
     return `${stripePaymentLink}?client_reference_id=${clientReferenceId}`;
   }
 
+  @cached
   get subscriptionData() {
     return this._subscriptionData;
   }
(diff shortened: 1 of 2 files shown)


packages/host/app/services/billing-service.ts
```diff
@@ -79,7 +80,7 @@ export default class BillingService extends Service {
     await this.fetchSubscriptionDataTask.perform();
   }
 
-  private fetchSubscriptionDataTask = dropTask(async () => {
+  private fetchSubscriptionDataTask = task(async () => {
```
jurgenwerk: Why was this needed? Don't we want to make sure that no multiple tasks are being performed at the same time?

richardhjtan: @jurgenwerk You're right, multiple task calls can occur simultaneously when we listen for the billing notification. I have reverted this one.

habdelra:
I want to call your attention to this issue. It’s probably related to this PR. https://linear.app/cardstack/issue/CS-7582/cleanup-task-cancellation-for-fetchsubscriptiondatatask

richardhjtan:
To be verified again after this [commit](https://github.com/cardstack/boxel/pull/1838/commits/471f02cad21c94fbd6d65b8fe5041631bb48ff1c) is up, mentioned in: https://linear.app/cardstack/issue/CS-7582/cleanup-task-cancellation-for-fetchsubscriptiondatatask#comment-d12537ad

jurgenwerk:
@richardhjtan is this good to merge? The PR you mentioned changed quite a bit, not sure how much it affects this one.

jurgenwerk (approved):

richardhjtan:
Tested with the changes in this [PR](https://github.com/cardstack/boxel/pull/1838/files), it works fine too. Thanks for the heads up @jurgenwerk

PR #1855, opened 2024-11-28
Fix admin data

tintinthong:
**Problem**

When not having admin data ie embedded in crm, the overlay will span the empty space provided for admin data. 

**Solution**

Make admin data as an allocated slot of the grid that places meta/adminData next to the card. there shudn be overlay over the card, otherwise, that wud just reside inside of the embedded template itself

### DIFF
--- a/packages/experiments-realm/components/grid.gts
+++ b/packages/experiments-realm/components/grid.gts
@@ -60,8 +60,10 @@ export class CardsGrid extends GlimmerComponent<CardsGridSignature> {
                     <card.component />
                   </CardContainer>
                 </div>
-                {{#if (has-block 'meta')}}
-                  {{yield card to='meta'}}
+                {{#if (eq @selectedView 'card')}}
+                  {{#if (has-block 'meta')}}
+                    {{yield card to='meta'}}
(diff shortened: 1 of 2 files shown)


packages/experiments-realm/components/grid.gts
```diff
@@ -44,22 +44,27 @@ export class CardsGrid extends GlimmerComponent<CardsGridSignature> {
...
-                  class='card'
-                  @displayBoundaries={{not (eq @selectedView 'card')}}
+              <li class='{{@selectedView}}-view-container'>
+                <div
+                  {{@context.cardComponentModifier
+                    cardId=card.url
+                    format='data'
+                    fieldType=undefined
+                    fieldName=undefined
+                  }}
```
burieberry, line 54: You can probably just put this modifier on the CardContainer below and not need the extra div. Otherwise the changes you made look good.

tintinthong: Yea thanks. I was worried of previous issues with the overlay but I think having an extra. div will cause more confusion. I dont see any issues atm

PR #1858, opened 2024-11-29
Cs 7588 create account isolated

lucaslyl:
1. Extract the company card and redefine its schema.
2. Create an account schema.
3. Add an account tab to the CRM app.

Take note: Some of the data is mocked since we haven’t finalized the schema yet. 
This is the first draft, and it is currently focused entirely on the layout.


burieberry (dismissed):
I know the order in linear seems wrong, but I think you should first implement CS-7414 "Create company card" ticket. That ticket has the fields you need to use. If the fields do not match, we should discuss. However, to be able to implement that ticket, you'll first need to implement any missing fields. It would be much easier to do code review, if these are implemented one by one instead of all at once. Also, make sure to research the best practices and semantic html for each template.

tintinthong (approved):

PR #1863, opened 2024-12-02
Create URL field

tintinthong:
Creates URL field 

**This displays the atom in a card**

Screenshot 2024-12-02 at 20 53 38

**Edit mode**

Screenshot 2024-12-02 at 20 57 37

**Invalid URL**

Screenshot 2024-12-02 at 20 57 28

**Valid URL**

Screenshot 2024-12-02 at 21 00 32

**Question**

Shud I put this in base?

### DIFF
--- a/packages/experiments-realm/url.gts
+++ b/packages/experiments-realm/url.gts
@@ -15,12 +15,19 @@ export class UrlField extends StringField {
   static displayName = 'Url';
 
   static edit = class Edit extends Component<typeof UrlField> {
+    get isValidUrl() {
+      if (!this.args.model) {
+        return false;
+      }
+      return isValidUrl(this.args.model);
+    }
     <template>
       <BoxelInput
         type='url'
         value={{@model}}
         @onInput={{@set}}
         @disabled={{not @canEdit}}
+        @state={{if this.isValidUrl 'valid' 'invalid'}}
       />
     </template>
   };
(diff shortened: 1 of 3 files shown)


packages/experiments-realm/url.gts
```diff
@@ -0,0 +1,66 @@
...
+  };
+}
+
+function isValidUrl(urlString: string): boolean {
+  try {
+    new URL(urlString);
+    return true;
+  } catch (err) {
+    return false;
+  }
```
backspace, line 60: `URL.canParse` [does exist](https://caniuse.com/mdn-api_url_canparse_static) for this use.

backspace (approved):
This seems like something that could be in `boxel-ui`, was there any discussion of that possibility?

lukemelia (approved):
@tintinthong use canParse per Buck's suggestion. Let's discuss whether this should be promoted to boxel-ui on Thursday

burieberry (commented):
Looks good with the improvement Buck mentioned. A note on link styling, Gardella mentioned that hover state should be same color but with underline text-decoration. So you can override the default link styling in this component for hover state.

packages/experiments-realm/url.gts
tintinthong: In order for this to work with types, I'd need to upgrade to typescript 5.2. Shall I upgrade our types to the most **latest** version of typescript or upgrade it to 5.2.2? I will merge first and resolve in another PR.

https://github.com/microsoft/TypeScript/issues/55623#issuecomment-1704705924

PR #1864, opened 2024-12-02
Create country field with dropdown edit mode

tintinthong:
This creates a country field which is a field definition but it gets data of country codes from a [cdn](https://cdn.jsdelivr.net/npm/country-data-find/+esm). This is meant to support the form filling of an address field https://linear.app/cardstack/issue/CS-7410/refine-address-field

Screenshot 2024-12-02 at 20 46 15
Screenshot 2024-12-02 at 20 46 09

Before we have feature of data realms, we would need this to enumerate fields inside of a dropdown

**Possible alternatives**

We can also hardcode the data inside of the country field itself


lukemelia (approved):

burieberry (approved):
Nice

PR #1865, opened 2024-12-02
Fix checkout.session.completed handler to properly handle buying extra credits 

jurgenwerk:
We had a flaw in our "credit buy" links where we forgot to include the client_reference_id, which tells us which user bought the credits. I added that query param in the links and adjusted the webhook handler. Also did a small refactor for encoding/decoding logic so that it's clearer what it does. 

Tested on staging and I was able to buy myself extra credits without problems.

### DIFF
--- a/packages/billing/stripe-webhook-handlers/index.ts
+++ b/packages/billing/stripe-webhook-handlers/index.ts
@@ -79,7 +79,7 @@ export type StripeCheckoutSessionCompletedWebhookEvent = StripeEvent & {
       id: string;
       object: 'checkout.session';
       client_reference_id: string;
-      customer: string;
+      customer: string | null; // string when payment link is for subcribing to the free plan, null when buying extra credits
       metadata:
         | {
             credit_reload_amount: string;
--- a/packages/billing/stripe-webhook-handlers/index.ts
+++ b/packages/billing/stripe-webhook-handlers/index.ts
@@ -79,7 +79,7 @@ export type StripeCheckoutSessionCompletedWebhookEvent = StripeEvent & {
       id: string;
       object: 'checkout.session';
       client_reference_id: string;
-      customer: string | null; // string when payment link is for subcribing to the free plan, null when buying extra credits
+      customer: string | null; // string when payment link is for subscribing to the free plan, null when buying extra credits
       metadata:
         | {
             credit_reload_amount: string;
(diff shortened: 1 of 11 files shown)


packages/billing/stripe-webhook-handlers/index.ts
```diff
@@ -79,7 +79,7 @@ export type StripeCheckoutSessionCompletedWebhookEvent = StripeEvent & {
       id: string;
       object: 'checkout.session';
       client_reference_id: string;
-      customer: string;
+      customer: string | null; // string when payment link is for subcribing to the free plan, null when buying extra credits
```
lukemelia:
```suggestion
      customer: string | null; // string when payment link is for subscribing to the free plan, null when buying extra credits
```

packages/host/app/components/operator-mode/profile/profile-subscription.gts
```diff
@@ -8,15 +9,28 @@ import {
...
 export default class ProfileSubscription extends Component<Signature> {
+  @service private declare billingService: BillingService;
+  @service private declare matrixService: MatrixService;
+
+  @action
+  urlWithClientReferenceId(url: string) {
+    return `${url}?client_reference_id=${encodeWebSafeBase64(
+      this.matrixService.userId as string,
+    )}`;
+  }
```
lukemelia:
```suggestion
  urlWithClientReferenceId = (url: string) => {
    return `${url}?client_reference_id=${encodeWebSafeBase64(
      this.matrixService.userId as string,
    )}`;
  };
```

Prefer this style for binding functions over @action per Ed

packages/host/app/components/operator-mode/profile/profile-subscription.gts
```diff
@@ -8,15 +9,28 @@ import {
 } from '@cardstack/boxel-ui/components';
 import { IconHexagon } from '@cardstack/boxel-ui/icons';
 
+import { encodeWebSafeBase64 } from '@cardstack/runtime-common';
+
 import WithSubscriptionData from '@cardstack/host/components/with-subscription-data';
 import BillingService from '@cardstack/host/services/billing-service';
+import MatrixService from '@cardstack/host/services/matrix-service';
```
lukemelia:
```suggestion
import type BillingService from '@cardstack/host/services/billing-service';
import type MatrixService from '@cardstack/host/services/matrix-service';
```

lukemelia (approved):

richardhjtan (approved):

jurgenwerk:
Thanks @lukemelia for code change suggestions!

PR #1866, opened 2024-12-02
Fix stale credit info in account popup

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7578/stale-credits-info-in-account-popup-problem

### What is changing

This is to fix stale credit info in the account popup after talking with the AI bot, credit is only updated until we refresh the page.

In this PR we fix it by fetching and showing up-to-date credit info in the account popup.

**Before**

https://github.com/user-attachments/assets/899d4b1e-490c-4b2f-875c-5e1d18307beb

**After**

https://github.com/user-attachments/assets/4a574873-996f-41fa-9d55-1b9553d6aba1

### DIFF
--- a/packages/host/app/components/operator-mode/profile-info-popover.gts
+++ b/packages/host/app/components/operator-mode/profile-info-popover.gts
@@ -5,7 +5,7 @@ import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 
 import { Avatar, BoxelButton } from '@cardstack/boxel-ui/components';
-import { cn, or } from '@cardstack/boxel-ui/helpers';
+import { cn } from '@cardstack/boxel-ui/helpers';
 
 import WithSubscriptionData from '@cardstack/host/components/with-subscription-data';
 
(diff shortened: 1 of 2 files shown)


jurgenwerk (commented):
I think the buttons that gray out on every open are annoying and actually that is probably a bug because we're caching stripe payment links - we're not loading these links every time user opens the popover so that visual behavior is unnecessary.

https://github.com/user-attachments/assets/417393a3-2cb3-4118-8d8e-ac403041c940

richardhjtan:
> I think the buttons that gray out on every open are annoying and actually that is probably a bug because we're caching stripe payment links - we're not loading these links every time user opens the popover so that visual behavior is unnecessary.
> 
>  Screen.Recording.2024-12-02.at.16.13.52.mov

I fixed this in latest commit

https://github.com/user-attachments/assets/09f2929a-b254-42ff-8951-ce34149b87b1

lukemelia:
I thought this was going to tie into the matrix notification work that @FadhlanR did. What's the relationship between that and the user's credit balance?

jurgenwerk:
@lukemelia that already works for work that happens in webhooks. But we have a missing place to trigger a notification - in the AI bot when it is adding to the credits ledger. We have a ticket for that but for now to see the updated credits amount after interacting with the bot (spending credits) the easier/faster solution was to refresh credits amount every time the user popover gets opened.

jurgenwerk (approved):

PR #1868, opened 2024-12-03
Store skills in matrix room state

lukemelia:
- [x] Update aibot to draw its skills from new roomstate config
- [x] add support for disabling/enabling skills
- [x] In UI, sort skills by name
- [x] queue skillsconfig updates to avoid the changes stepping on each other


packages/host/app/services/matrix-service.ts
```diff
@@ -627,6 +624,37 @@ export default class MatrixService extends Service {
...
+    );
+    let skillEventIdsStateEvent: Record<string, any> = {};
+    try {
+      skillEventIdsStateEvent = await this.client.getStateEvent(
+        roomId,
+        SKILLS_STATE_EVENT_TYPE,
+        '',
+      );
+    } catch (e) {
+      // this is fine, it just means the state event doesn't exist yet
```
lukemelia: make catch more precise

IanCal: Could have checked "instanceof MatrixError" but I think that would result in importing the whole matrix sdk that I think we've been trying to avoid. Instead I checked the errcode for the relevant error message.

packages/ai-bot/helpers.ts
```diff
@@ -120,6 +161,22 @@ export function constructHistory(history: IRoomEvent[]) {
   return latestEvents;
 }
 
+function getEnabledSkills(
+  eventlist: DiscreteMatrixEvent[],
+  cardFragments: Map<string, CardFragmentContent>,
+): CardResource[] {
+  let skillsConfigEvent = eventlist
+    .reverse()
```
lukemelia: Try using findLast here to save memory

PR #1869, opened 2024-12-03
Enhance address field

tintinthong:
- makes country chosen by dropdown
- change fields schema to be more standardised version of address

**Look at Vendor card**

Screenshot 2024-12-03 at 13 38 52

**Works with Map card**
Screenshot 2024-12-03 at 14 17 42

### DIFF
--- a/packages/experiments-realm/Vendor/1.json
+++ b/packages/experiments-realm/Vendor/1.json
@@ -30,12 +30,16 @@
         }
       ],
       "mailingAddress": {
-        "streetAddress": "50 California St Ste 1500",
+        "addressLine1": "50 California St Ste 1500",
+        "addressLine2": null,
         "city": "San Francisco",
-        "region": "CA",
+        "state": "CA",
         "postalCode": "94111",
-        "poBoxNumber": null,
-        "country": "USA"
+        "country": {
+          "name": "United States of America",
+          "code": "US"
+        },
+        "poBoxNumber": null
       },
       "preferredPaymentMethod": {
         "type": "Crypto Payment",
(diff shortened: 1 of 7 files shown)


packages/experiments-realm/address.gts
```diff
@@ -4,25 +4,89 @@ import {
...
+  }
+
+  if (countryCode) {
+    rows[2].push(countryCode);
+  }
+
+  if (poBoxNumber) {
+    rows[3].push(`PO Box ${poBoxNumber}`);
+  }
+  return rows;
```
richardhjtan:
```suggestion
 const addressComponents = [
    // Street address (including PO Box if present)
    [
      poBoxNumber ? `PO Box ${poBoxNumber}` : null,
      addressLine1,
      addressLine2,
    ].filter(Boolean),
    // City, state, postal
    [[city, state, postalCode].filter(Boolean).join(', ')].filter(Boolean),
    // Country
    countryCode ? [countryCode] : [],
  ];

  // Remove empty rows and return
  return addressComponents.filter((row) => row.length > 0);
```

Perhaps we can make it more readable with the suggestion above

richardhjtan: The PO Box number will be at first line, some reference: 
https://www.smarty.com/articles/postal-address-examples-formats

packages/experiments-realm/address.gts
```diff
@@ -4,25 +4,89 @@ import {
...
+      let rows = getAddressRows(
+        this.addressLine1,
+        this.addressLine2,
+        this.city,
+        this.state,
+        this.postalCode,
+        this.country?.name,
+        this.poBoxNumber,
+      );
+      return rows.filter((r) => r.length > 0).join(', ');
```
richardhjtan: @tintinthong if we agree to use the suggestion above, then we can just

```
 return rows.join(',\n');
```

Added `\n` so it can push another row of address to next line

packages/experiments-realm/Vendor/1.json
```diff
@@ -78,17 +82,20 @@
...
       "preferredPaymentMethod.cryptoPayment.chain": {
         "links": {
-          "self": "../Chain/1"
+          "self": "http://localhost:4201/experiments/Chain/1"
         }
       },
       "preferredPaymentMethod.cryptoPayment.token": {
         "links": {
-          "self": "../Token/1"
+          "self": "http://localhost:4201/experiments/Token/1"
```
richardhjtan, line 98: Will these affect the relationships in staging etc?

tintinthong: Good catch. Yes it will

lucaslyl (approved):

packages/experiments-realm/address.gts
tintinthong: No I specifically used this to be one single line so it can be used with a maps api

PR #1872, opened 2024-12-03
CS-7592 Save stripe email to user table 

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7592/save-users-stripe-email-to-users-table-so-that-we-can-prefill-the

### What is changing
- save the Stripe customer's email in the user table.
- prefill the email in the Stripe customer portal link URL.

**Screenshot**

https://github.com/user-attachments/assets/a3f7edbd-f53b-48c6-90e6-98bacbc79665


packages/billing/stripe-webhook-handlers/checkout-session-completed.ts
```diff
@@ -56,6 +58,14 @@ export async function handleCheckoutSessionCompleted(
       stripeCustomerId = user.stripeCustomerId;
     }
 
+    if (stripeCustomerEmail) {
```
jurgenwerk: Isn't email always present in the webhook payload?

packages/host/app/services/billing-service.ts
```diff
@@ -51,7 +52,13 @@ export default class BillingService extends Service {
...
 
   get customerPortalLink() {
-    return this.stripeLinks.value?.customerPortalLink;
+    if (!this.subscriptionData?.stripeCustomerEmail) {
+      return this.stripeLinks.value?.customerPortalLink?.url;
+    }
+    const encodedEmail = encodeURIComponent(
+      this.subscriptionData.stripeCustomerEmail,
+    );
+    return `${this.stripeLinks.value?.customerPortalLink?.url}?prefilled_email=${encodedEmail}`;
```
jurgenwerk: Avoid deep optional chaining as it can be confusing to read and reason about, especially when repeated. I find it better if we deal with the uncertainty right at the beginning and then operate on things we know they have a value:

```
if (!this.stripeLinks.value)
  return undefined; 
}

let customerPortalLink = this.stripeLinks.value.customerPortalLink.url;

... rest of the code 
```

jurgenwerk: What happens for existing users who do not have their email saved? Will the link be `...link...?prefilled_email=undefined`? If yes that needs fixing

packages/billing/stripe-webhook-handlers/checkout-session-completed.ts
richardhjtan: Right, since it is required when filling in customer details - I removed the email check

packages/host/app/services/billing-service.ts
richardhjtan: @jurgenwerk  Yes there is handling on that, for no email saved from us will be just `...link...`. I have tested the existing account with no email saved

jurgenwerk (approved):

PR #1876, opened 2024-12-03
Add support to returning instances in an error state in prerendered search

habdelra:
This PR updates the prerendered search to include card instances in an error state.

error-search

In the case where there is a last known good state we will render that. I opted to render with 25% opacity in a red font color--we don't really have the ability to mess with the prerendered HTML so this is a pretty light touch:

image

In the case there is no last known good state, we'll try our best to determine the card type for the card in error. This may be impossible depending on the kind of error. If we cannot determine the card type, then the item cannot be rendered in the cards grid (because the card type is table stakes for being included in the cards grid). If it is possible to know the card type (e.g. the card error is just a missing link), then we'll render the item with a default error tile.

image

As part of this i'm opting not to include error results in the card chooser and in the card search. Let's review this approach first, and if we want to incorporate error results in those areas let's followup with a new PR.

Also as part of this I ran into an issue where our primary key for the boxel_index was incorrect--we should not be including the `type` column in the PK, as this prevents an index entry from going into and out of an error state.


packages/base/cards-grid.gts
```diff
@@ -284,6 +285,10 @@ class Isolated extends Component<typeof CardsGrid> {
       .operator-mode .buried .add-button {
         display: none;
       }
+      .instance-error {
+        opacity: 0.25;
```
backspace: 0.25 vs 0.33:

Experiments Workspace 2024-12-03 17-44-52

Experiments Workspace 2024-12-03 17-44-27

I find the latter easier to read but I think it’s still clear that something is amiss. Also, maybe showing a tooltip on hover that indicates tapping the errored card will show the error would be helpful.

backspace (approved):
I tried this locally, great improvement! The tests look thorough.

packages/base/cards-grid.gts
habdelra: Sure

lukemelia (approved):

jurgenwerk (approved):
Nice!

PR #1880, opened 2024-12-04
Fix fitted task in anticipation for using in CRM

tintinthong:
I wanted to incorporate the round circle on the top-left as part of design into the fitted but I noticed it was very difficult to mangle with the container queries that dont conform in some way to the fitted driver in code made the container queries  more dumb. 

This shud unblock the crm stuff 

**Before**

Screenshot 2024-12-05 at 00 29 39
Screenshot 2024-12-05 at 00 29 32
Screenshot 2024-12-05 at 00 29 24
Screenshot 2024-12-05 at 00 29 11

**After**

Screenshot 2024-12-04 at 19 38 16
Screenshot 2024-12-04 at 19 38 08
Screenshot 2024-12-04 at 19 37 59
Screenshot 2024-12-04 at 19 37 30

A more explicit way of dealing with fitted task

Perhaps, we can fix the overflow: hidden separately

### DIFF
--- a/packages/experiments-realm/Tag/ad921cba-ffc7-4fdb-af34-ab8b93eae228.json
+++ b/packages/experiments-realm/Tag/ad921cba-ffc7-4fdb-af34-ab8b93eae228.json
@@ -2,7 +2,7 @@
   "data": {
     "type": "card",
     "attributes": {
-      "name": "Application",
+      "name": "App",
       "color": "",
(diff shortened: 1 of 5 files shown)


burieberry (approved):
Yeah we can make another ticket to handle overflow issue. Maybe the title font can be smaller in smaller views.

PR #1883, opened 2024-12-04
Fix CardError doc shape to be consistent with the HTTP card error response

habdelra:
We have a bug where there is an incongruity between the `CardError` shape and the HTTP error body response for card errors that resulted in a bug in rendering card errors during the sprint demo. This PR brings these shapes into alignment.

This PR will require that we clear the index as the old `CardError` shape is currently resident in the index for existing card errors.

### DIFF
--- a/packages/host/app/components/operator-mode/stack-item.gts
+++ b/packages/host/app/components/operator-mode/stack-item.gts
@@ -347,7 +347,7 @@ export default class OperatorModeStackItem extends Component<Signature> {
     }
     return this.cardError.status === 404 &&
       // a missing link error looks a lot like a missing card error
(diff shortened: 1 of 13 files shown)


backspace (approved):
Is there some underlying or shared type missing that could have prevented this mismatch?

lukemelia (approved):

habdelra:
> Is there some underlying or shared type missing that could have prevented this mismatch?

Yeah, I think the error JSONAPI response could probably use parts of the CardError type. It’s a little tricky because CardError is a class and the JSONAPI response is just an interface. There is some overlap there that would be beneficial if there was some shared type.

jurgenwerk (approved):

PR #1884, opened 2024-12-04
Move generator to experiments

backspace:
This doesn’t fully function, it may need work from #1909. But the prime goal is to get it out of the catalog realm, problems can be addressed afterward.


IanCal:
Can we move these to the experiments realm instead while they're being worked on?

lukemelia:
Does the module path of the ai app generator instance json need updating?

jurgenwerk:
I tried testing this locally and can confirm the generator is not in the catalog anymore, however if I visit /experiments then I get this: 

image

Not sure if this is a result of this PR, is this the problem you were mentioning in the description?

jurgenwerk:
Ignore the above, this is because I forgot to add experiments realm to my matrix account.

I was able to run the generator from the experiments realm and make an app but there was an error running loading the module because the generator made 2 classes with the same name... Unsure if related to this PR?

image

lukemelia (approved):

backspace:
> I was able to run the generator from the experiments realm and make an app but there was an error running loading the module because the generator made 2 classes with the same name... Unsure if related to this PR?

I’m also unsure! I’ve made peace with the brokenness, I think the current goal is to remove this from the catalog to not expose it to the alpha testers. I’ve merged in from `main` and it semi-works but I too got duplicate classes. It’s hard to know what is supposed to be working at this point as I constantly experience bugs.

PR #1885, opened 2024-12-05
Cs 7418/create contact

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7418/create-contact

1) restructure CRM contact schema


tintinthong (approved):
There is some stale things with social media links but we shall keep it there so when @burieberry creates the social media links field we can remove it. We will be responsible for cleaning this up

PR #1887, opened 2024-12-05
Fix AI panel not closing properly

FadhlanR:
https://github.com/user-attachments/assets/17ebbcc8-6596-4695-aba4-f41eabcbbb3b

In this PR, instead of directly addressing the bug, I decided to refactor resizable-panel by forking logic from the react-resizable-panels library. Now, the bug is fixed and the codes look cleaner.


richardhjtan:
There is some behaviour change on the resizable panel in code mode. The most noticeable one is that the data section no longer auto-hides when reduced to a certain width.

**Before**

https://github.com/user-attachments/assets/abc72907-8e7e-4998-a3f6-dadcfb934d0d

**Now**

https://github.com/user-attachments/assets/fdecfa27-5eeb-41dc-98a1-d3e65e6cbb5e

It shouldn't be considered a bug but just the decision if we need to preserve this behaviour

richardhjtan:
The AI assistant chat max-width has changed, but looks more natural to me

**Before**
ai-chat-max-width-before

**After**
ai-chat-max-width-now

lukemelia (changes requested):
File names should be consistent with repository style (dasherized)

packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -97,21 +96,22 @@ type SelectedAccordionItem = 'schema-editor' | null;
...
-  codeEditorPanel: (1 - defaultLeftPanelWidth) / 2,
-  rightPanel: (1 - defaultLeftPanelWidth) / 2,
-  emptyCodeModePanel: 1 - defaultLeftPanelWidth,
+  codeEditorPanel: (100 - defaultLeftPanelWidth) / 2,
+  rightPanel: (100 - defaultLeftPanelWidth) / 2,
+  emptyCodeModePanel: 100 - defaultLeftPanelWidth,
 };
 
 const CodeModePanelHeights = 'code-mode-panel-heights';
 const ApproximateRecentPanelDefaultFraction =
```
lukemelia: If we're going to make this out of 100, rename the constant to `ApproximateRecentPanelDefaultPercentage`

FadhlanR:
> There is some behaviour change on the resizable panel in code mode. The most noticeable one is that the data section no longer auto-hides when reduced to a certain width.
> 
> **Before**
> 
>  resize-panel-before.mov 
> **Now**
> 
>  resize-panel-now.mov 
> It shouldn't be considered a bug but just the decision if we need to preserve this behaviour

Thanks for checking, I forgot to remove the collapsible tag for the code panel. I've updated please check it again.

FadhlanR:
> The AI assistant chat max-width has changed, but looks more natural to me
> 
> **Before** ai-chat-max-width-before
> 
> **After** ai-chat-max-width-now

I think the previous min size is too small, so I updated it.

FadhlanR:
> File names should be consistent with repository style (dasherized)

Updated

burieberry (commented):
Looks good overall. In terms of ai-chat panel size, I noticed it starts getting cropped from the right when the viewport gets too small and some of the buttons become unreachable. I recommend not letting it get smaller than approximately 300px. The default size of the ai panel in the designs seems to be around 366px.

Buttons cropped on the right-side when too small:
cropped-ai-chat

lukemelia (approved):
The minSize of 25 for the AI assistant panel feels too large on a wide display. Are we limited to percentage-based min sizes?

FadhlanR:
> The minSize of 25 for the AI assistant panel feels too large on a wide display. Are we limited to percentage-based min sizes?

Yes, the pixel-based constraints added significant complexity, which is why we are limited to percentage-based constraints. However, I’ve adjusted the panel widths in submode-layout. It may not work perfectly when resizing the window, but when the AI panel is closed and reopened, it will open with a better size. cc: @burieberry

lukemelia:
I think there used to be a threshold behavior where if you shrank a panel below a certain threshold and it was collapsible it would collapse. Was that intentionally removed?

FadhlanR:
> I think there used to be a threshold behavior where if you shrank a panel below a certain threshold and it was collapsible it would collapse. Was that intentionally removed?

It's still there, you can try to resize code editor panel in code submode. It will collapse once the size is lower than minSize.

jurgenwerk:
image

Dragging this box is glitchy:

https://github.com/user-attachments/assets/cf8dc269-fc27-4c3a-9777-55f853be8c4d

FadhlanR:
> image
> 
> 
> 
> Dragging this box is glitchy:
> 
> 
> 
> https://github.com/user-attachments/assets/cf8dc269-fc27-4c3a-9777-55f853be8c4d
> 
> 
> 
> 
> 
> 

I'll check the box corner.

For the glitch, i think it's expected behavior since we define the minimum size for recent files panel. So the recent files panel will only open if the cursor dragging delta is more than the minimum size. Is it better to remove the minimum size of recent files panel? cc: @lukemelia

richardhjtan (approved):

jurgenwerk:
I think the previous resize behavior of recent files felt much nicer... you could resize it to a small size and when the handle was very close to to the border, it snapped into the closed position.

Another degradation that I found - I would expect to be able to narrow this column down even further but now it snaps closed when there is still plenty of available space for code:

https://github.com/user-attachments/assets/927f927a-25bd-460f-a260-dbb67f1edb98

FadhlanR:
> Another degradation that I found - I would expect to be able to narrow this column down even further but now it snaps closed when there is still plenty of available space for code:

I've updated the minimum size of code editor. Please check it again.


> image

Fixed!

burieberry (approved):

jurgenwerk:
In my opinion the boxes still snap close too quickly. It feels odd when there is plenty of space left for content and then the box just closes. 

Here is what I am experiencing:

https://github.com/user-attachments/assets/fb5b2b60-0e02-406d-b630-62d1078ea497

Here is the old behaviour where breakpoints aren't that extreme and the experience is nicer: 

https://github.com/user-attachments/assets/d0190242-b09b-4a79-ba8e-43e432c1c95a

FadhlanR:
> In my opinion the boxes still snap close too quickly. It feels odd when there is plenty of space left for content and then the box just closes.
> 
> Here is what I am experiencing:
> 
>  Screen.Recording.2024-12-10.at.21.53.29.mov 
> Here is the old behaviour where breakpoints aren't that extreme and the experience is nicer:
> 
>  Screen.Recording.2024-12-10.at.21.59.24.mov

It seems we need to introduce a minimum length in pixels. What do you think if I file separate tickets for the minimum length in pixels and the collapsed panel behavior? I believe the adjustment will take time with the current calculation logic. However, the current logic is much better than the previous one because it fixes ai panel closing issue, provides more precise cursor positioning when dragging the panel, and is more flexible in terms of allowing multiple tables to be resized simultaneously.

I've implemented a workaround to address your feedback regarding collapsed panel behavior [2946b15](https://github.com/cardstack/boxel/pull/1887/commits/2946b15739025ba46546e3389c121d80df4da162)

jurgenwerk:
I am not actually seeing any improvement with your workaround (tested in the PR preview link). I am ok if issues are addressed in follow up work.

PR #1892, opened 2024-12-05
ci: Add more Matrix tests shards

backspace:
An experiment


habdelra:
this looks nice!

habdelra (approved):

backspace:
> this looks nice!

👍🏻 I was going to test more matrix sizes but the results are so variable… the two runs at this size did all have shorter longest run times, so let’s just go ahead with it 😆

PR #1894, opened 2024-12-05
ContactLink field

burieberry:
- created ContactLink field def, inspired by loosey-goosey field type used in other places in experiments realm
- used in an author card with different static values
- added a card to be able to preview custom fields we're creating for apps
- added catalog entries so the new fields can be added from code-mode
- updated email field because it was erroring by default when empty

edit-preview
preview
author-atom-contact-links


tintinthong (approved):
theres a lint error

PR #1897, opened 2024-12-06
Save user's registration token to users table when signing up

jurgenwerk:
Will be used to track which channels users came through when registering for Boxel.


packages/host/app/services/realm-server.ts
```diff
@@ -82,19 +82,41 @@ export default class RealmServerService extends Service {
       window.localStorage.getItem(sessionLocalStorageKey) ?? undefined;
   }
 
+  async createUser(matrixUserId: string, registrationToken?: string) {
+    await this.login();
+    let response = await this.network.authedFetch(`${this.url.href}_user`, {
```
jurgenwerk: I copied the usage of `this.network.authedFetch`  from `createRealm` from this same service but this feels confusing... why are we providing `Authorization: Bearer ${this.token}` header when `authedFetch` has the  auth token functionality to do that for us? 

However `authedFetch` only supports authenticated querying of realm's content, and not the realm server, which is the case here. So given the logic for realm server auth lives in this service (realm-server), I think this should just be a normal `this.network.fetch` instead? @lukemelia @habdelra

habdelra: that seems like a reasonable conclusion

packages/host/app/components/matrix/register-user.gts
```diff
@@ -713,12 +714,18 @@ export default class RegisterUser extends Component<Signature> {
...
       throw e;
     }
+
     // If access_token and device_id are present, RegisterResponse matches LoginResponse
     // except for the optional well_known field
-    if (auth.access_token && auth.device_id) {
+    if (
+      auth.access_token &&
+      auth.device_id &&
+      this.state.type === 'waitForEmailValidation'
```
habdelra: i'm curious why we needed this condition, there is no guaranteed order when email validation occurs and when the user provides the registration token--they can happen in either order. wouldn't this force a particular order?

packages/realm-server/handlers/handle-create-user.ts
```diff
@@ -0,0 +1,45 @@
...
+import {
+  fetchRequestFromContext,
+  sendResponseForBadRequest,
+  sendResponseForSystemError,
+  setContextResponse,
+} from '../middleware';
+import { RealmServerTokenClaim } from '../utils/jwt';
+import { CreateRoutesArgs } from '../routes';
+
+export default function handleCreateUserRequest({
```
habdelra, line 12: in this function where are we asserting that the bearer token is correct for the user being created? is that happening before we reach this point? I just want to make sure a malicious account cannot create arbitrary users

packages/realm-server/tests/realm-server-test.ts
```diff
@@ -3677,6 +3678,52 @@ module('Realm Server', function (hooks) {
...
+        'matrix user ID is correct',
+      );
+      assert.strictEqual(
+        user.matrixRegistrationToken,
+        'reg_token_123',
+        'registration token is correct',
+      );
+    });
+
+    test('can not create a user without a jwt', async function (assert) {
```
habdelra, line 3722: should we also have a test where the JWT does not match the user being created?

packages/host/config/schema/1733393646040_schema.sql
habdelra, line 1: i always delete the old schema file (the diffs look a lot easier to read then too)

packages/realm-server/handlers/handle-create-user.ts
jurgenwerk: So there is `jwtMiddleware` being applied to this request which will verify the request's token, and in the token's payload there's `matrixUserId`. We use that to upsert the user by that field and save the registration token. So this only allows to create the user with the matrix id from the jwt, which prevents the situations where requests could be creating arbitrary users

packages/realm-server/tests/realm-server-test.ts
jurgenwerk: I don't think this situation is possible–the handler will read the users's matrix id from the verified JWT payload and use that to upsert this particular user (together with the registration token). This means the jwt will always match the user being created

packages/host/app/components/matrix/register-user.gts
jurgenwerk: So given that in the app, the (optional) registration token input phase always appears before the email validation I thought the order is guaranteed in a way where `waitForEmailValidation` is the last phase.

I actually added `this.state.type === 'waitForEmailValidation'` to satisfy the type check where `token` is currently only defined in `type: 'sendToken'` and `type: 'waitForEmailValidation';`. Otherwise the type checker will complain `token` doesn't exist in other types, for example when `type == 'register'`

So given `token` doesn't exist in all state types I think we have 2 options, either to narrow down the possible states in which `initializeNewUser(..., token)` can get called, or simply check if registration token is in state and use it if it is, something like:

```
if (auth.access_token && auth.device_id) {
      await this.matrixService.initializeNewUser(
        auth as LoginResponse,
        this.state.name,
        'token' in this.state ? this.state.token : undefined,
      );
    }
```

packages/host/config/schema/1733393646040_schema.sql
jurgenwerk: Ok, I removed those.

lukemelia (approved):

packages/host/app/components/matrix/register-user.gts
jurgenwerk: I added a comment on why I added this condition in https://github.com/cardstack/boxel/pull/1897/commits/7f0bea5e4dcb4a3576b4e3b53a5a6ec648040234. If this is still concerning we can talk more about it in a pairing session but for now I'll go ahead and merge this.

PR #1899, opened 2024-12-06
Fix task cancellation error when fetching subscription data

habdelra:
This issue was driving me crazy, I had to fix it....

This fixes the task cancellation error when fetching subscription data. Please make sure that tasks are only used in places where the lifetime of the owner is not infinte. Using a task in a service is a code smell. Anytime a Task.perform is `awaited` on the outside the task become non-cancellable. In this case the task lived in a service, which lives forever,. There is no need for cancellable promises that Tasks provide because the service never goes away.

image

This PR removes the task from the service, and adds it to the component that consumes this service's async, which is where it should always have been.

### DIFF
--- a/packages/host/app/components/operator-mode/stack-item.gts
+++ b/packages/host/app/components/operator-mode/stack-item.gts
@@ -328,9 +328,6 @@ export default class OperatorModeStackItem extends Component<Signature> {
 
   @cached
   get cardError() {
-    if (this.args.item.cardError) {
-      debugger;
-    }
     return this.args.item.cardError;
   }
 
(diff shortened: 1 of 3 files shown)


backspace (approved):
> Please make sure that tasks are only used in places where the lifetime of the owner is not infinte.

Is this lintable?

richardhjtan (approved):
Tested and looks good, thanks for the fix 👍

habdelra:
> > Please make sure that tasks are only used in places where the lifetime of the owner is not infinte.
> 
> Is this lintable?

probably, that's not a bad idea

PR #1903, opened 2024-12-09
Free plan credits should not carry over when upgrading to a paid plan

jurgenwerk:
Previously, if you had 1000 credits from the free plan, and upgraded to a paid plan (e.g. "creator" plan for 5000 credits), you'd get 6000 credits. This change adjusts the plan upgrading logic to not add unused free plan credits when a paid subscription is made.

### DIFF
--- a/packages/billing/billing-queries.ts
+++ b/packages/billing/billing-queries.ts
@@ -101,7 +101,7 @@ export async function getPlanByStripeId(
   return {
     id: results[0].id,
     name: results[0].name,
-    monthlyPrice: results[0].monthly_price,
+    monthlyPrice: parseFloat(results[0].monthly_price as string),
     creditsIncluded: results[0].credits_included,
     stripePlanId: results[0].stripe_plan_id,
   } as Plan;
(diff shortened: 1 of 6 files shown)


packages/billing/billing-queries.ts
```diff
@@ -101,7 +101,7 @@ export async function getPlanByStripeId(
   return {
     id: results[0].id,
     name: results[0].name,
-    monthlyPrice: results[0].monthly_price,
+    monthlyPrice: parseFloat(results[0].monthly_price as string),
```
lukemelia: Perhaps it would be better to cast the return value of the query to a response type rather than casting each monthly_price to a string?

lukemelia (approved):

richardhjtan (approved):

packages/billing/billing-queries.ts
jurgenwerk: I added a function to mitigate this repetitiveness in https://github.com/cardstack/boxel/pull/1903/commits/5c822f047cf58523138452b841fa518d45c367fb

PR #1906, opened 2024-12-09
Add live card support for showing and dismissing errors

habdelra:
This PR adds support for showing and dismissing errors in a live fashion in both stack items and in code mode.

sse-errors

(note from the screenshot that there is already an existing bug in main where the monaco editor is not live updating based on file system changes--that is unrelated to this PR)

### DIFF
--- a/packages/host/app/resources/card-resource.ts
+++ b/packages/host/app/resources/card-resource.ts
@@ -35,7 +35,7 @@ import type RealmService from '../services/realm';
 
 interface CardErrors {
   errors: {
-    id: string;
+    id?: string; // 404 errors won't necessarily have an id
     status: number;
     title: string;
     message: string;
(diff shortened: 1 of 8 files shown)


packages/host/app/resources/card-resource.ts
```diff
@@ -47,12 +48,13 @@ export interface CardError {
...
 }
 
+export type CardError = CardErrors['errors'][0];
+
 interface Args {
   named: {
     // using string type here so that URL's that have the same href but are
     // different instances don't result in re-running the resource
     url: string | undefined;
-    loader: Loader;
```
habdelra, line 55: The loader no longer has to be specified for a CardResource--there only ever is one loader.

lukemelia (approved):

PR #1907, opened 2024-12-09
Serialize matrix room mutations

lukemelia:
This will help us avoid network race conditions with swift actions like enabling/disabling skills or changes invoked rapidly via commands.


packages/host/app/components/matrix/room.gts
```diff
@@ -586,16 +585,13 @@ export default class Room extends Component<Signature> {
     return this.autoAttachmentResource.cards;
   }
 
-  // using enqueue task on this ensures that updates don't step on each other
-  private updateSkillIsActiveTask = enqueueTask(
```
backspace, line 590: I’m sure it didn’t since you’re making this change, but why didn’t `enqueueTask` work?

backspace (approved):

packages/host/app/components/matrix/room.gts
lukemelia: It did work, but it doing in the component means it would be bypassed by commands that enable/disable skills

packages/host/app/lib/mutex.ts
```diff
@@ -0,0 +1,43 @@
...
+const mutex = new Mutex();
+
+async function criticalSection() {
+  await mutex.dispatch(async () => {
+    // Your critical section code here
+    console.log('Critical section');
+  });
+}
+*/
+
```
habdelra: maybe provide some more description on what this is doing and when you'd use it. I think this makes it so that the critical section is only ever executed once at a time. is that right?

habdelra: Also can this be unit tested as well since it's a general lib?

lukemelia: Updated!

habdelra (approved):
looks great!

jurgenwerk (approved):

PR #1910, opened 2024-12-10
Allow work tracker instance to be created

tintinthong:
**Changes**
- Work tracker introduced link to  Project card 
- Based upon the project linked, the task query executes a filter for Project card id (this checks the relative path)
- If no project is linked, the isolated view will have a disabled overlay on top of it

**No Project**
Screenshot 2024-12-10 at 21 49 06
Screenshot 2024-12-10 at 21 58 01

**With Project**
Screenshot 2024-12-10 at 21 48 54
Screenshot 2024-12-10 at 21 48 47


packages/experiments-realm/work-tracker.gts
```diff
@@ -625,3 +645,22 @@ function removeFileExtension(cardUrl: string) {
...
+  let relativeTo = relativeURL(
+    new URL(absoluteUrl),
+    new URL(instanceDirectoryUrl),
+    realmURL,
+  );
+  if (!relativeTo) {
+    throw new Error(`Missing relative url`);
+  }
+  return relativeTo;
+}
```
tintinthong, line 666: I do not know why @field assignee has the indexed absolute url in the search doc whereas @field project has the indexed url relative in the search doc

This is even after clearing the db

tintinthong:
What is not handled here yet is I think task that are created are not automatically linked to the same project linked in the work tracker

packages/experiments-realm/work-tracker.gts
burieberry: Yeah I came across this issue too. Sometimes they're absolute, sometimes relative. I think when it's a different realm, it will have to be an absolute url.

burieberry (commented):
Looks much better. Here is a design decision to make: Should the `project` be a linksToMany? If not, we could eliminate the Project card and the work tracker could be per project.

tintinthong:
> Looks much better. Here is a design decision to make: Should the `project` be a linksToMany? If not, we could eliminate the Project card and the work tracker could be per project.

I understand that this was motivated by having task belong to different project in most work trackers. But I think the bottom-line is links-to-many support is limited. 

I think there are 2 points to consider for this design

**Searchability**

I am able to look at my project and identify which applications my project are linked inside of-- the reverse query. Currently, there is no way to do that. So the good-enough solution now is we often make use of design that end up going to a page with all possible app cards and enumarting them with project tags and searching the project. 

If we use `linksToMany`, the query in our good-enough solution will not work. Also, inside of the kanban board I don't think filtering ur task by a linksToMany works as well. 

**Creatability**

I think when I create a card from the kanban board, I want it to link to a particular project. It cud be that we want the creation of a task to just inherit the linksToMany of the project but I think handling the plural case is MORE confusing. Like what if the user just wants to have it in only 1 project but not 2 that the work tracker is linked to then we must make more designs to decide which project the task belongs and wud the task disappear if we chose 1 then. 


**Possible Next Step**

Perhaps, there shud be a ticket to scope and understand what features `links-to-many` support querying provides otherwise, we would always be asking this question of "Should it work" vs "Should I design it this way"? But in general, I am quite frugal with my usage of `linksToMany` field

tintinthong:
Make `linksToMany` project inside Work Tracker. Task can only `linksTo` a single project

tintinthong:
@burieberry I have discussed with chris and we will stick for linksTo for now. I will fix the conflicts and merge this today

PR #1912, opened 2024-12-10
Allow command execute to take in pojos rather than cards

IanCal:
The idea here is to allow the function to take in either a *card* or a pojo with the right fields.

Here all the custom fields in the type are optional, and it optionally allows carddef fields (e.g. title, description). 

I think I've got the typing right for allowing undefined things too, and only when set as the input type. 

If this looks like the right direction I'll put tests in for the things that should type check, and if there's a way of adding tests that check something *doesn't* type check that'd be great.

Examples of things that do and don't type check

```import { Command } from '@cardstack/runtime-common';
import {
  CardDef,
  StringField,
  contains,
  field,
} from 'https://cardstack.com/base/card-api';

export class ExampleInput extends CardDef {
  @field exampleInputField = contains(StringField);
  @field exampleSecondInputField = contains(StringField);
}

class ExampleResult extends CardDef {
  @field exampleOutputField = contains(StringField);
}

export class ExampleCommand extends Command<ExampleInput, ExampleResult> {
  inputType = ExampleInput;

  async getInputType() {
    return ExampleInput;
  }

  async run(input: ExampleInput) {
    console.log('just to use input', input);
    let otherCommand = new ExampleCommandWithNoInput(this.commandContext);
    return await otherCommand.execute(undefined);
  }
}

export class ExampleCommandWithNoInput extends Command<
  undefined,
  ExampleResult
> {
  inputType = ExampleInput;

  async getInputType() {
    return undefined;
  }

  async run(input: undefined) {
    console.log('just to use input', input);
    let result = new ExampleResult();
    result.exampleOutputField = 'test';
    return result;
  }
}

export class ExampleCommandUsingInput extends Command<
  ExampleInput,
  ExampleResult
> {
  inputType = ExampleInput;

  async getInputType() {
    return ExampleInput;
  }

  async run(input: ExampleInput) {
    let otherCommand = new ExampleCommand(this.commandContext);
    return await otherCommand.execute(input);
  }
}

export class ExampleCommandUsingInputPojo extends Command<
  ExampleInput,
  ExampleResult
> {
  inputType = ExampleInput;

  async getInputType() {
    return ExampleInput;
  }

  async run(input: ExampleInput) {
    let otherCommand = new ExampleCommand(this.commandContext);
    return await otherCommand.execute({
      exampleInputField: input.exampleInputField,
      exampleSecondInputField: input.exampleSecondInputField,
    });
  }
}

export class ExampleCommandUsingInputPojoTypeError extends Command<
  ExampleInput,
  ExampleResult
> {
  inputType = ExampleInput;

  async getInputType() {
    return ExampleInput;
  }

  async run(_input: ExampleInput) {
    let otherCommand = new ExampleCommand(this.commandContext);
    // Type error
    /*
Argument of type '{ noneExistingField: string; }' is not assignable to parameter of type 'ExampleInput | Omit<ExampleInput, keyof CardDef>'.
  Object literal may only specify known properties, and 'noneExistingField' does not exist in type 'ExampleInput | Omit<ExampleInput, keyof CardDef>'.
    (*/
    return await otherCommand.execute({
      noneExistingField: 'test',
    });
  }
}

export class ExampleCommandUsingInputAllowsMissingFields extends Command<
  ExampleInput,
  ExampleResult
> {
  inputType = ExampleInput;

  async getInputType() {
    return ExampleInput;
  }

  async run(_input: ExampleInput) {
    let otherCommand = new ExampleCommand(this.commandContext);
    return await otherCommand.execute({
      exampleInputField: 'test',
    });
  }
}

export class ExampleCommandUsingInputAllowOptionallyCardDefFields extends Command<
  ExampleInput,
  ExampleResult
> {
  inputType = ExampleInput;

  async getInputType() {
    return ExampleInput;
  }

  async run(_input: ExampleInput) {
    let otherCommand = new ExampleCommand(this.commandContext);

    return await otherCommand.execute({
      exampleInputField: 'test',
      exampleSecondInputField: 'test',
      title: 'test',
    });
  }
}
```


packages/runtime-common/commands.ts
```diff
@@ -66,20 +68,38 @@ export abstract class Command<
...
+  async execute(
+    input: CardInputType extends CardDef | undefined
+      ? CardInputType | Omit<CardInputType, keyof CardDef>
+      : never,
+  ): Promise<CardResultType> {
     // internal bookkeeping
     // todo: support for this.runTask being defined
     // runTask would be an ember task, run would just be a normal function
 
+    let inputCard: CardInputType;
```
IanCal, line 81: this section feels a bit awkward

packages/runtime-common/commands.ts
```diff
@@ -66,20 +68,38 @@ export abstract class Command<
...
+    if (input === undefined) {
+      inputCard = undefined as CardInputType;
+    } else if ('isCardDef' in input && input.isCardDef) {
+      inputCard = input as CardInputType;
+    } else {
+      let InputType = await this.getInputType();
+      if (!InputType) {
+        inputCard = undefined as CardInputType;
```
lukemelia: perhaps this conditional branch should be an error?

packages/runtime-common/commands.ts
```diff
@@ -51,7 +51,9 @@ export abstract class Command<
   CommandConfiguration extends any | undefined = undefined,
 > {
   // Is this actually type checking ?
-  abstract getInputType(): Promise<{ new (args: any): CardInputType }>; // TODO: can we do better than any here?
+  abstract getInputType(): Promise<
+    { new (args: any): CardInputType } | undefined
```
lukemelia: Could this be:

```suggestion
    { new (arg?: Partial<CardInputType>): CardInputType } | undefined
```

lukemelia (approved):
Left a few ideas but overall looks good and will clean up lots of code

packages/runtime-common/commands.ts
IanCal: I think this side is for when there's no input required.

lukemelia: no input is consumed but the user provided input -- sounds like a state we should be noisy about

IanCal: Oh of course, thanks, yeah this should throw an error

lukemelia (approved):

PR #1913, opened 2024-12-10
Cs 7651 create contact fitted

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7651/create-contact-fitted

This is actually an update of the contact field


tintinthong:
I will update these images with the new avatar thumbnails we have in s3

burieberry (approved):
This is fine, however I think we're missing some guidance on what the fitted cards should look like for other sizes as previewed in Code Mode. I created a ticket for it (ECO-125) so we can discuss at next sprint planning.

PR #1915, opened 2024-12-10
Fix for bug where monaco editor wasn't live updating based on index changes

habdelra:
This PR fixes a bug where the monaco editor was not live updating based on index changes. The issue was that we were being ambiguous as to the type of index event that would trigger the monaco update (or really, the `FileResource` update). previously we introduced a new index event that is triggered at the initiation of indexing. The index initiation event was actually triggering the `FileResource` update, but the issue was that the index was not done updating at the time the `FileResource` asked for the new content, so the old content was coming back. The fix was to be precise with the index event, and specifically wait for the `incremental` index event which is triggered _after_ the index has been updated.

As well as, we've added the missing test that should have caught this issue.

monaco-live

### DIFF
--- a/packages/host/app/resources/file.ts
+++ b/packages/host/app/resources/file.ts
@@ -199,12 +199,21 @@ class _FileResource extends Resource<Args> {
 
     this.setSubscription(realmURL, (event: { type: string; data: string }) => {
       let eventData = JSON.parse(event.data);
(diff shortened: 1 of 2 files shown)


lukemelia (approved):

jurgenwerk (approved):

richardhjtan (approved):

burieberry (approved):
Makes sense

PR #1916, opened 2024-12-11
fix status issue using a computed field. Remove status dropdown

tintinthong:


burieberry (approved):
Looks good. 

An idea -- how about using the same `StatusPill` component as the embedded and atom template of the `StatusTag` field? Currently it's not possible to preview because it's missing embedded view.

tintinthong:
We used `margin-top` to resolve the issue that occurs bcos of the inconsistency of field-component for atom. Perhaps, its not an issue but sometimes its annoying that atom StringField behave differently from atom CompoundField

Screenshot 2024-12-14 at 09 08 25

https://github.com/cardstack/boxel/pull/1916/commits/299ffcb90eb925c947f47a2c8fef0b8d4b0f07e3

PR #1917, opened 2024-12-11
Improve "buy credits" payment links display

jurgenwerk:
Adjusts order of links to be ascending, and fixes large numbers formatting. 

Before:
image

After:
image


FadhlanR (approved):
I think it would be good if you use the same number format for monthly credits

richardhjtan (approved):

burieberry (approved):

lukemelia (approved):

PR #1918, opened 2024-12-11
Fix: user credit amounts do not update in profile settings

FadhlanR:
After investigating the issue, I discovered that the realm server did not send a `billing-notification` after handling the `checkout.session.completed` event. This occurred because we were expecting all events to include `event.data.object.customer`. However, for the `checkout.session.completed` event, the matrix user id is stored in `event.data.object.client_reference_id` instead.

### DIFF
--- a/packages/billing/stripe-webhook-handlers/index.ts
+++ b/packages/billing/stripe-webhook-handlers/index.ts
@@ -1,9 +1,10 @@
-import { DBAdapter } from '@cardstack/runtime-common';
+import { DBAdapter, decodeWebSafeBase64 } from '@cardstack/runtime-common';
 import { handlePaymentSucceeded } from './payment-succeeded';
 import { handleCheckoutSessionCompleted } from './checkout-session-completed';
 
 import Stripe from 'stripe';
 import { handleSubscriptionDeleted } from './subscription-deleted';
+import { getUserByStripeId } from '../billing-queries';
 
 export type StripeEvent = {
   id: string;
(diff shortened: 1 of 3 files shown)


packages/realm-server/handlers/handle-stripe-webhook.ts
```diff
@@ -14,10 +15,23 @@ export default function handleStripeWebhookRequest({
     let response = await stripeWebhookHandler(
       dbAdapter,
       request,
-      async (stripeUserId: string) => {
-        let user = await getUserByStripeId(dbAdapter, stripeUserId);
-        if (user) {
-          await sendEvent(user.matrixUserId, 'billing-notification');
+      async ({
```
jurgenwerk: Why does this need to be a callback? Can it just be a normal function in `packages/billing/stripe-webhook-handlers/index.ts`?

packages/billing/stripe-webhook-handlers/index.ts
```diff
@@ -130,27 +136,34 @@ export default async function stripeWebhookHandler(
...
       break;
-    case 'checkout.session.completed':
+    }
+    case 'checkout.session.completed': {
       await handleCheckoutSessionCompleted(
         dbAdapter,
         event as StripeCheckoutSessionCompletedWebhookEvent,
       );
-      sendBillingNotification(event.data.object.customer);
+      await sendBillingNotification({
```
jurgenwerk: Add code comments explaining scenarios about when stripeCustomerId or encodedMatrixUserId are present. i think stripe customer id is present when subscribing to the free plan, and not present when buying credits

packages/realm-server/handlers/handle-stripe-webhook.ts
```diff
@@ -14,10 +15,23 @@ export default function handleStripeWebhookRequest({
...
+      }) => {
+        let matrixUserId = encodedMatrixUserId
+          ? decodeWebSafeBase64(encodedMatrixUserId)
+          : undefined;
+        if (stripeCustomerId) {
+          let user = await getUserByStripeId(dbAdapter, stripeCustomerId);
+          matrixUserId = user?.matrixUserId;
+        }
+
+        if (matrixUserId) {
```
jurgenwerk: What are the cases where we tolerate `matrixUserId` being undefined? Shouldn't that be a bug?

packages/billing/stripe-webhook-handlers/index.ts
jurgenwerk: Perhaps it would be simpler if you introduced a function `extractUser(stripeWebhookPayload)` and then pass the user to the function where you send the event. That way you could get rid of this parameter duality

richardhjtan (approved):

packages/realm-server/handlers/handle-stripe-webhook.ts
jurgenwerk: I see it's needed because it needs the `this.matrixClient` reference which lives in the `RealmServer`

jurgenwerk (approved):

PR #1920, opened 2024-12-11
ai: Add attachment for JSON parse failure

backspace:
It’s helpful when seeing an error like this to have the data attached that failed to parse:

AI-BOT-P - SyntaxError: Unexpected token o in JSON at position 1 - buck doyle@cardstack com - Cardstack Mail 2024-12-11 12-16-06

I’d prefer this to be applied through the whole codebase but it’s not clear to me how to easily accomplish that, something to think about if we keep getting errors like this elsewhere.


lukemelia (approved):
Nice!

PR #1924, opened 2024-12-12
Fix company not being displayed inside account

tintinthong:
- need to use computed. Filed https://linear.app/cardstack/issue/CS-7670/delegate-to-nested-field
- used address field and added the atom to it
- defaults shud be show icon but have no value. But inside account isolated we still show a default tho company info is missing

### DIFF
--- a/packages/experiments-realm/crm/account.gts
+++ b/packages/experiments-realm/crm/account.gts
@@ -18,38 +18,10 @@ import SummaryGridContainer from '../components/summary-grid-container';
 import BuildingIcon from '@cardstack/boxel-icons/captions';
 import AccountHeader from '../components/account-header';
 import { WebsiteField } from '../website';
-import MapPinIcon from '@cardstack/boxel-icons/map-pin';
 import { Avatar, Pill } from '@cardstack/boxel-ui/components';
 import { EntityDisplay } from '../components/entity-display';
 
 // Perhaps, can be address?
-export class LocationField extends AddressField {
-  static icon = MapPinIcon;
-  static displayName = 'Location';
-  static atom = class Atom extends Component<typeof this> {
-    <template>
-      {{#if @model.country.name}}
-        <div class='row'>
-          <MapPinIcon class='icon' />
-          <span>{{@model.city}}, {{@model.country.code}}</span>
-        </div>
-      {{/if}}
-      <style scoped>
-        .row {
-          display: inline-flex;
-          align-items: center;
-          gap: var(--boxel-sp-xxs);
-        }
-        .icon {
-          width: var(--boxel-icon-sm);
-          height: var(--boxel-icon-sm);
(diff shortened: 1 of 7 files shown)


packages/experiments-realm/crm/account.gts
```diff
@@ -18,38 +18,10 @@ import SummaryGridContainer from '../components/summary-grid-container';
 import BuildingIcon from '@cardstack/boxel-icons/captions';
 import AccountHeader from '../components/account-header';
 import { WebsiteField } from '../website';
-import MapPinIcon from '@cardstack/boxel-icons/map-pin';
 import { Avatar, Pill } from '@cardstack/boxel-ui/components';
 import { EntityDisplay } from '../components/entity-display';
 
 // Perhaps, can be address?
```
richardhjtan, line 24: Do we still need this comment?

richardhjtan:
Are we expecting the address and website info to be aligned center? Currently it looks like align start

Screenshot 2024-12-12 at 5 02 06 PM

richardhjtan (approved):
Looks good overall!

packages/experiments-realm/crm/account.gts
lucaslyl: for this special use case, we would like to make the icon smaller instead of changing the align-start to align-center

packages/experiments-realm/website.gts
```diff
@@ -9,13 +9,11 @@ export class WebsiteField extends UrlField {
...
-          <:thumbnail>
-            <WorldWwwIcon class='icon' />
-          </:thumbnail>
-        </EntityDisplay>
-      {{/if}}
+      <EntityDisplay @name={{@model}}>
+        <:thumbnail>
+          <WorldWwwIcon class='icon' />
+        </:thumbnail>
+      </EntityDisplay>
```
lucaslyl, line 16: Screenshot 2024-12-12 at 17 14 10
perhaps add a styled scoped CSS for the icon:
 
```.icon {
width: var(size);
height:var(size);
flex-shrink: 0;
}

and for .name CSS that from outside we can add CSS:
word-break: break-word
```

we can apply the same CSS code for the address field
```
  <MapPinIcon class='icon' />
```

or optionally:
What do you think about this approach? Do you think handling the styling of icon and name within the entity-display component would work well in this case?

tintinthong: Let's not touch this. It is not the focus of the PR. I intentionally ignored changing .name and any custom css bcos it wud make this PR unnecessarily confusing. Let's handle that tomorrow together.

packages/experiments-realm/crm/account.gts
tintinthong: There is a ticket for what @lucaslyl said. But, yes I will remove this comment since it is stale

PR #1927, opened 2024-12-12
Implement host command to enable and disable skills in a room

lukemelia:


packages/host/app/commands/update-skill-activation.ts
```diff
@@ -0,0 +1,52 @@
...
+    await matrixService.updateStateEvent(
+      roomId,
+      SKILLS_STATE_EVENT_TYPE,
+      '',
+      async (currentSkillsConfig) => {
+        let newSkillsConfig = {
+          enabledEventIds: [...(currentSkillsConfig.enabledEventIds || [])],
+          disabledEventIds: [...(currentSkillsConfig.disabledEventIds || [])],
+        };
+        if (value) {
```
backspace, line 37: I liked `isActive` as a name for this more, maybe there’s an interface reason for it to be `value` now but it’s less explanatory.

backspace (approved):
I confirmed that this (still) works by running locally and logging in with a separate browser, the setting is persisted 🎉 

Personal — Experiments Workspace 2024-12-12 15-00-12

packages/host/app/commands/update-skill-activation.ts
lukemelia: 👍  I'll adjust it in my next PR

PR #1928, opened 2024-12-12
Add live option to `getCards()`

habdelra:
This PR adds new options to the `getCards()` API that allows the search to be live as well as to return live cards. The cards from the search are actually provided via `CardResource`.

In this approach I follow a similar pattern to how we get or create realm resources from the realm service, only instead of realm resources I'm operating with card resources, which takes advantage of an identity map for all the cards it deals with.

Probably a future refactor is to retire the `CardService.search()`. i don't think it really does what we want since it is not live, and the cards that it returns are not live and does not leverage the identity map from `CardResource`, so it may create duplicative instances of cards when it deserializes the results.

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -53,9 +53,6 @@ export class Search extends Resource<Args> {
     waitForPromise(this.loaded);
 
     if (isLive) {
-      // TODO this triggers a new search against all realms if any single realm
-      // updates. Make this more precise where we only search the updated realm
-      // instead of all realms.
       this.subscriptions = this.realmsToSearch.map((realm) => ({
         url: `${realm}_message`,
         unsubscribe: subscribeToRealm(
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -56,7 +56,7 @@ export class Search extends Resource<Args> {
       this.subscriptions = this.realmsToSearch.map((realm) => ({
         url: `${realm}_message`,
         unsubscribe: subscribeToRealm(
-          `${realm}_message`,
+          realm,
           ({ type, data }: { type: string; data: string }) => {
             let eventData = JSON.parse(data);
             // we are only interested in incremental index events
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -196,10 +196,6 @@ export function getSearchResults(
   query: Query,
   realms?: string[],
   opts?: {
-    // A new search is triggered whenever the index updates. Consumers of this
-    // function that render their cards using {{#each}} should make sure to use the
-    // "key" field: {{#each #this.results.instances key="id" as |instance|}} in
-    // order to keep the results stable between refreshes.
     isLive?: true;
     // it is probably desirable that dynamic context action `doWithStableScroll()`
     // is used here. For example:
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -1,16 +1,27 @@
+import { registerDestructor } from '@ember/destroyable';
+
 import { service } from '@ember/service';
 import { waitForPromise } from '@ember/test-waiters';
-import { tracked } from '@glimmer/tracking';
+import { tracked, cached } from '@glimmer/tracking';
 
 import { restartableTask } from 'ember-concurrency';
 import { Resource } from 'ember-resources';
 import flatMap from 'lodash/flatMap';
 
-import { type RealmCards } from '@cardstack/runtime-common';
+import { stringify } from 'qs';
+import { TrackedMap } from 'tracked-built-ins';
+
+import {
+  subscribeToRealm,
+  isCardCollectionDocument,
+  SingleCardDocument,
+} from '@cardstack/runtime-common';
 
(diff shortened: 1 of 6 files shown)


packages/host/app/resources/search.ts
```diff
@@ -85,11 +195,32 @@ export function getSearchResults(
   parent: object,
   query: Query,
   realms?: string[],
+  opts?: {
+    // A new search is triggered whenever the index updates. Consumers of this
+    // function that render their cards using {{#each}} should make sure to use the
+    // "key" field: {{#each #this.results.instances key="id" as |instance|}} in
```
lukemelia: Is this actually needed? Aren't we keeping the instances the same on updates via updateFromSerialized?

lukemelia (approved):

packages/host/app/resources/search.ts
habdelra: ah, good point. I just copy pasted this from the previous implementation, but you are right.

packages/host/app/resources/search.ts
```diff
@@ -1,79 +1,189 @@
...
+  private getOrCreateCardResource(
+    urlOrDoc: string | SingleCardDocument,
+  ): CardResource {
+    let url = asURL(urlOrDoc);
+    // this should be the only place we do the untracked read. It needs to be
+    // untracked so our `this.cardResources.set` below will not be an assertion.
+    let resource = this.seenCard(url, false);
+    if (!resource) {
+      resource = getCard(this, () => urlOrDoc, {
+        isLive: () => true,
```
habdelra, line 128: I'm assuming that all cards we get back from search (whether the search is live or not) are live cards. we can easily amend this if that is a bad assumption.

packages/host/app/resources/search.ts
```diff
@@ -1,79 +1,189 @@
...
-      })
-      .filter((r) => r.cards.length > 0);
-
-    this._instancesByRealm = await Promise.all(
-      realmsWithCards.map(async ({ url, cards }) => {
-        let realmInfo = await this.cardService.getRealmInfo(cards[0]);
-        if (!realmInfo) {
-          throw new Error(`Could not find realm info for card ${cards[0].id}`);
-        }
-        return { url, realmInfo, cards };
```
habdelra, line 74: i did slightly change the API for `instancesByRealm`. it felt unnecessary to provide the realm info for each realm in the search results as that is something that can synchronously be obtained by the `RealmService`.

packages/host/app/resources/search.ts
```diff
@@ -1,79 +1,189 @@
...
     this.realmsToSearch = realms ?? this.realmServer.availableRealmURLs;
 
     this.loaded = this.search.perform(query);
     waitForPromise(this.loaded);
+
+    if (isLive) {
+      this.subscriptions = this.realmsToSearch.map((realm) => ({
+        url: `${realm}_message`,
+        unsubscribe: subscribeToRealm(
+          `${realm}_message`,
```
tintinthong: This is a bug. We `subscribeToRealm` to the `realm` not `${realm}_message` to get the auth token from local storage. 

If we use `getCards(..., {isLive: true})` we will get a `Error: Could not find JWT for realm http://localhost:4201/experiments/_message`

tintinthong:
```suggestion
          realm,
```

tintinthong:
I have tested and confirm that newly created cards inside of the search are being indexed and returned by the search resource! 

HOWEVER, the invalidations are still being ignored inside of the card resource (at least inside of the work tracker). Particularly, when I create a new card, I can see that a new card is appearing (with empty vals). Following that, it's typical that I try to edit the fields of that new card (inside edit mode), BUT, the invalidation of that newly created card is ignored bcos of clientRequestIds in the card service as stated [here](https://linear.app/cardstack/issue/CS-7632/cards-in-stackitems-ie-card-backed-by-cardresource-should#comment-87199972) (the previous test I crafted proved that this is more general than the work tracker). **But, this is the problem of the card resource and not the search itself**

tintinthong (approved):
There is one bug to fix that wud block this PR. But, once that is fixed I consider this good. 

Thank you Santa!

packages/host/app/resources/search.ts
habdelra: thanks!

habdelra:
> I have tested and confirm that newly created cards inside of the search are being indexed and returned by the search resource!
> 
> HOWEVER, the invalidations are still being ignored inside of the card resource (at least inside of the work tracker). Particularly, when I create a new card, I can see that a new card is appearing (with empty vals). Following that, it's typical that I try to edit the fields of that new card (inside edit mode), BUT, the invalidation of that newly created card is ignored bcos of clientRequestIds in the card service as stated [here](https://linear.app/cardstack/issue/CS-7632/cards-in-stackitems-ie-card-backed-by-cardresource-should#comment-87199972) (the previous test I crafted proved that this is more general than the work tracker). **But, this is the problem of the card resource and not the search itself**

yeah, @ef4 and I felt that the way that we were tracking our local echoes using clientRequestIds was very dubious, since they were being issued and being checked in very different contexts. let me resurrect that test and work that issue

PR #1929, opened 2024-12-12
Convert MatrixService#sendAiAssistantMessage to a command

lukemelia:


packages/host/app/services/matrix-service.ts
```diff
@@ -534,8 +530,8 @@ export default class MatrixService extends Service {
   async addCardsToRoom(
     cards: CardDef[],
     roomId: string,
-    cardHashes: Map<string, string>,
-    opts?: CardAPI.SerializeOpts,
+    cardHashes: Map<string, string> = this.cardHashes,
```
IanCal, line 533: Not required in this pr but I think we can just fully remove this as an argument - the default is what we use for everything except skills but I don't see why skills need to be their own map. If I attach a skill, we can still just use the same event id for the skill card list, there's no important distinction.

packages/host/app/services/matrix-service.ts
```diff
@@ -534,8 +530,8 @@ export default class MatrixService extends Service {
...
+    opts: CardAPI.SerializeOpts = { maybeRelativeURL: null },
```
IanCal, line 534: Similarly we could replace this with whether we should include computeds, that's the only serialisation option we distinguish between. Not key for this PR but could be a good cleanup.

IanCal:
*inception noises* 
image

IanCal (approved):

PR #1935, opened 2024-12-13
Add more fields to BlogPost card

FadhlanR:
Screenshot 2024-12-13 at 16 12 26

### DIFF
--- a/packages/experiments-realm/blog-post.gts
+++ b/packages/experiments-realm/blog-post.gts
@@ -560,7 +560,6 @@ export class BlogPost extends CardDef {
   });
   @field blog = linksTo(BlogAppCard, { isUsed: true });
   @field featuredImage = contains(FeaturedImageField);
-  @field blog = linksTo(BlogAppCard);
   @field lastUpdated = contains(DatetimeField, {
     computeVia: function (this: BlogPost) {
       let lastModified = getCardMeta(this, 'lastModified');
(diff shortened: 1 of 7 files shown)


packages/experiments-realm/blog-post.gts
```diff
@@ -607,7 +637,15 @@ export class BlogPost extends CardDef {
               <li class='pub-date'>
                 Published on
                 <time timestamp={{@model.datePublishedIsoTimestamp}}>
-                  {{this.formattedDatePublished}}
+                  {{@model.formattedDatePublished}}
```
tintinthong, line 640: I've never seen this pattern before and I don't know why I always assumed this to be wrong. 

You use a getter at the root of the CardDef and you use @model.formattedDatePublished to access the model value. This is great bcos you can share these across formats. But, is this different than a computed or not -- do you know? Typically, all my getters exist on the format template. 

In general, this is not a change request. Just a curiosity. I thought perhaps the types will fail or something

FadhlanR: I'm unsure if this is the correct pattern or not. However, it works, and I notice the same pattern in a few cards, such as catalog-entry and leaflet-gtfs. Are you referring to a computed field? This is different from a computed field.

burieberry: I've seen it used before too and it made sense to me. If the getter only applies to a specific format class, then it can be there, but if it applies to all formats, then it can go on the card class I think.

burieberry (approved):

PR #1938, opened 2024-12-16
Cs 7659 currency & amount with currency field

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7659/create-currency-amt-field

### What is changing
- add currency field based on the list ready in the `currency-code-symbol-map` package (via CDN)
- add the amount with the currency field 

**Screenshots**
Screenshot 2024-12-16 at 4 06 26 PM
Screenshot 2024-12-16 at 4 06 33 PM

### DIFF
--- a/packages/experiments-realm/fields/amount-with-currency.gts
+++ b/packages/experiments-realm/fields/amount-with-currency.gts
@@ -6,17 +6,13 @@ import { action } from '@ember/object';
 import { BoxelInputGroup } from '@cardstack/boxel-ui/components';
 import { guidFor } from '@ember/object/internals';
 
-// @ts-ignore
-import { getSymbolFromCode } from 'https://esm.run/currency-code-symbol-map';
-
 class View extends Component<typeof AmountWithCurrency> {
   get formatNumberWithSeparator() {
-    const num = this.args.model.amount;
-    const currencyCode = this.args.model.currency?.code;
-    const currencySymbol = getSymbolFromCode(currencyCode) || '';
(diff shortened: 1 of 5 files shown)


packages/experiments-realm/fields/amount-with-currency.gts
```diff
@@ -0,0 +1,119 @@
...
+        vertical-align: bottom;
+        width: var(--boxel-icon-sm);
+        height: var(--boxel-icon-sm);
+      }
+    </style>
+  </template>
+}
+
+export class AmountWithCurrency extends FieldDef {
+  static displayName = 'AmountWithCurrency';
```
tintinthong: I think display name shud be with spaces so it looks good where displayName is used

tintinthong:
can we default to USD somehow?

packages/experiments-realm/fields/amount-with-currency.gts
```diff
@@ -0,0 +1,115 @@
...
+    let newModel = new AmountWithCurrency();
+    newModel.amount = val;
+    newModel.currency.code = newModel.currency.code || 'USD';
+    this.args.set(newModel);
+  }
+
+  @action
+  setCurrency(val: CurrencyField) {
+    let newModel = new AmountWithCurrency();
+    newModel.amount = this.args.model.amount as number;
```
richardhjtan, line 42: @tintinthong lint is complaining to have a type on `amount`, so I keep the changes here

tintinthong (approved):

PR #1939, opened 2024-12-16
Cs 7682/enhance entity display with implement css var and replace title args

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7682/enhance-entity-display-with-implement-css-var-and-replace-title-args

### DIFF
--- a/packages/experiments-realm/components/entity-display.gts
+++ b/packages/experiments-realm/components/entity-display.gts
@@ -1,13 +1,14 @@
 import GlimmerComponent from '@glimmer/component';
 interface EntityDisplayArgs {
   Args: {
-    name?: string | null;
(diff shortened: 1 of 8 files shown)


packages/experiments-realm/components/entity-display.gts
```diff
@@ -1,13 +1,14 @@
 import GlimmerComponent from '@glimmer/component';
 interface EntityDisplayArgs {
   Args: {
-    name?: string | null;
     center?: boolean;
     underline?: boolean;
   };
   Blocks: {
+    title: [];
```
lucaslyl, line 8: Replace the name argument with a title block for greater flexibility in the future.

PR #1940, opened 2024-12-16
create activity-card component

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7680/activity-card-component

- create activity card reusable component


burieberry (commented):
Can you post some screenshots for how this component looks?

lucaslyl:
> Can you post some screenshots for how this component looks?

@burieberry ,I create an example card with mock data that uses this component 
Screenshot 2024-12-18 at 13 21 35

But, we are facing an issue with the entity-display component. The component includes a yield block for the thumbnail, but the thumbnail accepts various types of data, such as icons or avatars. The problem is that the avatar thumbnail has default width and height settings, which can potentially disrupt the visual layout of the entity-display, as shown in the screenshot below. Perhaps you could advise us on how to handle this scenario? or can we set a meeting to discuss this case @tintinthong @burieberry 
Screenshot 2024-12-18 at 13 24 14

lucaslyl:
We would be happy to merge this in first to ensure non-blocking development progress. Please feel free to bring up any additional requirements if needed.

PR #1942, opened 2024-12-16
Use the correct realm when creating cards in blog post app

jurgenwerk:
image

Fixes a bug where creating a post, author, or category put the newly created card into the default realm instead of the realm where the blog post app lives (currently in "experiments")

### DIFF
--- a/packages/experiments-realm/blog-app.gts
+++ b/packages/experiments-realm/blog-app.gts
@@ -399,7 +399,6 @@ class BlogAppTemplate extends Component<typeof BlogApp> {
         },
         meta: {
           adoptsFrom: ref,
-          realmURL: currentRealm.href,
         },
       },
     };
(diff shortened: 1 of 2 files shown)


richardhjtan (approved):

burieberry (approved):

packages/experiments-realm/blog-app.gts
```diff
@@ -399,6 +399,7 @@ class BlogAppTemplate extends Component<typeof BlogApp> {
         },
         meta: {
           adoptsFrom: ref,
+          realmURL: currentRealm.href,
```
lukemelia: Is it a smell that we need to specify the realm 3 times between lines 402 and 407?

lukemelia (approved):

packages/experiments-realm/blog-app.gts
jurgenwerk: Yeah it is, actually the `realmURL` in opts is there specifically for defining the target realm but the code in the card api didn't respect it in some cases. Fixed it in https://github.com/cardstack/boxel/pull/1942/commits/3d375c80927c9768adb832785367b19c093843c0 and removed the redundant `realmURL`

PR #1943, opened 2024-12-16
Fix for realm write performance degradation after deploy

habdelra:
This issue is 2 fold:
1. all the mtimes of the shared realm files where being modified on each deploy via the `cp` command thus rendering our logic to skip unchanged files moot.
2.  the session token was not being shared for HTTP GET requests, so every HTTP GET request required logging into the realm.

This PR addresses those 2 issues by using `rsync` to preserve mtimes of the unchanged shared realm files. and fixes a bug in the `RealmAuthDataSource` that was not returning a token when the request for a url matched `this.realmURL`. Really there is no need for `this.realmURL` in `RealmAuthDataSource`. the matrix client for the `RealmAuthDataSource` has all the necessary logic for the realm user. no particular realm should get any special treatment in  `RealmAuthDataSource`.


packages/runtime-common/realm-auth-data-source.ts
```diff
@@ -6,28 +6,22 @@ export class RealmAuthDataSource {
...
   }
 
   token(url: string): string | undefined {
     let targetRealmURL = this.toRealmURL(url);
     if (!targetRealmURL) {
       return undefined;
     }
-    if (targetRealmURL === this.realmURL) {
-      return undefined;
-    }
```
habdelra, line 30: this was the crucial logic that was blowing away our session token for each HTTP GET. there is no need for this, the worker has no special realm preference. I think this made sense when the worker lived on the realm. but now that workers are their own separate process we need to treat them as arms length away from any particular realm

lukemelia (approved):

habdelra:
hmmm, looks like `rsync` alone is not gonna cut it, the source files that we copy from get a new mtime on each deploy too via docker...

lukemelia (approved):

PR #1944, opened 2024-12-16
Add support for forcing tool use when you programatically

IanCal:


IanCal:
Note - I tried adding tests about the error, but had the same issue as before with uncaught exceptions because we throw in two ways. I'll put up a PR with a clearer case where this happens but didn't want to hold this up because of that.

packages/ai-bot/helpers.ts
```diff
@@ -293,31 +295,55 @@ export function getRelevantCards(
...
+
+export function getToolChoice(
+  history: DiscreteMatrixEvent[],
+  aiBotUserId: string,
+): ToolChoice {
+  let lastMessage = getLastUserMessage(history, aiBotUserId);
+  if (lastMessage && lastMessage.content.data?.context?.requireToolCall) {
+    let tools = lastMessage.content.data.context?.tools || [];
+    if (tools.length != 1) {
+      throw new Error('Forced tool calls only work with a single tool');
```
lukemelia, line 344: Is this recoverable, or does it "break" a room forever?

lukemelia (approved):

packages/ai-bot/helpers.ts
IanCal: I'll test this before merging, with the check on the host we shouldn't get into this situation anyway but with the current setup this should be temporary as the list of commands and the requiredtoolcall is just on the last user message so the next valid one sent will be fine

PR #1945, opened 2024-12-17
Add categories to blog posts

jurgenwerk:
This PR adds a blog category. Here are a couple of templates:

Isolated and embedded:
image

Fitted:
image

In blog post:

image

image

In blog app:

image


burieberry:
The color picker relates to CS-7513

packages/experiments-realm/blog-app.gts
```diff
@@ -303,17 +279,95 @@ class BlogAppTemplate extends Component<typeof BlogApp> {
         font: 600 var(--boxel-font-lg);
         letter-spacing: var(--boxel-lsp-xxs);
       }
+      :deep(.card-view.categories-grid) {
+        --grid-card-height: 150px;
```
jurgenwerk: This is to make the category cards in the blog app shorter than posts and author bios. Otherwise there is too much vertical space

packages/experiments-realm/blog-app.gts
```diff
@@ -303,17 +279,80 @@ class BlogAppTemplate extends Component<typeof BlogApp> {
...
+
+    let makeQuery = (codeRef: ResolvedCodeRef) => {
+      if (!blogId) {
+        throw new Error('Missing blog id');
+      }
+
+      return {
+        filter: {
+          on: codeRef,
+          eq: { 'blog.id': blogId },
```
jurgenwerk, line 294: Previously, this filter was searching by relative urls, but that triggers the WAF error in staging/prod, because it doesn't like relative paths in url queries (it detects it as a security threat). 

So I changed it to searching by absolute urls, but that will only start working when we land https://github.com/cardstack/boxel/pull/1955

packages/experiments-realm/blog-app.gts
lukemelia: Is there a way to do this without using :deep and reaching down? CSS variables cascade

lukemelia (approved):

packages/experiments-realm/blog-app.gts
burieberry: this can potentially move to BlogApp isolated template styles, by using `...attributes` on cards-grid component and passing in a class name

burieberry (approved):
Looks great. On the BlogApp home page, it would look better if the container boundaries also were displayed

burieberry:
@jurgenwerk I've just realized that there's confusion between category and tags. I marked them below. Up to you if you want to merge this in and create another ticket to update the designs, or make the changes in this one.

category
isolated-post
isolated

packages/experiments-realm/blog-app.gts
lukemelia: The fix has landed

burieberry:
Resolved merge conflicts

packages/experiments-realm/blog-app.gts
jurgenwerk: 🚀

jurgenwerk:
@burieberry that's right looks I was confused with tags and categories. I'll fix it here in this PR

jurgenwerk:
New design adjustments: 

image

image

image

image

image

PR #1946, opened 2024-12-17
Echo detection logic removal

habdelra:
This PR removes the echo detection logic. ultimately this was a hacky solution to deal with the fact that we have no store. probably this creates more problems than it solves.

Background:
In `CardService.saveModel()` (which drives the card context API's `save()` function) we attach a temporary listener to the card while the save is in flight to see if the card is mutated locally while we are saving. if the card is mutated locally during the in-flight save request then we will not update the card with the server's save response, and rather the next auto save that doesn't collide with local changes will update our card after the save.

If there was only ever a single instance of a card on the host at a time, this would be fine. The problem, though, is that it is still possible to have _multiple_ instances of the _same_ card living in the host runtime. These cards are not live bound to one another--changing one of them does not change all of them. so there are multiple sources of truth for what the current local value of the card is. It's only the `CardResource` that has the closest way of dealing with a single instance of a card via its identity map. Any other way of obtaining a card will be fraught with issues (this goes back to not having a store). The echo cancellation logic was a hacky solution to deal with the fact that there are lots of instances of the same card, and to try to prevent collisions between these instances by trying to detect the collision and prevent it from altering one of among many of the same instances. Really though, this is just kicking the can down the road. the collision doesn't go away, it's just pushed further in the future. there really is no way for us to reconcile all the different states of the same instances that may live on the host. and the echo cancellation actively gets in the way of the CardResource trying to do the right thing.


jurgenwerk:
Should we also delete the header from https://github.com/cardstack/boxel/blob/b7cfbfadca5ccb8d374d337b93799ce04380ff74/packages/runtime-common/realm.ts#L1331-L1332 and https://github.com/cardstack/boxel/blob/b7cfbfadca5ccb8d374d337b93799ce04380ff74/packages/realm-server/server.ts#L136?

habdelra:
> Should we also delete the header from
> 
> https://github.com/cardstack/boxel/blob/b7cfbfadca5ccb8d374d337b93799ce04380ff74/packages/runtime-common/realm.ts#L1331-L1332
> 
> and
> https://github.com/cardstack/boxel/blob/b7cfbfadca5ccb8d374d337b93799ce04380ff74/packages/realm-server/server.ts#L136
> 
> ?

thanks, yes!

lukemelia:
Should I be surprised that no tests fail as a result of this change?

habdelra:
> Should I be surprised that no tests fail as a result of this change?

no, i don't think we had any tests around this in the first place

tintinthong:
I wonder if this fixes https://linear.app/cardstack/issue/CS-7574/code-mode-autosave-bug

lukemelia:
Tried this branch out and the user experience is markedly worse: https://www.loom.com/share/615bcc3e250c42c286ae5c00ea3a3d8d?sid=3eb3e96d-caac-4f70-97d3-ca91ebeeba74

habdelra:
> Tried this branch out and the user experience is markedly worse: https://www.loom.com/share/615bcc3e250c42c286ae5c00ea3a3d8d?sid=3eb3e96d-caac-4f70-97d3-ca91ebeeba74

Hmmm, maybe we just need to hold off on this until we have some kind of store solution…

PR #1947, opened 2024-12-17
Update command results to be the result cards of command executions

lukemelia:
- Change how command results are stored in Matrix: old way: ReactionEvent + optional CommandResultEvent; new way: CommandResultEvent with optional data.content.cardEventId property that can point to a matrix-stored result card
- Update aibot to read new result event
- Update apply button state to read new result event
- Replace special casing of search command result with rendering result card in embedded format in AI panel
- Introduce CopyCard host command and refactor "copy to workspace" and other copy features to use it


packages/runtime-common/matrix-constants.ts
```diff
@@ -2,6 +2,7 @@ export const APP_BOXEL_CARDFRAGMENT_MSGTYPE = 'app.boxel.cardFragment';
 export const APP_BOXEL_MESSAGE_MSGTYPE = 'app.boxel.message';
 export const APP_BOXEL_COMMAND_MSGTYPE = 'app.boxel.command';
 export const APP_BOXEL_CARD_FORMAT = 'app.boxel.card';
+export const APP_BOXEL_COMMAND_RESULT_EVENT_TYPE = 'app.boxel.commandResult';
 export const APP_BOXEL_COMMAND_RESULT_MSGTYPE = 'app.boxel.commandResult';
```
IanCal: Do we still need a msgtype?

If we do, should we use this to distinguish between filled and undefined results? The event type says it's a command result, but the msgtype could be how to parse that (e.g. linked card or nothing). Notably we don't seem to use the msgtype on the bot side.

IanCal:
I like that responses with undefined don't trigger a response, that seems like a reasonable way to control some of this.

I failed to get this working with a command that returns a card defined in the same file. It's not known to the loader, and not in "localIdentities" so it fails:

image

Code https://gist.github.com/IanCal/78191aa82bfb51af2e88b81ca50b7234

lukemelia:
> I failed to get this working with a command that returns a card defined in the same file. It's not known to the loader, and not in "localIdentities" so it fails:

Hopefully will be fixed for cases where the command result card type is exported by this: https://github.com/cardstack/boxel/pull/2004. We should still improve the error message for cases where there is a missing export.

packages/runtime-common/matrix-constants.ts
lukemelia: Good idea. Updated this.

PR #1950, opened 2024-12-18
Make author field a linksToMany

FadhlanR:
https://github.com/user-attachments/assets/1b2bdfe7-bee3-4de3-910f-c41d7f8ae425


packages/experiments-realm/author.gts
```diff
@@ -302,7 +302,9 @@ export class Author extends CardDef {
         {{else}}
           <UserIcon class='author-icon' width='20' height='20' />
         {{/if}}
-        <@fields.title />
+        <span class='autor-title'>
```
burieberry: typo in class name -- doesn't match the name in stylesheet

packages/experiments-realm/blog-post.gts
```diff
@@ -626,15 +641,12 @@ export class BlogPost extends CardDef {
...
-                  @displayContainer={{false}}
-                />
-              </li>
-            {{/if}}
+            <li class='byline'>
+              {{#each @fields.authors as |AuthorComponent index|}}
+                <AuthorComponent @format='atom' class='author' />
+                <span>{{this.additionalTextForAuthorComponent index}}</span>
+              {{/each}}
+            </li>
```
burieberry: I think this will need a `{{if @model.authors.length}}` wrapper around it, otherwise blank space shows up on the template when there are no authors

packages/experiments-realm/review.gts
```diff
@@ -14,6 +14,21 @@ export class Review extends BlogPost {
...
+
+    if (titles.length === 2) {
+      return `${titles[0]} and ${titles[1]}`;
+    }
+
+    return titles.length > 2
+      ? `${titles.slice(0, -1).join(', ')}, and ${titles.at(-1)}`
+      : titles[0];
+  }
+
```
burieberry: If you put this on the BlogApp card definition, wouldn't need to repeat it here

burieberry (commented):
Looks good, suggested a couple of improvements

packages/experiments-realm/blog-post.gts
```diff
@@ -607,6 +612,44 @@ export class BlogPost extends CardDef {
...
+      text = text + ' ';
+    }
+
+    if (isLastTwo) {
+      text = text + 'and';
+    }
+
+    return text;
+  };
+
```
jurgenwerk: Was this in the design, I mean commas and "and"? Adding "," and "and" is not a good way of delimiting items where there are images or website elements involved. I think it's unnecessary and it should only be used in prose or copy.

image

packages/experiments-realm/blog-app.gts
```diff
@@ -202,13 +202,22 @@ class BlogAdminData extends GlimmerComponent<CardAdminViewSignature> {
...
+    const formattedEditors = editors.map((editor) =>
+      editor.email ? `${editor.name} (${editor.email})` : editor.name,
+    );
+
+    if (formattedEditors.length === 2) {
+      return `${formattedEditors[0]} and ${formattedEditors[1]}`;
+    }
+
+    return formattedEditors.length > 2
+      ? `${formattedEditors.slice(0, -1).join(', ')}, and ${formattedEditors.at(
```
jurgenwerk: I can't find any design for this in your ticket but I think "and" is unnecessary

packages/experiments-realm/blog-post.gts
FadhlanR: We don't have design for this but the commas and "and" added based on discussion with @lukemelia in our 1-on-1 session.

packages/experiments-realm/blog-app.gts
FadhlanR: reverted this one.

packages/experiments-realm/blog-post.gts
jurgenwerk: I see, I think it would look better without the noise.

image

lukemelia: OK, let's go with @jurgenwerk's suggestion

PR #1952, opened 2024-12-18
Fix sorting by PublishDate field

burieberry:
Sorting by a field requires specifying the card type, so I had to move sortOptions to the card definition. ReviewBlog subclass has its own filters, made adjustments accordingly.


burieberry:
This will cause conflict with your categories PR @jurgenwerk, I can wait to merge or update your pr.

lukemelia (approved):

PR #1953, opened 2024-12-18
[BREAKING] Update our custom matrix identifiers from org.boxel to app.boxel

lukemelia:
- Once merged, old messages will not appear and user realms will need to be set again


IanCal:
I've not had a chance to run this yet, but the changes look good and I like the move to consts.

I wondered if it makes sense to have something to auto migrate realms? Like get the current realms and if they're empty, check the old key and overwrite?

lukemelia:
> I wondered if it makes sense to have something to auto migrate realms? Like get the current realms and if they're empty, check the old key and overwrite?

Do you mean realms or rooms?

IanCal:
> > I wondered if it makes sense to have something to auto migrate realms? Like get the current realms and if they're empty, check the old key and overwrite?
> 
> Do you mean realms or rooms?

The realms a user has access to, so checking the account data for the new key, if it's not present copying it from the old key to the new one. Migrating the stuff just in the account data.

FadhlanR (approved):

IanCal:
@lukemelia proposed a migration step for the list of realms here: https://github.com/cardstack/boxel/pull/1959 to smooth over the change.

IanCal (approved):
LGTM, an import didn't work in the aibot so I changed it to pull from runtime common directly (other option would have been doing a re-export I think). Seems to work fine and I can't find missing updates in the rest of the code.

PR #1955, opened 2024-12-19
Serialize card into index using absolute URLs for linksTo/linksToMany and CodeRef

lukemelia:


lukemelia:
Closing in favor of #1958

PR #1956, opened 2024-12-19
Make activity card more pure presentational

richardhjtan:
### What is changing

We don't have an indicator to know whether the entity thumbnail is an image or an icon, the optimal way is to do the styling handling at the level we use the entity display component.

Hence, we make the activity card more presentational by displaying what it gives, and controlling the styling at the account card instead.

With this approach, we can control the image is not getting shrinked down:

Screenshot 2024-12-19 at 12 15 58 PM


richardhjtan:
Close as for demo and we can make further discussion on improvements

PR #1959, opened 2024-12-19
Org boxel to app boxel migration

IanCal:
Proposing adding a migration step, when logging in for the first time / using the app it will :+1: 
* Check for realms data under the new key
* If there is some carry on
* If not, check the old key
* * If in there, copy those to the new key
* * remove the old account data key (set to an empty object)


lukemelia:
Incorporated into #1953

PR #1960, opened 2024-12-20
Absolute urls on work tracker

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7725/work-tracker-change-url-to-absolute

### What is changing 
- change the work tracker app's relative URL to the absolute URL to avoid WAF errors


tintinthong (approved):
Yes tq!

PR #1961, opened 2024-12-20
Cs 7568 deals fitted (crm strip view)

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7568/deals-fitted-crm-strip-view

This is the pr for the CRM deal (for strip view only)

result:
image


richardhjtan (approved):

tintinthong (approved):
This is good to me. Note: not all fitted is handled but it wud be a progression in a future PR. But this is a good foundation

PR #1962, opened 2024-12-20
Fix word count logic

FadhlanR:


burieberry (commented):
This is better, but since the body is a markdown field, this will also count all the markdown syntax and some of the html elements. I think we should get the text content from the dom for the article body and do the count there.

FadhlanR:
> This is better, but since the body is a markdown field, this will also count all the markdown syntax and some of the html elements. I think we should get the text content from the dom for the article body and do the count there.

Thanks for the feedback! I wasn't aware that the body is a markdown field. I've updated the logic.

burieberry (approved):
Great

PR #1964, opened 2024-12-20
Fix safari icon bug

burieberry:
A relatively recent change in IconButton component was setting svg width and height to `auto`, which translated to 0 for some browsers, such as Safari.

Before:
icon-bug

After:
icon-fixed


jurgenwerk:
Percy has an option to include Safari. Should we enable it?

image

jurgenwerk (approved):

burieberry:
> Percy has an option to include Safari. Should we enable it?
> 
> image

Good question. What do you think @lukemelia ? One thing to consider is we're already going over our limit for percy so we may need to increase that too.

PR #1970, opened 2024-12-23
Aggregate tools so they remain available to call by the aibot

IanCal:
We plan to change how we enable/disable tools, but for now this avoids the problem of the bot being unaware of something it had one message ago.

Tools are pulled from all messages, and we assume the names must be unique (the same name appearing more than once doesn't break, but we keep the last one we see).

If "requireToolCall" is set to true, we set the required tool call to only that one.

If the last message from the user is not a regular message (e.g. is a reaction) we don't force tool calls so that we don't get into loops.

Once we have returning data, the system becomes significantly more useful

### DIFF
--- a/packages/ai-bot/helpers.ts
+++ b/packages/ai-bot/helpers.ts
@@ -6,7 +6,6 @@ import {
 import { ToolChoice } from '@cardstack/runtime-common/helpers/ai';
 import type {
   MatrixEvent as DiscreteMatrixEvent,
-  CardMessageEvent,
   CardFragmentContent,
   CommandEvent,
   CommandResultEvent,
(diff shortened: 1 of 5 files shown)


packages/ai-bot/helpers.ts
```diff
@@ -302,44 +301,51 @@ export function getRelevantCards(
...
   aiBotUserId: string,
 ): ToolChoice {
-  let lastMessage = getLastUserMessage(history, aiBotUserId);
-  if (lastMessage && lastMessage.content.data?.context?.requireToolCall) {
-    let tools = lastMessage.content.data.context?.tools || [];
+  const userMessages = history.filter((event) => event.sender !== aiBotUserId);
+  if (userMessages.length === 0) {
+    // If the user has sent no messages, auto is safe
+    return 'auto';
+  }
```
lukemelia: Try a reverse walk through history to find the most recent user message

lukemelia (approved):
👍

PR #1971, opened 2024-12-23
Show card type icon of CardsGird's filter list

FadhlanR:
Screenshot 2024-12-23 at 13 05 25

The idea is to capture the card type icon during indexing and include it in the card type summary in the realm meta table. To achieve this, I added a new render function in `RenderService`, allowing us to render the card type icon in FastBoot.


packages/base/cards-grid.gts
```diff
@@ -493,3 +497,21 @@ export class CardsGrid extends CardDef {
...
+    .replace(/\swidth="[^"]*"/, '')
+    .replace(/\sheight="[^"]*"/, '')
+    .replace('<svg', '<svg style="width: 100%; height: 100%;"');
+  return class extends GlimmerComponent<Signature> {
+    <template>
+      <div ...attributes>
+        {{htmlSafe updatedIconHtml}}
+      </div>
+    </template>
+  };
```
lukemelia: I don't think there is any point to generating a component for this. Perhaps instead we can pass along summary.attributes.iconHTML as is and make a helper that does the replace transformation from lines 507-509?

packages/host/app/components/card-prerender.gts
```diff
@@ -106,6 +106,7 @@ export default class CardPrerender extends Component {
       reader,
       indexWriter,
       renderCard: this.renderService.renderCard.bind(this.renderService),
+      render: this.renderService.render.bind(this.renderService),
```
lukemelia: I know this change just continued the pattern above, but let's change these methods to bind using fat arrows so we don't have to call `bind` here.

packages/host/app/components/card-prerender.gts
```diff
@@ -128,6 +129,7 @@ export default class CardPrerender extends Component {
         indexWriter,
         ignoreData: { ...ignoreData },
         renderCard: this.renderService.renderCard.bind(this.renderService),
+        render: this.renderService.render.bind(this.renderService),
```
lukemelia: Same thing here.

backspace:
I merged `main` into this, it was confusing because the Github UI wasn’t showing a merge conflict but I did see conflicts with the schema file. Hopefully I did it correctly!

packages/base/cards-grid.gts
FadhlanR: I updated the filter list component in Boxel UI so we can pass along the HTML string of card type icon. Please check it again.

packages/host/app/components/card-prerender.gts
FadhlanR: Updated

packages/boxel-ui/addon/src/components/filter-list/index.gts
```diff
@@ -68,11 +81,26 @@ export default class FilterList extends Component<Signature> {
...
+      :global(.filter-list__icon) {
         width: var(--boxel-icon-xs);
         height: var(--boxel-icon-xs);
         vertical-align: top;
       }
     </style>
   </template>
 }
+
+function convertSVGString(svgString: string, className: string) {
```
jurgenwerk: Probably a better name for this function would be `addClassToSVG`?

packages/host/config/schema/1735668047598_schema.sql
```diff
@@ -23,6 +23,7 @@
    fitted_html BLOB,
    display_names BLOB,
    resource_created_at,
+   icon_html TEXT,
```
jurgenwerk, line 26: The internal convention we have is to delete the old sql schema file, keeping only the latest one

jurgenwerk (approved):

richardhjtan (approved):

packages/postgres/migrations/1734686216941_prerendered-card-type-icon.js
```diff
@@ -0,0 +1,13 @@
...
+
+exports.up = (pgm) => {
+  pgm.addColumns('boxel_index', {
+    icon_html: 'varchar',
+  });
+};
+
+exports.down = (pgm) => {
+  pgm.dropColumns('boxel_index', ['icon_html']);
+};
```
habdelra, line 13: I'll make sure to capture column this in my index refactor work as well

habdelra (approved):

packages/host/app/components/card-prerender.gts
```diff
@@ -129,7 +129,7 @@ export default class CardPrerender extends Component {
         indexWriter,
         ignoreData: { ...ignoreData },
         renderCard: this.renderService.renderCard.bind(this.renderService),
-        render: this.renderService.render.bind(this.renderService),
+        render: this.renderService.render,
```
lukemelia: Let's make the same change for `renderCard`

habdelra:
there seem to be percy differences that dont have anything to do with this PR. probably you need to merge upstream in order to clean up the percy differences to just your changes. Also, please fix the lint issue too

habdelra:
@FadhlanR if you are unable to merge this by wed morning I'll just merge the index refactoring and i'll update this PR with the logic necessary after that PR lands.

PR #1977, opened 2024-12-23
Move BlogApp files to Seed Realm

burieberry:


packages/experiments-realm/BlogPost/ultimate-guide-remote-work.json
```diff
@@ -29,29 +29,14 @@
           "self": "../BlogApp/ramped"
         }
       },
-      "categories.0": {
```
lukemelia, line 32: Was the .0 .1 thing an old format?

lukemelia (approved):
Looks good once green

packages/experiments-realm/BlogPost/ultimate-guide-remote-work.json
burieberry: if value is null, it's `categories` only

lukemelia: Roger

PR #1980, opened 2024-12-24
Rename work tracker entities 

tintinthong:


packages/experiments-realm/task.gts
```diff
@@ -182,11 +180,14 @@ export class FittedTask extends Component<typeof TaskBase> {
       </div>
 
       <footer>
-        <@fields.assignee
-          class='card-assignee'
-          @format='atom'
-          @displayContainer={{false}}
-        />
+        {{!TODO: Actually, this shudn't be needed. atom default templates can't handle null values }}
```
lukemelia, line 183: Please ensure a bug is opened for this

lukemelia (approved):

packages/experiments-realm/task.gts
tintinthong: https://linear.app/cardstack/issue/CS-7734/default-atom-template-cannot-handle-model=null

PR #1982, opened 2024-12-24
Remove product with video and ratings

tintinthong:
Identical with video product I dono why there was issues creating productVideoWithRatings this instance but video product cud be created

The error was `worker: encountered error indexing card instance ProductWithVideoAndRatings/76529ed3-1c08-4644-9f6f-c392f9254b32.json: bug: could not identify card: Object`

There is already a catalog entry for video product


lukemelia (approved):

burieberry (approved):
You can remove but the difference is this one also has a ratings-summary field. Is there an error in ratings-summary?

tintinthong:
This errors out bcos you don't have `ratingsSummary` inside of the json field data when creating a card since `ratingsSummary` is a computed. I think there is an opportunity to understand why this problem occurs so I am investigating this

PR #1983, opened 2024-12-24
Copy scripts to seed realm

tintinthong:
This PR is intended to provide scripts that ease the developer into moving a file from seed realm to experiments realm

The usage is like this 

`pnpm copy-gts ./folder-with-files` 

`pnpm copy-gts ./a-file.gts` 

`pnpm copy-json ./folder-with-files` 

`pnpm copy-gts ./a-file.json` 

- the script will copy over all new files to an identical path at the root of the seed realm
- If there are diffs in the file, the script will not write that file. If you want to just write, you can use `copy-gts:force` `copy-json:force`

### DIFF
--- a/packages/experiments-realm/package.json
+++ b/packages/experiments-realm/package.json
@@ -18,7 +18,11 @@
   "scripts": {
     "lint": "concurrently \"pnpm:lint:*(!fix)\" --names \"lint:\"",
     "lint:hbs": "ember-template-lint .",
-    "lint:hbs:fix": "ember-template-lint . --fix"
+    "lint:hbs:fix": "ember-template-lint . --fix",
+    "copy-gts": "./scripts/copy-to-seed.sh gts",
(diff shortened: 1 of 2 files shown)


lukemelia:
Please explain in this PR description why this work is needed and how it is different from the package scripts that run rsync in realm-server package.

tintinthong:
> Please explain in this PR description why this work is needed and how it is different from the package scripts that run rsync in realm-server package.

 I don't think these copy scripts have necessarily have a strong use-case as compared to rsync. It provides "diff" outputs but those might be noisy. I think fundamentally we don't need the scripts. 

What we need to rsync seedRealm -> experimentsRealm and be signalled that a author.gts in seedRealm has a diff with experimentsRealm. That use-case will be a more useful utility of signalling to us that something in experiments has changed. 

@lukemelia shud I just add a simple script to seed realm package to rsync --dry-run or just close this PR

PR #1985, opened 2024-12-24
Update default atom template styles

burieberry:
- updated atom template styles for each field type:
  - linksTo and linksToMany atom fields have `displayContainer={{true | false}}` option
  - atom field is inline, inline-block, or display: contents; based on field type/mode
- updated atom template for edit mode (use Pill component and fix spacing between text and remove button)
- better field alignments for edit mode

Card with all atom template types:
### Before Changes:
- atom template is not inline
- container is not always controllable
before-1
before-2
before-3
before-4

### After Changes:
atom-1
atom-2

Edit mode better alignment:
edit-view-alignments

Nested atom field edit mode style fixes:
edit-view-nested-field

### DIFF
--- a/packages/base/contains-many-component.gts
+++ b/packages/base/contains-many-component.gts
@@ -70,6 +70,8 @@ class ContainsManyEditor extends GlimmerComponent<ContainsManyEditorSignature> {
           <AddButton
             class='add-new'
             @variant='full-width'
+            @iconWidth='12px'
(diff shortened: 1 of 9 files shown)


lukemelia (approved):

lukemelia:
Looking good!

jurgenwerk (approved):

PR #1986, opened 2024-12-25
intoduce --delete arg to rsync. Deleted files will be removed from persistent

tintinthong:


lukemelia (changes requested):
I think we don't want to do this with the experiments realm. If we do, changes made by us will be wiped out every deploy. Seems OK for base, seed, & catalog.

habdelra:
yeah, this is what I was thinking--but we should make it an option like @lukemelia was saying so that in the experiements realm we don't blow away stuff that was added outside of git

tintinthong:
I have updated this to only use --delete with catalog, seed, base.

richardhjtan (approved):

backspace (approved):

lukemelia (approved):

PR #1987, opened 2024-12-26
implement filter query of side panel

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7722/implement-account-side-panel-design

- Side panel filter list for CRM account
- Filter by value
- add icon to urgency tags

Screenshot 2024-12-27 at 17 01 49


packages/experiments-realm/crm-app.gts
```diff
@@ -73,7 +74,55 @@ const DEAL_FILTERS: LayoutFilter[] = [
 const ACCOUNT_FILTERS: LayoutFilter[] = [
```
tintinthong, line 71: Can we get all the data from one place? Like the exclude status field rather than making a new copy. I already had the thought of chnaging the list of field values but I have to change it in 2 places

packages/experiments-realm/crm-app.gts
```diff
@@ -265,6 +331,7 @@ class CrmAppTemplate extends Component<typeof AppCard> {
     this.selectedView = id;
   }
   @action private onSort(option: SortOption) {
+    console.log('option', option);
```
tintinthong: Remove

packages/experiments-realm/crm-app.gts
```diff
@@ -218,38 +267,55 @@ class CrmAppTemplate extends Component<typeof AppCard> {
...
+      ? [
+          {
+            any: [
+              {
+                on: activeFilter.cardRef,
+                contains: { name: searchKey },
+              },
+              {
+                on: activeFilter.cardRef,
+                contains: { 'company.name': searchKey },
```
tintinthong, line 247: Here I dono why 'company.name' is here. But it didn't seem right or general at all. I will look into it

packages/experiments-realm/crm-app.gts
```diff
@@ -218,38 +267,55 @@ class CrmAppTemplate extends Component<typeof AppCard> {
...
+              },
+            ],
+          },
+        ]
+      : [];
+
+    return {
+      filter: {
+        on: activeFilter.cardRef,
+        every: [baseFilter, ...accountFilter, ...searchFilter],
```
tintinthong: Why is accountFilter here? You are changing the way the query is supposed to be generic and query supposed to be declared. You need to look into loadAllFilters and LAYOUT_FILTER to implement ur change. Not swoop it inside here

tintinthong: This may require including a different sort of data inside LAYOUT_FILTER

tintinthong (changes requested):

packages/experiments-realm/crm-app.gts
lucaslyl: following the previous code commit, you might want to search queries by company name.

tintinthong (approved):

PR #1988, opened 2024-12-26
Refactor indexing to get mtimes for entire realm in a single HTTP request

habdelra:
This PR refactors the index worker to only use a single request to get all the mtimes for the underlying Realm filesystem.

### DIFF
--- a/packages/host/tests/integration/realm-indexing-and-querying-test.gts
+++ b/packages/host/tests/integration/realm-indexing-and-querying-test.gts
@@ -3401,6 +3401,12 @@ module(`Integration | realm indexing and querying`, function (hooks) {
                 { id: mangoID },
                 {
                   id: vanGoghID,
+                  firstName: 'Van Gogh',
+                  friends: [
+                    {
+                      id: hassanID,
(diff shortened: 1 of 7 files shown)


packages/host/tests/integration/realm-indexing-and-querying-test.gts
```diff
@@ -3401,6 +3401,12 @@ module(`Integration | realm indexing and querying`, function (hooks) {
                 { id: mangoID },
                 {
                   id: vanGoghID,
+                  firstName: 'Van Gogh',
+                  friends: [
+                    {
+                      id: hassanID,
+                    },
+                  ],
```
habdelra, line 3409: this is an interesting development--this search doc actually looks more correct now

lukemelia (approved):
Any idea why that test result changed?

habdelra:
> Any idea why that test result changed?

I suspect that there was a bug in the recursion for invalidation when we had a cycle in our links. Such that when I removed the invalidation recursion (we don’t need it anymore), the bug disappeared

PR #1989, opened 2024-12-27
Open AI panel when AI app generator starts generating

jurgenwerk:
This PR adjusts the following:
- Removes `experimentalAIEnabled` flag - I believe this is not experimental anymore
- Moves `currentRoomId` and `aiAssistantOpen` from component state to matrix and operator mode state services so that we can manipulate them from other parts of the app
- Adds `openRoom` command
- AI app generator will open the AI panel once it starts generating the app, allowing the user to observe the automatic prompts and generations from the AI bot

Here in this video you can see the AI panel opens when user clicks on "Let's get started":

https://github.com/user-attachments/assets/32662df2-4b30-43ba-83d9-aa5c3a3927b3


packages/host/tests/acceptance/commands-test.gts
```diff
@@ -622,7 +623,7 @@ module('Acceptance | Commands tests', function (hooks) {
     assert.dom('[data-test-card-url-bar-input]').hasValue(`${testCard}.json`);
     await click('[data-test-submode-switcher] button');
     await click('[data-test-boxel-menu-item-text="Interact"]');
-    await click('[data-test-open-ai-assistant]');
```
jurgenwerk, line 625: The behavior changed now that I moved the `aiAssistantOpen` from the component to the state service–previously, if the AI panel was open and the user toggled from interactive to code mode and then back to interactive, the AI panel was closed. But now, the AI panel remains open when going back to interactive mode. I think this behavior is actually more desirable, let me know if you think otherwise.

lukemelia (changes requested):
Instead of introducing a new action on context.actions, we should introduce a new host command.

jurgenwerk:
I added a new command for the purpose of opening an AI assistant room.

lukemelia (approved):
Nice work

PR #1992, opened 2024-12-30
Use CardResource to create card instances on the host

habdelra:
This PR closes a hole where we were using card instances that were not obtained from CardResources, specifically the creation of new cards. This was problematic for many reasons, in that it resulted in mulitple sources of truth for a card instance on the host which in turn leads to live update issues as well as an assortment of other bugs around stack items not being able to load properly.

This PR cleans up the `StackItem` class which is a bit schizophrenic in that it's maintaining 2 buckets of state--one using CardResources and another maintaining state for a newly created card. We collapse `StackItem` into _only_ using CardResources, and also give the `CardResource` the ability to create cards when it sees that the card document it is supplied with has no ID.


packages/host/app/lib/stack-item.ts
```diff
@@ -1,19 +1,21 @@
...
 
 import type { CardDef, Format } from 'https://cardstack.com/base/card-api';
-import type * as CardAPI from 'https://cardstack.com/base/card-api';
 
 interface Args {
   format: Format;
   owner: object;
   request?: Deferred<CardDef | undefined>;
   stackIndex: number;
-  card?: CardDef;
```
habdelra, line 16: this is super problematic--we should not allow card instances to be passed into the StackItems since we don't know the provenance of the instance (e.g. did it come from a resource?). Rather the StackItem should be responsible for getting the card instance from the CardResource.

packages/host/tests/acceptance/interact-submode-test.gts
```diff
@@ -1501,6 +1501,68 @@ module('Acceptance | interact submode tests', function (hooks) {
       assert.dom('[data-test-workspace-chooser]').exists();
     });
 
+    test<TestContextWithSave>('can create a card when 2 stacks are present', async function (assert) {
```
habdelra, line 1504: this test was one of many ways to reveal the card creation outside of a CardResource. previously this would just barf with an exception. but specifically it proves that the issue CS-7710 is fixed.

packages/host/tests/integration/components/operator-mode-test.gts
habdelra, line 1: these changes in this test were odd--i only started seeing these failures after upstream merge, i highly doubt this is related to my changes. it looks like a test selector changed... :man_shrugging:

lukemelia (approved):

tintinthong (approved):
Work Tracker new Card works! thanks!

PR #1993, opened 2024-12-31
Enhance catalog entry and code mode RHS panel to be able to create new catalog entry

tintinthong:
This PR allows a user to create a new catalog entry when focusing on a card declaration that can be used (exported card or field) to create a catalog entry

Screenshot 2024-12-31 at 11 15 48
Screenshot 2024-12-31 at 11 15 54
Screenshot 2024-12-31 at 11 16 09
Screenshot 2024-12-31 at 11 16 20
Screenshot_2024-12-31_at_11 17 38


packages/base/catalog-entry.gts
```diff
@@ -30,46 +33,12 @@ export class CatalogEntry extends CardDef {
       return new URL(this.ref.module, this[relativeTo]).href;
     },
   });
-  @field demo = contains(FieldDef);
```
tintinthong, line 33: I removed this for now and will treat adding examples of different boxel spec in separate PR

packages/base/cards-grid.gts
```diff
@@ -384,6 +384,12 @@ class Isolated extends Component<typeof CardsGrid> {
           on: catalogEntryRef,
           eq: { ref: activeFilterRef },
         },
+        sort: [
+          {
+            by: 'createdAt',
+            direction: 'desc',
+          },
+        ],
```
tintinthong, line 392: Focus on the last newly created catalog entry, if there are more than one

packages/host/app/components/operator-mode/create-file-modal.gts
```diff
@@ -765,6 +783,59 @@ export class ${className} extends ${exportName} {
       this.saveError = `Error creating card instance: ${e.message}`;
     }
   });
+
```
tintinthong, line 791: I make use of create-file-modal bcos other modals assume the existence of a catalog entry. This is not displayed in the top right for simplicity

packages/base/catalog-entry.gts
```diff
@@ -30,46 +33,12 @@ export class CatalogEntry extends CardDef {
...
-        <FieldContainer @label='Workspace Name' data-test-field='realmName'>
-          <@fields.realmName />
-        </FieldContainer>
-        <FieldContainer @vertical={{true}} @label='Demo' data-test-field='demo'>
-          <@fields.demo />
-        </FieldContainer>
-      </CatalogEntryContainer>
-    </template>
-  };
+  @field examples = linksToMany(CardDef);
```
tintinthong: stub this for now. We need to consider ways to model examples for both fieldDef and cardDef. I think @demo used to be the way to do this but let's think about this in another PR. We may very well just re-include demo but we need to investigate its usage more carefully. But, given fields are mostly displayed in cards anyway, maybe we only want to link to cards for now

packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -468,6 +469,13 @@ export default class CodeSubmode extends Component<Signature> {
     }
   }
 
+  get showBoxelSpecPreview() {
+    return (
+      !this.moduleContentsResource.isLoading &&
+      this.selectedDeclaration?.exportName
```
lukemelia: Should boxel spec pane be present for any module, or just for cards? What does this logic say?

packages/host/app/components/operator-mode/code-submode/boxel-spec-preview.gts
```diff
@@ -0,0 +1,274 @@
...
+      {{#if @showCreateBoxelSpecIntent}}
+        <div class='create-boxel-spec-intent-message'>
+          Create a Boxel Specification to be able to create new instances
+        </div>
+        <BoxelButton @kind='primary' {{on 'click' @createBoxelSpec}}>
+          Create Boxel Spec
+        </BoxelButton>
+      {{else}}
+        <div class='boxel-spec-selector'>
+          <RadioInput
```
lukemelia: Does it make sense to show a Radio button if there is only one spec instance?

packages/host/app/components/operator-mode/code-submode/boxel-spec-preview.gts
```diff
@@ -0,0 +1,274 @@
...
+
+  get realmURL() {
+    return this.realm.realmOfURL(this.operatorModeStateService.state.codePath!);
+  }
+
+  get realms() {
+    if (!this.realmURL) {
+      return [];
+    }
+    return [this.realmURL.href];
```
lukemelia: Why do we only want to search in the current realm for a spec?

packages/base/catalog-entry.gts
```diff
@@ -102,31 +71,116 @@ export class CatalogEntry extends CardDef {
   };
 
   static isolated = class Isolated extends Component<typeof this> {
+    get icon() {
+      return this.args.model.constructor?.icon;
+    }
+
     <template>
-      <CatalogEntryContainer class='container'>
```
lukemelia, line 106: Is `CatalogEntryContainer` used anymore? Should it be removed as part of this PR?

packages/host/app/components/operator-mode/code-submode.gts
tintinthong: In this logic, it models the boxel spec as present in any selected code with an exportName (including defaults)
- exported card 
- exported field 

However,  at the top-level, exported variable, function, class which are not card or fields, are ignored entirely by the RHS

Screenshot 2024-12-31 at 13 33 50

packages/base/catalog-entry.gts
```diff
@@ -3,23 +3,26 @@ import {
...
+import GitBranch from '@cardstack/boxel-icons/git-branch';
 
 export class CatalogEntry extends CardDef {
   static displayName = 'Catalog Entry';
   static icon = BoxModel;
-  @field title = contains(StringField);
-  @field description = contains(StringField);
+  @field readMe = contains(MarkdownField);
+
   @field ref = contains(CodeRef);
```
habdelra, line 45: the embedded view for this field type is super bare bones. could use a little CSS love.

packages/base/catalog-entry.gts
habdelra: this actually needed a lot more discussion in terms of how we render this as it should be a polymorphic card and not a field. but yeah, good to remove for now...

packages/base/catalog-entry.gts
tintinthong: This is used inside of the fitted view. We can just make this local to the fitted view if preferred since its not being re-used any where else

packages/host/app/components/operator-mode/code-submode/boxel-spec-preview.gts
tintinthong: There can be more than one boxel spec / catalog entry since the boxel spec is simply a card that points to a code ref. 

A user can make this themselves. And I don't think we are in control of the user making an extra code spec

packages/host/app/components/operator-mode/create-file-modal.gts
```diff
@@ -765,6 +783,57 @@ export class ${className} extends ${exportName} {
...
+  private createBoxelSpecInstance = restartableTask(async () => {
```
habdelra, line 792: i think its probably time to split the different files that can be created into different modules. right now this module is doing _a lot_. perhaps we can use a strategy pattern to figure out how to create the right type of file.

packages/host/app/components/operator-mode/code-submode/boxel-spec-preview.gts
tintinthong: Ahh good catch. I think this should be searching for all available realms that the user has access to that references this code.  https://github.com/cardstack/boxel/pull/1993#discussion_r1900223901 this wud then make more sense to enumerate the different code specs in a radio list too

packages/host/app/components/operator-mode/code-submode/boxel-spec-preview.gts
lukemelia: I understand that there can be more than one. But I also think that in most cases there will be zero or one. So I wonder if radio input should be a special case and not the usual case.

packages/host/app/components/operator-mode/create-file-modal.gts
habdelra: it could be a simple as having a module that exports an async function that is the meat of this task, and then setting up a map of file types to functions to make the file type and then just executing the async function that makes the file.

packages/base/catalog-entry.gts
lukemelia: I don't have a strong preference. Just wanted to make sure we didn't have an orphan.

tintinthong:
Im currently working on this design. 

Screenshot 2025-01-06 at 08 27 12

Screenshot 2025-01-06 at 08 34 37

The changes are 
- having a dropdown 
- in each dropdown item, cut the path and use realm icon. in the plural case, I notice there is no tag of the spec anywhere.. I am clarifying with chris now
- the accordion will display the n instances if there are any 
- if there are n = 1, the dropdown arrow shud not be there
- there shud be a create button at the accordion (but shud wait for example from ian on how to implement them)

packages/host/app/components/operator-mode/create-file-modal.gts
tintinthong: > it could be a simple as having a module that exports an async function that is the meat of this task, and then setting up a map of file types to functions to make the file type and then just executing the async function that makes the file.

Im not sure that I went overkill on this but I have done some work which might be a little bit of a restructuring. Tho particularly, I shud not include this in **as part of this PR** to ensure there are no regressions independently. 

The refactor is essentially
- **Use a service to store the state of the modal createFileService.** The service is responsible for triggering the open of the modal and deciding the strategy based upon the api call. So consumers are expected to inject this into them to open the modal. 
- **create separate strategies** that will have an interface of `create()`
- the strategies `CreateCardInstance`, `CreateCardDefinition`, `IneritCardDefinition`, `CreateCardInstanceFromDefinition, `CreateFieldDefinition`. Each of these strategies will require different things but the service will validate this when opening. 
- the strategies will take in a composition of `FileNameHandling` (fileName, displayName) and `CatalogEntryHandling` (catalogEntry). Each of these classes should handle the edit and default state internally. eg catalogEntry will hold the dropdown state within itself. Each of these tend to within itself check whether its the treatment for a card or a field, eg gts vs json or adoptsFrom vs inherit. 
- **the createFileModal then just relay the strategy** within it. The create button would just be a single button that will call create from the strategy.  A lot of the displays in the modal  should be displayed based upon `isCatalogEntryHandling` and `isFileNameHandling`. So, those compositions are key to the UI. 

**What do you think so far?** 

I feel there are some issues I run into but not too major
- I think I have instantiated the strategies as classes but I seem to want to include services inside that strategy but can't inject them by default it in unless I use some EmberObject stuff. I decided to pass it down using a context object into the instantiation of the class. 
- When opening the modal and creating a card, there is a redirect of the codePath and it sets everything to edit mode. The function setPreview is all the way up the top in `code-submode` that means consumers will actually want to call this action on `code-submode` perhaps this is fine since the preview-format exists on code-submode itself.

tintinthong:
This is the latest isolated and embedded view 

Screenshot 2025-01-14 at 00 33 33

Screenshot 2025-01-14 at 00 33 44

this is of an app boxel spec 

Screenshot 2025-01-14 at 00 49 27

tintinthong:
Here is a typical scenario for code-submode having 2 catalog entry/boxel spec instances

Screenshot 2025-01-14 at 00 39 04
Screenshot 2025-01-14 at 00 35 41

packages/boxel-ui/addon/src/components/select/trigger.gts
```diff
@@ -2,6 +2,7 @@ import Component from '@glimmer/component';
 import type { Select } from 'ember-power-select/components/power-select';
```
tintinthong, line 2: Changes in this component was to remove the arrow that suggest dropdown when the component was disabled

packages/base/cards-grid.gts
```diff
@@ -384,13 +384,19 @@ class Isolated extends Component<typeof CardsGrid> {
...
       };
     }
     let card = await chooseCard<CatalogEntry>(
       {
         filter: {
           on: catalogEntryRef,
-          every: [{ eq: { isField: false } }],
+          every: [{ eq: { type: 'card' } }],
```
tintinthong: I changed this to be clearer about what we are filtering for. But I aslo found that when isField was turned into a computed, it cudn query and I filed https://linear.app/cardstack/issue/CS-7821/filtering-by-computed-doesnt-work-at-the-first-level

packages/base/catalog-entry.gts
```diff
@@ -157,3 +308,52 @@ class CatalogEntryContainer extends GlimmerComponent<Signature> {
...
+      return StackIcon;
+    case 'app':
+      return AppsIcon;
+    case 'field':
+      return LayoutList;
+    case 'skill':
+      return Brain;
+    default:
+      return;
+  }
```
tintinthong, line 400: These are the different types of boxel specs we are supporting for now

packages/base/cards-grid.gts
habdelra: Just double checking, we don’t allow you to filter by fields that are not in the template. Is this field used the template. (Like is it part of the embedded field if this is a linked card)? If not is it set to is used? I’m positive we have passing tests that exercise the ability to filter by computeds

tintinthong: Confirm. Oh its another one of those. Thanks for pointing out.

isField was not used inside any of the template 

I confirm within the test that that computed **WORKS** when the computed field exist within the isolated template. So, the behaviour is as you point out. Although I would add that it won't work if the computed field only exist within embedded or fitted (not sure if its something we want to consider) but not the isolated template

habdelra: So the rationale for the behavior is the "social graph problem". Imagine a boxel app that is a social graph like facebook. each person card has a linkToMany relationship to all their friends, which are also person cards. If we have no limits on what feields are included in the search doc, we will likely end up including all the people in the world in each search doc. the template is a natural way to draw boundaries on the search doc. we do have an escape valve `isUsed` which can allow a field to be in the search doc even though its not in a template (and this could get abused to create the problem that I just raised).

packages/host/app/components/operator-mode/create-file-modal.gts
tintinthong: Im going to push a PR out for this separately

packages/base/cards-grid.gts
tintinthong: Thanks for the explanation. 

This is an aside. Shud not affect aspects of this PR. Since I filter by a field `type`, not `isField` anymore.

burieberry: How about you call this attribute `specType` instead of `type`? It is confusing because there's a real `type: "card"`. I think `isField` is helpful so we can distinguish between field defs and card defs. `specType` can then be freely used for the visual design reasons you need it for.

burieberry:
Some observations -- if these are outside of the scope of this PR, we should split up the work on linear:
I clicked on "Create" on Boxel Spec on `app-card` to create a spec for Tab fieldDef:
- "Create" on Boxel Spec panel only allows me to create a single spec
- Boxel Spec panel is only showing me the spec for Tab and not for AppCard
- The newly created spec does not know its `type` (it's null). `isField` is also marked as `false` even though I created it from a Field Def.
- "Add Cards" for Examples button displays everything, even when the ref is known. 
- The new specs I published did not appear when I clicked on Create New from interact mode cards grid

Visual design:
- The spacing and alignments can use some tidying up.
- The pill with the highlighted arrow on exported name looks like it will be a button, but it's not.

packages/base/catalog-entry.gts
```diff
@@ -3,73 +3,61 @@ import {
...
   });
-  @field demo = contains(FieldDef);
-  @field realmName = contains(StringField, {
+  @field examples = linksToMany(CardDef);
+  @field title = contains(StringField, {
     computeVia: function (this: CatalogEntry) {
-      return this[realmInfo]?.name;
+      return this.name || this.ref.name;
     },
   });
```
burieberry: This won't work well for default exports. The ref name will be 'default' for each one

burieberry (commented):
I came across a couple of issues as I wrote in my comment. I think you need a test that covers the case for a module with a default export and multiple FieldDef and CardDef exports and make sure it works as expected

tintinthong:
> Some observations -- if these are outside of the scope of this PR, we should split up the work on linear: I clicked on "Create" on Boxel Spec on `app-card` to create a spec for Tab fieldDef:
> 
> * "Create" on Boxel Spec panel only allows me to create a single spec

That is intended to encourage users not to create more than one spec per code ref

> * Boxel Spec panel is only showing me the spec for Tab and not for AppCard

This dependes on the code you have selected on the LHS definition picker. You need to choose AppCard as the selected export

> * The newly created spec does not know its `type` (it's null). `isField` is also marked as `false` even though I created it from a Field Def.

Good catch. 

> * "Add Cards" for Examples button displays everything, even when the ref is known.

I think this can be useful to have. 

> * The new specs I published did not appear when I clicked on Create New from interact mode cards grid

Hmmm weird. I think this is unexpected behaviour. I will look into it

> Visual design:
> 
> * The spacing and alignments can use some tidying up.
> * The pill with the highlighted arrow on exported name looks like it will be a button, but it's not.

Let me get the behaviour complete first and then maybe we can pair on this

tintinthong:
> The new specs I published did not appear when I clicked on Create New from interact mode cards grid

I can't seem to re-create this behaviour

tintinthong:
> "Add Cards" for Examples button displays everything, even when the ref is known.

this seems more involved than I anticipate. I have filed this here https://linear.app/cardstack/issue/CS-7831/link-to-examples-pertaining-to-code-ref

burieberry:
Tried this again and the spec type and isCard/isField logic is now working as expected. I was also able to use the newly generated card def catalog entry to create new instances. 👍

PR #1995, opened 2024-12-31
Fix delete in search resource

tintinthong:
Essentially, the instances array returned from the search resource [undefined, ...] bcos of the delete action

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -81,7 +81,9 @@ export class Search extends Resource<Args> {
 
   @cached
   get instances() {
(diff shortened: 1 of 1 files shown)


tintinthong:
Its likely this is a workaround that doesn't address the original issue

richardhjtan (approved):

tintinthong:
After pairing session with @habdelra, the main reason is that card resources clear the card instance upon receiving a 404 (assuming the card has deleted). 

https://github.com/cardstack/boxel/blob/fa5665d1b3caa8b0eb955c2e399c8417c66a6b9d/packages/host/app/resources/card-resource.ts#L389

The current fix is correct to only return searches that has card instances

PR #1998, opened 2025-01-03
CS-7723 account fitted view

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7723/implement-account-fitted-design

What is changing
- add account fitted view design which contains account header and tags (by condition).

**Screenshots**

https://github.com/user-attachments/assets/49e9575e-fe3b-48d6-9d86-0c3bfea5deff


tintinthong (approved):
👍

PR #2003, opened 2025-01-06
Flatten index to remove expensive joins during invalidation

habdelra:
This PR flattens our index so that remove the expensive join that was required to determine the index records that pertain to the current realm version. Instead of using a single table to control both the "production" index and the "working" index (the index that is being built), we use 2 tables. One table, `boxel_index_working`, is just for building the index for each mutation, and the other table, `boxel_index`, is for querying the final version of the index. 

TODO:
- [x] there is still another potentially expensive join used for invalidation. Specifically we are using a less than optimal query in order to preserve SQLite syntax compatibility in our invalidation query. Also consider breaking this compatibility and using a more optimal postgres query for this specific scenario. 
- [x] merge in `boxel_index.icon_html` column from https://github.com/cardstack/boxel/pull/1971


packages/host/tests/unit/index-query-engine-test.ts
```diff
@@ -484,15 +484,6 @@ module('Unit | query', function (hooks) {
...
 
-  test('can get paginated results that are stable during index mutations', async function (assert) {
-    await runSharedTest(indexQueryEngineTests, assert, {
-      indexQueryEngine,
-      dbAdapter,
-      loader,
-      testCards,
-    });
-  });
-
```
habdelra, line 495: when ed ad I reviewed this he felt that this was actually a weird feature in that this means that as you paginate the search results you are not necessarily seeing the latest query results since we are trying to maintain the results from when you originally made the request as you paginate thru the results as opposed to showing the latest values.

packages/host/tests/unit/index-writer-test.ts
```diff
@@ -48,14 +48,6 @@ module('Unit | index-writer', function (hooks) {
     });
   });
 
-  test('only invalidates latest version of content', async function (assert) {
```
habdelra, line 51: this becomes moot since we are now breaking out the working version and the production version into separate tables

packages/runtime-common/expression.ts
```diff
@@ -386,28 +386,6 @@ export function expressionToSql(query: Expression) {
   };
 }
 
-export function realmVersionExpression(opts?: {
```
habdelra, line 389: this is the expensive SQL that we are trying to eliminate

packages/runtime-common/index-writer.ts
```diff
@@ -469,13 +463,18 @@ export class Batch {
...
         'SELECT i.url, i.file_alias, i.type',
-        'FROM boxel_index as i',
+        'FROM boxel_index_working as i',
+        // TODO: If this query is still slow we'll need to consider dropping
+        // this join and writing a pg specific query here--as this query is
+        // currently written in a less performant manner in order to be
+        // compatible with SQLite. make sure to measure performance in hosted
+        // env first before sending for team review. Should strive to be less
+        // than 50ms (currently about 600-700ms) when this is run against a
+        // reasonably sized index.
```
habdelra: we should really just do this

lukemelia (approved):
LGTM 👍

PR #2005, opened 2025-01-06
Add sprint planner (work tracker) to seed realm

tintinthong:


jurgenwerk (approved):
I tested this locally–created a new personal realm and the sprint planner card was there.

richardhjtan (approved):

habdelra:
Probably wanna hold off on merging this until we figure out the linksToMany bug. if its related to work tracker code then users will not be able to get upstream changes for any bug fixes made after the create their realm

PR #2006, opened 2025-01-07
Add middleware to loader to show "broken" icon when import from boxel.ai icons fails

lukemelia:


packages/host/app/services/loader-service.ts
```diff
@@ -48,6 +48,22 @@ export default class LoaderService extends Service {
...
+    middlewareStack.push(async (req, next) => {
+      let response = await next(req);
+      if (
+        !response.ok &&
+        req.url.startsWith(
+          'https://boxel-icons.boxel.ai/@cardstack/boxel-icons/v1/icons/',
+        )
+      ) {
+        req = new Request(
+          'https://boxel-icons.boxel.ai/@cardstack/boxel-icons/v1/icons/error-404.js',
```
backspace: Could these use `iconsURL` [like here](https://github.com/cardstack/boxel/blob/5f92e6746b75c91c1db53e873ab1b8a3e5cb422a/packages/host/app/services/network.ts#L93)?

packages/host/tests/integration/components/loading-test.gts
```diff
@@ -0,0 +1,91 @@
...
+      },
+    };
+    let exampleCard = await createFromSerialized<typeof CardDef>(
+      doc.data,
+      doc,
+      undefined,
+    );
+    await renderCard(loader, exampleCard, 'fitted');
+
+    assert.dom('[data-test-card-title]').containsText('Example');
```
backspace, line 88: Maybe it’s fine but this is probably duplicated elsewhere and doesn’t seem to relate to this specific and otherwise-concise test 🤓

backspace (approved):

packages/host/tests/integration/components/loading-test.gts
lukemelia: Nice to ensure that the fallback retains the rest of the template as-is

PR #2008, opened 2025-01-07
Enhance component readability

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7761/enchance-activity-card

- simplify the card component with yield header and content

Screenshot 2025-01-07 at 15 56 53


richardhjtan (approved):

richardhjtan:
@lucaslyl  I assume that is no component uses it yet so it won't affected?

lucaslyl:
> @lucaslyl I assume that is no component uses it yet so it won't affected?

yes, I just pushed another commit to remove the dummy activity card, so it won't affect anything now.

PR #2013, opened 2025-01-07
BUG: resource id was resolving to wrong URL in LinksToMany deserialize

tintinthong:


packages/base/card-api.gts
```diff
@@ -1297,7 +1297,10 @@ class LinksToMany<FieldT extends CardDefConstructor>
           cachedInstance[isSavedInstance] = true;
           return cachedInstance;
         }
-        let resourceId = new URL(value.links.self, relativeTo).href;
+        let resourceId = new URL(
+          value.links.self,
+          !Array.isArray(doc.data) ? doc.data.id : relativeTo,
+        ).href;
```
tintinthong, line 1305: Before change 

```
relativeTo.href = 'http://localhost:4201/experiments/sprint-task'
value.links.self = '../Tag/f00a2177-df10-4ad7-8eac-2fbf8448de98'
resourceId =  'http://localhost:4201/Tag/f00a2177-df10-4ad7-8eac-2fbf8448de98'
```

After change 

```
doc.data.id = 'http://localhost:4201/experiments/SprintTask/259a2df6-6f0e-4795-abcf-2d16c437fb97'
value.links.self = '../Tag/f00a2177-df10-4ad7-8eac-2fbf8448de98'
resourceId = 'http://localhost:4201/experiments/Tag/f00a2177-df10-4ad7-8eac-2fbf8448de98'
```

The resourceId should be relative to instance, not the module

tintinthong:
In pairing, we were curious about the relativeTo value. Failing test is revealing something deeper about the expection of what the relativeTo is

habdelra (approved):

habdelra:
ultimately i'd like to remove the `relativeTo` symbol from the card-def. things should always be relative to the card id. and when we have a new card instance, it doesn't really make sense to even use relative URL's since we don't even know what we are relative to yet. this is probably a painful refactor (not necessarily part of this PR)--but i think it would make for easier to reason about code base.

tintinthong:
> In pairing, we were curious about the relativeTo value. Failing test is revealing something deeper about the expection of what the relativeTo is

Mind fart. doc.data.id could be undefined. Thats why the test fails. I maintain the fix AND default to relativeTo

PR #2015, opened 2025-01-07
shard realm server tests

habdelra:
make sure to hide whitespace when reviewing this PR. the lines changed is super deceptive as it's 99.9% whitespace. (there is also a lot of prettier updates caught up in this PR as someone doesn't have that enabled 😑...)


packages/realm-server/scripts/lint-test-shards.ts
habdelra, line 1: this script is used to make sure that the ci.yaml doesnt forget to be updated when a new test module is added

packages/realm-server/tests/auth-client-test.ts
```diff
@@ -7,93 +7,104 @@ import {
...
 import jwt from 'jsonwebtoken';
+import { basename } from 'path';
 
 function createJWT(expiresIn: string | number) {
   return jwt.sign({}, 'secret', { expiresIn });
 }
 
-module('realm-auth-client', function (assert) {
-  let client: RealmAuthClient;
+module(basename(__filename), function () {
```
habdelra, line 16: this is a tiny little boilerplate that we need to add to each test file made sure we can correlate filenames to test modules so we can filter them properly. this diff is a little hard to read without hiding white space, but this is an additive change--just wrapping the entire test module with this call.

habdelra:
So a general observation here. sharding the server tests doens't actually make things run faster. it looks like for all the server tests they finish pretty quickly, but after the tests runto completion and the test results are printed out, there is still some timer(s) that seems to be keeping the node process from exiting promptly. I'm guessing it's in some lib code that we are using.

habdelra:
perhaps we can write our own custom test reporter that wraps the existing reporter and just calls a `process.exit()` to force the test to end (i noticed that process.exit() does actually work to force the process to end)

packages/realm-server/tests/index.ts
```diff
@@ -12,3 +12,29 @@ import './queue-test';
...
+      todo: number;
+    };
+  }) => {
+    if (testCounts.failed > 0) {
+      process.exit(1);
+    } else {
+      process.exit(0);
+    }
+  },
+);
```
habdelra: this forces our tests to exit promptly despite some timer that seems to be preventing the node process from exiting

habdelra:
note the improved speeds of the realm server tests in the test results below (lol, it was the matrix tests that were slow this time)

lukemelia (approved):

packages/realm-server/tests/index.ts
```diff
@@ -12,3 +12,29 @@ import './queue-test';
 import './realm-server-test';
 import './virtual-network-test';
 import './billing-test';
+
+// There is some timer that is preventing the node process from ending promptly.
```
backspace, line 16: Is it feasible we could fix the underlying problem vs having to forcefully exit?

packages/realm-server/scripts/lint-test-shards.ts
```diff
@@ -0,0 +1,83 @@
...
+      console.error(
+        `Error: Test file '${filename}' exists in the filesystem but not in the ${yamlFilePath} file.`,
+      );
+      errorFound = true;
+    }
+  }
+  for (let filename of ciTestModules) {
+    if (!filesystemTestModules.includes(filename)) {
+      console.error(
+        `Error: Test file '${filename}' exists in the YAML file but not in the ${yamlFilePath} filesystem.`,
```
backspace: I appreciate this linting so we have a fallback instead of possibly losing new files. I tried it out locally and it worked as expected, though the Yaml path shouldn’t be here, right?

realm-server 2025-01-07 20-54-28


```suggestion
        `Error: Test file '${filename}' exists in the YAML file but not in the filesystem.`,
```

.github/workflows/ci.yaml
```diff
@@ -234,8 +234,24 @@ jobs:
     name: Realm Server Tests
     runs-on: ubuntu-latest
     concurrency:
-      group: realm-server-test-${{ github.head_ref || github.run_id }}
+      group: realm-server-test-${{ matrix.testModule || github.head_ref || github.run_id }}
```
backspace:
```suggestion
      group: realm-server-test-${{ matrix.testModule }}-${{ github.head_ref || github.run_id }}
```

I don’t think `||` makes sense here, each test file for a PR should be its own concurrency group… I think??

backspace (approved):
> a lot of prettier updates caught up in this PR as someone doesn't have that enabled 

irksome, I sometimes save files without autoformatting because I don’t want spurious diffs 😖

packages/realm-server/tests/index.ts
habdelra: Yeah I tried to find the source of the issue but it was very mysterious. Just pausing a debugger after the test finishes but before the process ends never breaks. It doesn’t look like something obvious we are doing in the test, and I see it happening for really simple test modules too. It makes me feel like some lib that we are loading is doing this as a side effect. But yeah, I’m unable to figure out where it’s coming from.

.github/workflows/ci.yaml
habdelra: Oh I thought `||` in this context was a concat operator, no?

habdelra:
> > a lot of prettier updates caught up in this PR as someone doesn't have that enabled
> 
> irksome, I sometimes save files without autoformatting because I don’t want spurious diffs 😖

I think it’s better to have consistent code formatting than spurious diffs. It makes things like find and replace much more predictable (like favoring ‘ vs “).

.github/workflows/ci.yaml
habdelra: Ah, i found the documentation for these expressions--yes, i was using it incorrectly https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#format

PR #2018, opened 2025-01-08
don't wait for ECS service stability for pg-migration deploy

habdelra:
we don't wait for ECS service stability for pg-migrations since it's not a service, just a task that needs to run in the same VPC--but the new ECS deploy GH action version gets mad that the service is not ready and exits with error status breaking the deploy. in the previous version of the GH action for ECS deploys, not having a heath check meant not waiting for the service to stabilize, but now we need to explicitly say it.


.github/workflows/manual-deploy.yml
```diff
@@ -100,10 +101,18 @@ jobs:
...
       service-name: "boxel-pg-migration-${{ inputs.environment }}"
       image: ${{ needs.build-pg-migration.outputs.image }}
+      wait-for-service-stability: false
+
+  post-migrate-db:
+    name: Wait for db-migration
+    needs: [migrate-db]
+    runs-on: ubuntu-latest
+    steps:
+      - run: sleep 240
```
habdelra, line 111: we still need to wait for the ECS task to be provisioned and start running. (the old action  version actually just waited for the timeout of 5 minutes--i think we can do a little better than that)

habdelra:
This issue is holding up deployments so i'm going to merge now. I'll be happy to address any feedback after it's merged

lukemelia:
Seems good to me

PR #2021, opened 2025-01-09
create atom format & replace with atom

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7697/contacts-in-account-shud-be-atom

result: 
image

### DIFF
--- a/packages/experiments-realm/crm/account.gts
+++ b/packages/experiments-realm/crm/account.gts
@@ -33,6 +33,8 @@ import Handshake from '@cardstack/boxel-icons/handshake';
 import ClockX from '@cardstack/boxel-icons/clock-x';
 import ClockUp from '@cardstack/boxel-icons/clock-up';
 import Contract from '@cardstack/boxel-icons/contract';
+import EntityDisplayWithThumbnail from '../components/entity-thumbnail-display';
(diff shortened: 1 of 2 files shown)


packages/experiments-realm/crm/account.gts
```diff
@@ -522,6 +514,80 @@ class FittedTemplate extends Component<typeof Account> {
   </template>
 }
 
+class AtomTemplate extends Component<typeof Account> {
```
tintinthong: Whats with this code?? You are creating an atom template for account? Where is this used? 

At first sight, I dont think you need to add this

tintinthong: Even if it was an atom template, you are using contact to display the atom format of an account. To me, I'd rather have the account name which shud be derived from the computed name of the account

lucaslyl: > Whats with this code?? You are creating an atom template for account? Where is this used?
> 
> At first sight, I dont think you need to add this

this atom format is used for activity card, but you are right, it seem like each activity card is linked to different contact of that account, I will remove this atom format from this pr.
Screenshot 2025-01-10 at 10 45 56

richardhjtan (approved):

tintinthong (approved):

PR #2011, opened 2025-01-07
Add Boxel UI components: ColorPicker, ColorPalette

jurgenwerk:
Adds 2 components that will help with composing color fields in Boxel:

image

image

Color spectrum open:
image

The work that follows will be using this in the Blog Category, Pill color field (design screenshot):

image

PR #2022, opened 2025-01-09
Use color palette for color fields in blog category

jurgenwerk:
Newly added [Boxel UI color components](https://github.com/cardstack/boxel/pull/2011) were the basis for creating the `ColorField`, which enables us to edit colors: 

image

I also assigned colors to our existing categories in seed realm:
image


packages/experiments-realm/blog-category.gts
```diff
@@ -18,8 +19,8 @@ export const categoryStyle = (category: Partial<BlogCategory>) => {
     return;
   }
   return htmlSafe(`
-    background-color: ${category.backgroundColor || '#FFFFFF'};
-    color: ${category.textColor || '#000000'};
+    background-color: ${category.backgroundColor?.hexValue || '#000000'};
+    color: ${category.textColor?.hexValue || '#ffffff'};
```
tintinthong, line 23: So this is the default for blog category. If I want a random default similar to how avatar computes the background color. What is the best way to do that? Wud that be an extend of this color field and handle that default internally?

I know I need this in particular use-case. But, I was just wondering if we would handle this in this PR. I think maybe NOT

packages/seed-realm/color.gts
```diff
@@ -0,0 +1,40 @@
+import {
```
tintinthong: My personal opinion is: The color.gts should exist inside the fields folder.. I think there are other examples inside of experiments which wud benefit from that treatment. Since they are somewhat of an enhanced /base fields

packages/seed-realm/color.gts
```diff
@@ -0,0 +1,40 @@
...
+  Component,
+  FieldDef,
+  StringField,
+  contains,
+  field,
+} from 'https://cardstack.com/base/card-api';
+import { ColorPalette } from '@cardstack/boxel-ui/components';
+import { ColorPicker } from '@cardstack/boxel-ui/components';
```
tintinthong: small cleanup here

packages/experiments-realm/blog-category.gts
jurgenwerk: We could make a function that converts the arbitrary string (category name) to some hex color value, in a deterministic way. And a function that returns a good contrast color to that color, similarly to how the avatar works. Then we could have a computed fields that either returns the computed colors, or user-set colors, in case they set them

packages/seed-realm/color.gts
jurgenwerk: Ok, I can put it there

tintinthong:
Is there a way within the color picker and palette component to remove the selection of color after uve picked it?? And just let the card display the default

jurgenwerk:
`<ColorPicker @color='#FF0000' @showHexString={{true}} @disabled={{true}}/>` will give you a non-editable color, like this:

image
 
 
 Is this what you had in mind?

tintinthong:
> true

What I mean is to remove the color that is already set. The UI doesn't give u a button to just REMOVE the color that is set inside of the card 

Im thinking from the users perspective of editing the color. Like the user wud probably want to go to the default color via the color picker or whatever its set as the default inside of the card

Or is it better UX to not give the user that option at all

tintinthong:
Screenshot 2025-01-09 at 11 11 26

The diagonal slash is what I mean @jurgenwerk. Maybe design will have input on this

It wudn be a bad idea to make an extra column white and (default)

jurgenwerk:
@tintinthong you ok if we put these additional features in separate tickets?

packages/seed-realm/color.gts
jurgenwerk: Done

tintinthong:
> @tintinthong you ok if we put these additional features in separate tickets?

Yep that sounds good. Prob good for discussion using sprint planning too

tintinthong (approved):

richardhjtan (approved):

PR #2023, opened 2025-01-09
Attach tool call results to the call & trigger responses on tool calls

IanCal:
Connects the command output to the original tool call event.

I had to change what we were doing with the event ids to get them to line up.

Issues

* We respond for every message from the user, so if there's results from *two* command executions then it replies with two things right now

Auto running for a small weather example with a custom embedded template
image


IanCal:
@lukemelia I'm wondering if something has gotten broken with the event IDs when there's streaming on the host side - the fix may not be to change it here. I think we've recently lost streaming which could be related?

Hynaotar:
> Connects the command output to the original tool call event.
> 
> 
> 
> I had to change what we were doing with the event ids to get them to line up.
> 
> 
> 
> Issues
> 
> 
> 
> * We respond for every message from the user, so if there's results from *two* command executions then it replies with two things right now
> 
> 
> 
> 
> 
> Auto running for a small weather example with a custom embedded template
> 
> image
> 
>

lukemelia (approved):

PR #2024, opened 2025-01-09
Indexing error catching

habdelra:
This is to hopefully catch the scenarios where indexing seems to drop off a cliff in production and our jobs never complete. We wrap a try/catch around the entire indexing functions and always make sure to return something. and log errors if they occur.

### DIFF
--- a/packages/host/app/lib/current-run.ts
+++ b/packages/host/app/lib/current-run.ts
@@ -321,11 +321,14 @@ export class CurrentRun {
         skipList.push(url);
       }
     }
-    if (skipList.length === 0) {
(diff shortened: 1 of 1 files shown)


backspace (approved):
Is it worth logging to Sentry too?

habdelra:
> Is it worth logging to Sentry too?

that's a good idea. although this is in the fastboot context. probably we'll need to project a special global into fastboot in order to do that. such a global could be abused by a malicious card too...

habdelra:
I'm gonn close this--it never actually was able to find the issue causing indexing to stop in prod. (that is solved in https://github.com/cardstack/boxel/pull/2026

PR #2027, opened 2025-01-09
Refactor hard-coded searchCard command into regular command implementation

lukemelia:


backspace (approved):

lukemelia:
@IanCal merge once you have reviewed if you wish

packages/host/app/commands/search-cards.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+    if (input.cardType) {
+      filter.eq = { _cardType: input.cardType };
+    }
+    if (input.type) {
+      filter.type = input.type;
+    }
+    return new SearchCardsByQueryCommand(this.commandContext).execute({
+      query: {
+        filter: filter as Filter,
```
habdelra, line 46: that's interesting that you don't need to suppply an `on` field which tells the search which card def you are talking about. something like this https://github.com/cardstack/boxel/blob/main/packages/runtime-common/tests/index-query-engine-test.ts#L156-L165. i would have thought we needed to provide the CardDef code ref in the `on` field.

packages/host/app/commands/search-cards.ts
```diff
@@ -0,0 +1,93 @@
...
+      throw new Error(
+        'At least one of title, cardType, or type must be provided',
+      );
+    }
+    let filter = {} as any;
+    if (input.title) {
+      filter.contains = { title: input.title };
+    }
+    if (input.cardType) {
+      filter.eq = { _cardType: input.cardType };
```
habdelra, line 39: because we are using the card type's `displayName` here (that is really what the `_cardType` field is searching against), there is the chance that different CardDef's may coincidentally have the same `displayName`. so this is not necessarily a super precise search. If we wanted to be more precise we should use the `on` field and provide the card's code ref that we are searching against.

lukemelia: Wait, what is the difference between `on` and `type` (see line 41-34 below)? That is where I am providing a CodeRef.

habdelra (approved):

packages/host/app/commands/search-cards.ts
habdelra: ah! i missed this. `type` is an alias for `on`. generally we use `on` because you can specify it at any level of the filter. but at the top level we provide a `type` which reads nice.

packages/host/app/commands/search-cards.ts
habdelra: i think your use of `type` is why we don't have to say `on`--but i would think `type` would need to default to the base `CardDef` if no `type` is supplied

lukemelia: The backend falls back to CardDef if not specified: https://github.com/cardstack/boxel/blob/e03162c5c2a4194ca25672153369fa723d3f2031/packages/runtime-common/index-query-engine.ts#L357

and

https://github.com/cardstack/boxel/blob/e03162c5c2a4194ca25672153369fa723d3f2031/packages/runtime-common/index-query-engine.ts#L542-L547

IanCal:
I'll try and pull this into a test but it seems like it can't search for a title and a type?

image

IanCal:
@lukemelia I've fixed a mixup between type and cardtype that caused the above error, and added a couple of small tests about searching. We should also improve the descriptions here as searching on a type is very restrictive (author brings back nothing, for example)

PR #2029, opened 2025-01-10
Cs 7789 deal card grid item view

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7789/deal-card-grid-item-view

### What is changing
- grid item view of the deal card

Note:
- The event part will be a separate task

**Screenshot**
Screenshot 2025-01-10 at 10 48 11 PM


lucaslyl:
Richard, for the grid item view is good.
But it's seems this changes is affect the CRM strip view, Perhaps we can create/edit aspect ratio to adjust the Visual in the Strip view as before :
image

richardhjtan:
@lucaslyl  thanks for reminding, close and open new PR to not affect existing strip view.

PR #2030, opened 2025-01-10
Make sure card types in index card are ordered by display_name

jurgenwerk:
The bug ticket specifies that the order of the card types is wrong:

image

However, during testing, I was not able to reproduce this. Even by creating new realms and checking the list - it was all perfectly in alphabetical order... These items are stored as a json array in realm_meta table in the following form:

```
realm_url             | realm_version | value                                                     
'https://example.com' | 1             | [{total:2, display_name: "Author",...}] 
```

It looks like the query that writes this data is not reliable and the sort by display name was "accidental" for most of the time–adding an explicit order should solve this rare bug. 

There is already a test that tests for the order of the display names: https://github.com/cardstack/boxel/blob/1a322fa545a781e0f7993457cf7d2b6bbd6382f9/packages/realm-server/tests/realm-server-test.ts#L4099-L4134

### DIFF
--- a/packages/runtime-common/index-writer.ts
+++ b/packages/runtime-common/index-writer.ts
@@ -305,6 +305,7 @@ export class Batch {
         any([['i.is_deleted = false'], ['i.is_deleted IS NULL']]),
       ]),
       `GROUP BY i.display_names->>0, i.types->>0`,
+      `ORDER BY i.display_names->>0 ASC`,
     ] as Expression);
 
     let { nameExpressions, valueExpressions } = asExpressions(


habdelra:
~~this was previously a revert from a working scenario. if there was already a test for this, why was it not failing? could we add tests to this so we can be sure there isn't another reversion of this functionality~~

EDIT: 
nevermind I just read thru the tests more carefully. it looks like this is not an easy failure case to reproduce

habdelra (approved):

backspace:
What was unreliable about the query? There’s no way to reproduce the out-of-order insertion?

jurgenwerk:
> What was unreliable about the query? 

The ordering of the `display_name`. It looks like it's one of those queries that occasionally deliver results in a different order when there's no explicit ORDER BY clause.

PR #2034, opened 2025-01-13
replace custom max-width with 100%

lucaslyl:
linear:  https://linear.app/cardstack/issue/CS-7787/crm-grid-max-width-should-be-100percent

- Remove any hardcoded max-width values
- Use CSS variables for flexible width control if required by design
- Let the content naturally fill the available space

Before:
Screenshot 2025-01-13 at 15 30 38

After:
Screenshot 2025-01-13 at 15 31 21


burieberry (commented):
Add the 1040px content-max-width for blog app. Otherwise looks good

lucaslyl:
> Add the 1040px content-max-width for blog app. Otherwise looks good

Sure, done add the max-width to blog app based at experiment-realms

richardhjtan (approved):

PR #2038, opened 2025-01-13
Fix missing stack-item headerColor and realm-icon border properties

burieberry:
**Before:** 
Stack Item's headerColor disappeared after opening and closing another card in the stack on top of it
before

**After:**
after

### DIFF
--- a/packages/boxel-ui/addon/src/components/card-header/index.gts
+++ b/packages/boxel-ui/addon/src/components/card-header/index.gts
@@ -1,4 +1,3 @@
-import { concat } from '@ember/helper';
 import { on } from '@ember/modifier';
 import Component from '@glimmer/component';
 import type { ComponentLike } from '@glint/template';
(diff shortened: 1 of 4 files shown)


packages/boxel-ui/addon/src/components/card-header/index.gts
```diff
@@ -37,33 +36,27 @@ export default class CardHeader extends Component<Signature> {
...
-  }
   <template>
     <header
       data-test-card-header
       class={{cn is-editing=(bool @onFinishEditing)}}
       {{setCssVar
-        boxel-card-header-background-color=this.headerColor
-        boxel-card-header-text-color=(getContrastColor this.headerColor)
+        --boxel-card-header-background-color=@headerColor
+        --boxel-card-header-text-color=(getContrastColor @headerColor)
```
lukemelia: setCssVar is supposed to add the "--" prefix if it is not present

lukemelia (commented):
Possible to get a test to prevent a regression on this?

jurgenwerk (approved):

richardhjtan (approved):

PR #2040, opened 2025-01-14
matrix: Add room retention policy

backspace:
This adds a retention policy to auth rooms created by a realm Matrix user. It also backfills the policy to existing rooms, this can be deleted after it has run in production. I already ran it in staging, you can confirm that your user’s rooms have a policy with this script after having stored  `matrix-service` in `$E` via Ember Inspector:

```
roomRows = [];
for (r of (await $E._client.getJoinedRooms()).joined_rooms) {
    roomRow = {roomId: r};
    roomRows.push(roomRow);

    roomRow.members = (Object.keys((await $E._client.getJoinedRoomMembers(r)).joined).join(', '));
    try {
      await $E._client.http.authedRequest('GET', `/rooms/${r}/state/m.room.retention`);
      roomRow.hasRetention = (true);
    } catch(e) {
      roomRow.hasRetention = (false);
    }
}

console.table(roomRows);
```

(This would only apply for users that existed when this branch was deployed.)


packages/matrix/tests/auth-rooms.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    expect(realmRoomIds.length).toBe(realmUsers.length);
+
+    for (let room of realmRoomIds) {
+      let retentionPolicy = await getRoomRetentionPolicy(
+        user.accessToken,
+        room,
+      );
+
+      expect(retentionPolicy).toMatchObject({
+        max_lifetime: 60 * 60 * 1000,
```
backspace, line 74: I don’t love having this duplicated vs importing the constant from `runtime-common` but if you look at the CI history you can see a type-checking [can](https://github.com/cardstack/boxel/actions/runs/12818334257/job/35743616398#step:17:14) of [worms](https://github.com/cardstack/boxel/actions/runs/12832477955/job/35785295504#step:12:41) 😞

lukemelia: I ran into the same challenge with matrix event name constants

lukemelia (approved):

packages/matrix/tests/auth-rooms.spec.ts
backspace: ugh… it’s really a shame, I just lack the understanding of how to get through the typechecking minefield

lukemelia: Maybe create an issue to review this with @ef4

PR #2041, opened 2025-01-14
Cs 7691 add event card

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7691/add-event-card

1. create Deal event card that extends from event card
2. enhance and create isolated, atom and edit format of deal event card
3. link event card to CRM Deal

Result: 
image


burieberry (approved):

packages/experiments-realm/crm/deal-event.gts
```diff
@@ -0,0 +1,242 @@
...
+        background-color: white;
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+      }
+    </style>
+  </template>
+}
+
+// @ts-ignore
```
richardhjtan, line 209: Is this line needed?

richardhjtan (approved):

packages/experiments-realm/crm/deal-event.gts
lucaslyl: Unfortunately, it is needed temporarily. When I use 'export class DealEvent extends Event,' it causes a type error.

richardhjtan: Got it

PR #2042, opened 2025-01-14
CS-7790 CS-7791 New CRM task tab and planner

richardhjtan:
Refer to
- https://linear.app/cardstack/issue/CS-7790/create-new-task-tab-in-crm-app
- https://linear.app/cardstack/issue/CS-7791/define-and-implement-work-tracker-into-task-tab

### What is changing
- add new `Task` tab in CRM app
- add new Task Planner into Task tab

**CRM Task Planner**
- pretty similar, but in isolated mode: we are depending on CRM id parameter to load tasks

**CRM Task**
- similar like sprint task, extends `Task`
- instead of Project, we tied a task to an CRM app ( this might solve the query mentioned [here](https://linear.app/cardstack/issue/CS-7700/enable-creation-of-new-crm-app) )
- instead of TeamMember, we use `Contact` with Representative as assignee because our representative (staff) save as contact right now
- links to `Account`
- links to `Deal` 

**Screenshot**
Screenshot 2025-01-14 at 11 04 33 PM


packages/experiments-realm/crm-app.gts
```diff
@@ -429,6 +439,12 @@ class CrmAppTemplate extends Component<typeof AppCard> {
             class='crm-app-grid'
           />
         {{/if}}
+        {{#if (eq this.activeTabId 'Task')}}
+          <CRMTaskPlannerIsolated
+            @model={{this.args.model}}
+            @context={{@context}}
+          />
+        {{/if}}
```
richardhjtan: Glint is not happy with these lines of code, might need to tweak the code

Screenshot 2025-01-14 at 10 23 17 PM

packages/experiments-realm/crm/crm-task.gts
```diff
@@ -0,0 +1,407 @@
...
+  @field subtasks = linksToMany(() => CRMTask);
+  @field status = contains(CRMTaskStatusField);
+
+  @field title = contains(StringField, {
+    computeVia: function (this: CRMTask) {
+      return this.name;
+    },
+  });
+
+  @field assignee = linksTo(() => Contact);
```
richardhjtan: Overriding assignee field but glint complains about requiring email, which initially declare from original Task
  `Property 'email' is missing in type 'Contact' but required in type 'User'`

packages/experiments-realm/crm/task-planner.gts
```diff
@@ -0,0 +1,643 @@
...
+
+type FilterType = 'status' | 'assignee';
+
+export interface SelectedItem {
+  name?: string;
+  label?: string;
+  index?: number;
+}
+
+export class CRMTaskPlannerIsolated extends Component<typeof CRMTaskPlanner> {
```
tintinthong: was hoping there was more re-use of a component here. Can we extract a component from the isolated work tracker first. 

If I had a list of todos and days of the week, cud I make a work tracker from that using the same template

richardhjtan: @tintinthong indeed it's a good time to extract and reuse the base task planner component before it grows up on other use cases.


In this new base task planner component, we are able to pass dynamic columns config from the consumer level. Hence its doable on your scenario to map columns by days of the week inside the config

You may see some examples within the sprint planner and crm task planner in this PR, thanks

packages/experiments-realm/components/base-task-planner.gts
```diff
@@ -0,0 +1,620 @@
...
+export interface TaskColumn {
+  title: string;
+  cards: TaskCard[];
+}
+
+export interface TaskCollection {
+  columns: TaskColumn[];
+}
+
+export class BaseTaskPlannerIsolated<
```
tintinthong, line 153: Is it more straightforward to get this to be a GlimmerComponent that is used inside the isolated template?

richardhjtan: @tintinthong  I was trying GlimmerComponent with straightforward type declaration like this

```
export class BaseTaskPlannerIsolated extends GlimmerComponent<BaseTaskPlannerIsolatedSignature> {
```
It turns out there is a bunch of lint errors because the base task planner is not extending to the type to SprintPlanner/TaskPlanner

https://github.com/cardstack/boxel/actions/runs/12823222126/job/35757386460?pr=2042


If I use something like this with GlimmerComponent it works now, but not sure is more straightforward
```
export class BaseTaskPlannerIsolated<
  T extends typeof CardDef = typeof CardDef,
> extends GlimmerComponent<
  BaseTaskPlannerIsolatedSignature & { Args: { model: InstanceType<T> } }
> {
```

richardhjtan: I reverted to original approach

tintinthong (approved):
Looks good. File a ticket for better refactoring of the task planner component

richardhjtan:
Ticket for refactoring: https://linear.app/cardstack/issue/CS-7888/refactor-task-planner-component

PR #2047, opened 2025-01-15
Refactor `loadMessage` function in room resource.

FadhlanR:
**Problem**
I initially investigated the scrolling position issue in the AI panel, where it would jump back to the top whenever new messages were added or updated. The root cause was that we were reinstantiating the messages in the room resource every time a new event occurred. This caused Ember to rerender all the messages in the room, preventing the browser from maintaining the scrolling position.

**Solution**
In this PR, I refactored the loadMessage function to avoid reinstantiating messages when new events occur. Instead, the event now updates only the necessary parts of the message.

Bugs that are fixed with this PR:
- Scrolling jumping issue ([CS-7745](https://linear.app/cardstack/issue/CS-7745/when-there-is-a-lot-of-back-and-forth-conversation-with-the-ai-and-you))
- Message not streaming in ([CS-7840](https://linear.app/cardstack/issue/CS-7840/bug-messages-not-streaming-in))
- Command result not always rendering ([CS-7841](https://linear.app/cardstack/issue/CS-7840/bug-messages-not-streaming-in))


packages/host/app/services/matrix-service.ts
```diff
@@ -1313,7 +1313,7 @@ export default class MatrixService extends Service {
           ? JSON.parse(event.content.data)
           : event.content.data
       ) as CommandResultWithOutputContent['data'];
-      this.ensureCardFragmentsLoaded(data.cardEventId, roomData);
+      await this.ensureCardFragmentsLoaded(data.cardEventId, roomData);
```
FadhlanR, line 1316: Sometimes there is an error says "no cardFragments found". And I noticed that we forgot to add await to this line. It caused card message event are able to be processed without card fragment event.

backspace: nice catch, I wish there were a way to lint against this but I gather it’s not that simple

packages/host/app/lib/matrix-classes/message-builder.ts
```diff
@@ -147,23 +144,51 @@ export default class MessageBuilder {
...
+    effectiveEventId,
+    status,
+    commandResultCardEventId,
+    owner,
+  );
+  return messageCommand;
+}
+
+export function formattedMessageForCommand(formattedBody: string) {
+  return `<p data-test-command-message class="command-message">${formattedBody}</p>`;
```
backspace: I know this is just copied from a different position but does anyone know if we have XSS protection or the like for this? Or is it a safe assumption that malicious messages of this event type can’t be produced?

backspace (approved):

jurgenwerk (approved):

packages/host/app/lib/matrix-classes/message-builder.ts
FadhlanR: I just added this [43de05a](https://github.com/cardstack/boxel/pull/2047/commits/43de05a15305aa015d71d4ff7278d3bc8df8cb92).

lukemelia (commented):
I think this is a good approach overall. I don't like that the logic of creating messages (in MessageBuilder) and the logic of updating messages (in RoomResource) is separated. It feels like it should be the responsibility of the same class.

packages/host/app/components/matrix/room.gts
```diff
@@ -81,7 +81,7 @@ export default class Room extends Component<Signature> {
           @registerConversationScroller={{this.registerConversationScroller}}
           @setScrollPosition={{this.setScrollPosition}}
         >
-          {{#each this.messages as |message i|}}
+          {{#each this.messages key='eventId' as |message i|}}
```
habdelra, line 84: nice!

packages/host/app/lib/matrix-classes/message-builder.ts
```diff
@@ -147,23 +147,53 @@ export default class MessageBuilder {
...
+    effectiveEventId,
+    status,
+    commandResultCardEventId,
+    owner,
+  );
+  return messageCommand;
+}
+
+export function formattedMessageForCommand(formattedBody: string) {
+  return `<p data-test-command-message class="command-message">${sanitizeHtml(
```
habdelra, line 196: @backspace  since we are not using triple curlies `{{{` this is not actually ever HTML. this is always only text. there is no need to sanitize this.

habdelra (approved):

PR #2048, opened 2025-01-15
load room events in timestamp order

lukemelia:
- I observed on my local machine and I suspect on a production account, that a cardFragment event was being processed after the message that referred to it, causing a load failure. This PR causes the order to be based on origin_server_ts


jurgenwerk (approved):
I experienced this too, but I thought I had some messed up data due to various testing. Thanks!

PR #2049, opened 2025-01-15
query the active deal & lifetime Value on CRM account

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7833/query-active-deals-in-crm-account

**Added active deals query functionality**
- Queries deals linked to current account with status label not equal to "Closed Won" or "Closed Lost"
- Calculates total number of active deals and their combined value
- Displays results in the "Active Deals" summary card

**Added lifetime value metrics**
- Queries all "Closed Won" deals for the account

  Calculates:
  1. Total lifetime value
  2. Current year's value
  3. Previous year's value
  4. Year-over-year growth
- Displays formatted results in the "Lifetime Value" summary card

Result: 
Screenshot 2025-01-16 at 10 54 41


packages/experiments-realm/crm/account.gts
```diff
@@ -127,6 +131,166 @@ class IsolatedTemplate extends Component<typeof Account> {
...
+      growth,
+    };
+  }
+
+  get formattedLifetimeTotal() {
+    const total = this.lifetimeMetrics.total;
+    if (total === 0) {
+      return '$0';
+    }
+    const formattedTotal = (total / 1000).toFixed(1);
```
richardhjtan: @lucaslyl it would be great if we include the thousand separator eg 1,250k. Can refer to deal summary usage

Number(formattedTotal).toLocaleString('en-US', {
  minimumFractionDigits: 2,
})

packages/experiments-realm/crm/account.gts
```diff
@@ -127,6 +131,166 @@ class IsolatedTemplate extends Component<typeof Account> {
...
+          module: new URL('./deal', import.meta.url).href,
+          name: 'Deal',
+        },
+        every: [
+          { eq: { 'account.id': this.args.model.id ?? '' } },
+          {
+            not: {
+              any: [
+                { eq: { 'status.label': 'Closed Won' } },
+                { eq: { 'status.label': 'Closed Lost' } },
```
tintinthong: I thought there is a computed `isActive`??

packages/experiments-realm/crm/account.gts
```diff
@@ -127,6 +131,166 @@ class IsolatedTemplate extends Component<typeof Account> {
...
+    const formattedTotal = (total / 1000).toFixed(1);
+    return `$${formattedTotal}k`;
+  }
+
+  get formattedCurrentYearValue() {
+    const currentYearValue = this.lifetimeMetrics.currentYear;
+    const formattedValue =
+      currentYearValue === 0
+        ? '+$0'
+        : `+$${(currentYearValue / 1000).toFixed(1)}k`;
```
richardhjtan: Same comment here

packages/experiments-realm/crm/account.gts
tintinthong: Maybe you have to add `isUsed` inside the `isActive` field bcos we don't use `isActive` in the isolated template

richardhjtan (approved):
Overall looks good!

packages/experiments-realm/crm/account.gts
lucaslyl: > I thought there is a computed `isActive`??

Yeah, I tried using isActive initially, but it seems I couldn't query the data. I'm not sure if it's because isActive is computed based on the status label, so it didn't work for the query. However, I tried querying from the root status label, and it worked. I will test it again to confirm

tintinthong: You need to add `isUsed` otherwise the query will return no data

lucaslyl: Thank for point out the issue! done implement it

packages/experiments-realm/crm/account.gts
lucaslyl: Thank for advise, done replace it with minimumFractionDigits

tintinthong (approved):

PR #2051, opened 2025-01-15
Add job priority support and prioritize user initiated indexing requests

habdelra:
This PR add job priority support by allowing you to configure workers that will only process jobs greater than or equal to a particular priority. This allows us to designate workers that will only process user initiated indexing requests (as well as workers that will process any jobs system or user initiated). 

This PR uses this new mechanism to set user initiated indexing jobs (incremental indexing and from-scratch indexing related to realm creation) to use the higher priority workers. This will ensure that workers are available to process user initiated index mutations while the realm server spins up and performs background indexing on all the realms on the server.

Paired with PR https://github.com/cardstack/infra/pull/571


backspace (approved):

lukemelia (approved):

lukemelia:
For the future, this would have been easier to review if the `private` -> `#` changes were in their own commit.

jurgenwerk (approved):

PR #2053, opened 2025-01-16
Do not use CardsGrid when using embedded-view

burieberry:
This is to prevent embedded-card view from getting cropped.


lukemelia (approved):
Good cleanup, as well as avoiding the problematic layout issue. 👍

tintinthong (approved):

jurgenwerk (approved):

PR #2054, opened 2025-01-17
Exclude service name of active LLM model

FadhlanR:
Screenshot 2025-01-17 at 10 19 28


jurgenwerk (approved):
I tested this in the deployment preview link and it looks good

PR #2056, opened 2025-01-17
Deleting cards that have an error

jurgenwerk:
This PR:

1. Adds an option to delete an error card when it's on the stack 

image

2. Fixes deleting such items from the index card (previously this wasn't working due to a card loading error - it wanted to load the card but this is not possible because the card has an error)

image

This needed a refactor in the card deletion logic where previously these functions were accepting a `CardDef`–but in case when card has an error we do not actually have a `CardDef` since the card cannot be loaded. So we simply pass down the card's `id` instead.


packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -278,31 +283,43 @@ export default class InteractSubmode extends Component<Signature> {
...
+              ?.cardTitle;
+
+            if (!cardTitle) {
+              throw new Error(`Could not get card title for ${card}`);
+            }
 
-        const stackItem = here.allStackItems.find(
-          (item) => item.card === loadedCard,
-        );
-        // if is workspace index card, do not allow deletion
```
jurgenwerk, line 296: We're not even showing a button to delete the card when the card is an index card, and that is covered with tests, so this check here is probably not needed.

packages/host/app/services/operator-mode-state-service.ts
```diff
@@ -172,27 +175,32 @@ export default class OperatorModeStateService extends Service {
     }
   }
 
-  async deleteCard(card: CardDef) {
+  async deleteCard(cardId: string) {
```
jurgenwerk, line 178: the parameter changed from `CardDef` to a `cardId` because we want to support deleting a card that has an error, but in this case it is not possible to load the card so we only have its `id`. Card's id is the lowest common denonimator of a loaded card and the card's error response, so we use that.

packages/runtime-common/realm-index-query-engine.ts
```diff
@@ -152,6 +153,7 @@ export class RealmIndexQueryEngine {
           errorDetail: instance.error,
           scopedCssUrls,
           lastKnownGoodHtml: instance.isolatedHtml ?? null,
+          cardTitle: instance.searchDoc?.title ?? null,
```
jurgenwerk, line 156: I'm adding this so that we can show the card's title in the delete modal (when the card has an error):

image

habdelra (approved):

packages/experiments-realm/SprintTask/cocojambo.txt
lukemelia, line 1: Is this meant to be part of this PR?

lukemelia (approved):

packages/experiments-realm/SprintTask/cocojambo.txt
jurgenwerk: Oh good catch, no it's a leftover from testing. Will delete this.

PR #2057, opened 2025-01-17
Invalidation performance improvement

habdelra:
This PR improves invalidation performance. I've used `EXPLAIN ANALYZE` on our query that discovers files to invalidate. In doing so I've found that our `SORT` incurs unnecessary overhead (there is no need to sort these results, they are not displayed), as well as our `COALESCE` which actually subverts our `GIN` index on the `deps` column. Also, I added a new compound index on the `type` and `realm_url` columns as indicated in the `EXPLAIN`. These items together cut the invalidation query to about <10ms in staging (which amounts to a lot in the aggregate).

The much bigger item, tho, is to skip the retrieval of the ignore rules when there are no ignore rules (which is almost all the time). This saves about 8-10 sec on the from-scratch indexing when booting.

TODO:
- [ ] measure performance after prod deployment


packages/host/app/lib/current-run.ts
```diff
@@ -284,21 +284,27 @@ export class CurrentRun {
...
-
     let mtimesStart = Date.now();
     let filesystemMtimes = await this.#reader.mtimes();
     perfLog.debug(
       `time to get file system mtimes ${Date.now() - mtimesStart} ms`,
     );
+
+    let ignoreFile = new URL('.gitignore', url).href;
+    // it costs about 10 sec to try to get the ignore file when it doesn't
+    // exist, so don't get it if it's not there.
```
backspace, line 295: wow 😯

backspace (approved):

lukemelia (approved):

PR #2060, opened 2025-01-20
Add task and query tasks in deal page

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7793/query-active-task-in-deal

What is changing
- add active task in deal page
- query active tasks in deal page


tintinthong (approved):

packages/experiments-realm/crm/deal.gts
```diff
@@ -360,6 +443,37 @@ class IsolatedTemplate extends Component<typeof Deal> {
...
+              >
+                New Task
+              </BoxelButton>
+            </:icon>
+            <:content>
+              <div>
+                {{#if this.activeTasks.isLoading}}
+                  Loading...
+                {{else if this.activeTasks.instances}}
+                  {{this.activeTasks.instances.length}}
```
tintinthong, line 467: We shud update all Loading to use the LoadingIndicatorIcon

richardhjtan: @tintinthong  Sounds good, I will separate this into a single UI ticket

richardhjtan: @tintinthong I think of it when there is a lot "loading" state from each section, a lot of loading indicators will show up, which may look cluttered. I will leave the ticket for trying to replace it, and see if the UX is good

PR #2061, opened 2025-01-20
Add copy operation to indexer

habdelra:
In this PR we add a new type of indexer operation that allows a realm's index to be copied to another realm (in the future I could see this being fine tuned so that you can copy just a specific list of items from one realm to another realm instead of the whole realm's contents). The idea is that the source realm has already incurred the cost of rendering all the various cards, so that we don't have to pay that cost again when we want to copy an item that is already in the index. Near term  this should dramatically speed up the ability to create new realms. in the long term this logic can be leveraged for copying a graph of cards and their modules from one realm into another.


packages/host/app/lib/current-run.ts
```diff
@@ -122,50 +122,26 @@ export class CurrentRun {
...
+  static async fromScratch(current: CurrentRun): Promise<IndexResults> {
     let start = Date.now();
     log.debug(`starting from scratch indexing`);
     perfLog.debug(
       `starting from scratch indexing for realm ${current.realmURL.href}`,
     );
-
     current.#batch = await current.#indexWriter.createBatch(current.realmURL);
     let invalidations: URL[] = [];
-    if (invalidateEntireRealm) {
```
habdelra, line 137: this was a shortcut that we were doing for creation of new realms that is less efficient than the new copy operation, so we are getting rid of this.

lukemelia (approved):
Looks good overall. Do we need validation somewhere that the provided source realm URL is for a realm that is managed by this realm server?

habdelra:
> Looks good overall. Do we need validation somewhere that the provided source realm URL is for a realm that is managed by this realm server?

sure, that's a good idea

PR #2063, opened 2025-01-20
Convert worker-manager TCP server to an HTTP server for DX

lukemelia:


lukemelia:
@habdelra thoughts about whether and where to test this?

packages/realm-server/worker-manager.ts
```diff
@@ -75,63 +81,93 @@ let isExiting = false;
...
+    let [{ queue_depth }] = (await query(dbAdapter, [
+      `SELECT COUNT(*) as queue_depth FROM jobs WHERE status='unfulfilled'`,
+    ])) as {
+      queue_depth: string;
+    }[];
+    result = {
+      ...result,
+      highPriorityWorkers: highPriorityCount,
+      allPriorityWorkers: allPriorityCount,
+      queueDepth: parseInt(queue_depth, 10),
```
jurgenwerk: Just my personal opinion that `pendingJobs` would be a bit clearer name for this.

jurgenwerk (approved):
I tested this locally in the browser and I was able to see the stats.

habdelra:
So i think the way I would approach testing for this would be to create a new realm server test module for this. in that module I'd set a random DB name using `prepareTestDB()` from the realm-server's test helpers. I would then spawn a child process to start the worker manager in the `beforeEach()`, which in turn will spawn workers and start the HTTP server.. You can hold a reference to the worker manager's process, and in the `afterEach()` kill the worker manager process which should kill the workers since they are subprocesses of the worker manager. since each test uses a different DB, hopefully that will help to accommodate the async used to spin down the worker in case the torn down workers haven't completed torn down by the time the next test starts. Within the body of each test then, you can use the pg queue's publisher to publish jobs to the queue. 

controlling queue depth might be a little tricky for the test. perhaps that means actually you publish jobs to the queue first before you spin up the worker manager--that way you can be sure of how many jobs are in the queue. and after the worker manager spins up you should see the queue depth start to decrease.

habdelra:
Also keep in mind that in production right now there are _2_ worker managers. because there are 2 ECS tasks in the worker ECS service. How do you intend to expose the worker manager HTTP for each ECS worker task? For state originating from the DB it probably doesn't matter, since it will be the same--but for worker manager specific state, like number of running workers (and perhaps the PID of the workers) this may not necessarily be the same

packages/realm-server/worker-manager.ts
```diff
@@ -75,63 +81,93 @@ let isExiting = false;
...
+  let result = {
+    ready: isReady,
+  } as Record<string, boolean | number>;
+  if (isReady) {
+    let [{ queue_depth }] = (await query(dbAdapter, [
+      `SELECT COUNT(*) as queue_depth FROM jobs WHERE status='unfulfilled'`,
+    ])) as {
+      queue_depth: string;
+    }[];
+    result = {
```
habdelra: another metric that might be nice to report on would be uncompleted jobs in jobs_reservations that have been running longer than some threshold and are not expired

packages/realm-server/worker-manager.ts
```diff
@@ -75,63 +81,93 @@ let isExiting = false;
...
+  .use(ecsMetadata);
 
-  process.on('SIGINT', shutdown);
-  process.on('SIGTERM', shutdown);
-  process.on('uncaughtException', (err) => {
-    log.error(`Uncaught exception in worker manager:`, err);
-    shutdown();
-  });
+webServer.on('error', (err: any) => {
+  console.error(`worker manager HTTP server error: ${err.message}`);
```
habdelra: let's use our logger instead

packages/realm-server/worker-manager.ts
```diff
@@ -75,63 +81,93 @@ let isExiting = false;
...
+
+const shutdown = (onShutdown?: () => void) => {
+  log.info(`Shutting down server for worker manager...`);
+  webServerInstance.closeAllConnections();
+  webServerInstance.close((err?: Error) => {
+    if (err) {
+      log.error(`Error while closing the server for worker manager HTTP:`, err);
+      process.exit(1);
     }
+    dbAdapter.close(); // warning this is async
```
habdelra: ignoring this async might make testing trickier.

packages/realm-server/worker-manager.ts
```diff
@@ -75,63 +81,93 @@ let isExiting = false;
...
+    log.info(`worker manager HTTP on port ${port} has stopped.`);
+    onShutdown?.();
```
habdelra, line 149: consider moving this into the `dbAdapter.close().finally()`...

packages/realm-server/worker-manager.ts
```diff
@@ -75,63 +81,93 @@ let isExiting = false;
...
+    process.exit(0);
```
habdelra, line 150: consider moving this also into the `dbAdapter.close().finally()`

lukemelia:
After discussion with @habdelra, I am going to update this branch to retain the TCP -> HTTP conversion but drop the queue depth. In a separate PR, I will add an endpoint to realm server to report on queue stats starting with queue depth.

habdelra (approved):

PR #2002, opened 2025-01-06
CS-7690 All deals summary section

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7690/query-summary-for-all-the-deals-in-deal-page

### What is changing
- add a summary section in All Deals page
- implement query to be fetch when `Deals` tab is active, and/or filter changes within Deals (eg: All Deals to Proposal )

**Screenshots**
Screenshot 2025-01-07 at 12 51 05 AM
Screenshot 2025-01-07 at 1 04 03 AM

PR #2068, opened 2025-01-21
CS-7889 task page lhs filter

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7889/task-page-lhs-filter-narrow-down-the-task-by-condition

What is changing
- new set of CRM task filter
- query changes when CRM active filter updating

Screenshot
https://github.com/user-attachments/assets/18a15839-7ef7-4851-ae87-7987cd8727c4


packages/experiments-realm/crm/task-planner.gts
```diff
@@ -119,8 +125,38 @@ export class CRMTaskPlannerIsolated extends BaseTaskPlannerIsolated<
       },
     };
     super(owner, args, config);
+
+    // Initialize query once
+    this.cardsQuery = getCards(this.getTaskQuery, this.realmHrefs, {
+      isLive: true,
+    });
```
tintinthong: Why isn't getTaskQuery the only one which is changing? I think that is the expectation for the resource pattern. Generally, we shudn't reassign the resource particularly inside of a getter

richardhjtan: @tintinthong yes it should be only query changing. I have same thought that `getCards` API should be reactive when query changes, but currently it is not. 

One example I did before was in deal summary page, we do a restartable tasks when the filter changes: https://github.com/cardstack/boxel/pull/2002. This PR currently use this approach as well.

packages/experiments-realm/crm-app.gts
```diff
@@ -193,6 +209,16 @@ class CrmAppTemplate extends Component<typeof AppCard> {
...
+  private setupTaskPlanner = (taskPlanner: CRMTaskPlannerIsolated) => {
+    this.taskPlannerAPI = taskPlanner;
+  };
+
+  private debouncedLoadTaskCards = debounce(() => {
+    if (this.activeTabId === 'Task') {
+      this.taskPlannerAPI.loadCards.perform();
+    }
+  }, 300);
+
```
richardhjtan: I was eager to look for a way to let the parent component that able to ask child component to do some operation (in this example we ask component to reload the cards), until I got the inspiration of stack and stack items communicating with each other. This approach helps

packages/experiments-realm/crm/task-planner.gts
tintinthong: this is surprising to me

tintinthong:
This should fix everything https://github.com/cardstack/boxel/pull/2088. Depends on core team review and follow-up

tintinthong (approved):

PR #2069, opened 2025-01-21
Split realm-server-test.ts into two files, one for realm endpoints and one for global endpoints

lukemelia:


backspace (approved):
I’m happy to see this, the file was beyond unwieldy!

PR #2070, opened 2025-01-21
Add _queue-status endpoint to realm server

lukemelia:


backspace (approved):

packages/realm-server/handlers/handle-queue-status.ts
```diff
@@ -0,0 +1,33 @@
...
+            },
+          },
+        }),
+        {
+          headers: { 'content-type': SupportedMimeType.JSONAPI },
+        },
+      ),
+    );
+  };
+}
```
habdelra: is it ok that this is an unprotected route? perhaps we should have a shared secret to access this? otherwise a malicious client might try to DDoS our DB impacting the system overall.

jurgenwerk: In our previous projects we had a `x-data-integrity-checks-authorization` header was used to authorize requests of this type. Perhaps we could have something similar here, depending on the service that will consume this

jurgenwerk (approved):
LGTM!

PR #2077, opened 2025-01-22
Add "copy from seed realm" checkbox

FadhlanR:
In this PR, I added a checkbox to the create workspace form to specify whether to copy files from the seed realm.


packages/realm-server/server.ts
```diff
@@ -336,15 +338,25 @@ export class RealmServer {
...
+          },
+        });
+        this.log.debug(
+          `seed files for new realm ${url} copied to ${realmPath}`,
+        );
+      } else {
+        copySync(
+          join(this.seedPath!.replace(/\/$/, ''), 'index.json'),
+          join(realmPath!.replace(/\/$/, ''), 'index.json'),
+        );
```
habdelra: good! i was worried about this, glad you noticed it

packages/realm-server/server.ts
```diff
@@ -336,15 +338,25 @@ export class RealmServer {
...
+      }
     }
 
     let realm = new Realm(
```
habdelra, line 369: you are going to have to also make sure to not set the `copiedFromRealm` value in the `Realm` constructor options below (after you merge from upstream) when `copyFromSeedRealm` is false.

packages/realm-server/server.ts
habdelra: I wonder if it's simpler to just use `writeJSONSync` to manually set this file like we do for `.realm.json`--it's really small. that way you don't depend on a `seedPath` which is optional
```suggestion
        writeJSONSync(join(realmPath, 'index.json'), {
          "data": {
            "type": "card",
            "meta": {
              "adoptsFrom": {
                "module": "https://cardstack.com/base/cards-grid",
                "name": "CardsGrid"
              }
            }
          }
        });
```

habdelra: (also, I thought the `join` took care of the trailing `/`, look at how we set the `.realm.json` above.)

packages/realm-server/server.ts
FadhlanR: Updated it here https://github.com/cardstack/boxel/pull/2077/files#diff-14c49ddb85659961b6177a036fc15d7166291f81169c9a1dd0dfead082ef68e5R385

jurgenwerk (approved):

packages/realm-server/server.ts
```diff
@@ -365,7 +380,13 @@ export class RealmServer {
...
       },
       {
-        ...(this.seedRealmURL ? { copiedFromRealm: this.seedRealmURL } : {}),
+        ...(this.seedRealmURL
+          ? {
+              copiedFromRealm: copyFromSeedRealm
+                ? this.seedRealmURL
+                : undefined,
+            }
+          : {}),
```
habdelra: it's easier to read if you just don't include it if `copyFromSeedRealm` is false (that's what this splat logic is doing). that way there is one less ternary to include in the logic.

```suggestion
        ...(this.seedRealmURL && copyFromSeedRealm
          ? {
              copiedFromRealm: this.seedRealmURL
            }
          : {}),
```

habdelra (approved):
Minor nit, but otherwise looks good

PR #2078, opened 2025-01-22
Do not exclude unused contained computed fields from search doc

lukemelia:


jurgenwerk (approved):

packages/host/app/lib/current-run.ts
```diff
@@ -505,6 +505,7 @@ export class CurrentRun {
           identityContext,
         },
       );
+      await api.flushLogs();
```
habdelra, line 508: make sure to await any async in the API so that the computeds can be calculated before inserting into the index. (likely this was a bug that was always there, but this PR exposed it)

habdelra (approved):

PR #2079, opened 2025-01-22
Fix issue where realm server tests were not exiting promptly

habdelra:
I was able to use `wtfnode` to identify the timers that were preventing the node tests from exiting promptly and used `.unref()` to inform node not to wait on the timer to finish before it can exit.

### DIFF
--- a/packages/runtime-common/realm.ts
+++ b/packages/runtime-common/realm.ts
@@ -728,14 +728,15 @@ export class Realm {
         if (!request.headers.get('X-Boxel-Building-Index')) {
           let timeout = await Promise.race<void | Error>([
             this.#startedUp.promise,
-            new Promise((resolve) =>
-              setTimeout(() => {
-                resolve(
-                  new Error(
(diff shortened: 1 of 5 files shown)


packages/runtime-common/realm.ts
```diff
@@ -728,14 +728,15 @@ export class Realm {
...
-              }, 60 * 1000),
+            new Promise(
+              (resolve) =>
+                setTimeout(() => {
+                  resolve(
+                    new Error(
+                      `Timeout waiting for realm ${this.url} to become ready`,
+                    ),
+                  );
+                }, 60 * 1000).unref?.(),
```
habdelra, line 739: `unref()` only exists in node and since this is isomorphic code we use `?` to make it nice for the browser which does not have `unref()`

lukemelia (approved):

jurgenwerk (approved):
Was nice to learn about `wtfnode` and `unref`, thanks

PR #2081, opened 2025-01-23
Default to anthropic/claude-3.5-sonnet in code mode

jurgenwerk:


packages/host/app/components/matrix/room.gts
```diff
@@ -239,6 +241,24 @@ export default class Room extends Component<Signature> {
     registerDestructor(this, () => {
       this.scrollState().messageVisibilityObserver.disconnect();
     });
+
+    this.setRoomLLM();
```
lukemelia: This looks like it running in the constructor of the room component. It seems like a surprising place to me that might miss certain cases. For example, creating a new room, or switching rooms?

jurgenwerk: My reasoning behind doing it this way is because I thought this made sense only for newly created rooms. But now I see it would be better to support switching rooms too, and move the logic out of the constructor. This is now fixed in later commits

packages/host/app/services/matrix-service.ts
```diff
@@ -1337,6 +1338,25 @@ export default class MatrixService extends Service {
       await this.client?.scrollback(room);
     }
   }
+
+  async setLLMModelForCodeMode() {
```
backspace: could be `setLLMForCodeMode` to not have the redundant “model” within the acronym

packages/host/app/services/operator-mode-state-service.ts
```diff
@@ -310,6 +312,10 @@ export default class OperatorModeStateService extends Service {
   updateSubmode(submode: Submode) {
     this.state.submode = submode;
     this.schedulePersist();
+
+    if (submode === Submodes.Code) {
+      this.matrixService.setLLMModelForCodeMode();
```
backspace: For two repetitions it seems okay but needing this here and in the assistant panel suggests to me we need a hook or something like that that lets us be able to specify setup during a state change.

backspace (approved):
I tried this in the preview deployment and it worked, there’s a brief moment before the model gets updated, that’s probably fine but does that indicate that there’s a different default that gets overridden, could it be the actual default instead?
screencast 2025-02-03 10-43-32

packages/host/app/services/matrix-service.ts
jurgenwerk: Sure, that does sound awkward when reading out loud, will change it

jurgenwerk:
https://github.com/user-attachments/assets/fe5f9b49-76f1-4614-b9dd-6def442f8af5

@backspace I tested it, it looks like the flicker is gone, perhaps https://github.com/cardstack/boxel/pull/2109 solved it?

jurgenwerk:
I'm going to merge this, if you notice the flicker again we can open a new ticket for that.

PR #2083, opened 2025-01-23
Isolated deal card style adjustments

burieberry:
Changes:
- Updated deal card background color and other styles & spacing to match latest designs on zeplin
- Fixed broken border style
- Fix html elements that were used incorrectly (liberal use of `label` element, `article` wasn't always appropriate)
- Alignment for tags on the side of the header if there's space

deal-isolated


tintinthong (approved):
Looks good thanks

PR #2085, opened 2025-01-24
Cleanup LLM prompt content, which also fixes sonnet model related issues

jurgenwerk:
This PR cleans up the "Thinking..." messages that are sent to the LLM in every prompt. I think these are unnecessary to include, and they cost more tokens. 

Luckily, this optimization also fixes AI bot crashes when using sonnet 😳

image

The real reason behind these crashes is unknown - all we can get from OpenRouter is "internal server error".

### DIFF
--- a/packages/ai-bot/constants.ts
+++ b/packages/ai-bot/constants.ts
@@ -1,0 +1,1 @@
+export const thinkingMessage = 'Thinking...';
--- a/packages/ai-bot/helpers.ts
+++ b/packages/ai-bot/helpers.ts
@@ -29,6 +29,7 @@ import {
   DEFAULT_LLM,
   APP_BOXEL_ACTIVE_LLM,
 } from '@cardstack/runtime-common/matrix-constants';
+import { thinkingMessage } from './constants';
 
 let log = logger('ai-bot');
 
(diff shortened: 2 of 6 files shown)


habdelra (approved):

IanCal:
Can we use the isStreamingFinished flag on the events? Looks like we should add that to the thinking... message but then we'd have a boolean to look for rather than doing string matching.

jurgenwerk:
@IanCal I am not sure what you mean here–`isStreamingFinished` is already present on the "Thinking..." events too. In this PR i just filter out thinking messages when building the prompt, I think it's just noise for the LLM.

IanCal:
@jurgenwerk We don't set it as we don't send it currently through the debounced sender, here's one of the recorded event lists from the tests
```
{
    "type": "m.room.message",
    "room_id": "!EtlFgGnAaFCLoozHPV:localhost",
    "sender": "@aibot:localhost",
    "content": {
      "body": "Thinking...",
      "msgtype": "m.text",
      "formatted_body": "Thinking...",
      "format": "org.matrix.custom.html",
      "m.new_content": {
        "body": "Thinking...",
        "msgtype": "m.text",
        "formatted_body": "Thinking...",
        "format": "org.matrix.custom.html"
      }
    },
    "origin_server_ts": 1736864365020,
    "unsigned": {
      "age": 69139
    },
    "event_id": "$VyCYG0DAuudwMGN8hStuSVfXrjMVZu2ZcJgIs9sIwLE",
    "user_id": "@aibot:localhost",
    "age": 69139
  },
```

(we don't have to use the debounced sender, but that's currently the only place we add the flag iiuc, we could add it when we send the thinking event too)

My thought around this was instead of looking for "thinking" messages by string matching, just to remove all events where we've not finished streaming.

burieberry (approved):

jurgenwerk:
After talking to @IanCal it looks like we have a deeper problem and this might be just a bandaid solution. We probably shouldn't have multiple "Thinking" events in between other events (events that we send to the LLM). I will take a deeper look tomorrow

jurgenwerk:
@IanCal I followed your suggestion to add `isStreamingFinished: false` to the `Thinking...` events that the bot sends, and filter these out when constructing the prompt. This way we can get rid of that last message that says `Thinking...` that was attached to the prompt. 

And regarding my claim we have **multiple** `Thinking...` events that we send to the LLM–that is actually false. Looks like I was inspecting these events at a wrong place.

PR #2088, opened 2025-01-26
Allow getCards to be reactive wrt to changes in query arg

tintinthong:
Ecosystem team had trouble using `getCards` like a resource. Particularly, changing query argument causse the search resource to recompute -- this did not work. This let to some confusing workarounds that force the re-instantiation of the resource. The main issue is for `ember-resources` we expect `Search.from(this,()=> objectWithArgs)` to exist in the root of the componet but we are wrapping it in a function and each arg item is passed as a value -- I  changed the expected reactive arg item into thunks

**Note**
- **Adding card api test**: this PR introduce some test to the usage of card api of card users. I added this test suite with the intent of making clear what card authors have access to and the correct way to use it
- **Promote getCards as a publicAPI**: I intentionally promoted since card users would want to use it
- **Suspicion**: My suspicion is that other resources use the same pattern and maybe accidentally re-instantiating as opposed to sitting on the component and recomputing

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -28,8 +28,8 @@ import type RealmServerService from '../services/realm-server';
 
 interface Args {
   named: {
-    query: Query;
-    realms?: string[];
+    query: Query | undefined;
+    realms: string[] | undefined;
     isLive?: true;
     doWhileRefreshing?: (ready: Promise<void> | undefined) => Promise<void>;
   };
@@ -47,7 +47,13 @@ export class Search extends Resource<Args> {
 
   modify(_positional: never[], named: Args['named']) {
     let { query, realms, isLive, doWhileRefreshing } = named;
-    this.realmsToSearch = realms ?? this.realmServer.availableRealmURLs;
+    if (query === undefined) {
+      return;
+    }
+    this.realmsToSearch =
+      realms === undefined || realms.length === 0
+        ? this.realmServer.availableRealmURLs
+        : realms;
 
     this.loaded = this.search.perform(query);
     waitForPromise(this.loaded);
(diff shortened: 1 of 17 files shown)


packages/host/app/resources/search.ts
```diff
@@ -202,32 +211,27 @@ export class Search extends Resource<Args> {
...
+}
+
+export function getSearch(
   parent: object,
-  query: Query,
-  realms?: string[],
+  getQuery: () => Query | undefined,
+  getRealms?: () => string[] | undefined,
   opts?: {
     isLive?: true;
```
tintinthong, line 225: The core change bottoms out here

packages/host/app/resources/search.ts
```diff
@@ -47,7 +47,13 @@ export class Search extends Resource<Args> {
 
   modify(_positional: never[], named: Args['named']) {
     let { query, realms, isLive, doWhileRefreshing } = named;
-    this.realmsToSearch = realms ?? this.realmServer.availableRealmURLs;
+    if (query === undefined) {
+      return;
+    }
```
tintinthong, line 52: I think its very possible that we build the query asynchronously. It is no good, sometimes to query a default value as it will cause blinking and an unnecessary fetch.

packages/host/tests/integration/components/card-api-test.gts
```diff
@@ -0,0 +1,258 @@
...
+        }
+        get realms() {
+          return [testRealmURL]; //TODO: Investigate where meta is returned. Offer a context where the symbol exists
+        }
+        resource = this.args.context?.actions?.getCards
+          ? this.args.context.actions.getCards(
+              () => this.query,
+              () => this.realms,
+            )
+          : undefined;
```
tintinthong, line 186: Here is an example of usage of the api

lukemelia (approved):
LGTM, perhaps consider explaining in a comment near the getCards signature why it requires functions instead of args passed directly

tintinthong:
This also fixes a bug in boxel spec preview whereby the boxel spec accordion doesn't re-render when u choose another card export

PR #2089, opened 2025-01-26
Support non tool call models

IanCal:
This change loads (at boot of the aibot) all models currently on openrouter that support tool calls.

While most parameters are safe to add in for providers that don't support them (they are just ignored) tool calls result in a 404 error.

I've tested swapping from a model that supports tools and using them (e.g. patching) to ones that don't and that seems to be fine. 

Note - R1 is currently fairly slow and we don't return the reasoning tokens so the latency can be very high.


lukemelia (approved):

jurgenwerk:
I'm trying this locally and I can't seem to get deepseek working:

image

I'm getting this error: ` 'missing finish_reason for choice 0'` (after several minutes)

PR #2090, opened 2025-01-27
remove aspect ratio from grid-view-container

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7918/remove-grid-view-container-aspect-ratio

- remove the aspect-ratio property when you need more control over the element's height
- we need to explicitly set a specific height for the element that does not align with the aspect ratio

example:The deal grid view requires more height on the card, but the aspect ratio will block the custom height.
Screenshot 2025-01-27 at 12 39 40


burieberry (commented):
Fine to remove these from here, however, can you add the aspect-ratio and possibly the height in Blog App css where this is used? Also, any change you make to files that also exist in seed realm, you'll need to make the same changes there and confirm that it works as expected.

burieberry (commented):
Blog app also needs the `--default-grid-view-height: max-content;`, otherwise items overlap when screen is narrowed.

overlap

lucaslyl:
> Blog app also needs the `--default-grid-view-height: max-content;`, otherwise items overlap when screen is narrowed.
> 
> overlap

I have applied --grid-view-height instead of --default-grid-view-height to the blog app CSS, and thank you for the advice on this.

image

burieberry (approved):

PR #2091, opened 2025-01-27
CS-7888 refactor task planner component

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7888/refactor-task-planner-component

### What is changing
- refactor and improve base task planner component to be more straightforward 
- refactor CRM task planner
- refactor Sprint Planner


tintinthong (approved):
This is good improvement that makes everything simpler. There are some changes in mind but I will handle it while ur on hols

richardhjtan:
Cool @tintinthong ,thanks for the pairing previously

PR #2098, opened 2025-01-28
Playground panel accordion item

burieberry:
- display playground panel only when a card def is selected
- add background and instance chooser
- instance chooser updates dropdown options when selected card def changes (within same file or when a different file is selected)
playground-panel
playground-panel-selected


packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -496,6 +498,17 @@ export default class CodeSubmode extends Component<Signature> {
     return undefined;
   }
 
+  private get displayPlayground() {
```
lukemelia: Consider renaming to `shouldDisplayPlayground` to avoid confusion with the verb usage of "display"

packages/host/app/components/operator-mode/code-submode/playground-panel.gts
```diff
@@ -0,0 +1,169 @@
...
+  @service private declare realmServer: RealmServerService;
+  @tracked private selectedItem?: CardDef;
+
+  private options = getCards(
+    () => ({
+      filter: { type: this.args.codeRef },
+      sort: [{ by: 'createdAt', direction: 'desc' }],
+    }),
+    () => this.realmServer.availableRealmURLs,
+  );
```
lukemelia, line 96: I was imagining this would be based on the recent files list in code mode (most recently viewed). But I like using this query. Maybe we should combine the ideas and run the query, then bubble up items on the recently accessed list to the top? Also, should have a max that we show in this list -- maybe 5?

lukemelia (approved):

PR #2102, opened 2025-01-29
Rename catalog entry to boxel spec

tintinthong:


lukemelia (commented):
Should we just call it a "spec" internally? Prefixing with Boxel is not something we do for anything else, and it is all in Boxel.

burieberry (approved):

tintinthong:
> Should we just call it a "spec" internally? Prefixing with Boxel is not something we do for anything else, and it is all in Boxel.

internally? Do you mean have a distinction between the way we name it in the UI vs code? 

An alternative is to change **Boxel Spec -> Code Spec** which was what Chris used to call it last time. 

I am ok with these options
- Option 1: Call it "Spec" in the ui and code
- Option 2: Call it "Code Spec" in the ui and code
- Option 3: Call it "Boxel Spec" in the ui (wherever design shows) and code refer to it as "Spec"

PR #2104, opened 2025-01-30
Refactor links to many component

tintinthong:


tintinthong:
This is wrong. We indeed need the @model (the outer containing box) being passed in and its not a substitute for arrayField. I shall just leave it as it is

PR #2105, opened 2025-01-30
new arg to filter by subclass ref

tintinthong:
Purpose of this PR is to enable spec to choose linkedExamples that link by type 

Screenshot 2025-02-05 at 16 36 17

Screenshot 2025-02-05 at 16 41 07

API change
- <@fields @subclassType={{ref}} />. This means users can control the linksTo and linksToMany editor via this arg. It will only work if you pass in a ref which is indeed a subclass of the type. Otherwise, it defaults to original behaviour
- I named it subclassType as the arg @lukemelia different from here https://discord.com/channels/584043165066199050/900021750367129641/1334556182001483826 bcos, I think its wrong to overwrite the field entirely. Now the api works st, you only pass the ref that way @fields.author is still true to the type of author but you can for example, <@fields.author @subclassType=bigAuthorRef>. The issue with adding another filter to the every = [ ] in the card chooser is that the index never supports 2 types so we always have to commit to one


packages/base/links-to-editor.gts
```diff
@@ -143,6 +147,7 @@ export class LinksToEditor extends GlimmerComponent<Signature> {
 
   private chooseCard = restartableTask(async () => {
     let type = identifyCard(this.args.field.card) ?? baseCardRef;
+    type = await getNarrowestType(this.args.subclassType, type, myLoader());
```
burieberry: Alternatively, you can only call this function if a subclassType arg was provided

burieberry: Actually do you even need to check if the `type` is the base CardDef? We know that all cards descend from it, so maybe in that case you can skip loading the whole chain and just use the provided subtype.

burieberry (commented):
This is all good, but I think it would be fine to make an exception and not check `isInsideAncestorChain` when the `type` we're comparing to is the base carddef. `isInsideAncestorChain` will end up loading all the cards in the ancestor chain when we already know that all card subtypes descend from card def. In that case, you can just confirm via `isCardDef` for the subtype if we're not sure if the type is a card.

packages/base/links-to-editor.gts
tintinthong: > Actually do you even need to check if the type is the base CardDef? We know that all cards descend from it, so maybe in that case you can skip loading the whole chain and just use the provided subtype.

I think we do. We need to check if type is base CardDef as a terminating condition to "stop checking".

tintinthong: @burieberry maybe we can talk thru this after standup. Probably some small misunderstanding

tintinthong: @burieberry I have made the change that you have suggested

packages/runtime-common/code-ref.ts
```diff
@@ -227,3 +227,27 @@ export function humanReadable(ref: CodeRef): string {
...
+async function isInsideAncestorChainOfCardDef(
+  codeRef: CodeRef,
+  loader: Loader,
+): Promise<boolean | undefined> {
+  let card = await loadCard(codeRef, { loader: loader });
+  if (isCardDef(card)) {
+    return true;
+  }
+  return false;
+}
```
burieberry: I thought about this and I think it would be good to add to this some of your previous logic. This makes sense if the `type` we're comparing to is base CardDef, or if we only care that the subtype is a card def, regardless of what the `type` we're comparing to is. If `type` is not base card def, then we should consider if we want to confirm that `subtype` is descended from it. We currently only use this `subtype` argument for BoxelSpec, but making this change would allow us to use it in different cases where the `type` is not base card def. I'm open to hearing what other people in the team thinks about this.

tintinthong: For context, this is the diff change https://github.com/cardstack/boxel/pull/2105/commits/02e8408b776f4dea179978271b6f1febc0246f9b. Previously, we use to have more generic handling of the subtype

packages/runtime-common/code-ref.ts
```diff
@@ -227,3 +227,27 @@ export function humanReadable(ref: CodeRef): string {
 function assertNever(value: never) {
   return new Error(`should never happen ${value}`);
 }
+
+async function isInsideAncestorChainOfCardDef(
```
habdelra: javascript has a similar kind of function called `typeof`. maybe that is a clearer name for this method: `isTypeOf()`

habdelra: Actually, i take that back, this looks like a synonmym for isCardDef. let's just overload that specific method so it takes either a codeRef and loader and an already running card:

```
export function isCardDef(card: any): card is typeof CardDef;
export function isCardDef(codeRef: CodeRef, loader: Loader): boolean
export function isCardDef(cardOrCodeRef: any, loader?: Loader): boolean { ... }
```
(not too sure how all the type assertion works out since you can't assert a CodeRef  param is a CarDef). if the overload doesn't work, at least try to name this is similar as you can to `isCardDef`, so that it is clear that we are using the same semantics.

packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -326,6 +326,33 @@ module('Spec preview', function (hooks) {
     await click('[data-test-create-spec-instance]');
     assert.dom('[data-test-exported-type]').hasText('card');
   });
+
+  test('can add linked examples to spec instance narrows the chooser by code ref inside of the instance', async function (assert) {
```
lukemelia:
```suggestion
  test('when adding linked examples, card chooser options are narrowed to this type', async function (assert) {
```

packages/base/card-api.gts
```diff
@@ -1092,7 +1093,13 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
     }
     return class LinksToComponent extends GlimmerComponent<{
       Element: HTMLElement;
-      Args: { Named: { format?: Format; displayContainer?: boolean } };
+      Args: {
+        Named: {
+          format?: Format;
+          displayContainer?: boolean;
+          subclassType?: ResolvedCodeRef;
```
lukemelia: Perhaps `typeConstraint` or `allowedClass` would be a better name?

packages/runtime-common/code-ref.ts
tintinthong: Ahh ok I think this worked but loadCard is async so I used .then within the isCardDef implementation. I also added a note on using `isCardDef` in `getNarrowestType`

lukemelia (approved):

burieberry (approved):
Looks much better

PR #2106, opened 2025-01-30
Host command to add a field to a card definition

jurgenwerk:
Adds a new host command so that the bot (AI assistant in code mode) will be able to add fields to a gts file. 

TODO:
- [ ] Use the command in the "Add field" functionality in the schema editor


jurgenwerk:
This is already reviewable, however Luke suggested to use this command in the "Add field" functionality in the schema editor, so I added it as a TODO.

packages/runtime-common/module-syntax.ts
```diff
@@ -434,12 +440,12 @@ function makeNewField({
...
-  if (
-    fieldRef.module.startsWith(baseRealm.url) &&
-    fieldRef.name === 'default'
-  ) {
-    // primitive fields
-    return `@${fieldDecorator.name} ${fieldName} = ${fieldTypeIdentifier.name}(${fieldCardIdentifier.name});`;
+  if (sourceCodeForComputedField) {
+    return `@${fieldDecorator.name} ${fieldName} = ${fieldTypeIdentifier.name}(${fieldCardIdentifier.name}, {
+              computeVia: function () {
+                return ${sourceCodeForComputedField}
```
lukemelia: This approach precludes being able to write a multi-line computed function. Perhaps instead, the end developer should be required to pass code that evaluates to a function, and then that function should we used as the computeVia.

packages/base/command.gts
```diff
@@ -54,6 +54,7 @@ export class WriteTextFileInput extends CardDef {
   @field content = contains(StringField);
   @field realm = contains(StringField);
   @field path = contains(StringField);
+  @field url = contains(StringField);
```
lukemelia: I see why this is convenient, but it's going to make it harder for the AI models to generate valid input to this command. Can we get away without it?

jurgenwerk: I see, I removed it, and added `realm` and `path` to `AddFieldToCardDefinitionInput` command:  https://github.com/cardstack/boxel/pull/2106/commits/8fd9004c5037ad7a8b5a04290de14e99d73e0d7e

packages/runtime-common/module-syntax.ts
jurgenwerk: Yeah that was unnecessary limiting. I modified the approach so that the user should provide the whole function implementation: https://github.com/cardstack/boxel/pull/2106/commits/9345fd9b7d779a67294e706caa8e5cbb2568f345

lukemelia (approved):

tintinthong (approved):
nice

PR #2107, opened 2025-01-30
Adding renderType flag to prerendered search endpoint

FadhlanR:
This PR introduces a new parameter called `renderType` to the prerendered search endpoint. If this parameter is defined, we will return the corresponding HTML; if it is not defined or if the HTML is not found, we will return whatever is available. Additionally, I have added a `usedRenderType` attribute to inform the client of the HTML that was ultimately used.


lukemelia (approved):
It would be good for @habdelra to review this when he gets back from vacation too (not a blocker)

burieberry (approved):

PR #2110, opened 2025-01-31
Remove description from top level of function calls

FadhlanR:
- [x] Removed the description from the function call parameters.
- [x] Used the content of the response message instead of the function call description.


packages/ai-bot/tests/responding-test.ts
```diff
@@ -260,87 +260,4 @@ module('Responding', (hooks) => {
       'The tool call event should replace the thinking message',
     );
   });
-
-  test('Sends tool call event separately when content is sent before tool call', async () => {
```
FadhlanR, line 264: I think this test is no longer relevant.

packages/ai-bot/helpers.ts
```diff
@@ -42,7 +42,8 @@ You may make multiple function calls, all calls are gated by the user so multipl
 If a user asks you about things in the world, use your existing knowledge to help them. Only if necessary, add a *small* caveat at the end of your message to explain that you do not have live external data. \
 \
 If you need access to the cards the user can see, you can ask them to attach the cards. \
-If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown.';
+If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown. \
+Always provide a content text explanation before calling a function';
```
lukemelia: @IanCal does this need to be phrased differently to help it keep the context text and the function in the same response?

lukemelia (approved):
I'd like @IanCal 's review on this before merging

packages/ai-bot/helpers.ts
IanCal: It'll never return more than one response, I would be tempted to leave this until we see lots of calls we want an explanation with coming without them, and ideally have evals to see the difference it makes. It may be different for different models.

packages/ai-bot/lib/send-response.ts
```diff
@@ -106,17 +109,19 @@ export class Responder {
...
-          this.roomId,
-          this.deserializeToolCall(toolCall),
-          this.initialMessageReplaced ? undefined : this.initialMessageId,
-        );
+        let optionPromise = sendOption({
+          client: this.client,
+          roomId: this.roomId,
+          content: msg.content,
+          functionCall: this.deserializeToolCall(toolCall),
+          eventToUpdate: this.initialMessageId,
```
IanCal, line 123: If there is more than one tool call, does this still work? It looks like it might replace the initial event with each tool call resulting in just one at the end

IanCal:
Without a change to host, it looks like this will still create the same wrapper around everything? Ideally we want to go from

```
toolCall({
description: blah
attributes: {things we care about}
})
```
to
```
toolCall({things we care about})
```

packages/ai-bot/lib/send-response.ts
FadhlanR: It seems we need to update our message structure. Currently, a message can have only one command, but we need it to support multiple commands.

FadhlanR:
I initially thought we wanted to remove the description from function call parameters because we could get the explanation from the content instead. However, after discussing it with Ian, I realized that the main goal is to eliminate the layering of parameters when triggering toolCall, and removing the description wouldn't address that. So, we decided to close this PR.

PR #2116, opened 2025-02-04
Move SSE to Matrix

backspace:
This repurposes existing Matrix “session rooms” (where the auth conversation is conducted) to also broadcast “realm events”, which were formerly transmitted via Server-Sent Events. It includes several TODOs to completely remove the traces of SSE, including:
* `service:message-service` no longer needs to exist
* `$REALM/…_message` endpoint for subscribing can be removed
* file watchers to generate realm events in development will go away, relying instead on Boxel Tools

It’s unfortunately a large PR but there were no obvious seams to further decompose it at. Thankfully most of the changes are mechanical, replicated across many files.

The high-level changes to `host` tests are:
* removal of `setupServerEvents` and the `TestContextWithSSE` that provided it
* flattening of `expectEvents`, no more callback
  * most assertions on expected events are gone, they are covered elsewhere
* occasional addition of `settled` and `waitUntil` to ensure realm events have propagated
* addition of `setupMockMatrix` when missing, the `MockUtils` it returns is required for realm setup now

Otherwise the changes include:
* adding Matrix event types for each flavour of realm event
* updating the interfaces of event listeners to use those new types
* removal of a similar `expectEvent` helper in the realm server’s `realm-endpoints-test`


packages/host/tests/helpers/mock-matrix.ts
```diff
@@ -53,6 +81,17 @@ export function setupMockMatrix(
...
+      'service:matrix-mock-utils',
+      {
+        async load() {
+          return mockUtils;
+        },
+      },
+      {
+        instantiate: false,
+      },
+    );
```
lukemelia, line 94: Maybe we can think of a more elegant way to do this?

lukemelia (approved):
I don't have time to try this out right now, but the code looks good. Very excited to see the test improvements land!

packages/host/tests/helpers/mock-matrix.ts
backspace: 👍🏻 I finally [faced it](https://github.com/cardstack/boxel/pull/2116/commits/1a1ef30b08ef9645a4bc8027b00d85a1d4b4c51c), thanks

lukemelia: Much better! 👏

habdelra (approved):
ditto with what luke said. overall looks great, but too large for me to get deep into it.

PR #2117, opened 2025-02-04
Add some more feedback that an AIbot response content is streaming in

lukemelia:
https://github.com/user-attachments/assets/f73243ca-edcd-470c-a4b5-32b57a5536b1

OLD
----
https://github.com/user-attachments/assets/7d2bf61b-160f-4609-b1c7-49c5fb23e20c

It would be nice to get the dot appear on the in-progress list item too but not sure how to target that with CSS

### DIFF
--- a/packages/host/app/components/ai-assistant/message/index.gts
+++ b/packages/host/app/components/ai-assistant/message/index.gts
@@ -129,6 +129,7 @@ export default class AiAssistantMessage extends Component<Signature> {
         'ai-assistant-message'
         is-from-assistant=@isFromAssistant
         is-pending=@isPending
+        is-streaming=@isStreaming
         is-error=@errorMessage
       }}
       {{MessageScroller index=@index registerScroller=@registerScroller}}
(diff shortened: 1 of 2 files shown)


packages/host/app/components/ai-assistant/message/index.gts
```diff
@@ -303,6 +304,21 @@ export default class AiAssistantMessage extends Component<Signature> {
...
+        height: 8px;
+        background: currentColor;
+        border-radius: 50%;
+        display: inline-block;
+        font-family: system-ui, sans-serif;
+        line-height: normal;
+        vertical-align: baseline;
+        margin-left: 5px;
+      }
+
```
backspace, line 350:
```
.is-streaming .content[data-scopedcss-b61d7f149b-f81bdaca25] > :not(div):last-child:after {
    content: '…';
}
```

It’s not the same look but does suggest continuation?

Boxel 2025-02-04 15-42-40

backspace (approved):
It looks good, I couldn’t figure out a way to continue the same look for list items sadly, it’s too bad there’s nothing to target text nodes after all these years.

(I merged from `main` to hopefully fix the Percy diffs)

jurgenwerk (commented):
When sending a message, just before the bot starts `Thinking...`, a dot is shown in the users's message:

image


Is this intended? Should we scope the css to target ai bot messages only?

backspace (approved):
I shied away from reaching into Javascript to add the marker, but a modifier seems like a minimal intervention and the consistent styling is great 🎉

PR #2121, opened 2025-02-05
Fix loading state rhs

tintinthong:
The idea is of this PR

- get loading state of spec to work inside its content
- change schema editor to do the same 
- schema editor to have the same inline style for playground, spec-preview and schema editor (the positioning is not the same tho). I guess we can also make schema editor work like the other but I notice the outer div with ...attributes means a lot of styling is gotten from the parent

**Schema editor**

Before: 

Screenshot 2025-02-05 at 16 40 18

After:

Screenshot 2025-02-05 at 20 46 36

**Spec**

Before: Spec also worked like this where  but I hid it in a previous PR. 

After: 

Screenshot 2025-02-05 at 20 49 31

### DIFF
--- a/packages/host/app/components/operator-mode/card-adoption-chain.gts
+++ b/packages/host/app/components/operator-mode/card-adoption-chain.gts
@@ -62,12 +62,7 @@ export default class CardAdoptionChain extends Component<Signature> {
     </div>
     <style scoped>
       .loading {
-        align-content: center;
-        text-align: center;
-        font-weight: 500;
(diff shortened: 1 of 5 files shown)


tintinthong:
@burieberry should we make spec to have the loading on the top left instead??

burieberry:
> @burieberry should we make spec to have the loading on the top left instead??

Would be good to keep them all consistent

burieberry (approved):
Looks good

PR #2123, opened 2025-02-06
Cs 6799 eliminate dialog creating spec

tintinthong:
Previously, when clicking Create for the boxel spec we navigated the user to a create-file-modal that gave the user the option which realm to create the spec at. Now, we just create it on-on-the-spot

For cards/fields that have no boxel spec, simply create spec instance in the same realm and not navigate the create-file-modal

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/spec-preview.gts
+++ b/packages/host/app/components/operator-mode/code-submode/spec-preview.gts
@@ -57,7 +57,6 @@ interface Signature {
   Element: HTMLElement;
   Args: {
     selectedDeclaration?: ModuleDeclaration;
-    isCreateSpecInstanceRunning: boolean;
   };
   Blocks: {
     default: [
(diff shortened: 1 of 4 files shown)


lukemelia (changes requested):
Shouldn't there be a corresponding change in tests?

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -329,6 +319,44 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
     };
   }
 
+  private createSpecInstance = restartableTask(
```
lukemelia: In what situation would this task restart?

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -329,6 +319,44 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
   }
 
+  private createSpecInstance = task(
+    async (ref: ResolvedCodeRef, specType: SpecType) => {
+      let relativeTo = new URL(ref.module);
+      let maybeRef = codeRefWithAbsoluteURL(ref, relativeTo);
+      let realmURL = this.operatorModeStateService.realmURL;
+      if ('name' in maybeRef && 'module' in maybeRef) {
+        ref = maybeRef;
+      }
```
burieberry: alternatively can use `isResolvedCodeRef` from runtime-common

burieberry (approved):
Looks good to me

lukemelia (approved):
Are we handling the case where you can't write to the realm of the module?

PR #2126, opened 2025-02-07
File Picker for Attaching Files

FadhlanR:
https://github.com/user-attachments/assets/b4d6a44d-1295-449b-9ed4-7c1a46ab20da

- [x] Refactored directory and file tree component so it can reused in the attach file modal.
- [x] Added feature flag for attaching files in AI Assitant.
- [x] Added attach file modal component.


jurgenwerk:
@FadhlanR I have a draft PR up for displaying `FileDef`s in the AI assistant (auto-attached file in code mode). Perhaps you can rebase work on top of that. It is also ok if you'd like to merge this first without integrating with my PR, and do the integration in the AI assistant later. But let's think about how to minimize merge conflicts. 

https://github.com/cardstack/boxel/pull/2135

FadhlanR:
> @FadhlanR I have a draft PR up for displaying `FileDef`s in the AI assistant (auto-attached file in code mode). Perhaps you can rebase work on top of that. It is also ok if you'd like to merge this first without integrating with my PR, and do the integration in the AI assistant later. But let's think about how to minimize merge conflicts.
> 
> #2135

@jurgenwerk if this PR is ok to merge, I'll merge it and will rebase your PR. The feature flag introduced in this PR also can be used in yours.

packages/host/app/components/pill.gts
```diff
@@ -5,85 +5,105 @@ import Component from '@glimmer/component';
...
-    return this.args.card.constructor.getComponent(this.args.card);
+    return this.args.item.constructor.getComponent(this.args.item);
   }
 
   get hideIconRight() {
-    return !this.args.onToggle && !this.args.removeCard;
+    return !this.args.onToggle && !this.args.remove;
+  }
+
+  get id() {
```
jurgenwerk: There is an alternative approach in https://github.com/cardstack/boxel/pull/2135 where we have a `CardPill` and a `FilePill`. Perhaps that is better than having so many conditions in a `Pill` that try to deal with 2 different things.

packages/host/app/components/pill.gts
```diff
@@ -5,85 +5,105 @@ import Component from '@glimmer/component';
...
   IconButton,
-  Pill,
+  Pill as BoxelPill,
   RealmIcon,
   Switch,
 } from '@cardstack/boxel-ui/components';
-import { cn } from '@cardstack/boxel-ui/helpers';
-import { IconX } from '@cardstack/boxel-ui/icons';
+import { cn, cssVar } from '@cardstack/boxel-ui/helpers';
+import { IconX, CodeFile } from '@cardstack/boxel-ui/icons';
```
jurgenwerk: Don't we already have an icon for this? It's named a bit differently: `import FileCode from '@cardstack/boxel-icons/file-code';`

packages/host/app/components/pill.gts
```diff
@@ -5,85 +5,105 @@ import Component from '@glimmer/component';
 
 import {
   IconButton,
-  Pill,
+  Pill as BoxelPill,
```
jurgenwerk: This looks like a code smell, let's not introduce another name for a Pill, I think this speaks in favor of my suggestion for introducing new (sub)classes, which can be a CardPill and a FilePill.

packages/host/app/components/pill.gts
jurgenwerk: Let me know your thoughts, we'll have to consolidate our approaches otherwise there will be a massive conflict

FadhlanR: @jurgenwerk I rebased this PR to your branch.

burieberry (commented):
Looks good and works as expected.

packages/host/app/components/matrix/room.gts
```diff
@@ -675,8 +710,9 @@ export default class Room extends Component<Signature> {
           .map((stackItem) => stackItem.card.id),
       };
       try {
-        let files = await this.matrixService.uploadFiles(this.filesToAttach);
-
+        if (files && !files[0].url) {
```
jurgenwerk: Why is the `!files[0].url` condition needed?

FadhlanR: To handle resend message, in that case, we don't want to reupload the files

jurgenwerk: I think we do, since the file can change in the meanwhile. Before we have file change detection I think we should reupload. I suggest getting rid of the whole condition (and making this.filesToAttach be an empty array by default)

jurgenwerk:
@FadhlanR I was testing this and see this PR already includes the autoattached behaviour from https://github.com/cardstack/boxel/pull/2135. So I see no point in maintaining a separate PR anymore since your PR now has that functionality almost in whole. I suggest you merge (or cherry pick the new commits if that is easier) my latest changes (my PR branch is green and ready to go). The latest changes include removing files from being attached, a test, and some type adjustments. I chose not to implement a resource for the autoattached files like we do with attached cards because here the situation is much simpler since we don't have any stacks, just the currently opened file. Let me know if you are ok with rebasing and then we can close my PR and move the rest of the todo into a separate ticket.

FadhlanR:
> @FadhlanR I was testing this and see this PR already includes the autoattached behaviour from #2135. So I see no point in maintaining a separate PR anymore since your PR now has that functionality almost in whole. I suggest you merge (or cherry pick the new commits if that is easier) my latest changes (my PR branch is green and ready to go). The latest changes include removing files from being attached, a test, and some type adjustments. I chose not to implement a resource for the autoattached files like we do with attached cards because here the situation is much simpler since we don't have any stacks, just the currently opened file. Let me know if you are ok with rebasing and then we can close my PR and move the rest of the todo into a separate ticket.

Sounds good. I'll include your latest updates to this PR.

packages/host/app/components/matrix/room.gts
```diff
@@ -598,19 +627,17 @@ export default class Room extends Component<Signature> {
       });
     }
 
+    return cards;
+  }
+
+  private get allFilesToAttach() {
```
jurgenwerk: I think using "all" in the name is redundant. We can just use `filesToAttach`. It should include all files shown in the attachments

packages/host/app/components/ai-assistant/attachment-picker/index.gts
```diff
@@ -49,7 +49,7 @@ export default class AiAssistantAttachmentPicker extends Component<Signature> {
   <template>
     <div class='item-picker'>
       {{#each this.itemsToDisplay as |item|}}
-        {{#if (this.isCard item)}}
+        {{#if (this.displayCardItem item)}}
```
jurgenwerk: I'm confused about logic - we already have `itemsToDisplay`, it's surprising to me that we have further logic on what to display.

FadhlanR:
@jurgenwerk I've rebased this PR to your latest updates. And I made some updates for the auto-attached file.

packages/host/app/components/ai-assistant/attachment-picker/index.gts
```diff
@@ -182,8 +182,16 @@ export default class AiAssistantAttachmentPicker extends Component<Signature> {
     this.areAllItemsDisplayed = !this.areAllItemsDisplayed;
   }
 
-  private isCard = (item: CardDef | FileDef): item is CardDef => {
-    return isCardInstance(item);
+  private displayFileItem = (item: CardDef | FileDef): item is FileDef => {
+    return (
+      !isCardInstance(item) &&
+      this.args.submode === 'code' &&
```
jurgenwerk: Cards and files are allowed to coexist. That was the point of the refactor introduced in https://github.com/cardstack/boxel/pull/2130.

packages/host/app/components/matrix/room.gts
```diff
@@ -587,8 +602,22 @@ export default class Room extends Component<Signature> {
 
   @action
   private sendMessage(prompt?: string) {
+    this.doSendMessage.perform(
+      prompt ?? this.messageToSend,
+      this.operatorModeStateService.state.submode === 'interact' &&
```
jurgenwerk: I don't think we should be making differences between submodes when sending a message. Cards and files can coexist

packages/host/app/components/matrix/room.gts
```diff
@@ -280,31 +281,47 @@ export default class Room extends Component<Signature> {
...
+      },
+    });
+
+    if (!this.autoAttachedFileUrl) {
+      state.value = undefined;
+    } else {
+      state.value = this.matrixService.fileAPI.createFileDef({
+        sourceUrl: this.autoAttachedFileUrl,
+        name:
+          this.autoAttachedFileUrl.split('/').pop() ?? this.autoAttachedFileUrl,
```
jurgenwerk: Why defaulting to the url? You probably don't want name to be the url

jurgenwerk (changes requested):

packages/host/app/components/matrix/room.gts
FadhlanR: I merged `autoAttachedFile` and `filesToAttach` directly in the `sendMessage` function, following the same approach we use for `autoAttachedCards` and `cardsToAttach`.

jurgenwerk (approved):

PR #2127, opened 2025-02-07
Avoid reprocessing matrix events

IanCal:
Currently a single message to the bot and single chunk reply results in reprocessing the entire event list 14-15 times for a room. This skips which events we've processed before.

To deal with things that are out of order, this actually just reprocesses events from the first unseen event. 

Even with this we still seem to try calling load room quite a lot, this PR just lowers the cost of doing so, and makes it a bit simpler where we want to start doing auto calling of commands.

Potentially this needs to change to more strictly process the events once.

### DIFF
--- a/packages/host/app/resources/room.ts
+++ b/packages/host/app/resources/room.ts
@@ -138,7 +138,7 @@ export class RoomResource extends Resource<Args> {
     return this.members.filter((m) => m.membership === 'join');
   }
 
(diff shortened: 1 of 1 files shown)


lukemelia (approved):
I'm fine with this provided all tests pass. There's a lot going on in the event processing so I don't know 100% that this won't cause any issues.

PR #2130, opened 2025-02-10
Refactor AiAssistantCardPicker to also support picking files

jurgenwerk:
This refactor is to support both `CardDef` and `FileDef` (new addition from https://github.com/cardstack/boxel/pull/2129) in the card picker in the assistant panel (so that users will be able to attach both files and cards).

This is currently just to satisfy the existing functionality, and more file related functionality and tests will be added in the work that follows (file auto attaching + the ability to select a file).

Follow the commits for a better overview of what changed.


packages/host/app/components/ai-assistant/attachment-picker/index.gts
```diff
@@ -10,48 +10,56 @@ import { TrackedSet } from 'tracked-built-ins';
...
+
+export default class AiAssistantAttachmentPicker extends Component<Signature> {
   <template>
-    <div class='card-picker'>
-      {{#each this.cardsToDisplay as |card i|}}
-        {{#if (this.isCardDisplayed card i)}}
-          {{#if (this.isAutoAttachedCard card)}}
+    <div class='item-picker'>
+      {{#each this.itemsToDisplay as |item|}}
+        {{#if (this.isCard item)}}
```
jurgenwerk, line 43: There will also be an `{{else if (this.isFile item)}}` here in the work that follows

FadhlanR (approved):

PR #2131, opened 2025-02-10
CS-7926 Link crm cards to crm app

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7926/link-all-card-to-the-crm-app-card

### What is changing
- link all CRM cards (deals, accounts, company) to CRM app

**Challenge**

Originally, each model—such as CrmApp, Company, Contact, Account, etc.—was defined in a separate file. This approach is common for maintaining a clear separation of concerns and modularity.

These models needed to reference each other. For instance, both Contact and Account referenced CrmApp, while Account also referenced Contact, and Company was referenced further down the chain. This results in circular dependency issues.

Since multiple models depended on each other (with several linking to CrmApp), the interdependencies became problematic when they were spread across different files.

**Resolution**

Merging Models into a Single File

All the interdependent models were merged into a single file. This also ensures that the required models are loaded beforehand before the model is used.


richardhjtan:
Luke mentioned that linksTo thunk type should resolve the circular dependency issue, I will try another approach to check linksTo thunk type works with the original models.

richardhjtan:
Thunk type method makes the approach cleaner compare to this approach, hence closing this PR.

PR #2132, opened 2025-02-10
Commands on skill cards

IanCal:
This covers manually approved/applied commands.

Key points

* Commands can be added to skill cards. Each one has a flag for whether it requires user approval (for use in a following PR but it's important it's nested) and a *function name*. 
* The function name is constructed from the coderef, so the name is stable and the same command added from two different skills is the same.
* The schemas are important *only* for the aibot, so the host adds them to the room when required but does not 
* There are some potential loading issues around skill cards, there is an async call inside a sync getter. This may be responsible for other cases where some things are showing or not.

Next steps:

- [x] Bring back the test for AIbot
- [x] Add host test
- [x] Neaten the hashing function in skill cards
- [x] Update all commands to export with the name as well
- ~~Cache commands that get added to the room?~~


packages/base/skill-card.gts
```diff
@@ -1,11 +1,56 @@
...
+
+  @field functionName = contains(StringField, {
+    description: 'The name of the function to be executed',
+    computeVia: function (this: CommandField) {
+      if (!this.codeRef?.module || !this.codeRef?.name) {
+        return '';
+      }
+
+      // Simple hash function that works in all environments
+      const djb2 = (str: string): string => {
```
IanCal: the md5 library seems to operate differently in host and here, so I used the simple djb2 hash function

packages/base/skill-card.gts
```diff
@@ -1,11 +1,56 @@
...
+        for (let i = 0; i < str.length; i++) {
+          const char = str.charCodeAt(i);
+          hash = (hash << 5) - hash + char;
+          hash = hash & hash; // Convert to 32-bit integer
+        }
+        return Math.abs(hash).toString(16).slice(0, 4);
+      };
+
+      const input = `${this.codeRef.module}#${this.codeRef.name}`;
+      return `${this.codeRef.name}_${djb2(input)}`;
```
IanCal: Function names look like `SwitchSubmodeCommand_ab43`

packages/host/app/lib/matrix-classes/message-command.ts
```diff
@@ -73,6 +73,21 @@ export default class MessageCommand {
...
 
+  async getCommandCodeRef() {
+    let roomResource = this.matrixService.roomResources.get(
+      this.message.roomId,
+    );
+    if (!roomResource) {
+      return undefined;
+    }
+    // We need to wait for the skills to actually be loaded
+    await roomResource.waitForAllSkills();
```
IanCal: Without this, the first time you load the page, they don't work, clicking again works fine.

packages/host/app/resources/room.ts
```diff
@@ -167,8 +170,9 @@ export class RoomResource extends Resource<Args> {
         continue;
       }
       let cardId = cardDoc.data.id;
+      let loadingPromise;
       if (!this._skillCardsCache.has(cardId)) {
-        this.cardService
+        loadingPromise = this.cardService
```
IanCal: Is there a better way of waiting for this all to be "settled"? Either existing or a better common setup?

IanCal:
Commands on skills lets us add things like grounding with wikidata easily
[Screencast from 11-02-25 15:02:35.webm](https://github.com/user-attachments/assets/d0b53606-452b-4dab-af91-9a046bfc4374)

packages/base/skill-card.gts
```diff
@@ -1,11 +1,55 @@
...
+
+  @field functionName = contains(StringField, {
+    description: 'The name of the function to be executed',
+    computeVia: function (this: CommandField) {
+      if (!this.codeRef?.module || !this.codeRef?.name) {
+        return '';
+      }
+
+      const hashed = djb2_xor(`${this.codeRef.module}#${this.codeRef.name}`);
+      return `${this.codeRef.name}_${hashed.slice(0, 4)}`;
```
lukemelia: Relying on the name without the module seems problematic to me. We don't really want to try to enforce how userland commands are organized do we?

packages/host/app/resources/room.ts
```diff
@@ -51,6 +53,7 @@ interface SkillId {
   skillCardId: string;
   skillEventId: string;
   isActive: boolean;
+  loadingPromise?: Promise<void>;
```
lukemelia: @ef4 was counseling against resources exposing promises -- let's talk this through.

packages/host/app/services/command-service.ts
```diff
@@ -130,6 +131,19 @@ export default class CommandService extends Service {
...
+      // If we don't find it in the one-offs, start searching for
+      // one in the skills we can construct
+      if (!commandToRun) {
+        // here we can get the coderef from the messagecommand
+        let commandCodeRef = await command.getCommandCodeRef();
+        if (commandCodeRef) {
+          let CommandConstructor = (await getClass(
+            commandCodeRef,
+            this.loaderService.loader,
+          )) as { new (context: CommandContext): Command<any, any> };
```
lukemelia, line 143: Maybe this type should get a name?

packages/host/app/services/matrix-service.ts
```diff
@@ -576,11 +580,61 @@ export default class MatrixService extends Service {
...
+      },
+    });
+  }
+
   async addSkillCardsToRoomHistory(
     skills: SkillCard[],
     roomId: string,
     opts?: CardAPI.SerializeOpts,
   ): Promise<string[]> {
+    const commandDefinitions = skills.flatMap((skill) => skill.commands);
```
lukemelia, line 638: Do we need to ensure uniqueness here?

lukemelia (commented):
This is coming together well overall.

PR #2136, opened 2025-02-11
Refactor CardResource to create primordial store

habdelra:
This PR refactors the `CardResource` so that we leverage a single backing service which becomes our primordial store. `StoreService` that is responsible for all the card loading and processing the live updates when the client receives invalidations from the server. This means that there is just a single handler for each realm that services all the `CardResource` instances that may exist. Notably, this is able to trigger a loader reset a single time when there are multiple invalidations that are part of an update to code. The `CardResource` then becomes mearly a reference counter that is used for lifetime management. Eventually we need to refactor callers of the `CardResource` to stop calling `CardResource.loaded` and simply be reactive to the fact that a card is initially undefined and will eventually be available.


burieberry (commented):
Tested this branch, and this resolves the playground preview error I was seeing 👍

lukemelia:
@ef4 please review

packages/host/app/resources/card-resource.ts
```diff
@@ -145,43 +90,36 @@ export class CardResource extends Resource<Args> {
...
-        this.removeLiveCardEntry(this.url);
-      }
-      this.unsubscribeFromRealm();
+      this.realmSubscription.unloadResource(this);
     });
   }
 
   get card() {
-    if (this.loadLiveModel.isRunning || this.loadStaticModel.isRunning) {
+    if (this.loadModel.isRunning || this.loadModel.isRunning) {
```
lukemelia: Is this redundant or is this redundant?

habdelra: Yeah, looks like that can be cleaned up

habdelra:
@ef4 I'm ready for another review. the `CardResource` has been trimmed down to the bare minimum (pending some more refactors to get rid of the `CardResource.loading()` promise.)

packages/host/app/resources/search.ts
```diff
@@ -156,53 +159,61 @@ export class Search extends Resource<Args> {
...
-            `${realm}_search?${stringify(query)}`,
-          );
-          if (!isCardCollectionDocument(json)) {
-            throw new Error(
-              `The realm search response was not a card collection document:
-        ${JSON.stringify(json, null, 2)}`,
+    // we cannot use the `waitForPromise` test waiter helper as that will cast
+    // the Task instance to a promise which makes it uncancellable. When this is
+    // uncancellable it results in a flaky test.
+    let token = waiter.beginAsync();
```
habdelra, line 165: FYI: casting a EC Task to a promise strikes again--please remember to never `await` an EC task outside of the task. Within the task `await` is just syntactic sugar for a `yield`.

lukemelia (commented):
LGTM

PR #2137, opened 2025-02-12
AI bot: ability to download the attached file and use it in the prompt to LLM

jurgenwerk:
When files are attached, the bot downloads them and includes them in the prompt. For now, we just put them in the chat context. In the following work, we will add skills to teach the bot about working with modules, and commands that will allow the bot to manipulate the attached file, for example a gts file.  

Here is how the attached files get included in the message sent to the LLM - for example we are attaching `todo.gts`:

image

The initial message sent to the LLM will have the contents of the attached files included (check the text below where it says "Attached files"):

```
`The user is using an application called Boxel, where they are working on editing "Cards" which are data models representable as JSON. The user may be non-technical and should not need to understand the inner workings of Boxel. The user may be asking questions about the contents of the cards rather than help editing them. Use your world knowledge to help them. If the user request is unclear, you may ask clarifying questions. You may make multiple function calls, all calls are gated by the user so multiple options can be explored.If a user asks you about things in the world, use your existing knowledge to help them. Only if necessary, add a *small* caveat at the end of your message to explain that you do not have live external data. If you need access to the cards the user can see, you can ask them to attach the cards. If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown. If you encounter code, please indent code using 2 spaces per tab stop and enclose the code within triple backticks and indicate the language after the opening backticks so that the code is displayed stylishly in Markdown.
The user currently has given you the following data to work with:

Cards: 

Attached files:
todo.gts: import {
  CardDef,
  field,
  contains,
  StringField,
} from 'https://cardstack.com/base/card-api';
import MarkdownField from 'https://cardstack.com/base/markdown';

export class Todo extends CardDef {
  static displayName = 'Todo';
  @field details = contains(MarkdownField);
}


The user has given you the following instructions. You must obey these instructions when responding to the user:
- If the user wants the data they see edited, AND the patchCard function is available, you MUST use the "patchCard" function to make the change.
- If the user wants the data they see edited, AND the patchCard function is NOT available, you MUST ask the user to open the card and share it with you.
- If you do not call patchCard, the user will not see the change.
- You can ONLY modify cards shared with you. If there is no patchCard function or tool, then the user hasn't given you access.
- NEVER tell the user to use patchCard; you should always do it for them.
- If the user wants to search for a card instance, AND the "searchCard" function is available, you MUST use the "searchCard" function to find the card instance.
Only recommend one searchCard function at a time.
If the user wants to edit a field of a card, you can optionally use "searchCard" to help find a card instance that is compatible with the field being edited before using "patchCard" to make the change of the field.
 You MUST confirm with the user the correct choice of card instance that he intends to use based upon the results of the search.
```


packages/ai-bot/tests/prompt-construction-test.ts
```diff
@@ -1270,7 +1277,7 @@ test('Return host result of tool call back to open ai', () => {
...
+                  body: 'Alice is a software engineer at Google.',
```
jurgenwerk, line 1443: I'm replacing lorem ipsums with something more reasonable, it's better for debugging test failures

packages/ai-bot/helpers.ts
```diff
@@ -321,6 +321,114 @@ export function getRelevantCards(
...
+          if (!response.ok) {
+            throw new Error(`HTTP error! status: ${response.status}`);
+          }
+          let content: string;
+
+          if (attachedFile.contentType?.startsWith('text/')) {
+            content = await response.text();
+          } else {
+            const buffer = await response.arrayBuffer();
+            content = Buffer.from(buffer).toString('base64');
```
jurgenwerk: Now that I'm reviewing this I see that I didn't actually test this case. I'll do that next. Need to test what happens in the LLM response if I attach an image for example.

jurgenwerk: I changed this to only support text files (code) for now. It looks like this is what we should be focussing on now (given Chris's workflow that he posted today). https://github.com/cardstack/boxel/pull/2137/commits/c0ecf7f8040303c3270a94f10da6b9e97d71e3bf

packages/ai-bot/helpers.ts
```diff
@@ -321,6 +321,113 @@ export function getRelevantCards(
...
+  if (
+    !mostRecentUserMessageContent ||
+    mostRecentUserMessageContent.msgtype !== APP_BOXEL_MESSAGE_MSGTYPE
+  ) {
+    return [];
+  }
+
+  // We are only interested in downloading the most recently attached files -
+  // downloading older ones is not needed since the prompt that is being constructed
+  // should operate on fresh data
```
lukemelia, line 356: The way we handle cards is to only attach the most recent version of any cards attached to any message in the room. Should we take the same approach for files?

lukemelia (approved):
Besides my question, I think this looks fine.

packages/ai-bot/helpers.ts
jurgenwerk: My reasoning behind keeping the previous versions attached in old messages is to have the ability to see the old file content, or revert to it if needed. Since the LLM will be dealing with much more complicated stuff (gts code) compared to cards data, the hallucination factor and thus the possibility breaking the file increases quite a bit, so I thought it would be good to have some means of versioning here. On the other hand, this does differ from the cards behaviour as you described, so there is a bit of a discrepancy. For the reasons described I'm leaning towards keeping it this way, but I am eager to hear your opinion as well!

jurgenwerk:
@lukemelia I'll go ahead and merge this since new work depends on this, but regarding your question we can discuss later and see if we'd like to change that behavior.

PR #2138, opened 2025-02-12
ci: Add namespace for realm server log artifacts

backspace:
The main workflow has a lot of artifacts attached, this should help clarify what jobs produced them. Before:

boxel@88164a5 2025-02-12 12-13-08

After:

boxel@881d5d1 2025-02-12 12-35-05


jurgenwerk (approved):
Nice

PR #2139, opened 2025-02-12
Add code copy button to AI assistant

habdelra:
We use a modifier to add a copy button into the AI assistant when code blocks exist.

The API for the clipboard has security requirements prevent the ability to use the clipboard API when chrome is run in headless mode (when it is not run in headless mode there are also other silly constraints that make automated testing awkward). so our tests just assert that the button appears.

copy-button

### DIFF
--- a/packages/ai-bot/helpers.ts
+++ b/packages/ai-bot/helpers.ts
@@ -42,7 +42,8 @@ You may make multiple function calls, all calls are gated by the user so multipl
 If a user asks you about things in the world, use your existing knowledge to help them. Only if necessary, add a *small* caveat at the end of your message to explain that you do not have live external data. \
 \
 If you need access to the cards the user can see, you can ask them to attach the cards. \
-If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown.';
+If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown. \
+If you encounter code, please enclose the code within triple backticks and indicate the language after the opening backticks so that the code is displayed stylishly in Markdown. ';
 
 export const SKILL_INSTRUCTIONS_MESSAGE =
   '\nThe user has given you the following instructions. You must obey these instructions when responding to the user:\n';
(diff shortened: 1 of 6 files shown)


backspace:
I’m not getting a copy button locally, I’ll try again tomorrow 🤔

It really seems sad to not be able to somehow attach a component via a modifier!

jurgenwerk (approved):
I tested this in the PR preview and it works great. Just one note, in other apps I think an established pattern is to change text from "Copy to clipboard" to "Copied" after clicking on the button (the "Copied" text then changes back to the original text after a couple of seconds) so that you can get some feedback on your click.

example:

https://github.com/user-attachments/assets/4a9d9b8d-a917-4093-913d-67ddbd6940be

packages/ai-bot/helpers.ts
```diff
@@ -42,7 +42,8 @@ You may make multiple function calls, all calls are gated by the user so multipl
 If a user asks you about things in the world, use your existing knowledge to help them. Only if necessary, add a *small* caveat at the end of your message to explain that you do not have live external data. \
 \
 If you need access to the cards the user can see, you can ask them to attach the cards. \
-If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown.';
+If you encounter JSON structures, please enclose them within backticks to ensure they are displayed stylishly in Markdown. \
+If you encounter code, please enclose the code within triple backticks and indicate the language after the opening backticks so that the code is displayed stylishly in Markdown.';
```
jurgenwerk: In your screenshot, I think I saw the LLM responded with code that has 4 spaces for tab indentation, should we additionally ask the LLM to format it with 2 spaces so that the code looks more compact?

habdelra: Good idea

habdelra:
> I tested this in the PR preview and it works great. Just one note, in other apps I think an established pattern is to change text from "Copy to clipboard" to "Copied" after clicking on the button (the "Copied" text then changes back to the original text after a couple of seconds) so that you can get some feedback on your click.
> 
> example:
> 
>  Screen.Recording.2025-02-13.at.08.32.41.mov

Sure. We should do this for the command output copy button too

lukemelia (approved):
Agreed with suggested changes from Matic, otherwise looks good

habdelra:
Updated to use 2 space indent for code as well as to show the "copied" message after copying (and updated the way the command copy works so that it uses the same behavior).

copy_2

PR #2141, opened 2025-02-13
CS-7926 link crm cards to crm app 

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7926/link-all-card-to-the-crm-app-card

What is changing
- link all CRM cards (deals, accounts, company) to CRM app
- compared to https://github.com/cardstack/boxel/pull/2131 , `linksTo` thunk type helps to resolve the issue when loading cards and we can keep our models stored by file separately

**Issue Found**
- There is an error when we try to create a card on Contact/Account. It is suspected due to an interdependency card. **FIXED: It's related to the import cycle**


habdelra:
I think the error is the result of an import cycle. make sure that you don't have a module A that imports module B, and in module B you import module A (the cycle may also be mulitple steps removed, like module A -> module B -> module C -> module A, etc.)

richardhjtan:
@habdelra let's take this CRM use case for example:

```
export class CrmApp extends CardDef {
  static displayName = 'CRM App';
  static prefersWideFormat = true;
  static headerColor = '#4D3FE8';
  static isolated = CrmAppTemplate;
}


export class Company extends CardDef {
   static displayName = 'Company';
   @field crmApp = linksTo(() => CrmApp); 
   ...
}

export class Account extends CardDef {
  static displayName = 'CRM Account';
  @field crmApp = linksTo(() => CrmApp); 
  @field company = linksTo(() => Company);
  @field primaryContact = linksTo(() =>Contact);
  @field contacts = linksToMany(() => Contact);
  ...
}

export class Contact extends CardDef {
  static displayName = 'CRM Contact'; 
  ...
  @field company = linksTo(Company);
  ...
}
```
***currently these models are stored in separate files

We have Company, Contact and Account referenced to CrmApp. This is to make sure we only load them which are relevant to this specific CrmApp only.

while Account referenced to CrmApp, and also Contact (and Contact referenced CrmApp)
^^^ I think this makes the cycle happened

--- 

Is there a better approach for handling this business use case? One solution I tried was consolidating them into a single file, which resolved the issue. However, this sacrifices the cleanliness of storing the models in separate files.

I've been wondering if the way of `createCard` read cards differently compared to how cards load in code/interact mode. The cards load well despite the circular dependency, but when we do `createCard` it encounters an error.

habdelra:
> @habdelra let's take this CRM use case for example:
> 
> ```
> export class CrmApp extends CardDef {
>   static displayName = 'CRM App';
>   static prefersWideFormat = true;
>   static headerColor = '#4D3FE8';
>   static isolated = CrmAppTemplate;
> }
> 
> 
> export class Company extends CardDef {
>    static displayName = 'Company';
>    @field crmApp = linksTo(() => CrmApp); 
>    ...
> }
> 
> export class Account extends CardDef {
>   static displayName = 'CRM Account';
>   @field crmApp = linksTo(() => CrmApp); 
>   @field company = linksTo(() => Company);
>   @field primaryContact = linksTo(() =>Contact);
>   @field contacts = linksToMany(() => Contact);
>   ...
> }
> 
> export class Contact extends CardDef {
>   static displayName = 'CRM Contact'; 
>   ...
>   @field company = linksTo(Company);
>   ...
> }
> ```
> 
> ***currently these models are stored in separate files
> 
> We have Company, Contact and Account referenced to CrmApp. This is to make sure we only load them which are relevant to this specific CrmApp only.
> 
> while Account referenced to CrmApp, and also Contact (and Contact referenced CrmApp) ^^^ I think this makes the cycle happened
> 
> Is there a better approach for handling this business use case? One solution I tried was consolidating them into a single file, which resolved the issue. However, this sacrifices the cleanliness of storing the models in separate files.
> 
> I've been wondering if the way of `createCard` read cards differently compared to how cards load in code/interact mode. The cards load well despite the circular dependency, but when we do `createCard` it encounters an error.

I don't think that's the cycle. you simply need to follow the `import` statements to see the cycle (if there is one)....

lucaslyl (approved):

PR #2147, opened 2025-02-14
CS-7991 Sort crm edit fields

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7991/sorting-of-all-crm-edit-fields-with-custom-edit-template

What is changing
- sort edit fields within the crm cards with custom edit template
- this can avoid those computed value to be displayed and causes confusion that is editable


tintinthong (approved):
There is some lint error. Otherwise looks good

PR #2148, opened 2025-02-14
Fix: LLM selector disappears when more than 1 attached item is shown

FadhlanR:
Before:
Screenshot 2025-02-14 at 14 08 34

After:
Screenshot 2025-02-14 at 14 07 03

### DIFF
--- a/packages/host/app/components/ai-assistant/llm-select.gts
+++ b/packages/host/app/components/ai-assistant/llm-select.gts
@@ -42,6 +42,8 @@ const LLMSelect: TemplateOnlyComponent<Signature> = <template>
       border: none;
       padding: 0;
       width: auto;
+      min-width: 80px;
+      height: 43.5px;
       flex: 1;
       padding-right: var(--boxel-sp-xxs);
       cursor: pointer;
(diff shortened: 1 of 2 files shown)


tintinthong (approved):

packages/host/app/components/ai-assistant/llm-select.gts
```diff
@@ -42,6 +42,8 @@ const LLMSelect: TemplateOnlyComponent<Signature> = <template>
       border: none;
       padding: 0;
       width: auto;
+      min-width: 80px;
```
jurgenwerk: I think we can consider giving the LLM selector more space to breathe. 

With your current code:

image

Enlarging min-width to 120px:
image

packages/host/app/components/ai-assistant/llm-select.gts
```diff
@@ -42,6 +42,8 @@ const LLMSelect: TemplateOnlyComponent<Signature> = <template>
...
+      height: 43.5px;
```
jurgenwerk, line 45: Is this value (43.5) calculated from something?

jurgenwerk (approved):

packages/host/app/components/ai-assistant/llm-select.gts
FadhlanR: It's calculated from the item pill height

PR #2152, opened 2025-02-17
CS-7891 Introduce sorting parameter at kanban resources helper

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-7891/kanban-resource-resources-api-to-deal-with-sorting

### What is changing
- add new optional `orderBy` function parameter so we can freely define sorting mechanism by dnd column
- crm task planner to consume sort by and sort order from crm app level
- implement sort option value from UI to crm task planner

**Screenshot**

https://github.com/user-attachments/assets/124e0424-ea15-437f-bcce-87e6c6bedc67

**Examples**
| Order by | Sort By | Result |
| ------------- | ------------- | ------------- |
| Due Date | Asc  |  duedate-asc |
| Due Date | Desc  | duedate-desc |
| Priority | Asc |  priority-asc |
| Priority | Desc | prority-desc |

### DIFF
--- a/packages/experiments-realm/crm-app.gts
+++ b/packages/experiments-realm/crm-app.gts
@@ -551,6 +551,8 @@ class CrmAppTemplate extends Component<typeof CrmApp> {
             @viewCard={{this.viewCard}}
             @searchFilter={{this.searchFilter}}
             @taskFilter={{this.taskFilter}}
+            @sortBy='priority'
+            @sortOrder='desc'
           />
         {{else if this.query}}
           {{#if (eq this.selectedView 'card')}}
--- a/packages/experiments-realm/crm-app.gts
+++ b/packages/experiments-realm/crm-app.gts
@@ -48,7 +48,7 @@ import { DEAL_STATUS_VALUES } from './crm/deal-status';
 import DealSummary from './crm/deal-summary';
 import { CRMTaskPlanner } from './crm/task-planner';
 import type { LooseSingleCardDocument, Sort } from '@cardstack/runtime-common';
-
+import type { TaskSortBy, TaskSortOrder } from './crm/task-planner';
 import type { TemplateOnlyComponent } from '@ember/component/template-only';
 import {
   Card as CardIcon,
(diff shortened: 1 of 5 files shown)


packages/experiments-realm/crm-app.gts
```diff
@@ -444,6 +470,17 @@ class CrmAppTemplate extends Component<typeof CrmApp> {
     return taskFilter;
   }
 
+  get taskSortBy() {
+    return this.selectedSort?.sort?.[0]?.by as
+      | 'dueDate'
+      | 'priority'
```
tintinthong: can make a union type for this like SortType

packages/experiments-realm/crm-app.gts
```diff
@@ -139,12 +155,22 @@ const TASK_FILTERS: LayoutFilter[] = [
...
         id: 'priorityDesc',
         displayName: 'Priority',
-        sort: sortByPriorityDesc,
+        sort: sortByPriority('desc'),
+      },
+      {
+        id: 'priorityAsc',
+        displayName: 'Priority',
+        sort: sortByPriority('asc'),
       },
```
tintinthong, line 174: Are the displayName meaningful here?

packages/experiments-realm/crm/task-planner.gts
```diff
@@ -16,10 +17,63 @@ interface CRMTaskPlannerArgs {
     viewCard: () => void;
     searchFilter?: Filter[];
     taskFilter?: Filter[];
+    sortBy?: 'dueDate' | 'priority';
+    sortOrder?: 'asc' | 'desc';
```
tintinthong: I think its meaningful to have this in a single argument since both pertain to sort

tintinthong: I dono why u split these args up when they are selected together anyway

packages/experiments-realm/crm-app.gts
richardhjtan: @tintinthong Yes `displayName` meant for these labels Screenshot 2025-02-18 at 2 49 28 PM

tintinthong (approved):
Looks good in general. Just minor comments

packages/experiments-realm/crm/task-planner.gts
richardhjtan: yeah agree, these shall be together as related to sorting - merged it into single argument

PR #2153, opened 2025-02-17
Show the attached file in message bubble

FadhlanR:
Screenshot 2025-02-17 at 16 13 34


lukemelia (approved):
I'm not sure how practical it is, but since FileDef and CardDef have a common base class, it would be nice to use polymorphism as much as possible to avoid having to repeat patterns for each, like CardPill/FilePill.

FadhlanR:
> I'm not sure how practical it is, but since FileDef and CardDef have a common base class, it would be nice to use polymorphism as much as possible to avoid having to repeat patterns for each, like CardPill/FilePill.

We decided to have a separate component for this because we'll be pulling data from different fields, and we want to avoid too many conditions. @jurgenwerk and I discussed it here: https://github.com/cardstack/boxel/pull/2126#discussion_r1950684602.

packages/host/app/components/ai-assistant/message/index.gts
```diff
@@ -215,6 +218,14 @@ export default class AiAssistantMessage extends Component<Signature> {
             </div>
           {{/if}}
 
+          {{#if @resources.files.length}}
+            <div class='cards' data-test-message-files>
```
jurgenwerk: Class name cards implies there will be cards here, but actually there are files. Would be good to rename the class.

jurgenwerk (approved):
Looks good just a minor comment. Can you also show a screenshot of how it looks like when there are both cards and files present?

FadhlanR:
> Looks good just a minor comment. Can you also show a screenshot of how it looks like when there are both cards and files present?

@jurgenwerk here's the screenshot:

Screenshot 2025-02-18 at 19 06 39

jurgenwerk:
The vertical and horizontal padding between pills is different which can be improved. And this looks different compared to how this collection is shown in the attachment picker. The pills are displayed inline there, regardless of its type. I would expect it to look the same here.

burieberry (commented):
Agree with @jurgenwerk about the visual treatment, otherwise looks good

PR #2154, opened 2025-02-17
code ref should serialize to relative

tintinthong:
Specs that get created now will serialize the code ref fields into relative urls. Similarly, if you had an absolute url, an auto-save will serialise it relative. These happen on the server. This changes will apply to any card with code ref field.

**Before**

Screenshot 2025-02-19 at 23 29 29

**After**

Screenshot 2025-02-19 at 23 29 37


packages/base/card-api.gts
```diff
@@ -2294,8 +2299,9 @@ async function getDeserializedValue<CardT extends BaseDefConstructor>({
 export interface SerializeOpts {
   includeComputeds?: boolean;
   includeUnrenderedFields?: boolean;
-  maybeRelativeURL?: ((possibleURL: string) => string) | null; // setting this to null will force all URL's to be absolute
+  useAbsoluteURL?: boolean;
```
habdelra, line 2307: thanks!

habdelra (approved):

PR #2155, opened 2025-02-17
Remember playground format via local storage

burieberry:
- in playground panel, the selected card preview format is now remembered


lukemelia (approved):

jurgenwerk:
Related question, I am using playground in another branch (https://github.com/cardstack/boxel/pull/2137) and after refreshing the playground doesn't stay open, I think this would be really helpful to persist too (in case it's not already implemented–I'm not sure if my branch is outdated)

burieberry:
> Related question, I am using playground in another branch (#2137) and after refreshing the playground doesn't stay open, I think this would be really helpful to have too (in case it's not already implemented–I'm not sure if my branch is outdated)

Merged #2157 just recently, see if that resolves this. Please report on any issues/ideas for improvement.

jurgenwerk:
@burieberry perfect, thank you.

PR #2156, opened 2025-02-17
Better handling for contained computed that consumes linked field

habdelra:
This PR includes better handling for a contained computed that consumes a linked field. This issue started to appear after we forced all contained fields to be rendered regardless if they were used in a template. As part of this we started to see contained computed's that consume linked fields throw errors in our indexer. In production instances that contained these types of fields actually triggered an indexing cycle. It looks like the API machinery can handle these fields just fine, so long as we just get out of the way. So the update here was to stop throwing when we encounter this, and just let the field temporarily be undefined while the API tries to resolve the field value.

Additionally, our cycle detection logic in our renderer has a threshold value that is set absurdly high (it was set back before HTTP GET's were used to visit cards in the indexer--before workers even). I lowered this threshold to something more reasonable so that cycles could more quickly be detected in our indexer. I also added some better debugging messages what will allow us to more easily spot any cycles in the future.

I'm not 100% positive this solves the cycle issue in prod, but at the very least it should greatly mitigate it and provide better forensics for finding and fixing indexing cycles in the future.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -57,11 +57,11 @@ import CaptionsIcon from '@cardstack/boxel-icons/captions';
 import RectangleEllipsisIcon from '@cardstack/boxel-icons/rectangle-ellipsis';
 import LetterCaseIcon from '@cardstack/boxel-icons/letter-case';
 
-interface CardorFieldTypeIconSignature {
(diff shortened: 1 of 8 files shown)


lukemelia (approved):
Nice spelling and capitalization improvements along the way 🐝

PR #2162, opened 2025-02-18
Can create new instance in playground

burieberry:
- separated next step into different pr to prevent blocker on turning feature flag off for playground: #2180 convert create instance to command


packages/host/app/components/operator-mode/code-submode/playground-panel.gts
```diff
@@ -488,6 +507,32 @@ class PlaygroundPanelContent extends Component<PlaygroundContentSignature> {
...
+    let doc: LooseSingleCardDocument = {
+      data: {
+        meta: {
+          adoptsFrom: ref,
+          realmURL,
+        },
+      },
+    };
+    try {
+      let card = await this.cardService.createFromSerialized(doc.data, doc);
```
habdelra: please use CardResource to create new cards, not the CardService directly. failing to do so will result in an instance that doesn't consistently rerender for live updates

habdelra (changes requested):
New instances should always be created via CardResource. Here is an example of what that looks like https://github.com/cardstack/boxel/blob/main/packages/host/app/lib/stack-item.ts#L55-L58

packages/host/app/components/operator-mode/code-submode/playground-panel.gts
burieberry: thanks, will do

burieberry: Updated this to use card-resource. Had some trouble persisting the new card and adding it to recent-cards, so you'll see the note I made in createNew task.

habdelra (approved):
Looks good!

PR #2163, opened 2025-02-19
Having an issue where PrerenderedCardSearch is not showing query data inside embedded/other formats

lucaslyl:
As per today’s sprint planning, we discussed the issue of displaying nested query data in an embedded format within the CRM App.

Issue:
**In the Account.gts Embedded Format, everything appears to be working correctly:**
Screenshot 2025-02-19 at 14 51 06

**However, in the CRM App, the nested query data is not displaying:**
Screenshot 2025-02-19 at 14 37 59

---------------------------

To address this issue,I created a **super-project-app** that replicates the same problem to make it easier to visualize the issue . I have outlined two methods:

- Using the original approach, PrerenderedCardSearch, which has the issue of not displaying the query data correctly.
- Using the getCards API, which temporarily resolves the issue.

Both methods have their pros and cons, which are detailed below:

https://github.com/user-attachments/assets/d2259c7d-60e2-48cb-ad39-f428d5596e75

### DIFF
--- a/packages/experiments-realm/SuperProjectAccount/1d5b6e71-067c-44bf-982e-d41285f5d53f.json
+++ b/packages/experiments-realm/SuperProjectAccount/1d5b6e71-067c-44bf-982e-d41285f5d53f.json
@@ -1,0 +1,23 @@
+{
+  "data": {
+    "type": "card",
+    "attributes": {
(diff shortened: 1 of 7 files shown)


lucaslyl:
@habdelra , Perhaps could get your assistance to review this issue

PR #2164, opened 2025-02-19
remove content-scroll-container from content-grid & add scroll to content

lucaslyl:
linear:https://linear.app/cardstack/issue/ECO-165/prevent-the-extra-scroll-in-crm-app-container

Result:

https://github.com/user-attachments/assets/6c06e042-151d-4b55-a4d9-16779b75c4f0


tintinthong (approved):
Works as  how I expected

richardhjtan (approved):
Looks good!

PR #2165, opened 2025-02-19
Playground instance previewed in edit view should save changes

FadhlanR:
- [x] Refactored the auto-save handler by moving it to the store service, based on @habdelra's suggestion.  
- [x] Made `cardResource` in the playground panel an auto-saving card.

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -203,7 +203,9 @@ export default class StoreService extends Service {
           onCardChange: resource.isAutoSave
             ? () => {
                 if (card) {
-                  this.initiateAutoSaveTask.perform(card);
+                  // Using the card ID instead, so this function doesn't need to be updated
+                  // when the card instance changes.
+                  this.initiateAutoSaveTask.perform(card.id);
                 }
               }
             : undefined,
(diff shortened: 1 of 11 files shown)


packages/host/app/services/store.ts
```diff
@@ -139,7 +170,7 @@ export default class StoreService extends Service {
 
     let url = cardOrError.id;
     if (!url || !resource.isLive) {
-      this.handleUpdatedCard(undefined, cardOrError);
+      await this.handleUpdatedCard(undefined, cardOrError);
```
habdelra, line 173: nice catch!

packages/host/app/services/store.ts
```diff
@@ -272,9 +316,27 @@ export default class StoreService extends Service {
     }
 
     if (maybeUpdatedCard?.id) {
+      if (!this.cardApiCache) {
+        this.cardApiCache = await this.cardService.getAPI();
+      }
```
habdelra, line 323: no need to cache this--it's actually cached in the loader, you can just get it each time that you need it.

habdelra (approved):
this looks really good! exactly like we discussed

tintinthong (approved):
I have tested this and I dont see any degression in auto save. LGTM

packages/host/app/services/store.ts
FadhlanR: It's cached so that it can be reused in the unloadResource function. Otherwise, unloadResource would have to be an async function, but we need it to be synchronous for use in registerDestructor. Do you have any other idea how to get cardAPI in unreloadResource?

habdelra: ah! I see! make sense let's leave it as is

PR #2166, opened 2025-02-19
Make case consistent for CreateAiAssistantRoomCommmand

lukemelia:


IanCal (approved):
Lgtm, this will cause issues with the example card we have for doing things with AI rooms @tintinthong it's a simple fix but this will likely impact code the students are working on

PR #2172, opened 2025-02-19
Add codeblock rendering support using monaco for AI messages

habdelra:
This PR adds codeblock rendering support using monaco for AI messages. As part of this there was a lot of cleanup that i had to do for `RoomResource` and `RoomMessageComponent`. Specifically there were a couple problems:

1. The main problem is that every `RoomMessageComponent` on the screen was re-rendering for every matrix event received, regardless of whether the matrix event had anything to do with the `RoomMessageComponent`. This was due to the fact that the `RoomMessageComponent` use an array of `Message` objects as one of it's args. The thing is that for every access of `RoomResource.message` the array was rebuilt (it was not cached), moreover, even when it is cached, it's still creating a brand new array for each new message received in a room. We need to stabilize our `RoomMessageComponent` on the `RoomResource` so that it is not rebuilding on every single matrix event.
2. The `RoomResource` has caches that span rooms. This is fine--as you jump between your rooms it's nice to use cached values. However, the cache was blown away every time you switched rooms, and the way this was done was a bit dubious. The better solution here is to retain the caches of items that span rooms and in the various getters filter out the pertinent items from the rooms, and moreover to use a `@cached` decorator to prevent the filter from running on every consumption of the property, and only recomputed when the room changes.

I believe that our message scrolling logic is actually dependent on the `RoomMessageComponent` re-render bug. so very likely we'll need to fix this _again_. (/cc @FadhlanR)

codeblock


packages/host/app/resources/room.ts
```diff
@@ -83,28 +83,12 @@ export class RoomResource extends Resource<Args> {
   @service declare private cardService: CardService;
 
   modify(_positional: never[], named: Args['named']) {
-    if (named.roomId) {
-      if (this.isNewRoom(named.roomId)) {
-        this.resetCache();
-      }
```
habdelra, line 89: this is not a good strategy. it defeats the purpose of the cache and its a really awkward way to constrain the values returned from this reource to only those that pertain to this room. A better approach is to keep this cache, but then filter out only items that pertain to this room in our various getters, and to make sure those getters use a `@cached` decorator

packages/host/app/resources/room.ts
```diff
@@ -120,32 +104,63 @@ export class RoomResource extends Resource<Args> {
...
 
+  @cached
   get joinedMembers() {
-    return this.members.filter((m) => m.membership === 'join');
+    if (this.roomId == undefined) {
+      return [];
+    }
+    return this.members.filter(
+      (m) => m.membership === 'join' && m.roomId === this.roomId,
+    );
```
habdelra, line 150: note that these caches span rooms. that is fine, it makes for quick access when switching between rooms, but it means that we should filter these values by room and use a `@cached` decorator so we only do this filtering when the room switches--not for every access of the property

packages/runtime-common/marked-sync.ts
```diff
@@ -1,53 +1,34 @@
...
-          html.push(
-            `<pre class="language-${language}" data-codeblock>${code}</pre>`,
-          );
-          return html.join('\n');
+          let id = `${CODEBLOCK_KEY_PREFIX}${uuidv4()}`;
+          // we pass the code thru using localstorage instead of in the DOM,
+          // that way we don't have to worry about escaping code. note that the
+          // DOM wants to render "<template>" strings when we put them in the
+          // DOM even when wrapped by <pre>. also consider a codeblock that has
+          // a "</pre>" string in it.
```
habdelra, line 20: this strategy actually worked out really well. so our AI bot's messages can have any manner of code within them., however we are not using this approach for messages that originate from the user, and those do have code escaping problems. like if a user tries to type `<template>` in a message, that message will be rendered without the string`<template>`. you can use your imagination for other things that might be problematic. I suspect that the HTML sanitization may be getting in our way here. users should be able to ask questions about code. filing this as a bug

packages/host/app/modifiers/code-block.ts
```diff
@@ -0,0 +1,172 @@
...
+        // producing spurious visual diffs
+        if (isTesting()) {
+          editorOptions.wordWrap = 'on';
+        }
+        let monacoContainer = document.createElement('div');
+        monacoContainer.setAttribute(
+          'class',
+          'preview-code monaco-container code-block',
+        );
+        monacoContainer.setAttribute('style', `height: ${lines + 4}rem`);
```
habdelra, line 108: This is a simple heuristic to get the codeblock to have a variable height based on the content. The monaco div itself does not automaticlaly grow based on the contents within it. (note that there is a min-height/max-height CSS prop on this to make sure this is well behaved)

packages/runtime-common/marked-sync.ts
habdelra: filed as CS-8030

packages/host/app/modifiers/code-block.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    for (let [index, codeBlockNode] of [...codeBlocks].entries()) {
+      let codeBlock = codeBlockNode as HTMLElement;
+      let id = codeBlock.id;
+      if (!id) {
+        continue;
+      }
+      // note that since the localstorage item was set from runtime-common there
+      // was no ember-window-mock available, so we can't use window mock here
+      let maybeContent = window.localStorage.getItem(id);
```
backspace, line 53: It’s worrisome to me that [this custom lint rule](https://github.com/cardstack/boxel/blob/3b2b6f8dcd5febe78c570be47e434c1836393c37/packages/eslint-plugin-window-mock/mock-window-only.js) isn’t triggering 😳

packages/host/tests/integration/components/ai-assistant-panel-test.gts
```diff
@@ -2664,6 +2664,25 @@ module('Integration | ai-assistant-panel', function (hooks) {
     );
 
     await waitFor('[data-test-message-idx="0"]');
+    assert
+      .dom('button.code-copy-button')
+      .exists('the copy code to clipboard button exists');
+
+    // assert that new messages don't destabilize the RoomMessage component
```
backspace, line 2671: What actually asserts this?

backspace (approved):
IIRC this is not the first time constant rerendering of room messages has been uncovered, I can’t think of how, but is there some way to detect this wastefulness? 😳

packages/host/tests/integration/components/ai-assistant-panel-test.gts
habdelra: Adding a new event would previously destabilize the RoomMessage component (as described in the PR description). The copy button and Monaco editors are from a modifier that adds DOM. A rerender of this component would blow away the DOM added by their modifier. Meaning that the assertions below this added event would fail.

habdelra:
> IIRC this is not the first time constant rerendering of room messages has been uncovered, I can’t think of how, but is there some way to detect this wastefulness? 😳

Yeah I added assertions that will fail (you commented on it actually) if this starts happening again. The CodeBlockModifier adds DOM to the RoomMessage. If it rerenders that added DOM from the modifier will get blown away.

backspace:
> Yeah I added assertions that will fail (you commented on it actually) if this starts happening again. The CodeBlockModifier adds DOM to the RoomMessage. If it rerenders that added DOM from the modifier will get blown away.

ah great, I wasn’t grasping that, thanks for the explanation 🤩

jurgenwerk (approved):
Nice optimization, and the feature works great, I tested it locally:

image

jurgenwerk:
I noticed a visual annoyance when the assistant is streaming file contents to the user. When the code is streamed, it looks like the code editor is continuously rerendering which produces flashing (check the video below). Probably this is out of scope for this PR and we should probably make a separate ticket: 

https://github.com/user-attachments/assets/6ea90989-c789-4d44-b0e7-958421cd30de

habdelra:
> I noticed a visual annoyance when the assistant is streaming file contents to the user. When the code is streamed, it looks like the code editor is continuously rerendering which produces flashing (check the video below). Probably this is out of scope for this PR and we should probably make a separate ticket:
> 
>  code-flashing.mov

i think the alternative would be to show it in a `<pre>` tag until it stops streaming, and once it stops streaming render it in monaco. I think actually there is a natural place to do that in this PR

habdelra:
@jurgenwerk I updated the PR in the latest commit so that we don't render the code block while the ai bot response is streaming.

lukemelia (approved):

PR #2173, opened 2025-02-19
Don't create a new RoomResource for the Room component, use the one from MatrixService

lukemelia:


backspace (approved):
Seeking: identity map 🙃

PR #2175, opened 2025-02-20
update summary card data to match zeplin design

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-171/company-info-should-change-to-deal-info-in-crm-deal-isolated

Before:
image

Zeplin design:
Screenshot 2025-02-20 at 15 05 57

Result:
I slightly changed and replaced the **icon** on the Deal Info Summary Card to make it more synced with the related data.

https://github.com/user-attachments/assets/22d4b2e5-255f-4e28-a5c9-048f237e0321

### DIFF
--- a/packages/experiments-realm/crm/deal.gts
+++ b/packages/experiments-realm/crm/deal.gts
@@ -48,6 +48,9 @@ import { DealEvent } from './deal-event';
 import { DealStatus } from './deal-status';
 import { DealPriority } from './deal-priority';
 import { CrmApp } from '../crm-app';
+import MapPin from '@cardstack/boxel-icons/map-pin';
+import Calendar from '@cardstack/boxel-icons/calendar';
+import UsersGroup from '@cardstack/boxel-icons/users-group';
 
 interface DealSizeSummary {
   summary: string;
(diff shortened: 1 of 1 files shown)


packages/experiments-realm/crm/deal.gts
```diff
@@ -117,19 +123,36 @@ class EditTemplate extends Component<typeof Deal> {
 class IsolatedTemplate extends Component<typeof Deal> {
   get logoURL() {
     //We default to account thumbnail
-    return (
-      this.args.model.thumbnailURL ?? this.args.model.account?.thumbnailURL
-    );
+    return this.args.model?.thumbnailURL?.length
+      ? this.args.model.thumbnailURL
+      : this.args.model.account?.thumbnailURL;
```
lucaslyl, line 128: This approach ensures that empty strings are treated as invalid and fall back to the account's thumbnailURL.

tintinthong:
There is an issue that @fields.linkedCard.propertyName doesn't work. https://linear.app/cardstack/issue/CS-7670/delegate-to-nested-field. This is why @lucaslyl uses the @model. I can look into this separately

tintinthong (approved):

packages/experiments-realm/crm/deal.gts
```diff
@@ -753,22 +788,16 @@ class IsolatedTemplate extends Component<typeof Deal> {
 
 class FittedTemplate extends Component<typeof Deal> {
   get logoURL() {
-    return (
-      this.args.model.thumbnailURL ?? this.args.model.account?.thumbnailURL
-    );
+    //We default to account thumbnail
```
richardhjtan, line 791: Not sure if the comment is necessary or perhaps 
// fallback to account thumbnail if deal has no thumbnail

richardhjtan (approved):
LGTM

PR #2177, opened 2025-02-20
Allow to search null field on search resources / improve parse query on null handling

richardhjtan:
We have a scenario to get CRM task cards with no assignee when user clicks 'Unassigned' filter

Two layer of issues found with @tintinthong 
- currently, when we passing `null` value on query, it parses as empty string - so search doc couldn't find any cards of field which has `null` value 
- at formatQuery card-api layer, we can't query by `assignee: null` as it gives error about compound field ( _**NOTE**: found this is not valid after fix the first issue and did several tests locally_ )

**Solution**
- able to parse and stringify search query url with null, by using qs feature handling of null values `strictNullHandling` 

**Examples**

**Search doc of CRM task card with no assignee**
crm-task-card-with-no-assignee

**Error when filtering assignee null before the fix**
Screenshot 2025-02-21 at 12 07 20 AM
Screenshot 2025-02-21 at 12 07 14 AM

**After fix**
Screenshot 2025-02-21 at 1 21 04 AM

### DIFF
--- a/packages/host/tests/integration/realm-indexing-and-querying-test.gts
+++ b/packages/host/tests/integration/realm-indexing-and-querying-test.gts
@@ -4391,6 +4391,22 @@ posts/ignore-me.json
       );
     });
 
+    test(`can use 'eq' to find empty linksTo field`, async function (assert) {
+      let { data: matching } = await queryEngine.search({
+        filter: {
+          on: {
+            module: `${testModuleRealm}friend`,
+            name: 'Friend',
+          },
+          every: [{ eq: { firstName: 'Mango' } }, { eq: { friend: null } }],
+        },
+      });
+      assert.deepEqual(
+        matching.map((m) => m.id),
+        [`${testRealmURL}friend2`],
+      );
+    });
+
     test(`can search for cards by using a computed field`, async function (assert) {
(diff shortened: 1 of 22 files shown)


packages/host/tests/integration/realm-indexing-and-querying-test.gts
```diff
@@ -4391,6 +4391,22 @@ posts/ignore-me.json
...
     });
 
+    test(`can use 'eq' to find empty linksTo field`, async function (assert) {
+      let { data: matching } = await queryEngine.search({
+        filter: {
+          on: {
+            module: `${testModuleRealm}friend`,
+            name: 'Friend',
+          },
+          every: [{ eq: { firstName: 'Mango' } }, { eq: { friend: null } }],
```
richardhjtan, line 4401: From what i test in the CRM app, both `{ eq: { assignee : null}}` or ` { eq: { 'assignee.id' : null}}` able to give the same expected result after the fix

burieberry (approved):

tintinthong (approved):
>  ( NOTE: found this is not valid after fix the first issue and did several tests locally )

Good find on this!

PR #2178, opened 2025-02-20
Allow field def containedExamples to render

tintinthong:
This PR builds on on #2189 and https://github.com/cardstack/boxel/pull/2105. 

Previously, Specs could not render fields

Screenshot 2025-02-25 at 11 02 49

Now, Specs render fields based upon the code ref it contains

Screenshot 2025-02-24 at 18 23 06

**Further Note**

Here, Specs are 1 to 1 with code refs so it doesn't really need a "different" subclass rendering of the field when showing containedExamples. However, if we want to include the UI affordance for selecting a template for field that is of a particular type, we can easily do so by dynamically specifying the typeConstraint (but this might not be the concern of Spec). 

I saw some rendering thing with the ColorField. I'll look into it independently


tintinthong:
Screenshot 2025-02-25 at 10 55 02

Currently, some issues with some fields. Color is another one

tintinthong:
Pause review here. There is some bug on primitive handling which requires more card-api tweaks

tintinthong:
> Pause review here. There is some bug on primitive handling which requires more card-api tweaks

The bug in here is when serializing (im not sure in deserializing) URL extends StringField. 

URL is a primitive so it doesn't serialize meta.fields. That means it doesn't have the ability to render a different template.

tintinthong:
I have updated https://github.com/cardstack/boxel/pull/2189. And will continue work once that is merged

tintinthong:
After some consideration, 

I think handling of primitive for polymorphism should be out-of-scope of this PR. 

It is quite involve to get this to work and might invite some workaround patterns. So it would be good to check-in with Ed in how to appraoch this

burieberry (approved):

PR #2180, opened 2025-02-20
Add "create card instance" command

burieberry:
Open for comments regarding usage of card-resource in a command in this way to create a card. If this is not a good idea, maybe this PR is no longer relevant.

### DIFF
--- a/packages/host/app/commands/create-card-instance.ts
+++ b/packages/host/app/commands/create-card-instance.ts
@@ -1,13 +1,17 @@
 import { type LooseSingleCardDocument } from '@cardstack/runtime-common';
 
+import { getCard } from '@cardstack/host/resources/card-resource';
+
+import type { CardDef } from 'https://cardstack.com/base/card-api';
 import type * as BaseCommandModule from 'https://cardstack.com/base/command';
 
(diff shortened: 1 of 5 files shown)


packages/host/app/commands/create-card-instance.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+    let doc: LooseSingleCardDocument = {
+      data: {
+        meta: {
+          adoptsFrom: input.module,
+          realmURL: input.realm,
+        },
+      },
+    };
+    await this.cardService.createFromSerialized(doc.data, doc);
```
habdelra: we should really be using the CardResource to make cards, not the CardService directly (it would be great if we could deprecate this so others don't fall into this trap). When you make a card with the CardService directly that means that the card will not be part of the identity map and will not receive live updates, as well as other realy bad side effects.

lukemelia (commented):
Result of team review: hold off on this PR, review with @ef4

burieberry:
This will be handled in https://linear.app/cardstack/issue/CS-7982/proposed-api-for-working-with-cards-cardresource-search-etc-that-will

PR #2181, opened 2025-02-21
add a href to website field

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7698/website-field-shud-not-need-to-https

Issue: The website field is not clickable because the atom format is overriding the URL field's atom format, which contains the href functionality for making links clickable.

Result:

https://github.com/user-attachments/assets/c72c5036-7c23-49a3-a15c-4d75bb7b45c8


tintinthong (approved):

richardhjtan (approved):
Looks good

PR #2183, opened 2025-02-21
Make text ellipsis in profile popover

FadhlanR:
Screenshot 2025-02-21 at 13 33 06


jurgenwerk (approved):
Not sure if there's a title attribute already present but I think it's a good practice to add it on elements with ellipsis overflow

habdelra (approved):

richardhjtan (approved):

PR #2184, opened 2025-02-21
CS-8016 promote CRM app to seed realm

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8016/manage-crm-app-deploy-to-seed-realm

### What is changing
- promote CRM app, related components and fields into seed realm 
- sync fields and component folder between experiment and seed realm

It has been pushed to staging by branch and it works as expected


backspace:
Does this need to wait for #2188?

tintinthong:
> Does this need to wait for #2188?

My opinion is that we shud merge what has been prepared in this PR here and work iteratively from there

richardhjtan:
> Does this need to wait for #2188?

@backspace we will address this separately but it should be immediately after this PR. Thanks for bringing up

PR #2186, opened 2025-02-21
CardService instance creation warning

habdelra:
Documenting our public service announcement to not create instances from CardService.


burieberry (approved):
Thanks, was thinking of doing the same

PR #2187, opened 2025-02-21
Fixes the bug where switching from module to an instance in code mode would show a loading indicator that never dismisses

habdelra:
Fixes this bug where switching from a module to an instance in code module would show a loading indicator that never dismisses.

Before:

loading-bug-before

After:

loading-bug-after2


tintinthong (approved):
Lgtm. thx thiz was annoying

lukemelia (approved):

PR #2188, opened 2025-02-21
CRM app more style fixes

burieberry:
@lucaslyl Your PR was merged before I could push these. I did some refactor on the SummaryCard component. There's more work to do. Please take over from here. 

TODO:
- [ ] Please please please organize the imports on `account.gts`
- [ ] Follow the header example in Isolated Deal page to fix the html and spacing for the Account Header component.
- [ ] Update the fitted views for Account, Deal, Contact cards. You can check in code-mode to see what looks off and work from there.
- [ ] Add any missing embedded views (Deal)

These will have to do for now because we don't have time. There are just so many duplications between Account and Deal files.


burieberry:
Closing this since the changes here are being addressed in a separate PR.

PR #860, opened 2023-11-28
Ensure that field components do not get recreated when state changes

jurgenwerk:
This is a result of pairing with @ef4 yesterday -

The issue we were solving is a bug where the "contains many" editor component will lose focus after inputting a single character, making it impossible to write a word in one go:

image

It turns out the whole editor component got replaced (i.e. we could see that  the id in `<div id="ember102"...` got replaced) in DOM after state changes. The fix is that we now cache the components by field name, using the same technique we already use in `getBoxComponent`. 

We were thinking about the possibilities about adding a test for this, but we don't actually know enough about the source of this problem. The described bug reproduction is one way to see the bug, but we assume that this is only one case among many, and that the glimmer engine is **usually** smart enough to not recreate components (but not in this case)...

PR #2189, opened 2025-02-24
Fix component cache for polymorphic field re-renders

tintinthong:
The point of this PR is to enable polymorphic fields to re-render upon assigning a subclass to a field of baseclass type. Previously, although the JSON is correctly serialised, the templates were not re-rendering

Screenshot 2025-02-24 at 16 03 18

Screenshot 2025-02-24 at 16 03 30

**Solution**

The main issue is that `getBoxComponent` was serving a stale cache (`componentCache`) bcos it did not take into account changes in the card or field class ie `cardOrField`. So we added a value inside the cache storing `cardOrField` which then serves the most up-to-date BoxComponent if `cardOrField` changed.

**Digression Risk**

~~As part of pairing on Thursday with @ef4, the outer cache at `fieldsComponentsFor` (`stableComponents`) was already serving a stale cache that was preventing the code from returning the most up-to-date cache in boxComponent. This several layers of caching made it difficult to debug. We initially thought that this outer-cache was not needed.~~

~~BUT afterwards I found out that the cache was made for the purpose preventing eager re-renders that caused `linksToManyComponent` and `containsManyComponent` to re-render -- so I kept the cache in place. Here is the relevant PR https://github.com/cardstack/boxel/pull/860 by @jurgenwerk. I eventually, decided to keep the cache for specific fields ie plural fields like `linksToManyComponent` and `containsManyComponent` for simplicity.. also to unblock Spec and Playground work.~~

~~I plan to pair further with @ef4 to find the best solution here.. I think the more general solution wud be to allow `getBoxComponent` to have better handling for plurals.~~


burieberry:
Is the template not rerendering when clicking the button still an issue? I tried on latest main and didn't have any issues.

tintinthong:
> Is the template not rerendering when clicking the button still an issue? I tried on latest main and didn't have any issues.

wdym. Where did u see that there was a re-render of the field? Particularly, when ur `somePersonField=contains(FieldDef)`

tintinthong:
> Is the template not rerendering when clicking the button still an issue? I tried on latest main and didn't have any issues.

I tried this test on main. I dont think it works

Screenshot 2025-02-27 at 02 46 23

burieberry:
> > Is the template not rerendering when clicking the button still an issue? I tried on latest main and didn't have any issues.
> 
> I tried this test on main. I dont think it works
> 

Sorry ok I was able to reproduce the issue. Previously didn't see it because I was switching between formats on code mode preview. Confirm that this PR resolves it.

tintinthong:
I think this is ready for review

The continuation of this PR. I have removed the fields proxy cache and moved the cache into the linksToManyComponent. containsMany and linksToMany examples should still have the stability they have had previously.

tintinthong:
Its great that the change in cache introduced inside getBoxComponent (that makes cardOrField as a dependency) here fixes the re-ordering template issue seen here https://github.com/cardstack/boxel/pull/2241. 

Essentially, it invalidates the componentCache of the box component but that also means any component state is destroyed (without auto-save).

ef4 (approved):
We think this is a reasonable step forward without taking on a big refactor. 

It's hard to be sure this doesn't change component stability somewhere else, but we haven't been able to find any examples in the tests. So we can move ahead with this and folks should be on the lookout for any component state loss regressions that we can debug.

PR #2195, opened 2025-02-24
realm can tolerate a null last_modified date in index when serving a module

habdelra:
it is useful to be able to set the last_modified date to null in the index in order to force a revisit of the instance during indexing (perhaps an indexer bug--not content--caused an error doc to be generated. In which case you'll want to force a revist of the instance regardless of whether or not it has changed.

Currently the realm is not at all tolerant of a null last_modified date and infact will actually not serve a module (like the card-api module) if the last modified date is null. there is no need for this level of strictness. let's relax this.


lukemelia (approved):

lukemelia (approved):

habdelra:
The percy tests are failing on a change to the user icon color. seems 100% unrelated to this PR. Unsure if I should approve those changes in percy...

backspace:
The [latest Percy build on `main`](https://percy.io/Cardstack-1/web/-cardstack-host/builds/38913986/unchanged/2103322433?browser=chrome&browser_ids=64%2C65&category=unchanged&viewLayout=overlay&viewMode=original&width=1280&widths=1280) has that same shade of green, so I merged `main` in here, hopefully the diffs disappear 🤞🏻

PR #2196, opened 2025-02-24
Don't treat non-URL CodeRef modules as URLs

lukemelia:
The issue I ran into is that a recent change assumed that a CodeRef's module would always be a URL. However, ours can be something like: @cardstack/boxel-host/commands/create-ai-assistant-room

This PR above accommodates this while retaining the relative URL behavior change.

### DIFF
--- a/packages/base/code-ref.gts
+++ b/packages/base/code-ref.gts
@@ -16,6 +16,14 @@ import {
 import { ResolvedCodeRef } from '@cardstack/runtime-common';
 import CodeIcon from '@cardstack/boxel-icons/code';
 
+function moduleIsUrlLike(module: string) {
+  return (
+    module.startsWith('http') ||
+    module.startsWith('.') ||
+    module.startsWith('/')
+  );
+}
+
 class BaseView extends Component<typeof CodeRefField> {
   <template>
     <div data-test-ref>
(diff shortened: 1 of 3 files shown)


packages/base/code-ref.gts
```diff
@@ -37,9 +37,11 @@ export default class CodeRefField extends FieldDef {
     _visited?: Set<string>,
     opts?: SerializeOpts,
   ) {
+    const moduleIsUrlLike =
+      codeRef.module.startsWith('http') || codeRef.module.startsWith('.');
```
habdelra: technically, starting with '/' would be "url like" as well

packages/base/code-ref.gts
```diff
@@ -70,12 +72,18 @@ function maybeSerializeCodeRef(
...
 ) {
   if (codeRef) {
-    // if a stack is passed in, use the containing card to resolve relative references
-    let moduleHref =
-      stack.length > 0
-        ? new URL(codeRef.module, stack[0][relativeTo]).href
-        : codeRef.module;
-    return `${moduleHref}/${codeRef.name}`;
+    const moduleIsUrlLike =
+      codeRef.module.startsWith('http') || codeRef.module.startsWith('.');
```
habdelra: ditto with above, starting with "/" is a valid "url like" path too

packages/base/skill-card.gts
```diff
@@ -1,11 +1,66 @@
...
 
+function djb2_xor(str: string) {
+  let len = str.length;
+  let h = 5381;
+
+  for (let i = 0; i < len; i++) {
+    h = (h * 33) ^ str.charCodeAt(i);
+  }
+  return (h >>> 0).toString(16);
+}
```
habdelra: we have a lib for this now. please use `simpleHash` from runtime-common

packages/base/skill-card.gts
```diff
@@ -1,11 +1,66 @@
...
+
+function friendlyModuleName(fullModuleUrl: string) {
+  return fullModuleUrl
+    .split('/')
+    .slice(-1)[0]
+    .replace(/\.gts$/, ' ');
```
habdelra: seems a little strange, why are we replacing the extension with a space character? isn't the goal to remove the extension? Also, `.pop()!` might be a little more easy to read than `.slice(-1)[0]`

lukemelia: I think the space is a mistake

habdelra (approved):

PR #2199, opened 2025-02-24
Revert ai assistant view-code container styles and add snapshot

burieberry:
Before:
view-code

After:
view-code


lukemelia (approved):

habdelra (changes requested):
wait--this breaks the monaco read-only editor when there are code blocks in the AI assistant response. 

try asking the AI assistant to give you an javascript class example. or another question that will result in inline code blocks (within a &lt;UL&gt;).

I think the better approach is to push forward the CSS updates to be inclusive of the command output as well as  code blocks. if you need help with this I can make a PR for that. (and we should add snapshots for this as well)

habdelra:
My bad, I jumped the gun on that and rejected this before I was able to see the output. codeblocks look ok:

image

but I'm updating the PR to also snapshot the code blocks

habdelra (approved):

PR #2201, opened 2025-02-25
Add edit mode to code ref field

habdelra:
This PR add a simple edit field to a code ref. Code refs can be entered in the manner of `module/exported_name` (e.g.  the `internalKeyFor()` serialization of a code ref)


packages/host/tests/integration/components/card-basics-test.gts
```diff
@@ -674,10 +674,53 @@ module('Integration | card-basics', function (hooks) {
...
 
       await renderCard(loader, driver, 'edit');
-      assert.dom('input').doesNotExist('no input fields exist');
+
       assert
-        .dom('[data-test-ref]')
-        .containsText(`Module: http://localhost:4202/test/person Name: Person`);
+        .dom('[data-test-ref] input')
+        .hasValue(
+          `http://localhost:4202/test/person/Person`,
```
tintinthong: {testRealmURL}??

lukemelia (commented):
What is the behavior of the module specifier parsing combined with  auto-save. Are we going to have a situation where we attempt to load modules that don't exist during indexing?

habdelra:
> What is the behavior of the module specifier parsing combined with auto-save. Are we going to have a situation where we attempt to load modules that don't exist during indexing?

yeah that could definitely happen. do you think we should try to load modules from the edit template itself to verify if they are valid?

lukemelia:
> yeah that could definitely happen. do you think we should try to load modules from the edit template itself to verify if they are valid?

Not a bad idea, and don't update underlying saveable field until we're able to confirm it is valid.

In the future, we could have a chooser that let's you pick a file from your realms and then choose an export.

burieberry (approved):

habdelra:
Validation has been added to our code ref field in commit https://github.com/cardstack/boxel/pull/2201/commits/4e68dff8733f62d1a6eafb723d642416d0f1ab6a

packages/base/code-ref.gts
```diff
@@ -35,6 +46,84 @@ class BaseView extends Component<typeof CodeRefField> {
...
+      if (parts.length < 2) {
+        this.validationState = 'invalid';
+        return;
+      }
+
+      let name = parts.pop()!;
+      let module = parts.join('/');
+      try {
+        if (moduleIsUrlLike(module)) {
+          await loadCard(
```
lukemelia: Is this OK to use even if the module you are loading may not be a card?

habdelra: if the code ref is a `ResolvedCodeRef` type then yes--it can be used to load any export. for any other types of code refs then no. However, we are inherently treating the serialized code refs as `ResolvedCodeRef` types, so it should be safe. (and technically the export you are loading could be a card, the module can never be a card)

lukemelia: Hmm, looking at the source of `loadCard`, it looks like it throws if the imported things is not a BaseDef via `isBaseDef` check.

habdelra: ah, you're right. let me use the same validation that we use for the non-URL like code refs

lukemelia (approved):

PR #2202, opened 2025-02-25
CRM styling update and to seed realm

richardhjtan:


richardhjtan:
Tested in staging by branch and it works well

PR #2205, opened 2025-02-25
AI message: do not render html when it's in a code block

jurgenwerk:
The problem happens when the bot is streaming code: 

image

Here is how it looks after the change in this PR:
image

It turns out the browser will render HTML tags, such as `<h1>`, or `<template>`, even when they are placed inside `<pre><code> ... </code></pre>`. So we need to escape these tags in order to show the actual code (and not render it as html).


jurgenwerk:
Just FYI, once the code is finished streaming, it changes into monaco editors:

image

lukemelia (approved):

burieberry (approved):

PR #2214, opened 2025-02-26
Improve code editing skill

jurgenwerk:
Added more concrete instructions for placing <style> tags


jurgenwerk:
Merged this so we can demo it today

PR #2215, opened 2025-02-26
VSCode extension - Sync version

IanCal:
This is a large change to how the extension works. There's likely cleanup we can do, this has been done as a matter of urgency since the recent cursor change broke the existing extension completely.

The new extension allows you to

* Find your realms
* Pull down updated files from those realms
* Push up updated files from your local filesystem
* Watch your local file system and push up changes/new files as you create them

It also keeps the existing skills loading, but stores them on disk too, along with your selections so you don't need to recreate anything.

Files are stored by default within the extensions storage folder.

With this, all cursor features should work much more simply, as it doesn't interact with our code.

I can run through this on a call with anyone about where the buttons are/etc.

See CS-8045 for a demo video and a built version of the extension


packages/vscode-boxel-tools/README.md
```diff
@@ -17,11 +10,66 @@ You can also build it [locally](#development).
 
 ## Usage
 
-Open a new window and in the command palette select "Boxel Tools: Attach to Boxel Workspaces".
+### Getting Started
+
+1. Install the extension from the VS Code Marketplace or [build it locally](#development).
+2. Open a new window and in the command palette (F1 or Ctrl+Shift+P), select "Boxel Tools: Attach to Boxel Workspaces".
+3. You will be prompted to log in with your Matrix credentials:
```
lukemelia: Maybe we call them Boxel credentials instead of Matrix credentials. We don't use the term Matrix in the app UI AFAIK.

packages/vscode-boxel-tools/README.md
```diff
@@ -17,11 +10,66 @@ You can also build it [locally](#development).
...
+
+To add the Boxel workspaces to your VSCode workspace, use the + button in the Boxel view or the command "Add Boxel Workspaces Folder to Workspace".
+
+### Working with Workspace Files
+
+Once you've downloaded workspace files:
+
+1. Navigate to the files in your Explorer view
+2. Edit files as you normally would in VS Code
+3. If file watching is enabled, changes will automatically sync to the remote workspace
```
lukemelia: Is the opposite true? Will changes on the server automatically sync to the local workspace? We should probably be clear about this in the docs.

packages/vscode-boxel-tools/README.md
```diff
@@ -17,11 +10,66 @@ You can also build it [locally](#development).
...
+
+The extension includes a Skills Manager that allows you to enable various coding assistants:
+
+1. Find the "Boxel" view in the Explorer sidebar
+2. Browse available skills grouped by source (Base, Catalog, or your custom workspaces)
+3. Check the checkbox next to skills you want to enable
+4. Enabled skills will be applied to your editing experience automatically
+
+Skills are stored in two locations:
+- Skill definitions are saved in your user folder
```
lukemelia, line 50: Do you have to worry about updating them if you change a skill in your realm? If so, how do you do it?

packages/vscode-boxel-tools/src/skills.ts
```diff
@@ -35,6 +82,148 @@ export class SkillsProvider
...
+
+      fs.writeFileSync(stateFilePath, JSON.stringify(stateData, null, 2));
+      console.log('Skill state saved successfully');
+    } catch (error) {
+      console.error('Error saving skill state:', error);
+    }
+  }
+
+  // Load skills from storage
+  private loadSkillsFromStorage(): void {
```
lukemelia, line 163: Why store/read skill details to/from here rather than storing references to the skill instance json?

lukemelia (approved):
Skimmed, did not read closely, but looks good overall. Left a few comments.

packages/vscode-boxel-tools/src/skills.ts
IanCal: Yeah I should change that to store the json as is and just read as necessary

packages/vscode-boxel-tools/README.md
IanCal: There's a reload, this should be made clearer (and maybe do this on load or something, or pull any changes when we enable one).

packages/vscode-boxel-tools/README.md
IanCal: It's not, no - only pushing changes live. We should make this clear both in the docs and also probably rename the part of the extension that does this so it's clear that this is the direction. "Auto push changes" or something.

We have a matrix client in this, so with moving sse to matrix events we could do bidirectional live syncing probably, although being clear about how conflicts are handled is important.

PR #2217, opened 2025-02-26
Improve styling of "cannot write" message in spec

tintinthong:
Screenshot 2025-02-27 at 00 09 42

Screenshot 2025-02-27 at 00 11 55

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/spec-preview.gts
+++ b/packages/host/app/components/operator-mode/code-submode/spec-preview.gts
@@ -21,7 +21,7 @@ import {
   RealmIcon,
   LoadingIndicator,
 } from '@cardstack/boxel-ui/components';
-import { not, cn } from '@cardstack/boxel-ui/helpers';
+import { cn } from '@cardstack/boxel-ui/helpers';
 
 import {
   type ResolvedCodeRef,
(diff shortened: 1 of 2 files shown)


richardhjtan (approved):

burieberry:
Some of the items in Base Realm actually have specs, for example the boolean field, but it's not possible to preview the spec on the Spec accordion item. Is this intentional?

tintinthong:
> Some of the items in Base Realm actually have specs, for example the boolean field, but it's not possible to preview the spec on the Spec accordion item. Is this intentional?

Hmmm. I see your point. Is it preferred that when its not writable to just render in isolated rather than displaying the "cannot write" message

tintinthong:
Yea. I can see how this got lost in translation abit. 

So, we already don't display create button when its not writable. We shud just render isolated and be done with it rather than displaying this particular message

burieberry (approved):

tintinthong:
Updated @burieberry. Adding test now

**Has boxel spec in non-writable realm**

Screenshot 2025-02-27 at 15 18 38

**No boxel spec in non-writable realm**

Screenshot 2025-02-27 at 15 18 48

PR #2221, opened 2025-02-27
Make Spec Title Editable

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-7932/computed-default-name-our-custom-edit-field-should-cover-this

**Demo**

https://www.loom.com/share/a9110260a6a04e4e8dc920e8272a113b?sid=ac1c1a0b-aebd-415b-bc9e-15870ef8e56e


tintinthong:
Iniitially, I thought that the edit view for Spec title was supposed to be big like a header. 

Bcos, the spec in the code mode doesn't change from edit to isolated when u look at `author.gts`. It is only in edit mode. So I thought that the input edit of title should be BIG. How does this look in this scenario?

tintinthong:
The removal of name is a great simplification --  thanks

tintinthong:
@lucaslyl I wud like to review after lint completes. can u get the lint to succeed

packages/base/spec.gts
```diff
@@ -136,10 +128,16 @@ export class Spec extends CardDef {
...
+            <div class='header-info-padding-container'>
+              <h1 class='title' id='title' data-test-title>
+                <@fields.title />
+              </h1>
+            </div>
+            <div class='header-info-padding-container'>
+              <p class='description' data-test-description>
+                <@fields.description />
+              </p>
+            </div>
```
burieberry: let's not do this template and css change here. we can create an edit template that handles the semantics and styles better. it can be outside of the scope of this PR.

lucaslyl: sure, i will create another pr of the edit template

lucaslyl:
> Iniitially, I thought that the edit view for Spec title was supposed to be big like a header.
> 
> Bcos, the spec in the code mode doesn't change from edit to isolated when u look at `author.gts`. It is only in edit mode. So I thought that the input edit of title should be BIG. How does this look in this scenario?

I c,I will create a separate edit template pr for handle this styling: https://linear.app/cardstack/issue/CS-8085/edit-template-for-spec

tintinthong (approved):
Looks good thx

PR #2224, opened 2025-02-27
Support for creating new instance of card that contains CodeRef field

habdelra:
Fixes bug where new instance of code ref field was unable to be created from playground. Interestingly this came down to the fact that during deserialization, the empty code ref field's serialized form is `{}`.

### DIFF
--- a/packages/base/code-ref.gts
+++ b/packages/base/code-ref.gts
@@ -33,6 +33,25 @@ function moduleIsUrlLike(module: string) {
   );
 }
 
+function maybeSerializeCodeRef(
+  codeRef: ResolvedCodeRef | {} | undefined,
+  stack: CardDef[] = [],
+) {
+  if (codeRef && 'module' in codeRef) {
+    if (moduleIsUrlLike(codeRef.module)) {
+      // if a stack is passed in, use the containing card to resolve relative references
+      let moduleHref =
+        stack.length > 0
+          ? new URL(codeRef.module, stack[0][relativeTo]).href
+          : codeRef.module;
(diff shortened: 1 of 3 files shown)


packages/base/code-ref.gts
```diff
@@ -33,6 +33,25 @@ function moduleIsUrlLike(module: string) {
...
+          ? new URL(codeRef.module, stack[0][relativeTo]).href
+          : codeRef.module;
+      return `${moduleHref}/${codeRef.name}`;
+    } else {
+      return `${codeRef.module}/${codeRef.name}`;
+    }
+  }
+  return undefined;
+}
+
```
habdelra: this got moved to the top because glint was providing garbage type hints when it was at the bottom. The actual change here is line 40 and line 37, where we assert that `{}` is a type that `codeRef` could be, and that we check for presence of `module` key in the `codeRef` in the initial guard.

lukemelia (approved):
LGTM

@backspace heads up new expectEvents usage

habdelra:
> LGTM
> 
> @backspace heads up new expectEvents usage

the event stuff was copy pasta, they are not neceesary for this test. i'm removing it

packages/base/code-ref.gts
burieberry: the function for checking `isResolvedCodeRef` in runtime-common does pretty much the same thing as line 40

burieberry (approved):

packages/base/code-ref.gts
habdelra: nice!

PR #2225, opened 2025-02-27
Replace PrerenderedCardSearch with direct API calls to prevent component recreation during autosave

richardhjtan:


richardhjtan:
After pairing with @tintinthong , we still appreciate the presents of `PrerenderedCardSearch` because it handles the query function and liveness check pretty well.

We also found:
- there is a pain point to acquire cards in PrerenderedCardSearch to be consumed at the parent component
- to solve that issue, there was an approach to use a method inside the PrerenderedCardSearch that trigger to save cards value at parent component level
- this approach will cause the component rerender(recreated) with ember id increments 

Resolution:
- use props passing to children pattern, which will propose in another PR

tintinthong:
> there is a pain point to acquire cards in PrerenderedCardSearch to be consumed at the parent component

Thanks for the explanation -- I think it is a good breakdown. 

**1.**

I think there might be a "missing" api here. 

There are use-case where we only want ids but the PrerenderedCardSearch doesn't allow a programatic way to call it in javascript (like a resource perhaps). I have encountered this twice. I think 1 in Lucas work in work planner and this scenario here. 

Specifically, the user probably wants an array of ids that update live. And the ids to be stable as possible. But change upon newer cards being introduced

**2.**

The other layer of the problem is that this yielding pattern shared with the schema editor uses these BoundArgs which doesn't allow any undefined / null for the types as they must be set in the code-submode

I'm open to other solution (ie your prop drillling one) but I find that pattern 1. and 2. roped us/me into this unconvention of setCardIds from the block of prerendered card search. We can keep this as an example if we encounter this again.

PR #2228, opened 2025-02-28
Test correct template returned from plural components

tintinthong:
~This broke in auto-save~

~It didn't break otherwise in card-basics~ This broke in card-basics!

~I think different async re-renders when calling `.children` ie `getBoxComponent` and `recompute` (field hydration)~

I removing `.prevChildren` cache in box to see if test breaks. Im missing context why this actually is a thing, there are other caches that support it at the component level. Less cache good.

I believe this will make clearer the behaviour of what we see in https://github.com/cardstack/boxel/pull/2178

My notes: The modifier still getes torn down when we reorder links unless we still have a cahce. We need, additionally, to put the cache in linksToManyComponent. Still haven't proven that re-render torn down the modifier

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -350,12 +350,12 @@ function callSerializeHook(
 
 function cardTypeFor(
   field: Field<typeof BaseDef>,
-  boxedElement: Box<BaseDef>,
+  boxedElement?: Box<BaseDef>,
 ): typeof BaseDef {
   if (primitive in field.card) {
     return field.card;
   }
-  if (boxedElement.value == null) {
+  if (boxedElement === undefined || boxedElement.value == null) {
     return field.card;
   }
   return Reflect.getPrototypeOf(boxedElement.value)!
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -3145,7 +3145,6 @@ export class Box<T> {
         });
       }
     });
-    this.prevChildren = newChildren;
     return newChildren;
   }
 }
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -3145,6 +3145,9 @@ export class Box<T> {
         });
       }
     });
+    if (state.useIndexBasedKeys) {
+      this.prevChildren = newChildren;
+    }
     return newChildren;
   }
 }
--- a/packages/base/field-component.gts
+++ b/packages/base/field-component.gts
@@ -386,10 +386,6 @@ function fieldsComponentsFor<T extends BaseDef>(
   target: object,
   model: Box<T>,
 ): FieldsTypeFor<T> {
-  // This is a cache of the fields we've already created components for
-  // so that they do not get recreated
-  let stableComponents = new Map<string, BoxComponent>();
-
   return new Proxy(target, {
     get(target, property, received) {
       if (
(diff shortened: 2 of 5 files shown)


tintinthong:
Of course these are the problems with the CI run


After removing the cache at the field proxy level, 

**1. linksToMany component was unstable and we can put a cache in linksToMany component**
not ok 96 Chrome 133.0 - [undefined ms] - Global error: Uncaught TypeError: Cannot read properties of undefined (reading 'prepare') at 

**2. The previousChildren were use for stability of containsMany component. Perhaps, we can put the same cache in containsMany**
not ok 99 Chrome 133.0 - [508 ms] - Integration | card-basics > cards allowed to be edited: component stability when editing containsMany primitive field

1. is fixable. 2. We will see 

EDIT: previousChildren is broken for linksToMany but containsMany works bcos its depenedent  on indexByKey and containsMany doesn't sort 

**Notes**
1. In addition, after this, we have to put test field subclass re-renders within linksToMany and containsMany component
2. Also, note-to-self the field proxy was not a good cache bcos it has too little information thats why linksToMany and containsMany component are new options for putting caches (this is whawt prevChildren are for)
3. I always wondered why @fields.someField gets called several times if you have this.args.someField[0] and this.args.someField loaded in ur javascript

tintinthong:
Note-to-self: 

Reorder thing doesn't work bcos of 2 things 
- stable component cache from top-level fields proxy
- prevChildren

tintinthong:
More Notes 

And it seems that a linksToManyComponent cache is not even backed by a Box[] key that has changed. Reordering of linksToMany field that does this.args.model.array = <some other array> doesn't actually do anything to give u an updated cache at all. It does change the value but that also means that when getBoxComponent is looking for a the component it is serving the old template with the new values. So things like re-ordering of linksToMany is fundamentally broken bocs we are thinking we serve a stale cache with their values changed. This is problematic for polymorphic things bcos the templates are different. 

In conclusion, the linksToMany component is not participating well in the idea of a Box. This is what ed said last time when solving other bugs. prevChildren is ok for now I think, we shudnt remove it (revisit when re-include linksToMany component)

What im trying to do now but I can't seem to understand 

- Why this.args.model.linksToMany = <something> doesn't trigger .set inside of the box. It triggers a re-render but there is no .set method run from the Box. -- I think maybe this is discovered already. EDIT: It should trigger a .set now if you use `arrayField.set` rather than mutating the values from underneath

- The 2nd layer is. If I changed something on the template like a tracked property (not a field), does that trigger the update in the box.

tintinthong:
Note-to-self 

After adding linksToMany cache. Look at booking

There is this bug 


Screenshot 2025-03-01 at 13 37 01


EDIT: The linksToMany cache breaks almost all the linksToMany cards. It actually has to do with a box element being passed that is undefined

tintinthong:
Behaviour in auto-save is differing from card-basics. The auto-save behaviour doesn't persist the template of the linksToMany. By https://github.com/cardstack/boxel/pull/2228/commits/88b2db90fcd7c43a009df028636d859848bc8103, card-basics does have the correct behaviour and maintains stability of the linksToMany templates but this doesn't account for round-trips

EDIT: Im leaning towards NOT fixing auto-save and don't consider that as an issue

EDIT: previous behavior in main returns a stable template for auto-save test. No reason to NOT still have stability -- I think. Unless there are changes in the field system or deserialized data

packages/base/card-api.gts
```diff
@@ -3145,7 +3145,9 @@ export class Box<T> {
         });
       }
     });
-    this.prevChildren = newChildren;
+    if (state.useIndexBasedKeys) {
+      this.prevChildren = newChildren;
+    }
```
tintinthong, line 3150: probably the wrong change here

tintinthong:
Looking at the re-order test independently. A changed in @tracked property via a button doesn't update any known caches of any of the box components. Hence, everytime there is a recompute we are serving an old cache on load. Perhaps this is the expected behaviour??

PR #2232, opened 2025-03-01
Improve playground card rendering

burieberry:
Changes:
- Previously, we were changing the args to card resource each time `persistSelections` was called. Now, we're only changing the args if they have actually changed.
- Stop relying on `cardType` resource, which was async. Instead now getting the card-ref and moduleId via available sync methods. This simplified the isLoading state.

Now I'm left with these options: 

Option 1: More elegant, with no screen flash, but the card-resource ends up loading the resource multiple times.
https://github.com/user-attachments/assets/5d4acf6f-666e-4a22-8ae3-cba67e8ee1d7

Option 2: Ensures that the card resource is only loaded once (when code mode's `moduleContentsResource` loading state is false), but because we're re-rendering the component, there's 1 screen flash with loading indicator shown.
https://github.com/user-attachments/assets/33ee53a2-a821-4a8f-8f51-18a9635dea5a

Note: Still planning on improving the dropdown by separating the trigger component and limiting the prerendered card search to the dropdown list component only.

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode.gts
+++ b/packages/host/app/components/operator-mode/code-submode.gts
@@ -29,6 +29,7 @@ import { File } from '@cardstack/boxel-ui/icons';
 
 import {
   identifyCard,
+  internalKeyFor,
   isCardDef,
   isCardDocumentString,
   hasExecutableExtension,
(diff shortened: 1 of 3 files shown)


lukemelia:
Option 1 is definitely a better user experience. Perhaps you can go with that and then review with @ef4 on Wednesday.

PR #2235, opened 2025-03-03
Fix muted syntax visual discrepancy in some code snippets

jurgenwerk:
When the assistant responded with code, the syntax coloring was odd in a couple of places, and this was happening because the `language` passed into the monaco editor was `typescript`, while the code was actually a part of a `gts` file.  So the editor didn't expect things like CSS and `template` tags in it, failing to render it properly.

This was fixed by adjusting the code module editing skill so that snippets now come wrapped in a gts markdown tag rather than typescript tag. 

image


packages/host/app/utils/editor-language.ts
```diff
@@ -11,12 +11,12 @@ export type MonacoLanguageConfig = {
...
 
 const gjsConfig: LanguageInfo = {
-  id: 'glimmerJS',
+  id: 'gjs',
   extensions: ['.gjs'],
 };
 
 const gtsConfig: LanguageInfo = {
-  id: 'glimmerTS',
+  id: 'gts',
```
jurgenwerk, line 19: This change was needed so that monaco can understand glimmer language when the assistant responds with ` ```gts ` code snippets. Alternatively we could change the skill to use ` ```glimmerTS ` but I think the former solution sounds more standard

lukemelia (approved):

burieberry (approved):
Nice

PR #2239, opened 2025-03-04
Improve way to verify is a skill card in spec

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8056/cover-boxel-spec-for-skill

This PR is to improve the identification of skill cards in the spec preview component 

We would determine the skill card when:
- the code ref is module under[ https://cardstack.com/base/skill-card](https://cardstack.com/base/skill-card)
- the code ref name is SkillCard
- also cover the class that skill card is a subclass for another card - finding under the chain of card type


lukemelia (commented):
This seems like it would not cover a skill that is a subclass of another skill.

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -446,14 +446,11 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
   }
 
   isSkill(selectedDeclaration: CardOrFieldDeclaration) {
-    if (selectedDeclaration.exportName === 'SkillCard') {
-      return true;
-    }
```
tintinthong, line 451: I thought we just wanted to remove this

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -446,14 +446,11 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
     if (
-      selectedDeclaration.super &&
-      selectedDeclaration.super.type === 'external' &&
-      selectedDeclaration.super.name === 'SkillCard' &&
-      selectedDeclaration.super.module ===
+      selectedDeclaration.cardType.type &&
+      isResolvedCodeRef(selectedDeclaration.cardType.type.codeRef) &&
+      selectedDeclaration.cardType.type.codeRef.name === 'SkillCard' &&
+      selectedDeclaration.cardType.type.codeRef.module ===
```
tintinthong: The original code was clearer that it didn't do much. 

A possible way to fix this is to introduce `specType` data  in card-type resource. Another way is to just recurse the chain on read and make a stopping condition based on the type data that is already

I'd recommend the 2nd way unless that is visibly inefficient

The test should have a 1-level super and a 2-level super

richardhjtan:
Thanks @lukemelia @tintinthong for the feedback. I switch the approach by finding the chain in `cardType.typeCache` that consists the boxel based skill card module

lukemelia (commented):
Is this typeCache guaranteed to be populated when this check is performed?

richardhjtan:
> Is this typeCache guaranteed to be populated when this check is performed?

To be assured, I introduce a function to find the class matches the desired codeRef at the first level or recursively.

https://github.com/cardstack/boxel/pull/2239/commits/693f9cc42518ad8f43777aabb6786a8e129f753f

lukemelia (approved):

PR #2240, opened 2025-03-04
Add a separate skill for code diffing

IanCal:
Heavily taken from Aider (https://aider.chat) and modified to use examples of gts files, and remove unnecessary things.

We may cut this down or alter it more, but this appears to get a decent result with 3.7 sonnet so we can move on.

Contents:

To alter code for the user, you must use a SEARCH/REPLACE block.

A SEARCH/REPLACE block has three sections. A filename for the file to update, a section of code to search for, and the code to replace it with. All code within the SEARCH will be replaced.

Example adding an import:

realm/user/filename.gts
```gts
<<<<<<< SEARCH
import { Component } from 'https://cardstack.com/base/card-api';
import { or } from '@cardstack/boxel-ui/helpers';
=======
import { Component } from 'https://cardstack.com/base/card-api';
import { MarkdownField } from 'https://cardstack.com/base/markdown';
import { or } from '@cardstack/boxel-ui/helpers';
>>>>>>> REPLACE
```

Example deleting a field by having an empty replace block.

realm/user/filename.gts
```gts
<<<<<<< SEARCH
  @field description = contains(StringField);
  @field categories = containsMany(Category);
  @field attemptsRemaining = contains(NumberField, {
    computeVia: function() {
      return 4; // Start with 4 attempts
    }
  });
=======
  @field description = contains(StringField);
  @field categories = containsMany(Category);
>>>>>>> REPLACE
```

Example changing text within a template

realm/user/filename.gts
```gts
<<<<<<< SEARCH
      <template>
      <div class="connections-game">
        <header class="game-header">
          <h1 class="game-title">Connections</h1>
          <div class="game-description">{{@model.description}}</div>
        </header>
=======
      <template>
      <div class="connections-game">
        <header class="game-header">
          <h1 class="game-title">Connections Game</h1>
          <div class="game-description">{{@model.description}}</div>
        </header>
>>>>>>> REPLACE
```

Every *SEARCH/REPLACE block* must use this format:
1. The *FULL* file path alone on a line, verbatim. No bold asterisks, no quotes around it, no escaping of characters, etc.
2. The opening fence and code language, eg: ```gts
3. The start of search block: <<<<<<< SEARCH
4. A contiguous chunk of lines to search for in the existing source code
5. The dividing line: =======
6. The lines to replace into the source code
7. The end of the replace block: >>>>>>> REPLACE
8. The closing fence: ```

Use the *FULL* file path, as shown to you by the user.

Every *SEARCH* section must *EXACTLY MATCH* the existing file content, character for character, including all comments, docstrings, etc.
If the file contains code or other data wrapped/escaped in json/xml/quotes or other containers, you need to propose edits to the literal contents of the file, including the container markup.

*SEARCH/REPLACE* blocks will *only* replace the first match occurrence.
Including multiple unique *SEARCH/REPLACE* blocks if needed.
Include enough lines in each SEARCH section to uniquely match each set of lines that need to change.

Keep *SEARCH/REPLACE* blocks concise.
Break large *SEARCH/REPLACE* blocks into a series of smaller blocks that each change a small portion of the file.
Include just the changing lines, and a few surrounding lines if needed for uniqueness.
Do not include long runs of unchanging lines in *SEARCH/REPLACE* blocks.

To move code within a file, use 2 *SEARCH/REPLACE* blocks: 1 to delete it from its current location, 1 to insert it in the new location.

Pay attention to which filenames the user wants you to edit, especially if they are asking you to create a new file.

If you want to put code in a new file, use a *SEARCH/REPLACE block* with:
- A new file path, including dir name if needed
- An empty `SEARCH` section
- The new file's contents in the `REPLACE` section


packages/catalog-realm/SkillCard/search-replace-editing.json
```diff
@@ -0,0 +1,18 @@
...
+      "commands": [],
+      "title": "Search/replace code editor",
```
jurgenwerk: Do you think we can come up with a better title so it's not too confusing for the user?

image

Perhaps this should just be "Code editing", and the other skill that has instructions for gts file can be "Card definition editing"?

jurgenwerk (approved):
I think this looks ok, I was doing some testing in my WIP with a very similar instruction copied from aider, and I got search/replace blocks returned to me.

Regarding my skill name comment, I think we can also change it later if needed.

packages/catalog-realm/SkillCard/search-replace-editing.json
lukemelia: "Source code editing"

lukemelia: and "Boxel Coding"

lukemelia (approved):

packages/catalog-realm/SkillCard/search-replace-editing.json
jurgenwerk: Also should this skill be in base realm instead? Next to code-module-editing.json (probably needs to be renamed to boxel-coding.json)

jurgenwerk: I have this commit in my work in progress PR that changes skill names as suggested, and adds them as default in code mode: https://github.com/cardstack/boxel/pull/2245/commits/0da24a98871378eec5da3757d89badbaefdee0aa

@IanCal If you like we can cherry pick it and add to this PR

PR #2241, opened 2025-03-04
Re-ordering linksToMany doesn't preserve templates (and UI component state)

tintinthong:
https://linear.app/cardstack/issue/CS-7936/reordering-linkstomany-cards-results-in-incorrect-fitted-components

Now template and component state should be preserved

https://github.com/user-attachments/assets/3afcc82b-48cc-4820-8923-10dd379a58a5

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -3100,6 +3100,7 @@ export class Box<T> {
   }
 
   private prevChildren: Box<ElementType<T>>[] = [];
+  private prevValues: ElementType<T>[] = [];
 
   get children(): Box<ElementType<T>>[] {
     if (this.state.type === 'root') {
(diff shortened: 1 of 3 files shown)


packages/base/card-api.gts
```diff
@@ -3117,10 +3118,10 @@ export class Box<T> {
       );
     }
 
-    let { prevChildren, state } = this;
+    let { prevChildren, prevValues, state } = this;
     let newChildren: Box<ElementType<T>>[] = value.map((element, index) => {
-      let found = prevChildren.find((oldBox, i) =>
-        state.useIndexBasedKeys ? index === i : oldBox.value === element,
```
tintinthong, line 3123: **The main issue was here** 

We were relying on getting  .value although that was already set during the array assigment. This caused us to always be checking arrays that was already of the same order

packages/host/tests/integration/components/card-basics-test.gts
```diff
@@ -2283,6 +2285,157 @@ module('Integration | card-basics', function (hooks) {
       await waitUntil(() => cleanWhiteSpace(root.textContent!) === 'Quint');
     });
 
+    test('Re-order linksToMany fields action will use the cached template that UI state', async function (assert) {
```
backspace: Is there a word missing here? I don’t understand the name

backspace (approved):
Nice tests 🎉

packages/host/tests/integration/components/card-basics-test.gts
burieberry: `will use the cached template OF that UI state`??? 🤔 Maybe there's a better way to phrase

burieberry (approved):

PR #2242, opened 2025-03-04
Cs 8060 make spec navigatable to the instance in code mode

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8060/make-spec-navigatable-to-the-instance-in-code-mode

Result:

https://github.com/user-attachments/assets/a64b98cf-1105-48af-943a-11272329db25


packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -222,6 +223,19 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
...
+  @action viewSpecInstance() {
+    if (!this.args.selectedId) {
+      return;
+    }
+
+    try {
+      const selectedUrl = new URL(this.args.selectedId);
+      this.operatorModeStateService.updateCodePath(selectedUrl);
+    } catch (error) {
+      console.error('Failed to open spec instance:', error);
```
lukemelia: Is this an expected circumstance? If not, I don't think we need the try/catch. We can just let the failure bubble. If so, we should have test coverage for it.

lucaslyl: Sure, you're right. This isn't an expected circumstance. I'll remove the try/catch and let any failures bubble up naturally. Thanks for the advice

tintinthong (approved):
"View Instance" seems like a better use of words imo.

packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -351,6 +353,8 @@ module('Acceptance | Spec preview', function (hooks) {
     assert
       .dom('[data-test-spec-selector-item-path]')
       .hasText('pet-entry-2.json');
+    assert.dom('[data-test-view-spec-instance]').exists();
+    await click('[data-test-view-spec-instance]');
```
tintinthong: You have to have another test that clicks on the button and assert that you have the spec instance in the RHS rendered in isolated mode

tintinthong (changes requested):
Need to add test and change to "View Instance"

packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -467,4 +471,30 @@ module('Acceptance | Spec preview', function (hooks) {
...
+
+    await waitFor('[data-test-editor]');
+    assert.dom('[data-test-editor]').hasAnyText();
+    assert.dom('[data-test-editor]').containsText('Person');
+    assert.dom('[data-test-editor]').containsText('Spec');
+    assert.dom('[data-test-editor]').containsText('specType');
+
+    await waitFor('[data-test-module-href]');
+    assert.dom('[data-test-module-href]').containsText(`${testRealmURL}person`);
+    assert.dom('[data-test-exported-name]').containsText('Person');
```
tintinthong: I said "Check that it opens in isolated mode". You can use the attributes here  (data-test-card-format)

```
<div id="ember1609" class="ember-view boxel-card-container boundaries field-component-card
                        isolated-format display-container-true" data-scopedcss-35a90eaf3c-4ee8da6495="" data-test-boxel-card-container="" data-test-card="http://localhost:4201/experiments/Spec/1" data-test-card-format="isolated" data-test-field-component-card="" data-scopedcss-b8fe299739-015c45ff6d="" data-scopedcss-6ef7616ef3-bc76e072f4="">
      
```

lucaslyl: Sure, I have studied other file patterns to apply this change with:
[data-test-card="${testRealmURL}person-entry"][data-test-card-format="isolated"]

I'm applying the check on data-test-card specifically to ensure we're targeting the exact card instance we need

tintinthong (approved):

PR #2244, opened 2025-03-04
Extend playground to preview field instances

burieberry:
Changes:
- Can preview compound fields on Playground panel in code submode
- "Create new instance": adds new field instance to field spec's "containedExamples" collection (generates spec if it doesn't already exist)
- "Choose another instance": Opens modal with instances from spec's "containedExamples", can change selected instance
- If a field has multiple specs, can change selected spec from Spec panel (this will update Playground spec selection)
- Displays next available field instance if persisted field index no longer exists due to deletion of an instance
- Display Add button if spec exists but has no examples
- Refactored FormatChooser component: Render only available formats when previewing fields

field-preview

field-instance-chooser

----

## TODOs: (Will create follow-up tickets for these)
- Do not display playground for primitive fields
- Improve live update UX for fields
- ~Hide Recent section in Instance Chooser dropdown for fields?~ Display instances in atom view
- (Related) Spec panel preview should update when selected declaration changes

### Feedback from design team 3/11:
- If spec has no examples, generate an example on selecting the spec without needing to press Add button
- If field has no spec, generate one on selecting the field without needing to press Add button
- Add "fitted" view for fields, only isolated should be not available
- Number index the examples list in playground & spec containsMany
- In Recent section of Instance dropdown, display atom view of instances


packages/host/tests/acceptance/code-submode/playground-test.gts
```diff
@@ -995,4 +1199,412 @@ export class BlogPost extends CardDef {
...
+        .hasText('Terrible product');
+      await realm.write('blog-post.gts', updatedCommentField),
+        await waitUntil(
+          () =>
+            document.querySelector('[data-test-embedded-comment-title]') ===
+            null,
+        );
+      assert.dom('[data-test-embedded-comment-title]').doesNotExist();
+    });
+  });
```
backspace: Is this file getting to the point where it should be broken up? I don‘t know that we have a metric for that but it might be time for a subdirectory 😯

backspace (approved):
I tried this via the preview deployment, works as described 😀

packages/host/tests/acceptance/code-submode/playground-test.gts
burieberry: Ha good point. I can separate the modules into different files in a subdir

packages/host/tests/acceptance/code-submode/playground-test.gts
```diff
@@ -995,4 +1199,412 @@ export class BlogPost extends CardDef {
...
+        .doesNotExist();
+
+      // TODO: spec panel updates when spec is created from playground
+      // await click('[data-test-accordion-item="spec-preview"] button');
+      // assert.dom('[data-test-boxel-input-id="spec-title"]').hasValue('Quote');
+      // assert
+      //   .dom(
+      //     '[data-test-contains-many="containedExamples"] [data-test-item="0"] [data-test-field="quote"] input',
+      //   )
+      //   .hasNoValue();
```
tintinthong: Noted on this. Its supposed to work but there was a digression here.

tintinthong (approved):

lukemelia (approved):

PR #2247, opened 2025-03-06
Make schema editor field row clickable

FadhlanR:
In this PR, I made the field name text in the schema editor field row clickable so that the cursor in the Monaco editor jumps to the line defining that field. Originally, the requirement was to make the entire field row clickable. However, since there are other buttons, such as a pill-type button and a three-dot button, inside the field row, I made only the field name text clickable to avoid nested buttons. I'm open to any other approaches.

Here’s a revised version of your sentence for better clarity and grammar:  

Note: These updates won't work in `CardDef`. There is a bug in `schema-analysis-plugin` when parsing `possibleFields` for `CardDef`. I'll submit the fix in a separate PR.

https://github.com/user-attachments/assets/7174ef8f-1659-4893-befb-6e9cff71d83b

### DIFF
--- a/packages/host/app/components/operator-mode/code-editor.gts
+++ b/packages/host/app/components/operator-mode/code-editor.gts
@@ -128,7 +128,9 @@ export default class CodeEditor extends Component<Signature> {
   @cached
   private get initialMonacoCursorPosition() {
     if (this.codePath) {
-      let recentFile = this.recentFilesService.findRecentFile(this.codePath);
+      let recentFile = this.recentFilesService.findRecentFileByURL(
+        this.codePath.toString(),
+      );
       if (recentFile?.cursorPosition) {
         return new Position(
           recentFile.cursorPosition.line,
--- a/packages/host/app/components/operator-mode/code-editor.gts
+++ b/packages/host/app/components/operator-mode/code-editor.gts
@@ -140,13 +140,14 @@ export default class CodeEditor extends Component<Signature> {
     }
     let loc;
     let selectedFieldName = this.operatorModeStateService.state.fieldSelection;
+    let { selectedDeclaration } = this.args;
     if (
       selectedFieldName &&
-      this.args.selectedDeclaration &&
-      'possibleFields' in this.args.selectedDeclaration &&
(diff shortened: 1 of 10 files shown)


tintinthong:
Screenshot 2025-03-06 at 21 50 37

Is there a way to set a cursor straight to the fieldName.

tintinthong (approved):
I have tested and the behavior seems pretty solid. Just nitpick about the cursor position (just seems weird position bcos the user will not type there straight away)

packages/host/app/components/operator-mode/code-editor.gts
```diff
@@ -136,14 +138,32 @@ export default class CodeEditor extends Component<Signature> {
...
+
+    if (!loc) {
+      loc =
+        this.args.selectedDeclaration?.path?.node &&
+        'body' in this.args.selectedDeclaration.path.node &&
+        'loc' in this.args.selectedDeclaration.path.node.body &&
+        this.args.selectedDeclaration.path.node.body.loc
+          ? this.args.selectedDeclaration?.path?.node.body.loc
+          : undefined;
+    }
```
lukemelia:
```suggestion
    let loc;
    let selectedFieldName = this.operatorModeStateService.state.fieldSelection;
    let { selectedDeclaration } = this.args;
    if (
      selectedFieldName &&
      selectedDeclaration &&
      'possibleFields' in selectedDeclaration &&
      selectedDeclaration.possibleFields
    ) {
      let possibleFields = selectedDeclaration.possibleFields;
      let field = possibleFields.get(selectedFieldName);
      loc =
        field?.path?.node && 'loc' in field.path.node && field.path.node.loc
          ? field.path.node.loc
          : undefined;
    }

    if (!loc) {
      loc =
        selectedDeclaration?.path?.node &&
        'body' in selectedDeclaration.path.node &&
        'loc' in selectedDeclaration.path.node.body &&
        selectedDeclaration.path.node.body.loc
          ? selectedDeclaration?.path?.node.body.loc
          : undefined;
    }
```

lukemelia (approved):

PR #2248, opened 2025-03-06
Prevent file from being auto attached in interact mode

FadhlanR:


packages/host/tests/acceptance/ai-assistant-test.gts
```diff
@@ -380,6 +380,32 @@ module('Acceptance | AI Assistant tests', function (hooks) {
     assert.dom('[data-test-llm-select-selected]').hasText('claude-3.5-sonnet');
   });
 
+  test('ensures to not display auto attached file in interact mode', async function (assert) {
```
lukemelia:
```suggestion
  test('auto-attached file is not displayed in interact mode', async function (assert) {
```

packages/host/app/components/matrix/room.gts
```diff
@@ -289,7 +291,10 @@ export default class Room extends Component<Signature> {
       },
     });
 
-    if (!this.autoAttachedFileUrl) {
+    if (
+      !this.autoAttachedFileUrl ||
+      this.operatorModeStateService.state.submode === Submodes.Interact
+    ) {
```
lukemelia: Let's put this conditional logic inside the `autoAttachedFileUrl` getter instead. i.e. have it return undefined unless this.operatorModeStateService.state.submode is code mode

lukemelia (changes requested):

lukemelia (approved):

PR #2249, opened 2025-03-06
add white-space and word-break property on boxel-button component lvl

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8100/improve-boxel-button-ui-to-handle-long-text

Before:
image

After:
Screenshot 2025-03-06 at 23 34 23

### DIFF
--- a/packages/boxel-ui/addon/src/components/button/index.gts
+++ b/packages/boxel-ui/addon/src/components/button/index.gts
@@ -103,7 +103,8 @@ export default class ButtonComponent extends Component<Signature> {
           height: min-content;
           align-items: center;
           border-radius: 100px;
-          white-space: nowrap;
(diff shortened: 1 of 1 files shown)


packages/boxel-ui/addon/src/components/button/index.gts
```diff
@@ -103,7 +103,7 @@ export default class ButtonComponent extends Component<Signature> {
           height: min-content;
           align-items: center;
           border-radius: 100px;
-          white-space: nowrap;
```
lucaslyl, line 106: i remove white-space: nowrap as this code didn't affect any file based on the Percy snapshot result, and this line of code causes the button text to exceed the button box

tintinthong (approved):

PR #2253, opened 2025-03-07
host: Add test to exercise clientRequestId check

backspace:
This test makes a card update request that uses an existing `clientRequestId` to ensure that the resulting invalidation realm event is ignored, so the card is not refetched. The `clientRequestId` code was previously untested.

When I [commented out the `clientRequestId` check](https://github.com/cardstack/boxel/pull/2253/commits/016a5c55784aa7ab4204f41cd38d4e24b1c253d9), the [test failed](https://github.com/cardstack/boxel/actions/runs/13726409222/job/38393782233#step:12:216).


jurgenwerk (approved):
Nice to have a test for this

PR #2254, opened 2025-03-10
Fix pre-rendered spec card icon

FadhlanR:
This is the minimal solution to fix the bug where the pre-rendered spec card displays a loading state as its icon. In this PR, we use the spec card's own card type icon as the icon for the pre-rendered spec card.

https://github.com/user-attachments/assets/8717a4a4-17fa-4218-87a1-c7ab593a296c

### DIFF
--- a/packages/base/spec.gts
+++ b/packages/base/spec.gts
@@ -61,12 +61,16 @@ class Isolated extends Component<typeof Spec> {
       value: undefined,
     });
     (async () => {
-      if (this.args.model.ref && this.args.model.id) {
-        let card = await loadCard(this.args.model.ref, {
-          loader: myLoader(),
-          relativeTo: new URL(this.args.model.id),
-        });
-        icon.value = card.icon;
+      try {
+        if (this.args.model.ref && this.args.model.id) {
+          let card = await loadCard(this.args.model.ref, {
+            loader: myLoader(),
+            relativeTo: new URL(this.args.model.id),
+          });
+          icon.value = card.icon;
+        }
+      } catch (e) {
+        icon.value = undefined;
       }
     })();
     return icon;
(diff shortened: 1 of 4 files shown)


packages/base/spec.gts
```diff
@@ -40,13 +40,704 @@ import StackIcon from '@cardstack/boxel-icons/stack';
...
+import { tracked } from '@glimmer/tracking';
 
 export type SpecType = 'card' | 'field' | 'app' | 'skill';
 
 class SpecTypeField extends StringField {
   static displayName = 'Spec Type';
 }
 
+class Isolated extends Component<typeof Spec> {
+  @tracked icon: CardOrFieldTypeIcon | undefined;
```
FadhlanR: I had to move out the `Isolated`, `Fitted`, and `Edit` views to satisfy the type check since I use `@tracked`.

tintinthong (approved):

packages/base/spec.gts
```diff
@@ -40,13 +39,704 @@ import StackIcon from '@cardstack/boxel-icons/stack';
...
+
+  get defaultIcon() {
+    return this.args.model.constructor?.icon;
+  }
+  constructor(owner: any, args: any) {
+    super(owner, args);
+    this.loadCardIcon.perform();
+  }
+
+  private loadCardIcon = restartableTask(async () => {
```
lukemelia: Since this is only ever called once in the constructor, there is no reason for it to be a restartableTask. It should just be a task.

packages/base/spec.gts
```diff
@@ -40,13 +39,704 @@ import StackIcon from '@cardstack/boxel-icons/stack';
...
```
lukemelia: Same note

packages/base/spec.gts
```diff
@@ -40,13 +39,704 @@ import StackIcon from '@cardstack/boxel-icons/stack';
...
```
lukemelia: Same note

packages/base/spec.gts
lukemelia: On that topic, perhaps this task *should* be executed again if the `ref` changes?

lukemelia (changes requested):

packages/base/spec.gts
FadhlanR: I used resource instead, so it will be reload if the value of ref is updated.

packages/base/spec.gts
```diff
@@ -40,13 +38,729 @@ import StackIcon from '@cardstack/boxel-icons/stack';
...
+class Isolated extends Component<typeof Spec> {
+  get defaultIcon() {
+    return this.args.model.constructor?.icon;
+  }
+
+  get icon() {
+    return this.loadCardIcon.value;
+  }
+
+  @use private loadCardIcon = resource(() => {
```
lukemelia, line 59:
```suggestion
  @use private cardIconResource = resource(() => {
```

Better to name this as a noun than a verb

lukemelia (approved):

burieberry (approved):

PR #2256, opened 2025-03-10
CS-8091 on click handler on links to many editor card / spec example card 

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8091/create-a-component-with-modifier-to-target-cards-inside-of-spec

### What is changing
- add card context and modifier to handle selected card action in links to component editor
- add cardComponentModifier at LinksToManyStandardEditor

Screenshot

https://github.com/user-attachments/assets/94530f10-fdda-450c-b61f-7c2e0c8d633d

### DIFF
--- a/packages/base/links-to-many-component.gts
+++ b/packages/base/links-to-many-component.gts
@@ -203,12 +203,6 @@ class LinksToManyStandardEditor extends GlimmerComponent<LinksToManyStandardEdit
                 <button
                   type='button'
                   class='card-button'
-                  {{this.cardContext.cardComponentModifier
-                    card=boxedElement.value
-                    format='fitted'
-                    fieldType='linksToMany'
-                    fieldName=@field.name
-                  }}
                   {{on 'click' (fn this.viewCard boxedElement.value)}}
                 >
                   <Item @format='fitted' />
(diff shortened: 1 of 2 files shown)


tintinthong:
I looked at this. Can't the component modifier just sit on the Spec card then. CardsGrid which is in base has that pattern. I think since componentModifier is in the context we can call it from inside the Spec card

packages/base/links-to-many-component.gts
```diff
@@ -195,7 +200,13 @@ class LinksToManyStandardEditor extends GlimmerComponent<LinksToManyStandardEdit
...
                 as |Item|
               }}
-                <Item @format='fitted' />
+                <button
+                  type='button'
+                  class='card-button'
+                  {{on 'click' (fn this.viewCard boxedElement.value)}}
+                >
+                  <Item @format='fitted' />
+                </button>
```
lukemelia, line 209: It was not this way in the past because we conceived of this behavior as something that interact mode added to card rendering. Are we clear that we want to change that? This may have far-reaching impact.

richardhjtan: @lukemelia Agreed it will be a risk to change inside links to component itself, I will introduce another method in this PR: https://github.com/cardstack/boxel/pull/2276

richardhjtan:
Close this PR as we found another method to able target the card by using a new base overlay component with card tracker.

https://github.com/cardstack/boxel/pull/2276

PR #2258, opened 2025-03-10
Add job identity to indexing log messages

habdelra:
In order to support correlation of log messages to indexing jobs, we are adding job identification to the indexing log messages.


backspace (approved):
Are we getting into structured logging territory? But looks good as it is

habdelra:
> Are we getting into structured logging territory? But looks good as it is

perhaps. really it's just logs in current-run that needs some keywords that i can parse in grafana

PR #2263, opened 2025-03-11
Auto attaches selected card in playground to message

FadhlanR:
https://github.com/user-attachments/assets/75621a0e-cfc9-409f-99f2-6455f7c27e08

### DIFF
--- a/packages/host/app/components/matrix/room.gts
+++ b/packages/host/app/components/matrix/room.gts
@@ -23,13 +23,14 @@ import { MatrixEvent } from 'matrix-js-sdk';
 
 import pluralize from 'pluralize';
 
-import { TrackedObject } from 'tracked-built-ins';
+import { TrackedObject, TrackedSet } from 'tracked-built-ins';
 
 import { v4 as uuidv4 } from 'uuid';
 
 import { BoxelButton } from '@cardstack/boxel-ui/components';
 import { not } from '@cardstack/boxel-ui/helpers';
 
+import { ResolvedCodeRef, internalKeyFor } from '@cardstack/runtime-common';
 import { DEFAULT_LLM_LIST } from '@cardstack/runtime-common/matrix-constants';
 
 import AddSkillsToRoomCommand from '@cardstack/host/commands/add-skills-to-room';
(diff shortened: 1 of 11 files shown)


packages/host/tests/helpers/index.gts
```diff
@@ -736,3 +736,80 @@ export function setupRealmServerEndpoints(
...
+        .dom(`[data-test-message-idx="${index}"] [data-test-attached-card]`)
+        .exists({ count: cards.length });
+      cards.map(async (card) => {
+        if (card.title) {
+          if (message != null && card.title.includes(message)) {
+            throw new Error(
+              `This is not a good test since the message '${message}' overlaps with the asserted card text '${card.title}'`,
+            );
+          }
+          // note: attached cards are in atom format (which display the title by default)
```
jurgenwerk: Not sure this comment is true. in CardPill there is currently no delegated render, it just renders its title directly.

jurgenwerk (commented):
"topmost card is shared automatically" tooltip sounds a bit odd when there is no stack (i.e. we are in code mode). Perhaps this should be changed to "Current card is shared automatically"

packages/host/tests/helpers/index.gts
FadhlanR: Correct, I removed the comment.

FadhlanR:
> "topmost card is shared automatically" tooltip sounds a bit odd when there is no stack (i.e. we are in code mode). Perhaps this should be changed to "Current card is shared automatically"

Updated.
Screenshot 2025-03-11 at 21 58 06

packages/host/app/components/matrix/room.gts
```diff
@@ -315,6 +326,40 @@ export default class Room extends Component<Signature> {
...
+      let cardResource = getCard(this, () => {
+        if (!this.args.selectedCardRef) {
+          return;
+        }
+        let moduleId = internalKeyFor(this.args.selectedCardRef, undefined);
+        return this.playgroundPanelService.getSelection(moduleId)?.cardId;
+      });
+      await cardResource.loaded;
+      state.value = cardResource.card;
+    })();
```
lukemelia, line 350: @ef4 please review this tomorrow for resource patterns

lukemelia: (not a blocker to merging)

lukemelia (approved):

packages/host/app/components/matrix/room.gts
```diff
@@ -144,6 +149,11 @@ export default class Room extends Component<Signature> {
                 @submode={{this.operatorModeStateService.state.submode}}
                 @autoAttachedFile={{this.autoAttachedFile}}
                 @filesToAttach={{this.filesToAttach}}
+                @autoAttachedCardTooltipMessage={{if
```
jurgenwerk, line 152: It would be clearer if both strings, "Current card is shared automatically", and "Topmost card is shared automatically" would be in the same logical construct, so that you don't have to check two different components to see which text would be rendered. For example:

`@autoAttachedCardTooltipMessage={{if this.isCodeMode 'Current card is shared automatically' 'Topmost card is shared automatically'}}`

jurgenwerk (approved):

PR #2268, opened 2025-03-11
Add "Show Card" command to "Card Editing" skill

lukemelia:


backspace (approved):

packages/host/app/services/command-service.ts
```diff
@@ -226,6 +210,44 @@ export default class CommandService extends Service {
...
+              adoptsFrom,
+            },
+            attributes: attributes ?? {},
+            relationships: relationships ?? {},
+          },
+        };
+        typedInput = await this.cardService.createFromSerialized(
+          inputDoc.data,
+          inputDoc,
+        );
```
ef4, line 240: I think `createFromSerialized` is bad because we're treating the storage/network format of the cards as their public API. In this case that leaks even further into the shape of the `attributes` and `relationships` arguments.

When you're trying to create a new instance, it would be better if we did something like:

```js
// this is existing api that is badly named. It doesn't load card instances, it loads card or field definitions
import { loadCard } from '@cardstack/runtime-common';

let Klass = await loadCard(adoptsFrom)
let instance = new Klass(properties);
await this.cardService.saveCard(instance);
```

In this case the properties are not serialized, they're the user-facing values. If some are cards, they are already card instances. If you need to store those somewhere, we can still have api for serialize/deserialize of existing instances and that doesn't leak the actual serialization format into the public API.

At the moment, you need to be careful using the card's constructor directly because it doesn't force you to properly setup the context the card will need if it attempts to load a relationship and ensure the related cards are identity mapped. (It's fine that the new card instance itself isn't identity mapped because it's new -- by definition. It can go into the identity map at `saveCard()`.)

I think there's a lot of benefit in making cards function as normal javascript values, and then enhancing them with more context only when needed. In this example, whether more context would be needed depends on the real types of `attributes` and `relationships`.

lukemelia: In this case, I'm trying to create a card instance for use as command input from JSON with attributes and relationships serialized by the AI according to the JSON schema we gave it. Using the constructor doesn't work for the relationships because it excepts card instances not json pointing to a card instance.

PR #2276, opened 2025-03-12
CS-8091 spec example card with card tracker overlay 

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8091/create-a-component-with-modifier-to-target-cards-inside-of-spec

**Objective**
We want the spec example cards ( which is the `linksToMany` field ) to be clickable without changing their structure. To do that, we need to create an overlay that listens to the hovered location and performs custom actions at the overlay level. This method can eliminate the risk of changing code in the links to many editor.

### What is changing
- Create a base overlay component that exists in operator mode overlay - it is useful in this case and reusable in future
- Refactor operator mode overlay extends to the base overlay component
- Implement base overlay to spec preview - so we can get on click card details that happening at links to many field editor

**Screenshot**

https://github.com/user-attachments/assets/1cc2ec15-f76f-428b-bb2f-4aa371c209e4


packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -200,6 +220,34 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
...
+
+  private get renderedCardsForOverlayActions(): RenderedCardForOverlayActions[] {
+    return this.cardTracker.elements
+      .filter((entry) => {
+        return (
+          entry.meta.format === 'data' ||
+          entry.meta.fieldType === 'linksTo' ||
+          entry.meta.fieldType === 'linksToMany'
+        );
+      })
```
tintinthong: we only want to narrow this down to the field `linkedExamples` right. If I introduce a field with a link, is the expectation that it will open in playground as well.

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -200,6 +220,34 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
...
+      .map((entry) => ({
+        element: entry.element,
+        cardDefOrId: entry.meta.card || entry.meta.cardId!,
+        fieldType: entry.meta.fieldType,
+        fieldName: entry.meta.fieldName,
+        format: entry.meta.format,
+        overlayZIndexStyle: htmlSafe(`z-index: 100`),
```
tintinthong: I thought the ZIndex is already derived in the overlay component

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -60,8 +61,19 @@ import type RealmServerService from '@cardstack/host/services/realm-server';
...
+  CardDef,
+  Format,
+  FieldType,
+} from 'https://cardstack.com/base/card-api';
 import { Spec, type SpecType } from 'https://cardstack.com/base/spec';
 
+import ElementTracker from '../../../resources/element-tracker';
+import Overlays, { RenderedCardForOverlayActions } from '../overlays';
+
+import { CardDefOrId } from '../stack-item';
```
tintinthong: probably don't create this dependency

packages/host/app/components/operator-mode/operator-mode-overlays.gts
```diff
@@ -0,0 +1,445 @@
...
+
+import { removeFileExtension } from '../search-sheet/utils';
+
+import Overlays from './overlays';
+import { StackItemRenderedCardForOverlayActions } from './stack-item';
+
+import type { CardDefOrId } from './stack-item';
+
+export default class OperatorModeOverlays extends Overlays {
+  @tracked overlayClassName = this.args.overlayClassName ?? 'actions-overlay';
```
tintinthong: this shud be hard-coded right. You don't need to have an overlayClassName here

tintinthong:
Perhaps shud add some test here. One thing that we shud test is if we re-order a links-to-many, will the correct action get triggered  (with the correct card as the argument)

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
richardhjtan: You are right @tintinthong , I will adjust to filter out by `linksToMany` - it's in latest changes

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
richardhjtan: @tintinthong the z index `auto` (derived from overlay helpers) is not working, so I pass a custom z index here. Try to put 1 for now, not necessarily to be 100

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
richardhjtan: Updated

packages/host/app/components/operator-mode/operator-mode-overlays.gts
richardhjtan: Updated in latest commit

richardhjtan:
> Perhaps shud add some test here. One thing that we shud test is if we re-order a links-to-many, will the correct action get triggered (with the correct card as the argument)

@tintinthong yes I added a test where overlay should exist when mouse enter or mouse leave. Manually tested the second it works but we can add the test when we implement the redirection to playground

richardhjtan:
Changes are made in latest commit and updated the screenshot on behaviour & testing

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -188,6 +198,14 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
     this.initializeCardSelection();
   }
 
+  cardTracker = new ElementTracker<{
+    cardId?: string;
+    card?: CardDef;
+    format: Format | 'data';
+    fieldType: FieldType | undefined;
+    fieldName: string | undefined;
```
lukemelia: Let's make this a named type instead of copy-pasting it from stack-item

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -200,6 +218,30 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
     return this.args.cards.map((card) => card.url);
   }
 
+  private get cardContext(): Omit<
```
lukemelia: consider being more explicit:

```ts
type SpecPreviewCardContext = Omit<CardContext, 'prerenderedSearchComponent'>;
```

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -200,6 +218,30 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
...
+        return entry.meta.fieldType === 'linksToMany';
+      })
+      .map((entry) => ({
+        element: entry.element,
+        cardDefOrId: entry.meta.card || entry.meta.cardId!,
+        fieldType: entry.meta.fieldType,
+        fieldName: entry.meta.fieldName,
+        format: entry.meta.format,
+        overlayZIndexStyle: htmlSafe(`z-index: 1`),
+      }));
```
lukemelia, line 229: Rather than duplicating this code already present in stack-item, consider how we can re-use it. e.g., maybe we should enhance ElementTracker with a method that can filter its elements based on a predicate and map them to an array of `RenderedCardForOverlayActions`

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -236,6 +278,11 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
     this.operatorModeStateService.updateCodePath(selectedUrl);
   }
 
+  @action viewCardInPlayground(card: CardDef | string) {
+    // TODO: Implement redirect to playground
```
lukemelia: Is this supposed to be left in?

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
richardhjtan: @lukemelia I make it as a helper to be reusable at the element tracker resource level: https://github.com/cardstack/boxel/pull/2276/commits/e17f6ede9eab5a0a99fe8501c380eeafe460bf5d

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
richardhjtan: Ah I will clear that out, just a reminder to be work on after this PR. Thanks!

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
richardhjtan: Refactored the same at stack item as well

lukemelia (approved):
👍  Nice improvements

PR #2277, opened 2025-03-12
Ability to apply code patches received by the AI assistant for card definitions 

jurgenwerk:
This is a minimal implementation for applying card implementation changes proposed by the AI assistant. You can ask the assistant to change the implementation, and it will leverage the following skills: 

- `boxel-coding` skill (the anatomy of a gts file)  
- `source-code-editing` skill (responding with a code patch in a way that our system understands and knows how to apply - it's a common search/replace notation)

Here is how this works: 

https://github.com/user-attachments/assets/f68d66ed-6843-4149-99eb-334e60ee8519

TODO (in other tickets):

- [ ] Code should be streamed directly into the monaco editor (without first rendering it in `<pre>` tags), and the experience should look good (without flashing/rerendering of the editor)
- [ ] Enable persistence of the applied code patch action status (use existing matrix<->command setup but without tool calls)
- [ ] Make code diffing human readable (don't show search/replace notation directly)


packages/host/app/components/ai-assistant/formatted-message.gts
```diff
@@ -1,18 +1,98 @@
...
+
+      await this.cardService.saveSource(
+        new URL(codeAction.fileUrl),
+        patchedCode,
+      );
+      this.loaderService.reset();
+
+      codeAction.state = 'applied';
+    } catch (error) {
+      console.error(error);
```
jurgenwerk, line 91: It's currently not clear what kind of errors we can anticipate here, so until we have a better idea we'll just log it to the console.

packages/host/app/components/ai-assistant/formatted-message.gts
```diff
@@ -1,18 +1,98 @@
...
 
+import ApplyButton from '../ai-assistant/apply-button';
+
 interface FormattedMessageSignature {
   sanitizedHtml: SafeString;
   monacoSDK: MonacoSDK;
   renderCodeBlocks: boolean;
 }
 
+interface CodeAction {
```
lukemelia, line 30: Perhaps we should consider making this a full-fledged CodeAction class.

If we did and we made it container-aware, it could have all the service dependencies and the task defined on it and keep FormattedMessage as more of a UI component.

packages/host/app/modifiers/code-block.ts
```diff
@@ -142,6 +169,16 @@ export default class CodeBlock extends Modifier<Signature> {
   }
 }
 
+function makeCodeActionsContainerDiv() {
+  let template = document.createElement('template');
+  let randomId = Math.random().toString(36).substring(2, 15);
```
lukemelia, line 174: Maybe we could use `guidFor(template)` instead?

lukemelia (approved):

packages/host/app/modifiers/code-block.ts
jurgenwerk: Yep, I'll use that instead.

packages/host/app/components/ai-assistant/formatted-message.gts
jurgenwerk: Good idea, will try something like that. On the other hand I'm also working on [CS-8124](https://linear.app/cardstack/issue/CS-8124/stream-code-directly-into-monaco-code-editor-fix-flashing-during) which might result in a refactor around this area so I'll keep this comment open until I figure that out

jurgenwerk:
I'll go ahead and merge this, and fix the suggested things in my next PR that will introduce a refactor in order to fix visual bugs during code streaming

PR #2278, opened 2025-03-12
SpecPreview was not changing based upon choosing new declarations

tintinthong:
The point of this PR is to get a stable auto-save while also allowing spec to react when navigating in LHS panel

For a small change, this PR was necessary bcos reactivity was hidden in many places


packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -435,11 +425,20 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
           },
         },
       };
-      await this.cardResource.loaded;
+      try {
+        let card = await this.cardService.createFromSerialized(doc.data, doc);
```
tintinthong: I know this is a sin. But, working with pre-rendered card search was problematic so I needed to work with `getCards`. (given there are not SO MANY specs I was ok to load the cards all at once)

Furthermore, it was useful to have `getCards` and not both (getCard and getCards) together. I confirm bcos `getCards` calls `getCard`, it does instantiate an identity context of the card

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -183,11 +177,6 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
   @service private declare realm: RealmService;
   @service private declare operatorModeStateService: OperatorModeStateService;
 
-  constructor(owner: Owner, args: ContentSignature['Args']) {
-    super(owner, args);
-    this.initializeCardSelection();
-  }
```
tintinthong, line 201: Here is the one of the issues. Is that the selection was made at constructor (by id). And the selection never re-rendered after that unless you interacted with the dropdown. 

This cause things like the LHS panel selection to never work

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -395,8 +386,7 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
   @service private declare realmServer: RealmServerService;
   @service private declare cardService: CardService;
   @service private declare playgroundPanelService: PlaygroundPanelService;
-  @tracked private _selectedCardId?: string;
-  @tracked private newCardJSON: LooseSingleCardDocument | undefined;
+  @tracked private _selectedCard?: Spec;
```
tintinthong, line 389: This was one of the saving grace. Using a getter that included lines like this always caused a re-render. By assinging at the unit of the card, we were able to keep th edit of the card stable

```
this.cards.find((card) => card.id === card.id);
```

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -600,49 +602,42 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
-              onSelectCard=this.onSelectCard
-              selectedId=this._selectedCardId
-              spec=this.card
-              isLoading=false
-              cards=cards
-            )
-          }}
-        {{/let}}
-      </:response>
-      <:loading>
```
tintinthong, line 678: The pre-rendered search api is not javascript driven. So, occasionally we wud have workarounds that made is bad. Particularly, assigning cardIds within the `:response` block. This caused more re-renders

tintinthong:
Update here. This PR is most likely the correct shape 

Few things to point out 
- persistence to local storage shud be confined to an action. But in the case where we default to the 1st selection, we should persist to local storage via a modifier. This is to keep the reaactivity system (of resources) separate from action-based persistence. In future, we shud local storage is not a great place to put cardSelections/cards, it should be in the url. Action: make that modifier and persist selected spec id so playground knows about it.
- createFromSerialized is a sucky pattern for now (its an old pattern that was previously discourage).  However, we confirm a creation will still hit the identity map. . In future, creating cards shud be simply new CARD and saveModel(CARDINSTANCE). Action: attempt to loadCard and call new and call save. If it dont work, We should write a comment about why its used
- getSearch returning a resource is fine for now but it has the quirkiness of returning card resources. In future, it should return instances. Things like auto-save should be handled inside getSearch (and other interactions with the store) similar to how getCard does it but in its own implementation. Action: No action.

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -211,7 +201,7 @@ class SpecPreviewContent extends GlimmerComponent<ContentSignature> {
   }
```
richardhjtan: We can remove the getter too `shouldSelectFirstCard`

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -463,29 +456,28 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
+      try {
+        let Klass = await loadCard(specRef, {
+          loader: this.loaderService.loader,
+        });
+        let card = new Klass({
+          specType,
+          ref,
+          title: ref.name,
+        }) as CardDef;
+        await this.cardService.saveModel(card);
```
tintinthong, line 471: new create card pattern

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -745,3 +703,54 @@ function getRelativePath(baseUrl: string, targetUrl: string) {
...
+    { id, codeRef }: ModifierSignature['Args']['Named'],
+  ) {
+    if (!id || !codeRef) {
+      throw new Error('bug: no id or codeRef');
+    }
+    let maybeAbsoluteRef = codeRefWithAbsoluteURL(codeRef, new URL(id));
+    if (!isResolvedCodeRef(maybeAbsoluteRef)) {
+      throw new Error('bug: ref in spec cannot be resolved to an absolute');
+    }
+    this.updateFieldSpecForPlayground(id, maybeAbsoluteRef);
```
tintinthong, line 766: modifier calls this. We dont even rely on actions anymore

packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -568,6 +630,42 @@ module('Acceptance | Spec preview', function (hooks) {
       .exists();
   });
 
+  test('navigation: spec preview updates when changing between different declarations inside inspector', async function (assert) {
```
tintinthong, line 633: this test checks for navigation by lhs. But also, deals with if you interacted with the dropdown

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -583,112 +575,74 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
+        if (
+          this._selectedCard?.ref.name ===
+          this.getSelectedDeclarationAsCodeRef.name
+        ) {
+          return this._selectedCard;
+        }
+      }
     }
-    return this.cardResource.card as Spec;
+    return this.cards?.[0] as Spec;
```
tintinthong, line 613: bcos we have to consolidate card with selection state (inside of a reactive usage with resources), we need to be smart to return the _selectedCard only when it matches with the selected code ref

tintinthong: It seems this is problematic for auto-save. 

Everything behaves well when we just return `this._selectedCard` but if you entangle with `getSelectedDeclarationAsCodeRef`the auto-save gets triggered everytime you type BUT only if you refer to `this._selectedCard`.

tintinthong:
I am closing this PR and rebasing the work on @habdelra https://github.com/cardstack/boxel/pull/2288. Wud be good if that goes in first

PR #2280, opened 2025-03-12
host: Remove flaky test for now

backspace:
This test has been [failing](https://github.com/cardstack/boxel/runs/38666936965), it may be related to the SSE removal, investigation pending.


lukemelia:
I tried making it more robust here: https://github.com/cardstack/boxel/pull/2273/files#diff-d27e368e108d1b5af519c4f070191f1ebf5e53cbd3683c2e12c7b9e6fc0e7b92L1448-R1463

PR #2281, opened 2025-03-12
Refactor playground tests to use helpers; make fitted field preview available on playground

burieberry:


richardhjtan (approved):
The refactor looks great

lukemelia (approved):

PR #2282, opened 2025-03-13
QUERY alternative to GET for _search endpoint

lukemelia:


burieberry (approved):
Nice

backspace (approved):
I couldn’t figure out how to exercise this via `host` as all my queries were using `_search-prerendered` but I was able to try it out via `curl` 🎉

PR #2283, opened 2025-03-13
Add support for using QUERY HTTP method against the _search-prerendered endpoint

lukemelia:
Targets https://github.com/cardstack/boxel/pull/2282


burieberry (commented):
The code looked fine, however, I get a CORS error when trying to run the app on localhost:4200
4200-cors

400 error when trying to run on localhost:4201
4201-400-err

lukemelia:
Thanks. Seems like I need to account differently for requests that route directly to test realm and those that go over http.

burieberry (approved):

PR #2288, opened 2025-03-14
Refactor to use dynamic context for getCard()

habdelra:
As part of our refactor for `getCard()` we are updating our use of this to use dynamic context. This means that from the host app you should use the following to obtain `getCard` instead of importing it from `card-resource.ts`.

```ts
import { consume } from 'ember-provide-consume-context';
import { type getCard, GetCardContextName } from '@cardstack/runtime-common';
  ...
class YourComponent extends Component {
  @consume(GetCardContextName) private declare getCard: getCard;
```
If you need to access getCard in the constructor it's a bit more tedious as `@consume` doesn't run before the Component's constructor is invoked. Rather its available after the DOM has been created for the component. As such, I made a modifier to make that a little easier. In this pattern you would establish `getCard` like this (extending the example above):

```ts
import { consume } from 'ember-provide-consume-context';
import { type getCard, GetCardContextName } from '@cardstack/runtime-common';
import consumeContext from '@cardstack/host/modifiers/consume-context';
  ...
class YourComponent extends Component {
  @consume(GetCardContextName) private declare getCard: getCard;
  @tracked private cardResource: ReturnType<getCard> | undefined; // the @tracked decorator is important here!!
  private makeCardResource = () => {
    this.cardResource = this.getCard(this, () => this.cardURL);
  };

  <template>
    <div {{consumeContext consume=this.makeCardResource}}>...</div>
  </template>
``` 

From a card you can access `getCard()` from the `@context`--no need to import it.

If you need to use `getCard()` in something that is not a Component, then let's talk. All the current non-component usages of `getCard` are slated for refactoring. Note that the refactor for `getCards()` is coming in the next PR.


packages/base/card-api.gts
```diff
@@ -145,6 +147,8 @@ export interface CardContext {
     };
   }>;
   prerenderedCardSearchComponent: any;
+  getCard: getCard<T>;
+  getCards: getCards;
```
tintinthong, line 151: There is additionally a getCards inside of actions. Whats the distinction between getCards here and getCards actions? 

Thanks for adding tho, I always thought not having `getCard` was weird

packages/host/app/components/operator-mode/code-submode/playground-panel.gts
```diff
@@ -484,33 +493,45 @@ class PlaygroundPanelContent extends Component<PlaygroundContentSignature> {
...
   @service private declare operatorModeStateService: OperatorModeStateService;
   @service private declare realm: RealmService;
   @service private declare realmServer: RealmServerService;
   @service private declare recentFilesService: RecentFilesService;
   @service private declare playgroundPanelService: PlaygroundPanelService;
-  @tracked newCardJSON: LooseSingleCardDocument | undefined;
-  @tracked fieldChooserIsOpen = false;
+  @tracked private newCardJSON: LooseSingleCardDocument | undefined;
+  @tracked private fieldChooserIsOpen = false;
+  @tracked private cardResource: ReturnType<getCard> | undefined;
```
tintinthong, line 507: So this is @tracked becos we are dynamically providing the context? is that correct?

tintinthong:
> If you need to use getCard() in something that is not a Component, then let's talk. All the current non-component usages of getCard are slated for refactoring. Note that the refactor for getCards() is coming in the next PR.

I feel that if this is the case, then we shud definitely put some README to say when this is actually used. Like if I want to get ids.. I will reach for `getCards`, otherwise, worse maybe pre-rendered search. Not important for this PR since it touches a lot of things and not a good idea to hang out. But wud be good to have a ticket to write docs about when to use it. Both addressing usage for card user and card developer; everything in the PR description is good too!

tintinthong (approved):

packages/base/card-api.gts
habdelra: the action for getCards will be going away in the next PR

packages/host/app/components/operator-mode/code-submode/playground-panel.gts
habdelra: no. it's `@tracked` because the resource is not set in the constructor. so we need to tell glimmer that it needs to pay attention to this property when it's consumed as it will no longer be undefined as soon as the DOM renders (when the resource is defined in teh constructor you don't need to make it tracked--this is just a general thing for resources when they are assigned outside of the constructor)

habdelra:
> > If you need to use getCard() in something that is not a Component, then let's talk. All the current non-component usages of getCard are slated for refactoring. Note that the refactor for getCards() is coming in the next PR.
> 
> I feel that if this is the case, then we shud definitely put some README to say when this is actually used. Like if I want to get ids.. I will reach for `getCards`, otherwise, worse maybe pre-rendered search. Not important for this PR since it touches a lot of things and not a good idea to hang out. But wud be good to have a ticket to write docs about when to use it. Both addressing usage for card user and card developer; everything in the PR description is good too!

there are about a dozen follow up tickets that will remove the `CardResource.loaded` property. eliminating that will make it pretty much impossible to consume this resource outside of a component. so the API will very much guide you in the right direction.

PR #2290, opened 2025-03-17
Improve boot time by implementing sliding sync

FadhlanR:
In this PR, I implemented sliding sync to improve boot time. The latest released version of `matrix-js-sdk` does not yet support the `simplified_msc3575` endpoint, so I had to patch the SDK. The [PR](https://github.com/matrix-org/matrix-js-sdk/pull/4400) to add support for `simplified_msc3575` in `matrix-js-sdk` was merged two days ago, so we can update the SDK version once it is released.  

The implementation introduces two sliding sync lists:  
1. One for filtering AI rooms.  
2. One for filtering authentication rooms.  

Initially, we only request **one** timeline event and will load all timeline events when the room is opened. The range for AI rooms increases when a user opens the past session list, and the range for authentication rooms increases when there is an update to the realms state event in account data.  

I haven't yet refactored to introduce separate `RoomResource` and `RoomStateResource` components, as we discussed, @lukemelia. If needed, we can create a ticket for that.  

I've tested this in staging, and it improves the AI panel's loading speed.


packages/host/app/components/ai-assistant/past-session-item.gts
```diff
@@ -209,7 +209,12 @@ export default class PastSessionItem extends Component<Signature> {
     if (!this.args.session.lastMessage) {
       return false;
     }
-    return !this.args.session.lastMessage.isStreamingFinished;
+
+    return (
+      this.args.session.lastMessage.author.userId !==
+        this.matrixService.userId &&
+      !this.args.session.lastMessage.isStreamingFinished
+    );
```
lukemelia: Would this work?

```suggestion
    return this.args.session.lastMessage.isStreamingFinished === false;
```

packages/host/app/services/matrix-service.ts
```diff
@@ -1547,6 +1664,61 @@ export default class MatrixService extends Service {
...
+        [0, newEndRange],
+      ]);
+    } finally {
+      this.isLoadingMoreAIRooms = false;
+    }
+  });
+
+  get loadingMoreAIRooms() {
+    return this.isLoadingMoreAIRooms;
+  }
```
lukemelia: Let's just make the member public. This getter doesn't add anything and the naming is confusing.

packages/host/app/services/matrix-service.ts
```diff
@@ -509,6 +529,66 @@ export default class MatrixService extends Service {
...
+  private async initSlidingSync() {
+    let accountData = await this.client.getAccountDataFromServer<{
+      realms: string[];
+    }>(APP_BOXEL_REALMS_EVENT_TYPE);
+
+    let lists: Map<string, MSC3575List> = new Map();
+    lists.set(SLIDING_SYNC_AI_ROOM_LIST_NAME, {
+      ranges: [[0, SLIDING_SYNC_LIST_RANGE_END]],
+      filters: {
+        is_dm: false,
```
lukemelia, line 541: Currently, this is adequate, since we only have AI rooms and Auth rooms, but generally we have been considering a room an AI room if the aibot it one if it's members. Is it possible to use that criteria here?

packages/host/app/services/matrix-service.ts
```diff
@@ -1547,6 +1664,61 @@ export default class MatrixService extends Service {
     await roomResource.loading;
     roomResource.activateLLM(model);
   }
+
+  loadMoreAIRooms() {
+    this.loadMoreAIRoomsTask.perform();
+  }
+
+  private loadMoreAIRoomsTask = restartableTask(async () => {
+    if (!this.slidingSync) return;
```
lukemelia: Is there a circumstance where this can happen?

packages/host/tests/helpers/mock-matrix/_client.ts
```diff
@@ -600,6 +629,71 @@ export class MockClient implements ExtendedClient {
...
+    let lists: MSC3575SlidingSyncResponse['lists'] = {};
+    let rooms: MSC3575SlidingSyncResponse['rooms'] = {};
+    Object.entries(req.lists || {}).forEach(([listKey, list]) => {
+      list.ranges.forEach((range) => {
+        let [start, end] = range;
+        //currently we only filter rooms using is_dm
+        let dmRooms =
+          this.getAccountData('m.direct')?.getContent()?.rooms ?? [];
+        let roomsInRange = this.serverState.rooms
+          .filter((r) => r.id.includes('mock'))
```
lukemelia: Why do we need this filter?

packages/host/app/services/matrix-sdk-loader.ts
```diff
@@ -92,7 +99,6 @@ export type ExtendedClient = Pick<
   | 'sendEvent'
   | 'sendReadReceipt'
   | 'sendStateEvent'
-  | 'setAccountData'
```
lukemelia: Why is this being removed?

lukemelia (changes requested):

packages/host/app/services/matrix-service.ts
```diff
@@ -509,6 +529,66 @@ export default class MatrixService extends Service {
...
+          0,
+          accountData
+            ? accountData?.realms.length
+            : SLIDING_SYNC_LIST_RANGE_END,
+        ],
+      ],
+      filters: {
+        is_dm: true,
+      },
+      timeline_limit: 1,
```
tintinthong: SLIDING_SYNC_LIST_TIMELINE_LIMIT?

packages/host/app/services/matrix-service.ts
FadhlanR: I updated to throw an error

packages/host/app/services/matrix-service.ts
FadhlanR: I renamed the field and getter since We don't want the value of the field can be updated from outside this class.

packages/host/app/services/matrix-service.ts
FadhlanR: Unfortunately, the filter options are very limited, as you can see [here](https://github.com/matrix-org/matrix-spec-proposals/blob/kegan/sync-v3/proposals/3575-sync.md?plain=1#L418-L478). We could use the room_types filter if we defined the room type during room creation, but we don't.

lukemelia (approved):

PR #2292, opened 2025-03-17
AI assistant: stream code directly into the monaco editor + code block refactor

jurgenwerk:
This is a significant refactor of how we display code blocks that the AI assistant sends to the user. The motivation for this was to fix visual annoyances during code streaming into the monaco editor. The reason for these visual bugs was that we were rerendering the AI message component on every code streaming chunk, which caused flickering/flashing during code streaming. Fixing component instabilites alone wasn't enough to get rid of the flickering–additional code delta calculation needed to be added for appending code chunks in monaco editors.

This PR:
- Introduces a new AI assistant message data structure which allows us to encapsulate streamed code, and have a stable code block component that reveals newly received code chunks in a visually smooth manner
- Introduces a new set of composable components for code editor and code actions (currently we have the "copy code", "apply code patch" action, but judging by designs it looks like we'll have more in the future, such as "undo")
- Overall gets rid of all places where we were manipulating the DOM to insert code action buttons, and removing using localstorage for passing code around

**Before** (see how code is streamed into text, then transformed into a monaco editor - we wanted to change this to stream directly into a monaco editor):

https://github.com/user-attachments/assets/934f0d61-2c4f-49e8-a078-6df29891e928

**Before the refactor** (see how it flickers during code streaming):

https://github.com/user-attachments/assets/206acf8b-25cf-46c2-a62f-67c31a6f5e53

**After refactor** (no flickering, smooth streaming, automatic scrolling as it streams):

https://github.com/user-attachments/assets/aa7debc8-ee6e-485b-9465-a2bc64837dfe


burieberry (approved):
This looks good. Did you account for what's causing the empty space under the code block in the message component in percy diffs?

packages/host/app/components/ai-assistant/code-block.gts
```diff
@@ -0,0 +1,358 @@
...
+    if (this.monacoState) {
+      let { editor } = this.monacoState;
+      let model = editor.getModel()!;
+
+      // Here we are appending deltas when code is "streaming" in, which is
+      // useful when code changes frequently in short periods of time. In this
+      // case we calculate the delta of the new code and the current code, and
+      // then apply that delta to the model. Compared to calling setValue()
+      // for every new value, this removes the need for re-tokenizing the code
+      // which is expensive and produces visual annoyances such as flickering.
```
habdelra, line 133: nice!

habdelra (approved):
Looks really good :+1:

lukemelia (approved):
Code looks good. Please think about tests. What behavior would you want to know about if someone broke accidentally? Write unit, integration and/or acceptance tests for that.

jurgenwerk:
@burieberry not sure–I'll look into that today.

@lukemelia these features are mostly already quite nicely covered with tests and I was relying on them while doing the refactor. I'll check again today to see if there's an opportunity to improve the coverage.

jurgenwerk:
I added a couple of tests specifically for `FormattedMessage` component. Fortunately this uncovered a bug with the "Apply" button which I fixed 👌

PR #2296, opened 2025-03-17
Add feature flag to revert to SSE for realm events

backspace:
This adds a `DISABLE_MATRIX_REALM_EVENTS` feature flag for both `host` and `realm-server` that switches back to using SSE for realm events while we address data being trampled. It’s only set for use in the production environment, where it’s currently deployed.

You can see the SSE coming in here when I edit a file in another tab, and eventually the card reloads:

screencast 2025-03-17 14-17-45

### DIFF
--- a/packages/host/config/environment.js
+++ b/packages/host/config/environment.js
@@ -53,7 +53,6 @@ module.exports = function (environment) {
     // ENV.APP.LOG_TRANSITIONS = true;
     // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
-    ENV.featureFlags = {};
   }
 
   if (environment === 'test') {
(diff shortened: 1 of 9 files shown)


packages/host/config/environment.js
```diff
@@ -50,7 +53,6 @@ module.exports = function (environment) {
     // ENV.APP.LOG_TRANSITIONS = true;
     // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
-    ENV.featureFlags = {};
```
backspace, line 53: This was making it so I couldn’t use SSE in development, it doesn’t seem correct to me to erase flags set via environment variables.

jurgenwerk (approved):

lukemelia (approved):

PR #2298, opened 2025-03-17
Refactor SearchResource to use Store Service

habdelra:
This PR refactors the SearchResource to internally use the StoreService to instantiate the search results instead of using CardResource.

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -1,4 +1,5 @@
 import { registerDestructor } from '@ember/destroyable';
+import { getOwner } from '@ember/owner';
 
 import { service } from '@ember/service';
 import { buildWaiter } from '@ember/test-waiters';
@@ -8,26 +9,22 @@ import { restartableTask } from 'ember-concurrency';
 import { Resource } from 'ember-resources';
 import flatMap from 'lodash/flatMap';
 
-import { TrackedMap } from 'tracked-built-ins';
+import { TrackedArray } from 'tracked-built-ins';
 
 import {
   subscribeToRealm,
   isCardCollectionDocument,
-  SingleCardDocument,
 } from '@cardstack/runtime-common';
 
 import type { Query } from '@cardstack/runtime-common/query';
 
-import type { CardDef } from 'https://cardstack.com/base/card-api';
+import { CardDef } from 'https://cardstack.com/base/card-api';
 
 import type { RealmEventContent } from 'https://cardstack.com/base/matrix-event';
 
-import { asURL } from '../services/store';
-
-import { type CardResource, getCard } from './card-resource';
-
 import type CardService from '../services/card-service';
 import type RealmServerService from '../services/realm-server';
+import type StoreService from '../services/store';
 
 const waiter = buildWaiter('search-resource:search-waiter');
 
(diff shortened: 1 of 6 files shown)


packages/host/app/resources/search.ts
```diff
@@ -191,35 +163,29 @@ export class Search extends Resource<Args> {
...
-      for (let [url, resource] of resultMap) {
-        if (!this.currentResults.has(url)) {
-          this.currentResults.set(url, resource);
-        }
-      }
+      this._instances.splice(
+        0,
+        this._instances.length,
+        ...results.map(({ card }) => card),
+      );
```
tintinthong, line 191: so here the whole entire instance array is cleared? 

Can the card know its the same after coming back from the server? Like checking with the identity context? So, it only mutates the part that has changed (assuming no new cards). Now it looks that if I invalidate a single card, my whole list of cards which is rendered will be replaced.

tintinthong: Cud it be beneficial to have a inPlaceEdit method that is only used internally?

habdelra: Yeah, I think I can add something that is a little more nuanced. The challenge would be to remove and insert cards in the array in the correct order. You might still have the same item in the results after the search, but now it’s the 7th result and not the 2nd result—so it’s still plucking it and re-adding it, but in a different position . My hope would be that glimmer can do the bookkeeping to see what has changed and only rerender the delta since this is a synchronous operation.  When you say that you invalidate a single card the whole list is rerendered , are you actually seeing that or are you guessing that that is what is happening?

habdelra: Also, consider in the template when you actually render the results, if you use an @key param in the {{each}} it should only rerender the difference. This is how the card grid works

habdelra: Although since we are using an identity map, @key is probably superfluous since we are not creating new instance. 

2 things to keep in mind:
1. The array is being mutated not replaced
2. The items in the array are from an identity map so we are never recreating an instance that already exists. 

Given that I think that removing all the items in the array and then re-adding them back in a synchronous operation  should be stable when rendered with glimmer.

tintinthong: > are you actually seeing that or are you guessing that that is what is happening?

I have tried to test this and in a plain card-basics environment. I have confirmed there is no re-render of the components actually. So I think ur right in that glimmer is handling it

tintinthong: > Although since we are using an identity map, @key is probably superfluous since we are not creating new instance.
> 
> 2 things to keep in mind:
> 
> 1. The array is being mutated not replaced
> 2. The items in the array are from an identity map so we are never recreating an instance that already exists.
> 
> Given that I think that removing all the items in the array and then re-adding them back in a synchronous operation should be stable when rendered with glimmer.

Cool. This is good to know that this pattern then works as expected

tintinthong (approved):
This is good. And I think I am convinced of the controlled re-renders returned from search.

burieberry (approved):

packages/host/app/resources/search.ts
```diff
@@ -8,56 +9,64 @@ import { restartableTask } from 'ember-concurrency';
...
 
 interface Args {
   named: {
     query: Query | undefined;
     realms: string[] | undefined;
-    isLive?: true;
+    isLive: boolean;
+    isAutoSaved: boolean;
     doWhileRefreshing?: (ready: Promise<void> | undefined) => Promise<void>;
+    storeService: StoreService;
```
lukemelia: Why pass this in as an arg rather than inject it?

packages/host/app/resources/search.ts
```diff
@@ -191,35 +163,29 @@ export class Search extends Resource<Args> {
...
-        if (!resultMap.has(url)) {
-          this.currentResults.delete(url);
-        }
-      }
-      for (let [url, resource] of resultMap) {
-        if (!this.currentResults.has(url)) {
-          this.currentResults.set(url, resource);
-        }
-      }
+      this._instances.splice(
```
lukemelia, line 187: Please add a comment giving some context to this for future readers.

lukemelia (approved):

packages/host/app/resources/search.ts
habdelra: nice catch, this was copy pasta

PR #2299, opened 2025-03-17
Avoid processing old realm events

lukemelia:


habdelra (approved):
This makes sense--if an event is recieved from before the client started we should ignore it, and presumably any event after that will have a client ID associated with it that we can correlate on.

packages/host/app/services/matrix-service.ts
```diff
@@ -1470,11 +1491,13 @@ export default class MatrixService extends Service {
           event,
         );
       } else {
+        (event.content as any).origin_server_ts = event.origin_server_ts;
```
backspace, line 1494: Is this for ease of debugging, so the `content` passed to subscribers has the timestamp?

lukemelia: I think we should probably get rid of this now.

PR #2302, opened 2025-03-18
Use dynamic context for `getCards`

habdelra:
This depends on https://github.com/cardstack/boxel/pull/2298. Please review and merge that first.

As with the `getCard()` API, the `getCards()` has been updated to use dynamic context. This means in host app code, please no longer import the `getSearch()` function. Rather use `@consume(GetCardsContextName)` to get a handle on `getCards()`, like so:

```ts
import { consume } from 'ember-provide-consume-context';
import { type getCards, GetCardsContextName } from '@cardstack/runtime-common';
  ...
class YourComponent extends Component {
  @consume(GetCardsContextName) private declare getCards: getCards;
```

From a card you can access getCards() from the @context--no need to import it (also, there is no longer a `getCards` action in the context).

If you need to use getCards() in something that is not a Component, then let's talk. Currently only components are consuming this so unlike `getCard()`, we are doing much better here.


packages/host/app/components/operator-mode/container.gts
```diff
@@ -61,37 +59,25 @@ export default class OperatorModeContainer extends Component<Signature> {
...
-  }
-
   @provide(GetCardContextName)
   // @ts-ignore "getCard" is declared but not used
   private get getCard() {
     return getCard;
   }
 
+  @provide(GetCardsContextName)
+  // @ts-ignore "getCard" is declared but not used
```
lukemelia:
```suggestion
  // @ts-ignore "getCards" is declared but not used
```

packages/host/app/components/operator-mode/container.gts
```diff
@@ -61,37 +59,25 @@ export default class OperatorModeContainer extends Component<Signature> {
 
   constructor(owner: Owner, args: Signature['Args']) {
     super(owner, args);
-    // TODO remove this
-    (globalThis as any)._CARDSTACK_CARD_SEARCH = this;
```
lukemelia, line 65: Good riddance!

lukemelia (approved):

packages/host/app/components/operator-mode/container.gts
habdelra: I think there is one last global used for invoking the card chooser from a card def. perhaps that can use the dynamic context as well....

packages/host/app/resources/search.ts
```diff
@@ -193,12 +194,18 @@ export class SearchResource extends Resource<Args> {
...
+// WARNING! please don't import this directly into your component's module.
+// Rather please instead use:
+// ```
+//   import { consume } from 'ember-provide-consume-context';
+//   import { type getCards, GetCardsContextName } from '@cardstack/runtime-common';
+//    ...
+//   @consume(GetCardsContextName) private declare getCards: getCards;
+// ```
+// If you need to use `getSearch()`/`getCards()` in something that is not a Component, then
+// let's talk.
```
backspace, line 211: Should we add a custom lint rule to discourage this programatically? [This one](https://github.com/cardstack/boxel/blob/main/packages/eslint-plugin-window-mock/wrapped-setup-helpers-only.js) could be a model as it looks at imports.

packages/host/app/resources/search.ts
```diff
@@ -42,11 +42,8 @@ export class SearchResource extends Resource<Args> {
   @service declare private realmServer: RealmServerService;
   @service declare private store: StoreService;
   @tracked private realmsToSearch: string[] = [];
-  // This is deprecated. consumers of this resource need to be reactive such
-  // that they can deal with a resource that doesn't have a search results yet.
-  loaded: Promise<void> | undefined;
+  private loaded: Promise<void> | undefined; // This should _always_ be private--do not make this public
```
backspace: Maybe an explanation of why would help the unfamiliar

backspace (approved):
🎉

jurgenwerk (approved):
Nice!

PR #2303, opened 2025-03-18
ci: Change all Actions uses to pinned SHAs

backspace:
This is [recommended by Github](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

### DIFF
--- a/.github/.yamllint.yml
+++ b/.github/.yamllint.yml
@@ -1,6 +1,8 @@
 extends: default
 
 rules:
+  comments:
+    min-spaces-from-content: 1
   line-length: disable
   document-start: disable
   quoted-strings:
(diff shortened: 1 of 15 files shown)


.github/.yamllint.yml
```diff
@@ -1,6 +1,8 @@
 extends: default
 
 rules:
+  comments:
+    min-spaces-from-content: 1
```
backspace, line 5: I chose to add this because my editor was removing the [double space the default calls for](https://yamllint.readthedocs.io/en/stable/rules.html#id6) when saving.

jurgenwerk (approved):

PR #2308, opened 2025-03-19
Refactor CardResource.loaded out of pill menu freestyle usage

habdelra:
This refactors `CardResource.loaded` out of pill menu freestyle usage. in this case it's totally not necessary to load a card from the store as we are just showing the freestyle usage. As part of this I added a method to get card instances that are detached from the store.

image

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -222,8 +222,7 @@ export default class StoreService extends Service {
 
   // This method is used for specific scenarios where you just want an instance
   // that is _not_ part of the identity map and detached from the store. This
-  // may include things like tests, freestyle guide, perhaps you just need a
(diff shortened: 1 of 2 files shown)


packages/host/app/services/store.ts
```diff
@@ -219,6 +220,24 @@ export default class StoreService extends Service {
...
+    if (!doc) {
+      return undefined;
+    }
+    return await this.cardService.createFromSerialized(
+      doc.data as LooseCardResource,
+      doc,
+      new URL(url),
+    );
+  }
+
```
habdelra: This is a low level method that consumers can use to get a card instance that is NOT part of the store.

burieberry (approved):

jurgenwerk (approved):

richardhjtan (approved):

PR #2310, opened 2025-03-19
Display instances in field playground dropdown

burieberry:
Changes:
- Display field instances from selected spec on playground panel's instance chooser dropdown. Dropdown items are numbered and use the atom template of field instances. (default to title field if there's one, fallback to card display name)
- Update dropdown display and `onSelect` behavior for fields
- Selected dropdown item name appears as `[spec title] - Example [index + 1]`
- Examples in Field chooser modal are also numbered now (cs-8151)

field-dropdown

### DIFF
--- a/packages/base/default-templates/atom.gts
+++ b/packages/base/default-templates/atom.gts
@@ -1,5 +1,5 @@
 import GlimmerComponent from '@glimmer/component';
-import { type CardDef, isCompoundField } from '../card-api';
+import { type CardDef } from '../card-api';
 import { cn, not } from '@cardstack/boxel-ui/helpers';
 
 export default class DefaultAtomViewTemplate extends GlimmerComponent<{
@@ -14,9 +14,6 @@ export default class DefaultAtomViewTemplate extends GlimmerComponent<{
     }
     if (typeof this.args.model.title === 'string') {
       return this.args.model.title.trim();
-    }
-    if (isCompoundField(this.args.model)) {
-      return;
     }
     return `Untitled ${this.args.model.constructor.displayName}`;
   }
--- a/packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
@@ -67,7 +67,7 @@ const PlaygroundPreview: TemplateOnlyComponent<Signature> = <template>
     <FittedFormatGallery
       @card={{@card}}
       @isDarkMode={{true}}
-      @isField={{@isFieldDef}}
+      @isFieldDef={{@isFieldDef}}
     />
   {{/if}}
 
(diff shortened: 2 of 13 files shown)


packages/base/default-templates/atom.gts
```diff
@@ -15,9 +15,6 @@ export default class DefaultAtomViewTemplate extends GlimmerComponent<{
     if (typeof this.args.model.title === 'string') {
       return this.args.model.title.trim();
     }
-    if (isCompoundField(this.args.model)) {
-      return;
-    }
```
burieberry, line 20: removing this allows compound fields' atom template to use `title` field as default (if the compound field has a title field) and `untitled [card type]` as fallback

backspace:
When I run this locally (and in the preview deployment) the entries aren’t numbered, is that something that changed after your screenshot?

blog-post gts in feb11 2025-03-20 11-22-23

burieberry:
> When I run this locally (and in the preview deployment) the entries aren’t numbered, is that something that changed after your screenshot?
> 
> blog-post gts in feb11 2025-03-20 11-22-23

Only the field instances are numbered, since they belong to a collection, not the cards. Hmm, that can be confusing, maybe we should reconsider.

burieberry:
> > When I run this locally (and in the preview deployment) the entries aren’t numbered, is that something that changed after your screenshot?
> > blog-post gts in feb11 2025-03-20 11-22-23
> 
> Only the field instances are numbered, since they belong to a collection, not the cards. Hmm, that can be confusing, maybe we should reconsider.

I can remove the numbering from the dropdown. CS-8151 only mentions numbering in the modal...

backspace:
> Only the field instances are numbered, since they belong to a collection, not the cards. Hmm, that can be confusing, maybe we should reconsider.

ah I was looking at the wrong thing then, thanks! I can review properly now 🤓

packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
```diff
@@ -64,7 +64,11 @@ const PlaygroundPreview: TemplateOnlyComponent<Signature> = <template>
       veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea
       commodo consequat.</div>
   {{else if (eq @format 'fitted')}}
-    <FittedFormatGallery @card={{@card}} @isDarkMode={{true}} />
+    <FittedFormatGallery
+      @card={{@card}}
+      @isDarkMode={{true}}
+      @isField={{@isFieldDef}}
```
backspace: is “field” conceptually the same as “fielddef”?

backspace (approved):

packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
burieberry: the way the types are set, `FieldDef` means it's an instance of field definition. `typeof FieldDef` means the constructor. meaning of `isField` is vague, I can change that to be `isFieldDef`

PR #2311, opened 2025-03-20
fix navigation issues with spec

tintinthong:
This PR fixes the issue that if you change a selection on LHS, it doesn't remain stale (particularly after you have chosen something in the dropdown) See this [test](https://github.com/cardstack/boxel/blob/b50af37b4dd788845c9ea09bef94c3c6a968c406/packages/host/tests/acceptance/code-submode/spec-test.gts#L990)

Other changes 
-  moved out of pre-rendered card search bcos shudn't rely on pre-rendered card to return meta data (so I used getSearch)
- selected spec is persisted 
- any writes to playground service is done thru a modifier since we shudn't rely on the readiness of the search resource before commiting the persisted playground state
- once you create a spec, it will open the accordion if its not already opened 

As part of this PR there is a digression in auto-save. I do not have a good solution for that and have skipped the test. But, the structure that we want is closer to this. PLS REVIEW THIS INDEPENDENTLY OF THE AUTO_SAVE ISSUE

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode.gts
+++ b/packages/host/app/components/operator-mode/code-submode.gts
@@ -63,6 +63,7 @@ import type { FileView } from '@cardstack/host/services/operator-mode-state-serv
 import type OperatorModeStateService from '@cardstack/host/services/operator-mode-state-service';
 import type RealmService from '@cardstack/host/services/realm';
 import type RecentFilesService from '@cardstack/host/services/recent-files-service';
+import type SpecPanelService from '@cardstack/host/services/spec-panel-service';
 
 import type { CardDef, Format } from 'https://cardstack.com/base/card-api';
 import { type SpecType } from 'https://cardstack.com/base/spec';
@@ -111,7 +112,7 @@ type PanelHeights = {
   recentPanel: number;
 };
 
-type SelectedAccordionItem =
+export type SelectedAccordionItem =
   | 'schema-editor'
   | 'spec-preview'
   | 'playground'
@@ -154,6 +155,7 @@ export default class CodeSubmode extends Component<Signature> {
   @service private declare environmentService: EnvironmentService;
   @service private declare realm: RealmService;
   @service private declare loaderService: LoaderService;
+  @service private declare specPanelService: SpecPanelService;
 
   @tracked private loadFileError: string | null = null;
   @tracked private userHasDismissedURLError = false;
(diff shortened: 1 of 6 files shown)


packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -932,7 +944,11 @@ export default class CodeSubmode extends Component<Signature> {
                       <A.Item
                         class='accordion-item'
                         @contentClass='accordion-item-content'
-                        @onClick={{fn this.selectAccordionItem 'schema-editor'}}
+                        @onClick={{fn
+                          this.selectAccordionItem
+                          'schema-editor'
+                          false
```
richardhjtan: @tintinthong is this boolean parameter needed?  
`@action private selectAccordionItem(item: SelectedAccordionItem) {`

packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -1034,10 +1050,8 @@ export default class CodeSubmode extends Component<Signature> {
                       {{#if this.showSpecPreview}}
                         <SpecPreview
                           @selectedDeclaration={{this.selectedDeclaration}}
-                          @onPlaygroundAccordionToggle={{fn
-                            this.selectAccordionItem
-                            'playground'
-                          }}
+                          @isLoadingNewModule={{this.moduleContentsResource.isLoadingNewModule}}
```
richardhjtan, line 1049: This might not be needed

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -514,34 +473,31 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
-          },
-        },
-      };
-      await this.cardResource?.loaded;
-      if (this.card) {
-        this._selectedCardId = this.card.id;
-        this.updateFieldSpecForPlayground(this.card.id);
-        this.newCardJSON = undefined;
+      try {
+        let Klass = await loadCard(specRef, {
```
richardhjtan: Perhaps a more readable variable name here, like `Spec` etc

packages/host/app/components/operator-mode/code-submode.gts
tintinthong: I do want to keep this so I can show a loading state while changing module and return a default title

packages/host/app/components/operator-mode/code-submode.gts
tintinthong: good catch

burieberry (approved):
Looks good

PR #2313, opened 2025-03-20
Make playground preview use available space

FadhlanR:
In this PR, the isolated and edit preview in playground panel will fill all available horizontal space.

https://github.com/user-attachments/assets/f953e96b-8872-4325-a4fb-a9f1137b1a47

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
@@ -29,7 +29,11 @@ interface Signature {
 const PlaygroundPreview: TemplateOnlyComponent<Signature> = <template>
   {{#if (or (eq @format 'isolated') (eq @format 'edit'))}}
     <CardContainer
-      class={{if @isFieldDef 'field-preview-container' 'full-height-preview'}}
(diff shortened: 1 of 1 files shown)


richardhjtan (approved):

packages/host/app/components/operator-mode/code-submode/playground/playground-preview.gts
```diff
@@ -29,7 +29,11 @@ interface Signature {
 const PlaygroundPreview: TemplateOnlyComponent<Signature> = <template>
   {{#if (or (eq @format 'isolated') (eq @format 'edit'))}}
     <CardContainer
-      class={{if @isFieldDef 'field-preview-container' 'full-height-preview'}}
+      class={{if
+        @isFieldDef
+        'field-preview-container isolated-and-edit-preview'
+        'full-height-preview isolated-and-edit-preview'
+      }}
```
burieberry: let's only add this new class for the card case and not for field def

burieberry (commented):
Good to go after removing the class from field def case. While reviewing realized that it looks better if field's edit view doesn't have full width, we can assume it's an "embedded" edit view.

PR #2315, opened 2025-03-20
Refactor CardResource.loading out of create-file-modal.gts

habdelra:


packages/host/app/components/operator-mode/create-file-modal.gts
```diff
@@ -389,6 +388,16 @@ export default class CreateFileModal extends Component<Signature> {
     return !!this.currentRequest;
   }
 
+  private get isReady() {
+    return this.definitionClass
+      ? true
+      : Boolean(this.defaultSpecResource?.isLoaded);
```
tintinthong, line 394: So now we shud check for the resource using the synchronous `isLoaded`?

are we supposed to be moving to a pattern where we never wait for the resource? at least, tht was my understanding but it cud be possible that in "setting up" use cases that it is not viable.

habdelra: Yes, that is exactly right. Awaiting for a resource is an anti-pattern. That means you are consuming it in a non-reactive manner.

habdelra: You should use a Boolean “isLoaded” in the template to wait for a resource to be ready. Never use a task in the constructor to wait for the resource to be ready. This goes for all resources. Never await a resource. Remove any promises that are publically available for a resource. @lukemelia just ran into this issue in the RoomReource that was causing problems yesterday too.

habdelra: Another way to think of it is that the resource handles all the async internally and only exposes a synchronous API. The resource starts  leaking its abstraction when it has public async methods and properties.

habdelra: This is also why this refactor is so challenging. I’m having to rethink so many of these resources’ consumers so that they actually become reactive—or more reactive as the case may be. Another original sin is when a resource is consumed outside of a Conponent (like a service). That is a huge no-no and a large reason why we have a lot of these problems.

jurgenwerk (approved):

packages/host/app/components/operator-mode/create-file-modal.gts
tintinthong: Thx for the explanation. 

I just thought tht even the idea of checking if "isLoaded" was an anti-pattern too. Bcos it lies about a state tht may change (but I can see how consumers expect tht value to be ready)

habdelra: nope--that is totally cool. as long as its synchronous it's :100:

tintinthong (approved):

packages/host/app/components/operator-mode/create-file-modal.gts
habdelra: In @lukemelia's case he added a "lastLoadTime" property to the resource to keep track of a specific load. that's also a totally valid scenario in the case where you just want to react to a particular `modify()` hook being called.

PR #2320, opened 2025-03-21
Refactor CardResource.loaded out of room-message.gts

habdelra:


packages/host/app/components/matrix/room-message.gts
```diff
@@ -151,71 +166,32 @@ export default class RoomMessage extends Component<Signature> {
...
     }
-
-    let hasResourceErrors = this.loadedResources.errors.length > 0;
-    if (hasResourceErrors) {
-      return 'Error rendering attached cards.';
-    }
-
-    return this.loadedResources.errors
-      .map((e: { id: string; error: Error }) => `${e.id}: ${e.error.message}`)
-      .join(', ');
```
habdelra, line 219: it was impossible to get to this place in the code, as we would always return above in this scenario, so i opted to remove it.

backspace (approved):

lukemelia (approved):

packages/host/app/components/matrix/room-message.gts
```diff
@@ -93,13 +103,14 @@ export default class RoomMessage extends Component<Signature> {
   });
 
   <template>
+    <div class='hide' {{consumeContext consume=this.makeCardResources}} />
```
jurgenwerk, line 106: I see there is one more place with this pattern of adding a hidden div and `.hide` style. Is this an opportunity to make a reusable component for this purpose?

jurgenwerk (approved):

packages/host/app/components/matrix/room-message.gts
habdelra: that's not a bad idea

PR #2322, opened 2025-03-21
Can autogenerate spec or containedExamples from playground

burieberry:
This is for fields only:
- If there's a spec for field but it has no `containedExamples`, then on opening playground panel, autogenerate a field instance and add to `containedExamples`.
- If there's no pre-existing spec for field, on opening playground panel, autogenerate a Spec instance and then add a field instance to `containedExamples`.

Review needed: See my comments below in using a search resource and a modifier to check if spec exists and to auto-generate it if it doesn't.


packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -484,15 +529,16 @@ export default class PlaygroundContent extends Component<Signature> {
...
+        this.card.id,
+        'edit',
+        this.args.isFieldDef ? 0 : undefined,
+      ); // open new instance in playground in edit format
     }
+    this.closeInstanceChooser();
   });
 
-  private createNewField = task(async (specCard: Spec) => {
+  private createNewField = restartableTask(async (specCard: Spec) => {
```
burieberry, line 514: this is now a `restartableTask` because `fieldInstances` getter is called twice and this task starts running if there are no instances

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -118,17 +127,14 @@ export default class PlaygroundContent extends Component<Signature> {
...
-            Add Field
-          </AddButton>
+        {{else if @isFieldDef}}
+          {{#if
+            (or this.searchFieldSpec.isLoading this.createNewCard.isRunning)
+          }}
+            <LoadingIndicator @color='var(--boxel-light)' />
+          {{else if this.canGenerateFieldSpec}}
+            <div {{runFn this.createNew}} />
+          {{/if}}
```
burieberry: a little awkward. using a modifier to kickstart `createNew` if conditions have been met

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -199,6 +201,36 @@ export default class PlaygroundContent extends Component<Signature> {
...
+      !this.canWriteRealm ||
+      this.createNewCard.isRunning
+    ) {
+      return false;
+    }
+    return (
+      !this.searchFieldSpec.isLoading &&
+      this.searchFieldSpec.instances?.length === 0
+    );
+  }
```
burieberry: this checks if there's a preexisting spec as well as other conditions for running createNew to autogenerate a new spec

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
lukemelia: Can this be conditionally initiated in the constructor instead?

burieberry: Sure, I can make it a component

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -30,6 +29,8 @@ import {
 
 import consumeContext from '@cardstack/host/modifiers/consume-context';
 
+import { getSearch } from '@cardstack/host/resources/search';
```
habdelra: no, please don't import `getSearch` use the provide consume context instead. https://github.com/cardstack/boxel/blob/main/packages/host/app/resources/search.ts#L202-L212

habdelra (changes requested):
use provide-consume context to get access to `getCards()`

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -199,6 +194,32 @@ export default class PlaygroundContent extends Component<Signature> {
...
+  private get specQuery(): Query {
+    return {
+      filter: {
+        on: specRef,
+        eq: { ref: this.args.codeRef },
+      },
+    };
+  }
+
+  private searchFieldSpec = getSearch(
```
habdelra: since this is in the constructor, you'll need to use the new `{{consumeContext}}` helper to get access to provided `getCards()` (provided contexts are yet available in the constructor--only after DOM has been created). https://github.com/cardstack/boxel/blob/main/packages/host/app/components/operator-mode/code-submode.gts#L793. As well as you'll need to use `@tracked` on `searchFieldSpec`.

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -457,15 +480,26 @@ export default class PlaygroundContent extends Component<Signature> {
 
   private createNewCard = task(async () => {
```
habdelra: does this need to be restartable? what happens if the user selects this again very quickly before the async finishes?

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -457,15 +480,26 @@ export default class PlaygroundContent extends Component<Signature> {
...
     if (this.args.isFieldDef) {
+      let fieldCard = await loadCard(this.args.codeRef, {
```
habdelra, line 467: I apologize for the name of this function. we should really rename it so it doesn't confuse people. maybe `loadCardDef`. I'll make a ticket for that.

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
burieberry: Yes, missed this during a recent change

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
tintinthong: actually @habdelra why is this the case? Isn't playground-content just a component that uses a resource. Its not a card.

burieberry: ^ @habdelra curious about this too. I did a refactor to minimize search consumption. ready for another review

habdelra: As per @ef4 , we want everything that uses the store to get it the same way, whether its a card or the host app. eventually we will use closure to setup our context dynamically and we'll control that from our context provider. this will allow us to have multiple stores, i.e. being able to have the playground use a different loader than the rest of the app. This will allow for much more efficient and safer state management. e.g.changing code in the playground shouldn't force you to have to reload code used for the host app. In order to get there though, it's important to lay the groundwork now so that we can use the store consistently everywhere.

habdelra (approved):
looks good thanks!

packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
tintinthong: ok thanks for the explanation

PR #2324, opened 2025-03-24
CS-8210 promote fields to base realm

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8210/promote-fields-to-base-realm

### What is changing
- promote reusable fields from experiment and seed to base realm
- EntityDisplay components (thumbnail, icon) to Boxel UI components
- added reusable country list helpers that used in the country field


packages/boxel-ui/addon/src/components/entity-thumbnail-display/usage.gts
```diff
@@ -0,0 +1,76 @@
+import { fn } from '@ember/helper';
```
lucaslyl, line 1: @richardhjtan  Just one small comment, and I'm not too sure about this, is it ideal to include the entity-icon-display inside boxel-ui component

lucaslyl (commented):
overall looks good

packages/boxel-ui/addon/src/components/entity-thumbnail-display/usage.gts
richardhjtan: @lucaslyl Yes, this PR covers both the entity display icon and the thumbnail. The entity display icon component was used quite frequently in some fields.

lukemelia (approved):

PR #2329, opened 2025-03-24
Move instance chooser to panel header

FadhlanR:
https://github.com/user-attachments/assets/4e7d81dc-bda7-40b2-b90e-b036662809cb


backspace:
The Percy diffs have a few snapshots where the card is no longer rendered despite showing as chosen in the dropdown menu, is that expected?

image

burieberry (commented):
Some feedback:
- The dropdown should only appear when playground panel is opened
- We will need to refactor the PlaygroundPanel component in a similar way to SchemaEditor and SpecPreview components, where the PlaygroundPanel component wraps the AccordionItem and yields the title and content components
- Some styling fixes: Dropdown trigger should get shorter when panel is narrower with the arrow visible; remove `padding-top: 0` from `.playground-panel` class

Let me know if you need help with the refactor

FadhlanR:
> The Percy diffs have a few snapshots where the card is no longer rendered despite showing as chosen in the dropdown menu, is that expected?
> 
> image

No it's not expected. Nice catch! I'll check it.

FadhlanR:
> Some feedback:
> 
> * The dropdown should only appear when playground panel is opened
> * We will need to refactor the PlaygroundPanel component in a similar way to SchemaEditor and SpecPreview components, where the PlaygroundPanel component wraps the AccordionItem and yields the title and content components
> * Some styling fixes: Dropdown trigger should get shorter when panel is narrower with the arrow visible; remove `padding-top: 0` from `.playground-panel` class
> 
> Let me know if you need help with the refactor

Ah I see, I didn't check the Schema editor. I'll update the code.

FadhlanR:
@backspace @burieberry this PR ready for another review. I made some updates based on your feedbacks.

packages/host/app/components/operator-mode/code-submode/format-chooser.gts
```diff
@@ -68,6 +69,7 @@ export default class FormatChooser extends Component<Signature> {
         margin-right: var(--boxel-sp-xxs);
         border: 1px solid
           var(--boxel-format-chooser-border-color, var(--boxel-700));
+        box-shadow: var(--boxel-deep-box-shadow);
```
FadhlanR, line 72: This update is not related to this PR but this is the small fix for format chooser button.

backspace:
blog-category gts in buck 2025-03-25 12-24-57

This menu should probably be wider when it’s empty. I think if truncation is possible it should also have `title` attributes so hovering works to show the full text. (Or maybe we have a different way to do that?)

burieberry (commented):
The refactor looks good. I agree with buck's comments on making the minimum for the dropdown larger. I guess the user will have to adjust their panel size if they go below the minimum. Left-align the dropdown headers "Recent" and "Action" while you're at it.

burieberry:
@FadhlanR Actually I merged my PR which caused conflicts here. I can resolve them now and also make the dropdown changes mentioned.

burieberry:
> @FadhlanR Actually I merged my PR which caused conflicts here. I can resolve them now and also make the dropdown changes mentioned.

@FadhlanR I fixed the merge conflicts. I did not do the dropdown changes. Please continue from here. I noticed another issue. It's no longer possible to use the keyboard to navigate to the dropdown and to open and select items in it.

FadhlanR:
I’m not sure if this makes sense, but I set the instance chooser width to a fixed 271px.

Screenshot 2025-03-26 at 13 31 54

burieberry (approved):

PR #2330, opened 2025-03-24
refactor consume context modifier into a helper

habdelra:


lukemelia (commented):
Interesting. Could consumeContext be a helper instead of a modifier, and would that address this more elegantly?

habdelra:
> Interesting. Could consumeContext be a helper instead of a modifier, and would that address this more elegantly?

i don't think so--the timing for when this is run is critical. we had to add this because there are scenarios where we want the use the provided context at the initiation of the component, however, you cannot consume a provided context in the constructor of a component--the provide-consume mechanism relies upon the DOM for providing the context. As such the earliest time that you can access a provided context is after the DOM has been rendered. so we are using the modifier as a timing mechanism that allows us access to the provided context at the safest earliest possible moment. I'm not sure that we can rely upon helpers for the timing, as I don't think we are guaranteed that the component's DOM has been fully established when the helper runs.

lukemelia:
You could use a helper with `schedule('afterRender', consume)`

habdelra:
> You could use a helper with `schedule('afterRender', consume)`

intersting, i hadn't considered that. I can refactor to do that instead

lukemelia:
Another argument against this: the `hide` div is messing up some CSS in the AI Assistant

https://github.com/user-attachments/assets/ac130190-26a6-4163-bf7d-cef92a0fcae4

lukemelia (approved):
👍

jurgenwerk (approved):
Nice!

PR #2336, opened 2025-03-25
Cs 8221 add showcase filter and create the skeleton layout

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8221/add-showcase-filter-and-create-the-skeleton-layout

Skeleton Placeholder demo:

https://github.com/user-attachments/assets/38c285a8-2c64-4c28-aa76-bc2403f60d5b


tintinthong (approved):
looks good to me. That was fast!

PR #2338, opened 2025-03-25
Upgrade to latest monaco editor version

jurgenwerk:
This upgrade is needed because in the newer versions, monaco supports folded diff editors, which will be useful for our code diffs in the AI panel (previous monaco version was from 3 years ago). 

After upgrading I experienced failing tests, failing with an error coming from monaco internals after trying to dispose the editor in `registerDestructor` when `CodeBlock` component gets destroyed. **But this only happens in tests**. It looks like a timing issue between destroying the component and disposing the editor (another open issue https://github.com/microsoft/monaco-editor/issues/4389 suggests this happens when destroying and recreating the editor within a few milliseconds). 

I opened an issue https://github.com/microsoft/monaco-editor/issues/4859 and in the meanwhile opted into adding a patch that comments out the code that rejects the promise that results in the error. Interacting with monaco editors in our app still works fine after the upgrade, I tested them manually and it looks good 👌

Code diffing feature that we want to try in our AI panel looks like this:

image


lukemelia (approved):

backspace (approved):
I’m glad the intervention required was minor! Should we be keeping up with newer versions more than every three years? hehe

jurgenwerk:
> Should we be keeping up with newer versions more than every three years

Hopefully 🤣

PR #2339, opened 2025-03-25
Refactor StackItem to use a card URL instead of wrapping a CardResource

habdelra:
This PR is very intense. unsure how to break it apart, as a StackItem is really central to our interact mode. The driving idea was to not deal in CardResources until the last possible moment in the consuming component--this means we don't pass around CardResource, nor CardDef--but rather use card ID's and rely on the store to do our caching. Also, of note, in order to do this refactor I had to change the API of the Card Chooser and Create Card to deal in CardResources instead of just raw CardDef's. This solves lots of problems, but makes the refactor even deeper. 

One of the big take aways is to be very very careful about dealing in `CardDef` in your component state. You almost always want to be using `CardResource` and not `CardDef` within a component. It's also too easy for `CardDef`'s to get detached from their component's lifetime when we pass them around to other components/services/commands, or maybe even getting consumed in a resource with a totally different life time than the originating component. 

we really need to think more deeply around "where did this CardDef come from?" and "when does it stop being used?". most of the time we probably just want to be using card ID's, and at the point we need to put a card on the page that is when we should get a `CardDef` (via a CardResource). Keep in mind that now that we have a store, we can take advantage of the store's cache when getting a `CardDef`--so passing around card ID's can be very efficient.

The CardResource has been modified to include 2 new methods:
- `getNewCardURL(): Promise<string>` This method is used to return the card URL for a newly created card. We have a couple scenarios were we create cards and really just need the ID. This breaks from the best practice of having a synchronous API on the resource, however, we don't actually return the card in this case, just a string so it would be difficult to abuse this.
- `dispose(): Promise<CardDef | CardError | undefined>` This method is used to deal with the places where we just don't care about the lifetime of the resource and we just want to consume the underlying card immediately and be done with it. The most common scenario here is the creation of a new card, which first you had to choose a spec instance. The card chooser gives you back a card resource for the spec (that has the lifetime of the component that asked for the card from the chooser). however, you really don't care about the resource in this case as you just want the code ref on the spec for the purposes of making a card. In that case I made a method that makes much more clear the thing we were back-dooring into before: this method will give you the underlying card instance and call destroy on the CardResource that was previously holding this. This means card instance you are handed is no longer receiving server updates, it is not auto saving, and the store has marked it for garbage collection.

Also, on the StoreService I have added a new method:
- `peek(url: string): Promise<CardDef | CardError | undefined>`. This is used for places where we like to get access to instances in the identity map but we don't have a clear lifetime (because we are consuming the instance in something other than a component, e.g. a Command). In this case the store will give you an instance from the identity map, or optionally fetch an instance if it's not in the identity map, and then add it to the identity map. The store will mark the instance for garbage collection so that it doesn't have to hang on to it forever since it's unclear what the life time should be. These instances do not have server updates and they do not autosave.


packages/host/app/components/ai-assistant/attachment-picker/index.gts
```diff
@@ -46,101 +54,104 @@ const MAX_ITEMS_TO_DISPLAY = 4;
...
-
-              <:content>
-                {{#if @autoAttachedCardTooltipMessage}}
-                  {{@autoAttachedCardTooltipMessage}}
-                {{else if (this.isAutoAttachedCard item)}}
-                  Topmost card is shared automatically
-                {{/if}}
-              </:content>
-            </Tooltip>
+      {{#if this.isLoaded}}
```
habdelra, line 58: do we need a loading state here?

packages/host/app/resources/auto-attached-card.ts
```diff
@@ -26,94 +23,60 @@ interface Args {
...
-  }
-
-  private wasPreviouslyRemoved(lastTopMostCard: CardDef) {
-    return this.lastRemovedCards.has(lastTopMostCard.id);
-  }
-
-  @action
-  onCardRemoval(card: CardDef) {
-    this.lastRemovedCards.add(card.id);
-  }
```
habdelra, line 105: This is very wrong. we should never mutate the state of a resource outside of the thunk.

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -380,32 +382,34 @@ export default class InteractSubmode extends Component<Signature> {
...
 
     // only save when closing a stack item in edit mode. there should be no unsaved
     // changes in isolated mode because they were saved when user toggled between
     // edit and isolated formats
-    if (item.format === 'edit') {
-      let updatedCard = this.args.saveCard(card);
-      request?.fulfill(updatedCard);
+    if (url && item.format === 'edit') {
+      this.store.save(url); // intentionally ignoring promise so caller is not waiting for save to complete
+      request?.fulfill();
```
habdelra: I'm anxious about throwing away this promise. my instinct is telling me this will come back to bite me...

habdelra: Also, since we have an identity map, it seems silly to return the card that you just saved. it's not a different card...

packages/host/app/services/matrix-service.ts
```diff
@@ -150,9 +150,12 @@ export default class MatrixService extends Service {
     () => this.cardAPI,
   );
 
+  // TODO This seems very bad. we should not be sharing Resources with anyone that
+  // wants one--resources are tied to the lifetime of their owner, who knows
+  // which owner made these and who is consuming these. we need to refactor this out..
   roomResourcesCache: TrackedMap<string, RoomResource> = new TrackedMap();
```
habdelra, line 156: we need to stop using resources like this

tintinthong: what is the pattern to handle something like this? 

I guess this shudnt even be in the matrix service in the first place

habdelra: confirm.

habdelra: if we need to have a cache it should be at the resource level, like how the store caches cards for card resources. but it does mean managing resource lifetime for the cache

packages/host/app/services/store.ts
```diff
@@ -221,11 +249,15 @@ export default class StoreService extends Service {
...
-  async getInstanceDetachedFromStore(
-    url: string,
-  ): Promise<CardDef | CardError> {
+  // that is not auto saving and not receiving live updates. This may include
+  // things like tests, freestyle usage guide, instances that we immediately
+  // consume, etc.
+  async peek(url: string): Promise<CardDef | CardError> {
+    let cached = this.identityContext.get(url);
+    if (cached) {
+      this.markForGarbageCollection(url);
```
habdelra: Don’t mark for garbage collection if a resource is associated with this instance

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -211,44 +214,34 @@ export default class InteractSubmode extends Component<Signature> {
...
           stackIndex = this.stacks.length;
         }
         let newItem = new StackItem({
-          owner: here,
-          url: cardOrURL instanceof URL ? cardOrURL : new URL(cardOrURL.id),
+          url: cardOrURL instanceof URL ? cardOrURL.href : cardOrURL.id,
           format,
           stackIndex,
         });
-        await newItem.ready();
```
tintinthong, line 251: I don't know how exactly. But, yesterday, the catalog listing had issues where linksToMany fields (ie return [undefined, undefined]) would suddenly disappear when popped onto the stack (after a refresh of the app)

Those links end up as undefined and will be accidentally saved too.

I merged in this PR into my branch and the links are consistently displaying

packages/runtime-common/index.ts
```diff
@@ -227,7 +230,7 @@ export async function chooseCard<T extends BaseDef>(
   }
   let chooser: CardChooser = here._CARDSTACK_CARD_CHOOSER;
 
-  return await chooser.chooseCard<T>(query, opts);
+  return await chooser.chooseCard<T>(owner, query, opts);
```
tintinthong: This needs an owner? This remains as resource bcos it bottoms out to `getSearch`. Is there any intention to not have this as a resource?

packages/runtime-common/index.ts
```diff
@@ -362,6 +371,7 @@ export interface SearchQuery {
 
 export interface Actions {
   createCard: (
+    owner: object,
```
tintinthong: is this a weird thing that createCard needs to have an `owner`? Or is it confirmed what we expect in the api.

habdelra: I don't think its weird at all. remember we are dealing in card resource now. it's not safe to pass raw instances around. that means that if you want a card you need to say who wants it because it's gonna give you a resource for the card, not the raw instance.

packages/runtime-common/index.ts
habdelra: remember we are dealing in card resource now. it's not safe to pass raw instances around. that means that if you want a card you need to say who wants it because it's gonna give you a resource for the card, not the raw instance.

tintinthong: Ok understood. But if chooseCard is returning the resource, it is being invoked inside a lot of restartableTask alot of the time.

Is this something we don't need that will be handled next? or is it still an ok pattern? (I understand there is a lot of dealing with legacy so its difficult to handle all in one-go)

habdelra: i don't see what the problem is. chosing a card is inherently an async operation, as you are awaiting the human to pick the card. it's totally fine to set a resource in a task. what is the concern that you have?

tintinthong (approved):
I have looked thru this PR and the changes make sense to me

I have seen on my side an improvement of a bug where linksToMany components rendered on the stack item do not appear null and get readily loaded properly

Wud be good if others cud feedback to

packages/base/links-to-editor.gts
```diff
@@ -162,8 +163,14 @@ export class LinksToEditor extends GlimmerComponent<Signature> {
...
-    if (chosenCard) {
-      this.args.model.value = chosenCard;
+    if (chosenCardResource) {
+      // this resource is our own (tied to this component's lifetime),
+      // but we need to make sure the card is loaded before we can set
+      // the value of the BoxModel
+      await chosenCardResource.loaded;
+      if (chosenCardResource.card) {
+        this.args.model.value = chosenCardResource.card;
+      }
```
habdelra: @ef4 I was able to get rid of references to the `CardResource.loaded` everywhere except in our links-to-editor and links-to-many editor. I tried a bunch of different ideas, but i'm unsure how to eliminate these.

packages/base/links-to-many-component.gts
```diff
@@ -119,9 +120,15 @@ class LinksToManyEditor extends GlimmerComponent<Signature> {
...
-      (this.args.model.value as any)[this.args.field.name] = selectedCards;
+    if (chosenCardResource) {
+      // this resource is our own (tied to this component's lifetime),
+      // but we need to make sure the card is loaded before we can set
+      // the value of the BoxModel
+      await chosenCardResource.loaded;
+      if (chosenCardResource.card) {
+        selectedCards = [...selectedCards, chosenCardResource.card];
+        (this.args.model.value as any)[this.args.field.name] = selectedCards;
+      }
```
habdelra: @ef4 ditto for this one

habdelra:
Ed feedback:
- chooser just returns ID instead of resource that is ultimately disposed. this way the consumer can decide if they want to peek, or they want an actual resource to live on in their state. (this also means that we no longer have to keep track of the owner in the chooser and create card modals since it will not be making resources).
- refactor how GC works so that we have a global GC timer. in the GC process we move thru the identity map and check to see if the instance is being referenced by an instance that has subscribers. if not then it is eligible for GC. After at least 2 consecutive GC sweeps where the instance is marked "dead" we remove the instance from the store.
- in the links to editors, let's use the peek mechanism to get the card from the store, and then perform the assignment. performing the assignment makes the instance we peeked no longer eligible for GC since it is now being referenced by an instance that has a resource that is subscribed to it.
- probably we can factor out the `CardResource.dispose()` and just rely on consumers that want immediate consumption of the instance to use the `Store.peek()`
- Move instance creation to the store (probably make a new ticket for this)

burieberry (approved):
These changes look good to me as well. It's nice to have all these use case examples. 

In light of these new methods, I'm wondering what the best pattern would be for a CreateCardInstance Command. In that case, it is sufficient to return the url of the new card. So, pass in the owner to the command, create new card resource, get the card url, get rid of the resource, return the url. Does this sound right?

habdelra:
> These changes look good to me as well. It's nice to have all these use case examples.
> 
> In light of these new methods, I'm wondering what the best pattern would be for a CreateCardInstance Command. In that case, it is sufficient to return the url of the new card. So, pass in the owner to the command, create new card resource, get the card url, get rid of the resource, return the url. Does this sound right?

Yup that sounds perfect. Return a card id, and the caller can then decide how they want to deal with it: either to hold the card in its state (because it’s a component) and use getCard() to make a resource, or use Store.peek() to get an instance that is immediately consumed. 

There should be no need to hold CardDef instances in the consumer’s state. Either use id’s or card resources. If you need a card and you have an id, use Store.peek immediate before the card is consumed and the store cache will serve you well.

habdelra:
FYI: Refactors suggested by @ef4 have been completed and is ready for another review.

We have a much more sophisticated GC algorithm such that when you `peek()` an instance from the store it will be eligible for garbage collection if no other `CardResource` is already subscribed to the instance. However, as soon that instance becomes consumed by another instance that has `CardResource` subscribers, it will no longer be a candidate for garbage collection. 

Consider how the links-to editor component works:
It uses the `chooseCard()` API to receive a card ID that the user has selected. It then uses `Store.peek()` to get the running instance that is a possible candidate for garbage collection. The rule is that if you use `Store.peek()` you must consume the instance immediately. The links-to editor then assigns the running instance it got from `Store.peek` to the `this.args.model.value` of the component (the `BoxModel`). In doing so, it actually adds the peek-ed value to the consumption graph of the instance for which you editing the links-to field. Now the peek-ed value is no longer eligible for garbage collection because it is part of the consumption graph for a `CardResource`--in this case the `CardResource` that is the parent of the links-to field. Assigning an instance is considered a form of immediate consumption.

habdelra:
> These changes look good to me as well. It's nice to have all these use case examples.
> 
> In light of these new methods, I'm wondering what the best pattern would be for a CreateCardInstance Command. In that case, it is sufficient to return the url of the new card. So, pass in the owner to the command, create new card resource, get the card url, get rid of the resource, return the url. Does this sound right?

@burieberry just an FYI: the API for creating a card has changed. you now use `Store.createInstance()` to create a card and it returns the url of the card.

PR #2340, opened 2025-03-25
newer version of a skill card replaces the older version

lukemelia:
- also ensure updated command definitions are uploaded if skill is attached


packages/host/app/lib/card-event-publisher.ts
```diff
@@ -0,0 +1,301 @@
...
+      await this.addCommandDefinitionsToRoomHistory(commandDefinitions, roomId);
+    }
+    return this.addCardsToRoom(skills, roomId);
+  }
+
+  private async addCommandDefinitionsToRoomHistory(
+    commandDefinitions: SkillCardModule.CommandField[],
+    roomId: string,
+  ) {
+    console.trace(
```
backspace: Is this meant to be merged?

backspace (approved):

packages/host/app/lib/card-event-publisher.ts
lukemelia: Nope!

PR #2351, opened 2025-03-27
poc of single copy source

richardhjtan:
This POC is to make sure gts file is able to copy through the new copy source action.

**Screenshot**

https://github.com/user-attachments/assets/1cd439a9-3bd6-4b7d-bda5-015688aca4b5


tintinthong:
If its a command to copy a card definition, we might want to just put it in separately into main. Not our skeleton catalog app. Then core team can review there

tintinthong:
This does unblock us but I think we want a command for this so other consumers like ai assistant can make use of this

richardhjtan:
Close this PR as this POC has transitioned into another PR

PR #2354, opened 2025-03-27
Fix broken file attachments after upgrading synapse from v1.99.0rc1 to v1.126.0

jurgenwerk:
The mentioned upgrade (made in https://github.com/cardstack/boxel/pull/2290/) broke our file attachments in the AI assistant. The upgrade:

1. Made all synapse endpoints authenticated -> I added an `Authorization` header where the bot is downloading attached files
2. Changed the upload endpoint from `_matrix/media/v3/download/SERVER/ID` to `media/v3/download/SERVER/ID` -> I added a patch for `getHttpUriForMxc` in the `matrix-js-sdk`. Ideally we should upgrade `matrix-js-sdk` which has the right endpoints but since newer versions have breaking changes, and our current version is heavily patched, upgrading it would probably be quite time consuming (would be good to plan this some time in the future, especially after `simplified_msc3575` is released as mentioned in the linked PR)

Before:
image

After:
image


lukemelia (approved):

backspace (approved):
Could an end-to-end test have caught the breaking change when the SDK was updated? 😳

jurgenwerk:
@backspace I believe so! I made a ticket for that and we can discuss at next planning.

PR #2356, opened 2025-03-27
realm: Remove Node file watcher

backspace:
This adds a `ENABLE_FILE_WATCHER` environment variable when starting the realm server that’s passed into `NodeAdapter`s as they’re constructed. If it’s not `true`, the `_message` subscription endpoint will be inert. In the future, `_message` (a relic of SSE) will be removed and the watcher will start automatically with the adapter’s construction.

There’s a new realm server `file-watcher-events-test` to ensure filesystem events are propagated as Matrix realm events. Alongside that I extracted a `setupMatrixRoom` helper and will return for another extraction.

To exercise this, first observe the default development setup:

1. Start everything as usual in development.
2. Visit a local file in the web interface.
3. Edit the file directly on the filesystem.
4. Observe that the change is reflected on the web.

Then turn off the watcher and observe the difference:

1. Edit `start-development.sh` to remove `ENABLE_FILE_WATCHER=true` and start the realm server again.
2. Visit a local file in the web interface.
3. Edit the file directly on the filesystem.
4. Observe that the change is _not_ reflected on the web.


.github/workflows/ci.yaml
```diff
@@ -293,10 +294,6 @@ jobs:
       - name: create realm users
         run: pnpm register-realm-users
         working-directory: packages/matrix
-      # TODO: remove with CS-8104
-      - name: Install watchexec
-        run: cargo install --locked watchexec-cli
-        if: ${{ matrix.testModule == 'realm-endpoints-test.ts' }}
```
backspace, line 299: I found that the test-hanging problem I originally added this workaround for didn’t happen when I used `node` as the `sane` file-watcher instead of `watchexec`, so this is no longer needed. The `testFileWatcher` is still in use though, so 8104 stays open.

packages/realm-server/tests/card-endpoints-test.ts
```diff
@@ -99,70 +105,14 @@ module(basename(__filename), function () {
             fileSystem,
           }));
 
-          request = supertest(testRealmHttpServer);
+          setTestRequest(supertest(testRealmHttpServer));
```
backspace, line 108: This callback formation is needed so `setupMatrixRoom` later in the file can be extracted as a test helper, it relies on this `request` but that’s not set until the `beforeEach` has executed.

`setupPermissionedRealm` is repeated in 13 files, I‘ll open a followup PR to extract it, with named properties as it’s getting to be That Time

habdelra (approved):
:+1:

PR #2358, opened 2025-03-27
Add copy source feature in card service

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8223/copy-gts-card-command-for-catalog

### What is changing
This PR introduces a new copy source service (e.g. copy .gts file) that can copy one file to another workspace.

There is a use case example & screenshot in this draft PR for catalog app: https://github.com/cardstack/boxel/pull/2351


backspace:
(I reran the host tests to hopefully remove the spurious “new” three Percy diffs, after I approved the restored snapshots on `main`, but now it’s worse 😖)

Should this have a test? The code looks good but it seems wise to exercise it via a test, but maybe there‘s context I’m unaware of.

richardhjtan:
> (I reran the host tests to hopefully remove the spurious “new” three Percy diffs, after I approved the restored snapshots on main, but now it’s worse 😖) 

After merge to latest main it seems okay now!

> Should this have a test? The code looks good but it seems wise to exercise it via a test, but maybe there‘s context I’m unaware of.

Thanks for bringing out @backspace , I included a command service for copy source and added a test as well in latest changes

backspace (approved):
Thanks for adding the test 😀

PR #2359, opened 2025-03-27
Index should include polymorphic field modules in deps

ef4:
Previously, the index entry for a card instance only depended on the module in its own `adoptsFrom` and the graph of modules consumed from there.

But cards with polymorphic contains or containsMany have other directly `meta.fields.*.adoptsFrom` references that also need to go into deps.


habdelra (approved):
looks good. I'll push this over the finish line

burieberry:
> looks good. I'll push this over the finish line

Thanks, I couldn't get to this due to other work

burieberry:
Whoops, I don't know how that happened

PR #2360, opened 2025-03-28
CS-8204-filter-create-catalog-app-tags-group-filter

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8204/filter-create-catalog-app-tags-group-filter

Result:
image

### DIFF
--- a/packages/catalog-realm/catalog-app/components/filter-section.gts
+++ b/packages/catalog-realm/catalog-app/components/filter-section.gts
@@ -27,22 +27,20 @@ export class FilterCategoryGroup extends GlimmerComponent<FilterCategoryGroupArg
 
   <template>
     <FilterGroupWrapper @title={{@title}} ...attributes>
-      <:filterList>
-        <div class='filter-list'>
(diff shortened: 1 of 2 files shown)


packages/catalog-realm/catalog-app/components/filter-section.gts
```diff
@@ -80,3 +73,106 @@ export class FilterCategoryGroup extends GlimmerComponent<FilterCategoryGroupArg
...
+}
+
+// FilterGroupWrapper
+interface FilterGroupWrapperArgs {
+  Args: {
+    title: string;
+  };
+  Element: HTMLElement;
+  Blocks: {
+    filterList: [];
```
tintinthong: just call this content or something generic

tintinthong (approved):

packages/catalog-realm/catalog-app/components/filter-section.gts
lucaslyl: sure, i think i will just yield to default for this scenario

PR #2363, opened 2025-03-28
Add "Ask AI" text box

burieberry:
Sends message to a new chat room and opens the room.

ask-ai-text-box

To Do items from design team review:
- [x] Remove disclaimer text
- [x] current open cards/files are attached by default to sent message

Follow up todos: 
- improve new room loading state as needed
- Convert search sheet trigger to a button


jurgenwerk:
Spinner has a slight offset
image

There is a state where the disclaimer and input have an overlapping conflict
image

Does it make sense to show the Ask AI input when the panel is already open?
image

packages/boxel-ui/addon/raw-icons/chevron-right.svg
backspace, line 1: There’s a `chevron-right` [in `boxel-icons`](https://github.com/cardstack/boxel/blob/main/packages/boxel-icons/src/icons/chevron-right.gts), could it work instead? I’m unsure if we have guidelines about when to use `boxel-icons` vs adding them to `boxel-ui`.

packages/host/tests/integration/components/ask-ai-test.gts
```diff
@@ -0,0 +1,270 @@
...
+        from: 'testuser',
+        message: 'Hello world',
+        cards: [{ id: cardId, title: 'Marco' }],
+      },
+    ]);
+    assert.dom('[data-test-ask-ai-input]').hasNoValue();
+    assert.dom('[data-test-message-field]').hasNoValue();
+  });
+
+  test('can send message to AI Assistant from workspace chooser', async function (assert) {
```
backspace, line 152: Seeing this test made me try it out with the chooser up:

Amara  Osei 2025-04-02 04-54-57

But I see that it’s the same in staging so I guess this is expected, I just am newly aware, it seems weird to me, but outside the scope of your PR.

packages/host/tests/integration/components/ask-ai-test.gts
```diff
@@ -0,0 +1,270 @@
...
+        from: 'testuser',
+        message: 'Goodbye',
+        cards: [{ id: marcoId, title: 'Marco' }],
+        files: [{ sourceUrl: petCardId, name: 'pet.gts' }],
+      },
+    ]);
+    await click('[data-test-past-sessions-button]');
+    assert.dom('[data-test-joined-room]').exists({ count: 2 });
+    await click('[data-test-past-sessions-button]');
+  });
```
backspace, line 269: When I tried this in code submode the panel didn’t open:

screencast 2025-04-02 04-59-19

but maybe this is because I’m using the preview deployment? I see a change in `packages/base/command.gts` that wouldn’t be in that deployment.

backspace (approved):
This looks good as long as the code submode problem I mentioned is only because I wasn’t using it locally 🎉

packages/boxel-ui/addon/raw-icons/chevron-right.svg
burieberry: I know for cards in realms we can use boxel-icons. For host, we don't use it and I copied this icon directly from boxel-icons into boxel-ui.

backspace: ah understood, thanks!

packages/host/app/services/operator-mode-state-service.ts
```diff
@@ -302,6 +311,39 @@ export default class OperatorModeStateService extends Service {
...
+
+  getOpenCards = restartableTask(async (selectedCardRef?: ResolvedCodeRef) => {
+    let cardIds = this.getOpenCardIds(selectedCardRef);
+    if (!cardIds) {
+      return;
+    }
+    let cards = (await Promise.all(cardIds.map((id) => this.store.peek(id))))
+      .filter(Boolean)
+      .filter(isCardInstance);
+    return cards;
```
burieberry, line 337: @habdelra is the store usage here correct in this context? I have the card ids, the existing SendAiAssistantMessage command input is expecting card defs

burieberry:
> Spinner has a slight offset 
> 
> There is a state where the disclaimer and input have an overlapping conflict
> 
> Does it make sense to show the Ask AI input when the panel is already open? 

Thanks, I fixed the spinner alignment. Chris said it's ok to remove the disclaimer, so it is removed now. There are designs to make the search sheet textbox into a button, which will help to prevent overlap as well. I'll make a ticket for this. Yes, according to design the input will still be present when panel is open. Sending a message from using the input will start a new session.

PR #2369, opened 2025-04-01
Making format changes not dependent on save completion in operator mode

habdelra:
This also addresses more card instances operating outside of the jurisdiction of the store.

TODO:
- [x] Move CardService.saveModel() into the Store service as private
- [x] Move CardService.reloadModel() into the Store service as private
- [x] Move CardService.patchCard() into Store service as public API

save


packages/host/app/commands/reload-card.ts
habdelra, line 1: This should no longer be necessary since we have an identity map driving our card instances

packages/host/app/resources/room.ts
```diff
@@ -379,6 +379,9 @@ export class RoomResource extends Resource<Args> {
     return this.allSkillEventIds.has(eventId);
   }
 
+  // TODO we should think about why we are caching these running instances
+  // outside of the store. it would be a good idea to bring these into the fold
+  // so that we can make the `createFromSerialized()` more private.
   private async ensureSkillCardCached(eventId: string) {
```
habdelra, line 385: FYI

packages/host/app/services/card-service.ts
```diff
@@ -380,10 +380,8 @@ export default class CardService extends Service {
         (card as any)[field] = value;
       }
     }
-    let updatedCard = await api.updateFromSerialized<typeof CardDef>(card, doc);
-    // TODO setting `this` as an owner until we can have a better solution here...
-    // (currently only used by the AI bot to patch cards from chat)
-    return await this.saveModel(updatedCard);
+    await api.updateFromSerialized<typeof CardDef>(card, doc);
```
habdelra: hmmm, this needs to using the identity map. let's move patchCard into the StoreService.

packages/host/app/services/store.ts
```diff
@@ -211,15 +209,34 @@ export default class StoreService extends Service {
...
 
+  save(id: string) {
+    this.initiateAutoSaveTask.perform(id, { isImmediate: true });
+  }
+
+  // Instances that are saved via this method are eligible for garbage
+  // collection--meaning that it will be detached from the store. This means you
+  // MUST consume the instance IMMEDIATELY! it should not live in the state of
+  // the consumer.
+  async saveInstance(instance: CardDef, realm?: string) {
```
habdelra: this should be used in lieu of CardService.saveModel()

packages/host/app/services/store.ts
```diff
@@ -231,7 +248,32 @@ export default class StoreService extends Service {
...
+    let api = await this.cardService.getAPI();
+    let deps = getDeps(api, instance);
+    for (let dep of deps) {
+      if (dep.id && this.identityContext.get(dep.id) !== dep) {
+        throw new Error(
+          `encountered a dependency, ${dep.id}, of ${instance.id} when adding ${instance.id} to the store, but the store ${
+            this.identityContext.get(dep.id)
+              ? "already has a different instance with this dependency's id"
+              : "has no entry for this dependency's id"
+          }. Please obtain the instances that already exists from the store`,
```
habdelra: in the case where the dep just doesn't exist yet in the identity map we should just add it instead of throwing.

packages/host/tests/acceptance/code-submode/inspector-test.ts
```diff
@@ -850,7 +850,7 @@ module('Acceptance | code submode | inspector tests', function (hooks) {
   });
 
   test<TestContextWithSave>('can duplicate an instance in same realm', async function (assert) {
-    assert.expect(10);
+    assert.expect(7);
```
habdelra, line 853: an unnecessary save was being issued which doubled up the onSave assertions, fixed.

packages/host/tests/acceptance/code-submode/inspector-test.ts
```diff
@@ -905,7 +905,7 @@ module('Acceptance | code submode | inspector tests', function (hooks) {
   });
 
   test<TestContextWithSave>('can duplicate an instance in different realm', async function (assert) {
-    assert.expect(11);
+    assert.expect(8);
```
habdelra, line 908: an unnecessary save was being issued which doubled up the onSave assertions, fixed.

packages/host/app/services/store.ts
```diff
@@ -231,7 +300,126 @@ export default class StoreService extends Service {
...
+
+  // This method is used for specific scenarios where you just want an instance
+  // that is not auto saving and not receiving live updates and is eligible for
+  // garbage collection--meaning that it will be detached from the store. This
+  // means you MUST consume the instance IMMEDIATELY! it should not live in the
+  // state of the consumer.
+  async search(query: Query, realmURL: URL): Promise<CardDef[]> {
+    await this.ready;
+    let json = await this.cardService.fetchJSON(
+      `${realmURL}_search?${stringify(query)}`,
```
richardhjtan: Are we allowed to search the query with `null` value within this case?

By default from qs stringify,`null` will become an empty string value unless we specify `strictNullHandling: true`

habdelra: good observation. I can update that

packages/host/app/services/store.ts
```diff
@@ -204,20 +245,75 @@ export default class StoreService extends Service {
...
+  }
+
+  // Instances that are saved via this method are eligible for garbage
+  // collection--meaning that it will be detached from the store. This means you
+  // MUST consume the instance IMMEDIATELY! it should not live in the state of
+  // the consumer.
+
+  // This method adds or creates a new instance to the store and returns an
+  // instance eligible for garbage collection.
+  async add<T extends CardDef>(
```
habdelra, line 288: this should be used in lieu of CardService.saveModel() and CardService.createFromSerialized()

PR #2370, opened 2025-04-02
realm: Extract realm server test duplication

backspace:
This extracts `setupPermissionedRealm` with `onRealmSetup` and `getRealmSetup` callbacks for containing test realm server properties needed by future hooks (only `setupMatrixRoom` for now). It also extracts `createJWT`, some shared variables, and removes some repetitive logging.

### DIFF
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -1,16 +1,15 @@
 import { module, test } from 'qunit';
-import supertest, { Test, SuperTest } from 'supertest';
-import { join, resolve, basename } from 'path';
+import { Test, SuperTest } from 'supertest';
+import { join, basename } from 'path';
 import { Server } from 'http';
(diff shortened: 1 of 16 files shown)


richardhjtan:
Question: There is an intention to remove `setGracefulCleanup` during the tests, is it because we have a reusable `dir` from the new helper?

richardhjtan (approved):
Nice refactor ✨

jurgenwerk (approved):

backspace:
> Question: There is an intention to remove `setGracefulCleanup` during the tests, is it because we have a reusable `dir` from the new helper?

I included `setGracefulCleanup` [here](https://github.com/cardstack/boxel/pull/2370/files#diff-f3ff005d711126a906edde86f1f57fdb06c296f97f96020f61471059d4fc3529R767), I think that means it’ll run anytime the temporary directory is created. But hopefully I didn’t miss anything!

richardhjtan:
Ah my bad! It looks good 👍

PR #2372, opened 2025-04-02
Remove SSE endpoint, remaining code, references

backspace:
This removes the vestigial `_message` endpoint that was still being used to start file-watching in some realms. Now that starts (when desired, in development by default) as part of the realm’s other calls in `start`.

With the final trace of SSE infrastructure being removed, it also made sense to remove the feature flags related to toggling back to SSE for realm events. This also includes comment and documentation updates to reflect that SSE is gone.


packages/runtime-common/realm.ts
```diff
@@ -1943,78 +1932,23 @@ export class Realm {
...
-          (w) => w !== writable,
-        );
-        if (this.listeningClients.length === 0) {
-          this.#adapter.unsubscribe();
-        }
-      },
-    );
-
-    // To remove as part of CS-8104, a workaround for a file watcher irregularity in tests
-    let subscribeOptions: { watcher?: string } = {};
```
backspace, line 1976: The workaround of using a different file-watcher to fix a hanging test is no longer needed, as the watcher now shuts down properly when the realm is disposed of.

habdelra (approved):
:tada:

jurgenwerk (approved):
nice!

packages/runtime-common/realm.ts
```diff
@@ -1943,78 +1932,23 @@ export class Realm {
...
-        this.drainUpdates();
-      }, subscribeOptions);
-    }
-
-    this.listeningClients.push(writable);
-
-    if (testFileWatcher) {
-      this.sendServerEvent({
-        type: 'test-file-watcher-ready',
+  private async startFileWatcher() {
```
habdelra, line 1935: @backspace  is there any reason why this should be async? usually the pattern is that a synchronous method calls an async "drain" method...

backspace: `start()` is async though?

habdelra: everything in this method is being called synchronously, no? I don't see any awaits in your function. why should it be async?

packages/runtime-common/realm.ts
```diff
@@ -425,6 +409,11 @@ export class Realm {
 
   async start() {
     this.#startedUp.fulfill((() => this.#startup())());
+
+    if (this.#adapter.fileWatcherEnabled) {
+      await this.startFileWatcher();
```
habdelra, line 414: there is no need to await this--its synchronous

packages/runtime-common/realm.ts
backspace: eep that’s actually a bug, it should be `await this.#adapter.subscribe(…`

habdelra: The `Realm.start()` is async because it's awaiting other things--but not this.

habdelra: ah, so there is async in here, then we are just missing the await

PR #2373, opened 2025-04-03
Add ability to copy a skill instance

tintinthong:
Now you should be able to copy cursor coding skill

### DIFF
--- a/packages/catalog-realm/catalog-app/listing/listing.gts
+++ b/packages/catalog-realm/catalog-app/listing/listing.gts
@@ -9,6 +9,7 @@ import {
 } from 'https://cardstack.com/base/card-api';
 import MarkdownField from 'https://cardstack.com/base/markdown';
 import { Spec, type SpecType } from 'https://cardstack.com/base/spec';
+import { SkillCard } from 'https://cardstack.com/base/skill-card';
 
 import { action } from '@ember/object';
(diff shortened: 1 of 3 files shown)


richardhjtan (approved):
Looks good, tested and it works

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -487,6 +490,7 @@ export default class InteractSubmode extends Component<Signature> {
     },
   );
 
+  //==catalog actions==
```
richardhjtan, line 493: Perhaps we can put something like this, not necessary to do now but I may help after that

// ===== catalog actions =====
...
// ===== end of catalog actions =====

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -76,18 +77,32 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
   }
 
+  get hasSkills() {
+    return (
+      this.args.model instanceof SkillListing &&
+      this.args.model?.skills?.length > 0
+    );
+  }
+
   get createButtonDisabled() {
```
richardhjtan, line 101: Let's anticipate this getter whether it can be refactor when logic grows up

PR #2376, opened 2025-04-03
Catalog listing install/fork action

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8292/implement-installfork

### What is changing
- When click add to workspace, also copy files based on code ref per spec that attached in a listing
- Copy source command action to be included in operator mode

**Screenshot**
https://github.com/user-attachments/assets/eddfc5e8-00e5-416e-83d6-6d08a7f26788


packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -70,13 +70,37 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
+      ),
     );
+
+    // Then, copy the gts file based on the attached spec's moduleHref
+    await Promise.all(
+      specsToCopy.map((spec: Spec) => {
+        const absoluteModulePath = spec.moduleHref;
+        const relativeModulePath = spec.ref.module;
+        // Remove the leading "../" to get the path relative to the target realm
+        const normalizedPath = relativeModulePath.replace(/^\.\.\//, '');
```
tintinthong: wudn't this code pattern be something we have already done in the past? The `.replace` seems unnecessary.

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -70,13 +70,37 @@ class EmbeddedTemplate extends Component<typeof Listing> {
   });
 
   _create = task(async (realmUrl: string) => {
```
tintinthong: can u fill all the required specs in the blogApp example so we can prove that this works

packages/catalog-realm/catalog-app/listing/listing.gts
richardhjtan: Yes I noticed the `pop` method in the codebase, I will change that

tintinthong (approved):

tintinthong:
Uhmmm is this new test failures with the code-ref. Looks like it

tintinthong (changes requested):
Pls resolve the test errors. Afterwards might be difficult to find

PR #2377, opened 2025-04-03
Improve source code editing skill to have better chance of including the file url

jurgenwerk:
Improve skill to greatly improve the chances of the model responding with a search/replace block that **includes the attached file url**. Previously it was happening that sometimes the url of the attached file was missing and our patching tool did not know the address of the file to patch, making our "apply code patch" feature not working. 

Better diff (using https://www.diffchecker.com/):
image


backspace (approved):
Thanks for including the better diff, much easier to understand what changed!

PR #2379, opened 2025-04-03
Do not run consumeContext helper after each render

burieberry:


habdelra (approved):
Thanks! looks good :+1:

burieberry:
> Thanks! looks good 👍

@ef4 wrote up this ticket for some more card-resource improvements https://linear.app/cardstack/issue/CS-8313/card-resource-still-does-too-much-synchronization-work

PR #2380, opened 2025-04-03
Add throttle to sliding sync matrix SDK logic

habdelra:
So it turns out that there is a different module in the matrix SDK that controls the sliding sync that we need to patch. I opted for a 2 second throttle, if folks feel that is too slow I can change it. I'm trying to balance respect for server resources with user experience. 

I added a console.log in this example so you can see the actual sync time. It be awesome to figure out how to make it stop this tight loop, but i think that's a synapse server bug. this is actual issue: https://github.com/matrix-org/synapse/issues/15824 which was migrated to here https://github.com/element-hq/synapse/issues/15824

https://github.com/user-attachments/assets/7b45d171-adba-4436-b145-019e6f74878c


habdelra:
This actually doesn't work. the sliding sync logic is different enough that the basic throttle that we used in the past breaks things horribly.... The failed matrix tests reveal this issue...

PR #2383, opened 2025-04-04
Use HTTP caching to prevent unnecessary network requests in loader

habdelra:
This PR introduces the use of ETag cache control to use HTTP caching in our loader so that when the loader needs to refresh we don't re-download modules that have not changed. 

We are issuing the correct ETag cache control headers to cloudfront as well when we fetch icons, but it looks like cloudfront is not honoring those. Rather, i think that cloudfront handles etags between the origin and the edge cache, not between the edge cache and the client. probably we should update cloudfront to request that browsers cache the icons for a long time. thoughts? (/cc @lukemelia )

https://github.com/user-attachments/assets/f7649341-e959-475b-94cd-8f9e7c0a511e


jurgenwerk (approved):

habdelra:
Note: added ticket https://linear.app/cardstack/issue/CS-8320/define-caching-rule-for-icons-on-cloudfront-to-cache-in-the-browser to handle browser caching of icons

backspace (approved):

PR #2385, opened 2025-04-07
Allow copy examples

tintinthong:


packages/host/app/services/card-service.ts
```diff
@@ -452,10 +452,10 @@ export default class CardService extends Service {
 
   private async saveCardDocument(
     doc: LooseSingleCardDocument,
-    realmUrl: URL,
+    destinationURL: URL, //directory
```
richardhjtan: is the comment needed

packages/host/app/services/card-service.ts
```diff
@@ -468,13 +468,16 @@ export default class CardService extends Service {
     return json;
   }
 
-  async copyCard(source: CardDef, destinationRealm: URL): Promise<CardDef> {
+  async copyCard(
+    source: CardDef,
+    destinationURL: URL, //directory
```
richardhjtan: also if this comment is intended

richardhjtan (approved):

PR #2387, opened 2025-04-07
Fix: nested pre tags issue in code block

FadhlanR:
https://github.com/user-attachments/assets/a79570a1-1f2c-43cf-b1dd-f418f8128271

**Issue**
There was an issue where code blocks couldn't display code properly when the code included nested `<pre>` tags. The previous implementation used regex to find content to display in code blocks, but it only found the first opening `<pre>` tag and the first closing `<pre>` tag. This approach failed to handle nested `<pre>` tags correctly.

**Solution**
In this PR, I've updated the way we identify code block content by using document.createElement to find the most outer pre tag as suggested by @jurgenwerk .


jurgenwerk (commented):
(ignore this, I pasted my feedback in the follow up review)

packages/host/app/components/ai-assistant/formatted-message.gts
```diff
@@ -249,25 +249,93 @@ class HtmlDidUpdate extends Modifier<HtmlDidUpdateSignature> {
...
-      // This is non <pre> tag HTML
-      result.push({
-        type: 'non_pre_tag',
-        content: match[2],
-      });
+  let result: HtmlTagGroup[] = [];
+  let tagStack: { tag: string; startPos: number }[] = [];
+  let currentPosition = 0;
+
+  let findNextTag = (
```
jurgenwerk: This parsing is getting quite big and complicated, move from formatted message component to a html utils file

packages/host/app/components/ai-assistant/formatted-message.gts
```diff
@@ -249,25 +249,93 @@ class HtmlDidUpdate extends Modifier<HtmlDidUpdateSignature> {
...
+      if (openTag === -1) return null;
+
+      if (htmlString.startsWith('<<<<<<<', openTag)) {
+        let endMarker = htmlString.indexOf('>>>>>>>', openTag);
+        if (endMarker === -1) {
+          currentPos = openTag + 7; // length of '<<<<<<<'
+          continue;
+        }
+        // Skip past the entire search/replace block
+        currentPos = endMarker + 7; // length of '>>>>>>>'
```
jurgenwerk: I am confused why you are dealing with search/replace blocks here. Parsing of the search/replace blocks is already done `extractCodeData`

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -279,4 +279,103 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
...
+<li>Styled <code>&lt;pre&gt;</code> block</li>
+<li>HTML code display using entities</li>
+</ol>
+<p>You can use these in your HTML documents to display formatted text, code snippets, or any content where you want to preserve spacing and line breaks exactly as written.</p>`;
+
+    simulateRemoteMessage(
+      roomId,
+      '@aibot:localhost',
+      {
+        body: complexHtmlMessage,
```
jurgenwerk: Other test examples in this file seem to have `body`s in markdown, for example `body: 'This is a code snippet that I made for you\n```javascript\nconsole.log("hello world");\n```\nWhat do you think about it?',`

Is there a specific reason you are using html here? AFAIK we are using `markdownToHtml` when the message is passed to the `AiAssistantMessage` I think you can use markdown here in order to fully test the data processing flow

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -279,4 +279,103 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
 
     await percySnapshot(assert);
   });
+
+  test('handles nested pre tags', async function (assert) {
+    let roomId = await renderAiAssistantPanel(`${testRealmURL}Person/fadhlan`);
+    let complexHtmlMessage = `<p>I&#39;ll help you create an example with pre tags in HTML. Here&#39;s a useful example that demonstrates different ways to use the pre tag:</p>
```
jurgenwerk: How come content is escaped here? For example why `Here&#39;s` instead of `Here's`

Is this how the bot is sending the message in the matrix message? Does it not send it in a markdown format? We should simulate that format as closely as possible

packages/host/app/lib/formatted-message/utils.ts
```diff
@@ -7,31 +7,61 @@ import {
...
+      code: null,
+      language: null,
+      searchReplaceBlock: null,
+    };
+  }
 
-  if (!content) {
-    return emptyCodeData;
+  // Find the matching closing tag considering nesting
+  let closingTagStart = findMatchingClosingTag(preElementString, openTagEnd);
```
jurgenwerk: Since this is now getting a bit more complicated, did you consider an alternative of creating an element using `document.createElement`, setting the html to it and use `querySelectorAll` to get the inner parts of the 1st level `<pre>`s? Wondering if that would be easier since the browser probably already has the functionality that you want.

jurgenwerk (changes requested):
In your screen video you have escaped html code that looks like `&lt;head&gt;`..., shouldn't that be displayed with proper formatting in a human readable way? Also would be good to write a top level description of what the issue was and how you solved it. It makes reviewing easier.

FadhlanR:
@jurgenwerk Thanks for the suggestion to use `document.createElement`. I've updated it please check it again.

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -279,4 +279,115 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
...
+const data = {
+    name: "John",
+    age: 30,
+    city: "New York"
+};
+</pre>
+
+<!-- Pre tag with HTML entities -->
+<pre>
+&lt;html&gt;
```
jurgenwerk: Is this supposed to be `<html>...` and not `&lt;html&gt;...`? And similarly for the rest of the tags

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -279,4 +279,115 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
...
+      },
+    );
+
+    await waitFor('[data-test-message-idx="0"]');
+
+    assert
+      .dom('.monaco-editor')
+      .exists(
+        { count: 2 },
+        'Should only have one monaco editor for the outer pre tag',
```
jurgenwerk: Count in assertion is 2 but message says "should have only one monaco editor". Which one is correct?

jurgenwerk (changes requested):
There's a bug, the diff editor is not showing diffs

Main:
image

This branch:
image

It's weird the tests didn't catch this, perhaps we need to improve them

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
FadhlanR: I copied this example from ai bot response and it seems everything under the nested pre tag should be treated as text that is why the ai bot return `&lt;html&gt;` instead of `<html>`.

packages/ai-bot/lib/matrix.ts
```diff
@@ -60,12 +67,13 @@ export async function sendMessageEvent(
   commandRequests: Partial<CommandRequest>[] = [],
   reasoning: string | undefined = undefined,
 ) {
+  let html = markdownToHtml(escapeHtmlOutsideCodeBlocks(body));
```
FadhlanR: This is another problem. The formatted message must be in html, but ai bot didn't convert the body which is in markdown into html.

packages/host/app/components/matrix/room-message.gts
```diff
@@ -112,6 +116,19 @@ export default class RoomMessage extends Component<Signature> {
...
   });
 
+  private get formattedMessage() {
+    if (
+      !this.message.formattedMessage ||
+      isHtml(this.message.formattedMessage)
+    ) {
+      return this.message.formattedMessage;
+    }
+    return markdownToHtml(
```
FadhlanR, line 130: The formatted message must be already in html. However, this handle the old cases where ai bot sent formatted message in markdown.

packages/runtime-common/dompurify-runtime.ts
```diff
@@ -1,15 +1,22 @@
 import DOMPurify from 'dompurify';
 
-let domPurify: DOMPurify.DOMPurifyI;
+let _domPurify: DOMPurify.DOMPurifyI | undefined = undefined;
+let customJsdom: any = null;
+
+export function setCustomJsdom(jsdomInstance: any) {
```
FadhlanR, line 6: We need to set the jsdom from the client since this can be used from browser and node js.

FadhlanR:
I fixed the bug @jurgenwerk found in this PR. It occurred because the AI bot sent the formatted message in Markdown instead of HTML. I added code to convert the message from Markdown to HTML before the AI bot sends it, and I also updated the code to handle cases where the formatted message is still in Markdown. This PR is ready for another review.

packages/host/app/lib/formatted-message/utils.ts
```diff
@@ -8,28 +8,26 @@ import {
...
-  let content = contentMatch ? contentMatch[1] : null;
-
-  if (!content) {
-    return emptyCodeData;
-  }
+  let language = preElement.getAttribute('data-code-language') || 'text';
 
+  let content = preElement.innerHTML;
+  // Decode HTML entities to handle special characters like < and >
+  let textarea = document.createElement('textarea');
```
jurgenwerk: Should this element be removed after we are done getting content?

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -279,4 +279,291 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
...
+    );
+
+    await waitFor('[data-test-message-idx="0"]');
+
+    assert
+      .dom('.monaco-editor')
+      .exists({ count: 2 }, 'Should have 2 monaco editors');
+
+    let monacoContent = getMonacoContent();
+    assert.ok(
```
jurgenwerk: I think a better test would be to just assert the whole content instead of many `.includes()`

packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -279,4 +279,291 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
...
+        origin_server_ts: new Date(2024, 0, 3, 12, 30).getTime(),
+      },
+    );
+
+    await waitFor('[data-test-message-idx="0"]');
+
+    // Check that HTML outside backticks is displayed as actual HTML
+    assert
+      .dom('[data-test-ai-message-content]')
+      .containsText(
```
jurgenwerk: Same here, I think it's a clearer and better test if you assert the whole content string, instead of many `containsText()`s

jurgenwerk (changes requested):
I am testing your branch locally and this is what I see

https://github.com/user-attachments/assets/13e6636b-704a-42ef-bab1-33fb91efa102

packages/host/app/lib/formatted-message/utils.ts
```diff
@@ -110,3 +108,44 @@ export function wrapLastTextNodeInStreamingTextSpan(
...
+
+export interface HtmlTagGroup {
+  type: 'pre_tag' | 'non_pre_tag';
+  content: string;
+}
+
+export function parseHtmlContent(htmlString: string): HtmlTagGroup[] {
+  let result: HtmlTagGroup[] = [];
+
+  let doc = document.createElement('div');
```
jurgenwerk, line 131: This one too - does this element need to be removed after working with it?

jurgenwerk, line 131: I think it would be helpful for other people who will work on this to add a comment explaining why exactly we are creating an element instead of just parsing the string directly (the way it was done previously)

jurgenwerk:
I'm testing this locally and can reliably see bugs where the code diff is not displayed. Observe the first code block, there is no code diff:

https://github.com/user-attachments/assets/cd9dba4f-3e05-4bd7-8e1b-581de36a7548

FadhlanR:
> I'm testing this locally and can reliably see bugs where the code diff is not displayed. Observe the first code block, there is no code diff:
> 
>  Screen.Recording.2025-04-14.at.11.06.34.mov

@jurgenwerk fixed.


https://github.com/user-attachments/assets/78dc2e95-747d-4b27-95fb-2cf4fdf122bd


https://github.com/user-attachments/assets/034124a6-07d1-49af-ace7-f11139443f4f

jurgenwerk:
I'm checking your 1st video and this (circled in red) is not supposed to show up in the editor. It means the search/replace block detection probably failed. 
 
image

FadhlanR:
> I'm checking your 1st video and this (circled in red) is not supposed to show up in the editor. It means the search/replace detection probably failed.
> 
> image

I asked AI to generate example code with `<pre>` tag and AI response with code with random file url. It seems the code-snippet.gts file is not found, what is expected behaviour if the file is not found?

jurgenwerk:
@FadhlanR We don't have support for creating new files from within the assistant. There is a ticket for creating new files in this sprint. I suggest you make a video of a real use case instead that demonstrates succesful handling on nested `pre`s (something like from the repro screenshots in the ticket)

FadhlanR:
> @FadhlanR We don't have support for creating new files from within the assistant. There is a ticket for creating new files in this sprint. I suggest you make a video of a real use case instead that demonstrates succesful handling on nested `pre`s (something like from the repro screenshots in the ticket)

@jurgenwerk here's the screenshot for the search and replace block that includes `<pre>` tag.

https://github.com/user-attachments/assets/4e8a1cf2-42a2-44ba-824e-4d6e7872322f

packages/host/app/lib/formatted-message/utils.ts
```diff
@@ -117,27 +119,40 @@ export interface HtmlTagGroup {
...
-    } else if (match[2] && match[2].trim() !== '') {
-      // This is non <pre> tag HTML
-      result.push({
-        type: 'non_pre_tag',
-        content: match[2],
-      });
+  let result: HtmlTagGroup[] = [];
+
+  // Using dom manipulation to find pre and non-pre tags
+  // because we need to avoid the nested html tags issue.
```
jurgenwerk:
```suggestion
  // We are creating a new element in the dom so that we can easily parse the content of the top level <pre> tags. Note that <pre> elements can have nested <pre> elements inside them and by querying the dom like that it's trivial to get its contents, compared to parsing the htmlString. 
```

jurgenwerk (approved):
I tested this again and didn't find any more issues. It was a wild ride but you now have pretty good understanding of how code streaming and code patching works which is awesome so thanks for going through this. I have some more comments:

1. You are using `document.createElement('div')` twice. Is it possible to only do it once? One case is for the nested pre tags. But it's not clear to me why there is another case for this. Try to consolidate if possible, or at least make a utility function for creating a new dom element end getting the content out, or at least improve the comment in the 2nd use case.

2. After merging, be on the lookout for bugs that Chris reports in the AI channel and see if this might have caused any regressions

3. Since there are now many fix/revert commits maybe squash all the commits into one or two to reduce all the noise

lukemelia (approved):

PR #2388, opened 2025-04-07
AI assistant: Allow uploading module code that has an error 

jurgenwerk:
The issue was that when there was an error in the code, the AI bot failed to download the file, so it did not include it in the prompt context. This meant the assistant was not able to fix the code if you asked it to. Here is how this looked (screenshot by @backspace):

image

### Before

There was a discrepancy between the content type between code that compiles ok and code that has en error. You can observe that here:

image

### After

My fix includes consolidating the content type in case when the code has an error. Coupled with an adjustment in the ai bot code, this fixes the bug of the ai bot not being able to download error source code. 

image


backspace (approved):
Nice sleuthing!

lukemelia (approved):

lukemelia (approved):

richardhjtan (approved):

PR #2390, opened 2025-04-07
Display "missing template" message for missing fitted field template in playground preview

burieberry:
missing-fitted


backspace (approved):
I couldn’t figure out how to reproduce this but the screenshot looks good

PR #2398, opened 2025-04-08
Update store to reason in local ID's

habdelra:
This PR updates the StoreService to reason about cards using the local ID's. The meat of the work is in the identity context class, as well as renaming params and vars in the Store to reflect the fact we are dealing in "id's" and not necessarily "url's" (also to be more clear about the fact that we deal in "instances" of cards). In order to keep things from getting too large I'll file subsequent PR's for:
-  serialization of unsaved cards, 
- handling local ID's in the serialized payload from the server,
- updating StackItems to use local id. 
- Also, before I get much farther, I will file another PR that tests the store service directly instead of relying on testing indirectly thru via UI tests.


packages/host/tests/integration/store-test.ts
```diff
@@ -0,0 +1,93 @@
...
+  // createSubscriber
+  // unloadResource
+  // create
+  // add
+  // save
+  // peek
+  // delete
+  // patch
+  // search
+  // getSaveState
```
habdelra, line 92: right now all of the store functionality is tested indirectly thru all the other acceptance and integration tests. As I move forward I'd like to test it more directly here as it's really painful debugging store related issues from indirect testing. In order to keep this PR manageable i'm going to follow up in another PR to add Store tests

packages/host/app/services/store.ts
```diff
@@ -117,18 +129,19 @@ export default class StoreService extends Service implements StoreInterface {
     clearInterval(this.gcInterval);
     this.subscriptions = new Map();
     this.onSaveSubscriber = undefined;
-    this.subscribers = new Map();
+    this.remoteIdSubscribers = new Map();
```
habdelra, line 132: need to reset state for `this.localIdSubscribers` and `this.localIds` too.

packages/runtime-common/package.json
```diff
@@ -15,6 +15,7 @@
     "@babel/types": "^7.20.0",
     "@cardstack/boxel-ui": "workspace:*",
     "@cardstack/ember-template-imports": "workspace:*",
+    "@lukeed/uuid": "^2.0.1",
```
backspace, line 18: Should we use the same UUID library everywhere? This and other packages also have `uuid@9`

packages/host/app/lib/gc-identity-context.ts
```diff
@@ -53,69 +105,88 @@ export default class IdentityContextWithGarbageCollection
...
+      .map((c) => this.gcVisit(c, consumptionGraph, cache))
+      .every((result) => result);
+    cache.set(localId, result);
     return result;
   }
 
-  private makeConsumptionGraph(): Map<string, string[]> {
-    let consumptionGraph = new Map<string, string[]>();
-    for (let { card: instance } of this.#cards.values()) {
+  // this consumption graph uses local ID's
```
backspace: Is there ever a pattern where a custom type that’s just a string is used? To avoid needing this kind of comment, so developers get a hint via typing

packages/host/app/lib/gc-identity-context.ts
```diff
@@ -53,69 +105,88 @@ export default class IdentityContextWithGarbageCollection
...
   gcVisit(
-    id: string,
-    consumptionGraph: Map<string, string[]>,
-    hasSubscribers = new Map<string, boolean>(),
-  ): boolean {
-    let cached = hasSubscribers.get(id);
+    localId: string,
+    consumptionGraph: Map<string, Set<string>>,
+    cache: Map<string, boolean>,
+  ): boolean /* true = eligible for GC, false = not eligible for GC */ {
```
backspace: Would changing the name of the function make this more clear, like `checkGcEligibility`

backspace (approved):
I appreciate the thorough unit tests and the plan for more!

packages/runtime-common/package.json
habdelra: maybe? this is specifically a uuid lib that is isomorphic and uses Math.random. it's not appropriate to use that for security purposes (however this lib does have a secure v4 uuid generator too that does rely upon a crypto module in the ambient runtime). unsure if our we have any uuid needs are security related in nature.

packages/host/app/lib/gc-identity-context.ts
habdelra: we could use `type LocalId = string` to make a type alias for string that helps to document our types better

packages/host/app/lib/gc-identity-context.ts
habdelra: sure--good suggestion

richardhjtan (approved):
👍

PR #2399, opened 2025-04-08
host: Add prefixed clientRequestIds to code mode editor requests

backspace:
This updates code mode to include `clientRequestId` in `POST` requests produced by the editor saving changes. This should prevent some instances of data loss like that seen here:

screencast 2025-04-08 09-29-30

`clientRequestId` is now prefixed based on the operation that prompted the request. If the user was editing the source, it’s `editor:`. There are other prefixes for different triggering events, like `bot-patch:`, so handlers that receive realm events can decide how to act. This may be improved in the future by being separated instead of prefixed so it can use typing.

I did have assertions that source saves included `clientRequestId` but I removed them in 6b0bc14 as it seemed superficial to me. When the `clientRequestId` lacked a prefix, the `card instance change made in monaco editor is auto-saved` test [failed](https://github.com/cardstack/boxel/runs/40200702549). There’s a corresponding Matrix test included here showing that a change made via editing a card preview is reflected in the source, but it’s skipped because there’s a caching problem preventing the invalidation-prompted refetch from getting the updated source.

An earlier iteration of this PR included the removal of debounce code within the Monaco modifier related to server echoes, which seems like it should entirely be taken care of at the `clientRequestId` level. Unfortunately it [broke](https://github.com/cardstack/boxel/runs/40600832970) a cursor position test and I was unable to figure it out in time.


packages/host/app/services/card-service.ts
```diff
@@ -156,11 +156,15 @@ export default class CardService extends Service {
     return response.text();
   }
 
-  async saveSource(url: URL, content: string) {
+  async saveSource(url: URL, content: string, type?: string) {
+    let clientRequestId = `${type ?? 'source'}:${uuidv4()}`;
```
lukemelia: Maybe we shouldn't default this and instead require that a type be passed in. This would avoid new usages of saveSource that originate outside the editor introducing problems.

backspace: 👍🏻 made that change, thanks. I have some concern that the semantics of these different prefixes are easily lost as the processing of realm events is scattered everywhere and not typed, but maybe this can be revisited if it becomes confusing.

packages/host/app/services/store.ts
```diff
@@ -63,6 +63,8 @@ export { CardError, CardSaveSubscriber };
 
 let waiter = buildWaiter('store-service');
 
+// FIXME this used to use serverEchoDebounceMs, does that need to exist? What will break
```
lukemelia: Is this supposed to still be here?

packages/host/app/services/store.ts
```diff
@@ -491,13 +493,33 @@ export default class StoreService extends Service implements StoreInterface {
           !event.clientRequestId ||
           !this.cardService.clientRequestIds.has(event.clientRequestId)
         ) {
+          console.debug(`store reloading ${invalidation}`);
```
lukemelia: Did you want to use the logging system for these?

packages/host/app/services/store.ts
```diff
@@ -491,13 +493,33 @@ export default class StoreService extends Service implements StoreInterface {
...
+          }
+
           this.reloadTask.perform(instance);
         } else {
           if (this.cardService.clientRequestIds.has(event.clientRequestId)) {
-            console.debug(
-              'ignoring invalidation for card because clientRequestId is ours',
-              event,
-            );
+            if (event.clientRequestId.startsWith('editor:')) {
```
lukemelia: Seems like we want to process anything that didn't come from a rendered card, not just editor: requests. i.e. cases where commands have updated things?

lukemelia (approved):

packages/host/app/services/store.ts
backspace: it was not! I missed my traditional once-over before submitting for review, oops! Thanks, removed now.

packages/host/app/services/store.ts
backspace: Correct! Done now, thanks.

packages/host/app/services/store.ts
backspace: It’s true, I’ve updated it to now react to all non-`instance:` events, thanks.

PR #2402, opened 2025-04-08
Prevent global monaco editor state from bleeding across tests

lukemelia:


jurgenwerk (approved):
There is an open issue in the monaco version we are using: https://github.com/microsoft/monaco-editor/issues/4702

We have a patch that addresses that, which might interfere with disposals, causing bleeding of monaco editor data: https://github.com/cardstack/boxel/pull/2338/commits/7cacc9e15020696b89e07e4badd4f6a124975109

I'm hoping we could delete the mentioned patch, after https://github.com/microsoft/monaco-editor/issues/4702 is fixed, perhaps then this cleanup in this PR won't be needed anymore as well.

PR #2404, opened 2025-04-09
Skeleton catalog app

tintinthong:
Refer to https://linear.app/cardstack/issue/CS-8410/move-skeleton-catalog-app-into-feature-flagged-environment

### What is changing
- add skeleton catalog app into `catalog-realm-dev`, which under feature flagged environment
- catalog app will not be shown in production
- add new `component` spec type 

**Notes**
- some tests that check the exact number of cards have been removed, as catalog cards can grow and it's not feasible to maintain those checks


habdelra:
This is marked as "Draft", do you still want us to review it or no?

tintinthong:
> This is marked as "Draft", do you still want us to review it or no?

yes pls review it @habdelra

richardhjtan:
@habdelra I drafted it back due to adding some new tests, but there will be no significant changes to that - please feel free to review it. Thanks!

packages/runtime-common/realm.ts
```diff
@@ -299,7 +299,7 @@ export class Realm {
     this.#dbAdapter = dbAdapter;
 
     this.#router = new Router(new URL(url))
-      .post('/', SupportedMimeType.CardJson, this.createCard.bind(this))
+      .post('(/|/.+/)', SupportedMimeType.CardJson, this.createCard.bind(this))
       .patch(
```
tintinthong, line 313: @habdelra there is this api change here. It opens the realm to creating card in a directory of their choice. I made this decision bcos I did not think that introducing a new directory api was a good idea since it may encourage empty directories (which we cant delete now)

tintinthong: This is to support the use-case that we copy examples into a single folder. And not just splate it all over the realm

burieberry (commented):
- Catalog app: change all bold font to use 600 or 500 font-weight
- Let's change all official cardstack apps/components like Blog App to be published by "Cardstack" and not by individual team member names

packages/runtime-common/realm.ts
habdelra: I'm ok with this, but let's make sure that you can't escape out of the realm's file system, e.g. ../../../../etc/passwd

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -343,6 +353,28 @@ export default class InteractSubmode extends Component {
...
+        targetRealm: string,
+        directoryName?: string,
+      ): Promise<CardDef[]> => {
+        return await here._copyCards.perform(cards, targetRealm, directoryName);
+      },
+      allRealmsInfo: async () => {
+        return await here.realm.allRealmsInfo;
+      },
+    };
+    return { ...actions, ...catalogActions };
```
richardhjtan: These are the API additions for interact submode, so listing card can consume these actions in context

richardhjtan:
> * Catalog app: change all bold font to use 600 or 500 font-weight
> * Let's change all official cardstack apps/components like Blog App to be published by "Cardstack" and not by individual team member names

Good catch @burieberry , addressed them in latest commit

packages/runtime-common/realm.ts
richardhjtan: Ticketed this improvement in Linear, which can be easier to review later: https://linear.app/cardstack/issue/CS-8436/improve-file-system-handling-creating-a-card-in-a-directory

tintinthong: I think here there is already a check inside realm.handle that ensures the request url is true from `paths.inRealm`. But I will add extra test here anyway

tintinthong:
before we merge this. lets make sure we have the same copy of everything thats in catalog/

richardhjtan (approved):

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -457,6 +493,67 @@ export default class InteractSubmode extends Component {
     },
   );
 
+  //==catalog actions==
```
habdelra, line 496: can we get rid of comments like this. it's not really helpful. comments need to describe the "why" not the "what"

packages/host/app/lib/externals.ts
```diff
@@ -92,4 +92,8 @@ export function shimExternals(virtualNetwork: VirtualNetwork) {
   });
   virtualNetwork.shimModule('super-fast-md5', superFastMD5);
   virtualNetwork.shimModule('matrix-js-sdk', matrixJsSDK);
+  virtualNetwork.shimAsyncModule({
+    id: 'uuid',
+    resolve: () => import('uuid'),
+  });
```
habdelra, line 98: we actually have a uuid module that is isomorphic if that is what this is all about. it's exported from runtime common here: https://github.com/cardstack/boxel/blob/main/packages/runtime-common/index.ts#L68. it works in both the browser and in node.

PR #2407, opened 2025-04-09
AI Assistant: ability to accept all offered code patches at once (Accept All button)

jurgenwerk:
This PR adds an "Accept All" button. When the assistant responds with multiple patches, you can now accept them all at once.

image

Observe here how the assistant offered multiple code patches to be accepted and I accept them all at the same time:

https://github.com/user-attachments/assets/4983a9af-9c30-4066-907b-9f101f3ca12d

When I was working on this ticket there was no design yet and idea was to "just add a button at the bottom". But just yesterday when I was finishing up this PR we got designs for this so I'm wondering if we can ship this and apply the design in a follow up work - I'll be checking this with Chris


packages/host/app/components/ai-assistant/formatted-message.gts
```diff
@@ -109,6 +121,60 @@ export default class FormattedMessage extends Component<FormattedMessageSignatur
     return index === this.htmlGroups.length - 1;
   };
 
+  private createCodePatchAction = (codeData: CodeData) => {
+    let codePatchAction = new CodePatchAction(getOwner(this)!, codeData);
```
jurgenwerk, line 125: CodePatchAction is a an object that holds the state of the code patch function (ready, applied, applying, failed) and has the functionality to apply the patch.

The owner is passed and set in `CodePatchAction` because the owner is needed in order to inject services, such as loader service, command service, card service (requirement of the framework)

FadhlanR (approved):
I've tested it in the staging preview—it works great! One small thing: I tried a scenario where I had applied one patch and then clicked the 'Accept All' button. I expected that the already applied patch wouldn't be affected, but its checkmark icon changed to the loading state. I'm not sure what the expected behavior is in this case.

richardhjtan (approved):
Screenshot 2025-04-10 at 10 47 50 PM

Looks good!

richardhjtan:
Some comments that can be looked into separately:
1. gpt-4o does not provide an apply all button, while sonnet 3.5 does
Screenshot 2025-04-10 at 10 46 04 PM
2. Monaco editor has green border when the page start resize on 90% downwards (in the above screenshot), not sure if is intended or default

packages/host/app/components/ai-assistant/formatted-message.gts
lukemelia: We should consider keeping this somewhere besides component state so that if a user closes and reopens the AI Assistant, the state is correct. Command execution has similar needs. Command execution state is tracked in the CommandService. In fact, if we made it a command, perhaps it would just work?

OK with me to merge as is and follow up though.

jurgenwerk:
Thank you all for thorough testing, I'll address the issues raised

packages/host/app/components/ai-assistant/formatted-message.gts
jurgenwerk: I would defer this to the commands integration work (where we will persist historic state of the code patch applied commands). I'll make a note in that ticket.

PR #2408, opened 2025-04-09
Fixing run away search issues in code mode

habdelra:
So the issue that triggered all the searches in code mode was a very subtle yet super pathological issue around reactivity resulting in a geometric expansion in the number of search queries run after each indexing event. Ultimately the ModuleContentsResource was rebuilding the declaration on the code changes from the module. that resulted in a bunch of calls to rebuild the declarations within this resource. reactivity in glimmer is triggered based on strict equality. think of strictEquals() vs deepEquals() in our unit tests. you can have 2 instances of `{blah: "foo"}` and `{blah: "foo"}`. These are not strictly equal, but they are deeply equal. This was basically the issue. about 8 degrees of separation live between the ModuleContentsResource.declations and the resulting search query for the spec, all chained together using getters that are reactive. the module declaration would change each time there was an edit in code module to the module source code, but ultimately the module declarations were deeply equal. Regardless, that is not how glimmer reactivity works. So even though the search query never changed from a deep equality standpoint, it was not strictly equal after each code change, and as a result the search would re-run (against all the realms). Moreover, this was set as a live search, and a new live search callback would be registered with each "query change"--which compounded the searches resulting in a geometric expansion of all the searches being run.

Also, as part of this I noticed that we were forcing all prerendered searches to be live bound. there are many scenarios where we don't want that actually, so i made this a param for prerendered search. As well as Prerendered Search had the same vulnerability where it would re-run for queries that were not stricty equal, but were deeply equal

Also, i removed the getSearch and used the provided/consume pattern instead.

After this change, there are no search queries that are issued after each code change  made in code mode:

https://github.com/user-attachments/assets/fda69d1f-e100-4c62-935c-08fe7abd716d

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/spec-preview.gts
+++ b/packages/host/app/components/operator-mode/code-submode/spec-preview.gts
@@ -5,6 +5,7 @@ import { next } from '@ember/runloop';
 import { service } from '@ember/service';
 import { htmlSafe } from '@ember/template';
 import GlimmerComponent from '@glimmer/component';
+import { tracked, cached } from '@glimmer/tracking';
 
 import AppsIcon from '@cardstack/boxel-icons/apps';
 import Brain from '@cardstack/boxel-icons/brain';
@@ -15,7 +16,9 @@ import StackIcon from '@cardstack/boxel-icons/stack';
 import { task } from 'ember-concurrency';
 import Modifier from 'ember-modifier';
 import { consume } from 'ember-provide-consume-context';
+import { use, resource } from 'ember-resources';
 import window from 'ember-window-mock';
+import { TrackedObject } from 'tracked-built-ins';
 
 import {
   BoxelButton,
@@ -45,13 +48,12 @@ import {
 } from '@cardstack/runtime-common/code-ref';
 
 import Preview from '@cardstack/host/components/preview';
-
+import consumeContext from '@cardstack/host/helpers/consume-context';
 import {
   CardOrFieldDeclaration,
   isCardOrFieldDeclaration,
   type ModuleDeclaration,
 } from '@cardstack/host/resources/module-contents';
-import { getSearch } from '@cardstack/host/resources/search';
 
 import type CardService from '@cardstack/host/services/card-service';
 import type EnvironmentService from '@cardstack/host/services/environment-service';
@@ -456,6 +458,8 @@ const SpecPreviewLoading: TemplateOnlyComponent<SpecPreviewLoadingSignature> =
   </template>;
 
 export default class SpecPreview extends GlimmerComponent<Signature> {
+  @consume(GetCardsContextName) private declare getCards: getCards;
+
   @service private declare operatorModeStateService: OperatorModeStateService;
   @service private declare environmentService: EnvironmentService;
   @service private declare realm: RealmService;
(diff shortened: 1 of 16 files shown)


packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -466,17 +470,41 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
+      value: undefined,
+    });
+    (async () => {
+      await Promise.resolve(); // buffer 1 microtask to prevent re-render cycles
+      if (this.args.selectedDeclaration?.exportName !== exportName.value) {
+        exportName.value = this.args.selectedDeclaration?.exportName;
+      }
+    })();
+    return exportName;
+  });
```
habdelra: This is how we adjust the reactivity to work in a deeply equal sense vs strictly equal. in this case we really only consume the export name of the selected declaration that was triggering the run away search, so i didn't opt to use `lodash.isEqual()` on the entire declaration arg.

packages/host/app/components/prerendered-card-search.gts
```diff
@@ -175,6 +176,8 @@ export default class PrerenderedCardSearch extends Component<Signature> {
     let { query, format, cardUrls, realms } = this.args;
 
     let realmsChanged = !isEqual(realms, this._lastRealms);
+    // guard against a query that is deeply equal to the last query, but not strictly equal
+    let queryChanged = !isEqual(query, this._lastSearchQuery);
```
habdelra: We make a similar guard in PrerenderedSearch as well which has the same vulnerability

packages/host/app/resources/search.ts
```diff
@@ -66,6 +68,17 @@ export class SearchResource extends Resource<Args> {
...
+      if (
+        this.subscriptions &&
+        (this.#previousQuery !== query || this.#previousRealms !== realms)
+      ) {
+        for (let subscription of this.subscriptions) {
+          subscription.unsubscribe();
+        }
+      }
+      this.#previousQuery = query;
+      this.#previousRealms = realms;
```
habdelra, line 81: not unsubscribing live searches when the query changes compounded the issue and resulted in geometric expansion of the queries being issued in this scenario

packages/host/app/components/prerendered-card-search.gts
```diff
@@ -265,7 +269,9 @@ export default class PrerenderedCardSearch extends Component<Signature> {
   };
 
   <template>
-    {{SubscribeToRealms @realms this.markRealmNeedsRefreshing}}
+    {{#if @isLive}}
+      {{SubscribeToRealms @realms this.markRealmNeedsRefreshing}}
+    {{/if}}
```
habdelra, line 300: consumers of the Prerendered search can now decide if they want a live search

packages/host/app/resources/search.ts
```diff
@@ -222,7 +235,7 @@ export function getSearch(
     named: {
       query: getQuery(),
       realms: getRealms ? getRealms() : undefined,
-      isLive: opts?.isLive != null ? opts.isLive : true,
+      isLive: opts?.isLive != null ? opts.isLive : false,
```
habdelra, line 238: searches using the SearchResource no longer default to being live.

packages/host/app/components/prerendered-card-search.gts
```diff
@@ -127,6 +129,20 @@ export default class PrerenderedCardSearch extends Component<Signature> {
...
+    });
+    (async () => {
+      await Promise.resolve(); // buffer 1 microtask to prevent re-render cycles
+      if (!isEqual(this.args.query, query.value)) {
+        query.value = this.args.query;
+      }
+    })();
+    return query;
+  });
+
```
habdelra, line 145: We make a similar guard in PrerenderedSearch as well which has the same vulnerability

habdelra:
cant make a deep equal reactive mechanism, glimmer strict equality reactivity is too deep... trying again...

PR #2413, opened 2025-04-09
Playground create-instance bug

burieberry:
- `this.store.create` call was missing the realm url and defaulting to personal realm, not the currently open realm.
- was able to reproduce the error in multiple-realm tests. Moved all existing card and field playground tests to nested module named `single realm` and added the new tests in `multiple realms` module. 

Note: Discovered a different bug -- when creating new field spec in multiple-realm situation, the spec is generated but it does not show up on spec panel until refreshed. --> Update: After merging in latest main, this is no longer an issue when I try it in the app, however, can't replicate the success in tests somehow

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
@@ -244,8 +244,12 @@ export default class PlaygroundPanel extends Component<Signature> {
     }
   }
 
+  private get currentRealm() {
+    return this.operatorModeStateService.realmURL.href;
+  }
+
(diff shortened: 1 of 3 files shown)


richardhjtan (approved):
Tried creating an instance in a different workspace, and it worked as expected

richardhjtan:
There is a minor lint error (this PR handles that https://github.com/cardstack/boxel/pull/2410#pullrequestreview-2755351548), the fail test looks fine when I run it locally

jurgenwerk (approved):
Nice catch

habdelra (approved):

PR #2414, opened 2025-04-09
stop run away search in code mode

habdelra:
So the issue that triggered all the searches in code mode was a very subtle issue around reactivity. Ultimately the ModuleContentsResource was rebuilding the declaration on the code changes from the module. that resulted in a bunch of calls to rebuild the declarations within this resource. reactivity in glimmer is triggered based on strict equality. think of `strictEquals()` vs `deepEquals()` in our unit tests. you can have 2 instances of `{blah: "foo"}` and `{blah: "foo"}`. These are not strictly equal, but they are deeply equal. This was basically the issue. about 6 degrees of separation live between the ModuleContentsResource.declations and the resulting search query for the spec in `SpecPreviewComponent`, all chained together using getters that are reactive. the module declaration would change each time there was an edit in code mode to the module source code, but ultimately the module declarations were deeply equal. Regardless, that is not how glimmer reactivity works. So even though the search query never changed from a deep equality standpoint, it was not strictly equal after each code change, and as a result the search would re-run (against all the realms). 

Compounding this issue was that this is  set as a live search in `SpecPreviewComponent`, and a new live search callback would be registered with each query object change triggered in the `modify()` hook of the search resource--which compounded the searches resulting in a geometric expansion of all the searches being run.

Also, as part of this I noticed that we were forcing all prerendered searches to be live bound. there are many scenarios where we don't want that actually, so i made this a param for prerendered search. As well as Prerendered Search had the same vulnerability where it would re-run for queries that were not stricty equal, but were deeply equal.

After this change, there is still a series of search queries made after each update, but it is very much under control. Specifically we search in each realm for the spec of the selected declaration after each realm indexing event is received. This is used for the spec preview (which i think shows the spec detail in the accordion panel header and is responsive to code changes). We also removed all the prerendered searches triggering on realm index events, as well as queries that were deeply equal but not strictly equal that were occurring after each code change in code mode

### DIFF
--- a/packages/host/app/components/card-catalog/modal.gts
+++ b/packages/host/app/components/card-catalog/modal.gts
@@ -9,7 +9,6 @@ import Component from '@glimmer/component';
 
 import { restartableTask, task, timeout } from 'ember-concurrency';
 import focusTrap from 'ember-focus-trap/modifiers/focus-trap';
-import { consume } from 'ember-provide-consume-context';
 
 import flatMap from 'lodash/flatMap';
 
@@ -21,8 +20,6 @@ import { eq, not } from '@cardstack/boxel-ui/helpers';
 import {
   type CodeRef,
   type CreateNewCard,
-  type getCards,
-  type getCard,
   createNewCard,
   baseRealm,
   Deferred,
@@ -30,8 +27,6 @@ import {
   RealmInfo,
   CardCatalogQuery,
   isCardInstance,
-  GetCardContextName,
-  GetCardsContextName,
 } from '@cardstack/runtime-common';
 
 import type {
@@ -75,7 +70,6 @@ interface Signature {
 }
 
 type Request = {
-  search: ReturnType<getCards>;
   deferred: Deferred<string | undefined>;
   opts?: {
     offerToCreate?: {
@@ -235,9 +229,6 @@ export default class CardCatalogModal extends Component<Signature> {
     </style>
   </template>
 
-  @consume(GetCardContextName) private declare getCard: getCard;
-  @consume(GetCardsContextName) private declare getCards: getCards;
-
   private stateStack: State[] = new TrackedArray<State>();
   private stateId = 0;
   @service private declare cardService: CardService;
@@ -366,7 +357,6 @@ export default class CardCatalogModal extends Component<Signature> {
         opts?.multiSelect,
       );
       let request = new TrackedObject<Request>({
-        search: this.getCards(this, () => query),
         deferred: new Deferred(),
         opts,
       });
--- a/packages/host/app/components/operator-mode/code-submode/playground/spec-search.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/spec-search.gts
@@ -42,7 +42,6 @@ export default class SpecSearch extends Component<Signature> {
       this,
       () => this.args.query,
       () => this.args.realms,
-      { isLive: true },
     );
   };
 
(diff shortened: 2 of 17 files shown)


packages/host/app/components/card-catalog/modal.gts
```diff
@@ -366,7 +357,6 @@ export default class CardCatalogModal extends Component<Signature> {
         opts?.multiSelect,
       );
       let request = new TrackedObject<Request>({
-        search: this.getCards(this, () => query),
```
habdelra, line 369: this was actually never being used! (this was refactored to use prerendered searches--so this is not necessary)

packages/host/app/components/operator-mode/code-submode/playground/spec-search.gts
```diff
@@ -42,7 +42,6 @@ export default class SpecSearch extends Component<Signature> {
       this,
       () => this.args.query,
       () => this.args.realms,
-      { isLive: true },
```
habdelra, line 45: i don't think we want this to be live (which is different than being responsive).

habdelra: I was wrong, this is necessary...

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -613,36 +616,38 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
-    () => this.realms,
-    { isLive: true, isAutoSaved: true },
-  );
+  private makeSearch = () => {
+    this.search = this.getCards(
+      this,
+      () => this.specQuery,
+      () => this.realms,
+      { isLive: true },
+    ) as ReturnType<getCards<Spec>>;
```
habdelra, line 625: This search is run after a realm indexing event is received in code mode. specifically this is the only search that is running as code is being changed in the code editor. (and it is one query per realm per realm index event received)

packages/host/app/components/operator-mode/code-submode/playground/spec-search.gts
tintinthong: If a user creates a new spec via a new json file, wudn't we want the new entry to appear?

habdelra: Yeah, but I think this component is created each time the drop down is opened. If not then we really need to think how we can share the search with the spec preview, as it is the same search, which is a request to all thr realms after each indexing event. It’s super wasteful not to share. Also why do we not have a test for this if that’s the case?

habdelra: @tintinthong if you feel this is an important feature could you please add a test to this PR for it, as you are more familiar with the expected functionality?

jurgenwerk (approved):
Thanks for a really good description of the problem!

packages/host/app/components/operator-mode/code-submode/playground/spec-search.gts
tintinthong: I don't think this is an important feature bcos I don't think this is a natural action that user takes. 

I think can de-prioritise and handle it if its an unusual behaviour that we notice in the future -- but I doubt so

tintinthong (approved):
Nice

packages/host/app/components/operator-mode/code-submode/playground/spec-search.gts
habdelra: sounds like a plan. this conversation has also sparked in me the idea that we need to share live searches in the store when they are the same query. I made this ticket for that https://linear.app/cardstack/issue/CS-8385/manage-live-search-subscriptions-in-the-store-to-optimize-for-search

PR #2415, opened 2025-04-10
Remove empty space in the bottom of the code block

FadhlanR:
In this PR, I removed the empty space at the bottom of the code block by setting `scrollBeyondLastLine` to `false` in the Monaco editor config and removing the height definition from the code block CSS class. Instead of defining the height in CSS, I added code to set the code block height based on the Monaco editor's content height, while also defining a max height to prevent it from becoming too long.

Before:

https://github.com/user-attachments/assets/979b1098-ebbf-43fa-8b5c-5ab058cd53c5

After:

https://github.com/user-attachments/assets/1da62ee5-7783-4274-9428-ba20d3cde8ad


jurgenwerk (approved):
I tested this locally and observed the code diff editor is now taller

image

I personally think this is a little bit too tall but otherwise I'm fine with it. We can see what design team thinks.

I do have a comment about very short snippets–in the bottom they appear as cut off.  Here's an example:
image

Adding some padding would give it more room to breathe and be more aesthetically pleasing IMO

image

PR #2418, opened 2025-04-10
Fix cross-origin worker loading in monaco (patch until new version gets released)

jurgenwerk:
Yesterday when we were demoing code diffs in our sprint planning we noticed the code diffs in the AI panel weren't working. 

I was unable to reproduce locally, but then I saw this in staging and prod (we also noticed that yesterday):
image

Then I found that people are already tracking this issue in https://github.com/microsoft/monaco-editor/issues/4741, the fix has been merged in but the version is not released yet, so I opted into patching our monaco to include this fix: https://github.com/microsoft/monaco-editor/pull/4742

I deployed this fix to staging and saw the mentioned error is gone, and the diffs are working again (previously just code was shown, no diffs):

image

### DIFF
--- a/packages/host/package.json
+++ b/packages/host/package.json
@@ -155,7 +155,7 @@
     "moment": "^2.29.4",
     "moment-locales-webpack-plugin": "^1.2.0",
     "monaco-editor": "^0.52.2",
-    "monaco-editor-webpack-plugin": "^7.0.1",
+    "monaco-editor-webpack-plugin": "^7.1.0",
     "ms": "^2.1.3",
     "path-browserify": "^1.0.1",
     "pluralize": "^8.0.0",
--- a/patches/monaco-editor-webpack-plugin@7.1.0.patch
+++ b/patches/monaco-editor-webpack-plugin@7.1.0.patch
@@ -1,0 +1,455 @@
+diff --git a/out/index.js b/out/index.js
+index 2e61d97b15096b2b48492ef35aad6188e1f416b1..880dfa27773d8c58f50c4b529a4eda31d7917dab 100644
+--- a/out/index.js
++++ b/out/index.js
+@@ -3,162 +3,211 @@ const path = require("path");
+ const loaderUtils = require("loader-utils");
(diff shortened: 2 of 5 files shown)


packages/host/package.json
```diff
@@ -155,7 +155,7 @@
     "moment": "^2.29.4",
     "moment-locales-webpack-plugin": "^1.2.0",
     "monaco-editor": "^0.52.2",
-    "monaco-editor-webpack-plugin": "^7.0.1",
+    "monaco-editor-webpack-plugin": "^7.1.0",
```
jurgenwerk, line 158: When 7.1.1 will be released, we will be able to upgrade to that version, and remove the patch.

patches/monaco-editor-webpack-plugin@7.1.0.patch
```diff
@@ -0,0 +1,455 @@
...
++  const globals = {
++    MonacoEnvironment: `(function (paths) {
+       function stripTrailingSlash(str) {
+         return str.replace(/\\/$/, '');
+       }
+@@ -174,7 +223,14 @@ function createLoaderRules(languages, features, workers, filename, monacoEditorP
+               if(/^(\\/\\/)/.test(result)) {
+                 result = window.location.protocol + result
+               }
+-              var js = '/*' + label + '*/importScripts("' + result + '");';
```
jurgenwerk, line 348: This here is the actual patch (https://github.com/microsoft/monaco-editor/pull/4742/files#diff-4c9d98877832d9f6dc71a783388abbe7fc64971386d338e4d3f9e7444137d713), the rest of the changes are from auto formatting after saving it in my editor

lukemelia (approved):

richardhjtan (approved):

patches/monaco-editor-webpack-plugin@7.1.0.patch
backspace: If you use VS Code (or Cursor) you can bypass autoformat-on-save with Command-K, S.

PR #2420, opened 2025-04-10
host: Add logger service

backspace:
With this, you can use Ember Inspector to get a handle on loggers and dynamically change their levels. This will facilitate debugging without requiring a deployment.

This also wires in the Matrix SDK’s logger (also `loglevel`) to the `LOG_LEVELS` environment variable and sets it to `info` in development, as it’s currently very noisy with information that is (almost?) never of use to us:

Boxel 2025-04-10 12-38-49

I couldn’t get it to work in deployed environments but at least it’s quieter locally:

index json in feb11 2025-04-10 15-21-58

In staging and production the logger service can still be used to change the level.


packages/host/config/environment.js
```diff
@@ -74,7 +74,7 @@ module.exports = function (environment) {
 
   if (environment === 'production') {
     // here you can enable a production-specific feature
-    ENV.logLevels = '*=warn,current-run=error';
+    ENV.logLevels = '*=warn,current-run=error,matrix=info';
```
backspace: This assignment means `LOG_LEVELS` has no effect in deployed environments, is that what we expect?

backspace: I’ve since removed this as it was having no effect. I still think it would be good to reduce the Matrix logging level on startup, but for now we’ll have to do it with the logger service.

habdelra (approved):
:+1:

PR #2422, opened 2025-04-10
Improve Boxel Coding skill

lukemelia:


lukemelia:
Passed here: https://github.com/cardstack/boxel/actions/runs/14404566548

backspace (approved):
I see it was already in there but should we be using Highcharts as an example considering its use of `document`?

lukemelia:
> I see it was already in there but should we be using Highcharts as an example considering its use of document?

Highcharts is popular and it would be good to show how to use it successfully, if possible

PR #2423, opened 2025-04-10
Markdown rendering should properly escape HTML in code blocks

lukemelia:


FadhlanR:
I think I fix the same issue in this PR https://github.com/cardstack/boxel/pull/2387/files#r2036548631.

lukemelia:
@FadhlanR does your change fix this problem in MarkdownField rendering? It looked like it was specific to codeblocks, and more complex than what I needed for rendering markdown.

FadhlanR:
@lukemelia Ah, I got it. Maybe you could use this function: https://github.com/cardstack/boxel/blob/4379525e32a877d4c5605b2f01a5361ccd646a98/packages/runtime-common/helpers/html.ts#L20.
I'll rebase my PR to main after this one is merged. Previously, we escaped HTML outside the markdownToHtml function, as you can see https://github.com/cardstack/boxel/blob/4379525e32a877d4c5605b2f01a5361ccd646a98/packages/host/app/services/matrix-service.ts#L757.

lukemelia:
@FadhlanR I made some changes to provide more flexibility and avoid changing behavior in ai assistant messages. Feel free to merge so that you can proceed with your PR.

jurgenwerk (approved):

FadhlanR (approved):

PR #2425, opened 2025-04-11
CS-8392 Feature flag on catalog listing app

richardhjtan:
### What’s Changing
- Enable a feature flag for the catalog listing app, limited to development and staging environments.
- Introduce a temporary catalog-realm-dev realm for development and testing.

This implementation differs slightly from typical feature flags because:
- The catalog realm’s `index.json` controls app card display, so logic can’t be applied in the JSON file.
- Listings should remain in dev/staging until it is production-ready.

**Follow-Up**
Revert these changes once the catalog listing app goes live in production.


packages/catalog-realm-dev/TODO.md
```diff
@@ -0,0 +1 @@
+Make sure to lint for hbs when we add gts files to this realm like we do in the experiments realm. Currently linting for templates fails because there are no gts files in this realm.
```
backspace: Because of [`--no-error-on-unmatched-pattern`](https://github.com/cardstack/boxel/pull/2425/files#diff-f3dac9dd7bab04c2b1ded3d6fefd05a1435a845aca4f7c5354b964ad793a9fe8R20) this can be removed. (We should remove it elsewhere too, assuming this is a copypaste thing.)

backspace (approved):

tintinthong (approved):
Nice

packages/catalog-realm-dev/TODO.md
richardhjtan: Thanks for noticing @backspace, I have cleaned up those files in this PR as well

PR #2427, opened 2025-04-11
Patch cards in code mode

IanCal:
This make sure with ask-ai, and in the rooms, that the current open card (including in playground) is auto attached and marked as open (so that a patch command is setup).


jurgenwerk (approved):
I tested this and it works OK. I was worried these two won't play well together but looks like it works nicely.

image

I really like it that this PR greatly decreases the chances of the assistant acting like it can't see the attached file in code mode (which happens sporadically when prompting). This is because after your change `tools` now has items and previously it didn't: https://github.com/cardstack/boxel/blob/f737dd9980fbe43029a8d81bdf17c550f9ded9b4/packages/ai-bot/helpers.ts#L751-L755

But what if there are no attached cards (user removes the autoattached card), and only keeps the autoattached file? In this case `tools` will not have any items, and the above message will be added to the system prompt, resulting in bigger chance of the model responding with sth like "I notice that no cards are currently open and accessible for me to edit.", even though there is an attached file present. Do you think we should change the condition to `if (attachedCards.length > 0 && tools.length == 0) {`? Edit: made a PR for that: https://github.com/cardstack/boxel/pull/2434

habdelra (approved):

packages/host/app/components/matrix/room.gts
```diff
@@ -810,9 +815,17 @@ export default class Room extends Component<Signature> {
 
   private get autoAttachedCardIds() {
     if (this.operatorModeStateService.state.submode === Submodes.Code) {
-      return this.playgroundPanelCardId
-        ? new TrackedSet([this.playgroundPanelCardId])
-        : new TrackedSet<string>();
+      // also get the card ids of the cards that are open in code mode
+      let cardIds = new TrackedSet<string>();
+      if (this.autoAttachedFileUrl?.endsWith('.json')) {
+        // remove the json extension. TODO: is there a way of getting the actual card id
```
lukemelia, line 821: The URL minus the .json _is_ the actual card ID for all cards that we have so far.

lukemelia (approved):

jurgenwerk:
Another question: since card can be autoattached in code mode should we also include the Card Editing skill?

IanCal:
> Another question: since card can be autoattached in code mode should we also include the Card Editing skill?

Good question - if not we probably want some skill that explains this stuff, or check if the patch tool itself has enough information. I'll file a ticket, there's work to be done sorting out the skills and this should either be part of it or done as well.

PR #2428, opened 2025-04-11
Refactor CardResource to remove synchronization work

habdelra:
This PR addresses @ef4 's concerns summarized in https://linear.app/cardstack/issue/CS-8313/card-resource-still-does-too-much-synchronization-work:

> host/app/resources/card-resource.ts still contains a `load` task and its own `@tracked _card`. This is bad and I suspect it's related to failure-to-update bugs that Burcu and I have been trying to track down.
>
> All tracked state should live in the store.
>
> The card resource itself should only be responsible for references and de-referencing the particular URL it's trying to access, and proxying all data reads through to the store. Example:
> 
>
> ```ts
> export class CardResource extends Resource<Args> {
>   #url: string | undefined;
>   declare private store: StoreService;
> 
>   modify(_positional: never[], named: Args['named']) {
>     let { url } = named;
>     if (url !== this.#url) {
>       if (this.#url) {
>         this.store.dropReference(this.#url);
>       }
>       this.#url = url;
>       this.store.addReference(this.#url);
>     }
>     registerDestructor(this, () => {
>       if (this.#url) {
>         this.store.dropReference(this.#url);
>       }
>     });
>   }
> 
>   get card() {
>     return this.store.peekCard(this.#url);
>   }
> 
>   get url() {
>     return this.#url;
>   }
> 
>   get isLoaded() {
>     // or maybe Boolean(this.store.peekCard(this.#url)), if that means the same thing
>     return this.store.isLoaded(this.#url);
>   }
> }
> ```
>
> Notice the store never gets a handle on the resource, and the resource has no tracked state.
>
> I don't really think `isLive` can make sense as an argument to the card resource, because it's supposed to return the same instance that every other card resource is returning. They will all either be live or not, that is a store-level policy decision, not something one card resource can decide.


packages/host/tests/integration/components/ai-assistant-panel/codeblocks-test.gts
```diff
@@ -98,7 +98,7 @@ module('Integration | ai-assistant-panel | codeblocks', function (hooks) {
       @field firstName = contains(StringField);
       @field firstLetterOfTheName = contains(StringField, {
         computeVia: function (this: Person) {
-          return this.firstName[0];
+          return this.firstName?.[0];
```
habdelra: I'm not sure why we are starting to see this now--but i got an error trying to render this card due to this error. computed's should be written such that we support any referenced properties potentially being undefined as that is a totally valid scenario for any field. It would be nice if TS could enforce this, and we spent lots of time trying to get the types to work out for that scenario a long while back but it was a crazy difficult slog and we never got there.

habdelra: ah i see why this happened. I had an async race condition bug where my `Store.isLoaded` property was not being set at the right time and we started rendering the card before the data had completed being loaded into the instance. but still--this is actually a more correct way to author the card.

packages/host/app/resources/card-resource.ts
```diff
@@ -3,157 +3,69 @@
...
     return this.#id;
   }
 
-  get cardError() {
-    return this._error;
-  }
-
   get isLoaded() {
-    return this._isLoaded;
+    return !!this.#id && this.store.isLoaded(this.#id);
```
habdelra: i think this is the same as `Boolean(this.store.peek(this.#id))`

packages/host/tests/acceptance/interact-submode-test.gts
```diff
@@ -1599,8 +1599,6 @@ module('Acceptance | interact submode tests', function (hooks) {
             // second save is after a field has been filled in
             assert.strictEqual(json.data.attributes?.name, 'Paper');
             deferred.fulfill();
-          } else {
-            assert.ok(false, 'unexpected save of Pet card');
```
backspace, line 1603: Can this not happen anymore?

packages/host/tests/acceptance/interact-submode-test.gts
```diff
@@ -828,7 +828,7 @@ module('Acceptance | interact submode tests', function (hooks) {
     });
 
     test<TestContextWithSave>('can create a card from the index stack item', async function (assert) {
-      assert.expect(9);
+      assert.expect(7);
```
backspace, line 831: Why does this change? 😯

backspace (approved):
I only have a superficial understanding of this but from that level it looks good

packages/host/tests/acceptance/interact-submode-test.gts
habdelra: yeah, this was awkward. now that we are making switching out of edit mode no longer await the save, it's not necessarily guaranteed how many number of saves that you'll have by the time the edit mode is dismissed. (which is what this test does). sometime this will flag an extra save by the time edit mode is dismissed. other times it will not. so I removed the assertion that a specific number of saves needs to happen by the time the edit mode is dismissed. (the test waiters changed enough in the store that this nuance was revealed)

packages/host/tests/acceptance/interact-submode-test.gts
habdelra: There was an extraneous save that was happen when the stack item was closed even though there were no card changes.

habdelra:
> I only have a superficial understanding of this but from that level it looks good

yeah, the north star in this PR is that the CardResource has no state other than the id of the card, that the store has no reference to any card resource instances, and that the identity map becomes tracked. everything else flowed from those ideas. and nicely there was an overall net reduction in lines of code since those ideas simplified our store

lukemelia (approved):

richardhjtan (approved):
Nice one!

richardhjtan:
This PR also fix the reactivity issue in Spec README editing, and also the sprint planner (not working lately)

jurgenwerk (approved):
I am testing the AI assistant and in current main, the card in playground does not refresh after accepting a code change. 

I tried it using the branch from this PR and the issue is solved!

PR #2430, opened 2025-04-14
fix linked cards occasionally missing. Use value.data.id

tintinthong:
This PR is to fix the issue where linked data is missing 

Screenshot 2025-04-09 at 23 40 10


packages/base/card-api.gts
```diff
@@ -966,7 +966,11 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
       cachedInstance[isSavedInstance] = true;
       return cachedInstance as BaseInstanceType<CardT>;
     }
-    let resourceId = new URL(value.links.self, relativeTo).href;
+    let relativeId =
+      value.data && Array.isArray(value.data)
```
habdelra: this is a LinksTo relationship, not a LinksToMany, why would the value here ever be an array? (keep in mind `value` here is the JSONAPI serialization for the relationship).

packages/base/card-api.gts
```diff
@@ -966,7 +966,11 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
...
+        ? undefined // we don't know the id of the parent
+        : value.data?.id ?? relativeTo;
```
habdelra: generally we strive to serialize with the `links.self` when sending the serialized form of the instance to the server. both `links.self` and `links.data.id` are supported, using both though, can lead to the potential of conflicts if they were ever configured differently from each other. which is why we have opted to pick one (links.self is more concise). i'm super curious what the scenario is when you are deserializing a payload and links.data.id is populated and links.self is not. I think that is the _actual_ bug if this is what is fixing it.

habdelra: another thought is to remove links.self altogether and purely only use links.data.id (this this also means we should not serialize this so we don't introduce confusion).

habdelra: and another though is to favor links.self, but fallback to links.data.id if links self is not populated and links.data.id is populated. the drawback is that that gets into the confusing place of having these at odds with one another....

tintinthong: Thanks for the input. The issue is really that `value.links.self` is relative. In addition, `doc.data.id` seems like the wrong id to be relative to anyway. 

The data values are 

```
value.links.self
../Category/someCategory

doc.data.id
http://test-realm/test/listing

value.data.id
http://test-realm/test/Category/someCategory

resourceId
http://test-realm/Category/someCategory
```

The expected value 

```
resourceId
http://test-realm/test/Category/someCategory
```

I have added the `serialization-test` test "can resolve relative URLs in linksToMany fields irrespective of location of linked cards"



> another thought is to remove links.self altogether and purely only use links.data.id (this this also means we should not serialize this so we don't introduce confusion).

I think value.data.id seems to return an absolute url which is actually what I want. Perhaps, this is what we want to do.

tintinthong: Screenshot 2025-04-15 at 14 35 59

The JSON API docs consider data to be kinda optional actually. Tho it seems links.self are also treated as optional. I can see why implementation wise it can get unnecessarily complex

I think in deserialization we can prefer value.data.id but I think in general, if we fallback to links.self the deserialization may be wrong

removal of serialization however, lets talk about it perhaps after standup

packages/base/card-api.gts
```diff
@@ -1322,8 +1322,8 @@ class LinksToMany<FieldT extends CardDefConstructor>
         }
         let resourceId = new URL(
           value.links.self,
-          !Array.isArray(doc.data) && 'id' in doc.data && doc.data.id
-            ? doc.data.id
+          !Array.isArray(value.data) && value.data?.id
+            ? value?.data?.id
             : relativeTo,
```
tintinthong, line 1327: Actually, the more I think about this. The more I feel that this is not an issue about the usage of value.data.id rather its just a simple bug wrt to the derivation of the absolute URL 

```
# Wrong but intuitive 
new URL('../Category/someCategory', 'http://test-realm/test/listing').href
'http://test-realm/Category/someCategory'

#Correct but unintuitive 
new URL('../Category/someCategory', 'http://test-realm/test/listing/').href
'http://test-realm/test/Category/someCategory'
```

packages/base/card-api.gts
habdelra: Yeah, it’s especially tricky too, because I don’t think you can look at your contained card (or contained contained card, etc as the case may be), to see the id of the outermost document that you are relative too. That would be perfect if you could.

habdelra: Perhaps we need to thread thru a new parameter thru the deserialzation: the id of the document (if there is one), aka “relativeTo” so that we can resolve any relative hrefs. If we did that, there’s about 4-5 places that the relativeTo value would need to be added to function parameters.

habdelra: If you like maybe we can pair on this after standup.

tintinthong: Yes pls that wud be helpful

habdelra: I was thinking one argument for not using links.data.id is that it makes it more challenging to copy a set of cards to a different realm. As relative references can be moved more easily.

packages/host/tests/integration/components/serialization-test.gts
```diff
@@ -5004,6 +5004,94 @@ module('Integration | serialization', function (hooks) {
...
+      );
+      let listingDoc: LooseSingleCardDocument = {
+        data: {
+          type: 'card',
+          id: `${testRealmURL}listing`,
+          attributes: {},
+          relationships: {
+            'category.0': {
+              links: {
+                self: '../Category/someCategory',
```
habdelra: the fact that this is intentionally a bad ref (`..` vs `.`) is good. it proves we are not paying attention to the links.self when loading the data from the included bucket.

packages/base/card-api.gts
```diff
@@ -1322,8 +1322,8 @@ class LinksToMany<FieldT extends CardDefConstructor>
         }
         let resourceId = new URL(
           value.links.self,
-          !Array.isArray(doc.data) && 'id' in doc.data && doc.data.id
-            ? doc.data.id
+          !Array.isArray(value.data) && value.data?.id
+            ? value?.data?.id
```
habdelra: this should just be value?.data?.id. do _not_ consider `links.self` at all.

tintinthong: here we shudnt be relative to anything at all

tintinthong: Yes. This is the key ingredient

habdelra (commented):
Based on our conversation in discord, the current code shows a misunderstanding of the purpose of `links.self` vs `data.id` in the JSON-API relationship entity.

`links.self` is used to tell the consumer of this payload how to get the resource via HTTP. `data.id` is used to tell the consumer of this payload how to get the resource from the side loaded `included` bucket. we need to strictly only consider `data.id` when calling the `resourceFrom()` function (which actually loads the resource out of the included bucket). we should never used `links.self` as part of that consideration. If there is a missing `data.id` in the resource entity that means that the serialization is incorrect and is not JSON-API compliant.

Let's make sure to add this note above as a comment as well.

habdelra:
So for the LinksTo deserialize, i think the solution would be:
```ts
   let cachedInstance = identityContext.get(
      new URL(value.links.self, relativeTo).href,
    );
    if (cachedInstance) {
      return cachedInstance as BaseInstanceType<CardT>;
    }
   /*
links.self is used to tell the consumer of this payload how to get the resource via HTTP. data.id is used to tell the 
consumer of this payload how to get the resource from the side loaded included bucket. we need to strictly only 
consider data.id when calling the resourceFrom() function (which actually loads the resource out of the included 
bucket). we should never used links.self as part of that consideration. If there is a missing data.id in the resource entity 
that means that the serialization is incorrect and is not JSON-API compliant.
    */
    // only this line changes--we don't even need resourceId
    let resource = resourceFrom(doc, value?.data?.id); 
    if (!resource) {
      if (loadedValue !== undefined) {
        return loadedValue;
      }
      return {
        type: 'not-loaded',
        reference: value.links.self,
      };
    }
```

And for LinksToMany the solution would be:
```ts
        let cachedInstance = identityContext.get(
          new URL(value.links.self, relativeTo).href,
        );
        if (cachedInstance) {
          return cachedInstance;
        }
   /*
links.self is used to tell the consumer of this payload how to get the resource via HTTP. data.id is used to tell the 
consumer of this payload how to get the resource from the side loaded included bucket. we need to strictly only 
consider data.id when calling the resourceFrom() function (which actually loads the resource out of the included 
bucket). we should never used links.self as part of that consideration. If there is a missing data.id in the resource entity 
that means that the serialization is incorrect and is not JSON-API compliant.
    */
        let resourceId = value?.data?.id;
        if (resourceId) {
          if (loadedValues && Array.isArray(loadedValues)) {
            let loadedValue = loadedValues.find(
              (v) => isCardOrField(v) && 'id' in v && v.id === resourceId,
            );
            if (loadedValue) {
              return loadedValue;
            }
          }
        }
        let resource = resourceId ? resourceFrom(doc, resourceId) : undefined;
        if (!resource) {
          return {
            type: 'not-loaded',
            reference: value.links.self,
          };
        }
```

tintinthong:
Actually https://github.com/cardstack/boxel/pull/2430/commits/81d88821834a33723f35a16101341623f157f229 test are already covered in serialization test

habdelra (approved):
looks good!

PR #2435, opened 2025-04-15
User customisable default skills

IanCal:
We want users to be able to override the default skills, and in an upcoming bit of work the models that appear in the dropdown.

For this, we're adding an AI settings card.

Right now this will only have the ability to set the default skills, but then we can add models into it.

Note - this is a draft of current progress but there are key parts about loading cards that feedback would be useful on.


packages/base/command.gts
```diff
@@ -134,8 +134,16 @@ export class SendAiAssistantMessageResult extends CardDef {
   @field eventId = contains(StringField);
 }
 
-export class GetBoxelUIStateResult extends CardDef {
+export class DefaultSkills extends CardDef {
```
IanCal, line 137: these can now be removed

packages/host/app/commands/index.ts
```diff
@@ -71,6 +73,14 @@ export function shimHostCommands(virtualNetwork: VirtualNetwork) {
     '@cardstack/boxel-host/commands/set-active-llm',
     SetActiveLlmModule,
   );
+  virtualNetwork.shimModule(
```
IanCal, line 76: these need updating to the ai settings

packages/host/app/services/matrix-service.ts
```diff
@@ -875,31 +888,65 @@ export default class MatrixService extends Service {
...
+      (await this.client.getAccountDataFromServer<{
+        id: string;
+      }>(APP_BOXEL_AI_SETTINGS_EVENT_TYPE)) || { id: undefined };
+    if (!useraiSettingsCardId) {
+      // load the default
+      console.log('no user ai settings, loading default');
+      return undefined;
+    }
+    let aiSettings =
+      await this.store.peek<AiSettingsCard>(useraiSettingsCardId);
```
IanCal, line 919: Ideally I'd use the store to keep a reference that gets automatically updated instead of using peek

PR #2436, opened 2025-04-15
ci: Add timeouts when waiting for Ember CLI

backspace:
This prevents a PR with a build error from holding up the entire Actions queue like this:

boxel@0b9ebe9 2025-04-15 09-08-19

**But** would it be better to add timeouts to the entire job instead of on this command only? 🤔


lukemelia (approved):

backspace:
For now I’ll just keep it as a timeout on this step but if this problem happens again, moving it up to the job level may make more sense.

PR #2439, opened 2025-04-15
Fix bug in reference counting in store and add new card collection resource

habdelra:
This PR fixes a bug in the StoreService were we were not incrementing the reference count when a reference was added. The resulted in all cards getting garbage collected regardless of whether they are being referenced in a component via a resource. this is a super nasty bug that prevents you from viewing a card after 4 minutes.

Also this fixes an issue where in scenarios where we create an array of card resource by mapping over a list of ids, the reference count is never decremented. To handle the scenario where we want to have a resource for an ad hoc list of card id's, I've added a new `CardCollectionResource` that can handle the Store service reference counting correctly in this scenario. All the places where we have the pattern mapping over an array to make a list of card resources have been retrofitted to use the new `CardCollectionResource`.

TODO:
- [x] Add unit test for Store.addReference()
- [x] Add unit test for Store.dropReference()

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -378,6 +378,7 @@ export default class StoreService extends Service implements StoreInterface {
     await this.withTestWaiters(async () => {
       this.newReferencePromises.push(deferred.promise);
       let maybeUrl: string | undefined;
+      let isDoc = typeof idOrDoc !== 'string';
       try {
         await this.ready;
         maybeUrl = asURL(idOrDoc);
@@ -391,8 +392,11 @@ export default class StoreService extends Service implements StoreInterface {
         let url = maybeUrl;
         let urlOrDoc = idOrDoc;
         let instanceOrError = this.peek(url);
-        if (!instanceOrError) {
-          instanceOrError = await this.getInstance({ urlOrDoc });
+        if (!instanceOrError || isDoc) {
+          instanceOrError = await this.getInstance({
+            urlOrDoc,
+            opts: { noCache: isDoc },
+          });
         }
         this.subscribeToRealm(new URL(url));
         await this.trackSavedInstance('start-tracking', instanceOrError);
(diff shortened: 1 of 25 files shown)


packages/host/app/services/store.ts
```diff
@@ -174,7 +184,11 @@ export default class StoreService extends Service implements StoreInterface {
     // synchronously update the reference count so we don't run into race
     // conditions requiring a mutex
     let currentReferenceCount = this.referenceCount.get(id) ?? 0;
-    this.referenceCount.set(id, currentReferenceCount++);
```
habdelra, line 177: Doh!!

backspace:
I opened these at the same time, success 🎉

0b9c06fd-3833-4947-a0b8-ac24b8e71ee7 json in mar10 2025-04-15 16-00-35

packages/host/app/services/store.ts
```diff
@@ -121,16 +121,26 @@ export default class StoreService extends Service implements StoreInterface {
...
       this.referenceCount.delete(id);
       let autoSaveState = this.autoSaveStates.get(id);
       if (autoSaveState?.hasUnsavedChanges) {
         this.initiateAutoSaveTask.perform(id, { isImmediate: true });
       }
-      this.autoSaveStates.delete(id);
+      // await for a microtask to prevent rerender dirty tag error so we don't
+      // get in trouble because we read this.autosaveStates in the same frame as
+      // we mutate this.autosaveStates
+      (async () => {
```
jurgenwerk, line 139: Is this solving the type of that ember error that goes something like `You attempted to update ... on
... but it had already been used previously in the same computation.`? Just curious

jurgenwerk (approved):

packages/host/app/services/store.ts
habdelra: yup. you cant read and update a tracked property in the same frame, as you may get different results based on whether you read first or set first. so this makes it so they are separated by the smallest increment of time possible so that reading and setting are deterministic

jurgenwerk: Got it. I was solving this specific issue with `scheduleOnce('afterRender')` in some other places. But unsure which solution is better...

habdelra: that is a totally reasonable approach as well for a component. if you are not in a component, I'm not sure how well that works.

PR #2441, opened 2025-04-16
AI Assistant: Be more specific that code patches should target only the currently attached file

jurgenwerk:
There was a situation that happened to Chris where a diff was applied to the wrong file (more info [here])(https://linear.app/cardstack/issue/CS-8434/code-diff-can-get-applied-to-the-wrong-file).

I was trying to reproduce this ad nauseam (switching between attached files in the same session), and using vague prompts that could target any of the attached files I was switching back and forth and usually the assistant will explicitly ask if you want to make a change to a file that is currently not attached. However, **I think** there was also a case where the assistant offered a code patch for another file that was not currently attached. It feels to me this case is very rare but possible. I added  some additonal instructions in the source editing skill for that case.


packages/base/SkillCard/source-code-editing.json
```diff
@@ -2,7 +2,7 @@
...
...ude enough lines in each SEARCH section to uniquely match each set of lines that need to change.\n\nKeep *SEARCH/REPLACE* blocks concise.\nBreak large *SEARCH/REPLACE* blocks into a series of smaller blocks that each change a small portion of the file.\nInclude just the changing lines, and a few surrounding lines if needed for uniqueness.\nDo not include long runs of unchanging lines in *SEARCH/REPLACE* blocks.\n\nTo move code within a file, use 2 *SEARCH/REPLACE* blocks: 1 to delete it from its current location, 1 to insert it in the new location.\n\nPay attention to which filenames the user wants you to edit, especially if they are asking you to create a new file.\n\nIf you want to put code in a new file, use a *SEARCH/REPLACE block* with:\n- A new file path, including dir name if needed\n- An empty `SEARCH` section\n- The new file's contents in the `REPLACE` section. \n\nAvoid detailed description of the search/replace blocks. For every search/replace block write 1 sentence description max. \n\n If you propose a search/replace block, it must be for the currently attached file(s), and not for those attached before the most recent one (unless you ask and get the user's approval)",
```
jurgenwerk, line 5: Better diff: image

jurgenwerk: Added text: "If you propose a search/replace block, it must be for the currently attached file(s), and not for those attached before the most recent one (unless you ask and get the user's approval)"

PR #2446, opened 2025-04-17
Allow card catalog modal to select cards from publicly readable realms

IanCal:
This wasn't a permissions issue, it was a missing case in the card catalog modal.


backspace (commented):
Is it possible to have a test that exercises this?

habdelra (commented):
I too would also like to see tests for this

backspace (approved):
🎉

PR #2448, opened 2025-04-17
Do not crash the process when matrix event fails to be sent (e.g. event too large)

jurgenwerk:
Previously, the process died when event was too large to send (matrix 64KB limit). We are talking about the events that the bot sends to the room when it is receiving chunks from the AI model. If a chunk gets too large in a way that violates the event limit, then the process exited:

image

The change allows us to keep the bot alive when that happens and let the user know:
image

We have further work planned that will normalize event sizes but for this fix only addresses the process crash.

### DIFF
--- a/packages/ai-bot/lib/responder.ts
+++ b/packages/ai-bot/lib/responder.ts
@@ -59,7 +59,9 @@ export class Responder {
     readonly roomId: string,
   ) {}
 
-  messagePromises: Promise<ISendEventResponse | void>[] = [];
+  messagePromises: Promise<
+    ISendEventResponse | void | { errorMessage: string }
+  >[] = [];
 
   initialMessageSent = false;
   responseEventId: string | undefined;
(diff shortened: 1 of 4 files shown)


jurgenwerk:
This is missing tests but can already be reviewed.

packages/ai-bot/lib/responder.ts
```diff
@@ -222,7 +228,14 @@ export class Responder {
...
-    await Promise.all(this.messagePromises);
+
+    let results = await Promise.all(this.messagePromises);
+
+    for (let result of results) {
+      if (result && 'errorMessage' in result) {
+        await this.onError(result.errorMessage);
+      }
+    }
   }
```
lukemelia, line 239: I think this is fine. It could also be modeled with `Promise.allSettled` which means that a message promise could reject if it fails, which may be a little more expressive.

lukemelia (approved):

packages/ai-bot/lib/responder.ts
jurgenwerk: I think even with using `allSettled`, we'd still need individual `.catch` handler on each promise, otherwise the unhandled rejection will still crash the process (I tested this). So even with `allSettled` it would probably look very similar. I think this would be a clearer option if we were interested in all results values but here we're just checking if there was an error.

PR #2449, opened 2025-04-17
Enable glint/ts linting in ai-bot package

jurgenwerk:
Adds glint linting to AI Bot package. 

This revealed many type check issues. Since this work wasn't planned I time boxed it but unfortunately some tests are using test fixtures whose types  differ from what they are supposed to be in a way that is not that trivial to solve. Since it would be great to have this feature (type checking inside the editor at least) asap, I opened a new ticket to fix those later and marked them as ignored lines for now.

### DIFF
--- a/packages/ai-bot/helpers.ts
+++ b/packages/ai-bot/helpers.ts
@@ -379,7 +379,7 @@ export function getRelevantCards(
   // Return the cards in a consistent manner
   let sortedCards = Array.from(attachedCardMap.values())
     .filter((card) => card.id) // Only include cards with valid IDs
-    .sort((a, b) => a.id.localeCompare(b.id));
+    .sort((a, b) => String(a.id!).localeCompare(String(b.id!)));
 
   return {
     mostRecentlyAttachedCard: mostRecentlyAttachedCard,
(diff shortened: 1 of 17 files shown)


habdelra (approved):

richardhjtan (approved):

packages/ai-bot/helpers.ts
```diff
@@ -827,3 +844,15 @@ export function isCommandResultEvent(
...
+  array: T[],
+  predicate: (item: T) => boolean,
+): T | undefined {
+  for (let i = array.length - 1; i >= 0; i--) {
+    if (predicate(array[i])) {
+      return array[i];
+    }
+  }
+  return undefined;
+}
```
lukemelia: Why can't we use Array#findLast?

lukemelia (approved):

packages/ai-bot/helpers.ts
jurgenwerk: Actually we can but a typescript upgrade is needed, and target change from es2022 to es 2023. I adjusted this in https://github.com/cardstack/boxel/pull/2449/commits/bb77ec5016c3a2f8cc148f1ef7b490bf797b1c4e so we can now use Array#findLast

PR #2454, opened 2025-04-18
Make garbage collection tolerant of NotLoadedErrors

habdelra:


packages/host/tests/unit/garbage-collection-test.ts
```diff
@@ -658,4 +660,47 @@ module('Unit | identity-context garbage collection', function (hooks) {
...
+          },
+        },
+        meta: {
+          adoptsFrom: {
+            module: `${testRealmURL}test-cards`,
+            name: 'Person',
+          },
+        },
+      },
+    };
```
habdelra, line 691: @tintinthong FYI this is what a serialization that has a NotLoadedError looks like

tintinthong (approved):

tintinthong:
I confirm in our branch  (`skeleton-catalog`) we dont see the garbage collection error from `makeConsumptionGraph`. But the behaviour  still persist for other reasons beyond this PR

jurgenwerk (approved):

habdelra:
> I confirm in our branch (`skeleton-catalog`) we dont see the garbage collection error from `makeConsumptionGraph`. But the behaviour still persist for other reasons beyond this PR

i think we can actually use the old `getDeps()` that blows up when it sees a NotLoadedError to help us to identity the fetch that has the NotLoadedError in it (if that's where its coming from). like just call getDeps on the newly deserialized instance just after the fetch is issued and see if it blows up. and if it does track it back to the server.

PR #2455, opened 2025-04-18
Nested paths fix

tintinthong:


packages/base/card-api.gts
```diff
@@ -1319,12 +1319,7 @@ class LinksToMany<FieldT extends CardDefConstructor>
           cachedInstance[isSavedInstance] = true;
           return cachedInstance;
         }
-        let resourceId = new URL(
-          value.links.self,
-          !Array.isArray(doc.data) && 'id' in doc.data && doc.data.id
-            ? doc.data.id
-            : relativeTo,
-        ).href;
+        let resourceId = new URL(value.links.self, relativeTo).href;
```
habdelra, line 1322: why do you make this change here and not LinksTo as well? if it's broken here likely it's broken in LinksTo as well. 

Please also include a test that shows why this change needs to be made

PR #2456, opened 2025-04-18
Fix cache control header setting in server response

habdelra:
This fixes the cache control header setting in the server response. Probably you'll need to clear your browser cache in staging and prod after this fix has been applied.


packages/realm-server/server.ts
```diff
@@ -161,7 +161,7 @@ export class RealmServer {
           Object.values(SupportedMimeType)
             // Actually, we want to use HTTP caching for executable modules which
             // are requested with the "*/*" accept header
-            .filter((m) => m === '*/*')
+            .filter((m) => m !== '*/*')
```
habdelra, line 164: 🤦🏽

lukemelia (approved):
You were just off by one character!

PR #2461, opened 2025-04-21
Allow modules to invalidate missing imports

habdelra:
This PR updates our invalidation logic such that we can invalidate missing imports for modules. This allows us to upload a graph of modules in any order, such that as missing imports are resolved, the broken modules are reindexed. @richardhjtan this should allow you to do the copy functionality in the catalog app.

Also, I took the liberty of renaming the `loadCard()` function that actually loads card defs to `loadCardDef()` in a separate commit in this PR--it was annoying me too much.


lukemelia (approved):

richardhjtan (approved):
Try to merge the changes into the catalog app, and it works really well 👍🏻 Thanks @habdelra

PR #2462, opened 2025-04-21
Display error-detail in create-file-modal

burieberry:
- Use the CardErrorDetail component in create-file-modal dialog
- Refactor CardError component
- Update error message handling so the correct error message is displayed

error-detail


packages/host/app/services/store.ts
```diff
@@ -951,7 +951,7 @@ export default class StoreService extends Service implements StoreInterface {
   }
 }
 
-function formattedError(
+export function formattedError(
```
habdelra: let's move this into a more common place, how about runtime-common/error?

burieberry:
Converted this back to draft -- doing more improvements

PR #2467, opened 2025-04-22
value.data.id to get identity context

tintinthong:


packages/base/card-api.gts
```diff
@@ -983,12 +983,13 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
...
-    if (cachedInstance) {
-      cachedInstance[isSavedInstance] = true;
-      return cachedInstance as BaseInstanceType<CardT>;
+    let cachedInstance: CardDef | undefined;
+    if (value.data?.id) {
+      cachedInstance = identityContext.get(value.data.id);
+      if (cachedInstance) {
+        cachedInstance[isSavedInstance] = true;
+        return cachedInstance as BaseInstanceType<CardT>;
+      }
```
habdelra, line 992: i don't think this is necessarily right, how do you know that the relationship is sideloaded? because using value.data.id will only work when a relationship is side loaded. we don't only want to use the identity context when the data is side loaded--we want to use it _always_

habdelra: to be more clear--this is the logic that prevents you from having 2 different instances of the same card. skipping this is not optional just because the instance is not side loaded.

habdelra:
you really need to include a test that shows why this change needs to be made

packages/base/card-api.gts
tintinthong: I agree. If we do not assume the side loading data.id is available. This wud go back to relative link not working correctly (particularly when we copy card). We need the change in this PR perhaps https://github.com/cardstack/boxel/pull/2455/files

PR #2469, opened 2025-04-22
AI Bot: Abort LLM processing when an error was detected while processing chunks

jurgenwerk:
The change here is that when message is streaming and gets too big to send using Matrix (64kb limit), we abort streaming which results in no more incoming chunks. This also applies to any other errors that matrix event sending can return.

Previously, when message was streaming and got too big, we kept sending it to matrix which resulted in a flood of failed requests and errors sent to the user. This change should prevent that from happening again.

Moreover, my hypothesis is that this flood of error events also contributed to crashing of the ai bot and matrix container (which restarted and had downtime), likely due to a flood of requests. I'm saying this because in the case I was inspecting, these crashes happened during the same period when Chris experienced the "message too large bug". If crashes keep happening even with this fix, then we'll have to look more closely at what exactly is raising the memory pressure in these containers (their allocated memory is on the low side, only 0.5 GB)


packages/ai-bot/main.ts
habdelra, line 1: just a general observation: the fact that we have async event callbacks is problematic. that could cause ai bot performance bottlenecks as totally unrelated event callbacks are now chained together because you need to await one callback before you can handle the next event. I strongly suggest that we have a followup PR to replace all the async event callbacks in `main.ts` with a work queue that we call `drain()` on synchronously from the event callback. Then in the `drain()` method we can decide the best way to tackle the synchronicity (process work in parallel or in serial). as well as to keep track of the promise that the callback is not awaiting in a `flush()` property that can be called from tests, or as part of a teardown. this can follow the pattern that we've used elsewhere where we need to schedule async work in event callbacks (like the matrix event callbacks in host).

habdelra (approved):

lukemelia (approved):

packages/ai-bot/main.ts
jurgenwerk: I totally agree, it's getting a bit unwieldy, definitely need a way to schedule jobs in a queue for better work management. I created a [ticket](https://linear.app/cardstack/issue/CS-8516/improve-ai-bot-event-handling-by-setting-up-a-queue-that-gets-drained) for this.

PR #2471, opened 2025-04-22
Add _lint endpoint to realm

lukemelia:


packages/realm-server/tests/realm-endpoints/lint-test.ts
```diff
@@ -0,0 +1,306 @@
...
+import '@cardstack/runtime-common/helpers/code-equality-assertion';
+import { MatrixClient } from '@cardstack/runtime-common/matrix-client';
+import { type PgAdapter } from '@cardstack/postgres';
+import { resetCatalogRealms } from '../../handlers/handle-fetch-catalog-realms';
+
+setGracefulCleanup();
+const testRealmURL = new URL('http://127.0.0.1:4444/');
+const testRealm2URL = new URL('http://127.0.0.1:4445/test/');
+const distDir = resolve(join(__dirname, '..', '..', '..', 'host', 'dist'));
+console.log(`using host dist dir: ${distDir}`);
```
backspace: I removed this copypasted line in #2370 and also added some test setup helpers, did this come from an old copy of the repository?

packages/realm-server/tests/realm-endpoints/lint-test.ts
```diff
@@ -0,0 +1,306 @@
...
+    let dir: DirResult;
+    let testRealmHttpServer2: Server;
+    let testRealm2: Realm;
+    let dbAdapter2: PgAdapter;
+    let publisher: QueuePublisher;
+    let runner: QueueRunner;
+    let testRealmDir: string;
+    let seedRealm: Realm | undefined;
+
+    function setupPermissionedRealm(
```
backspace: There’s now a helper for this in `packages/realm-server/tests/helpers/index.ts`

backspace (changes requested):
I think there’s room for further extraction of realm server test setup but this is missing some that already happened, otherwise looks good!

packages/realm-server/tests/realm-endpoints/lint-test.ts
lukemelia: I pulled this from an old branch. I will update!

backspace (approved):
Much shorter, 👍🏻 assuming CI passes (which I expect it to)

PR #2474, opened 2025-04-22
Update card JSON POST and PATCH to support lid's in request body

habdelra:
This PR updates the card json POST and PATCH endpoints to support lid's in the request body. An `lid` essentially becomes a client suggested ID for the instance. In doing so, the client will be able to correlate newly created instances on the server (and Realm indexing events for these newly created instances) with local instances that do not yet have an ID. Whenever the server encounters a POST or PATCH request with an `lid` in a resource it will create a new instance. This allow a single POST or PATCH request to actually create multiple instances in the request. This will be useful in scenarios were we optimistically want to create a new instance and the newly created and unsaved instance happens to be used as a link in another instance that is being autosaved. The server will see that the client wishes to link to a not yet created instance and create the said instance as part of handling a POST or a PATCH for the consuming instance.

Also worth noting this provides low level support for indexing operations where more than one instance/module is updated at a time. this could be useful if we have an realm endpoint that allows us to update multiple files simultaneously, e.g. "Batch update", which we might want to do for performance reasons in the future: consider the copy operation that we use for the catalog app. /cc @tintinthong @lukemelia


packages/base/card-api.gts
```diff
@@ -1364,10 +1363,6 @@ class LinksToMany<FieldT extends CardDefConstructor>
             `linksToMany field '${this.name}' cannot deserialize a list of resource ids`,
           );
         }
-        if (isLocalResourceID(value.data)) {
-          throw new Error(`lid is not supported in deserialization`);
-        }
```
habdelra, line 1369: this logic is incorrect. its not that lid's are not supported in deserialization, it's that they are _ignored_. we should not being making type assertions like this as `lid` is a supported member of the JSON API spec.

packages/base/card-api.gts
```diff
@@ -977,9 +976,6 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
         `linksTo field '${this.name}' cannot deserialize a list of resource ids`,
       );
     }
-    if (isLocalResourceID(value.data)) {
-      throw new Error(`lid is not supported in deserialization`);
-    }
```
habdelra, line 982: ditto with above

packages/base/card-api.gts
```diff
@@ -995,7 +991,10 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
     //consider data.id when calling the resourceFrom() function (which actually loads the resource out of the included
     //bucket). we should never used links.self as part of that consideration. If there is a missing data.id in the resource entity
     //that means that the serialization is incorrect and is not JSON-API compliant.
-    let resource = resourceFrom(doc, value.data?.id);
+    let resource =
```
tintinthong, line 994: 👍

tintinthong (approved):

lukemelia (approved):

PR #2478, opened 2025-04-23
Make negative reference counts visible in console

habdelra:
In order to help debug CS-8499 (negative reference count). make negative reference counts visible in the console.

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -138,9 +138,9 @@ export default class StoreService extends Service implements StoreInterface {
     );
     if (currentReferenceCount <= 0) {
       if (currentReferenceCount < 0) {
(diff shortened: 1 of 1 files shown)


tintinthong (approved):

backspace (approved):
Could this and the `.debug` calls use a named logger so we can more easily filter them?

habdelra:
> Could this and the `.debug` calls use a named logger so we can more easily filter them?

i want this to be extremely loud, like an error. having a negative reference count is very bad.

backspace:
> i want this to be extremely loud, like an error. having a negative reference count is very bad.

Just the `debug` then? If I have another logger set to print `debug` I also see all the reference logging, even if I don’t want to.

PR #2480, opened 2025-04-23
Pass patched code through our linting endpoint for auto-fixes

lukemelia:


backspace:
Running this locally, I asked for an edit and only applied the patch with the `or` but not the (incorrect) import, should the lint fix have been applied?

asset gts in Experiments Workspace 2025-04-23 14-04-32

lukemelia:
> Running this locally, I asked for an edit and only applied the patch with the or but not the (incorrect) import, should the lint fix have been applied?

Your question made me realize I failed to hook into the single codeblock Apply case, only the Accept All path. I will fix!

lukemelia:
@backspace updated

backspace (approved):
success!

asset gts in Experiments Workspace 2025-04-24 09-01-13

jurgenwerk (approved):
Nice!

PR #2483, opened 2025-04-23
Fix timing for starting matrix in tests

habdelra:
Currently there is an issue where in host tests there are requests for `/_info` that hit the network stack instead of being handled internally. This is a result of the fact that we are starting the matrix service in our tests before we have mounted the realm on the virtual network when `setupMockMatrix()` has the `autostart` option set to `true`. This PR synchronizes the starting of the matrix service so that it only happens after the test realm has been mounted.


lukemelia (approved):

backspace (approved):
Thanks for figuring this out 🎉

PR #2484, opened 2025-04-23
Correlate local IDs on client with newly saved instances from server

habdelra:
This PR is a part of the larger effort to enable optimistic new card creation in the host. This adds low level store features that coordinate with the recent realm server updates around persisting instances that have local id's in the JSON-API payload. This work will correlate the newly created instances with the local ID's on the host.


packages/host/tests/integration/components/card-copy-test.gts
```diff
@@ -894,7 +894,7 @@ module('Integration | card-copy', function (hooks) {
   });
 
   test<TestContextWithSave>('can copy a card that has a relative link to card in source realm', async function (assert) {
-    assert.expect(16);
+    assert.expect(15);
```
backspace, line 897: This is a widespread thing but when I was removing `expectEvents` I was able to use `await` in a way to no longer require this, is there any hope of that for this use? I find it so annoying 😆

packages/host/tests/integration/store-test.gts
```diff
@@ -533,6 +533,63 @@ module('Integration | Store', function (hooks) {
...
+    assert.strictEqual(
+      store.getSaveState(instance[localId])?.isSaving,
+      false,
+      'isSaving state is correct',
+    );
+    assert.strictEqual(
+      store.getSaveState(instance.id)?.isSaving,
+      false,
+      'isSaving state is correct (by remote id)',
+    );
```
backspace, line 583: Does `assert.false` work for these? And `assert.true` above

backspace (approved):

jurgenwerk (approved):

packages/host/tests/integration/store-test.gts
habdelra: sure no prob

packages/host/tests/integration/components/card-copy-test.gts
habdelra: can you point me to an example?

PR #2486, opened 2025-04-24
AI bot: Fix process crashing by not processing old events (it's wasteful and unneeded)

jurgenwerk:
On AI bot process boot, it tried to process old events, because we were missing the time check. That lead to massive amount of fetching from synapse server, especially in the `client.on(RoomEvent.Timeline)` handler where it called `roomInitialSync` for each of the room of the old events. 

Since the ai bot resources are tiny ( .25 vCPU .5 GB) this quickly resulted in a crash. The container kept being restarted every ~5mins due to container being out of memory.

I tested this by deploying to prod and crashes are not happening anymore. Bot is up and running.

### DIFF
--- a/packages/ai-bot/main.ts
+++ b/packages/ai-bot/main.ts
@@ -158,7 +158,10 @@ Common issues are:
   let assistant = new Assistant(client, aiBotUserId);
   await assistant.loadToolCallCapableModels();
 
-  client.on(RoomMemberEvent.Membership, function (_event, member) {
+  client.on(RoomMemberEvent.Membership, function (event, member) {
+    if (event.event.origin_server_ts! < startTime) {
+      return;
+    }
     if (member.membership === 'invite' && member.userId === aiBotUserId) {
       client
         .joinRoom(member.roomId)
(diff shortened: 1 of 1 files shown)


IanCal (approved):

packages/ai-bot/main.ts
```diff
@@ -326,12 +329,14 @@ Common issues are:
 
   //handle set title by commands
   client.on(RoomEvent.Timeline, async function (event, room) {
-    if (!room) {
-      return;
-    }
-    if (!isCommandResultStatusApplied(event)) {
+    if (
+      event.event.origin_server_ts! < startTime ||
```
lukemelia, line 333: Don't we want to try to process events that might have come in while the container is rebooting (i.e. during deploys)? @IanCal @jurgenwerk

jurgenwerk: Perhaps some grace time would be useful, but on the other hand, would it not be a bit odd to receive answers to your prompts/commands in a delayed manner? Depends on how long the deploy/switchover process takes, if it's only a couple of seconds then it would make sense.

I think this is part of a larger question of how to handle graceful shutdowns and rolling deploys of the AI bot. If we do that properly then we won't have to think at what time the events are arriving.

Something like:
- when a new version is deployed, the load balancer should direct the traffic to the new instance
- in the "old" instance, wait for the AI response (streaming) to be fully done before terminating it

PR #2488, opened 2025-04-24
Copy card command returns a card id and not a running card

habdelra:
it's problematic returning a running card from a command as creates an opportunity for mishandling the card (i.e. the running card could end up in the state of a consumer). a better pattern is to return a card ID, and the consumer of the command can use the store to load the card (which will absolutely be cached in the store after the command runs--which means that a consumer of this command can actually just get the card synchronously from the store). Also, for card copying, all the consumers of this command only want the id of the copied card anyways, so its also unnecessary to do this.

@tintinthong @richardhjtan this should fix the field validation errors that you are seeing. i think the negative reference count is not related to this issue actually.

### DIFF
--- a/packages/base/command.gts
+++ b/packages/base/command.gts
@@ -35,7 +35,7 @@ export class CopyCardInput extends CardDef {
 }
 
 export class CopyCardResult extends CardDef {
-  @field newCard = linksTo(CardDef);
+  @field newCardId = contains(StringField);
 }
 
 export class CopySourceInput extends CardDef {
--- a/packages/host/app/commands/copy-card.ts
+++ b/packages/host/app/commands/copy-card.ts
@@ -50,25 +50,19 @@ export default class CopyCardCommand extends HostBaseCommand<
       }
       doc.data.meta.adoptsFrom = input.codeRef;
     }
-    let maybeId = await this.store.create(doc, undefined, targetUrl);
-    if (typeof maybeId !== 'string') {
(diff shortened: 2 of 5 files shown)


backspace (approved):

tintinthong (approved):

tintinthong:
This is a good resolution and I can't seem to see those errors. I am going to close my PRs up about the relative issues.

habdelra:
Note the deeper reason why this is a problem:

The problem is that the loader resets as there are code changes during the copy process (because we are copying gts files too), however it looks like the commands are cached, and the commands include a static dependency on their input and output card classes.

Inside our card api as part of setting a field we validate that the field value is an `instanceof` the fields class.
So for the result card of CopyCardCommand, this was `newCard` of type `CardDef`. But after the loader resets because an indexed gts file triggered a loader reset via realm events. The CardDef class is different. `instanceof` will not see `CardDef` instances between 2 different loaders as deriving from the same  `CardDef` class.

So we really need to be careful about hanging on to card classes in state, even when they are indirectly retained (like thru a cached command). Using a cached command to instantiate an input card instance might be using an input card class from an out of date loader which could cause problems. This fixes one manifestation of this problem in a very narrow case. But likely there are more issues in this space that need a more general solution. 

I think the moral of the story is, much like with card instances, we should never hold card classes in state because they need to change when the loader changes. (https://linear.app/cardstack/issue/CS-8522/do-not-cache-card-classes-in-command-service)

PR #2489, opened 2025-04-24
Ensure that all field defs in base use default export

lukemelia:
Consistency will help our AI overlords


backspace (approved):
I guess things are unlikely to drift away from this standard because it’s `base`? Otherwise I’d wonder about linting to monitor this

lukemelia:
Correct. We set the convention. We can't make people follow it.

PR #2495, opened 2025-04-25
Preserve open/closed state of AI Assistant panel on refresh

FadhlanR:
https://github.com/user-attachments/assets/c2302416-c7c1-4538-b1f1-77fcd8342189


backspace:
I’m a bit puzzled by these test failures but would it make sense to exercise this once in a single test vs having the assertions repeat in the eight tests that use the helper?

but regardless I look forward to having this, I tried to add it long ago and couldn’t get it to work

lukemelia (approved):

FadhlanR:
> I’m a bit puzzled by these test failures but would it make sense to exercise this once in a single test vs having the assertions repeat in the eight tests that use the helper?
> 
> but regardless I look forward to having this, I tried to add it long ago and couldn’t get it to work

Do you mean this assertion `await expect(page.locator('[data-test-room]')).toHaveCount(1);` ?

backspace:
> Do you mean this assertion `await expect(page.locator('[data-test-room]')).toHaveCount(1);` ?

Yes, and the conditional about the query parameter, I think all that only needs to happen once.

PR #2496, opened 2025-04-25
AI Assistant: Add a debug button to help debug AI code patch actions & diffs

jurgenwerk:
I am currently debugging why in some cases the code diff in proposed code patches fails to show in production

While I am trying to find a reliable way of reproducing this issue, it is really helpful to have a way of dumping code patch data (search/replace block especially) to console so that I'm able to inspect what the input data is (in cases where the code diff fails to be displayed)

The new button that dumps the code patch action data to the console (the button shows up only if you add `debug=true` query param to the URL):
image

image


backspace (approved):
nice! 
asset gts in mar10 2025-04-25 13-19-47

I think we’ll start to want a lot more diagnostic tools

lukemelia (approved):

PR #2497, opened 2025-04-25
AI bot: Remove obsolete waypoint files

jurgenwerk:
This is my suggestion to upgrade ai bot resources from 0.25 vCPU to 0.5, and from 512MB memory to 1024MB in prod. 

In my opinion current resources are really tiny. We recently saw the bot crashing (due to memory running out) when there was a spike of requests sent to Matrix. Since we are about to be testing more and more, getting users, etc, I think it is a good idea to add some more breathing room and ensure stability. 

Currently, the bot is consuming around 70% of memory just while idling:
image


backspace:
I’m not too sure about this but isn’t it true that Waypoint is no longer in use? I think the `ecs` module in the infrastructure config is the place to configure this. The [deployment setup doesn’t mention Waypoint at all](https://github.com/cardstack/boxel/blob/1f2dc1e247b2c7aee137ba71fd4510cd3d824f30/.github/workflows/manual-deploy.yml#L17-L37).

But I think this is definitely worth trying, as long as we check up on things and see the effects.

lukemelia (changes requested):
This change needs to be made in the `infra` repo. I believe the waypoint config shoud be removed at this point.

jurgenwerk:
@backspace @lukemelia You're right, I was confused. I removed these files now. There is now another PR for this in the Infra repo.

backspace (approved):

lukemelia (approved):

PR #2498, opened 2025-04-25
Ai assistant: fix another case when code diffing is not working

jurgenwerk:
I was able to reproduce a bug where the assistant offered me some code to apply but there was no actual diff shown:

image

So I was inspecting what the search/replace block looks like (using my logging tool from https://github.com/cardstack/boxel/pull/2496) and discovered the search block had code that **didn't belong to the currently attached gts file** - and that's why the diff did not show up. I actually asked the model to tell me where it got this code from and it told me:

image

So I'm adding some instructions to the source coding skill to never rely on previous suggestions as they might not be applied.

Better diff:
image


backspace (approved):
I appreciate the better diff again, it’s so difficult to review these changes without that.

lukemelia (approved):

richardhjtan (approved):

PR #2500, opened 2025-04-25
Test to handle circular dependency

richardhjtan:


packages/realm-server/tests/indexing-test.ts
```diff
@@ -1117,6 +1117,274 @@ module(basename(__filename), function () {
...
+      );
+
+      let cModule = await realm.realmIndexQueryEngine.module(
+        new URL(`${testRealm}c`),
+      );
+      assert.strictEqual(
+        cModule?.type,
+        'module',
+        'Module C is in resolved module successfully',
+      );
```
habdelra: these assertions don't match your comment regarding the expectation that these result in error documents

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1117,6 +1117,274 @@ module(basename(__filename), function () {
...
+          export class BlogApp extends CardDef {
+            @field title = contains(StringField);
+          }
+        `,
+      );
+
+      let blogPostModule = await realm.realmIndexQueryEngine.module(
+        new URL(`${testRealm}blog-post`),
+      );
+      // This is failing because the currently module is not being resolved
```
habdelra, line 1184: what is the error message in the error document that you get when you try to load the module?

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1117,6 +1117,274 @@ module(basename(__filename), function () {
...
```
habdelra, line 1281: what is the error document that you get when you try to load the module?

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1117,6 +1117,274 @@ module(basename(__filename), function () {
...
+          import { B } from "./b";
+
+          export class C extends CardDef {
+            @field a = linksTo(() => A);
+            @field b = linksTo(() => B);
+          }
+        `,
+      );
+
+      // Verify that trying to load any of these modules directly results in an error
```
habdelra: since these are thunks inside the `linksTo` i don't think it will be an error since we arent consuming the imported declaration in the module scope.  If you made them so they weren't thunks, then i think you'd get an error.

packages/realm-server/tests/indexing-test.ts
richardhjtan: Yes this scenario is working fine here. I will remove this to avoid the confusion

packages/realm-server/tests/indexing-test.ts
richardhjtan: Here is the error 
```
{
  type: 'error',
  error: {
    deps: [],
    status: 406,
    message: 'encountered error loading module "http://test-realm/blog-post.gts": unable to fetch http://test-realm/author',
    additionalErrors: null
  }
}
```

packages/realm-server/tests/indexing-test.ts
richardhjtan: @habdelra same as without thunk:
```
{
  type: 'error',
  error: {
    deps: [],
    status: 406,
    message: 'encountered error loading module "http://test-realm/blog-post.gts": unable to fetch http://test-realm/author',
    additionalErrors: null
  }
}
```

packages/realm-server/tests/indexing-test.ts
richardhjtan: Right, it is unusual that only the `blog-post` has the module error. 

The main difference I observe in `blog-post` is that it has one field using `linksToMany`, while other files - which use `linksTo`, are working fine. **_UPDATE:_** It doesn't matter whether it's using `linksTo` or `linksToMany`, tried either approach will break the test

habdelra:
@richardhjtan is there a ticket for this issue? if not could you make one and assign me to it? that way I can remember to circle back and make this failing test pass

richardhjtan:
Sure @habdelra , ticketed as https://linear.app/cardstack/issue/CS-8540/loader-fails-to-handle-circular-dependency

habdelra:
This has been fixed in https://github.com/cardstack/boxel/pull/2555

PR #2501, opened 2025-04-25
Fix issue for negative reference counts in store

habdelra:
The issue came down the to the `registerDestructor()` that we had in `CardResource`, `CardCollectionResource`, and `SearchResource`. In all these cases we were re-running the `registerDestructor()` for each invocation of the modify hook for these respective resources. Thus wiring up duplicative callbacks when the resource's owner was destroyed. This resulted in multiple extra calls to `Store.dropReference()` when the resource was destroyed making the reference count unbalanced. The answer was to make sure we only ever wire this up once.

(also made a minor update to `Store.addReference()` to make it easier to reason about).


packages/host/app/services/store.ts
```diff
@@ -166,13 +166,9 @@ export default class StoreService extends Service implements StoreInterface {
     }
   }
 
-  addReference(idOrDoc: string | SingleCardDocument | undefined) {
```
habdelra, line 169: it was confusing that `Store.addReference()` could also take in a document--cleaned this up

packages/host/app/services/store.ts
```diff
@@ -369,45 +376,40 @@ export default class StoreService extends Service implements StoreInterface {
     return await Promise.allSettled(this.newReferencePromises);
   }
 
-  private async wireUpNewReference(idOrDoc: string | SingleCardDocument) {
+  getReferenceCount(id: string) {
+    return this.referenceCount.get(id) ?? 0;
+  }
```
backspace, line 381: nice convenience for testing 🎉

packages/host/tests/integration/store-test.gts
```diff
@@ -875,4 +880,246 @@ module('Integration | Store', function (hooks) {
...
+      store.getReferenceCount(jade),
+      0,
+      `reference count for ${jade} is 0`,
+    );
+    assert.strictEqual(
+      store.getReferenceCount(hassan),
+      0,
+      `reference count for ${hassan} is 0`,
+    );
+  });
```
backspace, line 1124: I appreciate the granularity of these tests, might be time soon to split the file up!

backspace (approved):

lukemelia (approved):

tintinthong (approved):

richardhjtan (approved):

PR #2502, opened 2025-04-25
Clear playground card-creation error on change

burieberry:
Previously, playground displayed an error when create-new-instance failed, however, selecting a different definition in the file did not clear the error display. This will clear the error.

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
@@ -102,12 +102,14 @@ export default class PlaygroundPanel extends Component<Signature> {
 
   @tracked private cardResource: ReturnType<getCard> | undefined;
   @tracked private fieldChooserIsOpen = false;
+  @tracked private cardCreationError: CardErrorJSONAPI | undefined = undefined;
 
   private get moduleId() {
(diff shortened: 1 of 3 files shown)


backspace (approved):

lukemelia (approved):

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -102,12 +102,14 @@ export default class PlaygroundPanel extends Component<Signature> {
...
   @tracked private cardResource: ReturnType<getCard> | undefined;
   @tracked private fieldChooserIsOpen = false;
+  @tracked private cardCreationError: CardErrorJSONAPI | undefined = undefined;
 
   private get moduleId() {
     return internalKeyFor(this.args.codeRef, undefined);
   }
 
   private get isLoading() {
+    this.clearCardCreationError();
```
lukemelia, line 112: Normally, it is not good practice to have a side effect on a getter. Is there a root source of state that we are missing that is pushing us in this direction?

burieberry: Maybe there's a reactive way to handle this in the store for the failing card-creation case? @habdelra

PR #2507, opened 2025-04-28
Do not reset loader for clientRequest prefixed with copy:

tintinthong:
When we save source, it causes a loader reset so instances that depend on the source can be updated
This caused a scenario when we click "install" in st 2 stacks will re-render and have all identity context get destroyed

Screenshot 2025-04-28 at 20 38 06

Screenshot 2025-04-28 at 20 32 37

**More Thoughts**

So rightfully, there certainly is some race condition occuring in how validations are handled but I can't pin point it
- this error doesn't occur when a single stack item is used
- this error typically occurs whenever you "install" a listing for a 2nd time

This may mean "a deeper issue"??

**Solution**
However, I believe it is not a "possible" scenario that the source that is newly copied will have instances that depend on it. So I executed a change to ignore all clientRequestIds prefixed with 'copy:' from handleInvalidations; this comes from copySource in the card service

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -488,7 +488,10 @@ export default class StoreService extends Service implements StoreInterface {
     }
     let invalidations = event.invalidations as string[];
 
-    if (invalidations.find((i) => hasExecutableExtension(i))) {
+    if (
+      invalidations.find((i) => hasExecutableExtension(i)) &&
+      !event.clientRequestId?.startsWith('copy:')
+    ) {
       // the invalidation included code changes too. in this case we
       // need to flush the loader so that we can pick up any updated
       // code before re-running the card


habdelra:
is it always the case that the realm you are copying to does not have the same card def that you are copying?
basically stomping on an existing card def? if you already have a black jack game in the destination realm, that lives at the same file location as the black jack game you are trying to copy so that they collide, that would require a loader reset. Perhaps this could be a scenario in which you want to copy an updated version to a realm that previously had the same app installed in it.

habdelra:
Based on our previous discussion it sounds like this PR was created because the loader reset causes the catalog app to flash. Instead of doing this we should really investigate why the catalog app is flashing on a loader reset and fix that. Otherwise we’re just kicking the can down the road until some other thing triggers a loader refresh when the catalog app is displayed. In general loader refreshes should not cause a rendered card to flash.

tintinthong:
> Based on our previous discussion it sounds like this PR was created because the loader reset causes the catalog app to flash. Instead of doing this we should really investigate why the catalog app is flashing on a loader reset and fix that. Otherwise we’re just kicking the can down the road until some other thing triggers a loader refresh when the catalog app is displayed. In general loader refreshes should not cause a rendered card to flash.

Agreed. Im investigating deeper here.

tintinthong:
> Based on our previous discussion it sounds like this PR was created because the loader reset causes the catalog app to flash. Instead of doing this we should really investigate why the catalog app is flashing on a loader reset and fix that. Otherwise we’re just kicking the can down the road until some other thing triggers a loader refresh when the catalog app is displayed. In general loader refreshes should not cause a rendered card to flash.

@habdelra I have identified the main issue; it has to do with the card-resource ability to re-run during a loader reset 

The sequence of the issue goes like this
- invalidation of a .gts file comes in due to copySource 
- the stack item already has a card-resource 
- the loader and identityContext gets thrown away 
- `isLoaded` getter in the card resource uses `store.peek` to look for an instance in the store but the instance has been thrown away 
-  the cardResource still exists with a private `this.#id` and doesn't wire up a NEW reference so it stays in `get isLoaded=false` state

**Possible Solution**
1 possible solution here is to pass in the loader nonce inside to the card resource as an argument. If the loader nonce has changed, the card resource is smart enough to know that the internal id is stale and can proceed to wire up a new reference in the store 

However, this will still mean that we experience a loading indicator anyway (this is actually typical when a stack item is opened and we edit its gts; the whole stack re-redners). However, It does seem quite unnatural that a install button clicked in a listing card will lead to it being re-rendering but this is due to the full loader reset.

tintinthong:
A deeper question that I have now is. "How come we never noticed this whenever 2 clients are opened?".  Based upon my testing, this bug is not experienced at all when editing gts in 1 and having stacks opened in the other

EDIT: So it seems fine. When you edit a gts file in another client, an `editor` is appended to the clientRequestId and you wud get returned an array of invalidations that includes 

- the gts
- instances that the gts depends on

Within the card resource, the instances will be reloaded even after their identities get thrown away

When we copy listing, that array DOES not include the invalidation of the cards that you see on a stack item, particularly, its not an instance of your new gts you copied. Hence, your stack items crash

One can re-create the issue by opening an instance that is not related to the gts file that ur editing. 

The behaviour is even worse in that a "refresh" will NOT recover the stack-item. (But I think this might be a separate issue of referenceCounting)

tintinthong:
There also exists a race condition when relaying an invalidation to matrix bcos of a COPY. In my scenario a copy consist of copying source and then an instance

It depends solely on whether the messageService.listenerCallback is registered in a subscription upon receiving an invalidation. This explains why the copy is actually occasionally succesful with no reload of the state (but the consequences of this might be bad). 

Reason for succesful copy (which is misleading) is bcos. 
- The subscription of the instance is never setup yet by the store
- the incremental copy: event comes in 
- the event is handled by all subscription handlers ( but not `handleInvalidations` hook)
- the incremental instance: event comes in 
- the event is handled by all subscription handlers ( including `handleInvalidations` hook) 
- but the `handleInvalidations` hook is only destructive for source fixing events like `:copy`

habdelra:
> There also exists a race condition when relaying an invalidation to matrix bcos of a COPY. In my scenario a copy consist of copying source and then an instance
> 
> It depends solely on whether the messageService.listenerCallback is registered in a subscription upon receiving an invalidation. This explains why the copy is actually occasionally succesful with no reload of the state (but the consequences of this might be bad).
> 
> Reason for succesful copy (which is misleading) is bcos.
> 
> * The subscription of the instance is never setup yet by the store
> * the incremental copy: event comes in
> * the event is handled by all subscription handlers ( but not `handleInvalidations` hook)
> * the incremental instance: event comes in
> * the event is handled by all subscription handlers ( including `handleInvalidations` hook)
> * but the `handleInvalidations` hook is only destructive for source fixing events like `:copy`

what if you copy the instance first and then the source? (the indexer can to handle any order and fix once all files are present)

habdelra:
We should probably close this

PR #2512, opened 2025-04-28
Sets a concurrency group on lint-source jobs

lukemelia:
- concurrency_group string includes a random character from 1 to 9, to allow for some concurrency within a realm but not unlimited


backspace (approved):
If `concurrencyGroup` is seemingly required, should the types be changed?

lukemelia:
> If `concurrencyGroup` is seemingly required, should the types be changed?

I'm not convinced of that just yet. Blank was originally supposed to mean no concurrency limit

PR #2513, opened 2025-04-28
Extract ESLint dependencies to catalog

backspace:


packages/host/config/fastboot.js
```diff
@@ -13,7 +13,9 @@ module.exports = function () {
     buildSandboxGlobals(defaultGlobals) {
       return Object.assign({}, defaultGlobals, {
         URL: globalThis.URL,
+        // eslint-disable-next-line n/no-unsupported-features/node-builtins
         Request: globalThis.Request,
+        // eslint-disable-next-line n/no-unsupported-features/node-builtins
```
backspace, line 18: These are new because I chose the newest version of `eslint-plugin-n` to use in the catalog and this package was behind.

lukemelia (approved):

PR #2516, opened 2025-04-28
lint: Add detection of unused disable directives

backspace:
In #2513 I added some `eslint-disable` directives that will become obsolete when we move to a newer Node version. Using the `--report-unused-disable-directives` flag causes ESLint to error on directives that don’t actually silence any errors, much like `@ts-expect-error` lets us know if formerly-missing type imports (for instance) are no longer missing.

`--fix` works on these, so the PR includes the fixes for the errors [the flag exposed](https://github.com/cardstack/boxel/actions/runs/14717788892/job/41305142427).

The Percy diff is spurious 😞


habdelra (approved):
nice!

PR #2517, opened 2025-04-28
Update aibot to make prompt specify when a SEARCH/REPLACE block has been applied

lukemelia:
This PR adds CodePatchResultEvent, a matrix event that is created when a code patch is applied, to memorialize the fact that is was applied in matrix history. This is used to show the correct button state in the UI and to reflect that the change was applied in the prompt we build for subsequent LLM messages.


jurgenwerk (approved):
Works great! I tested this manually in known scenarios, even with multiple code patches.

PR #2521, opened 2025-04-29
Set default skills in initial_state of room creation

FadhlanR:
This is another part of the effort to improve room creation time. The idea here is to set the default skill during room creation, instead of making a separate call after the room is created. Before this PR, room creation time was around 5–6 seconds. It has improved now, as shown in the screenshot below.

Screenshot 2025-04-29 at 15 21 21

### DIFF
--- a/packages/host/app/components/ai-assistant/panel.gts
+++ b/packages/host/app/components/ai-assistant/panel.gts
@@ -462,26 +462,17 @@ export default class AiAssistantPanel extends Component<Signature> {
 
   private doCreateRoom = restartableTask(async (name: string) => {
     try {
-      console.time('Total room creation time');
-      console.time('Load default skills');
       const defaultSkills = await this.matrixService.loadDefaultSkills(
         this.operatorModeStateService.state.submode,
       );
-      console.timeEnd('Load default skills');
-
-      console.time('CreateAiAssistantRoomCommand execution');
       let createRoomCommand = new CreateAiAssistantRoomCommand(
         this.commandService.commandContext,
       );
       let { roomId } = await createRoomCommand.execute({ name, defaultSkills });
-      console.timeEnd('CreateAiAssistantRoomCommand execution');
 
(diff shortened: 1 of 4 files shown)


lukemelia (approved):

packages/host/app/components/ai-assistant/panel.gts
```diff
@@ -464,29 +463,18 @@ export default class AiAssistantPanel extends Component<Signature> {
...
+      let createRoomCommand = new CreateAiAssistantRoomCommand(
         this.commandService.commandContext,
       );
-      await addSkillsToRoomCommand.execute({
-        roomId,
-        skills: defaultSkills,
-      });
-      console.timeEnd('AddSkillsToRoomCommand execution');
+      let { roomId } = await createRoomCommand.execute({ name, defaultSkills });
+      console.timeEnd('CreateAiAssistantRoomCommand execution');
```
jurgenwerk: Timers can be removed?

jurgenwerk:
I have a question–today I was updating skill cards locally and had a hard time debugging why they aren't updated when the assistant sends them to the bot. I realized I need to enable and disable them in the skill menu, otherwise they are cached. Is this expected? Is there / will there be a mechanism to update the default skills when they are changed?

jurgenwerk (approved):
Looks ok but I posted a question regarding cached skills.

FadhlanR:
> I have a question–today I was updating skill cards locally and had a hard time debugging why they aren't updated when the assistant sends them to the bot. I realized I need to enable and disable them in the skill menu, otherwise they are cached. Is this expected? Is there / will there be a mechanism to update the default skills when they are changed?

For now, this is expected since there's no mechanism to detect updates on skill cards or to re-upload them. I'm not sure if we have plans to provide such a mechanism—would be good to get input from @lukemelia on this.

lukemelia:
Yes, there is a Linear issue for this.

PR #2524, opened 2025-04-29
Spread out large AI responses over multiple matrix events

lukemelia:
Matrix has a maximum 64K size for the content of one event. aibot sometimes exceeds that when the model is generating a lot of code. This change introduces a new idea called "continuation events" where we can spread reasoning and/or body response content over multiple Matrix events, but have them treated as a single message for presentation and prompt generation purposes.

A CardMessageEvent (matrix event) will have a content key `app.boxel.has-continuation: true` if it has an event with additional content. The continuation event (also a CardMessageEvent) will have a content key `app.boxel.continuation-of: [event ID]` pointing to the event that it is "continuing". These keys can create a linked list of arbitrary length.

Tool calls present a particular problem that doesn't fit well with the solution outlined here. The OpenAI SDK converts streaming JSON content to snapshotted POJO that we are using. This makes the concept of chunking them up awkward. In addition we change the format of the tool call somewhat into a CommandRequest (also a JSON structure), so there's an extra layer of complexity there. For this PR, we decided to try to spread tool calls across multiple events. Since the observed problems to-date are from big codeblocks, this may be sufficient for quite a while.

- Note that this pull request includes some significant refactoring to the `ai-bot` package, focusing on modularizing code, improving maintainability, and enhancing the handling of Matrix events and responses. Key changes include moving the `constructHistory` function to a new history.ts file, introducing a `MatrixResponsePublisher` class for creating Matrix events to capture AIBot responses, and refactoring the `Responder` class to delegate responsibilities to the new publisher and a centralized response state.

- Ignoring the percy failure, which appears to be due to a timing issue of code highlighting.


jurgenwerk:
EDIT: this issue is not related to this PR. It's coming from upstream: https://github.com/cardstack/boxel/pull/2553

_____ 
I'm seeing a degradation, the code diff is not showing after streaming is done (but it will appear if I reload the page):

image

image

jurgenwerk:
I tried testing this by forcing sonnet 3.7 to respond with a huge message (more than 50KB) and the response got split up into multiple messages, where it appears that their content is related but incorrectly aggregated. I didn't find any errors in the AI bot, or in the browser console

Here is the message thread (I stitched it together using multiple screenshot):
chunked-msg

lukemelia:
Needs more work. Moving back to draft.

jurgenwerk:
I'm testing on this branch with all its updates, and getting the LLM to respond with a couple of base64 encoded images to test large responses and it was streaming up to around 38kb of text, and then another message showed up, which is unexpected, and streaming hangs:

image



When I reload the page the streaming resumes but it's scattered around in multiple messages, looks like the aggregation still has issues:

image

habdelra (approved):

jurgenwerk (approved):
I just tried streaming a message that contains 69KB of text and it worked OK - all the content is rendered in a single message. Really nice!!!

PR #2525, opened 2025-04-29
Pre-select card for playground preview if no card was selected

burieberry:
For CardDefs, if there's no selected card instance to display in playground preview, auto-select the most recent card from recent-cards. If you don't have any matching recent-cards, then the preview will still be empty (potentially auto-generating a card is covered in another ticket).

I commented in the code where the meat of the changes is. The rest is refactor and removal of duplicate code in playground files and the addition of `removeJsonExtension` util.


packages/host/app/components/operator-mode/code-submode/playground/instance-chooser-dropdown.gts
```diff
@@ -151,140 +159,154 @@ interface Signature {
...
+      if (!s) {
+        // if there's no selected card, choose the most recent card as selected
+        this.playgroundPanelService.persistSelections(
+          this.args.moduleId,
+          trimJsonExtension(prerenderedCards[0].url),
+          'isolated',
+          undefined,
+        );
+      }
+      return prerenderedCards[0];
```
burieberry, line 309: This is where we're persisting and returning the first available card if there was no selected card. Attempting to call `this.args.onSelect(prerenderedCards[0])` causes revalidation error. Using `this.playgroundPanelService.getSelection` causes revalidation error. That is why I added the `peekSelection` to the panel service.

backspace (approved):
I tried this out with the staging preview deployment, success!

PR #2526, opened 2025-04-30
Reduce chances of assistant using patchCard when user wants to edit gts file

jurgenwerk:
This situation was getting very problematic, especially in Claude 3.7 Sonnet:

image

image


packages/ai-bot/helpers.ts
```diff
@@ -681,14 +681,24 @@ Attached files:
 ${attachedFilesToPrompt(attachedFiles)}
 `;
 
+  let cardPatchTool = tools.find((tool) => tool.function.name === 'patchCard');
+  let sourceCodeEditingSkillCard = skillCards.find(
+    (card) => card.attributes?.title === 'Source Code Editing',
+  );
+
+  if (cardPatchTool && sourceCodeEditingSkillCard) {
+    systemMessage += `First determine if the user's request involves template editing, visual changes, or styling, or card logic changes.
```
jurgenwerk: Based on my experiments it's far more reliable if we put this message at the beginning, before skill prompts are added.

packages/base/SkillCard/boxel-coding.json
```diff
@@ -2,7 +2,7 @@
...
...ng: 0;\n        }\n        .info > li + li {\n          border-left: 1px solid black;\n          padding-left: var(--boxel-sp-xs);\n        }\n        .byline {\n          display: inline-flex;\n          align-items: center;\n          gap: var(--boxel-sp-xs) var(--boxel-sp);\n          flex-wrap: wrap;\n        }\n        .author {\n          display: contents; /* workaround for removing block-levelness of atom format */\n        }\n        .author-embedded-bio {\n          margin-top: var(--boxel-sp-xl);\n        }\n        .categories {\n          margin-top: var(--boxel-sp);\n          display: flex;\n          flex-wrap: wrap;\n          gap: var(--boxel-sp-xxs);\n        }\n        .featured-image + .categories {\n          margin-top: var(--boxel-sp-xl);\n        }\n        .category {\n          display: inline-block;\n          padding: 3px var(--boxel-sp-xxxs);\n          border-radius: var(--boxel-border-radius-sm);\n          font: 500 var(--boxel-font-xs);\n          letter-spacing: var(--boxel-lsp-sm);\n        }\n      </style>\n    </template>\n  };\n}\n```\n\n <style> tag must be at the root of the <template>, otherwise the app will break. It must not be nested.",
```
jurgenwerk: Added: `<style> tag must be at the root of the <template>, otherwise the app will break. It must not be nested.`. I got this error after code patch so I thought it would be a good idea to add this instruction

packages/base/SkillCard/card-editing.json
```diff
@@ -2,7 +2,7 @@
...
...able, you MUST ask the user to open the card and share it with you.\n- If you do not call patchCard, the user will not see the change.\n- You can ONLY modify cards shared with you. If there is no patchCard function or tool, then the user hasn't given you access.\n- NEVER tell the user to use patchCard; you should always do it for them.\n- If the user wants to search for a card instance, AND the \"searchCard\" function is available, you MUST use the \"searchCard\" function to find the card instance.\nOnly recommend one searchCard function at a time.\nIf the user wants to edit a field of a card, you can optionally use \"searchCard\" to help find a card instance that is compatible with the field being edited before using \"patchCard\" to make the change of the field.\n You MUST confirm with the user the correct choice of card instance that he intends to use based upon the results of the search.\n- Do NOT use patchCard function when user is asking for template changes, visual modifications, style updates, or card behavior changes. patchCard should ONLY be used when explicitly changing card data values bound to the card definition in gts files, or data relationships, and nothing else.",
```
jurgenwerk: Added: `- Do NOT use patchCard function when user is asking for template changes, visual modifications, style updates, or card behavior changes. patchCard should ONLY be used when explicitly changing card data values bound to the card definition in gts files, or data relationships, and nothing else."`

packages/base/SkillCard/source-code-editing.json
```diff
@@ -2,7 +2,7 @@
...
...you propose a search/replace block, it must be for the currently attached file(s), and not for those attached before the most recent one (unless you ask and get the user's approval). \n\n Your new SEARCH/REPLACE blocks must target ONLY the content of currently attached files - the search portion must not target any of your previous suggestions since it is not guaranteed that your previous SEARCH/REPLACE blocks were applied.\n\nIf you recognize the user wants to edit a template, do a visual change to a card, or describe a certain style, then you must use a SEARCH/REPLACE block to perform an edit to the attached gts file, by default in the isolated template. Do not default to using the patchCard tool function, unless the user asks you to change the supporting data of the card (bound model values for fields).\n\nWhen you respond with a SEARCH/REPLACE block, do not refer to it as a SEARCH/REPLACE block, as this is an internal structure that will get shown to user in a different format. If you need to refer to it, talk about it in a semantic way. For example, do not say 'I'll use a SEARCH/REPLACE block to add a template', but rather 'I'll add a border around the section', for example.",
```
jurgenwerk: Added: `if you recognize the user wants to edit a template, do a visual change to a card, or describe a certain style, then you must use a SEARCH/REPLACE block to perform an edit to the attached gts file, by default in the isolated template. Do not default to using the patchCard tool function, unless the user asks you to change the supporting data of the card (bound model values for fields).\n\nWhen you respond with a SEARCH/REPLACE block, do not refer to it as a SEARCH/REPLACE block, as this is an internal structure that will get shown to user in a different format. If you need to refer to it, talk about it in a semantic way. For example, do not say 'I'll use a SEARCH/REPLACE block to add a template', but rather 'I'll add a border around the section', for example`

jurgenwerk:
I think we wouldn't need to bother explaining this so much if we renamed `patchCard` to something more exact. Right now it's confusing and ambiguous because it's not clear what exactly it's patching.

habdelra:
> I think we wouldn't need to bother explaining this so much if we renamed `patchCard` to something more exact. Right now it's confusing and ambiguous because it's not clear what exactly it's patching.

Agreed. perhaps patch instance. and maybe we can discuss in the skill that a card encapsulates behavior and visual design which is implemented as a class, and an instance is an instantiation of a card class that is populated with data. And then we rename `patchCard` to `patchInstance`. In general even in our own code it can be super confusing about whether we are talking about instances or card classes as we use the word "card" so interchangeably. I think AI is just picking up on that confusion. This is actulaly what inspired me to clean up the terminology in the store service as it started to get really hard to reason about.

lukemelia:
Maybe `UpdateCardInstance`

packages/ai-bot/helpers.ts
lukemelia: I don't really like the idea of special casing some skills in the aibot.

jurgenwerk: Right, as discussed I think renaming the ambiguous `patchCard` could be beneficial in a way where we won't have to include special cases

jurgenwerk:
Ok I'll try to rename to reduce confusion.

burieberry (approved):
The name change looks good to me

jurgenwerk:
This is now ready for another review. Key things:

- Compared to my original PR, I removed extra instructions added to the system prompt. Extra instructions are now added in skills only
- `patchCard` was removed to `patchCardInstance`
- Added some slight adjustments to skills instructions

The result of that is that I am now unable to reproduce this type of problems where the assistant will fail to understand you want to change the card definition (and will use patchCard instead, which doesn't make too much sense):
image

image

lukemelia (approved):

.github/workflows/diff-skills.yaml
```diff
@@ -30,8 +30,8 @@ jobs:
           # Create a temporary directory for processing
           mkdir -p /tmp/skill-diffs
 
-          # Get the list of changed JSON files
-          CHANGED_FILES=$(echo "${{ steps.changed-files.outputs.all_changed_files }}" | tr ' ' '\n')
+          # Get the list of changed JSON files (with a "Skill" in the file path)
+          CHANGED_FILES=$(echo "${{ steps.changed-files.outputs.all_changed_files }}" | tr ' ' '\n' | grep -E 'Skill')
```
jurgenwerk, line 34: There are several other json files in our repo, we aren't interested in those here, and their content isn't what diff-skills expects it to be, leading to parsing errors

PR #2527, opened 2025-04-30
Extract some build dependencies to catalog

backspace:
This is a collection of build-related dependencies whose versions are now extracted to the pnpm catalog. I notably excluded `babel-plugin-ember-template-compilation` because it caused build failures. I didn’t include any Embroider changes as those seemed another likely culprit for causing failures in the bigger PR, but there’s a patch for `@embroider/compat` to not fail when encountering `catalog:` instead of a version for `@babel/core`.

The Percy diff is spurious, separate PR [here](https://github.com/cardstack/boxel/pull/2528).


habdelra:
babel-plugin-ember-template-compilation will probably be going away soon as we upgrade to the latest ember template compilation in https://github.com/cardstack/boxel/pull/2529

habdelra (approved):

habdelra:
or wait nevermind--we'll still use babel-plugin-ember-template-compilation for tests...

backspace:
ah well maybe reliance on it will decrease at least, at some point maybe I/we can dig into what’s wrong when using the newer version

PR #2529, opened 2025-04-30
Upgrade template compilation to use content-tag

lukemelia:


backspace:
The problem in the Actions environment:

server-9 log 2025-05-02 13-19-35

is the same I see locally:

realm-server 2025-05-02 13-18-55

It’s also the same problem that led me to abandon [this pnpm catalog PR](https://github.com/cardstack/boxel/pull/2481). When I was working on that it was confusing because I was originally able to run the realm server successfully locally and the `process.env` problem only showed up in Actions, but then when I switched away and back to the branch, I saw the problem locally. I spent some time trying to debug but got nowhere and eventually gave up.

In this PR, it only started when `vendor/ember-template-imports` was removed.

I’ve tried again today to figure out a fix but have been unsuccessful 😞

packages/host/ember-cli-build.js
```diff
@@ -74,6 +74,9 @@ module.exports = function (defaults) {
           plugins: [
             new GlimmerScopedCSSWebpackPlugin(),
             new MonacoWebpackPlugin(),
+            new webpack.DefinePlugin({
+              'process.env.NODE_DEBUG': JSON.stringify(process.env.NODE_DEBUG),
+            }),
```
backspace: In my attempts with this I also tried a workaround for `process.env.NODE_DEBUG` but the underlying problem was that `process.env` just didn’t exist, so the next failure was another attempt to use it. It seems like something is weird with the build environment that it wouldn’t exist anymore.

I was able to see this failure locally, maybe if you check out a fresh repository you can too so you don’t have to debug via CI?

packages/runtime-common/helpers/code-equality-assertion.ts
```diff
@@ -47,22 +47,18 @@ function standardize(code: string) {
 function preprocessTemplateTags(code: string): string {
   let output = [];
   let offset = 0;
-  let matches = parseTemplates(code, 'no-filename', 'template');
+  let matches = new ContentTagGlobal.Preprocessor().parse(code);
```
lukemelia, line 50: @ef4 Looks like we need this not only in tests, but also in the app. We attempt to run this in module-syntax.ts via the ModuleContentsResource when loading a module in code mode.

jurgenwerk (approved):
I tested this and looks like the issue is gone! Nice one

image

PR #2532, opened 2025-04-30
Support live updating card rendering error on create

habdelra:
In the scenario where we are creating a card and there is a rendering error, the POST for the instance will fail, but the  instance is persisted in the realm file system since the error rendering error is only realized at index time, which happens _after_ the file is written to the realm's file system. In such a scenario we use the local ID that the host knows about in the error doc to live update the card error with the running instance in the store using the local ID as the correlation key between the card error and the eventual fixed instance.

However, making the store more faithful for live updating in and out of errors opens up a can of worms, in that our UX seems to not reflect this reality. There were some scenarios (bugs really) where we were not live updating the card thru an error state and now we are. so for example in saving a card, if there is an error, the card will live update to the  error doc were it was not before (and it should live update out of the error as well). the error doc is _not_ a running card so in some scenarios the error was not severe enough to prevent rendering the edit format. but now in this particular scenario the identity context for the URL has an error doc and not a running instance (honestly it always did but due to bugs we were not being reactive enough about that fact), so we just never show a running instance anytime there is an error doc for the card (and we show a last known good state if that exists)--probably this is more correct. Unsure if we will need updated visual design for this. In this example the test used to show an auto save error message and the rendered card, but now we literally live update to error doc (in this case there happens to be no last known good state). before this commit the cached running instance may have been stable enough (depending on the error) to sill be able to render in the edit format and allow the user to get out of the error. Now you would always have to to into code mode to get out of an error state that is fixable via a data update.

I think keeping around the old instances when you are in an error state is possible, but it feels like it might need some deeper thinking from a visual design standpoint and an architectural standpoint. like how do we want to model this in the store? what visual affordance do you want to show when the instance you are looking at is stale, how do consumers know that the instance they are looking at is stale. I'm not too sure how likely such a scenario is--like a card that does not have a custom edit template would be a card that is more likely to survive an error that would otherwise make the card totally inoperable.

What do folks feel like? is this PR worth landing or should we do the deeper thing and work on caching stale instances in the store when the server informs us that the instance is in an error state on the server?

https://github.com/user-attachments/assets/d01841ee-3e52-4946-aaeb-957298bfc422

### UPDATE:
 The scenario above is no longer an issue. I've updated the store to understand stale cards such that the local state can diverge from the server state when the server state goes into an error for a particular card. This does mean a couple UI tweaks that we'll need to make to visualize these divergence which i filed in a ticket here: https://linear.app/cardstack/issue/CS-8553/in-playground-mode-when-there-is-a-card-error-error-show-message-in


packages/host/tests/integration/components/operator-mode-test.gts
```diff
@@ -892,18 +892,8 @@ module('Integration | operator-mode', function (hooks) {
...
-        document
-          .querySelector('[data-test-auto-save-indicator]')
-          ?.textContent?.trim() == 'Failed to save: Boom!',
-    );
-    setCardInOperatorModeState(`${testRealmURL}BoomPet/paper`);
-
-    await waitFor('[data-test-pet]');
-    // Card still runs (our error was designed to only fire during save)
-    // despite save error
-    assert.dom('[data-test-pet]').includesText('Paper Bad cat!');
```
habdelra: this is the test for which i recorded a video in the PR description. us used to be able to show a running card because in this particular scenario the live update was not actually working. now the live update is much more consistent thru error states.

packages/host/tests/integration/components/operator-mode-test.gts
```diff
@@ -947,16 +937,8 @@ module('Integration | operator-mode', function (hooks) {
...
-    await waitUntil(
-      () =>
-        document
-          .querySelector('[data-test-auto-save-indicator]')
-          ?.textContent?.trim() == 'Failed to save: Rejected by firewall',
-      { timeoutMessage: 'Waiting for "Failed to save" to appear' },
-    );
-    assert
-      .dom('[data-test-auto-save-indicator]')
-      .containsText('Failed to save: Rejected by firewall');
```
habdelra: This is another example of a live transition into an error state since the identity map replaces the running instance with the error. i think in this case we never have a last known good state as the error originates from the firewall and not the indexer

habdelra:
The more I think about it, the more it feels like we need to do the deeper thing: provide a running card if we have one even if we know the server state is a card error. We can provide in our card resource the current server state, so the caller can make the decision on which to reflect. Generally I think we’ll want to always show the current server state except if you are currently editing a card. Then I think we’ll need to show the stale card (with the hope that it is still renderable) and a message in the header reflecting that the current server state is in error. So probably from the consumer perspective not a huge divergence.

lukemelia:
Seems like our plans for finer-grained errors could help us best respond, too.

habdelra:
> Seems like our plans for finer-grained errors could help us best respond, too.

yeah, that's part of it--a finer grained error could return server state that includes a card that can be run. right now its a very binary decision... and this doesn't even get into the scenario where the error blows up thru the edit template. in that case the whole app blows up, as we don't have a way of isolating errors in the host template (e.g. try/catch helpers in the template) like we do in the indexer.

habdelra:
So after going thru this i think we just need a small visual update to deal with the fact that we prefer rendering stale cards when the server is in an error state. specifically in playground mode when there is an auto save error we should show the error message in the header like we do in isolated mode. I made a ticket for this: https://linear.app/cardstack/issue/CS-8553/in-playground-mode-when-there-is-an-autosave-error-show-message

/cc @burieberry

packages/host/app/services/store.ts
```diff
@@ -869,13 +918,14 @@ export default class StoreService extends Service implements StoreInterface {
...
-  ): Promise<void> {
-    await this.withTestWaiters(async () => {
+  ): Promise<CardDef | CardErrorJSONAPI> {
+    return await this.withTestWaiters(async () => {
       let cardChanged = false;
       function onCardChange() {
         cardChanged = true;
       }
       let api: typeof CardAPI | undefined;
+      let isNew = !instance.id;
```
burieberry, line 928: So in the original case described in the ticket, where the card is created but could not be rendered, returning `isNew` to be `true` means `cardError.meta.isCreationError` will be `true` and playground will display `Failed to create card` message, which seems wrong. Only because the file does actually appear in the system. 

If this will be unnecessarily tricky to sort out, we could also just get rid of that message and just say there's an error.

habdelra: Yeah, I was just trying to be true to the current functionality as there was another test that relied on this. In the scenario you mentioned I think I can have this flag be false.

packages/host/tests/acceptance/code-submode/card-playground-test.gts
```diff
@@ -1199,7 +1199,7 @@ module('Acceptance | code-submode | card playground', function (_hooks) {
       assert.dom('[data-test-error-detail]').hasText('Boom!');
     });
 
-    test('it renders error info when creating new instance causes error', async function (assert) {
+    test('it renders error info when creating new instance causes error after file was created in realm', async function (assert) {
```
habdelra, line 1202: @burieberry based on your comment I tweaked this test to be more reflective of the fact that in this particular scenario the instance was actually created in the realm

packages/host/tests/acceptance/code-submode/card-playground-test.gts
```diff
@@ -1234,15 +1236,15 @@ module('Acceptance | code-submode | card playground', function (_hooks) {
       assert.dom('[data-test-error-container]').doesNotExist();
     });
 
-    test('it can clear card-creation error when file is edited', async function (assert) {
+    test('it can clear card-creation error (that resulted in new file in the realm) when file is edited', async function (assert) {
```
habdelra, line 1239: ditto

packages/host/tests/acceptance/code-submode/card-playground-test.gts
```diff
@@ -1263,15 +1265,15 @@ module('Acceptance | code-submode | card playground', function (_hooks) {
       assert.dom('[data-test-error-container]').doesNotExist();
     });
 
-    test('it can clear card-creation error when different card-def is selected', async function (assert) {
+    test('it can clear card-creation error (that resulted in new file in the realm) when different card-def is selected', async function (assert) {
```
habdelra, line 1268: ditto

PR #2528, opened 2025-04-30
host: Hide Monaco diff sidebar in Percy

backspace:
This removes spurious diffs such as [this](https://percy.io/Cardstack-1/web/-cardstack-host/builds/40152188/changed/2164207198?browser=firefox&browser_ids=64%2C65&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=1280&widths=1280):

Build #12487 - Cardstack - Percy | Visual testing as a service 2025-04-30 12-53-41

This is the [second thing I’m hiding on this snapshot](https://github.com/cardstack/boxel/pull/2503) which makes me wonder whether we need it at all! Is this a custom Monaco view that we need to check for visual regressions on?

PR #2535, opened 2025-05-01
host: Change copy from seed to default to false

backspace:
I’m submitting this for review assuming I understand issue 8543 correctly 😆

The Percy diff is the same silenced by #2528, which is awaiting feedback.


tintinthong (approved):
Yep you got it spot on! Thanks!

burieberry (approved):
Tested this locally and looks right to me

PR #2539, opened 2025-05-02
Add gutter to RHS

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8558/add-gutter-to-rhs

Task:
image

Result:

https://github.com/user-attachments/assets/6e35ba25-1be1-4737-ad03-82714ef43b45

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
@@ -2,6 +2,7 @@ import { fn } from '@ember/helper';
 import { action } from '@ember/object';
 
 import { service } from '@ember/service';
+import { htmlSafe, type SafeString } from '@ember/template';
 
 import Component from '@glimmer/component';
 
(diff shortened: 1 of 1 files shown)


packages/host/app/components/operator-mode/code-submode/playground/playground-content.gts
```diff
@@ -129,6 +129,8 @@ export default class PlaygroundContent extends Component<Signature> {
         flex-direction: column;
         gap: var(--boxel-sp);
         min-height: 100%;
+        max-width: 800px;
```
burieberry: For stack-item, we're using `50rem`, so you can use the same here. Something to account for is the cards that have `prefersWideFormat` static property set to `true`, such as the app-card. For those, isolated card preview should still extend to 100% width, similar to how it is in interact mode.

lucaslyl: sure @burieberry 

I implemented the **styleForPlaygroundContent** getter in the PlaygroundContent component that dynamically sets the maximum width of the playground content area.:

1. When the format is not '**isolated**' or the card has the prefersWideFormat property, it sets the maximum width to '**100%**' to utilize the full available space
2. Otherwise, it constrains the width to '**50rem'** for a more controlled layout when displaying isolated content

burieberry (approved):

PR #2540, opened 2025-05-02
improve store api to accept localDir argument

tintinthong:


packages/runtime-common/index.ts
```diff
@@ -301,20 +301,25 @@ export type getCards<T extends CardDef = CardDef> = (
...
 
+export interface CreateOptions {
+  realm?: string; //this is meant to be a directory
+  localDir?: string; //this is meant to be a LocalPath authorDir/author
+  relativeTo?: URL;
+}
+
+export interface AddOptions extends CreateOptions {
+  doNotPersist?: boolean;
+}
```
habdelra: Looks good thanks!

packages/runtime-common/index.ts
```diff
@@ -301,20 +301,25 @@ export type getCards<T extends CardDef = CardDef> = (
   isLoading: boolean;
 };
 
+export interface CreateOptions {
+  realm?: string; //this is meant to be a directory
+  localDir?: string; //this is meant to be a LocalPath authorDir/author
```
habdelra: let' use the `LocalPath` type here to make it clear this is what it is

habdelra (approved):

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -515,13 +505,19 @@ export default class InteractSubmode extends Component {
   });
 
   private _copyCards = dropTask(
-    async (cards: CopyCardsWithCodeRef[], targetUrl: string) => {
+    async (
+      cards: CopyCardsWithCodeRef[],
+      realm: string,
+      localDir?: string, //
```
richardhjtan: // might be not needed?

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -515,13 +505,19 @@ export default class InteractSubmode extends Component {
...
+    ) => {
       let { commandContext } = this.commandService;
+      // let targetDir = combine realmUrl and localDir
```
richardhjtan: This variable might be not using as well

tintinthong:
During testing, @richardhjtan has found a bug that the stack item card doesn't persist in this PR (only in this PR)

packages/host/app/components/operator-mode/interact-submode.gts
habdelra: yes, please cleanup commented out code

habdelra:
> During testing, @richardhjtan has found a bug that the stack item card doesn't persist in this PR (only in this PR)

what do you mean by "doesn't persist"? like the stack item is not serialized into the URL? or that auto save is not working for the stack item?

tintinthong:
> During testing, @richardhjtan has found a bug that the stack item card doesn't persist in this PR (only in this PR)

@richardhjtan I can't re-create your behaviour. But, your behaviour in which the stack item persist but the loading state goes away; I think wud occur if the chessboard esm doesn't have issues with loading from the loader cache. Wud be interested to see your console output.

PR #2542, opened 2025-05-02
Bug fix for compound field subscriptions not notifying on enclosing instance

habdelra:
This fixes an issue where when a change is made in an instance that has a compound field, we were not sending a notification on the enclosing instance. This manifested itself as the inability to auto-save when a property within a compound field changes (e.g. the command field of a Skill card) as the notification event sent didn't include the enclosing instance.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -201,6 +201,10 @@ const subscribers = initSharedState(
   'subscribers',
   () => new WeakMap<BaseDef, Set<CardChangeSubscriber>>(),
 );
+const subscriberConsumer = initSharedState(
+  'subscriberConsumer',
+  () => new WeakMap<BaseDef, { fieldOrCard: BaseDef; fieldName: string }>(),
+);
 const fieldDescriptions = initSharedState(
   'fieldDescriptions',
   () => new WeakMap<typeof BaseDef, Map<string, string>>(),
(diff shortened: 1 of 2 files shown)


packages/host/tests/integration/components/card-basics-test.gts
```diff
@@ -1111,75 +1111,65 @@ module('Integration | card-basics', function (hooks) {
...
+            {
+              instance: firstWorkExperience,
+              fieldName: 'company',
+              value: 'First Company',
+            },
+            {
+              instance: mango,
+              fieldName: 'workExperiences.0.company',
+              value: 'First Company',
+            },
```
habdelra, line 1155: this is the key event that was missing: when the interior of a compound field changes we need to notify on the enclosing card instance.

backspace (approved):

tintinthong (commented):
Is there a need for cleanup of `subscriberConsumer` whenever `unsubscribeFromChanges` is called?

packages/base/card-api.gts
```diff
@@ -2035,10 +2039,20 @@ export type FieldDefConstructor = typeof FieldDef;
 export function subscribeToChanges(
   fieldOrCard: BaseDef | BaseDef[],
   subscriber: CardChangeSubscriber,
+  enclosing?: { fieldOrCard: BaseDef; fieldName: string },
```
tintinthong, line 2042: Is it reasonable to say that the subscriberConsumer values only applies to contain compound fields. Or is there an instance that enclosing is actually a "card"?

tintinthong:
Screenshot 2025-05-04 at 14 13 14

I have tested and the auto-save works based upon network request. JSON in the monaco editor doesn't react tho. I thought there was a ticket about this

jurgenwerk (approved):

habdelra:
> Is there a need for cleanup of `subscriberConsumer` whenever `unsubscribeFromChanges` is called?

you could potentially remove the entry if there are no subscribers anymore--but given how often the loader is refreshed it seems unnecessary.

packages/base/card-api.gts
habdelra: it's nested. so at the outer most layer its always a card, and then you have an entry for each layer of compound field that you go thru

habdelra:
> Screenshot 2025-05-04 at 14 13 14
> I have tested and the auto-save works based upon network request. JSON in the monaco editor doesn't react tho. I thought there was a ticket about this

yeah, it's the prefixed clientRequestId's. we are intentionally ignoreing the prefix that comes from the preview panel on the right hand side. i posted a question in discord about why why are doing this

PR #2544, opened 2025-05-02
host: Fix bot patching when search block empty

backspace:
Asking the bot to add to an empty file, previously:

screencast 2025-05-02 16-27-39

After:

screencast 2025-05-02 16-22-19

PR #2546, opened 2025-05-02
host: Extract HtmlGroupCodeBlock

backspace:
I found this useful when working on #2533, as having an explicit component with inspectable properties and arguments makes for easier debugging. I also wished for it when working on #2544.


jurgenwerk (approved):
Thanks, was definitely ripe for extraction

PR #2549, opened 2025-05-05
Implement files caching

FadhlanR:
The idea is to introduce a new attribute in `FileDef` called `contentHash`. This hash will be calculated when uploading files, cards, or command definitions. Since this data will be included in the Matrix event, we can build a cache when processing timeline event and use that cache during uploads.


packages/host/app/commands/add-skills-to-room.ts
```diff
@@ -53,7 +53,8 @@ export default class AddSkillsToRoomCommand extends HostBaseCommand<
             (newCommandDefinition) =>
               !commandDefinitions.some(
                 (commandDefinition) =>
-                  commandDefinition.name === newCommandDefinition.name,
+                  commandDefinition.sourceUrl ===
+                  newCommandDefinition.sourceUrl,
```
FadhlanR, line 57: I set the sourceUrl of the command definition using codeRef, as it seems more reliable than using name, which can be duplicated.

packages/host/tests/integration/components/ai-assistant-panel/general-test.gts
```diff
@@ -861,4 +871,189 @@ module('Integration | ai-assistant-panel | general', function (hooks) {
       .dom('[data-test-message-idx="1"] [data-test-ai-message-content]')
       .containsText('I sent a message from the background.');
   });
+
+  test('ensures cards are reuploaded when only content changes', async function (assert) {
```
jurgenwerk:
```suggestion
  test('ensures cards are reuploaded only when content changes', async function (assert) {
```

jurgenwerk:
I'll test this locally now

packages/host/app/lib/file-def-manager.ts
```diff
@@ -432,4 +506,49 @@ export default class FileDefManagerImpl implements FileDefManager {
...
+            this.contentHashCache.set(card.contentHash, card.url);
+          }
+        }
+      }
+    }
+
+    // Handle skills config
+    if (eventType === APP_BOXEL_ROOM_SKILLS_EVENT_TYPE) {
+      const skillsContent = content as SkillsConfigEvent['content'];
+      const skills = [
```
lukemelia:
```suggestion
      const skillsAndCommands = [
```

packages/host/app/lib/file-def-manager.ts
```diff
@@ -432,4 +506,49 @@ export default class FileDefManagerImpl implements FileDefManager {
...
+    const content = event.getContent();
+    if (!content) {
+      return;
+    }
+
+    const eventType = event.getType();
+
+    // Handle attached files and cards
+    if (eventType === 'm.room.message' && 'data' in content) {
+      const data = JSON.parse(content.data) as CardMessageContent['data'];
```
lukemelia: We are going to be doing a lot of JSON parsing here since we will process every aibot streaming (replace) message. Not sure how best to optimize.

lukemelia (approved):

packages/host/app/lib/file-def-manager.ts
FadhlanR: Updated here [51b7aed](https://github.com/cardstack/boxel/pull/2549/commits/51b7aed90344f3471cf9d0231c4939cada39b28b)

jurgenwerk (approved):

PR #2552, opened 2025-05-06
Fix: Prevent localStorage leak

FadhlanR:
When logging in with account B after previously logging in with account A, the new session incorrectly attempts to fetch account A’s realms. This happens because we didn’t properly clean up local storage during logout.

### DIFF
--- a/packages/host/app/utils/local-storage-keys.ts
+++ b/packages/host/app/utils/local-storage-keys.ts
@@ -22,8 +22,4 @@ export function clearLocalStorage() {
   window.localStorage.removeItem(RecentCards);
   window.localStorage.removeItem(RecentFiles);
   window.localStorage.removeItem(ScrollPositions);
-  // Remove all codeblock_ prefixed items
-  Object.keys(window.localStorage)
-    .filter((key) => key.startsWith('codeblock_'))
-    .forEach((key) => window.localStorage.removeItem(key));
 }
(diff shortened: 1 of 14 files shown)


lukemelia (approved):
Seems good. Should we consider colocating this more. e.g. having a "clearLocalStorage" function in local-storage-keys.ts that does this? This might mitigate the risk of a developer adding a key and not realizing they need to add the removal in matrix-service.

richardhjtan (approved):

habdelra (approved):

packages/host/app/utils/local-storage-keys.ts
```diff
@@ -6,3 +8,22 @@ export const CodeModePanelSelections = 'code-mode-panel-selections';
...
+  window.localStorage.removeItem(CodeModePanelWidths);
+  window.localStorage.removeItem(CodeModePanelHeights);
+  window.localStorage.removeItem(CodeModePanelSelections);
+  window.localStorage.removeItem(PlaygroundSelections);
+  window.localStorage.removeItem(RecentCards);
+  window.localStorage.removeItem(RecentFiles);
+  window.localStorage.removeItem(ScrollPositions);
+  // Remove all codeblock_ prefixed items
+  Object.keys(window.localStorage)
+    .filter((key) => key.startsWith('codeblock_'))
```
jurgenwerk: where are we setting `codeblock_` items? I think we aren't doing that anymore.

FadhlanR: Ah I didn't know that, I'll remove it. Thanks!

PR #2555, opened 2025-05-06
Fix for bug where we track module dependencies for modules that happen to be in an error state in the index

habdelra:
This fixes an issue where we were not tracking module dependencies correctly when one of the modules we depended on happened to be in an error state in the index. (originally https://github.com/cardstack/boxel/pull/2500)


packages/runtime-common/error.ts
```diff
@@ -225,7 +225,12 @@ export class CardError extends Error implements SerializedError {
           responseText: text,
         },
       );
-      if (response.status === 404) {
+      // 406 errors matter to us because in the scenario were you are depending on a
+      // module that is depending on another module, like a -> b -> c. if module c
+      // doesn't exist then module b is in an error state. when module a asks for
+      // module b, we return an error code, 406, meaning that we know about module b,
+      // but we can't give it to you because it is in an error state.
+      if ([404, 406].includes(response.status)) {
```
habdelra, line 233: The explanation of the issue and solution

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1118,6 +1118,200 @@ module(basename(__filename), function () {
...
+          export class BlogApp extends CardDef {
+            @field title = contains(StringField);
+          }
+        `,
+      );
+
+      let blogPostModule = await realm.realmIndexQueryEngine.module(
+        new URL(`${testRealm}blog-post`),
+      );
+      // This is failing because the currently module is not being resolved
```
richardhjtan: Probably not needed here as now it's resolved

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1118,6 +1118,200 @@ module(basename(__filename), function () {
...
```
richardhjtan: Same as above

richardhjtan (approved):
I installed the blog app listing via this branch, even installed the CRM app with complex circular dependencies files - every file can resolve the module without error and working great now ✨ Thanks for the fix @habdelra

PR #2560, opened 2025-05-07
fix spec preview auto-save bug

tintinthong:


packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -477,7 +512,23 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
...
+  private search = task(async () => {
+    this.cards = flatMap(
+      await Promise.all(
+        this.realms.map(
+          async (realm) =>
+            await this.store.search(this.specQuery, new URL(realm)),
+        ),
+      ),
+    ) as Spec[];
+  });
```
tintinthong, line 531: Ahh I can see how this search will not re-run from choosing a different code ref in lhs selector. Hence, why we may want a resource. Perhaps, there is an intent to maintain `getCards` (a resource). 

But, I did this change bcos I just wanted to use the store more directly; no reason why a resource wrapping the store.search wudn work also

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -224,6 +226,51 @@ type SpecPreviewCardContext = Omit<
...
+
+  private get displayIsolated() {
+    return !this.args.canWrite && this.args.id;
+  }
+
+  get spec() {
+    return this.specResource?.card as Spec;
+  }
+  <template>
+    {{consumeContext this.makeSpecResource}}
```
habdelra, line 254: this is good. correct creation for card resource :+1:

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
habdelra: so yeah, i think `getCards()` is what you want here right? the search query is a thunk that when it is changed the search re-runs. `Store.search()` is just a straight-up one-shot search.

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
```diff
@@ -477,7 +512,23 @@ export default class SpecPreview extends GlimmerComponent<Signature> {
   @service private declare recentFilesService: RecentFilesService;
   @service private declare specPanelService: SpecPanelService;
   @service private declare store: StoreService;
-  @tracked private search: ReturnType<getCards<Spec>> | undefined;
+  @tracked cards: Spec[] = [];
+
+  constructor(owner: Owner, args: any) {
+    super(owner, args);
+    this.search.perform();
```
habdelra, line 519: the fact that you are running this search in the constructor is a code-smell that you really want to be using a resource. in this case `getCards()`

packages/host/app/components/operator-mode/code-submode/spec-preview.gts
habdelra: the way you are consuming the search results, though is going to cause flickering, as you are just passing an array of cards to the child component. the array is probably different each time the search is run--there is probably no object stability. we solved in this other places by actually passing the _whole resource_ to the child component (in the ai assistant). the search resource is actually stable between different searches and will not trigger a re-render because of object equality

habdelra: in general, when you see an array being specified in a component signature, alarm bells should go off. arrays are notoriously unstable and can cause flickering if we are replacing arrays instead of mutating them.

PR #2561, opened 2025-05-07
Reestablish referenced cards in store after loader reset

habdelra:
This PR reestablishes referenced card instances in the store after a loader reset. We are correctly expunging instances after a loader reset since the implementation for the card instances is out of date. however, we were not reestablishing the references for the expunged cards. This PR fixes that.


tintinthong:
when we invalidate, it looks like fields in other cards will complain about `.validate`. How do we handle something like that at the field level?

Screenshot 2025-05-08 at 01 59 02

habdelra:
> when we invalidate, it looks like fields in other cards will complain about `.validate`. How do we handle something like that at the field level?
> 
> Screenshot 2025-05-08 at 01 59 02

this seems unrelated to this PR. are you on the latest in main? (I refactored the field validation logic very recently). this looks like a server issue during indexing, not a browser issue. the store does not live in the server, it is downstream from the indexer...

tintinthong (approved):
This looks good to me. It recovers the instances that are on the page. But, occasionally for other reasons the loader cache is misbehaving when interacting with the cdn

thx for doing this btw

habdelra:
> This looks good to me. It recovers the instances that are on the page. But, occasionally for other reasons the loader cache is misbehaving when interacting with the cdn
> 
> thx for doing this btw

please file an issue if you are able to nail down the problem

PR #2562, opened 2025-05-07
Start listening for realm events synchronously when adding reference to the store

habdelra:
This addresses a concern that @tintinthong has around when we start listening to a realm when adding a  reference to the store. This way there is no async between wiring up a reference and listening for realm events


tintinthong (approved):
This is fantastic! Now, the behaviour is very consistent. Rather than having loader resets only sometimes but not all the time. Now, I can see it happening ALL the time.

PR #2564, opened 2025-05-07
Auto-attach newly created card in AI assistant

habdelra:
This fixes an issue where a newly created card is not not auto-attached to the AI assistant. I opted to immediately kick off a save in the background after a card is created instead of just sending an unsaved card to the ai bot. The rationale is that trying to synchronize state between the matrix room with an unsaved card and the host seems very fraught with issues--specifically what happens if you reload the page? the matrix room still knows there was a card, but teh realm server and host no longer know about the card. reconciling the differing lifetimes of an unsaved card between matrix, the host, and the realm server seems very painful. so instead, while the initial saving is happening, a busy indicator is displayed in place of the card pill in the chat window. additionally i resolve any local ID's to remote ID's before the message is sent.

https://github.com/user-attachments/assets/431f9098-8466-4f6a-a4bb-4edb5927ca47


backspace (approved):
> reconciling the differing lifetimes of an unsaved card between matrix, the host, and the realm server seems very painful

agreed!

PR #2565, opened 2025-05-08
Autogenerate card instance for playground if there isn't one available to preview

burieberry:
Changes:
(Viewing a card-def in playground)
- If a card instance is not found in RecentFiles, but one exists in one of user's realms, display it on playground preview
(previously we only displayed if card was in RecentFiles)
- If there's no card instance of the card def in any of user's realms, autogenerate new card instance in current realm and preview it on playground
- Exception when previewing `CardDef` in `base/card-api`: this would result is a search of all descendants of type card-def, so skip expanded search here
- Sort recent cards in dropdown by `lastModified` date, instead of `created` date


jurgenwerk (approved):
Looking forward to using this!

habdelra:
does this also handle the scenario where we also check recent cards (as opposed to recent files) for a matching instance?

burieberry:
> does this also handle the scenario where we also check recent cards (as opposed to recent files) for a matching instance?

No, will handle that in my next PR

lukemelia (approved):

PR #2566, opened 2025-05-08
Always send updated skills and commands to AI

FadhlanR:
This PR addresses an issue where the AI bot was not receiving the latest skill updates. The idea is to ensure that skill cards are uploaded along with any attached cards every time a message is sent. FileDefManager uses a caching mechanism to determine whether skills and commands need to be re-uploaded and to update the room state accordingly.


packages/host/app/services/matrix-service.ts
```diff
@@ -755,8 +756,15 @@ export default class MatrixService extends Service {
       }
     }
 
+    // Upload skill cards and attached cards together to ensure:
+    // 1. Latest skill cards are deployed with the message
+    // 2. FileDefManager's cache prevents re-uploading unchanged skill cards
+    let skillCards = this.roomResources.get(roomId)?.skillCards ?? [];
+    let attachedCardsAndSkills = [...attachedCards, ...skillCards].filter(
+      (card, index, self) => index === self.findIndex((c) => c.id === card.id),
+    );
```
lukemelia: I assume this filter logic is supposed to enforce uniqueness. Let's make a small util named accordingly so that intent is clear.

packages/host/app/commands/send-ai-assistant-message.ts
```diff
@@ -106,10 +106,19 @@ export default class SendAiAssistantMessageCommand extends HostBaseCommand<
     if (files?.length) {
       files = await this.matrixService.uploadFiles(files);
     }
-    let cardFileDefs = await matrixService.uploadCardsAndUpdateSkillCommands(
-      input.attachedCards ?? [],
-      roomId,
+    // Upload skill cards and attached cards together to ensure:
+    // 1. Latest skill cards are deployed with the message
```
lukemelia: It is awkward that we are relying on the caller of `uploadCardsAndUpdateSkillCommands` to include the room's skill cards? Why not make that an internal implementation detail of the method?

lukemelia (changes requested):

jurgenwerk (approved):
In my https://github.com/cardstack/boxel/pull/2569 I am experiencing a bug where if I update a default skill, the bot keeps getting the old version. Even if I make a new session, enable/disable skill...nothing seems to work. I tried merging this branch into mine and it looks like the bug is fixed.

jurgenwerk:
Actually I take that back, I think I am still experiencing the caching bug even using this branch. It seems I need to sometimes still enable/disable skills in order for the bot to get the updated version. Will see if I can make repro steps

packages/host/app/commands/send-ai-assistant-message.ts
FadhlanR: I think it is better to remove the `uploadCardsAndUpdateSkillCommands` from file def manager and create a new function in matrix-service called `updateSkillsAndCommandsIfNeeded`. Please check it again.

packages/host/app/commands/send-ai-assistant-message.ts
```diff
@@ -106,9 +106,9 @@ export default class SendAiAssistantMessageCommand extends HostBaseCommand<
     if (files?.length) {
       files = await this.matrixService.uploadFiles(files);
     }
-    let cardFileDefs = await matrixService.uploadCardsAndUpdateSkillCommands(
+    await this.matrixService.updateSkillsAndCommandsIfNeeded(roomId);
```
lukemelia:
```suggestion
    await matrixService.updateSkillsAndCommandsIfNeeded(roomId);
```

lukemelia (approved):

PR #2569, opened 2025-05-09
AI assistant: Support creating new files

jurgenwerk:
### New features

1. Adds realm url to LLM context
2. If LLM recognizes you want to create a new file, it will respond with a search/replace block (with empty search portion) where the url of the file will be realm url + suggested file name
3. If file with that name already exists, it will create a new file

image

### Improvements

Add an error display of when the LLM responds with a search/replace where file url is missing (this can happen with some of the "mini" models but I don't have a reliable way of reproducing). Previously we blew up by throwing an unhandled error but now we have a new error display in this case:

image

### Missing related features / follow up work

Once the file(s) gets created, the user does not have a direct way of opening them. Currently, they have to do it by going to the file explorer and click on them. Follow up work will include a new feature where the assistant will offer to list the newly created files and offer to open them (we have a separate ticket for that)


packages/host/app/lib/formatted-message/utils.ts
```diff
@@ -76,13 +80,28 @@ export function extractCodeData(preElementString: string): CodeData {
...
+  let _isCompleteSearchReplaceBlock = isCompleteSearchReplaceBlock(
+    contentWithoutFileUrl,
+  );
+
+  // If search/replace block looks like this then we know this should result in a new file
+  // <<<<<<< SEARCH
+  // =======
+  // code ...
+  // >>>>>>> REPLACE
+  let isNewFile =
```
backspace: This means the bot can no longer add to a blank file:

test-h1v txt in Experiments Workspace 2025-05-09 07-33-01

backspace: When I added a test for that I wondered whether `host` was really the right place, or maybe my assertions just didn’t cover this case.

jurgenwerk: Ah nice catch, perhaps we should change this so that the LLM sends a special annotation that it wants to create a new file. So that we don't rely on detecting an empty search block (which can be the case of a blank file too as you pointed out)

jurgenwerk:
Moving back to draft until https://github.com/cardstack/boxel/pull/2569#discussion_r2081579207 is fixed

packages/host/app/lib/formatted-message/utils.ts
jurgenwerk: This should be fixed with the new updates (I tested it manually and it worked)

jurgenwerk:
For people trying to test this, you will unfortunately have a bad time because currently the skill cards are cached too aggresively and it's really hard to get them to update (to achieve that there's an involved process of switching between interact/code modes and reattaching the skill and creating a new session). Here are the PRs that need to be merged to get this fixed: https://github.com/cardstack/boxel/pull/2566 https://github.com/cardstack/boxel/pull/2577

lukemelia:
I have doubts about the __META idea, specifically about inventing our own format rather than trying to piggyback on formats the models are being trained on.

Looking at https://aider.chat/docs/more/edit-formats.html, should we consider instructing the model to use the "whole" format when it wants to create a file?

jurgenwerk:
@lukemelia I didn't know about the "whole" format. This is because in our existing source code editing skill (in main) we instruct the AI to use a search/replace block for new files:

```
If you want to put code in a new file, use a *SEARCH/REPLACE block* with:
- A new file path, including dir name if needed
- An empty `SEARCH` section
- The new file's contents in the `REPLACE` section.
```

What I'm worried about with the [whole](https://aider.chat/docs/more/edit-formats.html#whole) format approach is that we (the host app) has no way to recognize this is a code patch and it will render is as a normal code snippet.

IanCal:
What about adding create file to the commands on a the source code editing skill card?

IanCal:
I'm also not too clear on the problem we're trying to solve here, is the issue that it might try and create a new file and fully write the contents when that file already exists but the llm doesn't know, and then things are overwritten?

jurgenwerk:
@IanCal the main issue is that we need to support both creating a new file, and writing to a blank file (that already exists). In the current instructions, we tell it to use an empty search portion of the search/replace block in both cases. So we need a way to differentiate between the two flows. That's why I added a meta field that tells us whether this is a new file or not. Perhaps one way of simplifying this would be to see if the line above the search/replace block is a url (existing file), or just a file name (new file)

IanCal:
> @IanCal the main issue is that we need to support both creating a new file, and writing to a blank file (that already exists). In the current instructions, we tell it to use an empty search portion of the search/replace block in both cases. So we need a way to differentiate between the two flows. That's why I added a meta field that tells us whether this is a new file or not. Perhaps one way of simplifying this would be to see if the line above the search/replace block is a url (existing file), or just a file name (new file)

Can we tell if a file already exists, and create it if doesn't exist?

Alternatively, I'd be interested in seeing if a "create file" command. We have a write-text-file host command, and could have a thin wrapper for creating a new blank file and returns the url if successful.

jurgenwerk:
> Can we tell if a file already exists, and create it if doesn't exist?

I made this work in a way where we instruct the AI use file urls for existing files, and file names for new files. It will result in search/replace blocks like this:

Editing existing files:
```
http://localhost:4201/jurgen2/personal/simple-postcard.gts
<<<<<<< SEARCH
...code to replace...
=======
...code to replace it with...
>>>>>>> REPLACE
```

Creating new files:
```
wedding-invitation.gts
<<<<<<< SEARCH
=======
... code ...
>>>>>>> REPLACE
```

I'm using file urls because otherwise we would also need to pass in realm url to the AI to make it work and I'm worried that will bring additional problems. But in the 2nd case, file name is enough because we can figure out the realm url in the `PatchCodeCommand`. 

Regarding the "create file" command, I don't understand how it would help solve this problem, did you mean using it as a tool?

lukemelia:
Let's use the following logic:

```
Editing existing files:

http://localhost:4201/jurgen2/personal/simple-postcard.gts
<<<<<<< SEARCH
...code to replace...
=======
...code to replace it with...
>>>>>>> REPLACE

Creating new files:

http://localhost:4201/jurgen2/personal/wedding-invitation.gts
<<<<<<< SEARCH
=======
... code ...
>>>>>>> REPLACE
```

In the creating new file case (blank search block), use the following test to determine whether to create a new file:

1) If the URL does not exist, create a new file
2) If the URL exists and the file is empty, fill in the contents of the file
3) If the URL exists and the file is NOT empty, create a new file with a similar name (numeric algo already present in this PR)

lukemelia (commented):
Looking good. Left some comments with an adjusted plan while we chatted.

jurgenwerk:
Notes: 
- provide realm url as currently viewing context so that AI can form a url using a suggested filename (when it's trying to create new files) 
- with all that can probably get rid of the fileUrl/fileName/isNewFile triality and keep fileUrl only

jurgenwerk:
This is ready for another review. I also updated the PR description.

lukemelia (approved):

PR #2570, opened 2025-05-09
More resilient index worker

habdelra:
This PR adds better resiliency for index workers based on recent analysis of index issues causing unresponsive workers. This PR adds:
- stubbed out `setTimeout()` and `setInterval()` globals in the FastBoot context. code that was running in these timers was escaping the rendering context and causing the worker to crash and respawn.
- Fixed a bug in our loader when source code was unable to transpile. in our example it was due to very badly implemented template only components that would transpile to having multiple default exports. This would leave a hanging deferred promise that never got resolved causing the loader to pause indefinitely and hence making the worker freeze.
- added rendering timeouts such that we never spend more than 30 seconds trying to render HTML for slow renders that are interruptable. 
- added a worker manager watch dog. busy-wait bugs (synchronous infinite loops) will cause the browser to hang and cannot be interrupted from within the browser (which in this case is the entire worker since the indexing job is running wholly in fastboot). the worker manager checks the status of each worker every minute to see if it has work that has not finished by the lock expiration time (which is indicative of a stuck job). if the worker does appear to be stuck, the worker manager will mark the jobs as having resulted in an error and will recycle the worker. I tested this locally with the "sudoku" module that has a busy wait in it. it was able to discover the stuck job within 2 minutes and eliminate it.
- changed the job timeout for incremental indexing to 1 minute, and the job timeout for full indexing to 5 minutes. So in the event of a stuck job, generally the worker manager will be able to set everything right after 2 sweeps, e.g. 2 minutes.
- Added test cards that include un-transpilable code and timeouts to the node test realm. these would otherwise break the node test realm without the work inclkuded here


packages/postgres/pg-queue.ts
```diff
@@ -234,7 +234,7 @@ export class PgQueueRunner implements QueueRunner {
   constructor({
     adapter,
     workerId,
-    maxTimeoutSec = 10 * 60,
+    maxTimeoutSec = 5 * 60,
```
habdelra, line 237: i made this match the longest timeout that we use so we can be consistent. This is the value that the job will use internally to try to timeout (which it can't always do, e.g. busy waits that can only be interrupted externally)

packages/runtime-common/realm-index-updater.ts
```diff
@@ -145,7 +145,7 @@ export class RealmIndexUpdater {
       let job = await this.#queue.publish<IncrementalResult>({
         jobType: `incremental-index`,
         concurrencyGroup: `indexing:${this.#realm.url}`,
-        timeout: 4 * 60,
+        timeout: 60,
```
habdelra, line 148: generally incremental indexing should take less than a second, so a 1 minute timeout seems more than enough. this means we should not have to wait up to 10 minutes for things to resolve anymore

backspace (approved):
A great collection of improvements 🎉

lukemelia (approved):

PR #2571, opened 2025-05-12
custom isolated template not triggering not loaded error

tintinthong:
I dont know why isolated renderer is not triggering any not loaded errors here

Its causing sprint-tracker cards to fail to render and also not resolve its field missing link field error. It doesn't even go into trying loading again (which it shud)

https://linear.app/cardstack/issue/CS-8386/investigate-team-computed-todo-app

### DIFF
--- a/packages/realm-server/tests/indexing-test.ts
+++ b/packages/realm-server/tests/indexing-test.ts
@@ -1,4 +1,4 @@
-import { module, test } from 'qunit';
+import { module, test, only } from 'qunit';
 import { dirSync } from 'tmp';
 import {
   baseRealm,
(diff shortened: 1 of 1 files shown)


packages/realm-server/tests/indexing-test.ts
```diff
@@ -1024,6 +1004,167 @@ module(basename(__filename), function () {
...
+          }
+
+        )
+      );
+      let taskInstance = (await realm.realmIndexQueryEngine.instance(
+        new URL(`${testRealm}task`),
+      )) as any;
+      let taskInstanceWithoutCustomIsolatedTemplate = (await realm.realmIndexQueryEngine.instance(
+        new URL(`${testRealm}task-without-custom-isolated-template`),
+      )) as any;
```
tintinthong, line 1158: `taskInstance` is the failed case
`taskInstanceWithoutCustomIsolatedTemplate` is the success case

PR #2572, opened 2025-05-12
within a computed field, we should load missing field that have `isUsed=true`

tintinthong:
**Getting rid of this** 

Screenshot 2025-05-12 at 20 37 16

I have confirmed this fixes this

**Reason this happens** 

If a computed field depends of a link, even if the link uses `isUsed`, it is not trying to load the missing field bcos we handle computeds differently from other fields and may have missed it

**Coding pattern that is causing this (`isUsed` in team is not working for computeds)**

Screenshot 2025-05-12 at 17 27 12

### DIFF
--- a/packages/realm-server/tests/indexing-test.ts
+++ b/packages/realm-server/tests/indexing-test.ts
@@ -787,7 +787,7 @@ module(basename(__filename), function () {
         { ...realm.realmIndexUpdater.stats },
         {
           instancesIndexed: 1,
-          instanceErrors: 1,
+          instanceErrors: 0,
           moduleErrors: 0,
           modulesIndexed: 1,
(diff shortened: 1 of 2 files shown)


habdelra (approved):

packages/realm-server/tests/indexing-test.ts
```diff
@@ -277,26 +277,6 @@ module(basename(__filename), function () {
                 },
               },
             },
-            'bad-link.json': {
```
tintinthong: why was this removed

tintinthong: fixing now

lukemelia (approved):

PR #2563, opened 2025-05-07
host: Change code mode accordion to buttons

backspace:
This is a major overhaul of the module inspector panel which goes from an accordion:

ai-app-generator gts in Experiments Workspace 2025-05-30 10-44-16

to buttons, try it [here](https://hostcode-accordion-to-buttons-cs-8589.boxel-host-preview.stack.cards/):

ai-app-generator gts in Experiments Workspace 2025-05-30 10-45-53

You can see there that the format chooser looks different, and that the instance chooser shows alongside it. The spec and schema panels are pretty much unchanged other than container tweaks. Things that were previously within accordion headers are now within the view buttons, sometimes reduced.

The format chooser design changes also show in the card preview renderer:

0cd60785-af5d-4a9d-a56a-399ebbb720c6 json in Experiments Workspace 2025-05-30 10-58-52

Because the accordion children render even when they’re not open, some code with side-effects needed to move into the module inspector component.

There are a few issues mentioned in comments for followup work.

PR #2575, opened 2025-05-12
host: Extract code mode ModuleInspector

backspace:
This is a side quest to better prepare the way for #2563. It encapsulates the contents of the right-hand sidebar in code mode: schema editor, playground (soon to be preview), spec, and card renderer (renamed from `Preview`).

Next up: pulling up component functionality to remove sibling dependencies.


tintinthong (approved):
👍 This naming makes sense to me

jurgenwerk (approved):

packages/host/app/components/operator-mode/code-submode/rhs-panel.gts
```diff
@@ -0,0 +1,402 @@
...
+    isReadOnly: boolean;
+    moduleContentsResource: ModuleContentsResource;
+    previewFormat: Format;
+    readyFile: Ready;
+    selectedDeclaration: ModuleDeclaration | undefined;
+    setPreviewFormat: (format: Format) => void;
+  };
+}
+
+export default class RhsPanel extends Component<RhsPanelSignature> {
```
lukemelia: Is the rename going to come later, or is this the proposed name?

backspace: Maybe you opened an old notification? It’s `ModuleInspector` now 🤓

PR #2577, opened 2025-05-13
Avoid overwriting card instances with card docs from the matrix store

FadhlanR:
**Problem**

We instantiated the skill cards using the card documents that are uploaded to the Matrix media store. However, these card documents might be outdated compared to the ones in the Boxel index. We used `this.store.add` to instantiate the skill cards, but this function doesn't retrieve the card from the Boxel index—it simply creates a new card instance and stores it in the `identityContext`. As a result, we could potentially overwrite an existing skill card instance with an older version. Consequently, every time we create a new room, we might end up using this outdated version of the skill cards.

**Solution**

Instead of using `this.store.add` directly, we should first try to retrieve the card using `this.store.get`. If no card instance exists, we can fall back to using `this.store.add`.

There are also updates I made to address two issues with the `CommandField`.

* The `CodeRef` of `CommandField` must be an absolute URL. This bug wasn’t previously exposed because we always used card documents from the Matrix media store, where we set `useAbsoluteUrl` to `true` during serialization. To address this, I introduced a new field called `AbsoluteCodeRef`, but I’m open to other approaches.

* There is an inconsistency in the default value of `BooleanField`. When the field is first created, its value is `null`, so the radio buttons for `true` and `false` are both unchecked. However, after reloading the page, the `false` option becomes selected. This happens because the initial value comes from `field.emptyValue`, whereas after reloading, it comes from `field.deserialized`. To address this, I added a static getter called `[emptyValue]` to `BaseDef`, which can be overridden by specific `FieldDef` implementations. For `BooleanField`, this getter will return `false`. I'd appreciate input from @habdelra on this.


habdelra:
I think the empty value approach is a good one. we had the beginnings of this already in `Field.emptyValue`. this seems like it finishes the thought.

packages/host/app/resources/room.ts
```diff
@@ -355,9 +358,13 @@ export class RoomResource extends Resource<Args> {
...
-    let skillCard = await this.store.add<Skill>(cardDoc, {
-      doNotPersist: true,
-    });
+    let skillCard: Skill;
+    let maybeSkillCard = await this.store.get(cardId);
+    if (isCardInstance(maybeSkillCard)) {
+      skillCard = maybeSkillCard as Skill;
+    } else {
+      skillCard = await this.store.add<Skill>(cardDoc);
+    }
```
habdelra, line 376: Looks good. ultimately we have this ticket https://linear.app/cardstack/issue/CS-8367/clarify-how-command-result-cards-that-live-in-matrix-interact-with which will update the store to be able to load instances from matrix state instead of the store. perhaps this is what we'll want in the future?

packages/base/boolean.gts
```diff
@@ -41,6 +42,10 @@ export default class BooleanField extends FieldDef {
     return Boolean(val);
   }
 
+  static get [emptyValue]() {
+    return false;
+  }
+
```
habdelra, line 48: let's make sure to add a test for `[emptyValue]`. we can add this in the `card-basics-test` in the host tests. we should test for Boolean fields and we should test any other field that has the `BaseDef`'s `undefined` empty value.

packages/host/app/resources/room.ts
FadhlanR: Yes, it would be good if the store handled this.

packages/base/boolean.gts
burieberry: Do we want boolean fields to return `false` by default?

FadhlanR: From the [`[serialize]`](https://github.com/cardstack/boxel/blob/578cced5fdf5416173c0dd9c956dd726d9d875ca/packages/base/boolean.gts#L40) and the [`[deserialize]`](https://github.com/cardstack/boxel/blob/578cced5fdf5416173c0dd9c956dd726d9d875ca/packages/base/boolean.gts#L44) function in BooleanField, it seems we try to return false if the value null or undefined.

jurgenwerk (approved):
I tested this locally and it works! I confirmed that when a default skill is attached in the assistant and you make changes to the skill, the bot receives the updated skill on next prompt.

packages/base/boolean.gts
habdelra: we could also extend the boolean field class to make a new type of boolean field that returns true by default as well if that is a need.

habdelra (approved):

packages/base/boolean.gts
burieberry: what I was wondering was, do we want to have boolean field serialized as `null` or `undefined` if it has no value. if people are agreeing on `false` as default, that's fine with me 👍

PR #2578, opened 2025-05-13
Avoid runaway set size when tracking clientRequestIds

lukemelia:


packages/host/app/lib/limited-set.ts
```diff
@@ -0,0 +1,64 @@
+/* A bounded set that automatically removes the oldest entries when it reaches a maximum size */
+export default class LimitedSet<T> implements Iterable<T> {
```
habdelra, line 2: would it not be easier to just proxy `Set` and add the new capability in  methods that need it? what way you don't have to reimplement the methods that don't change.

habdelra: or i guess since you are managing your own internal state, you don't really have a choice.... nvm

habdelra (approved):

PR #2579, opened 2025-05-13
Optimistic card creation in code mode playground

habdelra:
This PR introduces optimistic card creation into the playground. 

Also as part of this PR I noticed that we are passing in an unused loader param into our test setup. I added a specific commit to clean that up.


packages/host/app/resources/import.ts
```diff
@@ -26,6 +27,14 @@ export class ImportResource extends Resource<Args> {
...
   modify(_positional: never[], named: Args['named']) {
     let { url, loader } = named;
+    // The loader service is shared between the realm server and the host. this
+    // resource can interfere with indexing in the browser by caching modules in
+    // the loader that we are trying to change in our index. you can use the
+    // `withoutLoaderMonitoring()` test helper to temporarily disable this
+    // resource in scenarios that it is interfering with the tests.
+    if (isTesting() && (globalThis as any).__disableLoaderMonitoring) {
+      return;
+    }
```
habdelra, line 37: Ed has often wondered how our tests can work when we are sharing the loader between the app and the realm server (running in the browser). it turns out they don't always work. this guy can actually hijack the loader in the realm server with cached stale code when you are trying to write an updated module in the realm server.

lukemelia (approved):

PR #2581, opened 2025-05-13
Include recent-cards in playground dropdown, ordered by last-viewed timestamp

burieberry:
- Include `timestamp` when persisting cards in RecentFilesService and RecentCardsService (had to update many tests due to this change)
- Combine recent-cards and recent-files ids in playground, order ids by timestamp, use the latest version if there are duplicates
- Sort prerendered-card search results by the same order provided by recentCardIds
- Moved recent-card and recent-file test helpers to a separate file from playground helpers

### DIFF
--- a/packages/host/tests/acceptance/code-submode/spec-test.gts
+++ b/packages/host/tests/acceptance/code-submode/spec-test.gts
@@ -18,10 +18,10 @@ import {
 import { setupMockMatrix } from '../../helpers/mock-matrix';
 import {
   getPlaygroundSelections,
-  getRecentFiles,
   assertCardExists,
   selectDeclaration,
 } from '../../helpers/playground';
+import { getRecentFiles } from '../../helpers/recent-files-cards';
 
 import { setupApplicationTest } from '../../helpers/setup';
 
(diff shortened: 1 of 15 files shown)


packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -935,11 +935,13 @@ module('Acceptance | Spec preview', function (hooks) {
     await click(`[data-test-card="${petId}"]`);
 
     // Verify the card was added to recent files
-    assert.deepEqual(
-      getRecentFiles()?.[0],
-      [testRealmURL, 'Pet/mango.json', null],
+    let recentFile = getRecentFiles()?.[0];
+    assert.strictEqual(
+      `${recentFile?.[0]}${recentFile?.[1]}`,
```
backspace, line 940: This is something that already existed before your PR, but seeing this `[0]` and `[1]` makes me wonder why the `RecentFile` type doesn’t have named properties instead of meaningful array indices. Then the first argument would be so much nicer: `${recentFile.realmURL}${recentFile.filePath}`

I’m sure it’s because that’s how the files are stored in local storage, but since we already have a parsing/serialisation layer to extract/store them, we might as well produce a nicer object!

Anyway, a possible future improvement that would make this assertion easier to read 🤓

backspace (approved):
I tried this in the preview deployment, works as expected! 🎉

packages/host/tests/acceptance/code-submode/spec-test.gts
burieberry: When accessing recent-files via the service, the RecentFile type is an object with named properties, but it is serialized into an array for local storage. We have tests that use the recent-file-service, but also some tests want to access the local-storage directly. When accessing it directly, we don't get the named properties. I was wondering if I should convert them to use the service as well, but it seemed ok that sometimes we're testing something without taking extra steps in the UI to fill local storage

backspace: ohhh yes, I was thinking of this being a DX annoyance when working in the application, not just with the test helper, oops! I guess the helper could use the same code as the service, but not a big deal.

PR #2585, opened 2025-05-14
Balance reference count  for search resource during index event

habdelra:
Fix bug where reference count was not balanced in live search resource when there is an index event

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -76,10 +76,6 @@ export class SearchResource extends Resource<Args> {
     this.#previousQuery = query;
     this.#previousRealms = realms;
 
-    for (let instance of this._instances) {
-      this.store.dropReference(instance.id);
-    }
-
     this.loaded = this.search.perform(query);
 
     if (isLive) {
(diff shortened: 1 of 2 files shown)


tintinthong:
Screenshot 2025-05-14 at 22 33 24

I dont understand why. But, I was experimenting spamming a card with saves (this is actually my personal branch). Somehow this actually gets to -1. 

Cud be more emblematic of https://linear.app/cardstack/issue/CS-8631/auto-save-issues-many-patches-should-we-reduce-this#comment-21e51573.

tintinthong (approved):
I have made a note about some testing behaviour. But code-wise everything still looks good to me

habdelra:
> Screenshot 2025-05-14 at 22 33 24
> I dont understand why. But, I was experimenting spamming a card with saves (this is actually my personal branch). Somehow this actually gets to -1.
> 
> Cud be more emblematic of https://linear.app/cardstack/issue/CS-8631/auto-save-issues-many-patches-should-we-reduce-this#comment-21e51573.

yeah, i think that is a different issue. I'll look into it as part of CS-8631. oh and also the PR for the optimistic card creation for playground had a bunch of dropReference cleanup around auto saving that might help here too. https://github.com/cardstack/boxel/pull/2579

PR #2586, opened 2025-05-14
Introduce the "fix it" button to send error to the ai assistant

FadhlanR:
In this PR, I implemented the "Fix it" button. When clicked, it sends the error message and the related file to the AI assistant.

Syntax Error:

https://github.com/user-attachments/assets/32bdff6e-bdcc-4f4b-b4b1-2953d953499a

Error in playground panel:

https://github.com/user-attachments/assets/55a8de5d-7be5-4cfc-87c9-e1e9e9d71bdd


packages/host/app/components/operator-mode/card-error-detail.gts
```diff
@@ -54,22 +67,31 @@ export default class CardErrorDetail extends Component<Signature> {
...
       >
         <:title>
-          <ExclamationCircle class='error-icon' />
-          {{if @headerText @headerText 'An error was encountered: '}}
-          <span data-test-error-title>
-            {{if @title @title @error.title}}
-          </span>
+          <div class='title'>
+            <ExclamationCircle class='error-icon' />
+            {{if @headerText @headerText 'An error was encountered: '}}
```
jurgenwerk:
```suggestion
            {{unless @headerText  'An error was encountered: '}}
```

jurgenwerk (commented):
1. I think "FixItButton" isn't the best name, it could be named in a way where it's clearer from the name what it does, for example `<SendToAiAssistant @fileToAttach={{fileToAttach}} @message={{message}},...>`

2. Consult with the design to get a design for the button (red isn't the best as it expresses negativity)

packages/host/app/components/operator-mode/fix-it-button.gts
```diff
@@ -0,0 +1,101 @@
...
+  private sendToAiAssistant = restartableTask(async () => {
+    await this.aiAssistantPanelService.openPanel();
+
+    if (!this.matrixService.currentRoomId) {
+      throw new Error('No room found');
+    }
+
+    await this.matrixService.sendMessage(
+      this.matrixService.currentRoomId,
+      `In the attachment file, I encountered an error that needs fixing:\n\n${this.errorMessage}.`,
```
jurgenwerk:
```suggestion
      `In the attached file, I encountered an error that needs fixing:\n\n${this.errorMessage}.`,
```

FadhlanR:
I've implemented the new design. Here is the latest screenshot: 


https://github.com/user-attachments/assets/653bad97-3450-4c01-aff3-27974b16cc82



https://github.com/user-attachments/assets/c3106325-273d-4eef-8b00-a1e360e125f1

jurgenwerk:
In your video when you fix the file it says "playground is currently not supported for this file type". Is this a new error?

FadhlanR:
> In your video when you fix the file it says "playground is currently not supported for this file type". Is this a new error?

Not it's not, it's expected since the cursor position in monaco editor is not in the author card definition.

burieberry (commented):
Looking at the `boom-pet.gts` error card and its instance on local, something looks broken:
In interact mode, no details show up when expanding the box. Also, is the "Fix It" button supposed to appear here too?
interact

Code mode playground:
- Not seeing error details when expanding the box
- Hovering on the "Fix It" button causes the button to shift up. I see some css `transform: translateY(-1px);` why is this needed?
error-details

If the details is only when there's a stack trace, maybe we shouldn't display it if there's no stack trace.

jurgenwerk:
> Not it's not, it's expected since the cursor position in monaco editor is not in the author card definition.

It looks to me it's disruptive to the user's flow. I would expect the instance in the playground to show after error is fixed. If this is out of scope for this PR we should add a ticket I think

FadhlanR:
> Looking at the `boom-pet.gts` error card and its instance on local, something looks broken: In interact mode, no details show up when expanding the box. Also, is the "Fix It" button supposed to appear here too? interact
> 
> Code mode playground:
> 
> * Not seeing error details when expanding the box
> * Hovering on the "Fix It" button causes the button to shift up. I see some css `transform: translateY(-1px);` why is this needed?
> 
> error-details
> If the details is only when there's a stack trace, maybe we shouldn't display it if there's no stack trace.

@burieberry For now this button will displayed for syntax error and the card error in playground panel. I removed the `transform: translateY(-1px);` and I fixed the blank details by adding better handling for cases where the stack trace is not provided. Instead of hiding the details area, I now display a message saying the stack trace is not available. Hiding the details area would also require hiding the "Show/Hide Details" button, and since there's only a copy button in the bottom-right, it looked weird.

With no stack trace:

Screenshot 2025-05-20 at 13 43 39

With stack trace:

https://github.com/user-attachments/assets/61985440-ba59-4bbd-bf36-e23510225965

FadhlanR:
> > Not it's not, it's expected since the cursor position in monaco editor is not in the author card definition.
> 
> It looks to me it's disruptive to the user's flow. I would expect the instance in the playground to show after error is fixed. If this is out of scope for this PR we should add a ticket I think

Maybe the screenshot wasn’t very clear—here’s the full screenshot before and after the error. As you can see in the video below, the playground panel is not displayed because the editing is happening in the JavaScript class, not in the card or field definition. That’s why the playground panel doesn't appear.


https://github.com/user-attachments/assets/c3061647-36ae-4683-abd7-413d2898ca36

jurgenwerk:
I am trying to test this and the button is not showing for me: 
image

packages/host/app/components/operator-mode/error-display.gts
```diff
@@ -0,0 +1,250 @@
...
+    <div class='error-display' data-test-error-display>
+      <div class='error-header'>
+        <div class='error-type' data-test-error-type>
+          <ExclamationTriangleFill class='error-icon' />
+          <span>{{this.headerText}}</span>
+        </div>
+        {{#if
+          (and
+            (bool @fileToAttach)
+            (not this.operatorModeStateService.aiAssistantOpen)
```
jurgenwerk: Why not show the fix button if the ai panel is open?

FadhlanR: To make more room: https://discord.com/channels/584043165066199050/584061276544827432/1372990016581799976.

FadhlanR: Or do you think it is better to not hide it?

jurgenwerk: I'm confused, room for which thing? I think the link you gave is talking about "Ask AI" input not the "Fix it" button

Yes, I don't think hiding this button when AI panel is open makes any sense. We need that button so that you can use it to quickly fix an error when you are working with the assistant.

FadhlanR: Ah I thought "Ask AI" is the "Fix it" button. I'll update it.

packages/host/app/components/operator-mode/error-display.gts
```diff
@@ -0,0 +1,250 @@
...
+                trace could be generated.</p>
+            {{/if}}
+          </div>
+        </div>
+      {{/if}}
+    </div>
+
+    <style scoped>
+      .error-display {
+        background: #ffba00;
```
burieberry: you can add this color to `variables.css` as `boxel-warning-200`

packages/boxel-ui/addon/src/icons.gts
```diff
@@ -20,6 +20,7 @@ import DropdownArrowFilled from './icons/dropdown-arrow-filled.gts';
 import DropdownArrowUp from './icons/dropdown-arrow-up.gts';
 import Exclamation from './icons/exclamation.gts';
 import ExclamationCircle from './icons/exclamation-circle.gts';
+import ExclamationTriangleFill from './icons/exclamation-triangle-fill.gts';
```
burieberry: There's a very similar icon `warning.svg` already existing here. Does that icon work?

packages/host/app/components/operator-mode/card-error-detail.gts
```diff
@@ -1,146 +1,57 @@
...
       .error-detail {
         flex: 1.5;
-        overflow: auto;
+        overflow: visible;
         margin-top: auto;
         max-height: fit-content;
-      }
-      .error-detail :deep(.accordion-item) {
-        height: auto;
+        margin: var(--boxel-sp);
```
burieberry: the `margin` in line 48 is overriding the `margin-top` in line 46. as a result, this component does not always appear at the bottom of the page.

packages/host/app/components/operator-mode/error-display.gts
```diff
@@ -0,0 +1,250 @@
...
+      {{#if @message}}
+        <div class='error-message' data-test-error-message>
+          {{@message}}
+        </div>
+      {{/if}}
+
+      <div class='error-actions'>
+        <CopyButton @textToCopy={{@message}} @width='16px' @heigth='16px' />
+        <button
+          class='toggle-details-button'
```
burieberry: Can probably remove a lot of extra css by using `<Button @kind='text-only' @size='extra-small' ...`

packages/host/app/components/operator-mode/send-error-to-ai-assistant.gts
```diff
@@ -0,0 +1,103 @@
...
+      [],
+      this.args.fileToAttach ? [this.args.fileToAttach] : [],
+    );
+  });
+
+  <template>
+    <Button
+      class='send-error-to-ai-assistant'
+      @kind='primary'
+      @size='small'
```
burieberry: maybe this is better: `@kind='secondary-dark' @size='extra-small'`

packages/host/app/components/operator-mode/send-error-to-ai-assistant.gts
```diff
@@ -0,0 +1,103 @@
...
+        --boxel-button-color: var(--boxel-ai-purple);
+        --boxel-button-border: 1px solid var(--boxel-ai-purple);
+        --boxel-button-text-color: var(--boxel-light);
+        padding: 6px 12px;
+        font-size: var(--boxel-font-size-xs);
+        font-weight: 500;
+        display: inline-flex;
+        align-items: center;
+        gap: 6px;
+        transition: all 0.2s ease;
```
burieberry: can use more specific transition properties if this is needed

burieberry (commented):
Made some additional suggestions, but overall it looks good to me

packages/host/app/components/operator-mode/error-display.gts
FadhlanR: @jurgenwerk updated, please check it again

burieberry (approved):
I think I saw a `console.log` leftover in the code, otherwise looks good

PR #2587, opened 2025-05-14
Debounce auto-saves

habdelra:
This PR debounces our auto-saves. previously we were issuing a PATCH request for each keystroke made in a field. now we gather up all the auto-save events that are issued while we are awaiting the response from the server for the previous auto-save and collapse them into a single save.


tintinthong (approved):
I have tested and the behaviour is stable for saves now

PR #2590, opened 2025-05-14
Fixing dollar curly parsing bug

ef4:
This replaces the whole JS-representation with one that hopefully isn't sensitive to string escaping problems.

~The previous implementation looks like it attempted to preserve offsets, but it definitely didn't actually, and we confirmed that it doesn't matter if it does.~

The previous implementation attempted to preserve offsets, but had off-by-some-small-amount errors when escaping backticks inside templates. Our new implementation only preserve lines, not exact offsets, which is good enough for our present use cases.

It would be good to refactor ModuleSyntax to have a more explicit API for asking about source locations, because we could fairly easily give precise, adjusted answers. At present it leaks a lot of AST internals instead, which are inconvenient to try to wrap/mutate.

### DIFF
--- a/packages/realm-server/tests/module-syntax-test.ts
+++ b/packages/realm-server/tests/module-syntax-test.ts
@@ -1,5 +1,9 @@
 import { module, test } from 'qunit';
-import { ModuleSyntax } from '@cardstack/runtime-common/module-syntax';
+import {
+  ModuleSyntax,
+  gjsToPlaceholderJS,
+  placeholderJSToGJS,
+} from '@cardstack/runtime-common/module-syntax';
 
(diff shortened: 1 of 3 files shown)


packages/realm-server/tests/module-syntax-test.ts
```diff
@@ -86,7 +86,7 @@ module(basename(__filename), function () {
         @field firstName = contains(StringField);
         @field age = contains(NumberField);
         static embedded = class Embedded extends Component<typeof this> {
-          <template><h1><@fields.firstName/></h1></template>                
+          <template><h1><@fields.firstName/></h1></template>
```
ef4, line 93: The tests were previously assertion that we **add** whitespace here, which was a bug.

PR #2594, opened 2025-05-15
add realm icon and max-height with overflow scroll

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8542/dropdown-realm-shud-have-icons

1. add realm icon and display realm name to remix dropdown
2. add max-height for the dropdown

Result:
image


richardhjtan (approved):
Looks good @lucaslyl !

PR #2595, opened 2025-05-15
CS-8351 Remix action by AI assistant in listing 

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8351/add-remix-button

Run remix action after the user selects a realm to install when clicking on catalog listing in fitted format. 

**Screenshot**

https://github.com/user-attachments/assets/a7c67012-95e2-43a1-b663-641245d2429d

What is remix action?
- Same as the 'install' action, we install the source file and copy the examples to the user's selected realm.
- After the 'install' action is done, we persist the first source file as code selection
- If there is example from the listing, we take and persist the first example in playground selection for preview
- Redirect them to the code submode
- Suggest them two example prompts which is related to the listing if possible.

### What is changing
- Add remix skill
- Add commands: 
  - Remix
  - Update playground selection
  - Update code path with selection


packages/catalog-realm-dev/catalog-app/components/listing-fitted.gts
```diff
@@ -348,9 +386,59 @@ export class ListingFittedTemplate extends Component<typeof Listing> {
...
 
-  @action remix(e: MouseEvent) {
+  _remixWithAI = task(async (realmUrl: string) => {
+    let commandContext = this.args.context?.commandContext;
+    if (!commandContext) {
+      throw new Error('Missing commandContext');
+    }
+
+    if (!this.roomId) {
+      if (!this.roomId) {
```
tintinthong: 2 room ids?

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -369,6 +369,9 @@ export default class InteractSubmode extends Component {
       allRealmsInfo: async () => {
         return await here.realm.allRealmsInfo;
       },
+      fetchCard: async (url: string) => {
+        return await here.store.get(url);
```
tintinthong: is the skill card in the identity context already? since its in base...

In which case, we prob want to use `store.peek` since that just looks into the identity map. There is no issue in using that bcos the `remixSkillCard` is consumed straight away

packages/catalog-realm-dev/catalog-app/listing/listing.gts
```diff
@@ -63,49 +64,168 @@ function nameWithUuid(listingName?: string) {
   return newPackageName;
 }
 
+export async function installListing(args: any, realm: string) {
```
tintinthong, line 67: can we be explicit about args. you are just using CatalogActions and types are availabe for that

packages/host/app/commands/remix.ts
```diff
@@ -0,0 +1,234 @@
...
+function nameWithUuid(listingName?: string) {
+  if (!listingName) {
+    return '';
+  }
+  // sanitize the listing name, eg: Blog App -> blog-app
+  const sanitizedListingName = deburr(listingName.toLocaleLowerCase())
+    .replace(/ /g, '-')
+    .replace(/'/g, '');
+  const newPackageName = `${sanitizedListingName}-${uuidv4()}`;
+  return newPackageName;
```
tintinthong, line 41: think can put in runtime common

packages/host/app/commands/remix.ts
```diff
@@ -0,0 +1,234 @@
...
+      throw new Error(`Invalid realm: ${realmUrl}`);
+    }
+
+    const copyMeta: CopyMeta[] = [];
+
+    const localDir = nameWithUuid(listing.name);
+
+    // first spec as the selected code ref with new url
+    // if there are examples, take the first example's code ref
+    let selectedCodeRef;
```
tintinthong, line 87: missing type for codeRef

packages/host/app/commands/remix.ts
```diff
@@ -0,0 +1,234 @@
...
+    let shouldPersistPlaygroundSelection = false;
+    let firstExampleCardId;
+
+    // Copy the gts file based on the attached spec's moduleHref
```
tintinthong, line 91: Presumably, you are proceeding to refactor this via `_install` as a command ticket

tintinthong:
Host test failing. I noticed @backspace removed a flaky test -- `_use`. Are these flaky too?

tintinthong:
we need test to prevent regression

packages/host/app/commands/remix.ts
richardhjtan: Yes, there will be continuing work for refactoring about new install command

packages/catalog-realm-dev/catalog-app/components/listing-fitted.gts
richardhjtan: opps - I've removed that, also fixed when click remix on another/same listing again it _should_ create a new room

packages/catalog-realm-dev/catalog-app/listing/listing.gts
richardhjtan: I will be on hold because we will transition this logic to become a command. And currently in remix, we don't have the `Listing` type in the base command

packages/base/commands/remix-input-result.gts
```diff
@@ -0,0 +1,57 @@
...
+} from '../card-api';
+import BooleanField from '../boolean';
+
+import IconTerminal2 from '@cardstack/boxel-icons/terminal-2';
+
+export class RemixInput extends CardDef {
+  @field realm = contains(StringField);
+  @field listing = linksTo(CardDef);
+}
+
```
tintinthong: do we still need remix result now? Is it meaningful? 

Chris mentioned its confusing and the tick should be enough to indicate a success. 

Unless the remix result is indicating a failure state that the user may want to know?

packages/catalog-realm-dev/catalog-app/listing/listing.gts
tintinthong: so if this entire install becomes a command.. will it solve our type woes?

packages/host/app/commands/remix.ts
```diff
@@ -0,0 +1,234 @@
...
+
+      await new UpdateCodePathWithSelectionCommand(this.commandContext).execute(
+        {
+          codeRef: selectedCodeRef,
+          localName: selectedCodeRef.name,
+          fieldName: undefined,
+        },
+      );
+
+      await timeout(1000);
```
tintinthong, line 220: Can there be a comment here of why there is a timeout? and what it is waiting for

tintinthong (approved):
I think this looks good. Lets merge this in and get everyone on ecosystem to test it

packages/catalog-realm-dev/catalog-app/components/listing-fitted.gts
```diff
@@ -348,9 +381,52 @@ export class ListingFittedTemplate extends Component<typeof Listing> {
     return this.args.model.tags?.[0]?.name;
   }
 
-  @action remix(e: MouseEvent) {
+  _remixWithAI = task(async (realmUrl: string) => {
```
tintinthong: All of this maybe is `openRoomWithSkillAndPrompt`

packages/base/commands/remix-input-result.gts
richardhjtan: It is not needed now after this ticket changes - https://linear.app/cardstack/issue/CS-8656/aibot-should-be-given-a-chance-to-respond-when-command-result-sent , I will revert it

packages/catalog-realm-dev/catalog-app/listing/listing.gts
richardhjtan: i think will solve most of it because most of them are commands, left the Listing type which needs to be addressed

packages/host/app/commands/remix.ts
richardhjtan: i will note it down to be refactored together with install command, thanks

PR #2597, opened 2025-05-15
Headless browser prerendering

ef4:
This should now have a working test that demonstrates capturing all the things that indexing captures from a running card, using headless Chrome via puppeteer.

TODOs:

- [ ] error handling. This can take advantage of full control over the browser window as needed, much more robust than our in-page, one-shot error trap.
- [ ] defer until async relationships are loaded. This is also easier for the same reason -- the existing error trap doesn't do re-renders, but now we have a normal re-render.
- [ ] auth: my test is using a public realm. Locally I tested personal realms by copying my own session. We need to automatically pass a session to the browser (see process.env.BOXEL_SESSION)
- [ ] pooling / reuse (the first draft here creates a whole browser for each call to `prerenderCard`)
- [ ] replace the fastboot usage in the real `current-run.ts` with this new implementation
- [ ] cleanup all the old implementation


ef4:
In order to implement the error handling, the browser needs to switch to loading the card in a way that doesn't itself hit the existing index (and thus get an error document back).

We could use `Accept: application/vnd.card+source` on the actual JSON file (see `getSource` on `CardService` and also the `File` resource), although it seems like it would be better to think about how the actual card json endpoint can be "progressively enhanced" with optional pre-rendering, or even split into more than one resource.

A related thing I'm noticing is that this branch can make a distinction about which parts are failing. If isolated render fails but other parts succeed, we don't really need to store the whole card as an error doc.

ef4:
I'm going to land this so we don't have to maintain a long-running branch. Right now it shouldn't have any production impact.

PR #2598, opened 2025-05-15
Add currencyFormat helper and update skill

lukemelia:


backspace (approved):
Build #12939 - Cardstack - Percy | Visual testing as a service 2025-05-15 14-21-10

Strange Percy diff 🥴

PR #2604, opened 2025-05-16
Change monaco-editor theme to be dark

burieberry:
Switch to "vs-dark" base monaco theme:
- regular editor background: #413e4e 
 - read-only editor background: #606060
 - kept default dark for AI-panel code blocks because #413e4e is too similar to AI assistant message background
 
Monaco editor and AI Panel view-code block:
view-code

Diff editor:
diff-editor

Read-only editor:
read-only-editor

### DIFF
--- a/packages/host/app/components/ai-assistant/formatted-aibot-message.gts
+++ b/packages/host/app/components/ai-assistant/formatted-aibot-message.gts
@@ -221,16 +221,16 @@ export default class FormattedAiBotMessage extends Component<FormattedAiBotMessa
         max-height: 30vh;
       }
 
-      /*
-        This filter is a best-effort approximation of a good looking dark theme that is a function of the white theme that
-        we use for code previews in the AI panel. While Monaco editor does support multiple themes, it does not support
-        monaco instances with different themes *on the same page*. This is why we are using a filter to approximate the
-        dark theme. More details here: https://github.com/Microsoft/monaco-editor/issues/338 (monaco uses global style tags
-        with hardcoded colors; any instance will override the global style tag, making all code editors look the same,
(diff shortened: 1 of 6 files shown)


habdelra (approved):

packages/host/app/components/ai-assistant/formatted-aibot-message.gts
```diff
@@ -221,16 +221,16 @@ export default class FormattedAiBotMessage extends Component<FormattedAiBotMessa
...
-        we use for code previews in the AI panel. While Monaco editor does support multiple themes, it does not support
-        monaco instances with different themes *on the same page*. This is why we are using a filter to approximate the
-        dark theme. More details here: https://github.com/Microsoft/monaco-editor/issues/338 (monaco uses global style tags
-        with hardcoded colors; any instance will override the global style tag, making all code editors look the same,
-        effectively disabling multiple themes to be used on the same page)
-      */
+      /* our dark-mode background-color is too similar to AI-Assistant message background,
+        so we are using a darker background for code-blocks */
       :global(.code-block .monaco-editor) {
-        filter: invert(1) hue-rotate(151deg) brightness(0.8) grayscale(0.1);
```
jurgenwerk, line 233: Really nice to see this hack being deleted!

jurgenwerk (approved):
- This is a personal opinion but I'm not a fan of this background color, it makes the text color look a bit dimmed/muted. Perhaps a slightly darker and grayer color would be better.

image

- When you toggle between files in the file tree explorer, the main monaco editor will flash with white background until the next content is shown. Perhaps we should set the background of the panel to the color of the editor.

richardhjtan (approved):

PR #2608, opened 2025-05-19
AI bot debug

IanCal:
This adds a new debug event type to distinguish between these debug responses and regular messages. These debug responses 

In debug mode (debug=true in the url), on every response the event list and prompt will be sent and can be downloaded.

Out of debug mode, you can send a message of

debug:promptandevents

To get the same. However, that event list would include that message sent, and you often want something to answer "what would have been sent for an older message". You can append a number to have that many events stripped off - debug:promptandevents:3 will remove the last 3 events from the list, etc. Without the 

This is not perfect, since there's no guarantee how files/etc were read at the time, but it's helpful if you see something without debug mode turned on. These files can be shared so someone with a bug can share the event list and we can dig into it more.

Other than tools (which are sent only to models that support them), in debug mode the prompt that's returned is the same as the prompt that gets sent to the models.


lukemelia (approved):

jurgenwerk (approved):
I tested this and I was able to download the JSON file containing the data dump. Nice!

image

PR #2609, opened 2025-05-19
add guard to prevent auto save queue from going into infinite loop

tintinthong:


habdelra:
do we know why there is an infinite loop? that seems like the actual fix. do you have a way to reproduce this issue?

tintinthong:
> do we know why there is an infinite loop? that seems like the actual fix

I dont know why. Its tied to the over-arching effect of using `getCards with isLive=true`. Causing promises to not resolve upon spamming the keystrokes

I created this PR with the intent of more debugging

PR #2611, opened 2025-05-19
simpler fix to spec auto-save

tintinthong:
This prevents component re-renders BUT 

there is still data streaming in when editing a spec causing user typing to be interrupted

https://linear.app/cardstack/issue/CS-8652/getcards-conflicts-with-autosaves


packages/host/app/resources/search.ts
```diff
@@ -203,6 +190,9 @@ export class SearchResource extends Resource<Args> {
         this._instances.length,
         ...results.map((instance) => instance),
       );
+      if (this.#doWhileRefreshing) {
+        this.#doWhileRefreshing();
+      }
```
tintinthong, line 199: This is a nice pattern I found that enables me to set the `activeSpec` in local storage when a search is returned. Then I handle the `activeSpec` independently with a card resource

This prevents reactivity getting intertwined with the list of specs being returned by the card resource

packages/host/app/resources/search.ts
```diff
@@ -139,14 +134,6 @@ export class SearchResource extends Resource<Args> {
...
   }
 
-  private doWhileRefreshing = restartableTask(
-    async (
-      doWhileRefreshing: (ready: Promise<void> | undefined) => Promise<void>,
-    ) => {
-      await doWhileRefreshing(this.loaded);
-    },
-  );
-
```
tintinthong, line 149: Oh perhaps I still need this pattern of returning the promise

habdelra:
i notice that the live search tests are failing, so i'm a bit hestitant to go with this approach....

tintinthong:
> i notice that the live search tests are failing, so i'm a bit hestitant to go with this approach....

thats bcos test rely on loaded property. I removed it without considering test uses that private property to await search

habdelra:
ah i see. I'm gonna try starting from this PR https://github.com/cardstack/boxel/pull/2612 and see if I land on a similar solution as this

habdelra:
it looks like we still have an unstable array as a component arg taht we discussed. that could end up being a landmine that someone runs int oin the future...

habdelra:
my bad, the array actually isn't unstable at all. I forgot that we actually mutate it instead of replace it in the search resource

habdelra:
@tintinthong this looks like it is working now. wanna check it our on your end too?

https://github.com/user-attachments/assets/c338c9d0-93bf-432c-8bb5-4864fa9ba08f

habdelra:
hmmm actually i see a problem. when I switch modules the spec does not change to the spec associated with the code ref of the different module

tintinthong:
> hmmm actually i see a problem. when I switch modules the spec does not change to the spec associated with the code ref of the different module

Yes i saw this yesterday before i went to bed. Its bcos doWhileRefreshing is only running the assignment when there is no existing active spec. The active spec selection is also changing when u choose a boxel dropdown.

tintinthong:
> > hmmm actually i see a problem. when I switch modules the spec does not change to the spec associated with the code ref of the different module
> 
> Yes i saw this yesterday before i went to bed. Its bcos doWhileRefreshing is only running the assignment when there is no existing active spec. The active spec selection is also changing when u choose a boxel dropdown.
> 

I also noticed possibly something bad. When u select a defintion the active spec in local storage is set to null. 

Meaning inside test, there may be unexpected behaviour if doWhileRefreshing occurs before selecting a definition -- u get spec id which is null

I think onenplayground test failed bcos of this

habdelra:
> > > hmmm actually i see a problem. when I switch modules the spec does not change to the spec associated with the code ref of the different module
> > 
> > 
> > Yes i saw this yesterday before i went to bed. Its bcos doWhileRefreshing is only running the assignment when there is no existing active spec. The active spec selection is also changing when u choose a boxel dropdown.
> 
> I also noticed possibly something bad. When u select a defintion the active spec in local storage is set to null.
> 
> Meaning inside test, there may be unexpected behaviour if doWhileRefreshing occurs before selecting a definition -- u get spec id which is null
> 
> I think onenplayground test failed bcos of this

yeah i think that is the last failing test

packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -623,6 +634,73 @@ module('Acceptance | Spec preview', function (hooks) {
     assert.dom('[data-test-create-spec-intent-message]').exists();
     await percySnapshot(assert);
   });
+  test('spec updates when different declaration selected in the module', async function (assert) {
```
tintinthong, line 637: 👍

packages/host/tests/acceptance/code-submode/spec-test.gts
```diff
@@ -623,6 +634,73 @@ module('Acceptance | Spec preview', function (hooks) {
...
+    assert.dom('[data-test-boxel-input-id="spec-title"]').hasValue('Person');
+    await click('[data-test-file-browser-toggle]');
+    await waitFor('[data-test-file="pet.gts"]');
+    await click('[data-test-file="pet.gts"]');
+    await click('[data-test-accordion-item="spec-preview"] button');
+    assert.dom('[data-test-boxel-input-id="spec-title"]').hasValue('Pet2');
+    await click('[data-test-file="person.gts"]');
+    assert.dom('[data-test-boxel-input-id="spec-title"]').hasValue('Person');
+  });
+  test('does not lose input field focus when editing spec', async function (assert) {
```
tintinthong, line 666: 👍

tintinthong:
I have tested and the spec auto-save is stable now

habdelra (approved):
great work @tintinthong

PR #2612, opened 2025-05-19
Better merge of server state into store

habdelra:
This PR performs a better merge of server state into the store. The idea is that the store represents the latest and greatest state, and the server is generally out of date compared to the store. This PR only merges the server state that the store doesnt know about. otherwise we don't interfere with the store's state.

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -9,6 +9,7 @@ import { formatDistanceToNow } from 'date-fns';
 import { task } from 'ember-concurrency';
 
 import cloneDeep from 'lodash/cloneDeep';
+import isEqual from 'lodash/isEqual';
 import mergeWith from 'lodash/mergeWith';
 
 import { stringify } from 'qs';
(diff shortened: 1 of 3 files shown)


packages/host/app/services/store.ts
```diff
@@ -1216,6 +1193,16 @@ function processCardError(
   }
 }
 
+function needsServerStateMerge(
+  instance: CardDef,
+  serverState: SingleCardDocument,
+): boolean {
+  return (
+    instance.id === serverState.data.id &&
+    isEqual(instance[meta], serverState.data.meta)
```
lukemelia: Is there some kind of version in meta? Why is this check meaningful?

habdelra: the idea is that we don't touch the identity context (which is tracked state) when nothing is different. anytime you set tracked state even if its not different you'll trigger glimmer invalidation. meta matters because that is where the realm info lives which the store doesn't know about. i was trying not to make assumptions about what is currently in meta and when it may change. (like perhaps a realm name changes or a realm background image changes).

habdelra: probably we can be even more precise here, in that we don't want to merge server state when the polymorphic field info changes. the store already knows about that...

habdelra: lol, i named this function backwards from what it is doing

habdelra:
@tintinthong i think the fix for spec is actually just the search form of this change. specifically when we search we don't want to merge the server state into the store for any instances that are already in our store.

lukemelia (approved):

PR #2618, opened 2025-05-20
Possibly better instructions for the LLM to not respond with omitted code placeholders

jurgenwerk:
I've been seeing this bug in really long assistant code sessions (with gemini 2.5)

image

I changed the skill instruction to be more exact, it looks like it helped because after this change I wasn't getting the placeholder output (in the same session)


IanCal:
as a heads up with the issues with matrix if you have long chats they may not have the skills and will be using gpt-4o

habdelra (approved):

jurgenwerk:
@IanCal due to the 1000 events limit?

PR #2619, opened 2025-05-20
Improve loading time when user logs in

FadhlanR:
In this PR, I implemented lazy loading for workspaces when the user logs in. This means the user doesn't need to wait for all workspace authentications to complete before seeing the workspace chooser or operator mode page. This works because on the workspace chooser page, we re-trigger authentication for each realm/workspace, and in the interact/code submode pages, we have a reauthentication mechanism that handles 401 errors when fetching cards or files. You can see in the screenshots below how workspaces and cards are now loaded lazily. The loading time should be faster on your end.

Login to workspace chooser:

https://github.com/user-attachments/assets/5da22f09-6bac-4936-8301-f0d12e0b80b9

Login and directly access submode:

https://github.com/user-attachments/assets/7388b614-d777-42ab-a002-4ab949b22639


packages/host/app/routes/index.ts
```diff
@@ -73,8 +73,30 @@ export default class Index extends Route<void> {
     let cardUrl: string | undefined;
     if (cardPath) {
       if (hostsOwnAssets) {
-        cardUrl = new URL(`/${cardPath}`, this.realm.defaultReadableRealm.path)
-          .href;
+        let realmUrl = Object.keys(this.realm.allRealmsInfo).find(
```
jurgenwerk: 1. Is it guaranteed `allRealmsInfo` will be loaded at this point?
2. Would be good to have a test for this new card-realm matching behaviour
3. This model hook is getting huge - extract your addition into a function?

packages/host/app/services/matrix-service.ts
```diff
@@ -510,18 +510,24 @@ export default class MatrixService extends Service {
...
         }>(APP_BOXEL_REALMS_EVENT_TYPE);
         await this.realmServer.setAvailableRealmURLs(
           accountDataContent?.realms ?? [],
         );
-        await Promise.all([
+
+        await this.initSlidingSync(accountDataContent);
+        await this.client.startClient({ slidingSync: this.slidingSync });
+
+        // Do not need to wait for theses to complete,
```
jurgenwerk, line 523: There are some typos in this comment that are making it hard to follow

packages/host/app/services/realm.ts
```diff
@@ -641,7 +638,7 @@ export default class RealmService extends Service {
   private identifyRealm = task(
     { maxConcurrency: 1, enqueue: true },
     async (url: string): Promise<void> => {
-      if (this.knownRealm(url)) {
+      if (this.knownRealm(url, false)) {
```
jurgenwerk, line 647: Wondering why this was needed

packages/matrix/tests/login.spec.ts
```diff
@@ -100,7 +100,58 @@ test.describe('Login', () => {
     await expect(page.locator('[data-test-login-btn]')).toBeEnabled();
     await page.locator('[data-test-login-btn]').click();
 
+    let boxelSessionBeforeCardLoaded = await page.evaluate(async () => {
+      return window.localStorage.getItem('boxel-session');
+    });
+    expect(boxelSessionBeforeCardLoaded).toBeNull();
+
+    await assertLoggedIn(page);
+    // the authentication to the realm could possibly processed when fetching card,
```
jurgenwerk: Same comment, I think some words are missing

packages/host/app/services/matrix-service.ts
```diff
@@ -510,18 +510,24 @@ export default class MatrixService extends Service {
...
+        // in the workspace chooser we'll retrigger login and wait fo them to complete
+        // and when fetching cards or files we have reautentication mechanism.
```
jurgenwerk, line 528: There are some typos in this comment that are making it hard to follow

habdelra:
Also make sure to test this by going straight to an interact mode or code mode URL that has query param state when not logged in. the idea is that the login page shows for the user to authenticate, and after login, the operator-state-service will load the query param state. there might be some edge cases in this scenario where you are directly rendering a stack after login instead of just showing the dashboard.

packages/host/app/routes/index.ts
FadhlanR: You're right — there's no guarantee that `allRealmsInfo` will be loaded here, so instead I use `availableActiveRealms` from `realmServerService`. We already have an existing test for the scenario where a user accesses Boxel using the canonical URL for a card that has an error. However, I added another test to ensure the new logic for card-realm matching works with URLs following the `[username]/[workspace name]` pattern. Please check it again.

packages/host/app/services/realm.ts
FadhlanR: Updated.

packages/matrix/tests/login.spec.ts
FadhlanR: Updated

packages/host/app/services/matrix-service.ts
FadhlanR: Updated

packages/host/app/services/realm.ts
FadhlanR: We need this to avoid the error: "Attempted to update a tracked variable, but it had already been used previously in the same computation." Also, we don't need to use the tracked variable in this task.

FadhlanR:
> Also make sure to test this by going straight to an interact mode or code mode URL that has query param state when not logged in. the idea is that the login page shows for the user to authenticate, and after login, the operator-state-service will load the query param state. there might be some edge cases in this scenario where you are directly rendering a stack after login instead of just showing the dashboard.

@habdelra I added a screenshot of this scenario to the description of the PR. I also added the matrix tests.

habdelra (approved):

jurgenwerk (approved):

PR #2620, opened 2025-05-20
Avoid re-executing already executed commands

lukemelia:
- commands with requiresApproval === false and no result card would be re-executed on load, wreaking havok


lukemelia:
Feedback welcome, but I'm going ahead to merge and ship because this is such a pain point at the moment.

PR #2621, opened 2025-05-21
add missing spec and instances

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8681/fix-catalog-sprint-planner-missing-spec-and-card-instances


tintinthong (approved):
This is good to merge in but we will resolve one of the issues in this ticket 

https://linear.app/cardstack/issue/CS-8685/should-not-have-source-files-determine-where-the-file-is-copied-to

PR #2622, opened 2025-05-21
Add more context to the LLM (what user is currently seeing in code mode)

jurgenwerk:
Something that occured quite often when coding using the assistant and you said something in the lines of: "I can't see the border", "where is the button"..., the LLM decided to automatically transition from code mode to interact mode in order for you to see and inspect the card instance. It didn't know that you are already looking at the instance in playground. You would get unexpected reactions like this:

image

image

You can observe in the above prompts I was trying to tell the assistant that its changes had no visual effect. I was expecting it to examine & fix code, but it decided to go to interact mode and open the card on the stack (which is not what I wanted)

Turns out we can mitigate this issue by adding some more context for the AI, and slightly adjust the instructions.


backspace (approved):

lukemelia:
I'd prefer to update the context we are providing to the aibot to include the playground/preview panel content and then evaluate whether we need additional skill guidance.

jurgenwerk:
@lukemelia I added the following to the system prompt:

```
The user is currently:
- operating in a realm with this URL: http://my-realm.com
- operating in code mode
- inspecting the file contents in this code mode panel: playground
- viewing this card instance in the UI: http://localhost:4201/postcards/1, in this format: isolated
```

However, this still wasn't enough, and I was still able to replicate the issue described in the PR description. I've started seeing success only after also adding this to the card editing skill:

`If the user is viewing the card instance whose card definition you are working with, do not use the showCard function since the user is already looking at it`

packages/host/app/components/operator-mode/code-submode/module-inspector.gts
```diff
@@ -490,6 +498,10 @@ export default class ModuleInspector extends Component<ModuleInspectorSignature>
       />
     {{/if}}
 
+    <ReportUserContextForAiAssistant
```
jurgenwerk: Properties like `selectedAccordionItem` are local component state, so there is a question of how to read this where we send this context to matrix. We could put it into a service but that would include refactoring in several different places. I opted into created a `ReportUserContextForAiAssistant` component which we can stick into any template where we are interested in gathering pieces of data that is sent to AI.

jurgenwerk: More info here: https://github.com/cardstack/boxel/pull/2622/files#diff-74107da2f57a38bed49ccc37a3cac7b46723735b154806573023d5262a6fcabcR33

IanCal:
I like the addition of the information - one thing I'd note is that we have some of this kind of information added to the user messages https://github.com/cardstack/boxel/blob/main/packages/ai-bot/helpers.ts#L585

This may be a better place, the model can see where the user *was* when it took certain actions. There has been a tendency for models to act like things in the system prompt were always there. If we choose to do it in the system prompt, we should probably move this per-message work up to the system prompt to have it in one place, though I think we moved it the other way originally.

jurgenwerk:
@IanCal makes total sense, that way we would give it even more valuable info (context over time), I can try to move this to the place that you linked

burieberry:
1- `playground-selections` in local storage provides the following info:
- key is selected module id in code mode (for example: `.../person.gts/Person` means `Person` CardDef is selected on `person.gts` file)
- values are: 
    - `cardId` (for cards: this is the card instance id selected on playground preview. for fields: this is the spec card Id)
    - `format` (this is the preview format selected: isolated, embedded, etc. existing cards default to isolated, fields default to embedded)
    - `fieldIndex` (this is `undefined` for cards, numbered for fields. it's the index of the field instance on Spec's `containedExamples` containsMany field)
 
 You can get this info by using PlaygroundPanelService (`peekSelection` is especially helpful when just viewing the selection per moduleId) or directly access from local storage if there's no access to the PlaygroundPanelService
    
  2- `code-mode-panel-selections` in local storage provides info on selected panel in code mode (Schema, Preview, Spec etc). Key is file address, value is the id for the selected panel. The default is `schema` unless this has been changed.

jurgenwerk:
@burieberry thanks, good to know. Moving this back to draft since it needs several adaptations

jurgenwerk:
@burieberry I changed my approach to match your suggestion, can you take a look? Just as a note, I added the module url in the persisted playground selections config so that it is more convenient to retrieve it using the current code path (currently opened file in code mode) that is in the operator mode state service.

burieberry (approved):

lukemelia (commented):
Flagging that this may have merge issues with the work that @backspace has in progress.

PR #2624, opened 2025-05-21
Introduce `actionVerb` for Command that is used in button text for AI Assistant

lukemelia:
- this required loading the command (async op) when creating the MessageCommand model from matrix events, which caused several methods in the matrix-events-to-messages loading to become async


packages/host/app/lib/matrix-classes/message-builder.ts
```diff
@@ -193,7 +195,7 @@ export default class MessageBuilder {
         command.commandRequest = decodeCommandRequest(encodedCommandRequest);
       } else {
         message.commands.push(
-          this.buildMessageCommand(
+          await this.buildMessageCommand(
```
habdelra, line 198: is it desired that this be performed serially rather than in parallel (via Promise.all)?

packages/host/app/lib/matrix-classes/message-builder.ts
```diff
@@ -236,15 +238,15 @@ export default class MessageBuilder {
...
+  private async buildMessageCommands(message: Message) {
     let eventContent = this.event.content as CardMessageContent;
     let commandRequests = eventContent[APP_BOXEL_COMMAND_REQUESTS_KEY];
     if (!commandRequests) {
       return new TrackedArray<MessageCommand>();
     }
     let commands = new TrackedArray<MessageCommand>();
     for (let commandRequest of commandRequests) {
-      let command = this.buildMessageCommand(
+      let command = await this.buildMessageCommand(
```
habdelra, line 249: ditto with above: serial vs parallel?

burieberry (approved):

packages/host/app/lib/matrix-classes/message-builder.ts
lukemelia: I think serialially is fine. The only async op is the loader.import and I expect the module to be loaded already in the majority of cases.

PR #2625, opened 2025-05-21
update search_doc during copy operation

habdelra:
This PR patches the id's of the search doc in the index when we are copying records to a different realm as part of an index copy operation.


lukemelia (approved):

tintinthong (approved):
I confirm this works for the seed realm apps 👍

jurgenwerk (approved):

PR #2626, opened 2025-05-21
Introduce new lint rule autofixing missing card-api imports

lukemelia:


packages/eslint-plugin-boxel/docs/rules/missing-card-api-import.md
```diff
@@ -0,0 +1,100 @@
...
+
+💼 This rule is enabled in the ✅ `recommended` config.
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+Auto-fixes missing imports for card-api related symbols in your code.
+
+If you refer to a card-api symbol without importing it:
```
habdelra: "symbol" is misleading. we do have a bunch of symbols that we export from the api, but that is not what you mean here. I think "export"  or "named export" is a better name here

packages/eslint-plugin-boxel/docs/rules/missing-card-api-import.md
```diff
@@ -0,0 +1,100 @@
...
+import { FieldDef, field, linksTo, contains } from 'https://cardstack.com/base/card-api';
+import StringField from 'https://cardstack.com/base/string';
+
+export class Payment extends FieldDef {
+  @field chain = linksTo(Chain);
+  @field address = contains(StringField);
+}
+```
+
+### Partially imported card-api symbols
```
habdelra: ditto. replace "symbols" with "exports"

packages/eslint-plugin-boxel/lib/utils/import-utils.js
```diff
@@ -0,0 +1,163 @@
...
+    // Module is already imported, so add to existing import
+    // Use the first non-type import declaration
+    const existingImport = importDeclarations[0];
+
+    // If it's a default import that we need to add
+    if (exportedName === 'default') {
+      // Check if the default import already exists
+      const hasDefaultImport = existingImport.specifiers.some(
+        (specifier) => specifier.type === 'ImportDefaultSpecifier',
+      );
```
habdelra, line 38: I'm a little confused about how you could fix a default export. For named exports we can see that the delcaration has a meaningful name like `CardDef`. so we can make some good guesses about what to import. In the case of a default export, it could be named anything. 

also in the case of a default export that already exists are you also confirming that the declaration for the default export is the same name that you are already using?

```ts
import Foo from 'foo'`;

foo.blah();

bar.blah(); // this is also a default export from foo that you discovered but its using a different declaration
```
I guess in this case it's not really a missing import, the import is there it' s a bug in the name of the declaration that you are using. but then I have no idea how you are able to detect that the name of the default export is a particular missing card-api default export.

packages/eslint-plugin-boxel/lib/utils/import-utils.js
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Adds an import statement for a missing import, or augments an existing import statement
+ * @param {import('eslint').Rule.RuleFixer} fixer The fixer instance
+ * @param {import('eslint').SourceCode} sourceCode The source code object
+ * @param {string} consumedName The local name that will be used in the code
+ * @param {string} exportedName The name that is exported from the module
+ * @param {string} module The module to import from
+ * @returns {import('eslint').Rule.Fix | null}
+ */
+function fixMissingImport(
```
habdelra, line 12: are we handling the case where the named import is actually renamed? this is a named export variant of the default export concern I noted above

```ts
import { CardDef as CardClass } from 'https://cardstack.com/base/card-api';

class MyCard extends CardDef {
```

in this case the import is not really missing, it's just been renamed to `CardClass` are you able to detect that?

habdelra: there is also the inverse of this:
```ts
import { Foo as CardDef } from 'foo';

class MyCard extends CardDef {
```
The CardDef import is really missing--and has been stomped on. are we able to fix this?

habdelra: (the idea is that you have to trace through the import declarations and the module scoped bindings to detect these scenarios), and when adding a declaration you may need to use a renamed import to not stomp on these eccentricities

habdelra: we actually do a lot of this from the schema editor using the ImportUtil that ed wrote. check this out. it might save you some work. it's smart enough to detect module scope binding collisions and rename accordingly https://github.com/cardstack/boxel/blob/35be8fd1111f97a78e1003fb5bd995def21ae8dc/packages/runtime-common/module-syntax.ts#L443-L449

lukemelia: As an auto-fix linting rule, this doesn't need to account for these edge cases, IMO. It is triggered only for identifiers that are not bound. So for:

```ts
import { Foo as CardDef } from 'foo';  

class MyCard extends CardDef {  
```

This is valid code and is not something this rule would get involved with.

In the case of:

```ts
import { CardDef as CardClass } from 'https://cardstack.com/base/card-api';  

class MyCard extends CardDef {  
```

THe `CardDef` identifier is unbound and since we will configure it to map to CardDef from card-api, the rule would auto-fix and the output would be:

```ts
import { CardDef as CardClass, CardDef } from 'https://cardstack.com/base/card-api';  

class MyCard extends CardDef {  
```

packages/eslint-plugin-boxel/lib/utils/import-utils.js
lukemelia: We would not try to solve a case where you have a random unbound identifier that doesn't match any of our config entries.

PR #2628, opened 2025-05-22
remove ember concurrency from setupAllRealms

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8686/white-tiles-in-prerendered-search-in-catalog

Before:
image

Result:
image


richardhjtan (approved):
Interesting, good find @lucaslyl

lucaslyl:
> Interesting, good find @lucaslyl

@richardhjtan @tintinthong 
But I realize it's still unstable. It's not just an async/await problem — it seems like prerenderCardSearch doesn't accept functions at all, whether they're asynchronous or synchronous. Even though I tried using a lot of guard clauses, it still fails. I'll try to find a better solution.

PR #2630, opened 2025-05-22
Add Listing type into host and remix command

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8704/add-listing-type-in-remix-command

### What is changing
- add Listing class type from catalog realm that able be consumed in `host`
- which strengthen the type in remix command


packages/host/app/commands/remix.ts
```diff
@@ -71,7 +74,8 @@ export class RemixCommand extends HostBaseCommand<
 
     let cardAPI = await this.loadCardAPI();
 
-    const listing = listingInput as any;
+    // this is intentionally to type because base command cannot interpret Listing type from catalog
+    const listing = listingInput as Listing;
```
richardhjtan, line 80: More details in: https://linear.app/cardstack/issue/CS-8705/catalog-listing-type-cannot-consume-in-base-folder-properly

tintinthong (approved):

PR #2633, opened 2025-05-22
Remove any 'future' events from the event list to avoid double responses

IanCal:
Repro:

please run two searches, for anything, in one message. It must be two tool calls in one message. You have to have at least something in the title in the query

Using gpt-4o in interact mode with no cards attached. You need to have two auto run commands in one message.

Previously what would happen is we'd get the command result for one of the commands, and between the bot triggering and getting the room events *the other command result* would come in. Now the code that checks all the command results are present says "yes" because it would see *both* command results. The other command result *also* triggers this check, and again both results are present so it runs as well. Now we have two going in parallel.

This fix removes anything in the event list *after* the event we're responding to.

This also fixes the problem of seeing "thinking" in the last message in the event list the bot is working with.

### DIFF
--- a/packages/ai-bot/lib/matrix/util.ts
+++ b/packages/ai-bot/lib/matrix/util.ts
@@ -147,7 +147,20 @@ function cleanupCache() {
   }
 }
 
-export async function getRoomEvents(roomId: string, client: MatrixClient) {
+/**
+ * Retrieves the last 1000 non-replace events from a Matrix room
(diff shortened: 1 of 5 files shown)


lukemelia (commented):
Did you miss adding matrix-util-test.ts?

IanCal:
> Did you miss adding matrix-util-test.ts?

derp, yes sorry.

lukemelia (approved):

PR #2634, opened 2025-05-22
Add 4.1 family and the new anthropic 4 models (their ids changed form…

IanCal:
…at :/


backspace (approved):
format change: weird!

lukemelia (approved):

PR #2636, opened 2025-05-22
Lint Autofix for unambiguous repeated imports

lukemelia:


backspace (approved):
I tried this locally by adding a spurious duplicate import and having the bot apply an unrelated suggestion, the duplicate disappeared 🎉

Is there any way to test these removals, maybe from an outside-in/API endpoint approach?

lukemelia:
Yes, there is a realm server test that can be expanded to have more coverage.

PR #2640, opened 2025-05-23
CS-8608 listing commands & refactors

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8608/listing-commands-and-refactors

### What is changing
- make install, use as commands
- introduce listing action initiation command to trigger whether to use/install/remix
- refactoring


tintinthong (approved):
Lets create a ticket to clean up catalog actions

PR #2643, opened 2025-05-23
Get card command

IanCal:


packages/host/app/commands/get-card.ts
```diff
@@ -0,0 +1,32 @@
...
+  static description = 'Get a card from the store by its ID';
+
+  async getInputType() {
+    let commandModule = await this.loadCommandModule();
+    const { GetCardInput } = commandModule;
+    return GetCardInput;
+  }
+
+  protected async run(input: BaseCommandModule.GetCardInput): Promise<CardDef> {
+    let card = await this.store.get(input.cardId);
```
lukemelia, line 25: I think @habdelra would want you to include some comments about the expected lifetime of this retrieved card.

habdelra: Absolutely. I’m a very dubious about consuming `getCard()` outside of a component. It’s really meant to be used by a component. We should really talk about the use case here. I think we might be better suited using a different store API.

habdelra: Or my bad I saw this command name and Luke’s comment and assumed it was wrapping `getCard()`, but I see that it is not. It’s still worth chatting about. But in the way we use the store api here the card returned is eligible for garbage collection and must be immediately consumed. A consumer of this command could hold on to the card in its state which would be very bad. So it needs to be clear to whoever uses this command that they must consume the returned card immediately.

IanCal: Yeah, I'm not sure what the best thing is here, I added this as part of a demo modifying a card because non-host commands have no way of accessing the store. Cards do, but commands don't.

Do we want to wrap the store in a series of commands like this? Or give access to the store to commands as well as cards?

lukemelia: I think the concept of exposing this as a command is good. Just needs docs to explain proper use I think.

habdelra: agreed

habdelra: also i think wrapping store API in commands like this is totally fine. we've already done it with the Store.patch(). this seems reasonable. i can I see us perhaps creating a command for creation or deleting too.

PR #2646, opened 2025-05-23
Synchronize monaco edits of instances with store

habdelra:
This PR synchronizes monaco edits of valid card instance JSON with the store so that we don't have to round trip thru the server in order to see monaco updates be reflected in any cards that are rendered from the store.

### DIFF
--- a/packages/host/app/components/operator-mode/code-editor.gts
+++ b/packages/host/app/components/operator-mode/code-editor.gts
@@ -9,11 +9,23 @@ import { cached, tracked } from '@glimmer/tracking';
 import { task, restartableTask, timeout, all } from 'ember-concurrency';
 
 import perform from 'ember-concurrency/helpers/perform';
+
+import isEqual from 'lodash/isEqual';
+
 import { Position } from 'monaco-editor';
 
 import { LoadingIndicator } from '@cardstack/boxel-ui/components';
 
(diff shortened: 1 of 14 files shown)


packages/host/app/components/operator-mode/code-editor.gts
```diff
@@ -305,12 +317,88 @@ export default class CodeEditor extends Component<Signature> {
...
+      ...(doc.data.attributes ? { attributes: doc.data.attributes } : {}),
+      ...(doc.data.relationships
+        ? { relationships: doc.data.relationships }
+        : {}),
+      ...(doc.data.meta.fields
+        ? { meta: { fields: doc.data.meta.fields } }
+        : {}),
+    };
+    await this.store.patch(this.args.file.url.replace(/\.json$/, ''), patch, {
+      doNotPersist: true,
```
jurgenwerk, line 398: Would it be clearer to say `{ persist: false }`?

jurgenwerk (approved):

packages/host/app/components/operator-mode/code-editor.gts
habdelra: I’m trying to be consistent with the Store.add() api. Also I think not persisting is the more seldom scenario.

richardhjtan (approved):

PR #2648, opened 2025-05-23
Update FilterList component to accommodate nesting

burieberry:
- this is now a "tree" component, but I did not rename it because it's used in various places
- accessibility and styling improvements

nested-tree


backspace:
I was able to get the failing test to pass by changing the assertion to this:

```
    assert
      .dom(`.is-selected [data-test-boxel-filter-list-button="All Cards"]`)
      .exists();
```

it’s less expressive though, I guess this is because the nesting means the style needs to persist down the tree?

richardhjtan:
Hi @burieberry, 

1) Is the PR screenshot an actual result from your machine? The child item may need a little gap after the category group label. 
By the way, I tried running these changes on my local, and it looks great.
Screenshot 2025-05-27 at 3 11 27 PM

2) Circular dependencies 
Screenshot 2025-05-27 at 3 10 17 PM
Perhaps we can introduce a shared file to be consumed by the filter list and filter list item?

burieberry:
> Hi @burieberry,
> 
> 1. Is the PR screenshot an actual result from your machine? The child item may need a little gap after the category group label.
>    By the way, I tried running these changes on my local, and it looks great.
>    Screenshot 2025-05-27 at 3 11 27 PM
> 2. Circular dependencies
>    Screenshot 2025-05-27 at 3 10 17 PM
>    Perhaps we can introduce a shared file to be consumed by the filter list and filter list item?

Yes, I did actually add the needed spacing after posting that screenshot. I'll update it. I will try to resolve the circular dependency.

burieberry:
@richardhjtan I put the components on the same file, which resolves the circular dependency between file imports. The components still embed each other, but that should be fine I think. Do you agree?

burieberry:
> I was able to get the failing test to pass by changing the assertion to this:
> 
> ```
>     assert
>       .dom(`.is-selected [data-test-boxel-filter-list-button="All Cards"]`)
>       .exists();
> ```
> 
> it’s less expressive though, I guess this is because the nesting means the style needs to persist down the tree?

Updated the data- attributes, and the tests now look like this:
```
 assert.dom(`[data-test-filter-list-item="All Cards"] > span`).hasClass('is-selected');
 assert.dom(`[data-test-selected-filter="All Cards"]`).exists({ count: 1 });
 ```

**Comment by richardhjtan** [comment]
> @richardhjtan I put the components on the same file, which resolves the circular dependency between file imports. The components still embed each other, but that should be fine I think. Do you agree?

Yes @burieberry sounds good too

File: packages/boxel-ui/addon/src/components/filter-list/index.gts [hunk 1]
```diff
@@ -1,100 +1,219 @@
...
       }
-      :global(.filter-list__icon) {
-        width: var(--boxel-icon-xs);
-        height: var(--boxel-icon-xs);
-        vertical-align: top;
+    }
+  </style>
+</template>;
+
+export default FilterList;
```
richardhjtan, line 58: Optional nitpick 🥜: would be easy to read if `export default` statement is at the bottom of the file

richardhjtan (approved):
LGTM

PR #2649, opened 2025-05-24
Fix card creation NOT returning in pre-rendered search for apps

tintinthong:
https://linear.app/cardstack/issue/CS-8660/card-creationdeletion-doesnt-update-live-search-cards

The issue is that realmURL set in the subscription was a URL and not a string

**Changes**

- add types to pre-rendered card search
-Moved pre-rendered card search types to runtime-common 
  - can be consumed by card-api
  - can be consumed by host 
- change lint errors that follow the change

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -57,7 +57,7 @@ import {
   type getCards,
   type getCardCollection,
   type Store,
-  type Query,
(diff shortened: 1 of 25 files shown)


lucaslyl (approved):
Great improvement on adding specific TypeScript.

PR #2652, opened 2025-05-26
update boxel card creation guide rc4

tintinthong:
https://gist.github.com/christse/491599c96dca6f6e14cd6adcd621ca9f


jurgenwerk:
Should this go to base skill too?

jurgenwerk (approved):
I have been using this locally for 2 days now, with more success than with the previous skill

habdelra:
skill updates are so painful to review. i wish we had a better mechanism to show the diffs....

backspace:
> skill updates are so painful to review. i wish we had a better mechanism to show the diffs....

There’s a workflow to provide helpful diffs in a comment but it [failed](https://github.com/cardstack/boxel/actions/runs/15249589456/job/42883027597#step:5:52) on this because it was too long. It should probably fail the job to be more noticeable.

PR #2660, opened 2025-05-27
Add missing lint in catalog realm

tintinthong:
This maybe can be linted on its own once catalog is phased to another repo. But bcos of not having linted the catalog properly, there is some code that was missed after a type change


habdelra (approved):
looks like a couple more lints to be fixed. but otherwise this is a good improvement

PR #2662, opened 2025-05-27
Replace skill card cache in RoomResource with the store

habdelra:


packages/host/app/resources/room.ts
```diff
@@ -356,34 +368,41 @@ export class RoomResource extends Resource<Args> {
...
-
-    let skillCard: Skill;
-    let maybeSkillCard = await this.store.get(cardId);
-    if (isCardInstance(maybeSkillCard)) {
-      skillCard = maybeSkillCard as Skill;
+    let skillCard = await this.store.get<Skill>(cardId);
+    if (isCardInstance(skillCard)) {
+      return skillCard;
     } else {
-      skillCard = await this.store.add<Skill>(cardDoc);
```
habdelra, line 376: this seems weird. when we are unable to find a skill with a particular ID we were making a new one. if there is a skill with a particular ID and you can't find it it seems like a really big problem, making a new skill is just masking over the problem

packages/host/app/resources/room.ts
```diff
@@ -356,34 +368,41 @@ export class RoomResource extends Resource<Args> {
...
-  private async createSkillCard(cardDoc: LooseSingleCardDocument) {
-    let cardId = cardDoc.data.id;
+  private async createSkillCard(doc: LooseSingleCardDocument) {
+    let cardId = doc.data.id;
     if (!cardId) {
-      console.warn(
-        `No card id found, this should not happen, this can happen if you add a skill card to a room without saving it, and without giving it an ID`,
+      throw new Error(
+        `SKill card document has no id, this should not happen: ${JSON.stringify(doc, null, 2)}`,
       );
```
habdelra, line 376: If this shouldn't happen we should throw an error so we can see that it is not happening and take steps to fix that

packages/host/app/resources/room.ts
lukemelia: This was the product of speculation. I was imagining a world where there were multiple users in a room and someone added a skill that another user didn't have access to. Not a requirement today though.

lukemelia (approved):

jurgenwerk (approved):

PR #2663, opened 2025-05-28
Fix broken auto-generate spec test

habdelra:
This PR un-comments the auto generated spec test assertions. The issue was actually around the test setup, specifically the active realms were being marked as active too late in the test setup and as such the realm matrix rooms used to broadcast realm events were never created. Feeding into this  was an unnecessary dependency on the realm-server service that made the timing of the active realm setup awkward.

### DIFF
--- a/packages/host/tests/acceptance/code-submode/field-playground-test.gts
+++ b/packages/host/tests/acceptance/code-submode/field-playground-test.gts
@@ -4,13 +4,12 @@ import { module, test } from 'qunit';
 
 import { specRef, type Realm } from '@cardstack/runtime-common';
 
-import type RealmServerService from '@cardstack/host/services/realm-server';
+import ENV from '@cardstack/host/config/environment';
 
 import {
   percySnapshot,
   setupAcceptanceTestRealm,
   setupLocalIndexing,
-  setupOnSave,
   setupUserSubscription,
   testRealmURL,
 } from '../../helpers';
@@ -33,6 +32,8 @@ import {
 } from '../../helpers/playground';
 import { setRecentFiles } from '../../helpers/recent-files-cards';
 import { setupApplicationTest } from '../../helpers/setup';
+
+const { resolvedBaseRealmURL } = ENV;
 
 const authorCard = `import { contains, field, CardDef, Component, FieldDef } from "https://cardstack.com/base/card-api";
   import MarkdownField from 'https://cardstack.com/base/markdown';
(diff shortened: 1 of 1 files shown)


packages/host/tests/acceptance/code-submode/field-playground-test.gts
```diff
@@ -998,18 +997,18 @@ module('Acceptance | code-submode | field playground', function (_hooks) {
 
   module('multiple realms', function (hooks) {
     let realm: Realm;
-    let personalRealmURL: string;
-    let additionalRealmURL: string;
+    let origin = new URL(resolvedBaseRealmURL).origin;
+    let personalRealmURL = `${origin}/testuser/personal/`;
+    let additionalRealmURL = `${origin}/testuser/aaa/`; // writeable realm that is lexically before the personal realm
```
habdelra, line 1002: if you look at where the realm-server service derives it's `url` property, for the test environment it is literally just the origin of the `resolvedBaseRealmURL` config setting for the test env. no need to use the realm server service to get this, we can simply import this from the config.

packages/host/tests/acceptance/code-submode/field-playground-test.gts
```diff
@@ -1018,13 +1017,6 @@ module('Acceptance | code-submode | field playground', function (_hooks) {
       });
       setupUserSubscription(matrixRoomId);
 
-      let realmServerService = this.owner.lookup(
-        'service:realm-server',
-      ) as RealmServerService;
-      personalRealmURL = `${realmServerService.url}testuser/personal/`;
-      additionalRealmURL = `${realmServerService.url}testuser/aaa/`; // writeable realm that is lexically before the personal realm
-      setActiveRealms([additionalRealmURL, personalRealmURL]);
```
habdelra, line 1026: it's too late to set this here. the state that this is set in has already been read and realm matrix rooms have already been created.

packages/host/tests/acceptance/code-submode/field-playground-test.gts
```diff
@@ -217,8 +218,6 @@ module('Acceptance | code-submode | field playground', function (_hooks) {
     let { setRealmPermissions, setActiveRealms, createAndJoinRoom } =
       mockMatrixUtils;
 
-    setupOnSave(hooks);
```
habdelra, line 220: this was unnecessary--we were never using this hook in these tests

backspace (approved):

PR #2672, opened 2025-05-29
Differentiate formatDateTime func and dayjsFormat helper usage

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8720/wrap-dayjsformat-so-it-does-both-date-formatting-as-a-helper-and-a

**Introducing New Date Formatting Function and Removing dayjsFormat Helper**

We're introducing a new `formatDateTime` function and eliminating the `dayjsFormat` helper to streamline our date formatting approach.

Current situation:
- We have a `dayjsFormat` helper for templates

Proposed change:
- We're introducing a new `formatDateTime` function that can be used in both templates and JS/TS code
- We will be removing the `dayjsFormat` helper completely


packages/base/Skill/boxel-development.json
lucaslyl, line 1: Remove this updated skill for now. Chris may have a better prompt to update the skill based on the information we provided.

lukemelia (changes requested):
There's no reason in this case to have two helpers that do the same thing. Templates can use plain JS functions, so let's just have one JS function and use that.

lucaslyl:
> There's no reason in this case to have two helpers that do the same thing. Templates can use plain JS functions, so let's just have one JS function and use that.

Thanks for your review and feedback! I removed the dayjsFormat template-only helper and unified everything under the formatDateTime JavaScript function

packages/runtime-common/etc/eslint/missing-invokables-config.js
```diff
@@ -8,7 +8,6 @@ module.exports = {
     array: ['array', '@ember/helper'],
     bool: ['bool', '@cardstack/boxel-ui/helpers'],
     concat: ['concat', '@ember/helper'],
-    dayjsFormat: ['dayjsFormat', '@cardstack/boxel-ui/helpers'],
```
lukemelia, line 11: Need to add `formatDateTime` here

packages/boxel-ui/addon/src/helpers/dayjs-format.ts
```diff
@@ -1,18 +1,17 @@
-import { helper } from '@ember/component/helper';
```
lukemelia, line 1: Let's rename this file to match the helper name

lukemelia (changes requested):
Update the relevant skills for this change. And when you'll merge, you'll want to notify the team in Discord since cards in people's realms may break

lucaslyl:
> Update the relevant skills for this change. And when you'll merge, you'll want to notify the team in Discord since cards in people's realms may break

sure, done update the relevant skills

lukemelia (approved):

tintinthong (approved):

PR #2673, opened 2025-05-29
Better test coverage on catalog app

richardhjtan:


tintinthong:
this change just made everything into a command-test?

richardhjtan:
> this change just made everything into a command-test?

@tintinthong yes right now it is. As currently, the use/install feature runs a command under the AI assistant, which is similar to remix now.

It's debatable whether these should be put into the commands-test, but I will suggest keeping it in `catalog-app-test` as:
- new catalog app tests will be added/amend in time being when UI/UX flow change
- commands test is quite large now (1400+ lines)

tintinthong:
Which dom selector was timing out? "Created Instances" is not a reliable state to be waiting for anymore bcos of the loader trashing

Also. I recall `waitUntil` was something to be wary of using as per comment from Ed

richardhjtan:
> Which dom selector was timing out? "Created Instances" is not a reliable state to be waiting for anymore bcos of the loader trashing
> 
> Also. I recall `waitUntil` was something to be wary of using as per comment from Ed

At the very beginning of waiting both install/use button exist

**Errors on the flaky tests**
Error: waitFor timed out waiting for selector "[data-test-catalog-listing-use-button]"
Error: waitFor timed out waiting for selector "[data-test-catalog-listing-install-button]"

We don't have "Created Instances" instance since we introduced use/install running in AI assistant, we only have loading state on use/install when initialling the room chat

tintinthong:
> [data-test-catalog-listing-use-button]



> data-test-catalog-listing-use-button

Based upon the test results that buck pasted, I think it originally was for the `waitUntil` right 

Screenshot 2025-05-29 at 11 27 13

Whats the change here, if we already don't have "Created Instances". The current test is completely non-applicable and test has to be rewritten. Is that correct?

tintinthong:
Based upon discussion with @richardhjtan, 

Few notes 
- not fixing flakiness, rather structuring better test coverage for catalog 
- click catalog button, check ai assistant room settled, check ai assistant room message prompt sent, check room event fired, read room event and create instance of command based upon data from room event and execute (since no ai assistant response; not handle other messages necessarily), check existence of files by navigating to 2nd realm

tintinthong (approved):

PR #2674, opened 2025-05-29
Fix: Cannot remove auto-attached card in code mode AI Assistant

FadhlanR:
This PR fixes an issue where auto-attached cards in code submode could not be removed. The problem was caused by the fact that auto-attached cards in code submode were not retrieved from the `autoAttachmentResource`, so when their IDs were added to `removeCards`, they were not processed correctly.


packages/host/app/resources/auto-attached-card.ts
```diff
@@ -82,13 +110,19 @@ export class AutoAttachment extends Resource<Args> {
 export function getAutoAttachment(
   parent: object,
   args: {
+    submode: () => Submode;
```
habdelra, line 115: if this is a thunk that means this resource responds to the user switching submodes. which means that you should add a test where there is an attachment made in code mode and the user switches to interact mode and vice versa, and you should confirm that the behavior is consistent when the submode changes.

packages/host/app/resources/auto-attached-card.ts
```diff
@@ -82,13 +110,19 @@ export class AutoAttachment extends Resource<Args> {
...
+    autoAttachedFileUrl: () => string | undefined;
+    playgroundPanelCardId: () => string | undefined;
```
habdelra, line 117: why is this treated specially? can't we just collapse values here into the `attachedCardIds` thunk?

packages/host/app/resources/auto-attached-card.ts
```diff
@@ -51,29 +58,50 @@ export class AutoAttachment extends Resource<Args> {
...
+          let card = await this.store.get(item.id);
+          if (card && isCardInstance(card)) {
+            let realmURL = card[realmURLSymbol];
+            if (realmURL && item.id === `${realmURL.href}index`) {
+              continue;
+            }
+          }
+          this.cardIds.add(item.id);
+        }
+      } else if (submode === Submodes.Code) {
```
habdelra, line 89: i'm questioning why you actually even care about the submode. can you not collapse these two different branches of the `if` into a single set of operations that works in either mode? like `autoAttachedFileUrl` should _always_ get the `.json` suffix treatment regardless of which submode you are in right? and why is  `playgroundPanelCardId` different from `attachedCardIds`, what makes its treatment so special, its just a card ID you want to attach, right?

FadhlanR:
@habdelra here's the scenario:

* In **interact** submode, the auto-attached card is the top-most item in the stack.
* In **code** submode, the auto-attached card is either the playground panel card or the card of the currently opened file — if the file has a `.json` extension.
* `attachedCardIds` stores the IDs of cards manually added by the user through the card chooser.

So, we need to check against `attachedCardIds` to determine whether to label the playground panel card, the opened file card, or the top-most stack item as an auto-attached card — but only if they are **not** already included in `attachedCardIds`.

I've added a test to cover these scenarios.

jurgenwerk (approved):

PR #2677, opened 2025-05-29
ci: Add path-based filtering to main workflow

backspace:
The main goal of this is to ensure that changes in `packages/ai-bot` don’t cause the tests and Percy jobs for `packages/host` to run, which are sadly flaky and slow. But I added path-based filtering to many of the jobs in the main workflow so PRs that focus on a particular package don’t produce unrelated Actions jobs.

I originally tried moving the `ai-bot` job into its own workflow but the `deploy` job checks its passing status, which wouldn’t work across workflows.

These are the jobs that ran and didn’t when I made a fake `ai-bot` change (and removed `.github/workflows/ci.yml` from the shared path filter):

boxel 2025-05-29 10-49-09


.github/workflows/ci.yaml
```diff
@@ -338,32 +425,6 @@ jobs:
...
-              - '.github/workflows/ci.yaml'
-              - 'packages/ai-bot/**'
-              - 'packages/base/**'
-              - 'packages/boxel-ui/**'
-              - 'packages/boxel-icons/**'
-              - 'packages/host/**'
-              - 'packages/realm-server/**'
-              - 'packages/runtime-common/**'
-              - 'pnpm-lock.yaml'
-
```
backspace, line 366: Since so many of the jobs now depend on this, I moved it to the top.

.github/workflows/preview-host.yml
```diff
@@ -1,4 +1,4 @@
-name: CI Preview Host
```
backspace, line 1: I renamed the file and workflow for clarity.

lukemelia (approved):

jurgenwerk (approved):

PR #2678, opened 2025-05-29
Refactor playground card creation error clearing

habdelra:
This PR refactors playground card creation error clearing so that it doesn't use side effects to clear errors. Rather through the use of local IDs that the host generates we can use the inherent ability of the store to clear error states when it sees a successful instance creation that is correlated to an error state via local ID.


packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -118,14 +119,15 @@ export default class PlaygroundPanel extends Component<Signature> {
...
+  // we rely on a unique stable local id per successful new card creation.
+  // this allows us to correlate on failed new card creation errors from the server
+  #newCardLocalId = uuidv4();
 
   private get moduleId() {
     return internalKeyFor(this.args.codeRef, undefined);
   }
 
   private get isLoading() {
-    this.clearCardCreationError();
```
habdelra, line 220: this is the side effect that we are eliminating

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -436,9 +435,14 @@ export default class PlaygroundPanel extends Component<Signature> {
...
-      this.cardCreationError = maybeId;
-    } else {
+    if (typeof maybeId === 'string') {
+      // reset the local ID for making new cards after each successful attempt
+      // such that failed attempts will use the same local ID. that way when we
+      // get an error doc in the store for a particular local id based on the
+      // server error response, when a new card is successfully created that
+      // uses a correlating local ID we will automatically clear the error in
+      // the store.
+      this.#newCardLocalId = uuidv4();
```
habdelra: FYI

burieberry:
Your changes look good, however when running it locally, I noticed that serialization error occurring during new card/field creation is no longer surfacing on playground as an error doc (although I see the error in browser console). I thought there were tests for these cases, but maybe they're no longer there.

Using `boom-pet.gts` file, clicking on `BoomPet` card def and opening playground panel:
card-serialization-error

Selecting `FailingField` field def on the same file:
field-serialization-error

habdelra:
> Your changes look good, however when running it locally, I noticed that serialization error occurring during new card/field creation is no longer surfacing on playground as an error doc (although I see the error in browser console). I thought there were tests for these cases, but maybe they're no longer there.
> 
> Using `boom-pet.gts` file, clicking on `BoomPet` card def and opening playground panel: card-serialization-error
> 
> Selecting `FailingField` field def on the same file: field-serialization-error

thanks for checking it out, i'll make sure we handle this scenario

habdelra:
@burieberry these issues should be fixed now. I updated the tests to use the specific scenarios that you outlined above.

burieberry (approved):

PR #2680, opened 2025-05-29
Update host to use content-tag 3.1.3 (with a parsing bugfix)

lukemelia:
- locked ember-basic-dropdown to 8.0.4 so that our patch would still apply


backspace (approved):
I’m just ignoring the weird Percy situation 😖

PR #2681, opened 2025-05-29
CardsGrid refactor

burieberry:
- CardGridLayout and CardList components (currently in base, but will probably replace the similar ones in catalog)
- SortDropdown component (in boxel-ui). Maybe this should also live in Catalog, but FilterList component was in boxel-ui, so I put the SortDropdown here as well.
- Currently the change for filters on index cards grid is now card-type filters are nested under the parent All Cards filter


lukemelia (approved):
Nice & clean!

PR #2683, opened 2025-05-30
Unify the width of all markers of the search/replace block

jurgenwerk:
Fixing a mistake in the name of consistency — the divider marker should be equal in width to the others.

Before:
```
╔═══ SEARCH ════╗
╠══════════════╣
╚═══ REPLACE ═══╝
```

After:

```
╔═══ SEARCH ════╗
╠═══════════════╣
╚═══ REPLACE ═══╝
```


IanCal:
I'm seeing it appear in responses: 
image

jurgenwerk:
@IanCal is this an old session or maybe a case of an outdated skill? I just tested this again and works OK for me

IanCal:
I'm concerned this is really sensitive to the number of = 

What was happening before when we used the previous format?

Was the extra ====== appearing just because it had *more* than we expected? Counting them in the PR it would match if the model just made the == line the same length as the search/replace lines.

jurgenwerk:
> Was the extra ====== appearing just because it had more than we expected? 

No. The extra `======` appeared after the actual divider (i.e. the first `======`), there was always some code in between.

IanCal:
> @IanCal is this an old session or maybe a case of an outdated skill? I just tested this again and works OK for me

Possibly, though this must have come from at least the old skill, you didn't change the replace part.

IanCal:
> > Was the extra ====== appearing just because it had more than we expected?
> 
> No. The extra `======` appeared after the actual divider (i.e. the first `======`), there was some code in between.

Ah it was randomly in the replace part? So something like

```
<<<< search

things

=====

replacement stuff

=====

more replacement stuff

>>>>> replace

```

IanCal (approved):

jurgenwerk:
Yep, that is correct. Sometimes it was also:

```
<<<< search

things

=====

replacement stuff

=====                               <----- the extra
>>>>> replace
```

PR #2685, opened 2025-06-02
Improve loading time when opening AI panel

FadhlanR:
Previously, all messages—including those generated during the streaming process—were retrieved when the user opened the AI panel. For a single final message, this could result in hundreds of intermediate streaming messages, significantly slowing down the loading time. In this PR, I introduce a new message event type, 'app.boxel.message.streaming', which is now filtered out when loading the message history. The streaming process itself remains unchanged; only the loading behavior when opening the AI panel is affected.


packages/ai-bot/lib/matrix/util.ts
```diff
@@ -87,7 +88,9 @@ export async function sendMessageEvent(
   return await sendMatrixEvent(
     client,
     roomId,
-    'm.room.message',
+    data.isStreamingFinished
+      ? 'm.room.message'
+      : APP_BOXEL_MESSAGE_STREAMING_EVENT_TYPE,
```
lukemelia, line 104: Fine for now, through it occurs to me that we may be able to eliminate isStreamingFinished as a property of the event now that we have a dedicated event type for streaming.

lukemelia (approved):

packages/ai-bot/lib/matrix/util.ts
FadhlanR: I'll submit another PR for that.

PR #2694, opened 2025-06-03
Trigger loginToRealms from index

FadhlanR:


FadhlanR:
I'll merge this PR directly, since this is a small change and still part of the loading time improvement when user logs in.

PR #2697, opened 2025-06-03
add skill listing test

tintinthong:


packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -26,87 +26,6 @@ import { setupApplicationTest } from '../helpers/setup';
 const catalogRealmURL = 'http://localhost:4201/catalog/';
 const testRealm2URL = `http://test-realm/test2/`;
```
tintinthong, line 28: Catalog is available. So lets not create a mock listing card source

packages/base/spec.gts
```diff
@@ -43,7 +43,7 @@ import Brain from '@cardstack/boxel-icons/brain';
 import { use, resource } from 'ember-resources';
 import { TrackedObject } from 'tracked-built-ins';
 
-export type SpecType = 'card' | 'field' | 'component' | 'app' | 'skill';
+export type SpecType = 'card' | 'field' | 'component' | 'app';
```
tintinthong, line 45: Ths idea of  a skill as a spec; doesn't really make so much sense. Spec is truer to code than instance. We already handle mangling of instances inside of a listing. This makes it clearer that a spec only pertains to code-related things -- it is unusual that a Skill which reference a single Skill base is indeed a spec; We will only ever have one spec with one specType ='Skill'. 

So after talking to Chris we agreed to remove this

packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -295,6 +213,46 @@ module('Acceptance | catalog app tests', function (hooks) {
     assert.strictEqual(message.content.body, expectedMessage);
   }
 
+  async function verifyInstanceExists(
```
richardhjtan, line 216: Nice, i will see any place i can clean up by using this helper after this PR get merge

richardhjtan (approved):
Looks good to merge after the lint and host test are fixed

PR #2699, opened 2025-06-03
Fix issue where instanceOf() did not work thru an interior field

habdelra:


packages/base/card-api.gts
```diff
@@ -243,9 +244,7 @@ export function instanceOf(instance: BaseDef, clazz: typeof BaseDef): boolean {
     if (isEqual(codeRefInstance, codeRefClazz)) {
       return true;
     }
-    instanceClazz = instanceClazz
-      ? (Reflect.getPrototypeOf(instanceClazz) as typeof BaseDef | null)
-      : null;
+    instanceClazz = instanceClazz ? getAncestor(instanceClazz) : null;
```
habdelra: it's important to use this `getAncestor()` function as opposed to the prototype directly, as this function will set the identity for interior cards/fields.

lukemelia (approved):

PR #2706, opened 2025-06-03
Can generate sample data with AI (playground preview)

burieberry:
- Refactored playground dropdown menu options using boxel-ui Menu component
- Added new menu items for filling in sample data and generating examples on instance-chooser dropdown and on the '...' menu in playground
- Implementation uses the `SendAiAssistantMessageCommand` to submit these requests from playground panel


lukemelia (approved):

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -153,24 +174,75 @@ export default class PlaygroundPanel extends Component<Signature> {
   }
 
   private get contextMenuItems() {
-    if (!this.card?.id) {
-      return undefined;
-    }
-    let cardId = this.card.id;
+    let cardId = this.cardResource?.id;
```
habdelra: be aware that there is a subtle difference in the change here. previously we used `this.card?.id` which is actually the remote id (aka URL). now we are using the cardResource?.id, which can be either the remote ID or the local ID. the store can handle either remote or local ID interchangably, but other systems (like the AI assistant) really only want the remote ID. there is also an `isLocalId()` function from runtime-common that you can use if this distinction is important for you.

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -153,24 +174,75 @@ export default class PlaygroundPanel extends Component<Signature> {
...
       new MenuItem('Copy Card URL', 'action', {
         action: () => this.copyToClipboard.perform(cardId),
         icon: IconLink,
+        disabled: !cardId,
       }),
       new MenuItem('Open in Code Mode', 'action', {
         action: () =>
-          this.operatorModeStateService.updateCodePath(new URL(cardId)),
+          this.operatorModeStateService.updateCodePath(
+            cardId ? new URL(cardId) : null,
```
habdelra, line 187: if `cardId` is a local id this will throw

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -153,24 +174,75 @@ export default class PlaygroundPanel extends Component<Signature> {
...
-    if (!this.card?.id) {
-      return undefined;
-    }
-    let cardId = this.card.id;
+    let cardId = this.cardResource?.id;
     let menuItems: MenuItem[] = [
       new MenuItem('Copy Card URL', 'action', {
         action: () => this.copyToClipboard.perform(cardId),
         icon: IconLink,
+        disabled: !cardId,
```
habdelra, line 182: you probably don't want to enable this if cardId is a local ID.

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
habdelra: additionally, if a card was created (as opposed to retrieved from the server), the `cardResource.id` will continue to remain a local id even after a remote ID has been assigned (for stability so we don't trigger rerenders unnecessarily). You may wish to make a computed that can return the remote ID for a given local ID, something like:
```ts
 get url() {
   if (!isLocalId(cardResource.id)) {
     return cardResource.id
   } else {
     let instance = store.peek(cardResource.id);
     return instance.id;
   }
 }
```

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -153,24 +174,75 @@ export default class PlaygroundPanel extends Component<Signature> {
...
+          this.operatorModeStateService.updateCodePath(
+            cardId ? new URL(cardId) : null,
+          ),
         icon: IconCode,
+        disabled: !cardId,
       }),
       new MenuItem('Open in Interact Mode', 'action', {
         action: () => this.openInInteractMode(cardId),
         icon: Eye,
+        disabled: !cardId,
```
habdelra, line 195: this you truly want to work for local ids and remote ids alike

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -191,11 +263,13 @@ export default class PlaygroundPanel extends Component<Signature> {
   }
 
   private get canEditCard() {
-    return Boolean(
-      this.format !== 'edit' &&
-        this.card?.id &&
-        this.realm.canWrite(this.card.id),
-    );
+    // in error case, we may not have `this.card?.id`, but we can get the id from `this.cardResource?.id`
+    let cardId = this.cardResource?.id;
```
habdelra: ditto on the remote vs local ID subtlety that was introduced on this line as above

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -191,11 +263,13 @@ export default class PlaygroundPanel extends Component<Signature> {
...
+    return cardId && this.realm.canWrite(cardId);
```
habdelra: the realm service is not going to know how to deal with a local ID. however, the instance does have a realm URL meta field. a more reliable solution here would be to pass that realm url meta field value for the instance instead of assuming you are dealing with a remote id.

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -678,6 +755,57 @@ export default class PlaygroundPanel extends Component<Signature> {
...
+          // otherwise it complains about not having an open card or
+          // generates instances for all card definitions on the file
+          prompt += ` for the selected module ${this.moduleId} in the attached file.`;
+        }
+      }
+
+      await sendMessageCommand.execute({
+        roomId: this.matrixService.currentRoomId,
+        prompt,
+        openCardIds: card?.id ? [card.id] : undefined,
```
habdelra, line 798: these are remote ID's which i think is correct in this context

habdelra (changes requested):
in general i would strongly suggest you revisit all the places where you are assuming that you are dealing with a remote ID, because in most cases this is an incorrect assumption.

burieberry:
> in general i would strongly suggest you revisit all the places where you are assuming that you are dealing with a remote ID, because in most cases this is an incorrect assumption.

Thank you for pointing out this distinction. I do not need the local id here, so I reverted those changes.

packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -153,24 +174,75 @@ export default class PlaygroundPanel extends Component<Signature> {
   }
 
   private get contextMenuItems() {
-    if (!this.card?.id) {
-      return undefined;
-    }
```
habdelra, line 158: you'll still want the logic to return undefined if there is no card ID

burieberry: The AI items don't need the card id. The moduleId is sufficient. I decided to disable the items that need card id. Another option is hiding them.

habdelra (approved):

PR #2708, opened 2025-06-04
Improve loading time when opening ai panel

FadhlanR:
**This PR is a follow-up to:** [https://github.com/cardstack/boxel/pull/2695](https://github.com/cardstack/boxel/pull/2695).

There was an issue that @IanCal encountered where the AI bot couldn't see its own messages, causing it to send another command request after the user had already applied the original one. This happened because our previous solution introduced a new event type, `'app.boxel.message.streaming'`, for streaming events, while `'m.room.message'` was used only for the final message. However, the final message couldn't replace the streaming message because they had different event types.

Instead of introducing a new event type, this PR filters out events with `'m.replace'`, following the same approach we use in the AI bot. I had to patch the `matrix-js-sdk` since it doesn't yet support that filter. The updates are now only on the host side, so there should be no side effects on the AI bot side.


jurgenwerk:
Do you have any stats about how much of a performance improvement this brings, especially in large sessions?

FadhlanR:
@jurgenwerk I don't have the exact number, but here's a screenshot showing the number of fetch message requests for the same room before and after this solution was implemented.

Before:
Screenshot 2025-06-05 at 14 29 33

After:
Screenshot 2025-06-05 at 14 30 56

jurgenwerk (approved):

lukemelia (approved):

PR #2709, opened 2025-06-04
Don't crash the assistant if it can't load skills

jurgenwerk:
This is a hot fix for situations where old matrix sessions are referencing skills that do not exist anymore (deleted or renamed). In that case the unhandled error results in an unfunctioning AI assistant (can't send messages). 

image

This fix enables users to continue using old sessions with broken skill references but they need to start a new session for the updated skills to be picked up. I made a new ticket for handling outdated skills (renamed/deleted) and make sure they get updated in the matrix config.

### DIFF
--- a/packages/host/app/resources/room.ts
+++ b/packages/host/app/resources/room.ts
@@ -133,6 +133,10 @@ export class RoomResource extends Resource<Args> {
       if (!memberIds || !memberIds.includes(this.matrixService.aiBotUserId)) {
         return;
       }
+      // TODO: enabledSkillCards can have references to skills whose URL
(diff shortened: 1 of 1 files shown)


IanCal (approved):

packages/host/app/resources/room.ts
```diff
@@ -396,7 +400,8 @@ export class RoomResource extends Resource<Args> {
     if (isCardInstance(skillCard)) {
       return skillCard;
     } else {
-      throw new Error(`Cannot find skill ${cardId}`);
+      // A known reason for this is that the skill has been renamed
+      return undefined;
```
habdelra, line 404: Probably we should console log the issue, as opposed to silently doing nothing.

PR #2711, opened 2025-06-04
Dont rely on card ID when determining if card is in writable realm

habdelra:
This solves a race condition where we were relying on a card's remote ID in order to determine if the card lives in a writable realm. because cards are optimistically created, we can't be sure that the card has a remote ID assigned yet. A more reliable way to determine the realm of a card is to use the `realmURL` symbol on the running instance. instances that have not yet been assigned a remote ID will have a `realmURL` symbol specified with the realm that the newly created card will live within.

### DIFF
--- a/packages/host/app/components/operator-mode/stack-item.gts
+++ b/packages/host/app/components/operator-mode/stack-item.gts
@@ -494,9 +494,10 @@ export default class OperatorModeStackItem extends Component<Signature> {
   private get canEdit() {
     return (
       this.card &&
+      this.card[realmURL] &&
       !this.isBuried &&
       !this.isEditing &&
-      this.realm.canWrite(this.card.id)
+      this.realm.canWrite(this.card[realmURL].href)
     );
   }
 


lukemelia (approved):

tintinthong (approved):
Tested. Card creation is real smooth now 👍

tintinthong:
Im noticing some behaviour that linking and creating a new blog app to a new created blog post is having Failed to Save error. But this might be due to blog app itself as I can't really re-create it anywhere else. It is also difficult to test this particular bug that I observe on main without this PR, otherwise, I run into the issue here. 

https://linear.app/cardstack/issue/CS-8812/new-blog-post-failing-to-save-when-linking-new-blog-app-to-it

habdelra:
> Im noticing some behaviour that linking and creating a new blog app to a new created blog post is having Failed to Save error. But this might be due to blog app itself as I can't really re-create it anywhere else. It is also difficult to test this particular bug that I observe on main without this PR, otherwise, I run into the issue here.
> 
> https://linear.app/cardstack/issue/CS-8812/new-blog-post-failing-to-save-when-linking-new-blog-app-to-it

yeah this issue looks different. specifically something happened during indexing where we are getting a not loaded error.

PR #2712, opened 2025-06-04
Defend against observed condition where command result was sent to the matrix room twice.

lukemelia:


IanCal (approved):

IanCal:
LGTM, I might spend a bit of time seeing if I can trigger this or find another path through that causes it, I don't see any harm in having this check here.

PR #2714, opened 2025-06-04
Fix omitted fieldIndex storage for playground preview

burieberry:
For playground local storage, we know if we have a field or card based on the stored `fieldIndex` property. For cards, this is not set or is explicitly `undefined`. For fields, we do not have card Ids, so we store the spec's cardId with `fieldIndex` that corresponds to the field example's index on spec's `containedExamples`.

There was a mixup and `persistToLocalStorage` was used instead of `persistSelections`. This is my bad for not writing a comment and making it clearer. The tests did catch it. `persistSelections` ensures the right fieldIndex is set, as well as checking for currently stored selection and preventing some of the unnecessary screen flashes. `persistToLocalStorage` was created to circumvent infinite-revalidation error for a specific legitimate case. This is no longer a problem, thanks to the 
code-mode accordion to buttons refactor. I was able to combine the two functions into one. Now there's only one `persistSelections` method.

I added more comments in the code so hopefully that helps explain the `fieldIndex`.

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
+++ b/packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
@@ -179,7 +179,7 @@ export default class PlaygroundPanel extends Component<Signature> {
     if (!this.card?.id) {
       return;
     }
-    this.persistToLocalStorage(this.card.id, format);
+    this.persistSelections(this.card.id, format);
   }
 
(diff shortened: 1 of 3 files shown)


backspace (approved):
Thanks for figuring this out and adding explanations 🤩

habdelra (approved):

PR #2717, opened 2025-06-05
Add schemas for query and coderef

IanCal:


lukemelia (approved):
Good catch. We need test failures. Can we throw in case of a missing mapping?

IanCal:
> Good catch. We need test failures. Can we throw in case of a missing mapping?

We can, I think we want to be gentler for some cases though (otherwise you'd not be able to patch a card with some unknown type). We should have something that throws for known inputs with host commands though. Maybe there's a strict argument or similar to pass around.

PR #2718, opened 2025-06-05
Avoid putting the system prompt between tool call and assitant

IanCal:


tintinthong (approved):
This works as I expect it.  And https://linear.app/cardstack/issue/CS-8777/remix-a-skill-has-provider-return-error#comment-770069e9 is fixed. 

I still need follow-up changes but that will be handled in another ticket

lukemelia (approved):
Is the test coverage here adequate to protect against regrressions?

IanCal:
> Is the test coverage here adequate to protect against regrressions?

I'll add something more. The tests will fail if we do this but I think the one that reliably checks isn't focused on this, and likely doesn't check each path.

PR #2719, opened 2025-06-05
Can add a link in a null links to many field

habdelra:
This fixes an issue where if a links to many relationship is set to null the add link button will still work.

### DIFF
--- a/packages/host/tests/integration/components/operator-mode-test.gts
+++ b/packages/host/tests/integration/components/operator-mode-test.gts
@@ -469,6 +469,24 @@ module('Integration | operator-mode', function (hooks) {
             }),
             pet: petMango,
           }),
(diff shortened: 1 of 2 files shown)


packages/host/tests/integration/components/operator-mode-test.gts
```diff
@@ -469,6 +469,24 @@ module('Integration | operator-mode', function (hooks) {
...
             pet: petMango,
           }),
+          'Person/hassan.json': {
+            data: {
+              attributes: {
+                firstName: 'Hassan',
+              },
+              relationships: {
+                friends: {
+                  links: { self: null },
```
habdelra, line 479: the links-to-many editor would previously blow up when trying to add a link in this structure

lukemelia (approved):

PR #2720, opened 2025-06-05
Add realm session cache to worker

habdelra:
Currently each worker performs a realm auth for each job it picks up, so that means each index mutation requires realm auth. we currently cache the matrix session in the worker, but we are not caching the realm session. This should allow for less matrix activity as a result of index mutations.


lukemelia (approved):
Nice!

backspace (approved):
The test run I linked to when opening this issue had 11 401s and this only has 5, great improvement 🎉

PR #2722, opened 2025-06-06
Follow-up to agentic command execution work

lukemelia:
The host's MatrixService generates a random "agentId" on instantiation. It includes this agent ID in the data.context.agentId node of message and result events sent to Matrix. The aibot captures this agentId when it responds to a matrix event and includes it in its own data.context.agentId node in all matrix events that it publishes, including messages with command requests and/or code patches. When the host's command service considers whether to automatically execute a command or automatically apply a patch, it will only do so if the agentId matches the matrix service's agentId. This should limit the agentic eagerness to one and only one tab while allowing multiple tabs to have their server state stay up-to-date via matrix broadcasts.

### DIFF
--- a/packages/ai-bot/helpers.ts
+++ b/packages/ai-bot/helpers.ts
@@ -25,7 +25,6 @@ import { logger } from '@cardstack/runtime-common';
 import {
   APP_BOXEL_ACTIVE_LLM,
   APP_BOXEL_CODE_PATCH_RESULT_REL_TYPE,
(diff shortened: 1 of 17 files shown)


IanCal:
Looking good, testing manually this works, am going to try a run through a remix. Have hopefully fixed one of the tests and will do the matrix one now, they were failing because the fake events we put in didn't have the agentId so the commands were not auto running.

habdelra (approved):

PR #2724, opened 2025-06-06
add sanitization of realmUrl that is executed by AI when running actions of listing

tintinthong:
The issue is that the AI constantly return thrown errors bcos realmUrl didn't have a trailing slash when attempting to run the remix command 

This fix uses RealmPaths abstraction to ensure the trailing slash is there

### DIFF
--- a/packages/host/tests/acceptance/catalog-app-test.gts
+++ b/packages/host/tests/acceptance/catalog-app-test.gts
@@ -571,5 +571,58 @@ module('Acceptance | catalog app tests', function (hooks) {
         .dom('[data-test-playground-panel] [data-test-boxel-card-header-title]')
         .hasText('Author');
     });
+
+    test('use is succesful even if target realm does not have a trailing slash', async function (assert) {
+      await visitOperatorMode({
+        stacks: [[]],
+      });
+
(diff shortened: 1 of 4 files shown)


habdelra (approved):
awesome--that is literally the reason that we made the `RealmPaths` abstraction was to deal with trailing slashes consistently in one place. i find that every time i don't use `RealmPaths` to do directory math i eventually regret it :smile:

packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -571,5 +571,58 @@ module('Acceptance | catalog app tests', function (hooks) {
         .dom('[data-test-playground-panel] [data-test-boxel-card-header-title]')
         .hasText('Author');
     });
+
+    test('use is succesful even if target realm does not have a trailing slash', async function (assert) {
```
backspace:
```suggestion
    test('use is successful even if target realm does not have a trailing slash', async function (assert) {
```

backspace (approved):

richardhjtan (approved):
👍

jurgenwerk (approved):

PR #2731, opened 2025-06-09
Ignore broken links when performing file serialization

habdelra:
The field option `{ isUsed: true }` will throw when it encounters a broken link during recompute--which is what we expect it to do. instances are recomputed after they are deserialized, however, there are scenarios where we really don't care about broken links as part of deserialization. Specifically when serializing an instance to the file system. Also when indexing, immediately before we begin the rendering process, we don't actually care about broken links yet. (the rendering process actually is what we use to deal with broken links).

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -42,6 +42,7 @@ import {
   meta,
   baseRef,
   getAncestor,
+  isCardError,
   type Format,
   type Meta,
   type CardFields,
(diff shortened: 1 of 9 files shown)


backspace (approved):

tintinthong (approved):
EDIT: I see issues. There were times it was succesful but I also see it unsuccesful. Looks like its failing on blog app side tht wants to rely on its existence of its own id via `setFilters`. Seems like a card-related issue. We can handle it separately. 


https://github.com/user-attachments/assets/26548d23-bbf1-44dd-9565-e8a36cf586d7

habdelra:
> EDIT: I see issues. There were times it was succesful but I also see it unsuccesful. Looks like its failing on blog app side tht wants to rely on its existence of its own id via `setFilters`. Seems like a card-related issue. We can handle it separately.
> 
>  test-blog-app-compressed.mp4

yeah this is a different problem. a card instance doesn't always have an ID. this card logic needs to be updated to reflect that fact.

PR #2732, opened 2025-06-09
host: Change operator mode state to private

backspace:
While there are currently no instances of external modules mutating operator mode state, this enforces that in the type system.

host 2025-06-09 13-24-43


habdelra (approved):
Love it!

PR #2733, opened 2025-06-09
host: Change updateCodePath to be redirect-aware

backspace:
(The failing realm server test is addressed in #2758. The Percy diff is spurious, of course.)

Currently the operator mode state service uses a [file resource `onRedirect` callback](https://github.com/cardstack/boxel/blob/f8a2f50f73391809c3e5fb612b973316b2eaa234/packages/host/app/services/operator-mode-state-service.ts#L762) to `replaceCodePath` when `updateCodePath` is called with a non-canonical URL, which is causing problems in #2704, because the module inspector view will be stored under disparate keys when modules are passed in without file extensions. This updates it to detect redirects and only actually update the code path once.

This necessitated making `updateCodePath` asynchronous, so most of the changes are just that cascading through various callbacks and enclosing functions.

Navigating to a parent in the schema used to involve two URL changes (delay introduced deliberately for visibility):

screencast 2025-06-09 14-47-26

With this, there’s no intermediate URL:

screencast 2025-06-09 15-04-46


packages/matrix/tests/registration-with-token.spec.ts
```diff
@@ -174,13 +174,6 @@ test.describe('User Registration w/ Token - isolated realm server', () => {
     let newRealmURL = new URL('user1/personal/', serverIndexUrl).href;
     await enterWorkspace(page, "Test User's Workspace");
 
-    // assert back button brings you back to workspace chooser
-    await page.goBack();
-    await expect(
-      page.locator(`[data-test-workspace="Test User's Workspace"]`),
-    ).toHaveCount(1);
-    await enterWorkspace(page, "Test User's Workspace");
```
backspace, line 182: This was failing in CI _only_, you can see a trace [here](https://boxel-matrix-playwright-reports.stack.cards/host/updateCodePath-redirects-cs-8833/20250611_185220Z/trace/index.html?trace=https://boxel-matrix-playwright-reports.stack.cards/host/updateCodePath-redirects-cs-8833/20250611_185220Z/data/97a5868a67e194d4354bd1085c1596ba062ad654.zip) (while it lasts): pressing the back button was not returning to the workspace chooser. Who knows the underlying cause, but I added a separate test that specifically exercises the back button, and it works in CI 🙃

lukemelia (approved):

PR #2734, opened 2025-06-09
Remix should open skill in preview

tintinthong:


packages/host/app/commands/listing-remix.ts
```diff
@@ -111,6 +111,15 @@ export default class RemixCommand extends HostBaseCommand<
...
       });
+    } else if ('skills' in listing) {
+      // A listing can have more than one skill
+      // The most optimum way for remixing is still to display only the first instance
+      let firstSkillCardId =
+        skillCardIds && skillCardIds.length > 0 ? skillCardIds[0] : undefined;
+      await new SwitchSubmodeCommand(this.commandContext).execute({
+        submode: 'code',
+        codePath: firstSkillCardId,
+      });
```
richardhjtan: @tintinthong do we need to handle if there is no skill card attached (undefined scenario)?

richardhjtan (approved):

PR #2736, opened 2025-06-10
Cs 8823 createcard not working prerendered cards search not returning live update

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8823/createcard-not-working-prerendered-cards-search-not-returning-new

After investigating the related file in the catalog-app, I can see that we omitted the @isLive argument in the PrerenderedCardSearch component, which is why we weren't getting live updates. Just adding this argument will fix the problem.

### DIFF
--- a/packages/catalog-realm/blog-app/components/card-list.gts
+++ b/packages/catalog-realm/blog-app/components/card-list.gts
@@ -31,6 +31,7 @@ export class CardList extends GlimmerComponent<CardListSignature> {
           @query={{@query}}
           @format='embedded'
           @realms={{@realms}}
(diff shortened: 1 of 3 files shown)


tintinthong:
Oh when was this removed? Who removed it? Was there a reason

lucaslyl:
> Oh when was this removed? Who removed it? Was there a reason

The file in catalog-realms existed before Hassan's changes( Prerenderedsearch @islive ...), and I guess he inadvertently omitted applying the @isLive argument to this file during his updates

richardhjtan (approved):

PR #2737, opened 2025-06-10
Cs 8835 sometimes when we click on dropdown from remix it still remains opened

lucaslyl:
linear:https://linear.app/cardstack/issue/CS-8835/sometimes-when-we-click-on-dropdown-from-remix-it-still-remains-opened

-----
The default BoxelDropdown component behavior - it will always need to be manually toggled to close the dropdown menu. If not, it will remain floating there even when we scroll.

Solution:
Add a new autoClose argument to the BoxelDropdown component that allows controlling whether the dropdown should automatically close when the mouse leaves the content area.

When autoClose is true (default), the dropdown will close on mouseleave. When set to False, the dropdown will remain open until explicitly closed.

This change improves the component's flexibility for different use cases where persistent dropdown menus are needed.

Result:
https://www.loom.com/share/c0d93317a2c0469e8a8b61e8042a09cf?sid=8c1eb50a-a9a6-47ac-b819-3a6b8ca49e71
https://www.loom.com/share/4113b618754e4611b815ed23b0ec3514?sid=3e6cc4e4-6c7e-4033-bb41-f20eb8ed0036


backspace:
With the nature of the failing test assertions, it seems like this changed something that either requires new assertions or a fix to the dropdown code.

lucaslyl:
> With the nature of the failing test assertions, it seems like this changed something that either requires new assertions or a fix to the dropdown code.

Yes, you're right, I refactor the dropdown component to maintain test compatibility while keeping the pointer-based detection. Thank you for the advice!

tintinthong:
Is this reproducible inside of a test? We dont seem to have test for boxel-ui but it will probably be useful in other tickets like handling dropdown with many realms

tintinthong:
Question: is this behaviour the same for boxel-select?

No, BoxelSelect doesn't have the same auto-close behavior as the BoxelDropdown. The BoxelDropdown has an explicit autoCloseOnOutsidePointer modifier that triggers close when the mouse moves outside the dropdown content,

 While BoxelSelect relies on PowerSelect's built-in behavior with click events that closes when an option is selected or when the user clicks outside the dropdown. Their APIs and behaviors are different.

lucaslyl:
> Is this reproducible inside of a test? We dont seem to have test for boxel-ui but it will probably be useful in other tickets like handling dropdown with many realms

Yes, this behavior is reproducible in tests. While we don't have direct tests in boxel-ui, we can see similar dropdown behavior being tested in the host application's integration tests (eg: in ai-assistant-test.gts). 

I will plan to add the tests in future tickets that handle dropdowns with many realms (since we still have one related ticket have overflow scroll issue - https://linear.app/cardstack/issue/CS-8836/dropdown-content-unable-to-scroll-if-too-many-realms-exist)

tintinthong:
> > Is this reproducible inside of a test? We dont seem to have test for boxel-ui but it will probably be useful in other tickets like handling dropdown with many realms
> 
> 
> 
> Yes, this behavior is reproducible in tests. While we don't have direct tests in boxel-ui, we can see similar dropdown behavior being tested in the host application's integration tests (eg: in ai-assistant-test.gts). 


Can u just add a new test in boxel-ui for dropdowns and build upon it later?

lucaslyl:
> > > Is this reproducible inside of a test? We dont seem to have test for boxel-ui but it will probably be useful in other tickets like handling dropdown with many realms
> > 
> > 
> > Yes, this behavior is reproducible in tests. While we don't have direct tests in boxel-ui, we can see similar dropdown behavior being tested in the host application's integration tests (eg: in ai-assistant-test.gts).
> 
> Can u just add a new test in boxel-ui for dropdowns and build upon it later?

Sure, Done add integration test to verify dropdown auto-close  functionality

richardhjtan (approved):

tintinthong (changes requested):
There is something leaky about this. Testing crashed my browser

packages/boxel-ui/addon/src/components/dropdown/index.gts
```diff
@@ -218,6 +219,57 @@ class BoxelDropdown extends Component<Signature> {
...
+            window.clearTimeout(closeTimeout);
+            closeTimeout = undefined;
+          }
+        }
+      };
+
+      window.addEventListener('mousemove', handleMouseMove);
+
+      return () => {
+        window.removeEventListener('mousemove', handleMouseMove);
```
tintinthong: I think capturing 'mousemove' events is very risky, is it better to use `onFocusOut` argument on `BasicDropdownContent`

lucaslyl: Sure, I added an autoClose argument to the BoxelDropdown component that lets user to control whether the dropdown menu automatically closes when the mouse leaves it. Also, update the dropdown test to handle autoclose behaviour

tintinthong (approved):
Nice extension of this dropdown component :)

PR #2741, opened 2025-06-10
shud never be able remix listing into same realm (altho writable)

tintinthong:
https://linear.app/cardstack/issue/CS-8815/someone-with-catalog-access-should-not-be-able-to-remix-into-catalog

### DIFF
--- a/packages/catalog-realm/catalog-app/components/listing-fitted.gts
+++ b/packages/catalog-realm/catalog-app/components/listing-fitted.gts
@@ -2,6 +2,7 @@ import {
   Component,
   CardDef,
   CardContext,
+  realmURL,
 } from 'https://cardstack.com/base/card-api';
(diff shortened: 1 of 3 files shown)


backspace (approved):
The code looks good to me but conceptually, if we don’t want people to be able to write to the catalog realm, why do they have write permissions to it?

tintinthong:
> The code looks good to me but conceptually, if we don’t want people to be able to write to the catalog realm, why do they have write permissions to it?

This PR is defending from ppl who have write access from accidentally remixing and polluting the catalog. 

We DONT expect users to have this access and they will NOT have catalog realm inside a dropdown option for remix (even before this PR)

But i knw in prod some of our internal users DO. Particuarly, i think in sprint planning there was a discussion to feature flag ctas via a user with write access to the catalog.

lucaslyl (approved):

PR #2744, opened 2025-06-10
Add indicator of current room to Past Sessions menu

lukemelia:
image


backspace (approved):
I was surprised to see a data attribute used for this but I do like how that means the style is added with the same mechanism that the assertion uses.

It’s a subtle highlight but better than before!

PR #2748, opened 2025-06-10
Update "Open In Interact Mode" to open stack with index card behind it

lukemelia:


backspace (approved):
I tried this out in the preview deployment, works as expected!

PR #2757, opened 2025-06-12
Fix code patch filename when code patch belongs to a different realm

jurgenwerk:
Fixes a bug where the currently selected realm does not match the realm of the code patch file (this can happen if you do some work in the ai assistant in one realm and then switch to another one while keeping the AI assistant open). It was actually not a good decision to parse the file name like it was previously–it's totally sufficient to just take the file name from the url and be done with it)

The bug:

image

Fixed:

image


jurgenwerk:
Actually this bug was already recorded in percy but didn't notice it before 🙈

https://percy.io/Cardstack-1/web/-cardstack-host/builds/40953716/changed/2199182560?browser=firefox&browser_ids=64%2C65&group_snapshots_by=similar_diff&subcategories=approved&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=1280&widths=1280

FadhlanR (approved):

habdelra (approved):

PR #2758, opened 2025-06-12
Add missing test for concurrency group in queue

habdelra:


lukemelia (approved):

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1156,7 +1156,7 @@ module(basename(__filename), function () {
           instanceErrors: 3,
           modulesIndexed: 7,
```
backspace:
```suggestion
          instanceErrors: 6,
          modulesIndexed: 10,
```

based on [this expectation only](https://github.com/cardstack/boxel/actions/runs/15616643505/job/43991050653?pr=2758#step:10:1153), I haven’t looked at whether it makes sense

PR #2764, opened 2025-06-13
Add support to deal with fatal vm errors in worker

habdelra:
This PR adds additional capabilities to the worker manager such that if it can see a signal that a worker's VM is about to crash that it can mark the job as failed and mark the in-progress indexed file with an error doc. This specifically arose from the graphing calculator example (https://gist.github.com/habdelra/6d8c99d81f05e2ec9147c182a8f4a23c). In this example there is some very poor float logic that causes a tight loop until the VM runs out of memory. If you tell claude that this gts file is causing out of memory errors, it is able to successfully fix the issue (the problem is `for` loop that renders the grid lines). This PR updates the worker such that as it visits files during indexing the worker reports back to the worker manager what file it is currently visiting. When an out of memory error occurs, the worker manager sees a specific log message that is indicative of VM failure (`FATAL ERROR`). When the worker manager sees a worker log this message it will mark the job as failed with the worker's dying message. as well as it will add an error doc to the index for the file the worker was currently visiting when it issued its dying message.

This then results an error message that looks like this in the playground:

Screenshot from 2025-06-13 14-55-31

AI should be able to fix the issue based on that error message.

### DIFF
--- a/packages/host/app/lib/current-run.ts
+++ b/packages/host/app/lib/current-run.ts
@@ -485,18 +485,15 @@ export class CurrentRun {
     url: string,
     resource: LooseCardResource,
   ) {
-    if (this.#reportStatus) {
-      this.#reportStatus?.({
-        status,
-        jobId: this.#jobInfo?.jobId ? String(this.#jobInfo.jobId) : undefined,
-        url,
-        realm: this.#realmURL.href,
-        deps: [
-          new URL((resource.meta.adoptsFrom as ResolvedCodeRef).module, url)
-            .href,
-        ],
-      });
-    }
+    this.#reportStatus?.({
+      status,
+      jobId: this.#jobInfo?.jobId ? String(this.#jobInfo.jobId) : undefined,
+      url,
+      realm: this.#realmURL.href,
+      deps: [
+        new URL((resource.meta.adoptsFrom as ResolvedCodeRef).module, url).href,
+      ],
+    });
   }
 
   private async indexCard({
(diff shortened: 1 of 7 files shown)


packages/host/app/lib/current-run.ts
```diff
@@ -490,231 +515,236 @@ export class CurrentRun {
...
-    let searchData: Record<string, any> | undefined;
-    let cardType: typeof CardDef | undefined;
-    let isolatedHtml: string | undefined;
-    let atomHtml: string | undefined;
-    let iconHTML: string | undefined;
-    let card: CardDef | undefined;
-    let embeddedHtml: Record<string, string> | undefined;
-    let fittedHtml: Record<string, string> | undefined;
+    this.reportStatus('start', fileURL, resource);
     try {
```
habdelra, line 516: make sure to hide white space when viewing this diff. all we added here was a try/finally with a report status on finish.

packages/realm-server/worker-manager.ts
```diff
@@ -308,19 +386,57 @@ async function startWorker(priority: number, urlMappings: URL[][]) {
     );
   }
   if (worker.stderr) {
-    worker.stderr.on('data', (data: Buffer) =>
-      log.error(`[worker ${name} priority ${priority}]: ${data.toString()}`),
-    );
+    worker.stderr.on('data', (data: Buffer) => {
+      let message = data.toString();
+      log.error(`[worker ${name} priority ${priority}]: ${message}`);
+      if (message.includes('FATAL ERROR')) {
```
habdelra, line 392: this is a dying message from a worker

lukemelia (approved):

PR #2766, opened 2025-06-16
feat(tabbed-header): Add search input yield block

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8868/add-a-search-input-to-tabheader-component

Based on sprint planning discussions and catalog design requirements, add a dedicated search input block in the TabbedHeader component. This change enables user to integrate their search input component while maintaining a consistent layout and styling within the header.

Result:
image


tintinthong (changes requested):
If there is a RHS yield block -- it doesn't have to be search. It could very much be a button such as "Go Home". 

"sub-content" or "side-slot" or "slot" or "right-content" or "right-slot"

lucaslyl:
> If there is a RHS yield block -- it doesn't have to be search. It could very much be a button such as "Go Home".
> 
> "sub-content" or "side-slot" or "slot" or "right-content" or "right-slot"

Make sense. I've updated the name from searchInput to sideContent.

tintinthong (approved):

PR #2767, opened 2025-06-16
Communicate in the action chin that AI is responding

FadhlanR:
- [x] Display "Generating results..." when ai bot responding.
- [x] Display "Apply Diff" when applying code patches and commands.

https://github.com/user-attachments/assets/4df2a00c-f354-438c-afaa-cd11c8bd68ba

### DIFF
--- a/packages/host/app/components/ai-assistant/action-bar.gts
+++ b/packages/host/app/components/ai-assistant/action-bar.gts
@@ -1,33 +1,46 @@
 import { TemplateOnlyComponent } from '@ember/component/template-only';
 import { on } from '@ember/modifier';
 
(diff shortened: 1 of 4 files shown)


lukemelia (changes requested):
The chin should not not say "Applying diff..." when it is executing a command. Instead it should is the `actionVerb` of the command to display something

FadhlanR:
> The chin should not not say "Applying diff..." when it is executing a command. Instead it should is the `actionVerb` of the command to display something

@lukemelia updated. Please check it again.



https://github.com/user-attachments/assets/a2ef63cf-5277-48c8-97de-4c3333ad54b9

lukemelia (approved):

PR #2769, opened 2025-06-16
ui: Update resizable panel look and feel

backspace:
Before, you could only resize using the handles, which were always visible:

screencast 2025-06-16 13-33-19

With this, you can resize from anywhere in the separator which is now narrower, and the handles only show when you hover over the separator:

screencast 2025-06-16 13-31-59


lukemelia (approved):
It seems like the handle should stay visible while you are dragging.

backspace:
> It seems like the handle should stay visible while you are dragging.

I agree, did you try it out or you mean in the GIF? I failed to make another after I added [5b6973f](https://github.com/cardstack/boxel/pull/2769/commits/5b6973f1429cf24e728d9d7867919befb1b958ea), which changes it to not require CSS hover to show the handle.

PR #2773, opened 2025-06-17
Fix todo mvc catalog listing folder structure

richardhjtan:
Refer to https://linear.app/cardstack/issue/ECO-233/remix-producing-file-with-no-filename

**Issue**
- When remixing todo-mvc, the installed file name shows as ".gts"
- Found that todo-mvc listing is not stored under a folder but placed at root of catalog realm
Screenshot 2025-06-17 at 2 25 04 PM

**Fix**
- move todo-mvc to proper listing structure
Screenshot 2025-06-17 at 6 03 53 PM


tintinthong:
what was the file output of the empty file `.gts` ? Completely empty?

richardhjtan:
> what was the file output of the empty file `.gts` ? Completely empty?

@tintinthong still have code inside surprisingly:
Screenshot 2025-06-17 at 6 17 02 PM

tintinthong:
This PR is functioning as a "quick-fix" 

The major fix for this is actually that this line assumes that the file is within a packaged folder-- the code is too primitive

https://github.com/cardstack/boxel/blob/a136ccbedda3f5526b4f2f161051cc2bea8af3fb/packages/host/app/commands/listing-install.ts#L67

The correct solution is actually to just capture a normalized path within the root

@richardhjtan I will proceed with the fix while its your EOD

richardhjtan:
> This PR is functioning as a "quick-fix"
> 
> The major fix for this is actually that this line assumes that the file is within a packaged folder-- the code is too primitive
> 
> https://github.com/cardstack/boxel/blob/a136ccbedda3f5526b4f2f161051cc2bea8af3fb/packages/host/app/commands/listing-install.ts#L67
> 
> The correct solution is actually to just capture a normalized path within the root
> 
> @richardhjtan I will proceed with the fix while its your EOD



> This PR is functioning as a "quick-fix"
> 
> The major fix for this is actually that this line assumes that the file is within a packaged folder-- the code is too primitive
> 
> https://github.com/cardstack/boxel/blob/a136ccbedda3f5526b4f2f161051cc2bea8af3fb/packages/host/app/commands/listing-install.ts#L67
> 
> The correct solution is actually to just capture a normalized path within the root
> 
> @richardhjtan I will proceed with the fix while its your EOD

Added a ticket to proceed this: https://linear.app/cardstack/issue/CS-8874/improve-normalize-file-path-handling-on-catalog-install. Thanks

PR #2774, opened 2025-06-17
Requires approval on catalog listing skill command

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8870/requires-approval-on-catalog-listing-skill-command

Make catalog listing action - eg. remix command requires approval instead of running automatically

Screenshot 2025-06-17 at 6 12 54 PM

### DIFF
--- a/packages/base/Skill/catalog-listing.json
+++ b/packages/base/Skill/catalog-listing.json
@@ -9,21 +9,21 @@
             "name": "default",
             "module": "@cardstack/boxel-host/commands/listing-use"
           },
-          "requiresApproval": false
+          "requiresApproval": true
         },
         {
           "codeRef": {
             "name": "default",
             "module": "@cardstack/boxel-host/commands/listing-install"
           },
-          "requiresApproval": false
+          "requiresApproval": true
         },
         {
           "codeRef": {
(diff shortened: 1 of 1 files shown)


tintinthong:
I thought the test would break

tintinthong:
@richardhjtan I thought `requireApproval` is only on the switching of the submode. The main issue is "actually" that the submode switch will change the focus of the user when he or she is doing something else

richardhjtan:
@tintinthong 

1. The test does not cover the response from AI, we only covered sending a room message and making sure we give the right prompt.

2. The code submode command is one of the operation commands we ran under Remix. One other way is we can tweak the remix command and skill to make another command request to code submode separately - which requires approval. 
So it will be = remix command (auto approval) + code submode command (requires approval)

What do you think?

tintinthong:
> The code submode command is one of the operation commands we ran under Remix. One other way is we can tweak the remix command and skill to make another command request to code submode separately - which requires approval.
So it will be = remix command (auto approval) + code submode command (requires approval)

Yes I thought it was this. The changes in the test will follow

richardhjtan:
Closing this with another approach

PR #2776, opened 2025-06-17
Fix path normalize in listing install

tintinthong:
The point of this PR is to normalize paths CONSISTENTLY when we install resources specified in a listing

The original issue was that resources were being installed in the wrong place. This PR also covers https://linear.app/cardstack/issue/CS-8685/should-not-have-source-files-determine-where-the-file-is-copied-to since now we have a opiniated strategy on how/where to install

### DIFF
--- a/packages/host/app/commands/copy-source.ts
+++ b/packages/host/app/commands/copy-source.ts
@@ -31,5 +31,6 @@ export default class CopySourceCommand extends HostBaseCommand<
     if (r.ok && r.url) {
       return new CopySourceResult({ url: r.url });
     }
+    return new CopySourceResult({});
   }
 }
--- a/packages/host/app/commands/copy-source.ts
+++ b/packages/host/app/commands/copy-source.ts
@@ -7,7 +7,8 @@ import HostBaseCommand from '../lib/host-base-command';
 import type CardService from '../services/card-service';
 
 export default class CopySourceCommand extends HostBaseCommand<
-  typeof BaseCommandModule.CopySourceInput
+  typeof BaseCommandModule.CopySourceInput,
+  typeof BaseCommandModule.CopySourceResult
 > {
   @service declare private cardService: CardService;
 
@@ -21,9 +22,14 @@ export default class CopySourceCommand extends HostBaseCommand<
 
   protected async run(
     input: BaseCommandModule.CopySourceInput,
-  ): Promise<undefined> {
+  ): Promise<BaseCommandModule.CopySourceResult> {
     const fromRealmUrl = new URL(input.fromRealmUrl);
     const toRealmUrl = new URL(input.toRealmUrl);
-    await this.cardService.copySource(fromRealmUrl, toRealmUrl);
+    let r = await this.cardService.copySource(fromRealmUrl, toRealmUrl);
+    let commandModule = await this.loadCommandModule();
+    const { CopySourceResult } = commandModule;
+    if (r.ok && r.url) {
+      return new CopySourceResult({ url: r.url });
+    }
   }
 }
--- a/packages/host/app/commands/listing-install.ts
+++ b/packages/host/app/commands/listing-install.ts
@@ -2,18 +2,22 @@ import { service } from '@ember/service';
 
 import {
   type ResolvedCodeRef,
-  codeRefWithAbsoluteURL,
-  isResolvedCodeRef,
-  type CopyCardsWithCodeRef,
   type Command,
-  listingNameWithUuid,
+  CommandContext,
   RealmPaths,
+  join,
+  type InstallPlan,
+  InstallOptions,
+  planModuleInstall,
+  planInstanceInstall,
+  PlanBuilder,
+  LocalPath,
 } from '@cardstack/runtime-common';
+
+import { logger } from '@cardstack/runtime-common';
 
 import * as CardAPI from 'https://cardstack.com/base/card-api';
 import * as BaseCommandModule from 'https://cardstack.com/base/command';
-
-import type { Skill } from 'https://cardstack.com/base/skill';
 
 import HostBaseCommand from '../lib/host-base-command';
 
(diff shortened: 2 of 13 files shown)


packages/host/app/commands/copy-source.ts
```diff
@@ -21,9 +22,14 @@ export default class CopySourceCommand extends HostBaseCommand<
...
-  ): Promise<undefined> {
+  ): Promise<BaseCommandModule.CopySourceResult> {
     const fromRealmUrl = new URL(input.fromRealmUrl);
     const toRealmUrl = new URL(input.toRealmUrl);
-    await this.cardService.copySource(fromRealmUrl, toRealmUrl);
+    let r = await this.cardService.copySource(fromRealmUrl, toRealmUrl);
+    let commandModule = await this.loadCommandModule();
+    const { CopySourceResult } = commandModule;
+    if (r.ok && r.url) {
+      return new CopySourceResult({ url: r.url });
```
richardhjtan, line 32: I think i missed out to improve this part - We need to return the copy source response. Currently it returns the response from get source, but actually we need response from saveSource. 
Screenshot 2025-06-24 at 9 10 07 PM

packages/host/app/commands/listing-install.ts
```diff
@@ -41,125 +42,93 @@ interface InstallListingResult {
...
+    })
+    .addIf(listing.examples?.length > 0, (opts) => {
+      let r = planInstanceInstall(listing.examples, opts);
+      firstExampleCardId = r.instancesCopy[0].sourceCard.id;
+      return r;
+    })
+    .addIf(listing.skills?.length > 0, (opts) => {
+      let r = planInstanceInstall(listing.skills, opts);
+      skillLocalDir = r.instancesCopy[0].localDir;
+      return r;
```
richardhjtan, line 112: 👍

packages/host/app/commands/listing-install.ts
```diff
@@ -41,125 +42,93 @@ interface InstallListingResult {
...
-      });
-      if (!firstExampleCardId) {
-        firstExampleCardId = newCardId;
-      }
-    }
-  }
+  const finalPlan = builder.build();
+  let results = await install(finalPlan, installOpts, commandContext);
+  log.debug('=== Final Results ===');
+  log.debug(JSON.stringify(results, null, 2));
```
richardhjtan, line 118: are these logs needed?

packages/runtime-common/catalog.ts
```diff
@@ -1,14 +1,273 @@
...
+  }
+}
+
+export class InstallOptions {
+  targetRealm: string;
+  private sourceRealmPath: RealmPaths;
+  private listingDirectoryPath: RealmPaths; // organizing folder of listing
+  installDirectoryName: string; //name of outer uuid  folder
+  sourceDirectoryPath: RealmPaths; // (best guess) listing folder of listing
+  removeListingDirectory: boolean = true; // if true, remove listting directory from source directory
```
richardhjtan:
```suggestion
  removeListingDirectory: boolean = true; // if true, remove listing directory from source directory
```

packages/runtime-common/catalog.ts
```diff
@@ -1,14 +1,273 @@
...
+
+    this.sourceRealmPath = new RealmPaths(sourceRealmURL);
+    this.listingDirectoryPath = new RealmPaths(
+      new URL(join(this.sourceRealmPath.url, listingDirectoryName)),
+    );
+    this.removeListingDirectory = allCodeRefsInSameDirectory(
+      listing,
+      this.listingDirectory,
+    );
+    //use source directory
```
richardhjtan: not sure if needed or to elaborate more

packages/runtime-common/catalog.ts
```diff
@@ -1,14 +1,273 @@
...
+import { logger } from './log';
 
-export function listingNameWithUuid(listingName?: string) {
-  if (!listingName) {
-    return '';
+import type { Listing } from '@cardstack/catalog/listing/listing';
+
+const baseRealmPath = new RealmPaths(new URL('https://cardstack.com/base/'));
+
+// sourceCodeRef -- (installs module) --> targetCodeRef
```
richardhjtan, line 15: perhaps can explain more what sourceCodeRef and targerCodeRef for understanding

packages/host/app/commands/listing-install.ts
tintinthong: These are optional logs based upon your local debug config. Yes. That is part of the point of this plan and install pattern is to have bugs for debeugging

packages/runtime-common/catalog.ts
tintinthong: Commented above

packages/host/app/commands/listing-install.ts
tintinthong: right now this pattern seems a little weird but my intent was to separet the side effects from the pure functional code which is in the planning. 

Now we infer the side effects to make before install occurs but you can see like skills get assign only after install. Can revisit this

It could be this pipe pattern is not really useful

packages/runtime-common/catalog.ts
tintinthong: https://github.com/cardstack/boxel/pull/2776/commits/670370f6e4872795e9e71e0ec7920519a6b1b088

packages/runtime-common/catalog.ts
tintinthong: https://github.com/cardstack/boxel/pull/2776/commits/670370f6e4872795e9e71e0ec7920519a6b1b088

richardhjtan (approved):
Tested and looks good! Great test coverage

tintinthong:
I have tested and this looks good 

Found a bug 
https://linear.app/cardstack/issue/ECO-237/fix-spec-featured-image-field-remix

With logs and test it is much easier to debug now

PR #2779, opened 2025-06-17
Add create-new-file options for interact mode

burieberry:
- top 2 options on the menu are from recent-cards, excluding realm index cards
- "Choose a card type" opens the card catalog
- "Open Code Mode" switches to code-submode and opens the NewFile dropdown menu

This PR includes changes to operator-mode state service's `realmURL` logic.

new-file-interact


packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -18,11 +19,14 @@ import FromElseWhere from 'ember-elsewhere/components/from-elsewhere';
 import { consume, provide } from 'ember-provide-consume-context';
 import window from 'ember-window-mock';
 
+import flatMap from 'lodash/flatMap';
```
lukemelia: Javascript array has flatMap now

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -650,14 +697,131 @@ export default class InteractSubmode extends Component {
...
+    }
+
+    await this.cardContext.actions?.createCard?.(spec.ref, new URL(specId), {
+      realmURL: this.operatorModeStateService.getWritableRealmURL(),
+    });
+  });
+
+  private createNewFromRecentType = restartableTask(
+    async (codeRef: ResolvedCodeRef) => {
+      // TODO: creating instance from non-exported card-def?
```
lukemelia: Is this still TODO?

lukemelia (approved):

packages/host/app/components/operator-mode/interact-submode.gts
burieberry: Ah good point. Is this something that we want to be able to do? It looks like in code-submode, we are not offering the UI option to create or inherit instance from non-exported cards.

lukemelia: I don't think it's necessary to support that

PR #2781, opened 2025-06-18
Cs 8848 restructure catalog layout to match updated design

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8848/restructure-catalog-layout-to-match-updated-design

Enhance catalog design

- Redesigned catalog homepage with sidebar filters that display a grid (pre-rendered card search) when any filter is selected
- Implemented navigation - the "Catalog Home" button resets all filters and returns to the homepage layout
- Added loading and responsive skeleton placeholders to the grid for a better user experience
- Added result not found state

https://github.com/user-attachments/assets/0c0d4db3-cbd5-4611-8086-06d80a792648


packages/catalog-realm/catalog-app/components/grid.gts
```diff
@@ -31,6 +33,10 @@ export class CardsGrid extends GlimmerComponent<CardsGridSignature> {
   @tracked hydratedCardId: string | undefined;
   cardResource = this.args.context?.getCard(this, () => this.hydratedCardId);
 
+  get mockSkeletonCards() {
+    return Array.from({ length: 4 });
+  }
+
```
tintinthong: Hopefully this is going away

packages/catalog-realm/catalog-app/catalog.gts
```diff
@@ -339,7 +337,7 @@ class Isolated extends Component<typeof Catalog> {
   }
```
tintinthong, line 326: remove the mockTags

packages/catalog-realm/catalog-app/components/filter-section.gts
```diff
@@ -2,76 +2,59 @@ import GlimmerComponent from '@glimmer/component';
...
-          {{/each}}
-        </div>
+        <FilterList
+          @filters={{@categories}}
+          @activeFilter={{@activeCategory}}
+          @onChanged={{@onCategorySelect}}
+          class='filter-category-list'
+        />
       {{/if}}
     </FilterGroupWrapper>
```
burieberry, line 42: Can you make the relevant changes to the boxel-ui/FilterList component or move this component there as `FilterGroup` component?

packages/catalog-realm/catalog-app/components/filter-section.gts
```diff
@@ -91,39 +74,41 @@ interface FilterTagGroupArgs {
...
-                {{on 'click' (fn this.handleItemClick item)}}
-                data-test-filter-pill={{item.id}}
+                {{on 'click' (fn this.handleItemClick tag)}}
+                data-test-filter-pill={{tag.id}}
               >
                 <:default>
-                  <span>{{item.name}}</span>
+                  <span>{{tag.displayName}}</span>
                 </:default>
               </Pill>
```
burieberry: It looks like we should create a TagList or TagGroup component in boxel-ui and use it here

packages/catalog-realm/catalog-app/components/grid.gts
```diff
@@ -54,69 +56,86 @@ export class CardsGrid extends GlimmerComponent<CardsGridSignature> {
...
-                        {{on 'click' (fn this.viewCard card)}}
-                      >
-                        <Component />
-                      </CardContainer>
-                    {{/let}}
+    {{#let
+      (component @context.prerenderedCardSearchComponent)
+      as |PrerenderedCardSearch|
+    }}
+      <PrerenderedCardSearch
```
burieberry, line 63: I created a `CardList` component in Base realm (could not move to boxel-ui because of type issues), so we can use that instead of re-rewriting this component again. It does need improvements, though. Currently it's only used for the index cards-grid. Do you think you can use it here?

Here's the link: https://github.com/cardstack/boxel/blob/main/packages/base/components/card-list.gts
Alternatively, we can move that component to the Catalog realm and then use it in base-realm cards-grid card.

burieberry (commented):
I made a couple of suggestions because we're trying to move away from creating custom components to reusable app-layout components that should accommodate this type of app UI. It would really help if we can focus on improving them.

packages/catalog-realm/catalog-app/components/filter-section.gts
burieberry: Also the FilterList component can now accommodate nesting if that is helpful.

packages/catalog-realm/catalog-app/components/grid.gts
lucaslyl: Thanks for pointing this out. Yes, we can reuse the CardList component from the Base realm. 

However, the catalog usage scenario differs — it includes its own hydration logic for cards and a customized empty state.

I’ve improved the CardList to dynamically handle hydration logic with the "enableHydration" argument and yield block states. Here is the PR,  perhaps I can get your opinion and review: https://github.com/cardstack/boxel/pull/2787

packages/catalog-realm/catalog-app/components/filter-section.gts
lucaslyl: I’ve done the update based on your request. This has been addressed in:https://github.com/cardstack/boxel/pull/2804

packages/catalog-realm/catalog-app/components/filter-section.gts
lucaslyl: I’ve done the update based on your request. This has been addressed in: https://github.com/cardstack/boxel/pull/2788

burieberry (approved):
Thanks for creating the tickets. Something noticeable is that the card-header text-color now doesn't match the tablist text color. If the text must be white, then I think we may need to make some changes either in `getContrastColor` boxel-ui helper or if the team agrees, add a new static property in card-def api that allows for overriding the header text color.

packages/catalog-realm/catalog-app/components/filter-section.gts
lucaslyl: > Can you make the relevant changes to the boxel-ui/FilterList component or move this component there as `FilterGroup` component?

actually the **FilterGroupWrapper** is just a wrapper, so it doesn't affect the filterlist
image

lucaslyl:
> Thanks for creating the tickets. Something noticeable is that the card-header text-color now doesn't match the tablist text color. If the text must be white, then I think we may need to make some changes either in `getContrastColor` boxel-ui helper or if the team agrees, add a new static property in card-def api that allows for overriding the header text color.

Sure, I filed a ticket for this for further investigation: https://linear.app/cardstack/issue/CS-8916/getcontrastcolor-should-have-property-override-the-header-text-color. We will need a discussion with Chris during design review cycle.

PR #2784, opened 2025-06-18
Use single matrix account for workers

habdelra:
This PR works around a synapse performance issue where repeated calls to the /login endpoint kill the matrix performance. To work around this all the workers now login with the realm server matrix creds (whose session is cached from job to job). In order to prevent security issues--e.g. using indexing as a back door to view content that you otherwise don't have access to view in a different realm, I've also added a new permission `assume-user`. This permission is currently only permitted to the realm server user. And when the `X-Boxel-Assume-User: @matrix_user:host` header is present and the session has `assume-user` permissions, then the permissions of the session will inherit the permissions of the assumed user. In this manner we can prevent the realm server user working on behalf of a realm bot user from being able to access more content than it is permitted to.

an interesting thought behind the assume user permissions, is that we could actually make the realm indexing see exactly the same permissions as the user that made the realm. this would probably make the system seem more consistent (less surprising) during indexing of links and modules between a user's own private realms.


packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -321,98 +321,6 @@ module(basename(__filename), function () {
           }
         });
 
-        test('POST /_create-realm without copying seed realm', async function (assert) {
```
habdelra, line 324: this test is superfluous. it should have been removed when we removed seed realms

backspace (approved):
Is it worth deploying this to production to see if it does fix the problem?

PR #2787, opened 2025-06-19
Cs 8903 enhance and implement cardlist from the base into the catalog

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8903/enhance-and-implement-cardlist-from-the-base-into-the-catalog

- Replace custom CardsGrid with base CardList component
- Add card hydration functionality with mouse hover interactions ( now cardlist can support the hydration feature)
- Implement proper loading and empty state handling with custom UI
- Remove redundant custom grid and skeleton components
- Enhance CardList with hydration, view card actions, and improved styling

image


packages/base/components/card-list.gts
```diff
@@ -20,66 +25,190 @@ interface Signature {
...
+          {{/if}}
+        </:loading>
+        <:response as |cards|>
+          <ul
+            class={{cn
+              'boxel-card-list'
+              grid-view=(eq @viewOption 'grid')
+              strip-view=(eq @viewOption 'strip')
+              card-view=(eq @viewOption 'card')
+              single-column-view=(eq cards.length 0)
```
lucaslyl, line 97: The single-column view is needed because empty states require a full-width layout instead of being constrained to a single grid column.

burieberry (commented):
This is a good attempt, but I see the issue you faced about including hydration. I didn't realize there was so much other stuff going on. I'm not fully understanding the reasoning behind mouseover hydration but it seems like it should not be a concern of this component to do hydration. It could render a given list of already-hydrated cards. Previously, we abstracted the overlay-related logic to card overlay component and had the @context.cardComponentModifier to handle adding it.

We can incorporate some of the other enhancements such as the empty state. I'm happy to take over and merge those, but it looks like this component doesn't work with catalog app without adding the hydration logic.

lucaslyl:
> This is a good attempt, but I see the issue you faced about including hydration. I didn't realize there was so much other stuff going on. I'm not fully understanding the reasoning behind mouseover hydration but it seems like it should not be a concern of this component to do hydration. It could render a given list of already-hydrated cards. Previously, we abstracted the overlay-related logic to card overlay component and had the @context.cardComponentModifier to handle adding it.
> 
> We can incorporate some of the other enhancements such as the empty state. I'm happy to take over and merge those, but it looks like this component doesn't work with catalog app without adding the hydration logic.

Sure you can take over and very Thank you for review, You’re right — the catalog card fitted format comes with quite a bit of event bubbling and z-index complexity. On top of that, the preview and remix functionalities added additional layers of logic. After discussing with Chris, we decided to include the hydration logic directly here, and avoid using cardComponentModifier within the catalog app.

Also, based on your feedback, I think it makes more sense to stick with my previous custom grid component instead of switching to CardList, since hydration can’t be moved into it.

I will close this pr for now ya.

lucaslyl:
> This is a good attempt, but I see the issue you faced about including hydration. I didn't realize there was so much other stuff going on. I'm not fully understanding the reasoning behind mouseover hydration but it seems like it should not be a concern of this component to do hydration. It could render a given list of already-hydrated cards. Previously, we abstracted the overlay-related logic to card overlay component and had the @context.cardComponentModifier to handle adding it.
> 
> We can incorporate some of the other enhancements such as the empty state. I'm happy to take over and merge those, but it looks like this component doesn't work with catalog app without adding the hydration logic.

Burcu, this is the ticket I filed: https://linear.app/cardstack/issue/CS-8918/card-list-can-include-loading-and-emptystate — so we can track what’s going on in future. If you need any help from me, please let me know.

PR #2789, opened 2025-06-19
Ability to stop the AI response

FadhlanR:
In this PR, I implemented a "Stop" button to halt the AI bot's response. When a message is canceled, a "{Generation Cancelled}" indicator is displayed beneath it.

https://github.com/user-attachments/assets/2f4ce05c-a103-4020-a66d-60c24a4df2e7

https://github.com/user-attachments/assets/e304be8d-51d3-4c9a-b3d2-4e1873d0f869


packages/ai-bot/lib/matrix/response-publisher.ts
```diff
@@ -135,6 +135,7 @@ export default class MatrixResponsePublisher {
 
       let extraData: any = {
         isStreamingFinished: responseStateSnapshot.isStreamingFinished,
+        isCancelled: responseStateSnapshot.isCancelled,
```
lukemelia: Let's use the American spelling: `isCanceled`

packages/ai-bot/lib/responder.ts
```diff
@@ -158,14 +161,15 @@ export class Responder {
     }
   }
   isFinalized = false;
-  async finalize() {
+  async finalize(isCancelled = false) {
```
lukemelia: make this an optional `options` argument with an isCanceled prop

packages/ai-bot/lib/responder.ts
```diff
@@ -158,14 +161,15 @@ export class Responder {
...
     if (this.isFinalized) {
       return;
     }
     this.isFinalized = true;
 
     let isStreamingFinishedChanged =
       this.responseState.updateIsStreamingFinished(true);
+    this.responseState.isCancelled = isCancelled;
```
lukemelia: maybe this should be an argument to `updateIsStreamingFinished` instead of a direct property setter?

lukemelia (changes requested):
Have we covered the case where generation is stopped during reasoning?

FadhlanR:
> Have we covered the case where generation is stopped during reasoning?

@lukemelia I've added a test in host to check whether an event canceled during reasoning displays the `{Generation Canceled}` marker in the message. I think at the moment that is the only test I could add. I also attached a screenshot showing the cancellation during reasoning.

https://github.com/user-attachments/assets/4b150bc9-94c1-47fc-ae7f-18c196355551

lukemelia (approved):

PR #2791, opened 2025-06-19
Fix incorrect generated matrix userIds

habdelra:
This fixes an issue in prod where the matrix user IDs that we are deriving are incorrect


habdelra:
I'm deploying this directly to prod as the workers are not logging in with the correct matrix user ID's

PR #2793, opened 2025-06-19
Use single matrix account for workers

habdelra:
Take 2!

This PR works around a synapse performance issue where repeated calls to the /login endpoint kill the matrix performance. To work around this all the workers now login with the realm server matrix creds (whose session is cached from job to job). In order to prevent security issues--e.g. using indexing as a back door to view content that you otherwise don't have access to view in a different realm, I've also added a new permission assume-user. This permission is currently only permitted to the realm server user. And when the X-Boxel-Assume-User: @matrix_user:host header is present and the session has assume-user permissions, then the permissions of the session will inherit the permissions of the assumed user. In this manner we can prevent the realm server user working on behalf of a realm bot user from being able to access more content than it is permitted to.

Additionally this PR performs the indexing using the permissions of the user that created the realm. this will address the inability to share instances and modules among a user's own private realms.

 :heavy_check_mark: This branch has been successfully tested on both staging and production environments


jurgenwerk (approved):
Looking good

PR #2797, opened 2025-06-19
File Tree should immediately update to show new directories

lukemelia:


backspace (approved):
I tried this with `curl` in the preview deployment, success! 🎉

PR #2798, opened 2025-06-19
Workspace sync package

IanCal:
This adds a new package to the monorepo. Its goal is to be a simple CLI interface for pulling and pushing the contents of workspaces (realms) to your local device.

By specifying a username, password and matrix url as env vars, the package

To try it out locally you can build and install it,

`pnpm pack && npm install -g cardstack-workspace-sync-cli-0.1.0.tgz`

Then with the env vars you can run

`workspace-pull http://workspace/location ./local-version`

and push with

`workspace-push ./local-version http://workspace/location`

There's a `--dry-run` and a `--delete` which deletes non-matching files. This package *does not check for changes or modified times* so it simply copies everything from one place to another.

This lets you copy a realm out of production and into a local version if you want.

The main reason for this is skills dev. With this package, you can have the workflow of

* Checkout skills repo
* Push skills repo contents to a workspace of your own (workspace-push)
* modify live, test, and work on the cards in boxel
* Pull the cards out (workspace-pull)
* Commit to git

I published this to npm under a test account I made to try it out in CI on the skills repo:
@iancalcardstack/workspace-sync-cli


lukemelia (approved):

habdelra:
Could you also optionally include the seed in lieu of the password for the matrix client? this will allow us to be able to inspect other realms without asking a user for their password. the `MatrtixClient` should work with a seed instead of a password.

IanCal:
> Could you also optionally include the seed in lieu of the password for the matrix client? this will allow us to be able to inspect other realms without asking a user for their password. the `MatrtixClient` should work with a seed instead of a password.

Good point, I've added CS-8935 to add this. This is using the secret to construct the password right?

habdelra:
> > Could you also optionally include the seed in lieu of the password for the matrix client? this will allow us to be able to inspect other realms without asking a user for their password. the `MatrtixClient` should work with a seed instead of a password.
> 
> Good point, I've added CS-8935 to add this. This is using the secret to construct the password right?

yup, that's right

PR #2800, opened 2025-06-19
Remove ability to include commands directly on a message

lukemelia:
- Introduce waitForMatrixEvent, waitForCompletedCommandRequest utils
- Add GetEventsFromRoom command that is used to implement waitForMatrixEvent util
- Removes ability to include commands directly on a message
- Removed Command#waitForNextCompletion
- Removed "invocations" internals of Command that are no longer needed


habdelra (approved):

IanCal (approved):
Looks good, I assume there are some things in the ai bot that can be removed too?

I think it'd be fine to just delete the things we're mostly commenting out in the experiments realm, we're changing things enough that removing things we don't support any more should be alright. Product requirements and app gen we don't have any more, really.

lukemelia:
> Looks good, I assume there are some things in the ai bot that can be removed too?

I don't think things look any different to the AI Bot. Am I forgetting something?

lukemelia:
> I don't think things look any different to the AI Bot. Am I forgetting something?

I see that are close to removing context.tools on CardMessageContent. We need to eliminate the auto per-type patch commands before we do though.

jurgenwerk (approved):

lukemelia:
Please check the changes since your last review.

PR #2802, opened 2025-06-20
Add dropdown menu to plus button in chat input

FadhlanR:
Previously, when the "Plus" button in the chat input was clicked, the file chooser or card chooser modal would open directly based on the current submode. With this PR, I added a menu that allows users to choose which modal to open, so they can attach files or cards regardless of the submode.

https://github.com/user-attachments/assets/132d162f-579f-43bf-ac99-92384c366346


jurgenwerk (approved):
Code looks good but I find the wording inconsistent:

1. It says "Attach a card" and "Choose a file", why is there a different verb when it does the same thing to either of those
2. When clicking "Attach a card", the modal will say "Choose a card instance" and when clicking "Choose a file", the modal will say "Attach File"

FadhlanR:
> Code looks good but I find the wording inconsistent:
> 
> 1. It says "Attach a card" and "Choose a file", why is there a different verb when it does the same thing to either of those
> 2. When clicking "Attach a card", the modal will say "Choose a card instance" and when clicking "Choose a file", the modal will say "Attach File"

Which one do you think better to use 'Attach' or 'Choose'? I'll update those text.

jurgenwerk:
I'd prefer "Attach" for both, but we can check with Chris

PR #2807, opened 2025-06-23
Catalog view card action on Details button

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8847/restyle-listing-fitted-format-to-match-updated-design

Result:

https://github.com/user-attachments/assets/ab96c0c8-9a11-4680-8dd1-7c5cc003f2b7

### DIFF
--- a/packages/catalog-realm/catalog-app/components/cards-display-section.gts
+++ b/packages/catalog-realm/catalog-app/components/cards-display-section.gts
@@ -1,7 +1,4 @@
 import GlimmerComponent from '@glimmer/component';
-import { action } from '@ember/object';
-import { on } from '@ember/modifier';
-import { fn } from '@ember/helper';
 
 import {
   CardDef,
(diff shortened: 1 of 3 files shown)


tintinthong:
how is the finger icon supposed to behave? 

It looks like now it disappears. but on overlay the finger icon will still show. 

last time, i think the finger icon was always showing even when we hover preview, now it changes to a cursor

lucaslyl:
> how is the finger icon supposed to behave? 
> 
> 
> 
> It looks like now it disappears. but on overlay the finger icon will still show. 
> 
> 
> 
> last time, i think the finger icon was always showing even when we hover preview, now it changes to a cursor

That might be the Mac default screen recording issue ,I'm able to see the pointer while I hover the button in localhost,it's disappear during recording

tintinthong (approved):

PR #2809, opened 2025-06-23
Able to omit relationships when copy card

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-8872/fix-remix-for-sprint-planner

These changes can resolve the issue of installing the sprint planner listing into the user's realm, as the attached project instance's UUID changes when it is copied. For now they have to create their own after remix.


habdelra (commented):
please add a test for this

IanCal:
I'm concerned we're adding lots of custom functionality into copy card.

I also think copy card is probably the wrong thing for this. All the files have relative links in them right? Can't we just do read text file -> write text file?

tintinthong:
> I'm concerned we're adding lots of custom functionality into copy card.
> 
> I also think copy card is probably the wrong thing for this. All the files have relative links in them right? Can't we just do read text file -> write text file?

This is what im doing now 
- renaming copy card command into copy card from stack <- this operation will work in the 2 stack use case. I will revert copy card command to its original state
- create a new copy card command. Optionally, if we want to use omitRelationships it should go to this new copy card command. 
- yesterday I mentioned exactly what you are talking about. 
   - Original implementation was to use the old copy card command to create a new card always. This has limitations bcos the links would break since new ids are created. Hence this PR. But, this had a benefit bcos all instances typically had all directory instances in the root of the target folder -- it was organized. 
   - I wanted to solve this similar to how copyFromSeed worked previously. The files for the instances are created. That way relative links are preserved. But, this is simply NOT immune to breakage of links.
   - @habdelra had another suggestion to use lids that have the ability to preserve relationship links and also create a card with a custom id. Im trying to get that to work now. Im not sure with the concerns about speed on this solution as that has become a concern now

IanCal:
> * But, this is simply NOT immune to breakage of links.

Can I ask what links would be broken? 

Broadly I'm wondering if we can constrain the problem such that copying a folder works, or at a minimum that identifying relative urls that link outside of the folder get rewritten as absolute urls.

edit - for example, could we have a way of generating a zip file *once* that contains everything we need, relative urls for things we copy, absolute for ones we don't, and that's the thing we copy around? Have an api to accept a zip of files writing to a folder, and we're down to two network requests. One to pull a few kb zip file and one to push one. Modifying and rewriting things on the fly without breaking an arbitrary app seems very complex.

habdelra:
> > I'm concerned we're adding lots of custom functionality into copy card.
> > I also think copy card is probably the wrong thing for this. All the files have relative links in them right? Can't we just do read text file -> write text file?
> 
> This is what im doing now
> 
> * renaming copy card command into copy card from stack <- this operation will work in the 2 stack use case. I will revert copy card command to its original state
> * create a new copy card command. Optionally, if we want to use omitRelationships it should go to this new copy card command.
> * yesterday I mentioned exactly what you are talking about.
>   
>   * Original implementation was to use the old copy card command to create a new card always. This has limitations bcos the links would break since new ids are created. Hence this PR. But, this had a benefit bcos all instances typically had all directory instances in the root of the target folder -- it was organized.
>   * I wanted to solve this similar to how copyFromSeed worked previously. The files for the instances are created. That way relative links are preserved. But, this is simply NOT immune to breakage of links.
>   * @habdelra had another suggestion to use lids that have the ability to preserve relationship links and also create a card with a custom id. Im trying to get that to work now. Im not sure with the concerns about speed on this solution as that has become a concern now

i would think that the lid solution will always be faster than issuing a single API request for each file. the reason is that during indexing, the lid solution will batch all the updates into a single indexing operation, where when you do a single request per file that is a separate indexing operation for each file (and indexing jobs run serially within a realm).

also keep in mind `lid` are strings and not URL's. but you can preserve the "relative-ness" by popping the last segment of the path for each link and using that as the `lid`. that way the same UUID corresponds to the same instance in the two different realms.

I think @IanCal's idea around zipping files is not a bad one. but keep in mind the indexing step too. in that case you'd want to do a batch index on all the contents of the zip instead of indexing each item in the zip separately--as i mentioned, indexing jobs run serially within a particular realm.

I do think, though, it can be challenging trying to draw a folder boundary around all the links a card instance may have. what a card links to is pretty arbitrary. a user can just select any instance that matches the card type in the card chooser regardless of what folder it lives in. so such a solution will very quickly be outgrown.

The deeper challenge: the fact that copy is never just a single card, but rather all the instances/modules it consumes gets back to realm permissions. the absolute URLs that you fashion for links that are not copied should always terminate on realms that are readable for the audience of the copied cards--which i think at this stage is probably realms that are world read. that is really the boundary that we draw around cards that are copied and ones that are not.

tintinthong:
> Can I ask what links would be broken? Broadly I'm wondering if we can constrain the problem such that copying a folder works, or at a minimum that identifying relative urls that link outside of the folder get rewritten as absolute urls.

**To Answer your Question**

Not that the links are brokened. We just need alot of relative file path handling. The current implementation is to re-position relative position of instances for cleaniness. Typically, all instances (skills & examples) are placed into a folder `Author/*` into the root of the target directory `author-xyz123`. 

Here is an example 

**Example 1: Copy files from realm, re-position instance**

Catalog
- author-code/author.gts <- link: ../components/form
- author-code/Author/1.json <- link: ../author
- components/form.gts
- some-folder/some-other-folder/SomeAuthor/2.json <- link: ../author

Target Realm 
- author-xyz123/author-code/author.gts 
- author-xyz123/components/form.gts
- author-xyz123/Author/{some-uuid-for-1}.json
- author-xyz123/Author/{some-uuid-for-2}.json

The Example above is meant to maintain relativity for modules only. **Relativity of instances are not of a concern NOW bcos we use CopyCard to point the instance to the correct module.**

**Does this have to be the case?**
Perhaps not. 

A way to think about this is we copy the files into `author-xyz123` as tho its the target realm
But you can see that we will expose our catalog realm directory structure into `author-xyz123` and will impose catalog users to instances in a certain way. 

**Example 1.a: Copy files from realm, DO NOT re-position instance**

Catalog
- author-code/author.gts <- link: ../components/form
- author-code/Author/1.json <- link: ../author
- components/form.gts
- some-folder/some-other-folder/SomeAuthor/2.json

Target Realm 
- author-xyz123/author-code/author.gts 
- author-xyz123/components/form.gts
- author-xyz123/author-code/Author/{some-uuid-for-1}.json
- author-xyz123/some-other-folder/Author/{some-uuid-for-2}.json

> edit - for example, could we have a way of generating a zip file once that contains everything we need, relative urls for things we copy, absolute for ones we don't, and that's the thing we copy around? Have an api to accept a zip of files writing to a folder, and we're down to two network requests. One to pull a few kb zip file and one to push one. Modifying and rewriting things on the fly without breaking an arbitrary app seems very complex.

I too totally agree with your zip idea. probably also good for a lot of the tar package debugging workflow that we have

tintinthong:
@IanCal @habdelra I have spent some time to think deeper about this issue and product requirements and I think we can simplify without being tied down to the current implementation 

We can use the previous infrastructure `copyFromSeed`. The implementation will be like this

- get rid of "copying instance to root of target directory" idea !! 
- relative files are handled  `sourceRealm -> target directory`. We treat target directory like a new realm and it will mirror the directory structure of how the files are laid out in the source realm
- we expose a new api endpoint on the realm `/copy` where we specify a payload of file urls [ Dir1/file1.json, file2.gts ] and source realm (catalog) 
- already placed infrastructure: 
  - copy calls api calls out to realm-index-updater method `copy` which queues the job
  - worker creates batch transaction and calls out `copyFrom`
  - (extend the api) apis here have to be updated to take in an array of file paths
- copy endpoint will return a succes result and look for side effects to focus on inside of playground
- zip feature can be worked on separately
- lid solution not needed

WDY think?

IanCal:
@habdelra totally agree on the copying being more complex in the general case - I'm not sure how much that can be automated as well (e.g. you may want a class to remain an actual subclass of another and fail if permissions aren't correct instead of succeeding but creating a new incompatible carddef). That's part of why I'm concerned with adding a lot to copycard because it's a complex problem. The listing problem for remixes may be simpler as we can impose more constraints on it. We're explicitly happy creating parallel types because the point is to give people code to mess about with.

@tintinthong I'm not too familiar with the internal details of how best to do the bulk work but if this can be moved to doing something internal where we copy some files or it's a single db transaction copying rows that should get the total time to remix down to ~the indexing time. Perhaps we could copy the prerendered indexing results while marking them stale/triggering the new run? Cards would be available near instantly and usually the reindex wouldn't change them. Not a requirement, just wondering how quickly remixing could complete. 

Sounds like this is very similar to copying the seed realm but 
* there's a target subdir
* it's filtered on only certain cards/carddefs

However we do it, as you both say, having one indexing run for a bulk operation will make a huge difference. 

The zip idea I like but definitely isn't a specific feature I'd suggest right now, I think it was more "if we make things work so that we can just move files, packaging becomes trivial".

PR #2819, opened 2025-06-23
Link up skills realm and host

IanCal:
This PR starts using the new skills realm. It should be merged *after* https://github.com/cardstack/boxel/pull/2844

The skills themselves are in https://github.com/cardstack/boxel-skills

The intended workflow is:

Local skills dev:
* Log in as skills_writer
* Edit the skills directly in boxel
* Commit as a new branch in the boxel-skills repo
* Raise a PR
* On merge, the changes get pushed to staging
* Tag the commit to release

Remote skills dev:

* Checkout the boxel-skills repo
* Create a new branch
* Use the cli tool to write the current skills to a realm you control on production
* Experiment, work directly in boxel to try out the skills
* Use the cli tool to pull changes back to your local folder
* Commit, push the new branch
* Raise a PR
* On merge, the changes get pushed to staging
* Tag the commit to release

This all adds an additional overhead that you need to update the skills repo (there is a helper script for this). 

We could use submodules or possibly subtrees later, however these have other tradeoffs with overhead.


backspace:
I tried this locally and was able to duplicate a skill and edit it, but where does it actually live? I looked at the `boxel-skills` repository, the explanation there is production-centric, but is there a version that applies for local development?

I had to run `pnpm register-skills-writer`, but should this just be part of `register-realm-users`?

Maybe this all needs to be added to `README`?

IanCal:
> I tried this locally and was able to duplicate a skill and edit it, but where does it actually live? I looked at the `boxel-skills` repository, the explanation there is production-centric, but is there a version that applies for local development?
> 
> I had to run `pnpm register-skills-writer`, but should this just be part of `register-realm-users`?
> 
> Maybe this all needs to be added to `README`?

I'll get things into the README, including a walkthrough of developing skills working locally and for those without a dev environment. Thanks for catching the missing info on register-skills-writer, it's part of register-all and will be done as part of a full reset but there's no info right now on current users.

For local dev, the boxel-skills repo is checked out into packages/skills-realm/contents . Editing the skills as skills_writer writes them back to that location, you can then commit to a new branch and push for review. The idea is you can then edit and try them out live (as you're editing them directly in the realm), then when happy push for review in boxel-skills and they can then go to staging / prod easily.

backspace (approved):
Thanks for the explanation, I was able to find the new file in that path 🎉

PR #2824, opened 2025-06-24
Do not require a Stripe account at signup 

jurgenwerk:
TODO:
- [x] Configure Stripe sandbox environment
- [ ] Stripe staging environment
- [ ] Stripe production environment

This PR makes a change where users who sign up for boxel do not need to go through the Stripe signup anymore. On signup, new users receive 1k boxel AI credits once and that's it. Then, they have 3 plans to choose from (all of them need a stripe subscription):

1. Starter ($0/mo, 2500 credits monthly) 
2. Creator ($12/mo, 6500 credits monthly)
3. Power user ($49/mo, 35000 credits monthly)

Here's how the profile popup looks like when on free non-stripe plan:
image

When clicking on "Upgrade plan", the new pricing table opens in a modal (ported from https://boxel.ai/pricing):
image

The plan links are stripe payment links. In case user already has a subscription, the stripe link will open the customer portal where the user can manage their subscription (upgrade, cancel):
image
image

Here's how it looks like when user has a subscription (see the "Current plan" badge", and the link says "Manage plan"):
image


jurgenwerk:
Looks like something went wrong with rebase and some other commits crept in. I can clean it up but in the meanwhile it's ready for review

habdelra (approved):

richardhjtan:
Tested and it works well 👍

jurgenwerk:
@richardhjtan thanks!

PR #2820, opened 2025-06-23
host: Update chat header design

backspace:
Current:

Experiments Workspace 2025-06-25 10-35-33

Design:

LATEST UI Jun 17 - LATEST UI Jun 17 - Zeplin 2025-06-25 10-32-41

Implementation (font and padding inside chat bubbles are a separate issue):

garden-design gts in Experiments Workspace 2025-06-25 11-41-33

---

**Beneath here the screenshots are from before I fixed the padding and spacing, but the aspects they’re highlighting don’t relate to that**

Hover to reveal full title:

garden-design gts in Experiments Workspace 2025-06-25 10-55-45

Loading past sessions:

screencast 2025-06-25 10-46-43

Past session activity:

screencast 2025-06-25 10-48-09

Because the sessions popover is also scheduled for a redesign I ignored this truncation:

garden-design gts in Experiments Workspace 2025-06-25 10-51-29

PR #2826, opened 2025-06-24
host: Change diff editor to bubble scroll events

backspace:
I noticed while working on #2820 that I had to place the cursor in a narrow strip to be able to scroll the chat sidebar freely. This adds `alwaysConsumeMouseWheel: false` for diff editors, which was already present for regular editors, and extracts shared editor settings.

It’s difficult to convey in a GIF because you can’t tell when I’m scrolling (on a trackpad) but you can see that on the left once the cursor is atop the editor, I can only scroll within it, I can’t escape it without moving the cursor outside the editor. On the right, when the editor has fully scrolled, the container then scrolls.

<table>
<tr>
<td>before</td>
<td>now</td>
</tr>
<tr>
<td>

screencast 2025-06-24 10-08-58

</td>
<td>

screencast 2025-06-24 10-10-26

</td>
</tr>
</table>

### DIFF
--- a/packages/host/app/components/ai-assistant/code-block.gts
+++ b/packages/host/app/components/ai-assistant/code-block.gts
@@ -457,6 +457,9 @@ class CodeBlockDiffEditor extends Component<Signature> {
       contextLineCount: 1,
     },
     readOnly: true,
+    scrollbar: {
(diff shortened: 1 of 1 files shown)


lukemelia (approved):
Good idea!

PR #2828, opened 2025-06-24
host: Fix background of inspector dividers

backspace:
It’s subtle but the background shouldn’t be different:

<table>
<tr>
<td>
Before
</td>
<td>
Now
</td>
</tr>
<tr>
<td>
index json in Experiments Workspace 2025-06-24 13-31-15
</td>
<td>
index json in Experiments Workspace 2025-06-24 13-30-57
</td>
</tr>
</table>

Another way to accomplish this without needing to keep the values in sync would be inheriting the background down the hierarchy, better?

bm3 (Git) 2025-06-24 13-34-32


lukemelia (approved):
Not sure what approach is best -- use your best judgement! :)

PR #2841, opened 2025-06-26
Cs 8965 skeleton loader should show 10 cards

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8965/skeleton-loader-should-show-10-cards

- Create a renderSkeletons function with configurable length
- Add default length of 10 skeletons
- Allow custom skeleton counts for different use cases

Screenshot 2025-06-26 at 10 55 12 AM

### DIFF
--- a/packages/catalog-realm/catalog-app/components/grid.gts
+++ b/packages/catalog-realm/catalog-app/components/grid.gts
@@ -43,6 +43,11 @@ export class CardsGrid extends GlimmerComponent<CardsGridSignature> {
     this.hydratedCardId = cardId;
   }
 
+  //default to rendering 10 skeletons
+  renderSkeletons(length: number = 10) {
+    return Array.from({ length }, (_, i) => i);
+  }
+
   isHydrated = (cardUrl: string) => {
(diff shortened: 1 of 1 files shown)


packages/catalog-realm/catalog-app/components/grid.gts
```diff
@@ -43,6 +43,11 @@ export class CardsGrid extends GlimmerComponent<CardsGridSignature> {
     this.hydratedCardId = cardId;
   }
 
+  //default to rendering 10 skeletons
+  renderSkeletons(length: number = 10) {
+    return Array.from({ length }, (_, i) => i);
+  }
```
tintinthong: no usages of renderSkeletons as a function 

make into getter. hardcode the number.

lucaslyl: Done update the func to getter

tintinthong: 👍

tintinthong (approved):

PR #2844, opened 2025-06-26
Add skills realm

IanCal:
This PR adds a new realm for storing skills. It pulls from https://github.com/cardstack/boxel-skills when run locally (the git checkout should be done automatically for you on start:all) and you can log in as the skills user with 

Due to an ordering issue, this PR *does not start using these skills*. We need the skills realm to exist, to have the right users, to have the right content and *then* we can move to using those new skills. This PR shows this step and moving on working together https://github.com/cardstack/boxel/pull/2819

The skills realm in staging and production will be written to using the API, using files simplifies things locally.


habdelra:
I think there are host app issues around calculating the default realm that for some reason think that the new skills realm is the default realm.

IanCal:
> I think there are host app issues around calculating the default realm that for some reason think that the new skills realm is the default realm.

I'll have a look into it, I think this might come down to a permissions issue in the mock matrix client, assuming all unknown realms are world writable and therefore skills comes up as a writable one but it isn't.

lukemelia (approved):

PR #2845, opened 2025-06-26
Prevent copy source running more than once

tintinthong:
- Cleaned up test 
- introduce modulesToInstall


richardhjtan (approved):
Great improvement and refactor of tests

PR #2850, opened 2025-06-26
Tests for the workspace cli

IanCal:
Built with claude code. Creates an isolated test realm like in matrix and tries pulling and pushing some things to it.


habdelra:
Is there a reason why we don't use qunit for testing? we use it for all the other tests, except the matrix playwright tests. It would be nice to just continue the trend of using the same test framework instead of introducing another.

habdelra:
Also, probably, like we do with the ai-bot package, we don't need to run all the tests when only changes are made in the workspace-cli package.

habdelra (changes requested):
Also, please do not forget to lint (typecheck) this package.

habdelra (approved):

IanCal:
@habdelra thanks, I'd had problems getting things to work when using qunit and had left it to fix later but should have just done it here. I've got them working with one remaining issue around connection pool limits with the DB so have kept the structure for one realm server & worker shared between all tests - I'll raise a ticket about fixing that.

Linting is enabled already, and it's supposed to be setup to only run these tests when this package or the other common things change (runtime common, pnpm file, that kind of thing, there's a shared definition I'm reusing for that). Because of the changed yml file everything is getting triggered.

packages/workspace-sync-cli/tests/integration-test.ts
```diff
@@ -0,0 +1,484 @@
...
+  hooks.afterEach(async function () {
+    console.log('\n🧹 Cleaning up test environment...');
+
+    // Only clean up test-specific temp directory
+    if (context?.tempDir) {
+      await fs.rm(context.tempDir, { recursive: true, force: true });
+    }
+  });
+
+  test('Pull files from realm to local directory', async function (assert: any) {
```
habdelra: out of curiosity, are you having type issues with `assert`? you shouldn't have to specify the type, as the `QUnit.test()` function is already typed

IanCal: I was, but now am not - perhaps it was something else I had broken part way through and never fixed. I'll remove those in the next pr on this

IanCal: oh I have a few issues and a lint issue here anyway so I'll fix in this pr

PR #2853, opened 2025-06-27
Retry realm auth when server responds with 500 error

habdelra:
we believe that realm is sometimes unable to login to matrix in CI which results in failed auth requests because the realm doesn't know who it is. in these cases the realm responds with a 500 error, so we try again...

### DIFF
--- a/packages/runtime-common/realm-auth-client.ts
+++ b/packages/runtime-common/realm-auth-client.ts
@@ -1,4 +1,4 @@
-import { unixTime } from './index';
+import { unixTime, delay } from './index';
 import { TokenClaims } from './realm';
 
 // iat - issued at (seconds since epoch)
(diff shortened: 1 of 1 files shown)


packages/runtime-common/realm-auth-client.ts
```diff
@@ -141,31 +141,55 @@ export class RealmAuthClient {
...
-    return this.fetch(`${this.realmURL.href}${this.sessionEndpoint}`, {
-      method: 'POST',
-      headers: {
-        Accept: 'application/json',
-      },
-      body: JSON.stringify({
-        user: this.matrixClient.getUserId(),
-        challenge,
-        ...(challengeResponse ? { challengeResponse } : {}),
+    return this.withRetries(() =>
```
jurgenwerk, line 161: Can this come in handy? https://github.com/cardstack/boxel/blob/38dd36fe2512c41e143d9ce8a6b2e45e3793f5f8/packages/runtime-common/utils.ts#L3-L24

jurgenwerk (approved):

packages/runtime-common/realm-auth-client.ts
habdelra: almost--this mechanism doesn't have a back off. maybe we could add a backoff function to this?

habdelra: oh wait--actually no. the `retry()` function above uses a `catch` block to trigger the retry,. in our case nothing is throwing--rather it's a status of 500 that triggers the retry. a 500 response is returned from a fetch without throwing. a throw would be more like the fetch failed which is not the case here.

PR #2861, opened 2025-06-30
feat: make carousel preview example and info section show listing details

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8967/fix-clicking-on-overlay

- Adjust the card styling and fix accessibility
- Add an acceptance test to cover the listing card navigation.
image

Demo:

https://github.com/user-attachments/assets/96cb9071-1a72-4cff-9796-1ebe85db2dc3


tintinthong:
No test for this? Seems like an awkward behaviour that we shud test

tintinthong (changes requested):
No test for this? Seems like an awkward behaviour that we shud test

lucaslyl:
> No test for this? Seems like an awkward behaviour that we shud test

I’ve added an acceptance test to cover the listing card navigation.

packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -333,118 +334,279 @@ module('Acceptance | Catalog | catalog app tests', function (hooks) {
       await waitFor('.showcase-center-div');
     });
 
-    test('after clicking "Remix" button, the ai room is initiated, and prompt is given correctly', async function (assert) {
-      await waitFor(
-        `[data-test-card="${mortgageCalculatorCardId}"] [data-test-card-title="Mortgage Calculator"]`,
-      );
-      assert
-        .dom(
+    module('listing card', async function () {
```
tintinthong, line 337: I dont think we need this module block right. Its not too meaningful imo

packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -333,118 +334,279 @@ module('Acceptance | Catalog | catalog app tests', function (hooks) {
...
-        .exists();
-    });
 
-    test('after clicking "Preview" button, the example card opens up onto the stack', async function (assert) {
-      await waitFor(
-        `[data-test-card="${mortgageCalculatorCardId}"] [data-test-card-title="Mortgage Calculator"]`,
-      );
-      assert
-        .dom(
+      test('after clicking "Preview" button, the example card opens up onto the stack', async function (assert) {
```
tintinthong: can you add "the first example card"

packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -333,118 +334,279 @@ module('Acceptance | Catalog | catalog app tests', function (hooks) {
...
+          )
+          .exists();
+        assert
+          .dom(
+            `[data-test-stack-card="${mortgageCalculatorCardId}"] [data-test-boxel-card-header-title]`,
+          )
+          .hasText('CardListing - Mortgage Calculator');
+      });
+
+      test('carousel arrows and dots appear only when multiple images exist', async function (assert) {
```
tintinthong: 👍

tintinthong: can you make this test as "no arrows appear when one image exist"

packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -333,118 +334,279 @@ module('Acceptance | Catalog | catalog app tests', function (hooks) {
...
+            .dom(`[data-test-card="${leafletMapCardId}"] .carousel-arrow-prev`)
+            .doesNotExist();
+          assert
+            .dom(`[data-test-card="${leafletMapCardId}"] .carousel-arrow-next`)
+            .doesNotExist();
+        }
+      });
+
+      // leaflet map has 3 slides, so index 2 is the last slide
+      test('carousel navigation works correctly when triggered by arrows or dots', async function (assert) {
```
tintinthong: can you make this test as 

carousel arrows and dots appear only when multiple images exist and works when triggered

tintinthong (approved):
Coverage is good enough for me. I made suggestions of organizing that you can make or not

packages/host/tests/acceptance/catalog-app-test.gts
lucaslyl: i just noticed that everything is categorized by module (see example screenshot below), and this test focuses on actions within a **listing card**. I think this makes it clearer for users to understand which part of the app’s UI is being targeted. Does that sound good to you, or do you still prefer not to categorize it by module

image

tintinthong: The way I see it is that you are interacting with the catalog app itself. I am fine with your change also

packages/host/tests/acceptance/catalog-app-test.gts
lucaslyl: Noted. but I will change it to "no arrows and dots appear when one image exist", cause this test cover both

tintinthong (approved):

PR #2863, opened 2025-06-30
set default LLM to GPT-4.1 Nano for remix

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8684/set-default-model-for-remix

- set default LLM to GPT-4.1 Nano for remix 
- add an acceptance test after clicking "Remix" button

image

### DIFF
--- a/packages/host/app/commands/listing-action-init.ts
+++ b/packages/host/app/commands/listing-action-init.ts
@@ -1,6 +1,7 @@
 import { service } from '@ember/service';
 
 import { baseRealm, isCardInstance } from '@cardstack/runtime-common';
+import { DEFAULT_REMIX_LLM } from '@cardstack/runtime-common/matrix-constants';
 
 import * as BaseCommandModule from 'https://cardstack.com/base/command';
 
@@ -12,6 +13,7 @@ import AddSkillsToRoomCommand from './add-skills-to-room';
 import CreateAiAssistantRoomCommand from './create-ai-assistant-room';
 import OpenAiAssistantRoomCommand from './open-ai-assistant-room';
 import SendAiAssistantMessageCommand from './send-ai-assistant-message';
+import SetActiveLLMCommand from './set-active-llm';
 
 import type RealmServerService from '../services/realm-server';
 import type StoreService from '../services/store';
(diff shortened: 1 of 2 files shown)


packages/runtime-common/matrix-constants.ts
```diff
@@ -25,7 +25,7 @@ export const APP_BOXEL_HAS_CONTINUATION_CONTENT_KEY =
   'app.boxel.has-continuation';
 export const APP_BOXEL_CONTINUATION_OF_CONTENT_KEY =
   'app.boxel.continuation-of';
-export const DEFAULT_LLM = 'openai/gpt-4.1';
+export const DEFAULT_LLM = 'openai/gpt-4.1-nano';
```
tintinthong: You are changing the DEFAULT_LLM for the whole app. I DO NOT think this is the intended behaviour at all. We just want nano to be run for remix action right

tintinthong (changes requested):
You want to look at `listing-action-init`. After creating the room, call `SetActiveLLMICommand`

packages/host/app/commands/listing-action-init.ts
```diff
@@ -73,6 +74,16 @@ export default class ListingActionInitCommand extends HostBaseCommand<
       });
     }
 
+    // Set active LLM to gpt-4.1-nano for remix actions
+    if (roomId && actionType === 'remix') {
```
richardhjtan: It's fine not to check the action type, checking `roomId` will be sufficient

richardhjtan (approved):

tintinthong:
Is there an improvement we can make st if a dev mutates the DEFAULT_LLM_LIST like removing `gpt-4.1-nano`. Will he know to also update the corresponding the model inside of remix?

It seems now that if this list is mutated -- we are exposed to Error(`Cannot find LLM model: ${model}`);

lucaslyl:
> Is there an improvement we can make st if a dev mutates the DEFAULT_LLM_LIST like removing `gpt-4.1-nano`. Will he know to also update the corresponding the model inside of remix?
> 
> It seems now that if this list is mutated -- we are exposed to Error(`Cannot find LLM model: ${model}`);

Yes, that's a good catch.
I've created a dedicated DEFAULT_REMIX_LLM constant in matrix-constants.ts specifically for Remix operations and updated listing-action-init.ts to use this constant instead of the hardcoded string. This prevents the maintenance issue you identified where developers might modify DEFAULT_LLM_LIST but forget to update the Remix-specific model configuration. Does that sound good?

tintinthong:
> > Is there an improvement we can make st if a dev mutates the DEFAULT_LLM_LIST like removing `gpt-4.1-nano`. Will he know to also update the corresponding the model inside of remix?
> > It seems now that if this list is mutated -- we are exposed to Error(`Cannot find LLM model: ${model}`);
> 
> Yes, that's a good catch. I've created a dedicated DEFAULT_REMIX_LLM constant in matrix-constants.ts specifically for Remix operations and updated listing-action-init.ts to use this constant instead of the hardcoded string. This prevents the maintenance issue you identified where developers might modify DEFAULT_LLM_LIST but forget to update the Remix-specific model configuration. Does that sound good?

sounds like an ok improvement

tintinthong (approved):

PR #2865, opened 2025-06-30
Add sliding sync support to AI bot

habdelra:
This PR updates the AI bot to use sliding sync so that we can hopefully eliminate performance issues when the ai-bot boots up. On my local the AI bot boots up and responds like normal with this setup. I did have to update the `pnpm patch` so that I also set the `/_matrix/client/unstable/org.matrix.simplified_msc3575` sliding sync HTTP prefix in the `lib/` folder of the npm package--which is what node will use.


packages/ai-bot/main.ts
```diff
@@ -440,9 +446,24 @@ Common issues are:
     }
   });
 
+  let lists: Map<string, MSC3575List> = new Map();
+  lists.set(SLIDING_SYNC_AI_ROOM_LIST_NAME, {
+    ranges: [[0, 0]],
```
habdelra, line 451: I believe this means that we do not care about room history in the sync response

packages/ai-bot/main.ts
```diff
@@ -440,9 +446,24 @@ Common issues are:
     }
   });
 
+  let lists: Map<string, MSC3575List> = new Map();
+  lists.set(SLIDING_SYNC_AI_ROOM_LIST_NAME, {
```
lukemelia, line 450: Do we need to add this list at all if we don't care about the list of rooms?

lukemelia (approved):
Exciting!

packages/ai-bot/main.ts
habdelra: I was seeing that without this, the ai bot was unresponsive. I'm not finding good documentation on what exactly this is for and how to use it though...

lukemelia: Maybe @IanCal or @FadhlanR have an idea?

habdelra: based on what you set here--i think this is a list of rules, not a list of rooms.

habdelra:
the percy diffs seem significant, but i don't think they are related to this PR....

packages/ai-bot/main.ts
lukemelia: Did you try just providing an empty map?

In any case, I'm OK to merge for further review after the fact.

habdelra: I did, like I said, when this is an empty map the ai bot is unresponsive.

jurgenwerk (approved):

habdelra:
tried it out on staging--it looks really good perf wise!

PR #2866, opened 2025-07-01
add skill section for listing isolated

tintinthong:


packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -268,8 +290,30 @@ class EmbeddedTemplate extends Component<typeof Listing> {
         {{/if}}
       </section>
 
+      <section
```
lucaslyl, line 294: Everything looks good! I suggest adding a divider at the top of this section because the spacing looks off.

Screenshot 2025-07-17 at 2 05 10 PM

lucaslyl (approved):

packages/catalog-realm/catalog-app/listing/listing.gts
tintinthong: Good catch thanks.

PR #2871, opened 2025-07-01
Support cli sync using realm secret

IanCal:
Builds on #2868 

This adds support for using a realm secret rather than a password, making it easier to pull contents from realms like catalog, skills/etc.


packages/workspace-sync-cli/src/realm-sync-base.ts
```diff
@@ -378,10 +427,23 @@ export function validateMatrixEnvVars(): {
...
+    );
     process.exit(1);
   }
 
-  return { matrixUrl, username, password };
+  // If password is not provided but realm secret is, generate password from secret
+  if (!password && realmSecret) {
+    const { realmPassword } = await import(
+      '../../matrix/helpers/realm-credentials.js'
+    );
```
habdelra: hmmm, we are dynamically importing thru a different mono repo workspace. this is very awkward. let's statically import `@cardstack/runtime-common/matrix-client#passwordFromSeed` instead

habdelra (dismissed):
please statically import `passwordFromSeed()` from runtime common instead of dynamically importing it from the matrix mono repo workspace--the matrix workspace is only meant to consume dependencies, not to be consumed.

lukemelia (approved):

PR #2872, opened 2025-07-02
Flaky test for resetting state in catalog app

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-9002/flaky-test-for-resetting-state-in-catalog-app

Results:
image

### DIFF
--- a/packages/host/tests/acceptance/catalog-app-test.gts
+++ b/packages/host/tests/acceptance/catalog-app-test.gts
@@ -1,7 +1,7 @@
 import { click, waitFor, waitUntil, fillIn } from '@ember/test-helpers';
 
 import { getService } from '@universal-ember/test-support';
-import { module, skip, test } from 'qunit';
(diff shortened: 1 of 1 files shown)


lucaslyl:
@IanCal  I have fixed the flaky test '**should be reset when clicking "All Apps" button**', and it passed CI. However, another error occurred: '**Integration | card-catalog > realm filters: can filter cards by selecting a realm (Chrome)**' I think this is unrelated to my branch

packages/skills-realm/contents
IanCal, line 1: We shouldn't add this to the monorepo, I'll make sure we get a gitignore in to skip it in the future

lucaslyl: Thanks for pointing that out! I’ve reset to the latest main to remove it from the monorepo. Appreciate you mentioning this. Also, it seems like the flaky test is fixed now.

tintinthong (approved):

PR #2883, opened 2025-07-03
AI assistant: persist agent id and current room selection scoped to individual browser tabs

jurgenwerk:
This fixes the problem where if you used multiple tabs to generate AI content in multiple different rooms, the rooms annoyingly switched to another room (either to most recently selected or most recently added) when navigating or reloading the tab. 

This assures you stay in the currently selected room in each individual browser tab.

Additionally, these changes also include persisting `agentId` in a similar manner–we talked with @lukemelia that would be wise to do regarding how command results are handled.


lukemelia (approved):
I assume this means that for a brand new tab, AI Assistant will be open to an empty room? Or will it choose the most recently updated room?

jurgenwerk:
In that case, it will choose the most recently created room

PR #2885, opened 2025-07-03
Update content hash re-caching with hash validation

lukemelia:


backspace:
The Matrix test failures all look like this so it does seem broken:

Playwright Trace Viewer 2025-07-04 07-44-15

backspace (approved):

PR #2888, opened 2025-07-07
Make card error detail sticky and scrollable

FadhlanR:
In this PR, I address an issue where the card error detail could not be scrolled. I also improved its positioning — instead of placing it at the bottom, I made it sticky.

Before:

https://github.com/user-attachments/assets/4094bd5b-8aed-42dd-a150-47f08d3ee4e8

After:

https://github.com/user-attachments/assets/4f64c61c-98d2-48ed-a8d8-8ffdec0199a8


lukemelia (approved):
Definite improvement. Is there a scroll in the case of a long stack trace?

FadhlanR:
> Definite improvement. Is there a scroll in the case of a long stack trace?

Yes, there is a scroll in that case, as you can see at second 12 of the video.

PR #2891, opened 2025-07-07
AI Assistant: option to restore content of a file before code patch was applied

jurgenwerk:
New option to restore the file contents to the point before the code patch was applied:

image


lukemelia (approved):
Looks good!

PR #2892, opened 2025-07-07
Run code in realm using the authority of the realm's owner

habdelra:
This PR updates the realm such that when we run code in the realm we run it using the authority of the realm's owner via the assume-user permission granted to the realm server. This logic should be temporary. When the headless chrome logic lands we should be able to back this all out as the realm will hopefully no longer be using the loader at that point.

### DIFF
--- a/packages/runtime-common/matrix-client.ts
+++ b/packages/runtime-common/matrix-client.ts
@@ -16,7 +16,7 @@ export class MatrixClient {
   private access: MatrixAccess | undefined;
   private password?: string;
   private seed?: string;
-  private loggedIn = new Deferred<void>();
+  private loggedInDeferred: Deferred<void> | undefined;
 
   constructor({
     matrixURL,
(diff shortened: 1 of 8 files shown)


packages/runtime-common/matrix-client.ts
```diff
@@ -48,10 +48,6 @@ export class MatrixClient {
     return this.access !== undefined;
   }
 
-  async waitForLogin() {
-    return this.loggedIn.promise;
-  }
-
```
habdelra, line 54: this was never actually being consumed anywhere

packages/runtime-common/matrix-client.ts
```diff
@@ -74,6 +70,10 @@ export class MatrixClient {
   }
 
   async login() {
+    if (this.loggedInDeferred) {
+      return await this.loggedInDeferred.promise;
+    }
+    this.loggedInDeferred = new Deferred();
```
habdelra, line 76: now that we are sharing the matrix client for the realm server amongst the realms we need to be smart about logging in such that all the realms can share the work of a single login attempt as opposed to re-running the login.

lukemelia (approved):

PR #2901, opened 2025-07-08
host: Add gradients at pill menu scroll edges

backspace:
Designs show gradients when the pill menu contents have overflow:

LATEST UI Jun 17 - LATEST UI Jun 17 - Zeplin 2025-07-09 11-31-39 
LATEST UI Jun 17 - LATEST UI Jun 17 - Zeplin 2025-07-09 11-32-17

Implementation in this PR, using a similar `scroll-timeline` approach from #2881, though using background gradients instead of `box-shadow` (which I now think was less than ideal):

screencast 2025-07-09 11-34-54

The skills-chooser doesn’t actually scroll at the moment (a separate issue) but this includes handling for when there’s a footer, to reposition the gradient:

screencast 2025-07-09 11-37-42


packages/host/app/components/pill-menu/index.gts
```diff
@@ -168,12 +170,69 @@ export default class PillMenu extends Component<Signature> {
...
+          to top,
+          rgba(0, 0, 0, 0.25) 0%,
+          transparent 100%
+        );
+
+        animation: scroll-pill-menu-content reverse linear backwards;
+        animation-timeline: --pill-menu-content-scroll-timeline;
+
+        bottom: var(--boxel-sp-sm);
+        margin-bottom: 16px;
```
backspace, line 213: This and the `bottom: 60px` are more hackish than I’m generally comfortable with, there’s something about the element structure of these components I don’t understand that makes positioning difficult:

screencast 2025-07-09 11-43-49

The header and footer elements don’t have padding themselves, the container does, so the gradients have to be positioned with offsets. I plan to improve this in a followup PR.

lukemelia (approved):

PR #2905, opened 2025-07-09
atomic write endpoint

tintinthong:
https://github.com/user-attachments/assets/3f0ad6d0-13c4-4843-825a-b2ae5587f6e8

```
curl --location 'http://localhost:4201/experiments/_atomic' \
--header 'Authorization: <AUTHORIZATION_HEADER>' \
--header 'Accept: application/vnd.card+json' \
--header 'Content-Type: application/json' \
--data-raw '{
  "atomic:operations": [
    {
      "op": "add",
      "href": "./place.gts",
      "data": {
        "type": "source",
        "attributes": {
          "content": "import { field, CardDef, contains } from \"https://cardstack.com/base/card-api\";\n            import StringField from \"https://cardstack.com/base/string\";\n            export class Place extends CardDef {\n              static displayName = '\''Place'\'';\n              @field name = contains(StringField);\n            }"
        },
        "meta": {}
      }
    },
    {
      "op": "add",
      "href": "./Place/kl.json",
      "data": {
        "type": "card",
        "attributes": {
          "name": "Kuala Lumpur"
        },
        "meta": {
          "adoptsFrom": {
            "module": "../place",
            "name": "Place"
          }
        }
      }
      
    },
    {
    "op": "add",
      "href": "./Place/new-york.json",
      "data": {
        "type": "card",
        "attributes": {
          "name": "New York"
        },
        "meta": {
          "adoptsFrom": {
            "module": "../place",
            "name": "Place"
          }
        }
      }
      

    }

  ]
}'
```


packages/runtime-common/realm.ts
```diff
@@ -526,6 +551,223 @@ export class Realm {
...
+      });
+    }
+    let sourceFiles = new Map<LocalPath, string>();
+    let instanceFiles = new Map<LocalPath, string>();
+    let sourceOperations = filterOperationsWithSourceData(operations);
+    let instanceOperations = filterOperationsWithCardData(operations);
+    let instanceWriteResults: WriteResult[] = [];
+    let sourceWriteResults: WriteResult[] = [];
+
+    // we intentionally write source files first so that fileSerialization can work
```
tintinthong: Just to point out. In the case, if an instance depends on a module in its adoptsFrom, fileSerialization tries to fetch the module somehow
- Previously, we thought we should allow the that fetch to go directly to the file system rather than the index (thats its default behaviour) by maybe including a header 
- I found that `getFileWithFallbacks` still relies on the file being written first. So, what I did, is I perform a write of the sources first 
- if we want better implemnentation for batch writes, we should introduce a new `bulkWrite` api that the realm can use

packages/realm-server/tests/helpers/indexing.ts
```diff
@@ -0,0 +1,100 @@
...
+  IncrementalIndexInitiationContent,
+} from 'https://cardstack.com/base/matrix-event';
+import { validate as uuidValidate } from 'uuid';
+
+interface IncrementalIndexEventTestContext {
+  assert: Assert;
+  getMessagesSince: (since: number) => Promise<MatrixEvent[]>;
+  realm: string;
+}
+
```
tintinthong, line 16: It looks like this is actually very flaky. If I change this matrixMessages to use some version of filter and count the incremental index events, it works "sometimes"

packages/realm-server/tests/helpers/indexing.ts
```diff
@@ -0,0 +1,100 @@
...
+    let matrixMessages = await getMessagesSince(since);
+
+    return matrixMessages.some(
+      (m) =>
+        m.type === APP_BOXEL_REALM_EVENT_TYPE &&
+        m.content.eventName === 'index' &&
+        m.content.indexType === 'incremental',
+    );
+  });
+}
```
tintinthong, line 31: It looks like this is actually very flaky. If I change this matrixMessages to use some version of filter and count the incremental index events, it works "sometimes"

tintinthong: I have excluded the check for indexing from the test but indexing success doesn't seem to be a requirement for the write

tintinthong:
Perhaps, I shud prettify the module content before writing?

packages/runtime-common/realm.ts
habdelra: you could say that the onus is on the caller of this API to ensure that the operations are ordered correctly, as that is part of the JSONAPI spec--that the operations are executed in order. if the caller of the API tries to write an instance before the module that the instance depends on, that seems like a client error. We could even go so far as to provide a helper that the client can use to order the operations properly by doing a tree traversal for all the items in the batch. and then once the operations are ordered correctly the client can issue the batch operations using the suggested order provided by the helper.

tintinthong: A correct order doesn't change or improve the way the "write" implementation occurs. I still need to have 2 separate writes. First one for the source writes. Second one for the instance writes. Key reason is `fileSerialization` needing to fetch the dependent gts file. With this implementation, I do not think order matters

packages/realm-server/tests/atomic-endpoints-test.ts
```diff
@@ -0,0 +1,788 @@
...
+                      name: 'Person',
+                    },
+                  },
+                },
+              },
+            ],
+          };
+          let response = await request
+            .post('/_atomic')
+            .set('Accept', 'application/vnd.card+json')
```
lukemelia: Does this Accept header make sense for posting to this endpoint? You aren't getting card json back are you?

packages/realm-server/tests/helpers/indexing.ts
lukemelia: Perhaps what you are seeing as flaky is the cause of the bug where cards fail to update in Boxel UI after a change, until you refresh.

packages/runtime-common/realm.ts
```diff
@@ -526,6 +551,223 @@ export class Realm {
...
+      let content = resource.attributes?.content as string;
+      sourceFiles.set(localPath, content);
+    }
+
+    if (sourceFiles.size > 0) {
+      sourceWriteResults = await this.write(
+        sourceFiles,
+        request.headers.get('X-Boxel-Client-Request-Id'),
+      );
+    }
```
lukemelia: Does it mean that we have to wait for an indexing run after the source files are written before we write the card data files?

packages/runtime-common/realm.ts
```diff
@@ -526,6 +551,223 @@ export class Realm {
...
+      meta: {
+        created,
+      },
+    }));
+    return createResponse({
+      body: JSON.stringify({ 'atomic:results': results }, null, 2),
+      init: {
+        status: 201,
+        headers: {
+          'content-type': SupportedMimeType.CardJson,
```
lukemelia: I think this should likely be `application/vnd.api+json` throughout

lukemelia (changes requested):

packages/realm-server/tests/atomic-endpoints-test.ts
tintinthong: Indeed it doesn make sense

packages/runtime-common/realm.ts
tintinthong: Yes. This is this behavior unfortunately. A failure to write one of the modules will stop the writes of ALL the instances. And modules will remain written as there is no cleanup currently. 

Do you have a proposed way to resolve this? Essentially, `fileSerialization` is the method that creates this dependence between instance and their modules. 

**Sanity-check**

I think we already wait for indexing of this already bcos the `update` method in realm-index-updated waits for the `job.done`. Otherwise, we'd have an issue of indexing of the module being incomplete before trying to write the instance.

lukemelia: @habdelra thoughts?

tintinthong: Just to recap the step forward in meeting with @lukemelia @habdelra and @IanCal 
- we introduce a new `.bulkWrite` method inside of the realm whereby, a list of files writes occur one by one before a final indexing step 
- update `write` as a singular method that delegates bulkWrite
- reduce the endpoint handler to call bulkWrite and move its implementation of indexing inside internal bulkWrite. It can remain to handle other endpointy stuff
- `fileSerialization` issues ie a module being dependent on a module becomes the onerous of the client to maintain a correct order. But, we still need to bypass the calls to the module index by accessing the file on disk (we shud use a special header for this). We can create a helper to ensure the order is correct

tintinthong: > fileSerialization issues ie a module being dependent on a module becomes the onerous of the client to maintain a correct order. But, we still need to bypass the calls to the module index by accessing the file on disk (we shud use a special header for this). We can create a helper to ensure the order is correct

This doesn't seem to be an issue if `fileSerialization` is handled internally. It seems that it DOES fallback to the file so we are good here. No need for any headers. It tries to get from index and then otherwise it gets from file.

packages/runtime-common/realm.ts
```diff
@@ -488,34 +503,51 @@ export class Realm {
...
       let url = this.paths.fileURL(path);
-      this.sendIndexInitiationEvent(url.href);
+      this.sendIndexInitiationEvent(url.href); //check if this does anything
       await this.trackOwnWrite(path);
-      let { lastModified } = await this.#adapter.write(path, content);
-      results.push({ path, lastModified });
+      try {
+        let doc = JSON.parse(content);
+        if (isCardResource(doc.data)) {
+          let serialized = await this.fileSerialization(
```
tintinthong, line 538: I moved `fileSerialization` into an internal implementation. There are consequences of this which to me make more sense. But open to debate. Since with SRP principles, we might want write to just write but that is not what we are doing with the current implementation anyway

packages/runtime-common/realm.ts
```diff
@@ -524,13 +556,214 @@ export class Realm {
...
+            headers: { 'content-type': SupportedMimeType.JSONAPI },
+          },
+          requestContext,
+        });
+      }
+    }
+
+    if (files.size > 0) {
+      try {
+        writeResults = await this.writeMany(
```
tintinthong: A single call to `writeMany` of the realm

packages/runtime-common/realm.ts
```diff
@@ -488,34 +503,51 @@ export class Realm {
...
+        }
+      }
+      let { lastModified, created, isNew } = await this.#adapter.write(
+        path,
+        content,
+      );
+      results.push({ path, lastModified, created, isNew });
       urls.push(url);
     }
     await this.#realmIndexUpdater.update(urls, {
```
tintinthong, line 556: A single indexing run for `writeMany`

packages/host/tests/integration/commands/transform-cards-test.gts
```diff
@@ -359,20 +341,14 @@ module('Integration | commands | transform-cards', function (hooks) {
     );
     let fluffyContent = await response.text();
     let fluffyData = JSON.parse(fluffyContent);
-    assert.strictEqual(
-      fluffyData.data.attributes.metadata,
-      'Added by TransformCardsCommand',
-    );
+    assert.strictEqual(fluffyData.data.attributes.name, 'Transformed: Fluffy');
```
tintinthong: this suggest one of the consequences of `write` handling `fileSerialization` internally. Transformation of the cards will break if it doesn't comply with the type of the card

packages/host/tests/acceptance/code-submode/editor-test.ts
```diff
@@ -384,8 +384,38 @@ module('Acceptance | code submode | editor tests', function (hooks) {
       throw new Error('file not found');
     }
     assert.deepEqual(
```
tintinthong: Here is another test where the changes were serialized but previously it didnt seem to be

packages/realm-server/node-realm.ts
```diff
@@ -173,15 +176,18 @@ export class NodeAdapter implements RealmAdapter {
     };
   }
```
tintinthong, line 178: `.write` method in any of the adapters is dealing with the filesystem write exclusively. There is no bulk method here.

habdelra (approved):

packages/runtime-common/realm.ts
```diff
@@ -524,13 +559,214 @@ export class Realm {
...
+          requestContext,
+        });
+      }
+    }
+
+    if (files.size > 0) {
+      try {
+        writeResults = await this.writeMany(files, {
+          clientRequestId: request.headers.get('X-Boxel-Client-Request-Id'),
+          serializeFile: true,
```
tintinthong, line 732: its easier to handle `serializeFile` internally bcos if we read from the header, its really not that clear weather an `_atomic` write is providing source or card. So we just include an option `serializeFile` to instruct all card resources to be serialized.

tintinthong:
@lukemelia this PR is awaiting your review

lukemelia (approved):

packages/realm-server/tests/atomic-endpoints-test.ts
```diff
@@ -0,0 +1,916 @@
...
+          let response = await request
+            .post('/_atomic')
+            .set('Accept', SupportedMimeType.JSONAPI)
+            .set(
+              'Authorization',
+              `Bearer ${createJWT(testRealm, 'user', ['read', 'write'])}`,
+            )
+            .send(JSON.stringify(doc));
+
+          console.log('response body', response.body);
```
lukemelia: Remove this and similar logging from the other tests in this file.

PR #2907, opened 2025-07-09
Add error handling to patching code with results reporting

lukemelia:
- Updated `PatchCodeCommand` to return detailed results for each code block, indicating whether the patch was applied or failed.
- Enhanced `apply-search-replace-block` command to throw an error if a patch does not apply cleanly.
- Modified the `matrix-event` interface to include an optional `failureReason` for failed patches.
- Updated tests to cover scenarios for both successful and failed code patches, ensuring proper event dispatching and state updates.
- Reflect patch results accurately in the UI.

The end result of this work is that the aibot should now know about failed patches and respond to them accordingly.


packages/runtime-common/commands.ts
```diff
@@ -95,7 +95,8 @@ export abstract class Command<
         inputCard = new InputType(input) as CardInstance<CardInputType>;
       }
     }
-    return this.run(inputCard);
+    let result = await this.run(inputCard);
+    return result;
```
backspace: Is this an artefact of having logged the result?

backspace (approved):

packages/runtime-common/commands.ts
lukemelia: Something like that. Rolled back this change.

packages/host/app/commands/patch-code.ts
```diff
@@ -87,19 +100,32 @@ export default class PatchCodeCommand extends HostBaseCommand<
...
+        let { resultContent } = await applyCommand.execute({
+          fileContent: content,
+          codeBlock: codeBlock,
+        });
+        content = resultContent;
+        results.push({ status: 'applied' });
+      } catch (error) {
+        results.push({
+          status: 'failed',
+          failureReason: error instanceof Error ? error.message : String(error),
```
jurgenwerk, line 122: Do we know what are some common failure reasons here?

jurgenwerk (approved):

packages/host/app/commands/patch-code.ts
lukemelia: The only one I know of is when the search term is not found in the doc.

PR #2911, opened 2025-07-09
Escape dollar sign to workaround glint bug

lukemelia:


habdelra:
should we fix the lint errors too?

lukemelia:
Lint errors fixed, though I don't understand how they got in in the first place.

Update: Oh, because the file wasn't getting checked due to the error.

habdelra (approved):

PR #2913, opened 2025-07-09
host: Update skills list design

backspace:
Design:

LATEST UI Jun 17 - LATEST UI Jun 17 - Zeplin 2025-07-10 13-16-09

Current:

index json in Experiments Workspace 2025-07-10 13-16-41

With this PR, including the gradients on overflow:

index json in Experiments Workspace 2025-07-10 13-18-13 index json in Experiments Workspace 2025-07-10 13-30-21

A noteworthy discrepancy is that there’s currently no way to designate an icon for a skill, so this excludes that part of the design. There’s an issue for adding that.


packages/host/app/components/card-pill.gts
```diff
@@ -81,18 +79,6 @@ export default class CardPill extends Component<CardPillSignature> {
...
-                @isEnabled={{@isEnabled}}
-                @onChange={{@onToggle}}
-                @label={{this.card.title}}
-                data-test-card-pill-toggle='{{@cardId}}-{{if
-                  @isEnabled
-                  "on"
-                  "off"
-                }}'
-              />
-            {{/if}}
```
backspace, line 95: The skills menu was the only use of `CardPill` with a toggle, which IMO was already overly stretched, and moving the toggles and icons outside of it made that moreso. So I extracted the relevant parts into `SkillToggle`.

packages/matrix/tests/skills.spec.ts
```diff
@@ -65,8 +65,8 @@ test.describe('Skills', () => {
       page.locator(`[data-test-pill-menu-item="${cardId}"]`),
     ).toHaveCount(1);
     await expect(
-      page.locator(`[data-test-card-pill-toggle="${cardId}-on"]`),
-    ).toHaveClass('switch checked');
+      page.locator(`[data-test-skill-toggle="${cardId}-on"]`),
+    ).toContainClass('checked');
```
backspace, line 69: The version update for Playwright is to enable this assertion, which lets us specifically address the relevant `checked` class and not need to include whatever else the element has (now `toggle` as well).

lukemelia (approved):

PR #2920, opened 2025-07-10
host: Update chat input design

backspace:
Designs:

LATEST UI Jun 17 - LATEST UI Jun 17 - Zeplin 2025-07-10 16-34-22

In order to meet the requirement of adding the divider when the input exceeds 150px, I removed the modifier that set element height by estimating how many lines the input value would render as, as it was not calculating reliably, causing the divider to appear and disappear as the height was set by the modifier:

screencast 2025-07-11 11-56-29

That technique looks similiar to [this](https://css-tricks.com/auto-growing-inputs-textareas/#aa-resizing-actual-input-elements).

Instead, the textarea is now contained within an element that also has a clone of the input value but in a `div`, which grows with the value automatically. Grid properties cause the two elements to render on top of each other, and `visibility: hidden` hides the clone element. It’s adapted from [this technique](https://css-tricks.com/the-cleanest-trick-for-autogrowing-textareas/), there’s no need to hook into `input` events to copy the value via `dataset` as Ember lets us just use the same `@value` for both elements.

See the divider appear and disappear here based on the textarea overflowing:

screencast 2025-07-11 11-45-43

### DIFF
--- a/packages/host/app/components/ai-assistant/attachment-picker/attached-items.gts
+++ b/packages/host/app/components/ai-assistant/attachment-picker/attached-items.gts
@@ -131,7 +131,6 @@ export default class AttachedItems extends Component<Signature> {
     </div>
     <style scoped>
       .attached-items {
-        border-top: 1px solid var(--boxel-400);
         background-color: var(--boxel-light);
         color: var(--boxel-dark);
         display: flex;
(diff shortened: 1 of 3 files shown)


burieberry (approved):
This autogrow feature would be a nice option to add to our boxel textarea component too.

habdelra (approved):

backspace:
> This autogrow feature would be a nice option to add to our boxel textarea component too.

Thanks for the suggestion, I tried doing that [here](https://github.com/cardstack/boxel/commit/1c5d5861e3df76709b32899af2514e53791e31b2) but gave up because I couldn’t get proper alignment for the placeholder or when the textarea was one line:

Experiments Workspace 2025-07-14 12-44-24 
Experiments Workspace 2025-07-14 12-45-13

Maybe there’s a way, if we need it elsewhere this attempt can be salvaged.

PR #2921, opened 2025-07-11
Add displayName with counter inside of cards grid LHS

tintinthong:
**Old**
Screenshot 2025-07-12 at 12 29 51

**New**
Screenshot 2025-07-12 at 12 25 03

**Change**
- we add a counter to duplicate display names
- the latest indexed will be the ones that have counters. this uses the earliest index time across code ref types to decide
- one thing it doesn't seem to cover is the icon to the left of the display name so counter will appear even tho its different icon
- this counter is not unique and persistent. So we dont expect if Author2 is seen to always exist if Author1's display name changes; Author2 will become Author 1

### DIFF
--- a/packages/runtime-common/index-query-engine.ts
+++ b/packages/runtime-common/index-query-engine.ts
@@ -625,6 +625,7 @@ export class IndexQueryEngine {
        FROM realm_meta rm
        WHERE`,
       ...every([['rm.realm_url =', param(realmURL.href)]]),
+      `ORDER BY indexed_at DESC`,
     ] as Expression)) as Pick<RealmMetaTable, 'value'>[];
 
     return (results[0]?.value ?? []) as unknown as CardTypeSummary[];
--- a/packages/base/cards-grid.gts
+++ b/packages/base/cards-grid.gts
@@ -224,6 +224,7 @@ class Isolated extends Component<typeof CardsGrid> {
       id: string;
       attributes: {
         displayName: string;
+        uniqueDisplayName: string;
         total: number;
         iconHTML: string | null;
       };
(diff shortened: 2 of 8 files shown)


packages/runtime-common/index-query-engine.ts
```diff
@@ -625,6 +625,7 @@ export class IndexQueryEngine {
        FROM realm_meta rm
        WHERE`,
       ...every([['rm.realm_url =', param(realmURL.href)]]),
+      `ORDER BY indexed_at DESC`,
```
tintinthong, line 628: This change here was needed. On main, if I restarted the server, the LHS will show the stale realm meta. This change guarantees we take the latest

habdelra: why don't you use a `COUNT 1` so that you can optimize plucking the first value in SQL instead of in javascript (query should be faster that way). it also reads better too so we can see why you are sorting

habdelra:
is the assignment of these number consistent? it looks like we are ordering by indexed_at, right? does that mean if I have an incremental index that effects the card def with the dupe display name, will that reassign the number?

tintinthong:
> is the assignment of these number consistent? it looks like we are ordering by indexed_at, right? does that mean if I have an incremental index that effects the card def with the dupe display name, will that reassign the number?

yes. that is something I am aware about. Essentially, if you have the following folders 

Author: jane, john
Author 1: juniper
Author 2: jeffrey, jim

If Author 1 display name  in its gts got changed from Author to Writer

then we will get 

Author jane, john
Author 1:  jeffrey, jim  //The Author 1 directory changed 
Writer: juniper 

Do we want persistent state of the assigned folder name? If so, id really need to re-think this alot more

packages/runtime-common/index-query-engine.ts
tintinthong: Ok I will try to optimise this

habdelra:
i think so, otherwise these numbers will jump around and Author 1 might become Author 2, and then switch back to Author 1. The UI will become inconsistent. perhaps we can lexically sort on the coderefs instead, since that will guaranteed to be different and stable

PR #2945, opened 2025-07-16
Move profile avatar icon from lower left to upper right

FadhlanR:
https://github.com/user-attachments/assets/fa455250-34f5-4fa8-942e-b11dc00e8793


lukemelia (approved):

jurgenwerk (approved):
Tested it in staging preview and it looks good

PR #2948, opened 2025-07-16
Cs-9023-enhance-listing-install-to-use-bulk-write-endpoint

tintinthong:


packages/host/app/services/playground-panel-service.ts
```diff
@@ -60,8 +60,9 @@ export default class PlaygroundPanelService extends Service {
     cardId: string,
     format: Format,
     fieldIndex: number | undefined,
+    codePath?: string,
   ) => {
-    let url = this.operatorModeStateService.codePathString;
+    let url = codePath ?? this.operatorModeStateService.codePathString;
```
tintinthong: shud revert this

tintinthong:
I somehow managed to break all atomic endpoint test

richardhjtan:
Much cleaner now thanks to the atomic endpoint @tintinthong

richardhjtan:
Feedback:
Task planner project instance has same UUID when copied (which is expected), but there is error about instance not found.
Screenshot 2025-07-16 at 3 39 29 PM

tintinthong:
> Feedback: Task planner project instance has same UUID when copied (which is expected), but there is error about instance not found. Screenshot 2025-07-16 at 3 39 29 PM

hmmm i dont get that error @richardhjtan

packages/host/app/commands/listing-install.ts
```diff
@@ -2,183 +2,155 @@ import { service } from '@ember/service';
...
+        let { sourceCard, targetCodeRef } = copyInstanceMeta;
+        let doc: LooseSingleCardDocument =
+          await this.cardService.serializeCard(sourceCard);
+        let cardResource = doc.data;
+        delete cardResource.id;
+        cardResource.lid = copyInstanceMeta.lid;
+        if (targetCodeRef) {
+          cardResource.meta.adoptsFrom = targetCodeRef;
+          cardResource.meta.realmURL = realmUrl;
+        }
```
tintinthong, line 123: Here. we need not delete relationships.data as they are done during fileSerialization. But, I think the complexity of demolishing a sourceDirectory is biting us in the butt. Bcos now links.self are relative with a folder prefixed that was removed. I guess we can resolve it but it is more relative path work

richardhjtan (approved):

packages/host/app/commands/listing-install.ts
tintinthong: I was digging. Actually a lot of these can be removed 

`cardResource.id` ,`cardResource.lid` , `cardResource.meta.realmURL`. They are removed from an internal fileSerialization handled by server. Im ok with this for now, but I can revisit this when I handle the relationships

PR #2950, opened 2025-07-16
AI bot: prevent duplicate event processing when multiple bots are running, add graceful shutdown

jurgenwerk:
This PR focuses on solving 2 problems in the AI bot, which were revealed to us at the time of a bot deploy, when there is a brief time period when there are 2 instances running (the old one and the new one). In that time, user got duplicated responses because both instances responded to the same user input, and the streaming from the old instance got terminated abruptly. 

## Graceful shutdown

In our deployed environments, when a new deploy happens and the old instance gets terminated, it is possible that the old instance is still processing (streaming) a response to a user event. In this case we now try to wait for that to finish. 

image

## Event locking (only one bot instance should process user's input)

When a bot starts responding to user's request (matrix event), it leverages a new database table (`ai_bot_event_processing`) where it writes a row that acts as a lock for that event. This makes it so that no other bot instance is allowed to respond to that event. 

image


packages/ai-bot/main.ts
```diff
@@ -45,6 +46,11 @@ import { PgAdapter } from '@cardstack/postgres';
 import { ChatCompletionMessageParam } from 'openai/resources';
 import { OpenAIError } from 'openai/error';
 import { ChatCompletionStream } from 'openai/lib/ChatCompletionStream.mjs';
+import { acquireLock, releaseLock } from './lib/queries';
+import { logger as matrixLogger } from 'matrix-js-sdk/lib/logger';
+
+// Silence FetchHttpApi Matrix SDK logs
+matrixLogger.setLevel('warn');
```
jurgenwerk, line 55: I did this because the output is very noisy:

image


So I adjusted the log level to only output warnings and errors

packages/ai-bot/main.ts
```diff
@@ -471,3 +504,54 @@ Common issues are:
...
+
+  log.info('Shutting down...');
+
+  waitForActiveGenerationsPromise = waitForActiveGenerations();
+  await waitForActiveGenerationsPromise;
+  waitForActiveGenerationsPromise = undefined;
+}
+
+async function waitForActiveGenerations() {
+  let minutes = 5;
```
jurgenwerk: This reminds me we need to match the timeout in the infra for process termination timeout

packages/ai-bot/main.ts
```diff
@@ -231,6 +250,17 @@ Common issues are:
           return;
         }
 
+        let eventLock = await acquireLock(
```
jurgenwerk: In case there are many bot instances, only the first one that succeeds writing a row to this table will acquire the lock for this event. The operation is atomic and with a constraint so there is no chance of multiple bots getting the lock for the same event

jurgenwerk:
Looks like the test error is coming from upstream and not from this PR (@lucaslyl please check  https://github.com/cardstack/boxel/actions/runs/16313839549/job/46074907319)

lucaslyl:
> Looks like the test error is coming from upstream and not from this PR (@lucaslyl please check https://github.com/cardstack/boxel/actions/runs/16313839549/job/46074907319)

Thanks for pointing that out! 🙏 I’ve fixed this test error with this pr: https://github.com/cardstack/boxel/pull/2954

packages/ai-bot/main.ts
```diff
@@ -211,7 +222,15 @@ Common issues are:
   client.on(
     RoomEvent.Timeline,
     async function (event, room, toStartOfTimeline) {
+      let eventId = event.getId()!;
```
lukemelia, line 232: Before we abandon the event due to isShuttingDown or failure to acquire a lock, I think we need to check to see if this event should abort a response that this aibot instance has in progress.

lukemelia (changes requested):
See inline comment

packages/ai-bot/main.ts
jurgenwerk: Nice catch, I addressed this in the new commits, and tested manually (stopping the AI generation works)

jurgenwerk:
@lukemelia ready for another review, I also spotted a problem of where I initially put the lock releasing - it was placed in a wrong place, outside of the async boundary (this problem became really apparent when I tested this with 5 running bots 😅)

lukemelia (approved):

lukemelia:
@jurgenwerk I thought I left this comment somewhere, but just in case: if the user sends Ctrl+C after the process is already handling shutdown, it should quit immediately. This will make for a better development experience.

jurgenwerk:
Good idea, I added this functionality (1st ctrl+c will wait for things to finish, 2nd will terminate it immediately):

```
Gracefully shutting down... (Press Ctrl+C again to force exit)
Shutting down...           <---- Here I pressed ctrl+c again
Exiting immediately...             
```

Will merge this tomorrow morning so I can stay put and see if it works ok in deployed environments

PR #2959, opened 2025-07-16
migrate old users into the users table in prod

habdelra:
This PR uses a list of users from the production synapse admin console and manually adds any users that don't already exist into the users table. I've omitted all the realm bot users from this list. I did confirm that on new user registration we are always creating an entry for a matrix user id--there is no stripe requirement for making a user in this table.


packages/postgres/migrations/1752685843945_migrate-old-users.js
```diff
@@ -0,0 +1,107 @@
+const users = [
```
habdelra: This list comes from the prod synapse admin console

jurgenwerk (commented):
~~I find it odd that there are so many accounts without a users table entry 😵‍💫 Is it possible that these users did not validate their email? Because in our current logic, we create a user account (in `this.matrixService.initializeNewUser`) only after they validate their email.~~

edit: I see it's just a list of all prod users (I initally thought these were fetched by using some filters) and it tries to create one if it doesn't exist yet.

packages/postgres/migrations/1752685843945_migrate-old-users.js
```diff
@@ -0,0 +1,102 @@
...
+  '@jacthong:boxel.ai',
+  '@janderson23:boxel.ai',
+  '@jbrown44:boxel.ai',
+  '@jhall77:boxel.ai',
+  '@jtjr:boxel.ai',
+  '@jurgen1:boxel.ai',
+  '@jurgen5:boxel.ai',
+  '@jurgen:boxel.ai',
+  '@jurgenwerk2:boxel.ai',
+  '@jurgenwerk:boxel.ai',
```
jurgenwerk, line 49: This is one of my usernames and actually there already is a user table entry for this username

jurgenwerk: Ah sorry I see now, it's just a list of all prod users. All good

jurgenwerk (approved):

packages/postgres/migrations/1752685843945_migrate-old-users.js
habdelra: yeah, i literally just got a csv of all the prod matrix usernames and added them. the script just skips over any ones that already are in the users table.

habdelra:
i've run this branch in prod, the missing users have been added

PR #2963, opened 2025-07-16
BoxelTag component and Base Tag card

burieberry:
Changes:
- Add BoxelTag component, which uses Pill component with styling
- Update Pill component for easier override of border and font color if desired
- Update former CardContentContainer component as BoxelContainer
- Use BoxelTag component in `tag.gts` in base-realm
- Add spec for tag.gts in base
- Use base-realm tag card in sprint-planner, with customized atom format


packages/boxel-ui/addon/src/components.ts
```diff
@@ -6,7 +6,7 @@ import BasicFitted from './components/basic-fitted/index.gts';
 import Button from './components/button/index.gts';
 import BoxelButton from './components/button/index.gts';
 import CardContainer from './components/card-container/index.gts';
-import CardContentContainer from './components/card-content-container/index.gts';
+import BoxelContainer from './components/card-content-container/index.gts';
```
backspace: Should the path be changed since it’s renamed here and in the file?

backspace (approved):
I tried this locally and got a weird crash creating a new tag but I don’t think that’s from this PR

PR #2967, opened 2025-07-17
fix remix for sprint planner examples have broken links

tintinthong:
Remove splatting logic for source directory during installation and refactoring of classes


tintinthong:
Discussed with @christse. 

What we will do is 

- dont resolve relationships 
- keep folder structure similar to catalog. dont splat the folders anymore
- no example links will break

Todo 
- remove code that does this heuristic to splat folders 


cc @lukemelia

tintinthong:
@richardhjtan I fixed to use json fetches

richardhjtan (approved):
It works!

PR #2971, opened 2025-07-18
fix flaky text

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-9083/should-be-reset-when-clicking-catalog-home-button-test-is-flaky

image

### DIFF
--- a/packages/host/tests/acceptance/catalog-app-test.gts
+++ b/packages/host/tests/acceptance/catalog-app-test.gts
@@ -1,7 +1,7 @@
 import { click, waitFor, waitUntil, fillIn } from '@ember/test-helpers';
 
 import { getService } from '@universal-ember/test-support';
-import { module, skip, test } from 'qunit';
+import { module, test } from 'qunit';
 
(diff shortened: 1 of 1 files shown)


packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -721,13 +720,13 @@ module('Acceptance | Catalog | catalog app tests', function (hooks) {
...
-          .dom('[data-test-showcase-view]')
-          .doesNotExist('Should be in list view after applying filter');
-
         await click('[data-test-navigation-reset-button="showcase"]');
 
+        await waitUntil(() => {
+          const cards = document.querySelectorAll('[data-test-showcase-view]');
+          return cards.length === 1;
+        });
+
```
habdelra, line 729: instead of doing this can we add a test waiter around the logic for showing the showcase view so that QUnit naturally waits for that promise?

tintinthong:
im handling these tests inside my PR

PR #2973, opened 2025-07-19
remove timeout for remix

tintinthong:
waiting to check if test fails. 1s doesn't seem like long so I doubt it makes a difference.


richardhjtan (approved):
That was catered for runtime user use case, I tried this PR with the latest `main` - it works fine now without waiting the timeout

habdelra (approved):

PR #2974, opened 2025-07-21
Handle card errors by removing the hydration logic when isError is true

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-9101/hydration-makes-a-card-with-error-doc-disappear

https://github.com/user-attachments/assets/f1e80940-59b1-49fa-b62e-6db9637ddd04


tintinthong:
why isnt the instance errors already served from `fitted_html` from the index?

tintinthong:
> why isnt the instance errors already served from `fitted_html` from the index?

Oh my bad. I can see that it is.

tintinthong (approved):

PR #2975, opened 2025-07-21
allow clicking auto-attached items to convert them to chosen items

FadhlanR:
- [x] Clicking on an auto-attached item (dashed border) converts it to a chosen item (solid border)
- [x] The item remains attached but is no longer auto-attached
- [x] Remove button continue to work correctly

https://github.com/user-attachments/assets/785482e4-2e0b-42d2-b2cf-67cc39189115


backspace (approved):

packages/host/app/components/card-pill.gts
```diff
@@ -62,19 +63,43 @@ export default class CardPill extends Component<CardPillSignature> {
...
+      this.args.removeCard(this.args.cardId);
+    }
+  }
+
+  private get pillKind() {
+    return this.args.isAutoAttachedCard && this.args.chooseCard
+      ? 'button'
+      : 'default';
+  }
+
```
lukemelia: It seems like CardPill is getting too much specific knowledge about attached cards in it. Let's refactor to make it more of a UI component and not have isAutoAttachedCard / chooseCard. It should be told whether to use a dottedBorder and what to do onClick.

PR #2976, opened 2025-07-21
Fix: Apply grid template column css layout in editor only when user has write permissions

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-9119/ui-issue-with-whole-div-when-accessing-edit-mode-in-read-only-realm

As a result:
When users did not have write permissions, the icon buttons were not displayed, but the grid still reserved space for them.
This led to extra empty space on either side of the main content, causing the layout to appear misaligned or less visually balanced in read-only mode.
**The improvement:**
By applying the **grid-template-columns** style only when the user has write permissions (using a conditional class such as .can-write), the layout now adapts appropriately:

- Three columns are used when the icon buttons are present (editable state).
- A single column is used when the buttons are absent (read-only state).

This approach ensures a consistent and visually balanced layout for all users.

Before:
Screenshot 2025-07-21 at 1 59 40 PM

After:
Screenshot 2025-07-21 at 2 05 57 PM

### DIFF
--- a/packages/base/contains-many-component.gts
+++ b/packages/base/contains-many-component.gts
@@ -74,7 +74,7 @@ class ContainsManyEditor extends GlimmerComponent<ContainsManyEditorSignature> {
           >
             {{#each @arrayField.children as |boxedElement i|}}
               <li
-                class='editor'
+                class='editor {{if permissions.canWrite "can-write"}}'
                 data-test-item={{i}}
                 {{sortableItem
                   groupName=this.sortableGroupId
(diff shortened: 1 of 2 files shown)


backspace (approved):
I’m unsure if this is a preexisting problem but I’m surprised about the left alignment of “Specs”, why is it outdented compared to the other fields?

image

lucaslyl:
> I’m unsure if this is a preexisting problem but I’m surprised about the left alignment of “Specs”, why is it outdented compared to the other fields?
> 
> image

Yes, I also noticed this and have filed a ticket to address it. **https://linear.app/cardstack/issue/CS-9120/adjust-field-alignment-in-edit-mode-read-only-state**. It seems to be an existing layout behavior where others fields have extra padding left/right behavior except links-to-many-editor/contains-many-editor.

image

PR #2980, opened 2025-07-21
make auto save occur only when write access is available

tintinthong:
we get 403 forbidden when people are playing with examples inside the catalog realm that essentially write to card state using `this.args.model.set` 

Screenshot 2025-07-21 at 21 00 08

Screenshot 2025-07-21 at 20 59 26

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -836,6 +836,13 @@ export default class StoreService extends Service implements StoreInterface {
     } else {
       instance = idOrInstance;
     }
+    let realmToWrite = instance[realmURLSymbol]?.href;
+    let permissionToWrite = realmToWrite
+      ? this.realm.permissions(realmToWrite).canWrite
+      : false;
+    if (!permissionToWrite) {
+      return;
+    }
     let autoSaveState = this.initOrGetAutoSaveState(instance);
     let queueName = instance.id ?? instance[localIdSymbol];
     let autoSaveQueue = this.autoSaveQueues.get(queueName);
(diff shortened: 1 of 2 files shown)


packages/host/app/services/store.ts
```diff
@@ -823,6 +823,23 @@ export default class StoreService extends Service implements StoreInterface {
     }
   }
 
+  // this function is used to determine if the instance will be auto-saved or not
+  // this is a temporary function that is likely to go away with the creation of completion emphemeral state solution of the store/realm
+  // the only use-case for this function is determining if a preview instance in catalog realm (which is a read-only),
+  // st a card can be mutable without persisting to the server
+  private useEphemeralState(instance: CardDef | undefined): boolean {
```
tintinthong, line 830: perhaps, another naming? but I intentionally name this to express the intent of "where we r going"

packages/host/tests/integration/store-test.gts
```diff
@@ -776,6 +778,24 @@ module('Integration | Store', function (hooks) {
     (instance as any).name = 'Air';
   });
 
+  test<TestContextWithSave>('an instance will NOT auto save when its data changes, if the user does not have write permissions', async function (assert) {
+    await realmService.logout(); //logout means store does not have write access to the realm
```
habdelra: this seems like an awkward way of saying this. it's not inconceivable that there might be other side effects of logging out that would impact the store. can we just mock the realm service so that this would indicate that this is read only so we hold everything else about the realm service constant.

habdelra: Like just set our mocked `RealmService.permissions()` to always return `{ canRead: true, canWrite: false}`.

habdelra (approved):

PR #2981, opened 2025-07-21
Fix bug around serialization/deserialization of polymorphic primitives

habdelra:
This PR fixes a bug where we weren't serializing and deserializing polymorphic primitives correctly.


lukemelia (commented):
Was eliminating the typescript constructor shorthand part of the bug fix?

habdelra:
> Was eliminating the typescript constructor shorthand part of the bug fix?

yeah, but specifically, the change was more about using an object to set the constructor args instead of a list of items.
anytime there are more than 3 params in a function call I stop passing it in as a list of args as its too much to keep track of what is doing what. sadly the constructor shorthand doesn't support this, but i prefer the clarity with the arguments over the shorthand . and in this case both the last booleans are optional, but making it a list of args with 2 optional params at the end is especially confusing:

so this:
```ts
new Field({
  cardThunk: () => card, 
  computeVia, 
  name, 
  description, 
  isUsed: true, 
  isPolymorphic: true
})
```

vs this:
```ts
new Field(() => card, computeVia, name, description, true, true)
```

backspace (approved):

PR #2989, opened 2025-07-22
Implement ask vs act toggle

FadhlanR:
This PR adds a new LLM mode toggle component that allows users to switch between 'ask' and 'act' modes. The default mode is 'ask' mode. If user switch to 'act' mode, commands and code patches are automatically executed without requiring explicit approval, but only for commands and code patches sent after the mode was enabled.

https://github.com/user-attachments/assets/f589f695-7da6-4564-a019-02f9c9cc2a04


packages/ai-bot/lib/responder.ts
```diff
@@ -33,10 +34,40 @@ export class Responder {
...
+        lastLLMModeEvent &&
+        (lastLLMModeEvent as any).content &&
+        (lastLLMModeEvent as any).content.mode === 'act'
+      ) {
+        return true;
+      }
+      return false;
+    }
+    // Fallback: do not respond to command/code patch result events by default
+    return false;
```
lukemelia: I'm not convinced this is the logic we want. Let's discuss it.

packages/host/app/services/matrix-service.ts
```diff
@@ -1592,6 +1598,22 @@ export default class MatrixService extends Service {
...
+    // Queue code patches for processing
+    if (
+      event.type === 'm.room.message' &&
+      event.content?.body &&
+      event.content?.isStreamingFinished
+    ) {
+      // Check if the message contains code patches by looking for search/replace blocks
+      let body = event.content.body as string;
+      if (
+        body.includes('╔═══ SEARCH ════╗') &&
```
jurgenwerk: We have this in constants

packages/host/app/components/ai-assistant/llm-mode-toggle.gts
```diff
@@ -0,0 +1,87 @@
...
+    disabled?: boolean;
+    onExpand?: () => void;
+    onCollapse?: () => void;
+  };
+  Element: HTMLElement;
+}
+
+export default class LLMModeToggle extends Component<Signature> {
+  <template>
+    <div class='llm-mode-toggle' ...attributes>
```
backspace, line 24: I know this probably isn’t in the design but maybe some hover text on this would help explain what the toggle does for the unfamiliar

packages/host/app/resources/room.ts
```diff
@@ -386,6 +389,71 @@ export class RoomResource extends Resource<Args> {
...
+    if (this.activeLLMMode === mode) {
+      return;
+    }
+    this.llmModeBeingActivated = mode;
+    try {
+      if (!this.matrixRoom) {
+        throw new Error('matrixRoom is required to activate LLM mode');
+      }
+      await this.matrixService.sendLLMModeEvent(this.matrixRoom.roomId, mode);
+      let remainingRetries = 20;
```
jurgenwerk, line 444: Why is retrying needed?

jurgenwerk (approved):
I have tested this locally and the act mode works nicely for the cases I tried.

packages/host/app/services/command-service.ts
```diff
@@ -142,14 +170,126 @@ export default class CommandService extends Service {
...
+      let timeout = Date.now() + 60_000; // reset the timer to avoid a long wait if the room resource is processing
+      let currentRoomProcessingTimestamp = roomResource.processingLastStartedAt;
+      while (
+        roomResource.isProcessing &&
+        currentRoomProcessingTimestamp ===
+          roomResource.processingLastStartedAt &&
+        Date.now() < timeout
+      ) {
+        // wait for the room resource to finish processing
+        await delay(100);
```
jurgenwerk, line 228: Is this always enough of a delay? Is it possible to use a more reliable mechanism to check for readiness?

packages/host/app/services/command-service.ts
```diff
@@ -142,14 +170,126 @@ export default class CommandService extends Service {
...
+      let messageTimestamp = message.created.getTime();
+      let activeModeAtMessageTime =
+        roomResource.getActiveLLMModeAtTimestamp(messageTimestamp);
+      console.log('activeModeAtMessageTime', activeModeAtMessageTime);
+      // Only auto-apply if in 'act' mode
+      if (activeModeAtMessageTime !== 'act') {
+        continue;
+      }
+
+      // Auto-apply all ready code patches from this message
```
lukemelia, line 258: This must be redundant code for what is run when the user clicks Accept All. Can we consolidate?

packages/host/app/resources/room.ts
FadhlanR: This is for waiting the room state event of LLM_MODE to be received by room resource.

packages/host/app/services/command-service.ts
FadhlanR: This async delay is inside a loop that stops only when the room resource processes the messages.

PR #2993, opened 2025-07-22
Remove redundant throw for allowed cyclic relationship

burieberry:
The "cycle in ancestor" error thrown in below example seems redundant, and causes various parts of the code-submode to break. If there is a legit "max stack size exceeded" error, it is caught elsewhere and the error reporting appears on the UI, the error-causing code is not saved to card.

Before: Follow the `employmentInfo` field trail in Employee card def

bug

After:
fixed

## Bonus bug fix: 
As I was working on the above, came across a bug in schema field-edit-modal, where the card-type was not calculated for local (non-exported) types.

Before: No field type

local-field-edit-bug
After:
field-edit-fixed
Code:
local-class-code

### DIFF
--- a/packages/host/app/resources/card-type.ts
+++ b/packages/host/app/resources/card-type.ts
@@ -135,16 +135,7 @@ export class CardType extends Resource<Args> {
     if (superCard && card !== superCard) {
       superType = await this.toType(superCard, [card, ...stack]);
     }
-    if (isCodeRefType(superType)) {
-      throw new Error(
-        `bug: encountered cycle in card ancestor: ${[
-          superType,
-          ...stack.map((c) => identifyCard(c)),
-        ]
-          .map((r) => JSON.stringify(r))
-          .join()}`,
-      );
-    }
+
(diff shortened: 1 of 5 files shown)


lukemelia (approved):

lukemelia:
@habdelra should take a look as well

packages/host/app/resources/card-type.ts
```diff
@@ -135,16 +135,7 @@ export class CardType extends Resource<Args> {
...
-    if (isCodeRefType(superType)) {
-      throw new Error(
-        `bug: encountered cycle in card ancestor: ${[
-          superType,
-          ...stack.map((c) => identifyCard(c)),
-        ]
-          .map((r) => JSON.stringify(r))
-          .join()}`,
-      );
-    }
```
habdelra, line 147: i agree. this was really really old code, back when it was possible to do this kind of thing. i agree with @burieberry here--i think it's actually impossible to do this now. there are many layers that will prevent this from happening before it reaches this point.

habdelra (approved):

PR #2999, opened 2025-07-23
Centralize custom serializers/deserializers

habdelra:
This PR centralizes all the custom serializers/deserializers into runtime common. A followup PR will eliminate custom code for serializers/serializers in our card API


lukemelia (approved):

habdelra:
it just occurred to me that we'll also need to centralize the BaseDef and CardDef serializers too. I can do that in a separate PR...

Or well, maybe the deeper question is: why do we need to fully deserialize a card on the server (vs just a serializing/deserializing a specific field). it really comes down to normalizing the on disk format from teh wire format of the card. but maybe we can do that without having to actually run the card....

habdelra:
> it just occurred to me that we'll also need to centralize the BaseDef and CardDef serializers too. I can do that in a separate PR...
> 
> Or well, maybe the deeper question is: why do we need to fully deserialize a card on the server (vs just a serializing/deserializing a specific field). it really comes down to normalizing the on disk format from teh wire format of the card. but maybe we can do that without having to actually run the card....

and i think the indexed serialization can help with this....

PR #3003, opened 2025-07-23
Eliminate top level `dialog` element

lukemelia:
- Fixes issue where LastPass popovers were not clickable


lukemelia:
Percy diff looks to be an improvement to me, though I don't understand why they had the extra border/padding before

backspace (approved):
weird about the former padding in Percy, but it’s nice to no longer have that incorrect

PR #3007, opened 2025-07-24
AI panel: offer to copy submitted content

jurgenwerk:
This will copy the original file content that was submitted in a form of an attached file:

image

<br><br>
In this case, it will this content to the clipboard:

```
import { concat } from '@ember/helper';
import { formatDateTime } from '@cardstack/boxel-ui/helpers';
// ═══ [EDIT TRACKING: ON] Mark all changes with ⁿ ═══
import { CardDef, field, contains, Component } from 'https://cardstack.com/base/card-api'; // ¹ Core imports
import StringField from 'https://cardstack.com/base/string';
import DatetimeField from 'https://cardstack.com/base/datetime';
import TextAreaField from 'https://cardstack.com/base/text-area';
import UrlField from 'https://cardstack.com/base/url';
import PartyIcon from '@cardstack/boxel-icons/cake'; // ² Icon import

export class BirthdayInvite extends CardDef { // ³ Card definition
  static displayName = 'Birthday Invite';
  static icon = PartyIcon;
  
  @field celebrantName = contains(StringField); // ⁴ Primary fields
  @field age = contains(StringField); // Can be "25th" or "Sweet 16"
  @field partyDate = contains(DatetimeField);
  ...
  ...
  ...
  ...
 ```

### CODE REVIEW

File: packages/host/app/components/ai-assistant/code-block/diff-editor-header.gts [hunk 1]
```diff
@@ -204,6 +248,13 @@ export default class CodeBlockDiffEditorHeader extends Component<CodeBlockDiffEd
     return items;
   }
 
+  copySubmittedContent = async () => {
+    if (this.loadSubmittedContent.isRunning) {
+      await this.loadSubmittedContent.perform(); // Should be dropped if loading is already running
```
jurgenwerk, line 257: Hm now that I am looking at this, I am not sure this await assures the currently running task is finished. Need to check this

habdelra: also, its a little dangerous to have an async method in a component. this should be a task, as `this` may go out of scope if this component is destroyed but the promise hasn't resolved.

lukemelia (approved):

packages/host/app/components/ai-assistant/code-block/diff-editor-header.gts
jurgenwerk: Added a new task for async work.

PR #3008, opened 2025-07-24
Use fieldSerializer string to specify serializer for primitive fields

habdelra:
This PR adds the capability for a primitive FieldDef to provide a serializer's name instead of the actual code when specifying a serializer. We will lookup the named serializer from among our known serializers and use that. In this PR I've updated only the Date field to use this approach. After merging this PR we'll followup with all the other primitives. This capability will ultimately allow us to no longer run cards in the Realm server in subsequent PRs.

### DIFF
--- a/packages/base/date.gts
+++ b/packages/base/date.gts
@@ -4,7 +4,12 @@ import { fn } from '@ember/helper';
 import { BoxelInput } from '@cardstack/boxel-ui/components';
 import { not } from '@cardstack/boxel-ui/helpers';
 import CalendarIcon from '@cardstack/boxel-icons/calendar';
-import { DateSerializer, fieldSerializer } from '@cardstack/runtime-common';
+import {
+  type SerializerName,
+  DateSerializer,
+  fieldSerializer,
+  getSerializer,
+} from '@cardstack/runtime-common';
 
 // The Intl API is supported in all modern browsers. In older ones, we polyfill
 // it in the application route at app startup.
@@ -31,7 +36,7 @@ class View extends Component<typeof DateField> {
 export default class DateField extends FieldDef {
   static icon = CalendarIcon;
   static [primitive]: Date;
-  static [fieldSerializer] = 'date';
+  static [fieldSerializer]: SerializerName = 'date';
   static displayName = 'Date';
   static embedded = View;
   static atom = View;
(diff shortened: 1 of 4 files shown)


packages/base/date.gts
```diff
@@ -38,12 +31,8 @@ class View extends Component<typeof DateField> {
 export default class DateField extends FieldDef {
   static icon = CalendarIcon;
   static [primitive]: Date;
+  static [fieldSerializer] = 'date';
   static displayName = 'Date';
-
-  static [serialize] = DateSerializer.serialize;
-  static [deserialize] = DateSerializer.deserialize;
-  static [queryableValue] = DateSerializer.queryableValue;
-
```
habdelra, line 46: these are no longer necessary for primitive fields

packages/base/date.gts
```diff
@@ -38,12 +36,8 @@ class View extends Component<typeof DateField> {
 export default class DateField extends FieldDef {
   static icon = CalendarIcon;
   static [primitive]: Date;
+  static [fieldSerializer]: SerializerName = 'date';
```
habdelra, line 39: Now we can provide simply the name of the serializer. The idea is that anytime you use a [primitive] symbol, you can pair that with a [fieldSerializer] symbol with the name of the serializer you want to use for the field.

packages/base/date.gts
```diff
@@ -70,7 +64,9 @@ export default class DateField extends FieldDef {
       if (!this.args.model) {
         return;
       }
-      return DateField[serialize](this.args.model);
+      return getSerializer(DateField[fieldSerializer]).serialize(
+        this.args.model,
+      );
```
habdelra, line 68: you may also be doing serialization for the UI too, you can use the same mechanism to get access to your serializer for the UI as well.

packages/base/date.gts
```diff
@@ -38,12 +35,8 @@ class View extends Component<typeof DateField> {
 export default class DateField extends FieldDef {
   static icon = CalendarIcon;
   static [primitive]: Date;
+  static [fieldSerializer] = 'date' as const;
```
habdelra, line 38: Now we can provide simply the name of the serializer. The idea is that anytime you use a `[primitive]` symbol, you can pair that with a `[fieldSerializer]` symbol with the name of the serializer you want to use for the field.

lukemelia (approved):

PR #3015, opened 2025-07-24
Prerender data loading

ef4:
Closes https://linear.app/cardstack/issue/CS-8900/consume-card-source-api-to-load-card-in-render-route and starts on the infrastructure for https://linear.app/cardstack/issue/CS-8805/capture-errordocuments-like-we-do-today-under-headless-browser


habdelra:
Looks like from the tests that we have introduced an infinite recursion that is blowing up the stack

PR #3018, opened 2025-07-25
add skills to rooms

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-8995/add-skills-to-room-command

When the "Test Skills" button is clicked:
- The system checks if there is an existing AI assistant room .
- If no room exists, a new AI assistant room is created.
- The room is then opened (brought into focus for the user).
- The selected skill(s) from the listing are attached to the room skill menu using the command. (Support adding multiple skills)

Demo:

https://github.com/user-attachments/assets/de6c997f-82b4-4dfb-87fe-be4279160401


packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -45,6 +48,7 @@ class EmbeddedTemplate extends Component<typeof Listing> {
   @tracked selectedAccordionItem: string | undefined;
   @tracked writableRealms: { name: string; url: string; iconURL?: string }[] =
     [];
+  @tracked roomId: string | null = null;
```
tintinthong: how about we just use matrix service `getRoomId`. so we know we are getting state from one place. If matrix service has another way to know about the roomId, we don't dig into internals of the window storage. 

You can either add a new command that wraps AddSkillsToRoom or have a field that suggests to look for roomId inside current room

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -88,13 +92,53 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
+  }
+
   get remixDisabled() {
     return (
       (!this.isSkillListing && !this.hasOneOrMoreSpec) ||
       (this.isSkillListing && !this.hasSkills)
     );
   }
 
+  @action testSkills() {
```
tintinthong: lets call it exactly addSkills  or addSkillsToCurrentRoom

packages/catalog-realm/catalog-app/listing/listing.gts
tintinthong: Or use UseAiAssisstantCommand and add the current room feature there

packages/catalog-realm/catalog-app/listing/listing.gts
lucaslyl: Yeah, I’ve done the changes to rename it back to addSkills

packages/host/tests/integration/commands/use-ai-assistant-test.gts
```diff
@@ -375,6 +375,55 @@ module('Integration | commands | ai-assistant', function (hooks) {
...
 
+  test('adds skill cards to room without sending prompt', async function (assert) {
+    let roomId = createAndJoinRoom({
+      sender: '@testuser:localhost',
+      name: 'room-with-skills',
+    });
+
+    let store = getService('store');
+
+    // Load skill cards
```
richardhjtan, line 387: We can clear the comments out since the assertions explain the expectations

richardhjtan (approved):

packages/host/app/commands/ai-assistant.ts
```diff
@@ -66,27 +65,45 @@ export default class UseAiAssistantCommand extends HostBaseCommand<
...
     input: BaseCommandModule.UseAiAssistantInput,
   ): Promise<string> {
+    // If a specific roomId is provided and it's not 'new', use that room
     if (input.roomId && input.roomId !== 'new') {
       return input.roomId;
     }
+
+    // Check if there's a current room open (only when roomId is not 'new')
+    let currentRoomId = this.matrixService.currentRoomId;
+    if (input.roomId !== 'new' && currentRoomId) {
```
lucaslyl, line 102: @tintinthong  Added a guard clause to check if input.roomId !== 'new'; it only proceeds to check if currentRoomId exists.

tintinthong (approved):

PR #3023, opened 2025-07-28
Create a new room with the same skills as the currently open room

FadhlanR:
This PR makes the "Add same skills" option in the new session hover menu work. When this option is checked, the new session will copy the skills from the currently open room.

Additionally, this PR updates how to open the new session settings: users now need to Shift+click to access them. A tooltip has been added to explain this behavior.

https://github.com/user-attachments/assets/5872c6a6-ca05-4102-9a6e-5904fb77a68b


lukemelia:
The modal that appears should have two buttons: "Cancel" and "Start New Session"

packages/base/command.gts
```diff
@@ -151,9 +145,18 @@ export class PatchCodeInput extends CardDef {
...
+
+  @field enabledSkillCards = containsMany(JsonField);
+  @field disabledSkillCards = containsMany(JsonField);
+  @field commandDefinitions = containsMany(JsonField);
+}
+
 export class CreateAIAssistantRoomInput extends CardDef {
   @field name = contains(StringField);
   @field defaultSkills = linksToMany(Skill);
+  @field skillConfig = contains(SkillConfigField);
```
lukemelia: This seems to overlap with `defaultSkills`. Perhaps we should consolidate?

packages/base/command.gts
```diff
@@ -151,9 +145,18 @@ export class PatchCodeInput extends CardDef {
   @field codeBlocks = containsMany(StringField);
 }
 
+export class SkillConfigField extends FieldDef {
+  static displayName = 'Skill Configuration';
+
+  @field enabledSkillCards = containsMany(JsonField);
+  @field disabledSkillCards = containsMany(JsonField);
+  @field commandDefinitions = containsMany(JsonField);
```
lukemelia: Should the command definitions come from the enabled skills and not be stored separately?

packages/base/command.gts
FadhlanR: Should we only copy the enabled skills? With the skillConfigField approach, I copy the enabled skills and disabled skills from the room state of the open room—so not only the enabled skills.

lukemelia: If we had top level enabledSkills and disabledSkills field, I think that could replace the need for defaultSkills

FadhlanR:
@lukemelia I made updates based on your feedbacks. Please check it again.


https://github.com/user-attachments/assets/b21dd22a-5514-4ef2-89a9-e47d174a6e0d

packages/host/app/commands/create-ai-assistant-room.ts
```diff
@@ -42,15 +42,37 @@ export default class CreateAiAssistantRoomCommand extends HostBaseCommand<
...
+    const uniqueCommandDefinitions = allCommandDefinitions.filter(
+      (command, index, array) =>
+        array.findIndex((cmd) => command.functionName === cmd.functionName) ===
+        index,
+    );
+
+    if (uniqueCommandDefinitions.length) {
+      commandFileDefs = await matrixService.uploadCommandDefinitions(
+        uniqueCommandDefinitions,
+      );
```
lukemelia: Doesn't this code exist somewhere already? Can we have it in one place instead of duplicating it here?

packages/host/app/components/ai-assistant/new-session-button.gts
```diff
@@ -40,41 +35,76 @@ export default class NewSessionButton extends Component<Signature> {
...
+    // Check if Shift key is pressed
+    if (event.shiftKey) {
+      event.preventDefault();
+      this.showMenu = !this.showMenu;
+      return;
+    }
+
+    // Normal click behavior
+    const addSameSkills = this.selectedOptions.has('Add Same Skills');
+    this.args.onCreateNewSession(addSameSkills);
```
lukemelia: Normal click should always create the new session as if no special options have been selected. The path with the options should only happen if you shift-click and then click the Create button in the modal.

FadhlanR: What happens if the user clicks the "+" button while the new-session-settings panel is open and some options are selected?

lukemelia: It should close the modal and not take any other action.

FadhlanR: Updated 

https://github.com/user-attachments/assets/ffd83abe-8e92-441f-80b8-4946b0fe5d7c

lukemelia (approved):
👍

PR #3025, opened 2025-07-28
set created as an attribute of file

tintinthong:
birhttime is not a thing on linux systems so we shud not rely on this filesystem api

prod shows that its not very meaningful `resource_created_at=0` for all of the entries

This PR introduces usage of `fs-xattr` library to set a 'user.created' attribute on the file that is newly created so it will be persistent across re-indexes

https://linear.app/cardstack/issue/CS-9160/look-back-at-payload-why-created-is-returning-0

To test this branch 
- you can sort the cards grid by created at 
- create a new instance and the card shud appear at the top
- you can also verify the attribute on the file like `x-attr` using mac


packages/realm-server/package.json
```diff
@@ -74,9 +74,11 @@
     "uuid": "catalog:",
     "wait-for-localhost-cli": "catalog:",
     "yaml": "catalog:",
-    "yargs": "catalog:"
+    "yargs": "catalog:",
+    "fs-xattr": "catalog:"
   },
   "scripts": {
+    "postinstall": "npm rebuild fs-xattr",
```
tintinthong: this script recompiles the C bindings needed for this native addon

tintinthong:
I think maybe a less accidental way to setup for `fs-xattr` binaries is to use `onlyBuiltDependencies` which can only be set at the root package.json here https://github.com/cardstack/boxel/pull/3025/commits/ef1f227caa31588551df1fcc016c12fcacb51366

tintinthong:
Pause on review. I can see in some aspects of test, the files are just `copySync` and they never go thru any adapter when writing. Seems a bit problematic if we rely on test that have a special filesystem state like that introduced in this PR

Also, rsync writes the file presumably some process can set the attribute too

packages/realm-server/scripts/set-created-attributes.ts
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env ts-node
+
+import { readdirSync } from 'fs-extra';
+import { join } from 'path';
+import { getAttributeSync, setAttributeSync } from 'fs-xattr';
+
```
tintinthong, line 6: its much easier to just delegate to this library so the correct system library is used

packages/realm-server/tests/helpers/index.ts
```diff
@@ -269,18 +263,27 @@ export async function createRealm({
...
 }): Promise<{ realm: Realm; adapter: RealmAdapter }> {
   await insertPermissions(dbAdapter, new URL(realmURL), permissions);
 
-  for (let [filename, contents] of Object.entries(fileSystem)) {
-    if (typeof contents === 'string') {
-      writeFileSync(join(dir, filename), contents);
-    } else {
-      writeJSONSync(join(dir, filename), contents);
+  let adapter = new NodeAdapter(dir, enableFileWatcher);
+
```
tintinthong, line 252: we did this so we use `.write` as the main interface so file attributes are set inside of tests -- particular realm-server-tests

packages/realm-server/scripts/setup-files.sh
```diff
@@ -0,0 +1,50 @@
...
+    exit 1
+fi
+
+# Create destination directory if it doesn't exist
+mkdir -p "$DEST"
+
+# Use rsync without extended attributes first
+# -r: recursive
+# --size-only: only copy if sizes differ (for efficiency)
+rsync -r --size-only $DELETE_FLAG "$SOURCE/." "$DEST/"
```
tintinthong, line 44: actually, there should be `-X` here which was the case in my previous implementation. @lukemelia ur concern of the rsync not preserving `user.created` attributes was valid without `-X`.

habdelra (approved):
👍

jurgenwerk (approved):

tintinthong:
Screenshot 2025-08-13 at 16 09 48

Oh goddamn. extended attributes is not available in the flavor of nfs in efs

Only figured it out after trying to run `getfattr` on the container

tintinthong:
im re-opening another PR to address this

PR #3027, opened 2025-07-28
Fix loader dependency state leak

habdelra:
This PR fixes a state leak in our loader consumption graph. specifically we were caching consumption state, but that state is all predicated on what path you took to get to a specified consumed module. you can't necessarily share the same results when you are taking different paths thru the consumption graph. Also, i don't really think the consumption cache was saving much work, since we are already separately caching deps for each module (that the consumption cache was trying to optimize).

### DIFF
--- a/packages/realm-server/tests/cards/f.js
+++ b/packages/realm-server/tests/cards/f.js
@@ -1,0 +1,6 @@
+import { b } from './b';
+import { g } from './g';
+
+export function f() {
+  return b() + g();
+}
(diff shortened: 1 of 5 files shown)


tintinthong:
I wonder if this gives us a speed boost in indexing as `deps` array is reduced, particularly in running `itemsThatReference`  query

habdelra:
> I wonder if this gives us a speed boost in indexing as `deps` array is reduced, particularly in running `itemsThatReference` query

i think you're right, we were over calculating our deps

tintinthong (approved):

lukemelia (approved):

richardhjtan (approved):
Dependencies are now accurate after the instance was incremented—appreciate the fix! @habdelra

PR #3029, opened 2025-07-28
Add CardDef meta to index

habdelra:
This PR adds CardDef meta to our index, such that any CardDef's that are encountered during the indexing process will be summarized in the index.


packages/runtime-common/tests/index-writer-test.ts
```diff
@@ -1252,6 +1254,10 @@ const tests = Object.freeze({
     );
   },
 
+  // TODO can get a card-def entry
+
+  // TODO can get a card-def entry from working index
+
```
habdelra: don't forget

packages/runtime-common/tests/index-writer-test.ts
```diff
@@ -207,6 +207,8 @@ const tests = Object.freeze({
     ]);
   },
 
+  // TODO can perform invalidations for a card-def entry
```
habdelra: don't forget

packages/runtime-common/tests/index-writer-test.ts
```diff
@@ -543,6 +545,7 @@ const tests = Object.freeze({
     delete (copiedIndex as Partial<BoxelIndexTable>).indexed_at;
     delete (copiedModule as Partial<BoxelIndexTable>).indexed_at;
 
+    // TODO need to add the card-def entry...
```
habdelra: don't forget

lukemelia (approved):
Looks good. A few thoughts:

1) is "card-def" the right name considering some entries are fields? I think it's OK, just wondered as I reviewed it.
2) Would it make sense to capture the card type icon here, too? I think we render it with each instance currently.

habdelra:
> Looks good. A few thoughts:
> 
> 1. is "card-def" the right name considering some entries are fields? I think it's OK, just wondered as I reviewed it.
> 2. Would it make sense to capture the card type icon here, too? I think we render it with each instance currently.

1. no entries are fields actually [[1]](https://github.com/cardstack/boxel/pull/3029/files#diff-144210b0ce628518954b980c432570c07d3b732efd29c0d101153d8e312de58bR492). hence `card-def`. all pertinent field info (recursively) can be determined from each entry. from the server we never, out of the blue, try to load a field def. we are always loading a field def because we are navigating the fields of a card def. If we come across a scenario that requires us to get field meta outside the context of a card def, then let's add field-def entries. but let's not do that until we have a need for it.
2. i thought about that, but the card type icon requires rendering. and I opted to not make more changes than necessary. I don't want to fall into the trap of premature optimization. It will also make merging with ed's work easier to hold off on that. let's tackle that in the future if we feel that will make the system more optimal.

PR #3030, opened 2025-07-28
Remove caching for computed fields

lukemelia:
This PR transitions computed field execution to an on-demand model, eliminating caching of computed values and stale value tracking. There was a bunch of positive fallout from that, detailed below:
### Computed Fields Behavior:
- **On-demand Execution:** Computed fields are now executed on-demand when accessed and are no longer cached or stored in the data bucket. This ensures fresh values on every access. (`docs/card-serialization-deserialization.md` [[1]](diffhunk://#diff-8d8cd1eb40a449da525000e31dabdb83f3c92ced3ecc8e637c080dbe75f844efL172-R172) `docs/computed-fields.md` [[2]](diffhunk://#diff-d5208e7c25550e4411a642dc7b518f9426a53602e2c1cff4d79f1a6369e60f94L3-R9)

### Codebase Simplification:
- **Removal of Stale Value Mechanism:** The `StaleValue` type and related methods (e.g., `markAllComputedsStale`) have been removed, simplifying the logic for field updates. (`packages/base/card-api.gts` [[1]](diffhunk://#diff-abdbf2e4d24040dd1eea3fc9a720d36711a16847f57b97f8bb7522faccbe11e9L203-L220) [[2]](diffhunk://#diff-abdbf2e4d24040dd1eea3fc9a720d36711a16847f57b97f8bb7522faccbe11e9L3153-L3165)
- **Recompute Functionality Replaced:** Most calls to the old `recompute` function calls have been replaced with `cardTracking.set(instance, true)` for notifying Glimmer when cards are ready to re-render. (`packages/base/card-api.gts` [[1]](diffhunk://#diff-abdbf2e4d24040dd1eea3fc9a720d36711a16847f57b97f8bb7522faccbe11e9L700-R686) [[2]](diffhunk://#diff-abdbf2e4d24040dd1eea3fc9a720d36711a16847f57b97f8bb7522faccbe11e9L764-R750) [[3]](diffhunk://#diff-abdbf2e4d24040dd1eea3fc9a720d36711a16847f57b97f8bb7522faccbe11e9L798-R784)
- **Renamed recompute to ensureLinksLoaded:** The card-api's `recompute` method only remaining responsibility is to ensure that Links/LinksToMany fields (direct and indirect) are loaded, so we renamed it to `ensureLinksLoaded`.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -3174,7 +3174,7 @@ async function _updateFromSerialized<T extends BaseDefConstructor>({
     if (isCardInstance(instance) && resource.id != null) {
       (instance as any)[meta] = resource.meta;
     }
-
+    cardTracking.set(instance, true);
     if (isCardInstance(instance) && resource.id != null) {
       // importantly, we place this synchronously after the assignment of the model's
       // fields, such that subsequent assignment of the id field when the model is
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -384,8 +384,7 @@ export interface Field<
   handleNotLoadedError(
     instance: BaseInstanceType<CardT>,
     e: NotLoaded,
-    opts?: RecomputeOptions,
-  ): Promise<
+      ): Promise<
     BaseInstanceType<CardT> | BaseInstanceType<CardT>[] | undefined | void
   >;
 }
@@ -1279,8 +1278,7 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
   async handleNotLoadedError(
     instance: BaseInstanceType<CardT>,
     e: NotLoaded,
-    opts?: RecomputeOptions,
-  ): Promise<BaseInstanceType<CardT> | undefined> {
+      ): Promise<BaseInstanceType<CardT> | undefined> {
     let deserialized = getDataBucket(instance as BaseDef);
     let identityContext =
       identityContexts.get(instance as BaseDef) ?? new SimpleIdentityContext();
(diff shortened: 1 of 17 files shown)


packages/base/card-api.gts
```diff
@@ -3121,14 +3191,11 @@ async function _updateFromSerialized<T extends BaseDefConstructor>({
...
-    // we always want this to run after we set values as we're not sure which
-    // values we set that depend on computeds. Currently we do the thing that
-    // is always correct.
-    markAllComputedsStale(instance);
+    // Note: Computed fields are now always executed on-demand and never cached,
+    // so server-provided computed values would be ignored. We run recompute here
+    // primarily to populate non-computed fields and trigger any necessary loading
+    // of related cards. For updateFromSerialized() we always want this to run
+    // after we set values as we're not sure which values we set that depend on computeds.
     await recompute(instance, {
```
habdelra: we shouldn't need to call this anymore. it was put there to deal with literally the bug you are solving specifically right after deserialization.

habdelra (commented):
I think your refactor probably doesn't go far enough. My hope would be that we can remove the recompute() function and all the private/module scoped functions that it consumes. Recompute does 2 things: evaluates computeds to store  in the deserialized cache and to optionally load links.

Definitely the first case should be eliminated with your refactor. in terms of loading links, i don't think we want to use recompute to do this any more. link loading during indexing uses a completely different process involving rendering. and on the client, there shouldnt really be a need to load links since the store is very much over fetching links. so my hunch is that we could probably ditch the recompute(). although doing that as a separate PR seems totally reasonable.

habdelra:
regardless, i might have missed it, but you should remove the call to recompute in the setter property descriptor too. That is where we populate the cache for computeds, so if we are not reading from the cache anymore, there is no need to populate it.

habdelra:
Probably a good idea would be to locate all the places we call recompute() and figure out if the need to call it has been obviated now.

habdelra (approved):

PR #3033, opened 2025-07-29
Implement copy file history choice

FadhlanR:
https://github.com/user-attachments/assets/22c978b4-deb0-4627-a3ab-f71a2c9eb37c


packages/host/app/services/ai-assistant-panel-service.ts
```diff
@@ -226,6 +289,22 @@ export default class AiAssistantPanelService extends Service {
...
+            this.collectFileHistoryFromCurrentRoom();
+
+          if (attachedFiles.length > 0 || attachedCards.length > 0) {
+            await this.matrixService.sendMessage(
+              roomId,
+              'This session includes files and cards from the previous conversation for context.',
+              attachedCards,
+              attachedFiles,
+            );
+          }
```
lukemelia, line 305: Is there anything to stop this message from generating a response from the aibot?

lukemelia (approved):

packages/host/app/services/ai-assistant-panel-service.ts
FadhlanR: No for now, if we need we can introduce a new flag that can be used for ai bot to ignore a message.

PR #3034, opened 2025-07-29
CS-9178 making create listing command to be more scripted oriented

richardhjtan:
Refer to https://linear.app/cardstack/issue/CS-9178/create-listing-command-to-be-scripted-until-the-need-of-ai-bot

### What is changing
- get rid the need of catalog skill in order to trigger create listing command
- only ask AI operation when needed, eg: update card details


richardhjtan:
@tintinthong  I will merge this first as the flow was presented to you physically and this PR includes to fix the host test

PR #3036, opened 2025-07-29
Cs 9137 fix issue where deleted file still appears from search bcos it was not tombstoned

tintinthong:


packages/host/app/lib/current-run.ts
```diff
@@ -349,6 +349,16 @@ export class CurrentRun {
       return invalidationList;
     }
 
+    // Check for deleted files - files that exist in index but not on filesystem
```
tintinthong, line 352: I made a concious decision to add deletedUrls to the invalidation list. Although I was told to tombstone everything inside of the db. I am still but indirectly.

Reason being is batchUpdates fundamentally apply to invalidation array. That way I can be confident that I only tombstone the working index and the batch update gets promotes to boxel_index for me, otherwise more machinery needed. Also confine tombstoning to the `invalidate` function

habdelra: i think that's perfectly fine

packages/runtime-common/index-writer.ts
```diff
@@ -512,31 +512,7 @@ export class Batch {
...
-        ]),
-      ];
-      invalidations = [...new Set([...invalidations, ...workingInvalidations])];
-    }
-
-    if (invalidations.length === 0) {
-      return [];
-    }
-
+  async tombstoneEntries(invalidations: string[]) {
```
habdelra: I would like for you to hoist this into the constructor. we lose cohesion when we start doing side effects like this. this method is really only for doing the thing that it is named after.  it means you need to add a little extra logic to set the `this.ready` (like `this.ready = this.setNextRealmVersion().then(() => this.tombstoneEntries()`), but it will make for less surprising code.

tintinthong: I see your point about cohesion. 

How do I know what to tombstone before calling `.invalidate`?  Previously, this side effect of tombstoning was called inside of `invalidate` -- I guess for good reason. 

By putting tombstoning into the constructor, the question then becomes "am I just tombstoning everything at every batch run?" 

If we want this side-effect to be manually run by the current run operation. I think it makes sense to manually call `tombstoneEntries` from outside -- typically after every call to `invalidate` in the current-run file.

habdelra: So the point of the invalidation is to tombstone everything. And then use the visit process in the index to resurrect the entries from death. Ideally the entire realm should be tombstoned and then we visit all items in the entire realm and resurrect the ones we are still using. I believe we made optimization to speed up the from scratch indexing that would make that process impossible now….. 

Given that I think it’s ok to have current run manually tombstone the things that it sees as having been deleted for from-scratch from the outside.

habdelra: But only for from-scratch. The normal invalidate for incremental should still tombstone from the inside as part of calling invalidate()

habdelra: (I’m away from my keyboard right now and don’t remember if we call Batch.invalidate() directly for from scratch indexing). I’m assuming we don’t which is why it’s a side effect of incrementing the realm version in your PR.  It should VERY much be part of the Batch.invalidate() method. I’m not suggesting removing it there. Tombstoning is part of the invalidation process, and it doing it in that method does not lead to lack of cohesion, and it’s not surprising that it happens there.

habdelra: Another thought: why not call Batch.invalidate() multiple times during from scratch? And just call Batch.invalidate() for the files that we see as missing from the filesystem?

habdelra: Sorry for all the replies to this, I think I finally understand why you did this: it’s because I said we should invalidate everything in from scratch. In hindsight that wrong because we added the mtime checks to see what has actually changed. If this tombstoning as part of the increasing the realm version is a reaction to that, then I think we can just remove it. The fact that we added the files that were in the index but not on disk to the invalidation list should be all that is required I believe, as that is technically something that has “changed”.

tintinthong: > Sorry for all the replies to this, I think I finally understand why you did this: it’s because I said we should invalidate everything in from scratch. In hindsight that wrong because we added the mtime checks to see what has actually changed. If this tombstoning as part of the increasing the realm version is a reaction to that, then I think we can just remove it. The fact that we added the files that were in the index but not on disk to the invalidation list should be all that is required I believe, as that is technically something that has “changed”.

I believe that is all I have done. I think what is confusing is that I exposed a new tombstoneEntries method which actually should be private to the batch (now its public)

habdelra: my bad-- i must have imagined it, i thought that the increment realm version method was calling this. i see now that it is not. sorry about that. let's just make the tombstoning a private method then. thanks!

habdelra (approved):

habdelra:
Also, i think this branch will have conflicts with https://github.com/cardstack/boxel/pull/3029. let's merge yours first, and I'll deal with the conflicts on my end.

PR #3039, opened 2025-07-30
Simplify UI remix handling

tintinthong:


packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -96,10 +96,10 @@ class EmbeddedTemplate extends Component<typeof Listing> {
     if (!commandContext) {
       throw new Error('Missing commandContext');
     }
```
lucaslyl, line 98: https://github.com/user-attachments/assets/c8ee4ca5-e95c-4c57-a763-d243bacc9466

I can see the changes make the remix process extremely fast now, but I noticed that incorrect parameter passing is causing the remix listing isolated template to not work.

```_remix = task(async (realm: string)```
should be changed to
```_remix = task(async (realmUrl: string)```

Comment (tintinthong) [reply packages/catalog-realm/catalog-app/listing/listing.gts:98]: good catch! thx

Comment (tintinthong) [reply packages/catalog-realm/catalog-app/listing/listing.gts:98]: haiz really gotta fix this ci LINT for catalog

File: packages/host/app/commands/listing-remix.ts [hunk 2]
```diff
@@ -108,5 +110,20 @@ export default class RemixCommand extends HostBaseCommand<
         });
       }
     }
+
+    let prompt =
+      'Remix done! Give me example prompts on how to edit this card.';
```
richardhjtan:
```suggestion
      'Remix done! Please suggest two example prompts on how to remix this card.';
```

richardhjtan: Initially, we ask for two example prompts for remix, however it's fine to change the number or no limit and let AI decide

packages/host/app/commands/listing-remix.ts
```diff
@@ -108,5 +110,20 @@ export default class RemixCommand extends HostBaseCommand<
...
+    const skillCardIds = [
+      skillCardURL('boxel-environment'),
+      skillCardURL('boxel-development'),
+      skillCardURL('source-code-editing'),
+      skillCardURL('catalog-listing'),
+    ];
+    await new UseAiAssistantCommand(this.commandContext).execute({
+      roomId: 'new',
+      prompt,
+      openRoom: true,
```
richardhjtan, line 126:
```suggestion
      openRoom: true,
      attachedCards: [listing as CardAPI.CardDef],
```

richardhjtan: Giving the listing card context helps to generate more related prompts relates to the module

PR (without)
Screenshot 2025-07-31 at 10 22 01 PM

With the attached Listing card
Screenshot 2025-07-31 at 10 22 51 PM

richardhjtan (changes requested):

tintinthong:
im honestly still half-hearted about this PR. It simplifies things and makes things more straightforward but, occasionally it can just look slow if room opening is that slow;

richardhjtan (approved):

PR #3041, opened 2025-07-30
Ensure that user does not need to re-enter payment card info

jurgenwerk:
Previously, our setup was the following: 

1. Configure payment links in Stripe (for buying credits, subscribing, customer portal)
2. Realm server responds with payment links
3. Host app renders them and users can click on them (it takes them to Stripe)

The issue with this is that the payment methods by users were not memorized, and users needed to re-enter whenever buying something again, which was annoying and added friction. The solution is to use "Stripe checkout sessions", where we can **explicitly specify that we want to persist previously entered payment cards** (which is not possible using Stripe payment links).

This is how the new flow looks like:

image


richardhjtan (approved):
Have to run this with the latest upstream locally. Tested and it works! Great work 👍 

Screenshot 2025-08-01 at 12 23 44 PM

habdelra (approved):

packages/host/app/services/billing-service.ts
```diff
@@ -54,141 +39,65 @@ export default class BillingService extends Service {
     this.reset.register(this);
   }
 
+  extraTokensPricing: Record<number, number> = {
+    2500: 5,
+    20000: 30,
+    80000: 100,
+  }; // in USD
```
lukemelia: Should this come from a source of truth (i.e. stripe?)

packages/realm-server/handlers/handle-create-stripe-session.ts
```diff
@@ -0,0 +1,254 @@
...
+          'Either aiTokenAmount or plan parameter is required',
+        );
+        return;
+      }
+
+      let tokensToPriceMap: Record<number, number> = {
+        2500: 5,
+        20000: 30,
+        80000: 100,
+      };
```
lukemelia: Is this the source of truth for these prices?

lukemelia (approved):

packages/host/app/services/billing-service.ts
jurgenwerk: The source of truth is actually this exact object–we're not pulling prices from Stripe anymore

packages/realm-server/handlers/handle-create-stripe-session.ts
jurgenwerk: The source of truth is actually this exact object–we're not pulling prices from Stripe anymore

packages/host/app/services/billing-service.ts
jurgenwerk: I see the issue, this price config is defined in two places in our code and one of them is redundant. Both are the sources of the same truth. Will fix this

jurgenwerk: The source of truth is in our code but I see the issue, we're defining the same value in 2 places. Here's the fix: https://github.com/cardstack/boxel/pull/3041/commits/6e6b9aac1eeadfce4dc699fa3bcbf73d4d62914f

PR #3042, opened 2025-07-30
Create Solitaire listing

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-272/create-solitaire-listing

solitaire-listing-01

### DIFF
--- a/packages/catalog-realm/solitaire/solitaire.gts
+++ b/packages/catalog-realm/solitaire/solitaire.gts
@@ -1,4 +1,10 @@
-import { CardDef, Component } from 'https://cardstack.com/base/card-api';
+import {
+  CardDef,
+  Component,
+  field,
+  contains,
+} from 'https://cardstack.com/base/card-api';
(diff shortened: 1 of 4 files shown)


packages/catalog-realm/solitaire/solitaire.gts
```diff
@@ -0,0 +1,847 @@
...
+  </template>
+
+  // Define helper functions
+  get multiply() {
+    return (a: number, b: number) => a * b;
+  }
+
+  get subtract() {
+    return (a: number, b: number) => a - b;
+  }
```
richardhjtan: @lucaslyl  Is this needed as we have helpers about multiply/subtract?

richardhjtan (approved):

packages/catalog-realm/solitaire/solitaire.gts
lucaslyl: Good Catch! Done remove the custom getter

PR #3044, opened 2025-07-30
Create Astral Garden Listing

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-271/create-astral-garden-listing

astral-garden-listing-01


richardhjtan (commented):
Screenshot 2025-07-30 at 4 41 54 PM

Found there are two scrollbars when seeing the example, can we fix this because the inner scrollbar cannot scroll to the end

lucaslyl:
> Screenshot 2025-07-30 at 4 41 54 PM
> Found there are two scrollbars when seeing the example, can we fix this because the inner scrollbar cannot scroll to the end

Sure, Done remove the overflow auto behaviour

tintinthong (approved):

PR #3045, opened 2025-07-30
Serve card def meta from Realm

habdelra:
This PR exposes the card def meta as a Realm endpoint `/_card-def?coderef=XXXX`


packages/runtime-common/realm.ts
```diff
@@ -2298,6 +2300,75 @@ export class Realm {
...
 
+  private async getCardDef(
+    request: Request,
+    requestContext: RequestContext,
+  ): Promise<Response> {
+    let href = new URL(request.url).search.slice(1);
+    let payload = parseQuery(href);
+    if (!payload.codeRef) {
+      return badRequest({
+        message: `The request body is missing the coderef parameter`,
```
backspace:
```suggestion
        message: `The request body is missing the codeRef parameter`,
```

The message should match the expected name, I think, same for below

backspace (approved):

lukemelia (approved):

PR #3048, opened 2025-07-31
Update host mode detection

backspace:
This is based on the key insight that `host` shouldn’t need to be aware of the domain/subdirectory structure of published realms. If I have a card at `buck.staging.boxel.dev/may8/Product/13e1dc95-e84d-4ddd-ad17-fa6b3c59cca5`, it should just be able to request it, as the realm server will exist at that domain. Same for `buck.boxel.site` when we support custom subdomains, and `whateverdomain.net` when we support custom domains.

How then to know when a request is for host mode or operator mode? `hostsOwnAssets: false` means `host` is being served via the realm server vs directly from Cloudfront-S3 (preview deployments). The realm server also configures `host` with its `realmServerDomain`, so if `window.location.hostname` is `realms-staging.stack.cards`, we know it’s not host mode.

There’s also a special query parameter to specify the origin that can be used on preview deployments, `host-mode-origin=`. With this branch deployed to staging, it makes this URL work `https://custom-subdomain-host-mode-cs-9205.boxel-host-preview.stack.cards/may8/Product/13e1dc95-e84d-4ddd-ad17-fa6b3c59cca5?host-mode-origin=https://buck.staging.boxel.dev`:

Custom Whiskey Glass with premium quality side and bottom engraving 2025-08-07 12-12-00

And with this branch deployed to staging, this URL works  `https://buck.staging.boxel.dev/may8/Product/13e1dc95-e84d-4ddd-ad17-fa6b3c59cca5`:

Custom Whiskey Glass with premium quality side and bottom engraving 2025-08-07 12-12-35

This was facilitated by me manually adding a row in staging `published_realms` and copying my `may8` realm to the special `_published` realms directory.


packages/host/app/routes/card.gts
```diff
@@ -36,21 +34,8 @@ export default class Card extends Route<ReturnType<StoreService['get']>> {
...
-    let realmUrlString = `${config.realmServerRoot}${this.hostModeService.userSubdomain}/${realm}/`;
-
-    await this.realm.ensureRealmMeta(realmUrlString);
-
-    let realmUrl = this.realm.url(realmUrlString);
-
-    if (!realmUrl) {
-      throw new Error(`Realm not found: ${realmUrlString}`);
-    }
+    let cardUrl = `${this.hostModeService.hostModeOrigin}/${params.path}`;
```
backspace, line 37: Fully delegating resolution to DNS, the realm server, and realm makes this much simpler.

packages/host/app/services/host-mode-service.ts
```diff
@@ -8,38 +8,55 @@ export default class HostModeService extends Service {
...
 
-    return window.location.hostname.split('.')[0];
+    return window.location.origin;
+  }
+
+  get originIsNotMatrixTests() {
+    return (
+      this.hostModeOrigin !== 'http://localhost:4202' &&
+      this.hostModeOrigin !== 'http://localhost:4205'
+    );
```
backspace, line 59: I don’t love these exceptions but it’s what restored the Matrix tests. Once this is more settled, it might make sense to have another config property passed from the realm server to host that indicates host mode shouldn’t be detected in these cases. Or I’m open to suggestions!

lukemelia (approved):

PR #3049, opened 2025-07-31
CS-9155 fix invalid skills for llm chooser

richardhjtan:
- Fix invalid skill linkage causing card error
- Add LLM & environment chooser module into the experiment realm


lucaslyl (approved):
LGTM!

PR #3052, opened 2025-07-31
Persist ai panel width

FadhlanR:
This PR persists the AI panel's width using local storage, so the width remains unchanged after a host reload.


packages/host/app/components/operator-mode/submode-layout.gts
```diff
@@ -112,14 +114,30 @@ export default class SubmodeLayout extends Component<Signature> {
   private suppressSearchClose = false;
   private declare doSearch: (term: string) => void;
 
+  @action
+  private onLayoutChange(layout: number[]) {
+    if (layout.length === 2) {
```
jurgenwerk, line 122: Perhaps it would be good to add a comment what values come in `layout` here and why we want 2 but only take the 2nd one. I'm not too familiar with `onLayoutChange` this but perhaps this means previous and current value?

jurgenwerk (approved):

PR #3058, opened 2025-07-31
add get-all-realm-metas

tintinthong:
This PR cards (inside of code mode) to query realm infos and display a dropdown of realms to write to

Screenshot 2025-08-11 at 21 43 00


packages/host/app/resources/command-data.ts
```diff
@@ -0,0 +1,95 @@
...
+ * Data loading abstraction for commands - provides reactive state and automatic context handling.
+ *
+ * @example
+ * ```typescript
+ * const commandResource = commandData(this, GetAllRealmMetasCommand, undefined);
+ * if (commandResource.current?.isSuccess) {
+ *   console.log(commandResource.current.result);
+ * }
+ * ```
+ */
```
tintinthong: I added this data loading resource from commands. The point of it is to ready load data coming from commands in a component. It actually follows the patterns from an ember resource util called`remoteData`

packages/host/app/resources/command-data.ts
```diff
@@ -0,0 +1,95 @@
...
+  parent: object,
+  commandClass: new (
+    context: CommandContext,
+  ) => Command<CardInputType, CardResultType>,
+  executeArgs: CardInputType extends CardDefConstructor
+    ? Partial<FieldsOf<CardInstance<CardInputType>>>
+    : undefined,
+) {
+  return maybe(parent, (_) => {
+    const commandContext = inspectContext(parent, CommandContextName);
```
tintinthong: inspectContext is a programmatic way to get context variables at class property level

packages/host/app/resources/command-data.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+export function commandData<
+  CardInputType extends CardDefConstructor | undefined,
+  CardResultType extends CardDefConstructor,
+>(
+  parent: object,
+  commandClass: new (
+    context: CommandContext,
+  ) => Command<CardInputType, CardResultType>,
+  executeArgs: CardInputType extends CardDefConstructor
```
tintinthong: I havent checked reactivity with executeArgs here but should do in future command

packages/host/app/commands/get-all-realm-metas.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import type Realm from '../services/realm';
+
+export default class GetAllRealmMetasCommand extends HostBaseCommand<
+  undefined,
+  typeof BaseCommandModule.GetAllRealmMetasResult
+> {
+  @service declare private realm: Realm;
+
+  static actionVerb = 'Get All Realms Info';
```
lukemelia: This should be shorter: "Fetch Realms"

packages/host/app/resources/command-data.ts
```diff
@@ -0,0 +1,95 @@
...
+  setError(error: Error) {
+    this.status = 'error';
+    this.value = null;
+    this.error = error;
+  }
+}
+
+/**
+ * Data loading abstraction for commands - provides reactive state and automatic context handling.
+ *
```
lukemelia: What options did you consider for making this functionality available besides dynamic context? Could it live in runtime common? base realm?

tintinthong: I think base realm is a possiblity. This can be a good place. I think `maybe` is a good abstraction to carry to base too. This is cleaner! 

Ive always wondered whether there is a reason ember-resources was never put int runtime-common as there are duck types all over the place.

tintinthong: interesting. I when putting the dependency in base and the prerendered html that is indexed was BEFORE the resource data got loaded.

tintinthong: Oh my bad this is the behaviour even BEFORE the change. Our indexer doesn't wait for async data loading so fitted formats dont display UI

packages/experiments-realm/simple-search-card.gts
```diff
@@ -0,0 +1,135 @@
...
+      typeof SearchCardsByTypeAndTitleInput,
+      typeof SearchCardsResult
+    >(this, SearchCardsByTypeAndTitleCommand, () => {
+      if (!this.args.model.titleSearch) {
+        return undefined; // not executing command
+      }
+      return new SearchCardsByTypeAndTitleInput({
+        title: this.args.model.titleSearch,
+      });
+    });
```
tintinthong: I have included the api usage example here with reactive arguments. 

So here we expect, 
- command to run when our input changes 
- resource with execution state if no context is provided

so card users will have something stable to work with

packages/host/app/resources/command-data.ts
tintinthong: @lukemelia 

I did the following 
- the resource to base
- I added some treatment for loading of the menu
- fixed a bug to fallback focus-trap when data was not available <- in main
- added a card in experiments that will show how to use this resource

packages/experiments-realm/simple-search-card.gts
lukemelia: I wonder if we can get TypeScript to infer the generics from the command's type. Maybe a question for @ef4.

packages/experiments-realm/simple-search-card.gts
```diff
@@ -0,0 +1,135 @@
...
+
+    get searchResults() {
+      const resource = this.searchResource?.current;
```
lukemelia: We have a question mark after this.searchResource in a bunch of places here. Is it required? Can searchResource every be undefined?

lukemelia (approved):
Left a couple of comments. I'd like to have @ef4 review this. In particular, I'm unsure about the inspect-context stuff and whether there is a way to do that that involved less private API.

packages/experiments-realm/simple-search-card.gts
tintinthong: No. its not needed

tintinthong:
Main changes here https://github.com/cardstack/boxel/pull/3058/commits/aa1eb342ab03bd4fff4b800e3becab0e1c6388e9
- remove maybe wrapper
- resource returns error state when context doesnt exist
- overloading so typescript executeArgs

@ef4 was concerned bcos the api of using commands to get data which was already synchronous and question whether we want to remove publicAPI actions. But there are pros and cons.

tintinthong:
This PR is ready as is @lukemelia . I can either merge or wait for another review from you

lukemelia (approved):

PR #3059, opened 2025-07-31
Add theming to card-def

burieberry:
Changes:
- Added `cardInfo` compound field with only `theme` field (other fields to be added in separate PR)
- Added `extractCssVariables` function in runtime-common and related unit tests in host. The code here is my revision of some css-parsing code that Chris had sent me.
- `base/field-component` handles adding extracted css-variables from theme card to the card-container
- Hid the `cardInfo` field from default templates until we add the better templates for it as well as more theming support in our ui components
- Updated boxel card-container component to use new css-variables for theming (we are using css-variables from shadcn: https://ui.shadcn.com/docs/theming)
- Added some sample themes to experiments realm from tweakcn (https://tweakcn.com/editor/theme) and from some examples that Chris had sent me. The `Theme/cs.json` file is an unofficial/in-progress draft of our boxel theme as I came up with through testing the css-variables in our components.

Next steps after this PR:
- Adding the remaining `cardInfo` fields to card-info
- Adding the cardInfo templates from designs on zeplin
- Creating more tickets for theming next steps and discussing with the team


packages/base/field-component.gts
```diff
@@ -205,6 +207,13 @@ export function getBoxComponent(
     };
   }
 
+  function getThemeStyles(css?: string) {
+    if (!extractCssVariables) {
+      return;
+    }
+    return sanitizedHtml(extractCssVariables(css));
+  }
```
burieberry, line 215: I had to add this instead of directly using `extractCssVariables` as a helper because otherwise sometimes it is undefined for some reason and causes the card to error, especially when changing the linked theme. This only became a problem when I moved the css-parsing code from this file to runtime-common.

packages/base/default-templates/isolated-and-edit.gts
```diff
@@ -12,22 +12,22 @@ export default class DefaultCardDefTemplate extends GlimmerComponent<{
...
-    let fieldInstance = field?.card;
-    return fieldInstance?.icon;
-  };
+    return Object.fromEntries(fields);
+  }
+
   <template>
     <div class={{cn 'default-card-template' @format}}>
-      {{#each-in @fields as |key Field|}}
+      {{#each-in this.displayFields as |key Field|}}
```
burieberry, line 26: The cardInfo field is hidden currently as explained in description. However, here's the code if you want to see theming in action:
```
{{#if (eq @format 'edit')}}
        <FieldContainer
          @label='Card Info'
          @icon={{getFieldIcon @model 'cardInfo'}}
          data-test-field='cardInfo'
        >
          <@fields.cardInfo />
        </FieldContainer>
      {{/if}}
 ```
 
Currently only `CardContainer` and only its certain css variables are updated, so you'd only see the color and box-shadow changes. I have separately tested the variables by making changes to more components (cards-grid components, other base-realm default components, field-container and its related components, etc) and it works.

File: packages/base/card-api.gts [hunk 3]
```diff
@@ -2300,6 +2302,56 @@ export class MaybeBase64Field extends StringField {
...
+        }
+      </style>
+    </template>
+  };
+}
+
+export class CardInfoField extends FieldDef {
+  static displayName = 'Card Info';
+  @field theme = linksTo(() => Theme);
+}
```
habdelra, line 2353: card-api is pretty big as it is. i'm always looking for opportunities to reduce it in size. can we move these out into their own modules?

packages/base/helpers/sanitized-html.gts
habdelra, line 1: is there  a reason this is in base? if it was in runtime common it would be less code that would need to be dynamically imported when loading cards.

packages/base/card-api.gts
```diff
@@ -2389,6 +2442,12 @@ export class CardDef extends BaseDef {
   }
 }
 
+export class Theme extends CardDef {
+  static displayName = 'Theme';
+  static icon = ThemeIcon;
+  @field cssVariables = contains(CSSField);
+}
+
```
habdelra, line 2450: as with the comment above, this is a good opportunity to move into a separate module to keep card-api from getting too bloated.

habdelra (approved):
just some minor comments, but otherwise looks good

packages/base/card-api.gts
```diff
@@ -2321,6 +2373,7 @@ export class CardDef extends BaseDef {
     cardTracking.set(this, true);
   }
   @field id = contains(ReadOnlyField);
+  @field cardInfo = contains(CardInfoField);
```
habdelra, line 2376: there is some circularity here which might be what drove you to placing CardInfoField in this module. but while the modules would have a circular reference if you pull out CardInfoField into its own module, the exports that are consumed should be able to be resolved properly in spite of circular module references as they are not circular. (I believe we have loader tests that prove this is ok: `'can dynamically load modules with cycles'` in `realm-server/tests/loader-test.ts`). (this also might explain why StringField is in this module as well which i think could also be moved out for the same reason...)

packages/base/card-api.gts
burieberry: the issue is I actually had to move more things in because somehow we can't import them. it's the field-defs and card-defs

packages/base/card-api.gts
burieberry: > there is some circularity here which might be what drove you to placing CardInfoField in this module. but while the modules would have a circular reference if you pull out CardInfoField into its own module, the exports that are consumed should be able to be resolved properly in spite of circular module references as they are not circular. (I believe we have loader tests that prove this is ok: `'can dynamically load modules with cycles'` in `realm-server/tests/loader-test.ts`). (this also might explain why StringField is in this module as well which i think could also be moved out for the same reason...)

Came across this error:
err

packages/base/default-templates/isolated-and-edit.gts
```diff
@@ -12,22 +12,22 @@ export default class DefaultCardDefTemplate extends GlimmerComponent<{
...
         {{#unless (eq key 'id')}}
```
lukemelia: do we still need this since you are excluding `id` in the `displayFields` accessor?

lukemelia (approved):
Let's try to figure out the circular deps issues with @habdelra's help. Otherwise looking good.

packages/base/card-api.gts
habdelra: hmmm, i think there might be a bug in our circular loading. let's commit what you have, and ill file a ticket for this issue for us to dig in deeper.

habdelra: filed as https://linear.app/cardstack/issue/CS-9219/potential-bug-when-circular-loading-with-our-loader

PR #3064, opened 2025-08-02
Add full reindex script and grafana endpoint

habdelra:
This PR adds a full reindex script and grafana endpoint to allow all the realms to be fully reindexed. This will be necessary in order to generate the card-def meta documents in the index in order to allow searching.

you can use the script from packages/realm-server to do a full reindex on your dev env:
```
pnpm full-reindex
```

TODO:
- [x] deploy https://github.com/cardstack/infra/pull/611 to staging
- [x] deploy https://github.com/cardstack/infra/pull/611 to prod

### DIFF
--- a/packages/realm-server/handlers/handle-reindex.ts
+++ b/packages/realm-server/handlers/handle-reindex.ts
@@ -2,12 +2,17 @@ import Koa from 'koa';
 import {
   type FromScratchResult,
   type FromScratchArgs,
-  fetchUserPermissions,
+  type QueuePublisher,
+  type Job,
+  type Expression,
+  type Realm,
+  query as _query,
+  param,
   userInitiatedPriority,
   SupportedMimeType,
   RealmPaths,
+  DBAdapter,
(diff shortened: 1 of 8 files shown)


packages/realm-server/handlers/handle-reindex.ts
```diff
@@ -72,3 +67,35 @@ export default function handleReindex({
...
+}) {
+  let realmUsername = await realm.getRealmOwnerUsername();
+  let args: FromScratchArgs = {
+    realmURL: realm.url,
+    realmUsername,
+  };
+  await query(dbAdapter, [
+    `UPDATE boxel_index SET last_modified = NULL WHERE realm_url =`,
+    param(realm.url),
+  ]);
```
tintinthong, line 90: 👍

tintinthong (approved):

habdelra:
FYI: I deployed this branch to staging so that I could generate card-def meta files in our index so that we can continue to search in staging.

PR #3065, opened 2025-08-02
Fix worker connection leakage inside matrix tests

tintinthong:
When running matrix test locally, we get "too many clients connected" error. 

I managed to get matrix test to run "mostly" green except 2 tests. The other 2 tests might be flaky since they pass when run isolated

Screenshot 2025-08-02 at 16 41 39

There are all these connections 

Screenshot 2025-08-02 at 16 49 15

### DIFF
--- a/packages/realm-server/package.json
+++ b/packages/realm-server/package.json
@@ -112,6 +112,7 @@
     "lint:test-shards": "ts-node --transpileOnly scripts/lint-test-shards.ts",
     "full-reset": "./scripts/full-reset.sh",
     "full-reindex": "./scripts/full-reindex.sh",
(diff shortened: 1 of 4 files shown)


habdelra:
you can always run `./remove-test-dbs.sh` to clean these up. but I appreciate you hunting down the leaks

tintinthong:
> you can always run `./remove-test-dbs.sh` to clean these up. but I appreciate you hunting down the leaks

Yea I noticed that but I think there was some class of errors which relied on me to clear the process that held on to that connection. But in any case, this PR allows me to run matrix client test without is having it fail (in vs code), particularly, when I run > 5 test or at one go.

habdelra (approved):

PR #3066, opened 2025-08-02
Remove loader from Realm file serialization

habdelra:
This PR removes the loader necessary for serializing instances to the file system from the Realm. As part of this it became apparent that we actually do need to store field-defs meta as well as card-def meta in our index (I wasn't considering the needs around polymorphic fields). in a subsequent PR, i'll refactor the index type that we use for these from 'card-def' to 'meta' to be more inclusive of the fact that we are capturing both card-def and field-def meta (https://github.com/cardstack/boxel/pull/3067).

:warning: Note that this PR needs to have `pnpm full-reindex` performed after merging!

Consoder merging this PR and https://github.com/cardstack/boxel/pull/3067 at the same time to reduce the amount of full indexing that everyone has to do.


packages/host/tests/integration/realm-test.ts
```diff
@@ -761,69 +759,6 @@ module('Integration | realm', function (hooks) {
     );
   });
 
-  test('realm returns 400 error for POST requests that set the value of a polymorphic field to an incompatible type', async function (assert) {
```
habdelra, line 791: now that we are no longer running cards on the server as part of serializing instances to the filesystem, it's really hard to perform this validation as you don't know ahead of time what all field types are compatible with a given field without running the card. I'm skipping this for now, but probably we can come up with a non-loader based solution if we think harder about this....

packages/host/tests/integration/realm-test.ts
```diff
@@ -2411,124 +2344,6 @@ module('Integration | realm', function (hooks) {
     );
   });
 
-  test('realm returns 400 error for PATCH requests that set the value of a polymorphic field to an incompatible type', async function (assert) {
```
habdelra, line 2490: ditto for this as well

packages/realm-server/tests/atomic-endpoints-test.ts
```diff
@@ -632,92 +630,6 @@ module(basename(__filename), function () {
             'the card source is correct',
           );
         });
-        test('writes will get file directly from disk enabling writes that rely on file system but not index', async function (assert) {
```
habdelra, line 635: this test now becomes too low level in its approach. the issue is that now we need to have meta for a card-def in order to even serialize an instance that uses it to disk. it is not possible to write an instance to disk if you dont have meta in the index for the necessary card def. which is why we are now batching out the `Realm.writeMany()` to assert that you always have index meta available to write instances to disk.

habdelra: moreover we no longer load code in the realm server--so we don't even need the realm server form of this (probably worker only).

packages/runtime-common/realm.ts
```diff
@@ -318,7 +320,6 @@ export class Realm {
           owner = await this.getRealmOwnerUserId();
         }
         req.headers.set('X-Boxel-Assume-User', owner);
-        req.headers.set('X-Boxel-Disable-Module-Cache', 'true');
```
habdelra, line 321: we no longer run code in the realm so its not necessary to set this header anymore

packages/runtime-common/realm.ts
```diff
@@ -5,7 +5,6 @@ import {
   type SingleCardDocument,
 } from './document-types';
 import { type CardResource } from './resource-types';
-import { Loader } from './loader';
```
habdelra, line 8: No more loader in the realm 🎉

packages/matrix/tests/card-chooser.spec.ts
habdelra, line 1: The tests in this spec are obsolete. we no longer add realm read permission when linking to a card in a different realm that you own. Rather the realm is indexed under the authority of the realm owner so the permissions never have to be modified when making links among your realms. Moreover, the ability to read between your own realms during indexing (e.g. create an instance from a spec in a different realm that you own) is already tested

packages/host/tests/integration/realm-test.ts
lukemelia: If our meta indexing captured inheritance hierarchy, I think we could verify that a specified field type is or inherits from the type specified in the field. Perhaps the exception would be if we have never seen the specified field type before?

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1923,12 +1926,17 @@ module(basename(__filename), function () {
         // we splat because despite having the same shape, the constructors are different
         { ...realm.realmIndexUpdater.stats },
         {
+          // this is a little misleading because now that we are batching out the
+          // modules and instances to ensure that the meta is generated before
+          // we trying file serialization, this will only report the last batch
+          // of indexing when in fact there where actually 2 batches of indexing
+          // generated by this writeMany()
```
lukemelia, line 2227: I think it is OK to think of it as each batch having up to 2 parts.

packages/runtime-common/realm.ts
```diff
@@ -531,7 +526,38 @@ export class Realm {
...
+          ? 'instance'
+          : undefined;
+      if (lastWriteType === 'module' && currentWriteType === 'instance') {
+        // we need to generate/update possible card-def meta in order for
+        // instance file serialization that may depend on the included module to
+        // work. TODO: we could be more precise here and keep track of what
+        // modules the instances depend on and only flush the modules to index
+        // when you you see that you have an instance that you are about to
+        // write to disk that depends on a module that is part of this
+        // operation.
```
lukemelia, line 557: Do we need to / want to trigger an incremental index event for each stage of this, or only once both are complete?

packages/runtime-common/realm-index-query-engine.ts
```diff
@@ -191,13 +192,20 @@ export class RealmIndexQueryEngine {
     return await this.#indexQueryEngine.getModule(url, opts);
   }
 
-  async cardDef(
+  async getOwnCardDef(
```
lukemelia, line 195: What is the difference between getOwnCardDef and getCardDef?

lukemelia (approved):

packages/host/tests/integration/realm-test.ts
habdelra: yeah, so the idea is that the indexing would definitely encounter an error (since it uses the same deserialization that used to happen in the realm as part of file serialization) and you'd end up with an error doc. i guess the error expresses itself in a different way. i think we can rewrite this test so we see the error in the new way. I don't think we have to do anything additional with the way we write to disk or perform any new type of validation. the error just happens now after the file has been written to disk as opposed to before (which breaks some of this test's assertions since it expects  that the instance is never serialized to the file system when it is in this state)

packages/runtime-common/realm-index-query-engine.ts
habdelra: getOwnCardDef only returns the card def meta that are in this realm (used for serving card meta from the realm HTTP API) and uses the DB to look up the requested card def meta. getCarDef will use HTTP to look up any card def meta from any realm--this is used for loading meta for search queries. in search queries you can totally provide a code ref for a type that doesn't live on the realm your are searching within (like filtering all instances in this realm by a card def that lives in the catalog). Or for file serialization, you can serialize an instance to a realm that uses a card def that doesn't come from the realm you are writing to.

packages/runtime-common/realm.ts
habdelra: I don't have a strong opinion here. I could go either way.

packages/realm-server/tests/indexing-test.ts
habdelra: A batch could have more than 2 parts. it is all up to the client in terms of how they organize the operation list. If the client creates a list that is a bunch of modules, followed by a bunch of instances, followed by a bunch of modules, followed by a bunch of instances, then the patch will have 4 parts. the idea is that to be safe we always flush any modules to the index before we start processing instances, since the instances may have card def meta necessary for file serialization that is being provided by the preceding modules. so it's really up to the client to organize the operation in an optimal manner. if the client separates the operation into one segment of modules and another segment of instances that would be ideal...

packages/host/tests/integration/realm-test.ts
```diff
@@ -2495,37 +2496,11 @@ module('Integration | realm', function (hooks) {
       }),
     );
 
-    assert.strictEqual(response.status, 400, '400 server error');
-    let fileRef = await adapter.openFile('dir/driver.json');
```
habdelra, line 2575: this scenario now becomes a little different. we still detect the error, but only after writing the instance to disk. in return the index now holds the error as an error document (with a message noted in the new assertion that indicates a field validation problem)

habdelra:
This PR had nothing to do with the percy diff, but this diff seems concerning--it's materially different.... 

image

/cc @lukemelia @jurgenwerk

tintinthong (approved):

packages/runtime-common/realm.ts
lukemelia: If the contract here is that this could be used from the "atomic" endpoint, I think I would only expect one notification, after all of the requested updates have been indexed.

habdelra:
Closing this in favor of the rolled up PR: https://github.com/cardstack/boxel/pull/3067

PR #3067, opened 2025-08-02
Remove loader from Realm file serialization

habdelra:
This PR removes the loader necessary for serializing instances to the file system from the Realm. As part of this it became apparent that we actually do need to store field-defs meta as well as card-def meta in our index (I wasn't considering the needs around polymorphic fields).

⚠️ Note that this PR needs to have pnpm full-reindex performed after merging!

TODO:
- [x] rename "meta" => "definition" (in commit https://github.com/cardstack/boxel/pull/3067/commits/8ab498e0154a7156210cfea8bb38f2aed40cd078)
- [x] issue a single Realm event for the atomic operations that need to be batched


packages/host/tests/acceptance/catalog-app-test.gts
```diff
@@ -392,7 +392,7 @@ module('Acceptance | Catalog | catalog app tests', function (hooks) {
         ],
       });
 
-      await waitFor('.catalog-content');
+      await waitFor('.catalog-content', { timeout: 5_000 });
```
tintinthong, line 396: :thumbsup. We've been meaning to not depend on the catalog rather a mock catalog inside the test realm

packages/runtime-common/file-serializer.ts
```diff
@@ -0,0 +1,369 @@
+import {
```
tintinthong, line 1: 👍

tintinthong (approved):
lgtm

PR #3072, opened 2025-08-04
Elevate context provision into container

tintinthong:
Most important change is to elevate the context provision into container so code-submode can consume from it

- just provide entire context without explicitly breaking them into keys
- ensure overrides occur at the bottom of the context getter 
- remove cardContext arg from card renderer

This will be followed up by the PRs such as https://linear.app/cardstack/issue/CS-9195/allrealmsinfo and the general project of moving commands and context accesible in code mode

### DIFF
--- a/packages/host/app/components/card-renderer.gts
+++ b/packages/host/app/components/card-renderer.gts
@@ -1,5 +1,3 @@
-import { service } from '@ember/service';
-
 import Component from '@glimmer/component';
 
 import { provide, consume } from 'ember-provide-consume-context';
@@ -17,15 +15,12 @@ import {
   type getCardCollection,
 } from '@cardstack/runtime-common';
 
-import type StoreService from '@cardstack/host/services/store';
-
 import type {
   BaseDef,
   Format,
   Field,
+  CardContext,
 } from 'https://cardstack.com/base/card-api';
-
-import PrerenderedCardSearch from './prerendered-card-search';
 
 interface Signature {
   Element: any;
@@ -34,7 +29,6 @@ interface Signature {
     format?: Format;
     field?: Field;
     codeRef?: ResolvedCodeRef;
-    cardContext?: Record<string, any>;
     displayContainer?: boolean;
   };
 }
@@ -44,7 +38,7 @@ export default class CardRenderer extends Component<Signature> {
   @consume(GetCardsContextName) private declare getCards: getCards;
   @consume(GetCardCollectionContextName)
   private declare getCardCollection: getCardCollection;
-  @service private declare store: StoreService;
+  @consume(CardContextName) private declare cardContext: CardContext;
 
   @provide(DefaultFormatsContextName)
   // @ts-ignore "defaultFormat is declared but not used"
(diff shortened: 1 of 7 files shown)


packages/host/app/components/card-renderer.gts
```diff
@@ -17,24 +15,20 @@ import {
...
-import PrerenderedCardSearch from './prerendered-card-search';
-
 interface Signature {
   Element: any;
   Args: {
     card: BaseDef;
     format?: Format;
     field?: Field;
     codeRef?: ResolvedCodeRef;
-    cardContext?: Record<string, any>;
```
tintinthong, line 37: im not convinced that there is a direct use for passing in this argument when we are already using dynamic context

packages/host/app/templates/card.gts
```diff
@@ -68,7 +69,9 @@ class HostModeComponent extends Component<HostModeComponentSignature> {
     return false;
   }
 
-  private get cardContext(): HostModeCardContext {
+  @provide(CardContextName)
+  // @ts-ignore "context" is declared but not used
+  private get context(): HostModeCardContext {
```
tintinthong, line 74: im not sure if we possibly want to provide another context name here like HostCardContextName since HostCardContext by type seems to exclude prerenderedSearch but I dont know why

packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -129,6 +134,7 @@ function urlToFilename(url: URL) {
 
 export default class CodeSubmode extends Component<Signature> {
   @consume(GetCardContextName) private declare getCard: getCard;
+  @consume(CardContextName) private declare cardContext: CardContext;
```
tintinthong, line 137: consumes cardContext here so cards do thing in code mode

packages/host/app/components/operator-mode/stack-item.gts
```diff
@@ -109,13 +110,12 @@ export interface StackItemRenderedCardForOverlayActions
   stackItem: StackItem;
 }
 
-type StackItemCardContext = Omit<CardContext, 'prerenderedCardSearchComponent'>;
```
tintinthong, line 112: why wud we specify type to Omit the prerenderedCardSearchComponent here. Dont think this is needed

packages/host/app/components/card-renderer.gts
habdelra: i agree. is it something that we provide for testing? in which case your comment is still correct, and tests should use a provider if necessary....

packages/host/app/templates/card.gts
habdelra: I like to err on the side of adding more context names as opposed to overloading the default context. but i have no strong feelings here...

habdelra (approved):
this is a nice refactor

lukemelia (commented):
I'm not convinced that we should be using dynamic context outside of cards, which this refactor will invite people to do. I'd like to get @ef4's perspective before merging.

PR #3079, opened 2025-08-05
Add more cardInfo fields

burieberry:
- Adds the cardInfo fields: `title`, `description`, `thumbnailURL`, `notes`
- Change CardDef's `title`, `description`, `thumbnailURL` fields to be computed

Updated default isolated and edit templates:
- Default isolated template has a header with the result of computed fields, and notes footer
- Default edit template's header displays the cardInfo's `title`, `description`, `thumbnailURL`, if they are used to calculate the computed `title`, `description`, `thumbnailURL` fields
- Fallback for title is `Untitled [CARD_TYPE_DISPLAY_NAME]`
- Fallback for thumbnailURL is card's static icon
meeting-isolated
meeting-edit

## Different example with card's own computed title and thumbnailURL fields:
- If the card already has one of these fields (`title`, `description`, `thumbnailURL`) as a computed of a different field, we display the computed field because in this case, changing the cardInfo field would not change the actual title of the card. For example, the below card has a `title` field, which is a computed of the `name` field, and a `thumbnailURL` field, which is a computed of the `picture` field. In this case, this card's `title` and `thumbnailURL` fields override CardDef's `title` and `thumbnailURL`:

default-isolated
edit-1
edit-2

## Next Steps -- I'll open tickets for the following:
- Add `tags` field to `cardInfo`. Latest decision is that `tags` should be a `containsMany` and not a linksToMany as previously thought.
- Add the `cardInfo` fields to the `...` context menu of a card with custom isolated/edit templates


tintinthong (approved):

lukemelia (approved):
Besides already discussed issue of not bloating card-api.gts, it looks good

PR #3083, opened 2025-08-06
Create online store listing

lucaslyl:
Linear: https://linear.app/cardstack/issue/ECO-277/create-online-store-listing

online-store-listing-01


tintinthong (approved):
is there any opportunity for breaking down the field and cards here? can be done in a separate PR but just want to make sure we r thinking about it

lucaslyl:
> is there any opportunity for breaking down the field and cards here? can be done in a separate PR but just want to make sure we r thinking about it 

Oh, I just realized there was this comment.
Yes, I’ll break it down into three separate cards for orders, products, and customers in a new PR

PR #3086, opened 2025-08-06
Add caching to card-def meta

habdelra:
This PR adds a caching layer to our card def meta access. Which similar to the now removed Loader will invalidate whenever modules change in the index.

### DIFF
--- a/packages/host/tests/unit/index-query-engine-test.ts
+++ b/packages/host/tests/unit/index-query-engine-test.ts
@@ -15,6 +15,7 @@ import {
   type ResolvedCodeRef,
   type CardDefMeta,
 } from '@cardstack/runtime-common';
+import { DefinitionsCache } from '@cardstack/runtime-common/definitions-cache';
 
 import { runSharedTest } from '@cardstack/runtime-common/helpers';
 // eslint-disable-next-line ember/no-test-import-export
@@ -45,6 +46,7 @@ module('Unit | query', function (hooks) {
   let indexQueryEngine: IndexQueryEngine;
   let loader: Loader;
   let testCards: { [name: string]: CardDef } = {};
+  let definitionsCache: DefinitionsCache;
 
   hooks.before(async function () {
     dbAdapter = await getDbAdapter();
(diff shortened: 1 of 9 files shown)


packages/host/tests/unit/index-query-engine-test.ts
```diff
@@ -238,13 +240,15 @@ module('Unit | query', function (hooks) {
...
-    indexQueryEngine = new IndexQueryEngine(dbAdapter, virtualNetwork.fetch);
+    definitionsCache = new DefinitionsCache(virtualNetwork.fetch);
+    indexQueryEngine = new IndexQueryEngine(dbAdapter, definitionsCache);
   });
 
   test('can get all cards with empty filter', async function (assert) {
     await runSharedTest(indexQueryEngineTests, assert, {
       indexQueryEngine,
       dbAdapter,
+      definitionsCache,
```
habdelra, line 251: all these `definitionsCache` items in the tests will get elided in the future PR to remove Loader form the node context as we move the tests around. so this is just temporary...

packages/runtime-common/definitions-cache.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+    let url = `${realmURL}_card-def?${qs.stringify({ codeRef })}`;
+    let response: Response;
+    try {
+      response = await this.#fetch(url, {
+        headers: { accept: SupportedMimeType.JSONAPI },
+      });
+    } catch (e) {
+      throw new Error(
+        `Your filter refers to a nonexistent type: import { ${codeRef.name} } from "${codeRef.module}"`,
```
lukemelia, line 71: Possible improvement: cache not founds

lukemelia (approved):

packages/runtime-common/definitions-cache.ts
habdelra: i'm not understanding what the suggestion is. if the type is non existent isn't it obvious that the cache entry was not found?

habdelra: oh sorry, i got it now--add these URL's to our cache. so just FYI, the loader does not actually do this caching. i think without cross realm invalidation, it might be a bad idea to cache 404's, since there would be no way to invalidate the 404 if it came from a different realm in the scenario where you fixed the broken link.

lukemelia: I don't think its a blocker, and don't know whether it is a good or bad idea.

PR #3089, opened 2025-08-07
Fix stub listing missing tag

richardhjtan:


packages/host/app/commands/listing-create.ts
```diff
@@ -69,7 +69,9 @@ export default class ListingCreateCommand extends HostBaseCommand<
   }
 
   get catalogRealm() {
-    return this.realmServer.catalogRealmURLs[0];
+    return this.realmServer.catalogRealmURLs.find((realm) =>
+      realm.includes('catalog'),
+    );
```
richardhjtan: This is to fix sometime the first catalog realm URL is returning skill realm. I can reproduce it after full-reset and Lucas faces the same issue

packages/catalog-realm/CardListing/contract.json
```diff
@@ -1,56 +1,63 @@
 {
```
richardhjtan, line 1: The file structure was changed, which might be due to the serializer

packages/catalog-realm/CardListing/product-catalog.json
```diff
@@ -1,58 +1,63 @@
...
       "images": [
         "https://dummyimage.com/400x300/f0fdfa/0d9488&text=Product+Catalog"
       ],
-      "isStub": true,
```
tintinthong, line 10: what happened to `isStub`

tintinthong: I can see it still in all the other listings

richardhjtan: @tintinthong is the `isStub` adding manually within the JSON when we newly introduce the stubs? I can't see there is field to store `isStub`value

but rather right now we check whether `isStub` by a getter

tintinthong: ok then remove isStub from the json

richardhjtan: Sure, I will trigger to edit all stub files to clean up and have the same format, which is processed by the serializer

packages/catalog-realm/CardListing/contract.json
habdelra: yeah, we have a new file serializer now. the structure is slightly different and more correct. also it's reordering a bunch of the properties in the JSON

packages/host/app/commands/listing-create.ts
habdelra: this is a pretty loose criteria for search. maybe make it more precise? like `.endsWith("/catalog/")`

tintinthong (approved):

PR #3090, opened 2025-08-07
Implement rendering of card in host submode

FadhlanR:
This PR implements card rendering in host submode to provide a preview of the public host mode. Key changes:

- **Card Rendering**: Host submode now displays actual rendered cards instead of just showing card IDs or trails
- **Wide Format Support**: Uses the `prefersWideFormat` field to automatically adjust card width (full width vs standard width)
- **Improved Code-to-Host Logic**: Enhanced logic when switching from code submode to host submode to determine which card to display based on context (card instances, playground selections, or related cards)

https://github.com/user-attachments/assets/c8286d48-c934-4448-9b37-01dd1692da6e

### DIFF
--- a/packages/host/app/components/operator-mode/host-submode.gts
+++ b/packages/host/app/components/operator-mode/host-submode.gts
@@ -8,14 +8,7 @@ import { consume } from 'ember-provide-consume-context';
 
 import { BoxelButton, CardContainer } from '@cardstack/boxel-ui/components';
 
-import {
-  type getCard,
-  type getCards,
-  type getCardCollection,
-  GetCardContextName,
-  GetCardsContextName,
-  GetCardCollectionContextName,
-} from '@cardstack/runtime-common';
+import { type getCard, GetCardContextName } from '@cardstack/runtime-common';
 import { meta } from '@cardstack/runtime-common/constants';
 
 import CardRenderer from '@cardstack/host/components/card-renderer';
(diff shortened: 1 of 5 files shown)


packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -15,53 +27,200 @@ interface HostSubmodeSignature {
 }
 
 export default class HostSubmode extends Component<HostSubmodeSignature> {
+  @consume(GetCardContextName) private declare getCard: getCard;
```
lukemelia: Don't do this. Use a service.

lukemelia (approved):

packages/host/app/components/operator-mode/host-submode.gts
habdelra: ooof this was what you feared @lukemelia. perhaps there is a good way to teach this to the rest of the team?

habdelra: we actually have a bunch of call sites that are not in cards where we do this--i think @FadhlanR was just following a pattern. perhaps we need to open a ticket to refactor this?

lukemelia: I'll create one

lukemelia: Likely should just do `import { getCard } from '@cardstack/host/resources/card-resource';`

habdelra: So when we did the refactor that introduced getCard, @ef4 recommended using consume over a direct import so that we could close over getCard. I’d love to get Ed’s feedback.

lukemelia: Not sure what you mean by "close over" in this case?

PR #3092, opened 2025-08-07
Allow restoring/copying content on user provided file attachments

jurgenwerk:
image

image

image


packages/host/app/components/ai-assistant/attached-file-dropdown-menu.gts
```diff
@@ -0,0 +1,176 @@
...
+    ) {
+      return;
+    }
+    this.operatorModeStateService.updateSubmode(Submodes.Code);
+    this.operatorModeStateService.updateCodePath(
+      new URL(this.args.file.sourceUrl!),
+    );
+  }
+
+  copySubmittedContent = async () => {
```
habdelra: this casts a EC task to a promise which bascially removes the cancellability and will throw if cancelled. Please only use EC tasks in a component. If this is public API (called by the outside world) i think you might want to rethink this...

jurgenwerk: My reason for making it so is because of the `MenuItem`, whose API is to pass an action and it doesn't support passing EC tasks:

```
new MenuItem(`Copy ${submittedOrGenerated} Content`, 'action', {
  action: this.copySubmittedContent,
  icon: Copy,
  disabled: !this.args.file?.sourceUrl || this.args.isNewFile,
  }),
```

lukemelia (approved):

packages/host/app/components/ai-assistant/attached-file-dropdown-menu.gts
habdelra: can we just pass the perform() of the EC task instead of casting it to a promise? like literally just remove the `async` part of the method signature.

jurgenwerk: good point, I think we can also just replace `action: this.copySubmittedContent` with `action: this.copySubmittedContentTask.perform`, and get rid of that `copySubmittedContent` method

jurgenwerk: Fixed in https://github.com/cardstack/boxel/pull/3092/commits/120fca9bd5b3b34a0fb0d93ecc66288d7e6c4846

PR #3093, opened 2025-08-07
Refactor submode handling to use SwitchSubmodeCommand and remove from publicApi actions

lukemelia:


ef4:
The type issue was harder than I expected.

I started working through all the tsconfigs and extending them so they'd understand all the types inside host. This is necessary because as of this PR, anything that depends on base will now also pull in all the types from host.

I figured I would get type checks passing and then recommend we do a bigger refactor of the way the tsconfigs all relate to each other, because the current way is really unworkable. But before I could get there I reached a limit where typescript is erroring on "excessively deep type instantiation". So I think the bigger config change might be mandatory.

ef4:
Pushed [a branch](https://github.com/cardstack/boxel/tree/typescript-config-wip) with the in-progress work for how far I got before I hit the "excessively deep" error.

lukemelia:
I've used a ts-ignore to consume a command from base. We have a separate effort underway to address this.

tintinthong (approved):

PR #3094, opened 2025-08-07
Refactor to remove delete from publicApi.actions

lukemelia:


packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -476,6 +439,45 @@ export default class InteractSubmode extends Component {
     this.selectCards.perform(selectedCards, stackItem);
   }
 
+  @action
+  private async requestDeleteCard(card: CardDef | URL | string): Promise<void> {
+    let cardToDelete: CardToDelete | undefined;
+    let here = this;
```
backspace: Is this still needed if it’s an action?

backspace (approved):

packages/host/app/components/operator-mode/interact-submode.gts
lukemelia: Good catch

lukemelia (commented):
Fixed.

PR #3100, opened 2025-08-08
add skill from /skills to boxel tools

tintinthong:


packages/vscode-boxel-tools/src/skills.ts
```diff
@@ -254,7 +261,17 @@ export class SkillsProvider
           }),
         );
 
-        await Promise.all(loadingPromises);
+        const results = await Promise.allSettled(loadingPromises);
```
tintinthong, line 264: surprisingly this has been broken bcos /seed/ which is non-existing is still part of the realms that we r trying to login to

PR #3101, opened 2025-08-08
Eco 290 stubs break online store into smaller stubs

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-290/stubs-break-online-store-into-smaller-stubs

online-product-listing-01
online-customer-listing-01
online-order-listing-01


tintinthong (approved):
1 listing into 3 listings 👍. Possible to make a lot of comments about the smaller cards being standalone but I think this is good for now

PR #3103, opened 2025-08-11
Implement proxy endpoint in realm server

FadhlanR:
This PR introduces a new endpoint, `_request-proxy`, which acts as a proxy to call external endpoints. It helps avoid CORS issues and, for specific known endpoints, allows us to inject our API keys and charge for usage.


packages/realm-server/lib/credit-strategies.ts
```diff
@@ -0,0 +1,103 @@
...
+// Default AI Bot Credit Strategy (reused from AI bot)
+export class OpenRouterCreditStrategy implements CreditStrategy {
+  name = 'openrouter-credit-strategy';
+  openRouterApiKey: string;
+
+  constructor(openRouterApiKey: string) {
+    this.openRouterApiKey = openRouterApiKey;
+  }
+
+  async validateCredits(dbAdapter: DBAdapter, matrixUserId: string) {
```
jurgenwerk, line 37: Can this actually be made reusable? I.e. extracting the logic around `availableCredits < MINIMUM_AI_CREDITS_TO_CONTINUE` so that it can be imported both here and in the AI bot

packages/realm-server/handlers/handle-request-forward.ts
```diff
@@ -0,0 +1,352 @@
...
+          'Request body must include url, method, and requestBody fields',
+        );
+        return;
+      }
+
+      // 3. Validate external endpoint is whitelisted
+      const endpointsConfig = ExternalEndpointsConfig.getInstance(
+        externalEndpointsConfig,
+      );
+      if (!endpointsConfig.isEndpointWhitelisted(json.url)) {
```
jurgenwerk: Is it beneficial to have a whitelisted property? Would it be simpler to have a rule that the config includes whitelisted ones only? So if it's not in the config then it's not whitelisted

packages/realm-server/handlers/handle-request-forward.ts
```diff
@@ -0,0 +1,352 @@
...
+        const allowedEndpoints = endpointsConfig.getAllowedEndpoints();
+        await sendResponseForBadRequest(
+          ctxt,
+          `Endpoint ${json.url} is not whitelisted. Allowed endpoints: ${allowedEndpoints.join(', ')}`,
```
jurgenwerk: From a security standpoint - do we really wanna expose all our allowed endpoints config (in an error response in this case)?

packages/realm-server/handlers/handle-request-forward.ts
```diff
@@ -0,0 +1,352 @@
...
+  headers?: Record<string, string>;
+  stream?: boolean;
+}
+
+export default function handleRequestForward({
+  dbAdapter,
+  externalEndpointsConfig,
+}: {
+  dbAdapter: DBAdapter;
+  externalEndpointsConfig: string;
```
jurgenwerk: Suggestion to rename this to `externalEndpointsForProxying` to make it clearer (and env var also)

jurgenwerk (approved):
DIdn't test it but it looks like it makes sense 👍 Some minor comments.

packages/realm-server/handlers/handle-request-forward.ts
lukemelia: `allowedProxyDestinations`

lukemelia (approved):

PR #3105, opened 2025-08-11
Remove places where we run custom card code

habdelra:
This removes running the `[deserialize]` custom card function, as well as running the `[emptyValue]` custom card function.


lukemelia (approved):
Any docs or skills that need updating wit this change?

habdelra:
> Any docs or skills that need updating wit this change?

lemme double check...

habdelra:
looks like no skills nor readme's mention the deserializer and empty value base def hooks.

PR #3106, opened 2025-08-11
Add requireInputFields and ignoreInputFields properties to command

lukemelia:
- they influence JSON Schema generation

### DIFF
--- a/packages/host/tests/acceptance/commands-test.gts
+++ b/packages/host/tests/acceptance/commands-test.gts
@@ -1245,7 +1245,6 @@ module('Acceptance | Commands tests', function (hooks) {
             description:
               'Displaying the card with the Latin word for milkweed in the title.',
             attributes: {
-              unsupportedAttribute: 'http://test-realm/test/Person/hassan',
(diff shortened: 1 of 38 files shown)


packages/host/tests/acceptance/commands-test.gts
```diff
@@ -1265,7 +1264,7 @@ module('Acceptance | Commands tests', function (hooks) {
     assert
       .dom('[data-test-boxel-alert="warning"]')
       .containsText(
-        'Command "show-card_566f" validation failed: data/attributes must NOT have additional properties',
+        'Command "show-card_566f" validation failed: data/attributes must have required property \'cardId\'',
```
backspace, line 1267: This seems like a sort of opposite error, did it change?

backspace (approved):

packages/host/tests/acceptance/commands-test.gts
lukemelia: Yes, I removed the restrictions about additional fields and added the ability to specify required fields.

PR #3108, opened 2025-08-12
add loading state to menu for fetching data

tintinthong:
https://github.com/user-attachments/assets/ee104b11-e7d3-47da-85a2-23a86a49e1c7

### DIFF
--- a/packages/boxel-ui/addon/src/components/dropdown/index.gts
+++ b/packages/boxel-ui/addon/src/components/dropdown/index.gts
@@ -105,7 +105,7 @@ class BoxelDropdown extends Component<Signature> {
             )
             onDeactivate=dd.actions.close
             allowOutsideClick=true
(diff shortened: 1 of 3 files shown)


richardhjtan (approved):

packages/boxel-ui/addon/src/components/dropdown/index.gts
```diff
@@ -105,6 +105,7 @@ class BoxelDropdown extends Component<Signature> {
             )
             onDeactivate=dd.actions.close
             allowOutsideClick=true
+            fallbackFocus=(concat "#ember-basic-dropdown-content-" dd.uniqueId)
```
tintinthong: There are focus trap errors here that occurs in low-level `focus-trap` library 

```
Your focus-trap must have at least one container with at least one tabbable node in it at all times
```

It happens when there are no buttons to focus on so we put a fallback

tintinthong:
Better design 

Screenshot 2025-08-12 at 14 55 21

PR #3109, opened 2025-08-12
Eco 324 geo point field

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-324/geo-point-field

- Integrate a Leaflet map with a geo-point field
- Update latitude and longitude when the map is clicked

Screenshot 2025-08-12 at 4 10 35 PM


packages/catalog-realm/fields/geo-point.gts
```diff
@@ -0,0 +1,134 @@
...
+import NumberField from 'https://cardstack.com/base/number';
+import MapIcon from '@cardstack/boxel-icons/map';
+import { MapRender } from '../components/map-render';
+import { action } from '@ember/object';
+
+import { FieldContainer } from '@cardstack/boxel-ui/components';
+
+class AtomTemplate extends Component<typeof GeoPointField> {
+  get latValue() {
+    return this.args.model?.lat ?? 'N/A';
```
tintinthong, line 16: seems like a bad default

packages/catalog-realm/components/map-render.gts
```diff
@@ -0,0 +1,224 @@
...
+
+        // Use helper function for initial popup setup
+        setupMarkerPopup(
+          this.currentMarker,
+          { lat: defaultLat, lng: defaultLon },
+          map,
+          'Current Location',
+        );
+
+        map.on('click', (e: any) => {
```
tintinthong: e seems like a weird usage. You want event right

tintinthong (approved):
Nice looks good

packages/catalog-realm/components/map-render.gts
lucaslyl: Yes, this is a Leaflet map event type. I’ve updated my code to replace it with this event type

PR #3111, opened 2025-08-12
Update grid layout in command header for better responsiveness

lukemelia:
The problem:

image

With the solution:

image

### DIFF
--- a/packages/host/app/components/ai-assistant/code-block/command-header.gts
+++ b/packages/host/app/components/ai-assistant/code-block/command-header.gts
@@ -44,7 +44,7 @@ const CodeBlockCommandHeader: TemplateOnlyComponent<CodeBlockCommandHeaderSignat
     <style scoped>
       .code-block-header {
         display: grid;
-        grid-template-columns: 1fr auto;
+        grid-template-columns: minmax(0, 1fr) max-content;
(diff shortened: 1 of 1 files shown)


packages/host/app/components/ai-assistant/code-block/command-header.gts
```diff
@@ -61,6 +61,7 @@ const CodeBlockCommandHeader: TemplateOnlyComponent<CodeBlockCommandHeaderSignat
         letter-spacing: var(--boxel-lsp-xs);
         line-height: 1.5em;
         text-wrap: pretty;
+        overflow-wrap: break-word;
```
burieberry, line 64: Are we ok with the word potentially breaking into something like this:
```
Canc
el
```

burieberry (approved):
Looks good. Made a note about the word-break though to confirm.

packages/host/app/components/ai-assistant/code-block/command-header.gts
lukemelia: I don't love that, but I am OK with it if that is the only way to avoid breaking the layout. WDYT?

burieberry: Alternative is using ellipsis. Up to you

PR #3117, opened 2025-08-13
Show code diff error state if search&replace does not apply

jurgenwerk:
This introduces a new error state in the diff editor for the case when a diff does not apply to the source file being edited. Some known case of when this can happen:

1. File contents has changed during streaming
2. LLM does not have proper context about the current file contents
3. LLM hallucinates or makes textual errors in the search/replace block

It looks like this:

image

This is how it looks when there are multiple code patches, and some of them are good and some bad. The error state will show up before accepting:
image

The user can still hit "Accept all", and for the faulty diff, a new `CodePatchResult` event with status `failed` will be sent. It includes a verbose `failureReason` so that the LLM knows the code patch did not work (due to faulty search block in the search&replace block). After receiving the mentioned event with failed status, it will automatically send another patch. In ask mode it will ask to accept, but in the act mode it will loop until a valid code diff can be applied.

I used this prompt to test this:

```
change border from red to green, and add some emojis to the footer.

make 2 search/replace blocks for this.

in the first search&replace block, deliberately introduce the X character at the end of the search portion in the search& replace block. do this only for this response and not for next responses
```


packages/host/app/commands/apply-search-replace-block.ts
```diff
@@ -8,6 +8,14 @@ import type * as BaseCommandModule from 'https://cardstack.com/base/command';
 
 import HostBaseCommand from '../lib/host-base-command';
 
+let standardErrorMessage =
+  'Unable to process the code patch due to invalid code coming from AI';
+export const APPLY_SEARCH_REPLACE_BLOCK_ERROR_MESSAGES = {
+  SEARCH_BLOCK_PARSE_ERROR: `${standardErrorMessage} (search block parse error)`,
+  REPLACE_BLOCK_PARSE_ERROR: `${standardErrorMessage} (replace block parse error)`,
+  SEARCH_PATTERN_NOT_FOUND: `${standardErrorMessage} (search pattern not found in the attached source file)`,
```
lukemelia:
```suggestion
  SEARCH_PATTERN_NOT_FOUND: `${standardErrorMessage} (search pattern not found in the target source file)`,
```

lukemelia (approved):
I'm thinking we should proactively send a failed/invalid CodePatchResult event after the message completes (in CommandService) rather than wait for the user to press Accept All. WDYT?

packages/host/app/components/ai-assistant/code-block/patch-footer.gts
```diff
@@ -0,0 +1,33 @@
...
+    <style scoped>
+      .code-block-patch-footer {
+        display: flex;
+        align-items: center;
+        justify-content: space-between;
+        gap: var(--boxel-sp-xs);
+        background-color: var(--boxel-650);
+        color: var(--boxel-light);
+        padding: 8px 12px;
+        font-size: 14px;
```
burieberry: let's make css variables for these for in a parent component. they seem to also be used in header. this way it'll be easier to make any changes consistent.

burieberry: This font-size value seems to be redundant, given that the Alert component is overriding it. In general, it would be good to stick to boxel-font variables for consistency in size and line-height.

packages/host/app/components/ai-assistant/code-block/patch-footer.gts
```diff
@@ -0,0 +1,33 @@
...
+        height: 50px;
```
burieberry: This should probably be `min-height` to account for longer text.
long-text

packages/host/app/components/ai-assistant/code-block/patch-footer.gts
```diff
@@ -0,0 +1,33 @@
...
+      }
+
+      :deep(.code-patch-error) {
+        padding: 0;
+      }
+    </style>
+    <div class='code-block-patch-footer'>{{yield}}</div>
```
burieberry: we can use the `footer` html tag here instead of `div`

burieberry (commented):
Made some template/css comments. Looks good otherwise

PR #3118, opened 2025-08-13
Refactor IdentityContext to "CardStore" and add a load() method

habdelra:
This PR refactors the IdentityContext include a `load()` method and renames it to a `CardStore`. I opted to only perform the fetching in the `load()` and not the deserialization, as the deserialization requires special handling because the store we use in the host app is tracked. without the special handling we would see all the sausage making of the deserialization process in our templates due to the tracked nature of the store. Additionally, I provided a fallback store that uses a native fetch. this is only something that would be used in tests.


packages/base/card-api.gts
```diff
@@ -220,9 +222,10 @@ const loadLinksPromises = initSharedState(
   'loadLinksPromises',
   () => new WeakMap<BaseDef, Promise<any>>(),
 );
-const identityContexts = initSharedState(
-  'identityContexts',
-  () => new WeakMap<BaseDef, IdentityContext>(),
+// TODO do we really need this???
+const stores = initSharedState(
+  'stores',
+  () => new WeakMap<BaseDef, CardStore>(),
```
habdelra, line 228: look into this more...

packages/host/app/lib/current-run.ts
```diff
@@ -408,8 +408,11 @@ export class CurrentRun {
     if (hasExecutableExtension(url.href)) {
       await this.indexModule(url, fileRef);
     } else {
-      if (!identityContext) {
-        identityContext = new IdentityContextWithErrors();
+      if (!store) {
+        // the special building index header is set on the middleware stack
+        // specifically in the loader's fetcher (as opposed to the virtual
+        // network's fetcher), so we need that fetcher specifically for our store.
+        store = new CardStoreWithErrors(this.loaderService.loader.fetch);
```
habdelra, line 415: @lukemelia why do we set the `X-Boxel-Building-Index` header on the middleware used in the loader's fetcher and not on the virtual-network's fetcher? It seems weird that I needed to expose the fetcher that the loader uses in order to get access to a fetcher that uses this special header.

packages/base/card-api.gts
habdelra: this seems related to how we deal with the NotLoaded errors. I'll dive more into the need for this in CS-9036

packages/host/app/lib/current-run.ts
lukemelia: I'm not sure if I have a good answer for that. This header is only ever relevant during indexing, so perhaps we thought the loader was a narrower way to target it?

lukemelia (approved):

PR #3120, opened 2025-08-13
Make sure fitted components are rendered with the right CSS and that our code which renders them does so with proper sizing.

lukemelia:


burieberry (commented):
The linksTo card here looks 100% height and width, too. Maybe we need a max-height per linked item.
percy

lukemelia:
I made it so that delegated fitted format renders have a min-height of 40px and a max-height of 600px, which is how are skills and playground describes the valid range.

burieberry:
I'll take a better look later but some items now seem to be cropped at 40px and their basic-fitted template is not respecting the 40px height restriction. This is the experiments-realm Person card:
fitted

I saw that you added some extra `div`s around linked items, is that something the users will be required to do when using linked fields?

burieberry (commented):
Here's what I suggest: `.fitted-template` and `.field-component-card.fitted-format > :deep(*)` should be set to 100% height and width. 

`.field-component-card.fitted-format` should have by default a fitted size (I think we picked 65px height). Then the card author should be able to pass in other fitted view sizes as desired. For example, `@size="small-tile"` would set the `.field-component-card.fitted-format` to be 150px wide x 170px high. 

If you want `.field-component-card.fitted-format` to have width and height 100%, then we need to move the `container-name: fitted-card; container-type: size;` to a parent that's one level higher. And then you can set the `height: 65px` there and the fitted variables would be passed in to that container.

I pushed a spike I did a while ago that explores this idea, branch name: `spike-prerendered-search-card-sizes`. Screenshot below. The main idea is that the inner component needs to have 100% width and height and the outer component needs to have the container properties with at least the specific height value for the sizing to be respected.

fitted-sizes

PR #3121, opened 2025-08-14
Gaming hub listing feedback

richardhjtan:


packages/catalog-realm/base-entity/game.gts
```diff
@@ -0,0 +1,15 @@
...
+import StringField from 'https://cardstack.com/base/string';
+import MarkdownField from 'https://cardstack.com/base/markdown';
+
+import GamepadIcon from '@cardstack/boxel-icons/gamepad-2';
+
+export class Game extends CardDef {
+  static displayName = 'Game';
+  static icon = GamepadIcon;
+
+  @field title = contains(StringField);
```
richardhjtan, line 12: I overrode the title because currently the default title is computed from `cardInfo.title`. Mostly AI will write sample data with `title` at root level - not under cardInfo. Else it will display as `Untitled {CardName}`

tintinthong (approved):

PR #3124, opened 2025-08-14
Implement a command that calls proxy endpoint

FadhlanR:
The proxy endpoint in the realm server was already implemented in another PR. In this PR, we wrap the call to that endpoint in a command, allowing cards to access it. The command implemented here is fairly simple—it just wraps the request to, and the response from, the proxy endpoint.

### DIFF
--- a/packages/host/app/commands/request-forward.ts
+++ b/packages/host/app/commands/request-forward.ts
@@ -1,0 +1,69 @@
+import { service } from '@ember/service';
+
+import type * as BaseCommandModule from 'https://cardstack.com/base/command';
+
+import HostBaseCommand from '../lib/host-base-command';
+
+import type RealmServerService from '../services/realm-server';
+
+export default class RequestForwardCommand extends HostBaseCommand<
+  typeof BaseCommandModule.RequestForwardInput,
+  typeof BaseCommandModule.RequestForwardResult
+> {
+  @service declare private realmServer: RealmServerService;
+
+  static actionVerb = 'Forward Request';
+  description = 'Forward a request to an external API through the realm server';
+
+  async getInputType() {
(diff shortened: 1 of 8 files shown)


packages/host/app/commands/request-forward.ts
```diff
@@ -0,0 +1,69 @@
+import { service } from '@ember/service';
+
+import type * as BaseCommandModule from 'https://cardstack.com/base/command';
+
+import HostBaseCommand from '../lib/host-base-command';
+
+import type RealmServerService from '../services/realm-server';
+
+export default class RequestForwardCommand extends HostBaseCommand<
```
lukemelia:
```suggestion
export default class SendRequestViaProxyCommand extends HostBaseCommand<
```

I think this name is better, and we should update corresponding input, result, etc

packages/host/app/commands/request-forward.ts
```diff
@@ -0,0 +1,69 @@
...
+  typeof BaseCommandModule.RequestForwardInput,
+  typeof BaseCommandModule.RequestForwardResult
+> {
+  @service declare private realmServer: RealmServerService;
+
+  static actionVerb = 'Forward Request';
```
lukemelia:
```suggestion
  static actionVerb = 'Send';
```

packages/host/app/commands/request-forward.ts
```diff
@@ -0,0 +1,69 @@
...
+  description = 'Forward a request to an external API through the realm server';
```
lukemelia:
```suggestion
  description = 'Make a request to an external API through the Boxel proxy';
```

lukemelia (changes requested):

lukemelia (approved):

PR #3126, opened 2025-08-14
Eco 338 geo search point field

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-338/geo-search-point-field

Screenshot 2025-08-14 at 9 43 16 PM

Screenshot 2025-08-14 at 9 43 35 PM


packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,323 @@
...
+      }
+    </style>
+  </template>
+}
+
+// @ts-ignore
+export class GeoSearchPointField extends GeoPointField {
+  static displayName = 'Geo Search Point';
+
+  @field searchKey = contains(StringField);
```
lucaslyl, line 257: I chose to use a StringField instead of an AddressField because AddressField is a compound field that requires filling in a long list of subfields, such as postal code, street, and so on. We’re aiming to follow the approach used by Google Maps and other common UIs, which typically provide just a single input field for searching an address. 
so we can still perform a callback function to update the coordinates accordingly.

packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,323 @@
...
+      if (data && data.length > 0) {
+        (this.args.model as any)[latField] = parseFloat(data[0].lat);
+        (this.args.model as any)[lonField] = parseFloat(data[0].lon);
+      }
+    } catch (error) {
+      console.error('Error geocoding address:', error);
+    }
+  }
+
+  @action
```
tintinthong, line 99: ember concurrency to handle async here

packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,323 @@
...
+  async geocodeAddress() {
+    await this.geocodeAddressInternal(
+      this.args.model.searchKey || '',
+      'lat',
+      'lon',
+    );
+  }
+
+  @action
+  updateSearchAddress(event: Event) {
```
tintinthong: just use the standard debounce pattern

packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,323 @@
...
+    this.addressTimeout = setTimeout(() => {
+      this.geocodeAddress();
+    }, 500);
+  }
+
+  @action
+  updateMarkerColor(color: string) {
+    if (!this.args.model.config) {
+      // Initialize config if it doesn't exis
+      (this.args.model as any).config = {
```
tintinthong: dont use any

packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,323 @@
...
+        markerColor: '#22c55e',
+      };
+    }
+    (this.args.model.config as any).markerColor = color;
```
tintinthong: dont use any

packages/catalog-realm/components/map-render.gts
```diff
@@ -4,9 +4,11 @@ import { action } from '@ember/object';
```
tintinthong, line 4: I can see this component is important. 
- config should have a type here even tho leaflet modifier might not accept type. 
- it will save you a lot of heartache to start grouping lat,lon into a single object 
- I dono how it fits into route. But we should have route as an array of points maybe. How does route handle the points?? There has to be some encoding

packages/catalog-realm/fields/geo-search-point.gts
tintinthong: Yes this is expected for me

tintinthong:
As discussed. Lets remove config from geo point field and forget about colors for now

lucaslyl:
> As discussed. Lets remove config from geo point field and forget about colors for now

You're right! I've already implemented exactly what you suggested. I've grouped lat/lon into coordinate objects, also remove config from geo point/geo search point field from now

packages/catalog-realm/fields-preview/geo-search-point.gts
```diff
@@ -0,0 +1,52 @@
...
+
+      <style scoped>
+        .fields {
+          display: grid;
+          gap: var(--boxel-sp-lg);
+          padding: var(--boxel-sp-xl);
+        }
+      </style>
+    </template>
+
```
tintinthong, line 42: apart from getCardComponent. This is another one we always use. We shud find a way to clean this up. TBD

packages/catalog-realm/fields-preview/GeoSearchPointPreview/2b042a26-1932-47bd-b7c8-92d37e9a7e59.json
```diff
@@ -0,0 +1,34 @@
...
+      "cardInfo": {
+        "notes": null,
+        "title": null,
+        "description": null,
+        "thumbnailURL": null
+      },
+      "geoSearchPoint": {
+        "lat": 3.1336268,
+        "lon": 101.6299203,
+        "config": {
```
tintinthong: remove all config data

packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,268 @@
...
+      }
+    },
+  );
+
+  @action
+  geocodeAddress() {
+    this.geocodeAddressInternal.perform(
+      this.args.model.searchKey || '',
+      'lat',
+      'lon',
```
tintinthong: why is this such a weird api. You pass in the data right. This should be coordinate le

packages/catalog-realm/fields/geo-search-point.gts
```diff
@@ -0,0 +1,268 @@
...
+
+  get latValue() {
+    return this.args.model?.lat ?? 'N/A';
+  }
+
+  get lonValue() {
+    return this.args.model?.lon ?? 'N/A';
+  }
+
+  private geocodeAddressInternal = task(
```
tintinthong: lets call this fetchCoordinates bcos the naming is confusing

packages/catalog-realm/fields/geo-search-point.gts
lucaslyl: Thanks for catching up. I've simplified the fetchCoordinates task to take just the address parameter instead of the address plus hardcoded field name strings

packages/catalog-realm/fields/geo-search-point.gts
lucaslyl: Done rename

tintinthong:
If its one single point it should be Coordinate (singular) -- not CoordinateS

tintinthong (approved):
There are a few things I want to change. But this is a good starting point.

lucaslyl:
> There are a few things I want to change. But this is a good starting point.

Don't worry! we can continue to make things better together and upgrade them in future

PR #3128, opened 2025-08-14
Refactor API to not use NotLoaded errors to load missing fields

habdelra:
When in the /render route we lazily load the missing fields instead of throwing NotLoaded errors when the link has not yet been loaded.


habdelra:
cosing in favor of https://github.com/cardstack/boxel/pull/3131

PR #3130, opened 2025-08-15
host: Add Connect button for host mode

backspace:
This adds a Connect button in host mode:

fakepublished 2025-09-02 11-44-10

If the user has a session, that shows:

fakepublished 2025-09-02 11-44-47

**However**, this doesn’t implement the transition between having no session and becoming logged in.


packages/host/app/templates/card.gts
```diff
@@ -44,6 +46,10 @@ class HostModeComponent extends Component<HostModeComponentSignature> {
 
   @service private declare store: StoreService;
 
+  get connectUrl() {
+    return 'http://localhost:4200/connect';
```
lukemelia: should this be based on config?

lukemelia (approved):

packages/host/app/templates/card.gts
backspace: oops, yes, definitely, thanks

PR #3135, opened 2025-08-18
host: Add cross-port CORS development middleware

backspace:
CORS errors in Matrix test logs are confusing red herrings, and maybe they actually cause problems:

Playwright Trace Viewer 2025-08-18 13-15-14

With this in-repo addon, they no longer happen:

Playwright Trace Viewer 2025-08-18 13-19-26

This is a development-only change, as the realm server that serves `index.html` is responsible for setting these headers in deployed environments.

### DIFF
--- a/packages/host/lib/development-cors-middleware/package.json
+++ b/packages/host/lib/development-cors-middleware/package.json
@@ -1,0 +1,13 @@
+{
+  "name": "development-cors-middleware",
+  "ember-addon": {
+    "before": "broccoli-serve-files"
+  },
+  "keywords": [
+    "ember-addon"
+  ],
+  "dependencies": {
+    "ember-cli-babel": "*",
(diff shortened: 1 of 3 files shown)


packages/host/lib/development-cors-middleware/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "development-cors-middleware",
+  "ember-addon": {
+    "before": "broccoli-serve-files"
```
backspace, line 4: As with many Ember CLI things, this is scantily-documented, but this adds the middleware before the [internal addon that serves the assets](serve-files-middleware). In older documentation this is called [`serve-files-middleware`](https://blog.201-created.com/secrets-of-the-ember-cli-server-bde80bb546dd).

habdelra (approved):
nice!

PR #3142, opened 2025-08-19
Add ability to preview theming in boxel-ui freestyle

burieberry:
Added a couple of sample theme files.

theming-preview


lukemelia (approved):
Could be nice to have a little "cycle themes" checkbox so that it will continuously rotate through the themes while you're live reloading and tweaking code.

lukemelia (approved):

PR #3144, opened 2025-08-20
Add image search via proxy endpoint

FadhlanR:
https://github.com/user-attachments/assets/0856d349-bcd5-47fe-a859-e10a01c43405


packages/experiments-realm/google-image-search.gts
```diff
@@ -0,0 +1,504 @@
...
+      const resource = this.searchResource;
+      if (resource?.isSuccess && resource.value?.images) {
+        return resource.value.images;
+      }
+      return [];
+    }
+
+    get searchInfo() {
+      const resource = this.searchResource;
+      console.log('resource', resource);
```
backspace:
```suggestion
```

packages/experiments-realm/google-image-search.gts
```diff
@@ -0,0 +1,504 @@
...
+      <div class='google-image-search'>
+        <div class='search-form'>
+          <FieldContainer @label='Search Query'>
+            <input
+              type='text'
+              class='search-input'
+              placeholder='Enter your search query...'
+              value={{@model.searchQuery}}
+              {{on 'input' this.setSearchQuery}}
+              {{on 'keypress' this.handleKeyPress}}
```
backspace: I suggest `submitOnEnter` or something like that, as it describes what’s done with the keypress vs just that it’s being handled, then when I see the handler attached here I don’t need to look at the code to know what it does.

packages/host/app/commands/search-google-images.ts
```diff
@@ -0,0 +1,105 @@
...
+    const { SearchGoogleImagesResult } = commandModule;
+
+    try {
+      const maxResults = input.maxResults || 10;
+      const startIndex = input.startIndex || 1;
+      const searchQuery = encodeURIComponent(input.query);
+
+      // Build the Google Custom Search API URL
+      // The API key and search engine ID will be handled by the proxy endpoint
+      const searchUrl = `https://www.googleapis.com/customsearch/v1?q=${searchQuery}&searchType=image&num=${Math.min(maxResults, 10)}&start=${startIndex}&cx=064501294a5c9430a`;
```
backspace, line 34: It seems impossible to not expose this custom search engine id but is it protected against abuse? I haven’t been following this proxy work or any infrastructure attached to it.

packages/realm-server/tests/request-forward-test.ts
```diff
@@ -56,17 +56,24 @@ module(basename(__filename), function () {
...
+        {
+          url: 'https://www.googleapis.com/customsearch/v1',
+          apiKey: 'google-api-key',
+          creditStrategy: 'no-credit',
+          supportsStreaming: false,
+          authMethod: 'url-parameter',
+          authParameterName: 'key',
+        },
       ]);
+      console.log('testConfig', testConfig);
```
backspace:
```suggestion
```

backspace (approved):

packages/host/app/commands/search-google-images.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      const sendRequestViaProxyCommand = new SendRequestViaProxyCommand(
+        this.commandService.commandContext,
```
lukemelia:
```suggestion
        this.commandContext,
```

And remove the `@service commandService` line. This means that this command doesn't have to extend from host command

packages/host/app/commands/search-google-images.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  static actionVerb = 'Search Google Images';
+  description = 'Search for images on Google using the Custom Search API';
+
+  async getInputType() {
+    let commandModule = await this.loadCommandModule();
+    const { SearchGoogleImagesInput } = commandModule;
+    return SearchGoogleImagesInput;
+  }
+
```
lukemelia, line 20: Consider adding `requireInputFields = ['query'];`

lukemelia (approved):

packages/host/app/commands/search-google-images.ts
FadhlanR: I asked Gemini, and it said that the Custom Search Engine ID is not private, so I think it’s fine to expose it as long as we don’t expose the API key.

PR #3146, opened 2025-08-20
Update pnpm version to 10.15.0

lukemelia:


backspace (approved):
I’ll be trying out 10.12’s [global virtual store](https://github.com/pnpm/pnpm/releases/tag/v10.12.1) feature, not that I find installation to be slow ever!

PR #3148, opened 2025-08-20
Theme boxel-ui Button component

burieberry:
Updates:
- Kept pre-existing button kind and size variants but added some new ones
- Some changes to `extra-small` size styling and `disabled` button styling based on this design: https://app.zeplin.io/project/687e6476950590c5570a1cec/screen/687e64cc6e540965d28ae846

New variants: 
- `@size='auto'` : useful when you don't want min-width or min-height or padding set on boxel button component 
- `@kind='secondary'`: similar to `secondary-light` and `secondary-dark` but has a background-color instead of transparent background (uses theme `secondary` background-color and `secondary-foreground` color)
- `@kind='outline'`: uses main `background` and `foreground` colors, displays `accent` colors on hover

We use `@kind='text-only'` instead of `ghost` and `@kind='danger'` instead of the `destructive` variant. Interested to hear opinions on if I should add those names along with keeping the existing ones (for backwards-compatibility reasons).

Preview link: https://cs-9271-button-theming.boxel-ui-preview.stack.cards/?s=Components&ss=%3CButton%3E

### DIFF
--- a/packages/boxel-ui/addon/src/components/button/index.gts
+++ b/packages/boxel-ui/addon/src/components/button/index.gts
@@ -8,18 +8,41 @@ import LoadingIndicator from '../loading-indicator/index.gts';
 
 export type BoxelButtonKind =
   | 'primary'
+  | 'secondary'
+  | 'outline'
   | 'secondary-dark'
   | 'secondary-light'
   | 'danger'
   | 'primary-dark'
   | 'text-only';
 
+export const buttonKindOptions: BoxelButtonKind[] = [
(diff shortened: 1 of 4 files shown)


backspace (approved):
> Interested to hear opinions on if I should add those names along with keeping the existing ones (for backwards-compatibility reasons).

I feel like until we’re at 1.0-sort of situation we don’t need to preserve backward compatibility but I’m not strongly attached to that!

lukemelia (approved):

PR #3149, opened 2025-08-21
Incorporate CardContainer in pre-rendering

lukemelia:
Previous to this PR, our indexing process' prerendering step did not capture the card container element typically rendered around an isolated, embedded or fitted card HTML. This was annoying, because it meant you had to treat rendering of PrerenderedCardSearch results differently than when you render the same cards dynamically. We lived with the annoyance, but recently we started relying on the card-container element to set theme CSS variables. In order for theming to work with prerendering, something had to change.

With this change, we now include the card container element when prerendering. It adds the ability for the components yielded by PrerenderedCardSearch to support `...attributes`, too.

It also adds a `hide-boundaries` CSS class to CardContainer, which suppresses box shadows. (Prerendering captures HTML/CSS with boundaries visible and sometimes consumers need to render without them.)

The rest of the PR removes the extra CardContainers we were explicitly rendering when using PrerenderedCardSearch and updates tests accordingly.

Lastly, we replaced the static `styleConversions` export with a new `getStyleConversions()` function, which provides stable output for tests improving test reliability for style-related assertions.

After deployment, we will need to perform a full re-index of all realms.

### DIFF
--- a/packages/boxel-ui/addon/src/components/card-container/index.gts
+++ b/packages/boxel-ui/addon/src/components/card-container/index.gts
@@ -17,42 +17,48 @@ interface Signature {
 const CardContainer: TemplateOnlyComponent<Signature> = <template>
   {{#let (element @tag) as |Tag|}}
     <Tag
-      class={{cn 'boxel-card-container' boundaries=@displayBoundaries}}
+      class={{cn
+        'boxel-card-container'
+        boxel-card-container--boundaries=@displayBoundaries
+      }}
       data-test-boxel-card-container
       ...attributes
     >
       {{yield}}
     </Tag>
   {{/let}}
+
+  {{! Note: styles for this component use :global to avoid issues with
+      cached HTML if this component changes. This is important because it
+      ends up in nearly every card's prerendered HTML
+  }}
   <style scoped>
-    @layer {
-      .boxel-card-container {
-        position: relative;
-        background-color: var(--background, var(--boxel-light));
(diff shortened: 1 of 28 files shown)


packages/boxel-ui/addon/src/components/card-container/index.gts
```diff
@@ -45,6 +45,9 @@ const CardContainer: TemplateOnlyComponent<Signature> = <template>
           0 0 0 1px var(--border, var(--boxel-border-color)),
           var(--shadow, 0 0 0 1px var(--boxel-border-color));
       }
+      .boundaries.hide-boundaries {
+        box-shadow: none;
+      }
```
burieberry: if we want to keep a theme's box-shadow but not have the card-boundary, this line can be `box-shadow: var(--shadow);` or `box-shadow: var(--shadow, none);`

burieberry (commented):
Overall good changes, but the cards using the default fitted template don't seem to have the `CardContainer` integrated. Only the cards with custom fitted template have it.

missing-card-container

packages/boxel-ui/addon/src/components/card-container/index.gts
lukemelia: Hmm, would we consider any shadow or border on card-container to be the "boundary" and so we should hide it when so directed?

burieberry: Sure, that works for me

PR #3139, opened 2025-08-19
Summarize Current Session choice

FadhlanR:
In this PR, I implemented the third option for the new session settings. I added a new command, `SummarizeSessionCommand`, which retrieves the summary of the currently open room by calling `SendRequestViaProxyCommand`, which then forwards the request to the OpenRouter endpoint. I also need to extract helpers from ai-bot package so it can be reused in the host.

https://github.com/user-attachments/assets/4689a1b1-561f-4d6f-9b88-41cadb3320cf

PR #3151, opened 2025-08-21
AI bot: do not crash ai-bot if non-text attachments are encountered

jurgenwerk:
Gracefully ignores unsupported files without causing error noise

### DIFF
--- a/packages/host/app/lib/file-def-manager.ts
+++ b/packages/host/app/lib/file-def-manager.ts
@@ -315,7 +315,11 @@ export default class FileDefManagerImpl
           throw new Error('File needs a realm server source URL to upload');
         }
 
-        let response = await this.network.authedFetch(file.sourceUrl);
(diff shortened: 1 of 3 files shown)


packages/host/app/lib/file-def-manager.ts
```diff
@@ -315,11 +315,7 @@ export default class FileDefManagerImpl
           throw new Error('File needs a realm server source URL to upload');
         }
 
-        let response = await this.network.authedFetch(file.sourceUrl, {
-          headers: {
-            Accept: 'application/vnd.card+source',
```
jurgenwerk, line 320: I anticipate we'll support files other than code source so it makes sense to remove this header IMO

tintinthong (approved):

richardhjtan (approved):

lukemelia (approved):

jurgenwerk:
looks like this was addressed in https://github.com/cardstack/boxel/pull/3139

PR #3152, opened 2025-08-21
Tooltip component with theme variables

richardhjtan:
### What is changing
- Tooltip component to use theme variables
- Patch pill variable name

**Screenshot**

https://github.com/user-attachments/assets/7e95703b-232c-41d1-87ed-096a344a5266

### DIFF
--- a/packages/boxel-ui/addon/src/components/tooltip/index.gts
+++ b/packages/boxel-ui/addon/src/components/tooltip/index.gts
@@ -1,3 +1,4 @@
+import { concat } from '@ember/helper';
 import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 import { getOwner } from '@ember/owner';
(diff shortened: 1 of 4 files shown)


packages/boxel-ui/addon/src/components/tooltip/index.gts
```diff
@@ -19,29 +23,105 @@ interface Signature {
...
+      }
+      // eslint-disable-next-line ember/no-side-effects
+      this.overlayContainer = container;
+      // Do initial sync when overlay is created
+      this.syncCustomProps();
     }
+    return this.overlayContainer!;
+  }
+
+  private syncCustomProps() {
```
richardhjtan, line 59: This part is tricky — we need to pass variables to the overlay programmatically since theme variables aren’t available at that level. The overlay sits one level below `<body>`, while the theme variables may be defined several levels deeper.

Screenshot 2025-08-21 at 6 04 44 PM

backspace (approved):
I tried this out in the preview deployment, looks good!

I don’t think this is related to your PR but the font for StarryNight theme is missing:

Boxel Components 2025-08-21 08-33-36

PR #3153, opened 2025-08-21
fix listing remix flaky test

tintinthong:
we are relying on`getRoomEvents()` which is a server side consideration. 

By doing so, we are waiting for client events to get back to the server but expecting that waiting for `data-room-settled` on a component to be sufficient. 

We shud just be waiting for the message in the ai assistant chat


tintinthong:
There were 2 failed test but it seems unrelated 

Screenshot 2025-08-21 at 22 13 14

backspace:
> There were 2 failed test but it seems unrelated
> 
> Screenshot 2025-08-21 at 22 13 14

yes `failed to fetch` for chunks is always a red herring, IMO 😵

backspace (approved):

lukemelia (approved):

PR #3159, opened 2025-08-22
Select component with theme variables

richardhjtan:
https://github.com/user-attachments/assets/37cdab07-09ff-49b9-94b8-266ed407cf41


packages/boxel-ui/addon/src/components/select/index.gts
```diff
@@ -20,103 +23,608 @@ interface Signature<ItemT = any> {
...
+      }
+
+      .variant-primary {
+        --select-background-color: var(
+          --boxel-select-background-color,
+          var(--primary, var(--boxel-600))
+        );
+        --select-border-color: var(
+          --boxel-select-border-color,
+          var(--primary, var(--boxel-600))
```
burieberry, line 327: fallback should be `--boxel-highlight` for `--primary`

packages/boxel-ui/addon/src/components/select/index.gts
```diff
@@ -20,103 +23,608 @@ interface Signature<ItemT = any> {
...
+        );
+        --select-text-color: var(
+          --boxel-select-text-color,
+          var(--primary-foreground, var(--boxel-light))
```
burieberry, line 331: boxel fallback should be `--boxel-dark` for `--primary-foreground`

packages/boxel-ui/addon/src/components/select/index.gts
```diff
@@ -20,103 +23,608 @@ interface Signature<ItemT = any> {
...
+      }
+
+      .variant-destructive {
+        --select-background-color: var(
+          --boxel-select-background-color,
+          var(--destructive, var(--boxel-600))
+        );
+        --select-border-color: var(
+          --boxel-select-border-color,
+          var(--destructive, var(--boxel-600))
```
burieberry, line 384: fallback is `--boxel-danger` for `--destructive`

packages/boxel-ui/addon/src/components/select/index.gts
```diff
@@ -20,103 +23,608 @@ interface Signature<ItemT = any> {
...
+        );
+        --select-text-color: var(
+          --boxel-select-text-color,
+          var(--destructive-foreground, var(--boxel-light))
```
burieberry, line 388: fallback is `--boxel-light` for `--destructive-foreground`

burieberry (changes requested):
Thanks for working on this tricky one. A couple of things:

1- Make sure the fallbacks for the `default` are matching the current Select component, including the placeholder color and the disabled state, or even better, if they are matching gardella's design. Looks like we're missing the checkmark icon for the selected item and the trigger border should turn black on hover and active. 

Here's the existing:
existing

Here's the new one, which is not matching:
new

Gardella design:
dropdown-design

2- Bonus fix: It would be great if the usage doc didn't display [object Object]. Maybe convert it to text, or create a ticket for it to be handled at a later time.

richardhjtan:
Thanks for the feedback @burieberry , addressed the feedback and followed Gardella's design as possible
Screenshot 2025-08-25 at 4 22 54 PM

burieberry (commented):
The `default` variant's default styles look good now. However I noticed that the  keyboard accessibility of this component is broken and seems to be related to our styling. Ember-power-select by itself (https://ember-power-select.com/docs/how-to-use-it) has the accessibility features.

packages/boxel-ui/addon/src/components/select/trigger.gts
```diff
@@ -57,6 +57,8 @@ export class BoxelTriggerWrapper extends Component<TriggerSignature> {
         width: 100%;
         gap: var(--boxel-sp-xxxs);
         padding: var(--boxel-sp-xs);
+        font: var(--boxel-font-sm);
```
burieberry, line 60: whenever this shorthand `font` property is used, need to add `font-family: inherit` after it

richardhjtan: Ah missed that out, added!

richardhjtan:
Fixed the accessibility which should be similar to `ember-power-select` now. Have to clear the cache when seeing preview https://cs-9307-select-with-theme-variables.boxel-ui-preview.stack.cards/?s=Components&ss=%3CSelect%3E

burieberry (approved):

PR #3164, opened 2025-08-24
generate spec readme button

tintinthong:
Creating commands
- generate-readme
- generate-readme-spec

Screenshot 2025-08-24 at 23 52 24

### DIFF
--- a/packages/host/app/commands/generate-readme.ts
+++ b/packages/host/app/commands/generate-readme.ts
@@ -1,7 +1,6 @@
 import { service } from '@ember/service';
 
 import { isCardInstance } from '@cardstack/runtime-common';
-import { skillCardsToMessage } from '@cardstack/runtime-common/ai/prompt';
 
 import type * as BaseCommandModule from 'https://cardstack.com/base/command';
 import { Skill } from 'https://cardstack.com/base/skill';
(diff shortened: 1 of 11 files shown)


tintinthong:
@IanCal I think this PR is blocked on usage of commands in base. But wud be good to get your feedback on the usage of the command here (eg sharing of the skill etc). All should be in place generally apart from the type woes

packages/host/app/commands/generate-readme.ts
```diff
@@ -0,0 +1,145 @@
...
+
+import ReadTextFileCommand from './read-text-file';
+import SendRequestViaProxyCommand from './send-request-via-proxy';
+
+import type CommandService from '../services/command-service';
+import type MatrixService from '../services/matrix-service';
+import type RealmServerService from '../services/realm-server';
+import type StoreService from '../services/store';
+
+export default class GenerateReadmeCommand extends HostBaseCommand<
```
lukemelia: Perhaps this command should be more generic, such as "OneShotLlmRequestCommand". It could have an "attachedFilePaths" input field to accommodate the ReadTextFileCommand logic. Nothing else about it is really relevant to generating a README specifically.

lukemelia (approved):

lukemelia (approved):

tintinthong:
I have made some refinement to the prompt based upon more testing

PR #3166, opened 2025-08-25
Fix: Schema editor not refreshing after applying a code patch

FadhlanR:
This PR fixes an issue where applying a code patch does not refresh the schema editor because the refresh mechanism relies on whether the loader is reset. The same issue also exists in restore content.


lukemelia (approved):
Good job tracking this down. We can merge but please make a note to review with @habdelra when he returns from vacation.

PR #3167, opened 2025-08-25
GCCardStore should use authed fetch

tintinthong:
This came about when I noticed that spec in the RHS panel was missing inspite of having search return network request for specs

Screenshot 2025-08-25 at 18 40 18

Callstack will hit `loadCardDocument` which is from the GCCardStore (when loadMissingField)

Screenshot 2025-08-25 at 19 49 14

I saw this as part of the `getCard` resource which uses the store to fetch. Particularly, in the spec panel where the spec was not loading

 But the fetch instantiated is not the authedFetch from the network service and I get authorization errors when loading missing fields from card-api
 
 The most recent PR to touch this https://github.com/cardstack/boxel/pull/3118

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -105,7 +105,7 @@ export default class StoreService extends Service implements StoreInterface {
 
   constructor(owner: Owner) {
     super(owner);
-    this.store = new CardStore(this.referenceCount, this.network.fetch);
+    this.store = new CardStore(this.referenceCount, this.network.authedFetch);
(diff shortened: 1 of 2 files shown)


lukemelia (approved):
This seems good to me and OK to merge when green. Please make a note to review with @habdelra when he returns from vacation.

tintinthong:
After alot of trial and error adding the `disable-dbus` flag to testem file in host seems to solve the CI errors. Im not too confident. But today I have seen this 3 times on different PRs

I will merge in first

PR #3168, opened 2025-08-25
Update icon button component for theming

burieberry:
- Use `Button` component and accept button `kind` arguments
- Update fallback styles for `default` variant to remain the same as before, while accepting theming variables

Preview: https://cs-9310-icon-button.boxel-ui-preview.stack.cards/?s=Components&ss=%3CIconButton%3E

API:
icon-button

Themed example:
themed

### DIFF
--- a/packages/boxel-ui/addon/src/components/icon-button/index.gts
+++ b/packages/boxel-ui/addon/src/components/icon-button/index.gts
@@ -1,10 +1,9 @@
-import { on } from '@ember/modifier';
-import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
 
 import cn from '../../helpers/cn.ts';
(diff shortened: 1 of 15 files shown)


lukemelia (approved):

packages/boxel-ui/addon/src/components/icon-button/index.gts
```diff
@@ -1,102 +1,110 @@
...
+        height: var(--boxel-icon-button-height, var(--boxel-icon-lg));
         display: inline-flex;
         align-items: center;
         justify-content: center;
-        padding: 0;
-        background: var(--boxel-icon-button-background, none);
-        border: 1px solid transparent;
-        color: var(--boxel-icon-button-color, currentColor);
+        padding: var(--boxel-icon-button-padding, 0);
+        background: var(
```
richardhjtan: Perhaps in default we can use `--background` as second fallback

richardhjtan (approved):

PR #3169, opened 2025-08-26
Move ALLOWED_PROXY_ENDPOINTS to db

FadhlanR:


lukemelia (changes requested):
I would prefer to see a dedicated `proxy_endpoints` table. That way, the addition of a new one is a migration that just adds a row to that table, and won't have to fuss with JSON values in a generically named table.

packages/postgres/migrations/1756175337456_create-proxy-endpoints.js
```diff
@@ -0,0 +1,23 @@
+exports.up = (pgm) => {
+  pgm.createTable('proxy_endpoints', {
+    id: { type: 'uuid', primaryKey: true, notNull: true },
+    url: { type: 'varchar', notNull: true, unique: true },
+    api_key: { type: 'varchar', notNull: true },
```
lukemelia: Maybe this should be nullable?

packages/postgres/migrations/1756175337456_create-proxy-endpoints.js
```diff
@@ -0,0 +1,23 @@
...
+    updated_at: {
+      type: 'timestamp',
+      notNull: true,
+    },
+  });
+};
+
+exports.down = (pgm) => {
+  pgm.dropTable('proxy_endpoints');
+};
```
lukemelia, line 23: Should we also have a migration to add the initial row for the open router proxy?

lukemelia (approved):

PR #3172, opened 2025-08-26
move viewCard into command

tintinthong:
- added stackInfo to commandContext


lukemelia (commented):
I know this is what we discussed, but it sure seems like a worse developer experience for card authors. Also it seems wrong that we would have a ShowCard command and a ViewCard command. Let's discuss.

lukemelia:
Closing in favor of #3210

PR #3173, opened 2025-08-26
don't show focus pill for line number when no text is selected

lukemelia:
- Only show line range when text is actually selected, not just cursor positioning
- Add unit tests for all selection scenarios

### DIFF
--- a/packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
+++ b/packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
@@ -1,0 +1,124 @@
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+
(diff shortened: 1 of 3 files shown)


packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
```diff
@@ -0,0 +1,124 @@
...
+  setupTest(hooks);
+
+  test('focusPillCodeRange returns undefined when there is no selection', function (assert) {
+    let service = this.owner.lookup(
+      'service:ai-assistant-panel-service',
+    ) as AiAssistantPanelService;
+
+    let monacoService = this.owner.lookup(
+      'service:monaco-service',
+    ) as MonacoService;
```
backspace: I think if you use `getService` as seen in other test files you get the type automatically

backspace (approved):

PR #3182, opened 2025-08-28
Implement ai-app-generator and it's embedded view

FadhlanR:
https://github.com/user-attachments/assets/fdf73147-eee9-4863-944c-63801917b61a


tintinthong (approved):
is this supposed to be in act mode after creation?

FadhlanR:
> is this supposed to be in act mode after creation?

I don't see it in the requirement but it makes sense. I'll update it.

PR #3191, opened 2025-09-01
turn card-type and module-contents business logic into service so it can be used in a command

tintinthong:
I want to make create-spec into command https://github.com/cardstack/boxel/pull/3193 but the usage of resources get in the way of me wanting to compute the data from module contents

2 affected abstractions
- cardType is a resource that loads the entire inheritance hierarchy of a card and its put into a resource
- moduleContents is a resource meant to be a consolidator of syntactical information + card type information to derive declarations that we are interested in (essentially imports the module and loads the types using cardType)

I need moduleContents to be a service in order to create specs. I decided to peel out the business logic of these 2 resources and put them into a service. That way I can load information from them without weirdly having resources inside resources or resources inside commands

**Note:** 

- If there are concerns with caching etc. the service also does the same thing. In fact, inspector is always slow not bcos of caching in place but its bcos of the loader trashing that eliminates the cache when editing a file. 
- this doesn't touch the code semantic service api
- I think module contents is not particularly clear about what it is. Wud like to get feedback first before adding some comments

### DIFF
--- a/packages/host/app/components/operator-mode/card-adoption-chain.gts
+++ b/packages/host/app/components/operator-mode/card-adoption-chain.gts
@@ -14,7 +14,7 @@ import { stripFileExtension } from '@cardstack/host/lib/utils';
 
 import type { Ready } from '@cardstack/host/resources/file';
 
-import type { Type } from '@cardstack/host/services/card-type';
+import type { Type } from '@cardstack/host/services/card-type-service';
 import { isOwnField } from '@cardstack/host/utils/schema-editor';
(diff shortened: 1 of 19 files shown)


richardhjtan (approved):

tintinthong:
This commit may have fixed the flakiness https://github.com/cardstack/boxel/pull/3191/commits/7cbb459fa71108c48563c8e75142ca6486f630a1 that caused us to thread thru resources in the first place

PR #3192, opened 2025-09-01
Index card improvement

FadhlanR:
https://github.com/user-attachments/assets/9d8c04fa-643f-460f-a58b-c978ef8fcf8e


backspace (approved):

packages/matrix/helpers/index.ts
```diff
@@ -24,6 +24,7 @@ interface ProfileAssertions {
 }
 interface LoginOptions {
   url?: string;
+  showAllCards?: boolean; //default true
```
lukemelia, line 27: This seems very specific to a particular type of index card. Are there any other ways to conceptualize and solve the need here?

packages/base/cards-grid.gts
```diff
@@ -272,6 +281,55 @@ class Isolated extends Component<typeof CardsGrid> {
       this.loadFilterList.perform();
     }
   };
+
+  private loadHighlightsCards = restartableTask(async () => {
```
burieberry, line 285: Should the highlights cards shown on the index card be part of a linksToMany field? This way you can link to different cards as highlights from different index cards. Question for @lukemelia @christse

burieberry: Never mind, I see that this Highlights page needs to be the same for every index card

packages/base/join-the-community.gts
```diff
@@ -0,0 +1,213 @@
...
+          padding: var(--boxel-sp);
+          background: white;
+          border-radius: var(--boxel-border-radius);
+          box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
+          transition: box-shadow 0.2s ease;
+        }
+
+        .community-card:hover {
+          box-shadow: 0 2px 8px rgba(0, 0, 0, 0.15);
+        }
```
burieberry: I wonder if the whole card should be clickable instead of just the button inside?

packages/base/join-the-community.gts
```diff
@@ -0,0 +1,213 @@
...
+          >{{@model.platform}}</h4>
+          <p
+            class='community-description'
+            data-test-community-description={{@model.platform}}
+          >{{@model.description}}</p>
+          <button
+            class='community-link'
+            data-test-community-link={{@model.platform}}
+            {{on 'click' this.openLink}}
+          >{{@model.url}}</button>
```
burieberry: If this button is not a visual treatment only, you can use the Button component with `@kind='muted'` and maybe even the `@as='anchor'` for the link href. Then the only css would be to change the text color to blue.

PR #3193, opened 2025-09-02
make create-spec into a command

tintinthong:
make create-spec a command


richardhjtan (approved):

tintinthong:
one thing that might be useful here to prevent spec bloat. Is to check if a spec of the same code ref already exists

tintinthong:
I have added "create if exist" functionality to the spec command. I also return the new specs and the larger set of new specs + specs that already exist so listing-create can just link as needed

PR #3194, opened 2025-09-02
Add support for missing links in prerender route

habdelra:
This PR adds support for handling missing links in our prerender route for linksTo and linksToMany.


packages/host/tests/acceptance/permissioned-realm-test.gts
habdelra, line 1: This module actually has no tests. there is no reason to keep this around anymore.

lukemelia (approved):

PR #3195, opened 2025-09-02
Include format in Focus Pill

lukemelia:
https://github.com/user-attachments/assets/d89a0704-0162-4e13-a814-f5e86094b681

### DIFF
--- a/packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
+++ b/packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
@@ -90,4 +90,93 @@ module('Unit | Service | ai-assistant-panel-service', function (hooks) {
       'returns "Line 1" for single character selection',
     );
   });
+
+  test('focusPillMetaPills returns array combining itemType and codeRange', function (assert) {
+    // Mock selection for codeRange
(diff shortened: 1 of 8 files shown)


packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
```diff
@@ -90,4 +90,93 @@ module('Unit | Service | ai-assistant-panel-service', function (hooks) {
...
+    const result = service.focusPillMetaPills;
+
+    // Build expected array based on what the individual getters return
+    const expected: string[] = [];
+    if (itemType) {
+      expected.push(itemType);
+    }
+    if (codeRange) {
+      expected.push(codeRange);
+    }
```
backspace: Is there really dynamism, where does it come from? The test below also has a conditional but I don’t see why.

backspace (approved):

packages/host/tests/unit/services/ai-assistant-panel-service-test.ts
lukemelia: Cleaned it up, tx

PR #3196, opened 2025-09-02
FocusPill Pill component usage refactor

burieberry:
- Removed a bunch css that seems to be redundant with the current Pill component. 
- Main change in the template is that the meta-pills are now nested inside the main Pill component.


lukemelia (approved):
Nice simplification!

PR #3202, opened 2025-09-04
Introduce /_publish-realm endpoint

FadhlanR:
Path: `/_publish-realm`
Method: POST
Request:
```
{
    "sourceRealmURL": "http://localhost:4201/user/my-public-workspace/"
    "publishedRealmURL": "http://user.localhost:4201/my-public-workspace/" //OPTIONAL
} 

```

Response:
```
{
    "data": {
        "type": "published_realm",
        "id": "d9fd8eec-b95b-4894-beee-81dfc26a3cfd",
        "attributes": {
            "sourceRealmURL": "http://localhost:4201/user/my-public-workspace/",
            "publishedRealmURL": "http://user.localhost:4201/my-public-workspace/",
            "lastPublishedAt": "2025-09-04T13:03:38.585Z"
        }
    }
}

```

## Publishing Realm Process

The `/_publish-realm` endpoint allows users to publish a private realm to make it publicly accessible. The process:

1. **Validates permissions** - Ensures the user has `realm-owner` permission on the source realm
2. **Creates or updates database record** - Either inserts a new published realm entry or updates the `last_published_at` timestamp for existing publications
3. **Registers Matrix user** - Creates a bot user for the published realm (only for new publications)
4. **Copies realm files** - Duplicates the source realm's file system to a published directory
5. **Creates and mounts realm** - Instantiates a new Realm instance with the copied files and mounts it to the virtual network
6. **Copies search index** - Runs a background `copy-index` job to copy the source realm's search index to the published realm
7. **Returns published realm info** - Provides the published realm URL, ID, and timestamp


packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -0,0 +1,270 @@
...
+      if (existingPublishedRealm) {
+        virtualNetwork.unmount(existingPublishedRealm.handle);
+      }
+      let realm = createAndMountRealm(
+        publishedRealmPath,
+        publishedRealmURL,
+        realmUsername,
+        new URL(sourceRealmURL),
+      );
+      await realm.start();
```
lukemelia, line 218: This is process-specific, and we have multiple processes running (workers, worker manager, potentially multiple realm server instances). Seems problematic.

packages/runtime-common/index-writer.ts
```diff
@@ -763,19 +764,24 @@ export class Batch {
...
+              module: this.copiedRealmURL(
+                fromRealm,
+                new URL(fieldDefinition.fieldOrCard.module),
+              ).href,
+              name: fieldDefinition.fieldOrCard.name,
+            },
           },
-        },
-      ]),
+        ]),
```
habdelra: This is fixed in PR https://github.com/cardstack/boxel/pull/3206

packages/realm-server/handlers/handle-publish-realm.ts
FadhlanR: Do you mean that if an existing published realm is republished, we don’t need to re-create and re-start the realm?

FadhlanR: Ah, I see what you mean by process-specific. But I also don’t think we have any special handling for that in the ‘create realm’ endpoint. CMIIW.

backspace:
I was able to use this after deploying this branch merged with #3206, now I have a properly-published realm. Something I’m wondering about is validation of `publishedRealmURL`; do we really just allow anything to be put there? Only certain (sub)domains will work with the current DNS setup, if someone submits something that won’t work, is it fine, just won’t work?

packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -0,0 +1,270 @@
...
+          password: await passwordFromSeed(realmUsername, realmSecretSeed),
+          registrationSecret: await getMatrixRegistrationSecret(),
+        });
+        userId = newUserId;
+
+        await insertPermissions(dbAdapter, new URL(publishedRealmURL), {
+          [userId]: DEFAULT_PERMISSIONS,
+          [ownerUserId]: DEFAULT_PERMISSIONS,
+          '*': ['read'],
+        });
```
backspace, line 191: I’m a bit confused with the `DEFAULT_PERMISSIONS` abstraction here but when I used the endpoint I got these for the newly-published realm:

realm_user_permissions — boxel — port forwarding boxel realm server staging 2025-09-05 13-15-41

But I think Luke mentioned that the owner shouldn’t have write permissions because the realm should be immutable?

habdelra:
@FadhlanR for the conflict, just accept my new updates and that will fix the copied code ref stuff

packages/realm-server/handlers/handle-publish-realm.ts
FadhlanR: You're right, I was just copying this part from create realm endpoint. I'll update it.

packages/realm-server/handlers/handle-publish-realm.ts
FadhlanR: This will be solved in this ticket https://linear.app/cardstack/issue/682dbe41-530a-4316-a4f9-18cf00dc0777.

packages/realm-server/handlers/handle-publish-realm.ts
FadhlanR: Updated.

FadhlanR:
> I was able to use this after deploying this branch merged with #3206, now I have a properly-published realm. Something I’m wondering about is validation of `publishedRealmURL`; do we really just allow anything to be put there? Only certain (sub)domains will work with the current DNS setup, if someone submits something that won’t work, is it fine, just won’t work?

I updated to ensure publishedRealmURL using the valid domains.

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -1501,6 +1504,340 @@ module(basename(__filename), function () {
...
+                realmSecretSeed,
+              )}`,
+            )
+            .send(
+              JSON.stringify({
+                sourceRealmURL: testRealm2.url,
+              }),
+            );
+
+          assert.strictEqual(response.status, 400, 'HTTP 400 status');
```
backspace, line 1668: I’m deploying and will review more thoroughly after that but I think this should be 422, 400 is more like bad syntax, where this is correct JSON but an invalid value.

packages/host/app/services/host-mode-service.ts
```diff
@@ -24,7 +24,10 @@ export default class HostModeService extends Service {
   }
 
   get isRealmServerDomain() {
-    if (this.simulatingHostMode) {
+    if (
+      this.simulatingHostMode ||
+      (config.environment !== 'production' && config.environment !== 'staging')
```
backspace: it’s confusing but `staging` is never used that I know of, even in staging the value is `production`:

Boxel 2025-09-08 12-57-53

since this is overwritten in tests, maybe it could just be `config.environment === 'development'`

but I’m not sure I understand what this change is for…?

packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -0,0 +1,256 @@
...
+      ? json.sourceRealmURL
+      : `${json.sourceRealmURL}/`;
+    let publishedRealmURL = json.publishedRealmURL.endsWith('/')
+      ? json.publishedRealmURL
+      : `${json.publishedRealmURL}/`;
+
+    // Validate publishedRealmURL domain
+    try {
+      let publishedURL = new URL(publishedRealmURL);
+      if (validPublishedRealmDomains && validPublishedRealmDomains.length > 0) {
```
backspace, line 86: I added [this commit](https://github.com/cardstack/boxel/commit/1d3180e59461a3565384fc7990f9a54a8a33e1a8) to try this out in staging, as without it anything was still accepted. Does it make sense to accept anything if `VALID_PUBLISHED_REALM_DOMAINS` isn’t configured? Should this PR include this configuration for staging and production?

lukemelia (approved):

packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    if (!json.sourceRealmURL) {
+      await sendResponseForBadRequest(ctxt, 'sourceRealmURL is required');
+      return;
+    }
+
+    if (!json.publishedRealmURL) {
+      await sendResponseForBadRequest(ctxt, 'publishedRealmURL is required');
+      return;
+    }
```
backspace, line 74: hmm well I commented elsewhere that 400 was incorrect for this but now I see that this helper is used in many places that I also think 400 is incorrect so… I guess it’s fine? 😆

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -1501,6 +1504,340 @@ module(basename(__filename), function () {
...
+          );
+          assert.ok(
+            response.body.data.attributes.publishedRealmURL,
+            'published realm URL is present',
+          );
+          assert.ok(
+            response.body.data.attributes.lastPublishedAt,
+            'last published at timestamp is present',
+          );
+        });
```
backspace, line 1558: Is an assertion needed that the correct directory within `_published` was created, or is that covered elsewhere?

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -1501,6 +1504,340 @@ module(basename(__filename), function () {
...
+              'Authorization',
+              `Bearer ${createRealmServerJWT(
+                { user: ownerUserId, sessionRoom: 'session-room-test' },
+                realmSecretSeed,
+              )}`,
+            )
+            .send(
+              JSON.stringify({
+                sourceRealmURL: testRealm2.url,
+                publishedRealmURL: 'http://subdomain.localhost/test-realm/',
```
backspace: Is this really different than the `testuser.localhost` test above? It seems like in the absence of testing the `,`-parsing, this could just be removed.

backspace (approved):
I tried this out in staging (with a commit to add a domain for validation) and it worked as expected 🎉

realm-server 2025-09-08 14-03-49

packages/host/app/services/host-mode-service.ts
FadhlanR: Ah, I didn’t know that staging never uses the environment. It would work with config.environment === 'development'. This change allows host mode to be accessed from the same domain as operator mode. We need this validation in staging and production because host mode uses a different domain there. On localhost, however, host mode and operator mode share the same domain.

FadhlanR: I can revert this check, since there is `!currentHost.endsWith('.localhost')` check after I merge to main.

packages/realm-server/handlers/handle-publish-realm.ts
FadhlanR: I think so

PR #3203, opened 2025-09-04
Dropdown theming

richardhjtan:
https://github.com/user-attachments/assets/9a008194-83d0-4cd6-9e6b-2cb286091a2b


lukemelia (approved):

burieberry (changes requested):
Your changes look good in the boxel-ui package, however when I tried to run it things seems to be broken on host app. Percy is not very good at capturing changes with wormhole elements and overlays, so I recommend doing a search for BoxelDropdown usage in the codebase to ensure things remain mostly the same. A couple of things I could catch:

Past session menu is doubled:
Screenshot 2025-09-04 at 11 13 28 AM

Maybe needs a `document?.`. But also menus were all open.
Screenshot 2025-09-04 at 11 16 15 AM

Broken styling:
Screenshot 2025-09-04 at 11 18 18 AM

Tooltip styles are confused & trigger hover looks strange:
Screenshot 2025-09-04 at 11 24 32 AM

richardhjtan:
Thanks for checking @burieberry , it should be okay now

https://github.com/user-attachments/assets/8a75316a-1649-4e68-9596-2e6711f6bf8c

burieberry (approved):
Looks much better. The remaining items seem to be the extra border on the submode-switcher and the hover color on the un-themed dropdown menu items seems different. It doesn't look like the themed menu items inherit the theme font.

richardhjtan:
> Looks much better. The remaining items seem to be the extra border on the submode-switcher and the hover color on the un-themed dropdown menu items seems different. It doesn't look like the themed menu items inherit the theme font.

I took a look but I can't really reproduce them in my local, perhaps it's related to the cache. I will merge it and do another check in environment.

PR #3204, opened 2025-09-04
Alert component theming

richardhjtan:
https://github.com/user-attachments/assets/085e08e2-3522-43cb-bd56-aaee6be7443c


lukemelia (approved):
Do we need to consider contrast with the red and yellow warning icons?

burieberry (commented):
Maybe we need some design guidance here. I'm not sure what the themed version is expected to look like. You can also change `--icon-background-color` and `--icon-color`. For the way you have it, error icon background would be `--destructive-foreground` and fill would be `--destructive`. 

If we're following our own designs, then the closest is for error container to use `var(--foreground)` as background and `var(--background)` as foreground, since it's in dark mode always... Or maybe `--destructive-foreground` for the background and `--destructive` for the foreground.

jurgenwerk (approved):

PR #3210, opened 2025-09-04
[BREAKING] Move viewCard, createCard, editCard, saveCard funcs

lukemelia:
- Updated the `publicAPI` method in `InteractSubmode` to remove the actions object and directly use the new methods.
- Modified `Overlays` to use the new `viewCard` function from the arguments.
- Updated `OperatorModeStackItem` to utilize the new card CRUD functions context.
- Added a new dynamic context `CardCrudFunctionsContextName` for better organization of card-related functions.

This will break cards that use this.args.context.actions.[viewCard|editCard|saveCard|createCard]


lukemelia:
@IanCal we shouldn't still have those skills in base, should we?

packages/catalog-realm/catalog-app/resources/helpers/listing-action-resolver.gts
```diff
@@ -100,12 +101,25 @@ export class BaseListingAdapter {
...
   async preview(listing: Listing): Promise<void> {
     if (!listing.examples?.length) throw new Error('No examples available');
-    this.context.actions?.viewCard?.(listing.examples[0]);
+    const commandContext = this.context.commandContext;
+    if (!commandContext) {
+      return;
+    }
+    await new ShowCardCommand(commandContext).execute({
+      cardId: listing.examples[0].id,
+    });
```
tintinthong, line 110: The behaviour here has changed. Now when I preview an example or view a listing. A second stack is being popped open.

tintinthong:
Are we not putting the `CardCrudFunctionsContextName` provide higher up so that code mode can consume it too? Or is that a separate concern of another PR.

tintinthong:
CRM App seems to be broken so I wasn't able to confirm if the actions work there but others look ok  https://linear.app/cardstack/issue/CS-9366/investigate-why-crm-app-createcard-is-not-working. 

EDIT: this create in CRM not working seems independent of this PR too

tintinthong (approved):

lukemelia:
> Are we not putting the CardCrudFunctionsContextName provide higher up so that code mode can consume it too? Or is that a separate concern of another PR.

We can but will need a different implementation as these use stackIndex.

PR #3213, opened 2025-09-08
Fix serialization for nested polymorphic plural fields

tintinthong:
This issue occurs when we have a card > field in card > plural field in field. 

The behaviour that can be observed is `fileSerialization` will break ie editing a field will not serialize the json properly upon save based upon the subtype. 

However, its not easy to test or see at this point bcos contains-many editor is not rendering these nested fields -- probably needs discussion

### DIFF
--- a/packages/host/tests/test-helper.js
+++ b/packages/host/tests/test-helper.js
@@ -8,6 +8,8 @@ import start from 'ember-exam/test-support/start';
 import { useTestWaiters } from '@cardstack/runtime-common';
 import * as TestWaiters from '@ember/test-waiters';
 
+QUnit.dump.maxDepth = 20;
+
 useTestWaiters(TestWaiters);
 setApplication(Application.create(config.APP));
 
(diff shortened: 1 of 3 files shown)


packages/host/tests/test-helper.js
```diff
@@ -8,6 +8,8 @@ import start from 'ember-exam/test-support/start';
 import { useTestWaiters } from '@cardstack/runtime-common';
 import * as TestWaiters from '@ember/test-waiters';
 
+QUnit.dump.maxDepth = 20;
```
habdelra, line 11: if you set this to 0 it will treat it as always expanding the nested objects infinitely deep.

packages/runtime-common/realm.ts
```diff
@@ -2659,7 +2659,16 @@ export class Realm {
   ): Promise<void> {
     for (const [fieldName, fieldValue] of Object.entries(fields)) {
       const fieldPath = basePath ? `${basePath}.${fieldName}` : fieldName;
-
+      if (isMeta(fieldValue) && fieldValue?.fields) {
```
habdelra, line 2662: 👍

habdelra (approved):

PR #3215, opened 2025-09-08
Cleanup copy card command code

tintinthong:
- This cleans up the command copy-card api that previously was using an additional arg (code ref) that made this api confusing to use 
- this pr also revive some bulk copy functionality that was not working at all. Now at least if you copy from a public realm u have permissions to, the copy works as expected (but I think in the past there were still some quirks about permissions of copying from a private realm which I did not handle)

### DIFF
--- a/packages/base/command.gts
+++ b/packages/base/command.gts
@@ -30,10 +30,15 @@ export class SaveCardInput extends CardDef {
   @field localDir = contains(StringField);
 }
 
-export class CopyCardInput extends CardDef {
+export class CopyCardToRealmInput extends CardDef {
+  @field sourceCard = linksTo(CardDef);
+  @field targetRealm = contains(StringField);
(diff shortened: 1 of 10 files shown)


tintinthong:
Actually @lukemelia I suddenly rememebered that maybe you might want copy card to be separate command into 2. 

copyCard from stack and copyCard which just handles realm arg. Wud u prefer to split to 2?

lukemelia:
> Actually @lukemelia I suddenly rememebered that maybe you might want copy card to be separate command into 2.
> 
> copyCard from stack and copyCard which just handles realm arg. Wud u prefer to split to 2?

Yes, I think it would be cleaner to have CopyCardToRealmCommand and CopyCardToStackCommand, where the latter could delegate to the former.

lukemelia (approved):

PR #3218, opened 2025-09-09
Add auth support to prerender function

habdelra:
This PR adds auth support to the prerender function by generating JWTs for the requested userId and setting that in the headless browser's local storage as part of generating the prerendered result


packages/realm-server/prerender.ts
```diff
@@ -113,7 +151,9 @@ async function captureResult(
       let element = document.querySelector('[data-prerender]') as HTMLElement;
       let status = element.dataset.prerenderStatus as 'ready' | 'error';
       if (status === 'error') {
-        return { status, value: element.innerHTML! };
+        // there is a strange <anonymous> tag that is being appended to the
+        // innerHTML that this strips out
```
lukemelia, line 155: weird

lukemelia (approved):

PR #3219, opened 2025-09-09
Add PatchFields command

lukemelia:
The `PatchFieldsCommand` is a new command that allows updating specific fields of a card instance rather than patching the entire card. This provides more granular control and better performance for targeted updates.


packages/host/app/lib/field-path-parser.ts
```diff
@@ -0,0 +1,514 @@
...
+ * Supports: "address.city", "tags[0]", "authors[0].name", "tags[]" (append)
+ * Similar to Lodash path syntax but with custom append support.
+ *
+ * Examples:
+ * - "firstName" → ["firstName"]
+ * - "address.city" → ["address", "city"]
+ * - "tags[0]" → ["tags", "[0]"]
+ * - "authors[0].name" → ["authors", "[0]", "name"]
+ * - "tags[]" → ["tags", "[]"] (append operation)
+ * - "tags[-1]" → ["tags", "[-1]"] (append operation)
```
backspace, line 27: I don’t understand how these mean “append”, can you share examples?

lukemelia: It is conventional from rails and form libraries. I showed GPT-5 each of the following code snippets and asked what it meant and it understood both as appending "ok" to the tags array:

```
updateFields(instance, {
  "tags[]": "ok",
});
```

```
updateFields(instance, {
  "tags[-1]": "ok",
});
```

backspace: 👍🏻 thanks, I didn’t understand what was being appended, but I see now

backspace (approved):

PR #3221, opened 2025-09-09
Add already available pagination results to the search

tintinthong:
The meta object in the collection doc shud abide by json api


backspace (approved):
Could this change be verified in any tests?

tintinthong:
Ok I think there shud be enough coverage
- I have added pagination test for the components prerendered search and search resource 
- there are existing test for pagination that now also check for the return meta.page.total

PR #3224, opened 2025-09-10
Include lastPublishedAt to realm info

FadhlanR:
This PR includes a new property to realm info called lastPublishedAt. The data type of that field depends on whether the realm is a source or published realm. If it is a published realm, the type is a timestamp string. If it is a source realm, the type is a map from published realm URL to the timestamp string.


packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -1505,6 +1505,19 @@ module(basename(__filename), function () {
           });
         });
 
+        test('GET /_realm-info returns lastPublishedAt as null for unpublished realm', async function (assert) {
```
lukemelia, line 1508:
```suggestion
        test('GET /_info returns lastPublishedAt as null for unpublished realm', async function (assert) {
```

packages/runtime-common/realm.ts
```diff
@@ -2566,6 +2569,66 @@ export class Realm {
...
+          publishedVersions.map((p) => [
+            p.published_realm_url,
+            p.last_published_at,
+          ]),
+        );
+      }
+
+      return null; // Never published
+    } catch (error) {
+      this.#log.warn(`Failed to get lastPublishedAt: ${error}`);
```
lukemelia, line 2595: Seems like this would be a good error to surface to sentry

packages/runtime-common/realm.ts
```diff
@@ -2566,6 +2569,66 @@ export class Realm {
...
+  } | null> {
+    try {
+      let results = (await query(this.#dbAdapter, [
+        `SELECT last_published_at FROM published_realms WHERE published_realm_url =`,
+        param(this.url),
+      ])) as { last_published_at: string }[];
+
+      return results.length > 0 ? results[0] : null;
+    } catch (error) {
+      this.#log.warn(`Failed to query published realm: ${error}`);
```
lukemelia, line 2611: Here as well -- let's send to sentry

packages/runtime-common/realm.ts
```diff
@@ -2566,6 +2569,66 @@ export class Realm {
...
+  > {
+    try {
+      let results = (await query(this.#dbAdapter, [
+        `SELECT published_realm_url, last_published_at FROM published_realms WHERE source_realm_url =`,
+        param(this.url),
+      ])) as { published_realm_url: string; last_published_at: string }[];
+
+      return results;
+    } catch (error) {
+      this.#log.warn(`Failed to query source realm publications: ${error}`);
```
lukemelia, line 2627: ditto: sentry

lukemelia (approved):

PR #3226, opened 2025-09-10
Realm server: full reindex when boxel-ui changes

jurgenwerk:
This PR addreses the problem where cards' prerendered html will get outdated when boxel-ui package changes (for cards that depend on boxel-ui). Here is a more detailed description of the problem:

> 1) A card's format uses a boxel-ui component
> 2) Indexing prerender captures HTML with then-current data-scoped-… attribute
> 3) We update the boxel-ui component's css and deploy a new release of host
> 4) Rendering prerendered HTML outputs HTML with now-outdated data attribute, so the boxel-ui component CSS is no longer applied

To fix this, we decided to reindex all cards when we detect the boxel-ui package changed. There was some earlier work already done related to this, in this PR: [Write boxel-ui-checksum to the host dist when building the host app](https://github.com/cardstack/boxel/pull/3220). We leverage that in this PR.

Below you can see the schema of how this mechanism is wired together:

image


.github/workflows/manual-deploy.yml
```diff
@@ -159,4 +159,32 @@ jobs:
       cluster: ${{ inputs.environment }}
       service-name: "boxel-realm-server-${{ inputs.environment }}"
       image: ${{ needs.build-realm-server.outputs.image }}
-      wait-for-service-stability: false
+      wait-for-service-stability: true
```
jurgenwerk, line 162: Assures that all realm-server tasks pass their health checks (ECS task health + target group health), and that new tasks are running and old tasks are terminated. I think that's a pretty good assurance that the running addressable tasks have the newest deployed code running

packages/realm-server/handlers/handle-post-deployment.ts
```diff
@@ -0,0 +1,55 @@
...
+      for (let realm of realms) {
+        await reindex({
+          realm,
+          queue,
+          dbAdapter,
+          priority: systemInitiatedPriority,
+        });
+      }
+      writeCurrentBoxelUIChecksum(boxelUiChangeCheckerResult.currentChecksum);
+    }
```
jurgenwerk: We discussed the idea of making and enqueueing a new job for this block of code, but since `reindex` will produce new jobs for reindexing realms perhaps this is fine?

packages/realm-server/lib/boxel-ui-change-checker.ts
```diff
@@ -0,0 +1,53 @@
+import fs from 'fs';
+import path from 'path';
+
+const fileChecksumPath = path.join(
+  process.cwd(),
+  '/../../../persistent/boxel-ui-checksum.txt', // A folder that persists between deployments (AWS EFS)
+);
+
+async function fetchCurrentBoxelUIChecksum(distURL: URL): Promise<string> {
+  const response = await fetch(new URL('/boxel-ui-checksum.txt', distURL));
```
jurgenwerk, line 11: Fetches the boxel-ui-checksum of the host dist that this realm server uses

packages/realm-server/handlers/handle-post-deployment.ts
lukemelia: I'm worried about this request timing out in the time it takes to queue all those reindex jobs. Better to do it in a job IMO

packages/realm-server/lib/boxel-ui-change-checker.ts
```diff
@@ -0,0 +1,53 @@
...
+    throw new Error(
+      `The boxel-ui change checker expects 'persistent' path to exist but it doesn't: ${persistentPath}`,
+    );
+  }
+
+  try {
+    previousChecksum = fs.readFileSync(fileChecksumPath, 'utf8').trim();
+  } catch (error) {
+    // File doesn't exist or can't be read
+    // Create file with the checksum
```
lukemelia, line 42: This should really only happen once per environment, right? Maybe we should surface the error to sentry or at least log it then, because it does seem exceptional.

lukemelia (approved):

lukemelia:
Lovely PR description and diagram!

packages/realm-server/handlers/handle-post-deployment.ts
jurgenwerk: I thought enqueueing all these jobs was pretty instantaneous. But I did a check and the endpoint call took more than a minute, on staging. Which really is a good reason to make a new job to do this work. I'll add it!

packages/realm-server/lib/boxel-ui-change-checker.ts
jurgenwerk: That's true, the first time only, per environment. Agree, I'll send to Sentry

habdelra (approved):
pending creating the single job to kick off all the reindex jobs,  i think this looks good. (probably the full reindex route could use that same single job too)

packages/realm-server/handlers/handle-post-deployment.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    let boxelUiChangeCheckerResult =
+      await compareCurrentBoxelUIChecksum(assetsURL);
+
+    if (
+      boxelUiChangeCheckerResult.currentChecksum !==
+      boxelUiChangeCheckerResult.previousChecksum
+    ) {
+      await queue.publish<void>({
```
jurgenwerk, line 35: Here is the new job that reindexes all the specified realms

packages/runtime-common/worker.ts
```diff
@@ -198,6 +213,8 @@ export class Worker {
     this.#runner = indexRunner;
     this.#reportStatus = reportStatus;
     this.#realmServerMatrixUsername = realmServerMatrixUsername;
+    this.#dbAdapter = dbAdapter;
+    this.#queuePublisher = queuePublisher;
```
jurgenwerk, line 217: Added these two to the worker so that jobs can spawn new jobs (to support this case where we have a new job that spawns realm reindex jobs)

packages/runtime-common/worker.ts
```diff
@@ -341,6 +359,45 @@ export class Worker {
...
+  private fullReindex = async (
+    args: FullReindexArgs & { jobInfo?: JobInfo },
+  ) => {
+    this.#log.debug(
+      `${jobIdentity(args.jobInfo)} starting reindex-all for job: ${JSON.stringify(args)}`,
+    );
+    this.reportStatus(args.jobInfo, 'start');
+    let { realmUrls } = args;
+
+    const realmOwners = await fetchAllRealmsWithOwners(this.#dbAdapter);
```
jurgenwerk, line 371: Added this new query so that we can avoid N+1's for fetching realm owner usernames when triggering reindexing for individual realms

jurgenwerk:
@habdelra @lukemelia this is now ready for you to take another look. I added a new job that we discussed which kicks off all the reindex jobs, and used that in the full reindex route too

lukemelia (approved):

habdelra (approved):

PR #3234, opened 2025-09-11
server: Fix clash detection on published realms

backspace:
We can’t currently create new realms as published realms are interpreted as being served from the realm server’s root despite actually being at different hostnames:

Boxel 2025-09-11 12-30-55

I added logging to confirm a published realm was the problem:

Service logs | Elastic Container Service | us-east-1 2025-09-11 10-21-00

This adds a hostname comparison to the code that searches for a realm served at `/`.

The realm server test failure is [also on `main`](https://github.com/cardstack/boxel/actions/runs/17642021964/job/50131086553#step:10:102), so a red herring.

### DIFF
--- a/packages/realm-server/server.ts
+++ b/packages/realm-server/server.ts
@@ -302,14 +302,18 @@ export class RealmServer {
     backgroundURL?: string;
     iconURL?: string;
   }): Promise<Realm> => {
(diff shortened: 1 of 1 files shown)


tintinthong (approved):
👍

PR #3236, opened 2025-09-11
host: Add login via host mode Connect button

backspace:
This adds redirection when you click Connect from a card page in host mode, first to the base realm server URL, and then back to the host mode URL after completion.

screencast 2025-09-12 16-05-57

The valid host mode domains are added to the `host` environment so incoming messages from the iframe can be verified to be from a trusted source. I renamed the development-facilitating `host-mode-origin` to `hostModeOrigin` while promoting it to a true Ember query parameter to follow existing conventions, which was needed to preserve it through the URL changes.

Deployed to staging, you can see that the Connect button is still showing even after going through the login process, so further work is needed, but that’ll be in a future PR.

screencast 2025-09-12 16-02-39


lukemelia (approved):
Per our discussion, we should be able to do this without `validPublishedRealmDomains` config

backspace:
> Per our discussion, we should be able to do this without `validPublishedRealmDomains` config

hmm indeed… since this is the parent page that includes the iframe, whose `src` is a getter on the component, we already know what origin to expect messages from, I’ll simplify this, thanks.

PR #3238, opened 2025-09-11
Capture error doc from prerender result

habdelra:
This PR allows us to capture an error doc from the prerender result


packages/host/app/services/store.ts
```diff
@@ -826,17 +842,19 @@ export default class StoreService extends Service implements StoreInterface {
     }
   }
 
-  // this function is used to determine if the instance will be auto-saved or not
-  // this is a temporary function that is likely to go away with the creation of completion emphemeral state solution of the store/realm
-  // the only use-case for this function is determining if a preview instance in catalog realm (which is a read-only),
-  // st a card can be mutable without persisting to the server
+  // this function is used to determine if the instance will be auto-saved or
+  // not this is a temporary function that is likely to go away with the
```
lukemelia:
```suggestion
  // note this is a temporary function that is likely to go away with the
```

packages/host/app/services/store.ts
```diff
@@ -826,17 +842,19 @@ export default class StoreService extends Service implements StoreInterface {
...
+  // creation of completion ephemeral state solution of the store/realm the
+  // only use-case for this function is determining if a preview instance in
+  // catalog realm (which is a read-only), st a card can be mutable without
```
lukemelia, line 853: "st a card": typo?

packages/realm-server/prerender.ts
```diff
@@ -50,28 +58,73 @@ export async function prerenderCard({
...
+    let atomHTML: string | null = null,
+      iconHTML: string | null = null,
+      embeddedHTML: Record<string, string> | null = null,
+      fittedHTML: Record<string, string> | null = null;
+    if (meta?.types) {
+      let results = [
+        await renderAncestors(page, 'fitted', meta.types),
+        await renderAncestors(page, 'embedded', meta.types),
+        await renderHTML(page, 'atom', 0),
+        await renderIcon(page),
```
lukemelia, line 108: Do you mean for these to run in parallel? If so, I don't think they are with this syntax.

packages/realm-server/prerender.ts
```diff
@@ -120,29 +182,32 @@ async function renderHTML(
...
   format: string,
   ancestorLevel: number,
-): Promise<string> {
+): Promise<string | RenderError> {
   await transitionTo(page, 'render.html', format, String(ancestorLevel));
   let result = await captureResult(page, 'innerHTML');
   if (result.status === 'error') {
-    throw new Error('todo: error doc');
+    let error = JSON.parse(result.value) as CardErrorJSONAPI;
+    return { ...error, error: error.message };
```
lukemelia, line 190: My reading of this is that it will produce an object that has a `message` property and an `error` property, both with the same value. Is that right?

lukemelia (approved):

packages/realm-server/prerender.ts
habdelra: actually no, we cannot run these in parallel. the rendering is actually all messed up when they are

packages/realm-server/prerender.ts
habdelra: no. it will make an object that has an error property that is set to a mesasge value.

packages/host/app/services/store.ts
habdelra: probably, my change was just to realign the comment, I think perhaps @backspace or @tintinthong wrote this comment initially

packages/realm-server/prerender.ts
habdelra: keep in mind these are all sharing the same chrome window. you need to render one at a time otherwise the requests will fight for the same window and you won't really know what you are looking at

packages/realm-server/prerender.ts
lukemelia: plus all the properties that are on the error object (from the `...error`)

packages/realm-server/prerender.ts
lukemelia: Ah, yes

packages/realm-server/prerender.ts
habdelra: confirm. the idea here is that i'm using the error property as a way to easily discriminate the error result type (see isRenderError())

PR #3239, opened 2025-09-11
Ability to add theme css imports

burieberry:
- Implement ability to add theme css imports, such as fonts. Files to review: `base/card-api`, `base/field-component`, `boxel-ui/components/card-container/index`
- Moved `htmlSanitize` helper to boxel-ui because of issues with trying to import it to boxel-ui from runtime-common
- Updated sample theme files with font import urls

A single url can contain multiple fonts but here's an example for multiple imports:
theme-css-imports


packages/base/field-component.gts
```diff
@@ -243,6 +243,14 @@ export function getBoxComponent(
...
   }
 
+  function getCssImports(card?: CardDef) {
+    // for Theme card and its descendants, directly use the `cssImports` field;
+    // for all other cards, get imports via the Theme card linked from cardInfo
+    return card && 'cssImports' in card
+      ? (card as Theme).cssImports
+      : card?.cardInfo?.theme?.cssImports;
+  }
+
```
habdelra: so technically the `card` that is passed in should really be typed as a `Theme` card def right? because there is no guarantee that some random card that is passed in here does also coincidentally happen to have a field called `cssImports`. A safer way to check if you can use this field is to actually check the coderef for the field to see if it's really the field you think it is. I think there is a `getField()` functino in the API somewhere that will return a `Field` instance, and from there you can check `Field.card` to see what the backing card is for the field. that will be much more solid in terms of making sure this field is what you think it is.

burieberry: I want the Theme card and its subclasses and subclasses of subclasses etc. Do we have a method that recursively loads the ancestry line? Another option might be to add a symbol to the Theme card. What do you think? @habdelra

burieberry: ok I see what you mean. it worked out I think, thanks!

habdelra: there is a functino that does exactly this. in card-api.ts there is an `instanceOf()` function that uses the prototype chain and code refs to determine if an instance inherits from a particular type.

lukemelia (approved):

PR #3240, opened 2025-09-12
Implement /_unpublish-realm endpoint

FadhlanR:


packages/runtime-common/index-writer.ts
```diff
@@ -282,6 +282,22 @@ export class Batch {
...
 
+  async removeRealm(): Promise<void> {
+    await this.ready;
+
+    // Delete all index entries for this realm from working table
+    await this.#query([
+      `DELETE FROM boxel_index_working WHERE realm_url =`,
+      param(this.realmURL.href),
+    ] as Expression);
+
```
backspace: Is deletion from `boxel_index` also supposed to happen? I noticed these entries left behind after I unpublished a realm:

boxel_index — boxel — boxel realm server 2025-09-12 13-53-18

packages/runtime-common/index-writer.ts
```diff
@@ -282,6 +282,22 @@ export class Batch {
...
+    // Delete realm version tracking
+    await this.#query([
+      `DELETE FROM realm_versions WHERE realm_url =`,
+      param(this.realmURL.href),
+    ] as Expression);
+  }
```
backspace: I still have the unpublished realm in this table:

realm_versions — boxel — boxel realm server 2025-09-12 13-57-09

These both seem like something to have assertions on in the tests but I couldn’t find any.

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -91,9 +92,8 @@ module(basename(__filename), function () {
         copySync(join(__dirname, 'cards'), dir.name);
       });
 
-      module('various other realm tests', function (hooks) {
+      module('publish and unpublish realm tests', function (hooks) {
```
backspace: It’s annoying but I feel like this file is big enough that it should be split, like the `realm-endpoints` directory, into named groups.

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -335,366 +289,267 @@ module(basename(__filename), function () {
...
+          );
+          assert.notEqual(
+            secondResponse.body.data.attributes.lastPublishedAt,
+            firstTimestamp,
+            'Timestamp is updated on republish',
           );
         });
 
-        test('POST /_create-realm with invalid JSON', async function (assert) {
+        test('POST /_publish-realm returns bad request for missing sourceRealmURL', async function (assert) {
```
backspace: For ease of review, I try to minimise diffs so it’s clear what’s really changing; it appears to me that the “deletions” are really movements in the file, which makes it harder to understand.

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -706,55 +561,894 @@ module(basename(__filename), function () {
...
             );
-          }
+
+          assert.strictEqual(unpublishResponse.status, 200, 'HTTP 200 status');
+          assert.strictEqual(
+            unpublishResponse.body.data.type,
+            'unpublished_realm',
+          );
+          assert.ok(
+            unpublishResponse.body.data.id,
```
backspace: Not that there’s much danger of this not being true but wouldn’t it make sense to assert that the id returned is the same as the one created by `_publish-realm`?

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -706,55 +561,894 @@ module(basename(__filename), function () {
...
+              )}`,
+            )
+            .send(
+              JSON.stringify({
+                publishedRealmURL:
+                  'http://testuser.localhost/non-existent-realm/',
+              }),
+            );
+
+          assert.strictEqual(response.status, 404, 'HTTP 404 status');
```
backspace: I don’t know how much we’re adhering to this but a REST-esque style would have this be `POST [realm URL]/_unpublish` or the like, which is a situation where it would make sense to 404. But 404 on `POST /_unpublish-realm` seems a bit off to me, as the route _does_ exist, just not the thing the body references. But maybe this is a niche perspective.

backspace (changes requested):
I tried this locally and it worked as expected other than what I noted about the database records. Maybe I’m misunderstanding how it’s supposed to work though?

FadhlanR:
> I tried this locally and it worked as expected other than what I noted about the database records. Maybe I’m misunderstanding how it’s supposed to work though?

Sorry for not putting the PR description. You're right, this endpoint should reverse all processes of the publish realm endpoint. This PR must do these processes:

1. Unmount realm from virtual network.
2. Remove entries from boxel_index.
3. Remove the folder of the published realm.
4. Remove the user permissions for the published realm

packages/runtime-common/index-writer.ts
```diff
@@ -282,6 +282,22 @@ export class Batch {
     ]);
   }
 
+  async removeRealm(): Promise<void> {
```
habdelra: this approach is wrong. we need to use invalidation. please leverage the incremental() and use the delete operation.

packages/runtime-common/index-writer.ts
```diff
@@ -282,6 +282,22 @@ export class Batch {
...
+    await this.ready;
+
+    // Delete all index entries for this realm from working table
+    await this.#query([
```
habdelra: this is very wrong. it bypasses all of our invalidation logic

FadhlanR:
Please hold off this PR. I will update this to use `incremental` job type instead of introducing a new one.

habdelra (changes requested):
refactor this to use incremental index with a delete operation to unpublish a realm

FadhlanR:
I've implemented the approach to unpublish realm using incremental indexing. In realm, we have a function to delete card source or instance. I created a new function named `deleteAll` following that delete function to delete multiple files from a directory and trigger the incremental indexing with delete operation for those files. The result of this function is different from before - now the deleted files will still exist in the `boxel_index` table with the latest realm version, but they have the is_deleted field set to `true`.

packages/realm-server/tests/server-endpoints-test.ts
FadhlanR: I've created a new file for publish and unpublish realm endpoint.

packages/realm-server/tests/server-endpoints-test.ts
FadhlanR: You're right, I updated to make the endpoint return `422` error code for this case.

habdelra:
> I've implemented the approach to unpublish realm using incremental indexing. In realm, we have a function to delete card source or instance. I created a new function named `deleteAll` following that delete function to delete multiple files from a directory and trigger the incremental indexing with delete operation for those files. The result of this function is different from before - now the deleted files will still exist in the `boxel_index` table with the latest realm version, but they have the is_deleted field set to `true`.


yes, that is the correct approach. deleted items should have a "tombstone" entry in the index.

packages/realm-server/handlers/handle-unpublish-realm.ts
```diff
@@ -0,0 +1,214 @@
...
+      let entries = readdirSync(currentPath, { withFileTypes: true });
+
+      for (let entry of entries) {
+        let fullPath = join(currentPath, entry.name);
+
+        if (entry.isDirectory()) {
+          traverseDirectory(fullPath, basePath);
+        } else {
+          // Calculate relative path from the original realm root
+          let relativePath = fullPath.replace(basePath, '').replace(/^\//, '');
```
habdelra, line 43: you should use the RealmPaths.local() here--this is what it was made for

packages/runtime-common/realm.ts
```diff
@@ -881,6 +881,36 @@ export class Realm {
     });
   }
 
+  async deleteAll(paths: LocalPath[]): Promise<void> {
```
habdelra, line 884: Good!

packages/realm-server/handlers/handle-unpublish-realm.ts
```diff
@@ -0,0 +1,214 @@
...
+        realmsRootPath,
+        PUBLISHED_DIRECTORY_NAME,
+        publishedRealmInfo.id,
+      );
+      let allFilePaths = collectAllFilePaths(publishedRealmPath);
+
+      if (allFilePaths.length > 0) {
+        await existingPublishedRealm.deleteAll(allFilePaths);
+      }
+      removeSync(publishedRealmPath);
```
habdelra, line 151: why do we do this? isn't this what deleteAll() is doing in the `removePromises()`?

FadhlanR: This is to remove the root folder of published realm directory. In deleteAll function, we only delete the file contents under published realm directory.

lukemelia (approved):

packages/realm-server/tests/server-endpoints-test.ts
backspace: 👍🏻 thanks, I feel like our status codes are chaotic/inconsistent, but it doesn’t have to get worse 😆

backspace (approved):
I tried this again locally and the index/working index entries do now stay with `is_deleted = TRUE`, and I got a 422 as described

habdelra (approved):

PR #3244, opened 2025-09-12
Allow spec to be displayed in RHS 

tintinthong:
- spec is meant to support commands and classes but top-level file incompatibility message is preventing renders of the 3 col tab on rhs
- main issue is fileIncompatibility and itemIncompatibility (incompatibility related to the declaration) is handled at the top-level of module-inspector
- in this PR, I separated  fileIncompatibility with itemIncompatibility. 
   - fileIncompatibility remains at the top-level of module-inspector
   - itemIncompatibility calculated at the component. This means schema, playground preview and spec handle those in their own component (if needed)
   
   
Screenshot 2025-09-15 at 12 39 21

   
**Before**
Screenshot 2025-09-15 at 12 34 12

**After**
Screenshot 2025-09-15 at 12 25 30

**A command needs a spec**

Here are the UI treatments for a command

Screenshot 2025-09-15 at 12 45 55

Screenshot 2025-09-15 at 12 46 01

Screenshot 2025-09-15 at 12 46 07

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/module-inspector.gts
+++ b/packages/host/app/components/operator-mode/code-submode/module-inspector.gts
@@ -437,13 +437,6 @@ export default class ModuleInspector extends Component<ModuleInspectorSignature>
       <SyntaxErrorDisplay
         @syntaxErrors={{@moduleAnalysis.moduleError.message}}
       />
-    {{else if @card}}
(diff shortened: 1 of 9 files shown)


backspace (approved):
Nice hierarchy rearrangement, and I appreciate the thorough explanation 🤩

PR #3252, opened 2025-09-15
Include published realms table to sqlite

FadhlanR:
This PR includes published_realms table to sqlite. Since uuid is not compatible in sqlite, I added codes to convert uuid to autoincrement integer in sqlite.


lukemelia (commented):
I'll defer to @habdelra on this review. Does it make sense to have published realms in sqlite at all?

habdelra:
> I'll defer to @habdelra on this review. Does it make sense to have published realms in sqlite at all?

given how the published at value is populated in our realm response it does. see this ticket: https://linear.app/cardstack/issue/CS-9413/need-to-add-published-realms-to-sqlite

habdelra (changes requested):
So in the ticket i requested that we _keep_ UUID as the column type and adjust the script that generated the sqlite to handle that. i don't think we should change the type for the ID just because of sqlite. it should be the otherway around. we should be building the DB optimized for postgres and the way we use sqllite has to change to accomodate this.

Please take another stab at this. and instead of changing our DB migration, change the script that generates SQLlite so that it can handle UUID like I outlined in the ticket

habdelra:
@FadhlanR take a look at this branch--i think this is almost or maybe exactly what we want. https://github.com/cardstack/boxel/tree/sqlite-uuid-support

FadhlanR:
> @FadhlanR take a look at this branch--i think this is almost or maybe exactly what we want. https://github.com/cardstack/boxel/tree/sqlite-uuid-support

Merged that branch here. Thanks! Please check again.

habdelra (approved):

PR #3253, opened 2025-09-15
Introduce prerender server

habdelra:
This PR introduces a prerender server. The idea is that this server will ultimately be run in it's own ECS container and accept prerender requests that originate from the same VPC (enforced by AWS infra). kudos to GPT 5.0! huge time save here

I'm holding off on populating the README more fully until more of this vision is fleshed out...


packages/realm-server/tests/prerender-server-test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+module(basename(__filename), function () {
+  module('Prerender server', function (hooks) {
+    let request: SuperTest<Test>;
+    let prerenderServer: Server;
+    const testUserId = '@jade:localhost';
+
+    hooks.before(() => {
+      // Ensure chrome is available for puppeteer in CI
+      execSync('pnpm puppeteer browsers install chrome');
```
habdelra: GPT 5 thought it was a good idea to add this 🤷

backspace: That’s taken care of by [this script line](https://github.com/cardstack/boxel/blob/47cec66516ce4431868ca7c635b9136072f61a6f/.github/workflows/ci.yaml#L196)

jurgenwerk (approved):

PR #3254, opened 2025-09-15
Extract card serialization/field support types and functions from card-api into new modules

lukemelia:
Reduce card-api line count by about 15%


backspace (approved):

habdelra (approved):
nice!

habdelra:
probably we could move other module scoped state out too--but yeah i think this particular one gives us the biggest bang for our buck

PR #3256, opened 2025-09-17
Introduce page pooling in prerender server

habdelra:
This PR introduces page pooling in the prerender server. The idea is that prerenders of pages from the same realm will reuse the same page so that we can take advantage of the fact that the store is already loaded.


packages/realm-server/prerender/index.ts
```diff
@@ -29,154 +37,280 @@ export type PermissionsMap = {
...
+      );
+      if (result.status === 'error') {
+        error = JSON.parse(result.value) as CardErrorJSONAPI;
+      } else if (isRenderError(result)) {
+        error = result;
+      }
+      const isolatedHTML = result.status === 'ready' ? result.value : null;
+
+      // TODO consider breaking out rendering search doc into its own route so
+      // that we ran fully understand all the linked fields that are used in all
```
lukemelia:
```suggestion
      // that we can fully understand all the linked fields that are used in all
```

lukemelia (approved):

jurgenwerk (approved):

PR #3258, opened 2025-09-18
Implement publish realm button and make publishing to default url work

FadhlanR:
https://github.com/user-attachments/assets/0457a02f-9726-4609-9c25-f98c4bb338f1


backspace:
I’ll try this locally instead but in the preview deployment the published realm URL is invalid, is there a followup to address that?

buck 2025-09-22 10-20-25

packages/boxel-ui/addon/raw-icons/undo-arrow.svg
backspace, line 1: Would this work vs using a new file?

Boxel Components 2025-09-22 10-30-37

(I looked for one for publish but it’s either a book with a corner arrow or a file with a centre arrow)

packages/host/config/environment.js
```diff
@@ -39,6 +39,8 @@ module.exports = function (environment) {
     minSaveTaskDurationMs: 1000,
     renderTimeoutMs: 30_000,
     iconsURL: process.env.ICONS_URL || 'https://boxel-icons.boxel.ai',
+    defaultPublishedRealmDomain:
+      process.env.DEFAULT_PUBLISHED_REALM_DOMAIN || 'localhost:4201',
```
backspace, line 43: I see that this is what would need to be set for a correct domain in staging, which wouldn’t work in the preview. But is there an issue about setting these for deployed environments, or why not include that in this PR?

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -89,12 +133,37 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
...
+            {{#if this.isPublishing}}
+              <BoxelButton
+                @kind='primary'
+                @size='tall'
+                class='publish-realm-button publishing'
+                {{on 'click' this.togglePublishingRealmPopover}}
+                data-test-publish-realm-button
+              >
+                <Refresh width='22' height='22' class='publish-icon' />
+                Publishing...
```
backspace:
```suggestion
                Publishing…
```

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -89,12 +133,37 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
...
+            {{else}}
+              <BoxelButton
+                @kind='primary'
+                @size='tall'
+                class='publish-realm-button'
+                {{on 'click' this.openPublishRealmModal}}
+                data-test-publish-realm-button
+              >
+                <PublishSiteIcon width='22' height='22' class='publish-icon' />
+                Publish...
```
backspace:
```suggestion
                Publish…
```

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -0,0 +1,465 @@
...
+                      {{#if (this.isUnpublishingRealm this.generatedUrl)}}
+                        <LoadingIndicator />
+                        Unpublishing...
+                      {{else}}
+                        <UndoArrow
+                          width='11'
+                          height='11'
+                          class='unpublish-icon'
+                        />
+                        Unpublish
```
backspace, line 256: I left a comment on the design but will also ask here whether we think this is a clear term for average people. What differentiates this from “Delete”, which is more immediately-grasped, and seems more direct re the finality?

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -0,0 +1,465 @@
...
+                      @kind='text-only'
+                      @size='extra-small'
+                      @disabled={{this.isUnpublishingRealm this.generatedUrl}}
+                      class='unpublish-button'
+                      {{on 'click' (fn @handleUnpublish this.generatedUrl)}}
+                      data-test-unpublish-button
+                    >
+                      {{#if (this.isUnpublishingRealm this.generatedUrl)}}
+                        <LoadingIndicator />
+                        Unpublishing...
```
backspace:
```suggestion
                        Unpublishing…
```

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -0,0 +1,465 @@
...
+              @kind='primary'
+              @size='tall'
+              {{on 'click' (fn @handlePublish this.selectedPublishedRealmURLs)}}
+              @disabled={{this.isPublishDisabled}}
+              class='publish-button'
+              data-test-publish-button
+            >
+              {{#if this.isPublishing}}
+                <LoadingIndicator />
+                Publishing...
```
backspace:
```suggestion
                Publishing…
```

backspace:
Worth some investigation: I tried publishing and unpublishing locally and it worked initially, but when I did it more I got into a weird state and seemingly can’t recover:

Boxel 2025-09-22 10-48-32

packages/postgres/migrations/1758008721717_alter-last-published-at-to-bigint.js
```diff
@@ -0,0 +1,53 @@
...
+  // Drop the bigint column
+  pgm.dropColumn('published_realms', 'last_published_at');
+
+  // Rename the temp column back to the original name
+  pgm.renameColumn(
+    'published_realms',
+    'last_published_at_temp',
+    'last_published_at',
+  );
+};
```
backspace, line 53: What’s this for, just easier date calculations?

backspace (approved):

packages/host/config/environment.js
FadhlanR: You're right. I've included it in the new commit.

packages/host/app/components/operator-mode/publish-realm-modal.gts
FadhlanR: Yeah it might be worth for a discussion. We need design perspective for this.

packages/postgres/migrations/1758008721717_alter-last-published-at-to-bigint.js
FadhlanR: Yes for easier date calculation since the previous type does not store the timezone. I also followed approach we use in other tables.

FadhlanR:
> Worth some investigation: I tried publishing and unpublishing locally and it worked initially, but when I did it more I got into a weird state and seemingly can’t recover:
> 
> Boxel 2025-09-22 10-48-32

I cannot reproduce it in my local. I'll file a ticket and investigate this bug.

PR #3260, opened 2025-09-18
Login to multiple realms with one request on matrix service start

jurgenwerk:
This PR adds a change where a bulk realm login is performed on matrix service start. This should reduce the amount of requests for realms login after user logs in. After that, we still have the functionality for individual realm logins, which is used for cases when creating a new realm, or reauths after token expiry for individual realms.


backspace:
What should it look like for this to work? I ran the branch locally and still have a `POST _session` for each realm:

Boxel 2025-09-26 07-49-45

backspace:
> What should it look like for this to work? I ran the branch locally and still have a `POST _session` for each realm:
> 
> Boxel 2025-09-26 07-49-45

DISREGARD I failed to run the branch properly, sorry 😬 But when I’m actually running it, I can’t log in:

Boxel 2025-09-26 08-07-56

Let me know if there‘s any debugging I can provide.

jurgenwerk:
@backspace hmm that doesn't happen at my end, can you see the realm error message?

backspace:
> @backspace hmm that doesn't happen at my end, can you see the realm error message?

if you mean the realm server log, not much there:

```
<-- GET */* http://localhost:4201/_catalog-realms
--> GET */* http://localhost:4201/_catalog-realms: 200
<-- POST application/json http://localhost:4201/_server-session
--> POST application/json http://localhost:4201/_server-session: 401
<-- POST application/json http://localhost:4201/_server-session
--> POST application/json http://localhost:4201/_server-session: 201
<-- OPTIONS */* http://localhost:4201/_realm-auth
--> OPTIONS */* http://localhost:4201/_realm-auth: 204
<-- POST application/vnd.api+json http://localhost:4201/_realm-auth
--> POST application/vnd.api+json http://localhost:4201/_realm-auth: 422
```

The 422 is this:

```
{"errors":["User in JWT not found"]}
```

jurgenwerk:
I can't reproduce, I also just tried making a new user and it works ok for me 🤔 Is something wrong with your user in a way where it doesn't have the matrix id in the users table? I guess we can pair to see what's going on...

packages/host/app/components/matrix/login.gts
```diff
@@ -210,7 +210,11 @@ export default class Login extends Component<Signature> {
       // component being removed from the DOM and destroyed. Keep in mind that in EC tasks,
       // awaits are really just syntactic sugar for yields, and that we yield to
       // this.matrixService.start()
-      await this.matrixService.start({ auth, refreshRoutes: true });
+      await this.matrixService.start({
+        auth,
+        refreshRoutes: true,
+        loginToAllAccessibleRealmsInBulk: true,
```
habdelra: is there a scenario where we would not want to do this?

jurgenwerk: Actually I can't think of any, I think I can remove this option and make it so that it always logs into all realms that are not logged in when matrix service starts. I'll make the adjustment

packages/host/app/services/realm-server.ts
```diff
@@ -349,8 +382,6 @@ export default class RealmServerService extends Service {
 
   private loggingIn: Promise<void> | undefined;
 
-  // login happens lazily as you need to interact with realm server which
-  // currently only constitutes creating realms
```
jurgenwerk, line 353: Actually this comment isn't true anymore so I am deleting it. We are logging into the realm server in various other places too

packages/host/app/components/matrix/login.gts
jurgenwerk: I removed this option - now there will be a bulk login to realms every time matrix service starts

lukemelia (approved):

PR #3263, opened 2025-09-18
Evict page from prerender pool on unrecoverable error/timeout

habdelra:
This PR evicts a chrome page from our prerender pool when an unrecoverable error occurs (ember is unresponsive) or the page times out.


packages/realm-server/prerender/prerenderer.ts
```diff
@@ -0,0 +1,480 @@
...
+    }
+  >();
+  #lru = new Set<string>();
+  #pendingByRealm = new Map<string, Promise<void>>();
+  #maxPages: number;
+  #secretSeed: string;
+  #stopped = false;
+  #evictionMetrics = {
+    byRealm: new Map<string, { unusable: number; timeout: number }>(),
+  };
```
habdelra, line 62: ultimately i'd like to expose the eviction metrics to grafana so that we can more easily spot issues that are causing ember to crash during prerendering

lukemelia (approved):

PR #3266, opened 2025-09-19
Implement full width card in host mode

FadhlanR:
Before:
Screenshot 2025-09-19 at 14 41 18

After:

Screenshot 2025-09-19 at 14 40 23


backspace (approved):
I was able to see this in the preview deployment, which can simulate host mode with a `hostModeOrigin` parameter:

mar10 2025-09-22 17-25-38

PR #3268, opened 2025-09-19
Introduce Prerender Manager to orchestrate prerender servers

habdelra:
Introduces a Koa-based prerender manager to coordinate multiple prerender servers with realm affinity, multiplexing, capacity-aware routing, pressure fallback, and health-based eviction.

Details
  - Endpoints: GET/HEAD / (health), POST /prerender (proxy), POST/DELETE /prerender-servers (register/unregister), DELETE /prerender-servers/realms/:encodedRealm (realm disposal).
  - Routing: sticky realm affinity, multiplex rotation (>1), capacity-first selection, pressure fallback via global LRU realm.
  - Health: sweep function pings servers and removes unreachable ones; logs unhealthy detection and removals.


packages/realm-server/scripts/start-prerender-manager.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/sh
+
+# Start the prerender manager in staging
+
+NODE_ENV=production \
+  NODE_NO_WARNINGS=1 \
+  ts-node \
+  --transpileOnly prerender/manager-server \
+  --port=${PRERENDER_MANAGER_PORT:-4222}
```
backspace, line 9: I guess it follows the realm server pattern but is there a reason to have production and staging scripts if they’re not different?

packages/realm-server/prerender/manager-app.ts
```diff
@@ -0,0 +1,391 @@
...
+      if (!list.includes(any)) list.push(any);
+      if (list.length > multiplex) list = list.slice(-multiplex);
+      registry.realms.set(realm, list);
+      return any;
+    }
+    return null;
+  }
+
+  // health sweep: remove unreachable servers and clean up realm mappings
+  async function sweepServers() {
```
backspace, line 278: What actually calls this? I could only find it being used in a test.

packages/realm-server/prerender/manager-app.ts
```diff
@@ -0,0 +1,391 @@
...
+          requestBody = JSON.parse(raw);
+        } catch (e) {
+          log.debug('Invalid JSON body on registration; treating as empty:', e);
+        }
+      }
+
+      let attrs = requestBody?.data?.attributes || {};
+      let capacity: number = Number(attrs.capacity ?? 4);
+      let url: string | undefined = attrs.url;
+      if (!url) {
```
backspace, line 120: Is auth needed for this or the other endpoints that make changes?

backspace (approved):
I imagine we’ll want telemetry etc for insight in deployed environments but maybe that’s for the future?

I didn’t run this locally but did read through the tests, looks sensible

packages/realm-server/prerender/manager-app.ts
jurgenwerk: is it going to be a cron job?

jurgenwerk (approved):

packages/realm-server/scripts/start-prerender-manager.sh
habdelra: good point. let's just have one

packages/realm-server/prerender/manager-app.ts
habdelra: good catch--this was supposed to be called by a setInterval in the manager app. fixing....

packages/realm-server/prerender/manager-app.ts
habdelra: No auth necessary, the callers (our workers) will all be in the same VPC as the server--this is not exposed to the outside world, so we can just use a security group to make sure there no unintended access.

PR #3269, opened 2025-09-22
Fix: AI panel stuck after sending a message from ai-app-generator

FadhlanR:
I initially investigated the issue where the AI panel gets stuck after clicking the "create this for me" button in `ai-app-generator`. I found that we don’t have a mechanism to ensure a `roomResource` exists before the `enterRoom` or `enterRoomInitially` functions are called. To fix this, I added logic to wait for the first sync and to wait until the room is included in the sync. With this fix, this PR resolves the following issues:

* [x] A new room is always created every time a user accesses host.
* [x] The AI panel gets stuck after sending a message from `ai-app-generator`.
* [x] The AI panel gets stuck after creating a new room with settings to copy file history or to include the previous open room summary.

### DIFF
--- a/packages/matrix/tests/live-cards.spec.ts
+++ b/packages/matrix/tests/live-cards.spec.ts
@@ -101,18 +101,6 @@ test.describe('Live Cards', () => {
       },
     });
 
-    // The indexing indicator might appear and disappear so fast
-    // so no guarantee we can check it
-    // await expect(
-    //   page.locator('[data-test-realm-indexing-indicator]'),
-    // ).toHaveCount(1);
-    // await expect(
-    //   page.locator(`[data-test-card="${instanceUrl}"]`),
-    // ).toContainText('updated card title');
-    // await expect(
-    //   page.locator('[data-test-realm-indexing-indicator]'),
-    // ).toHaveCount(0);
-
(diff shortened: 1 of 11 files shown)


packages/matrix/tests/live-cards.spec.ts
```diff
@@ -101,15 +101,17 @@ test.describe('Live Cards', () => {
...
+    // so no guarantee we can check it
+    // await expect(
+    //   page.locator('[data-test-realm-indexing-indicator]'),
+    // ).toHaveCount(1);
+    // await expect(
+    //   page.locator(`[data-test-card="${instanceUrl}"]`),
+    // ).toContainText('updated card title');
+    // await expect(
+    //   page.locator('[data-test-realm-indexing-indicator]'),
+    // ).toHaveCount(0);
```
backspace: Should we just remove these then? If we really want to assert on this in tests there’s probably a way with fine-grained promises

backspace (approved):

packages/matrix/tests/live-cards.spec.ts
FadhlanR: I'll remove it for now; if needed, we can add it back later with a better promise implementation.

PR #3270, opened 2025-09-22
add created into realm file meta table (2nd attempt)

tintinthong:
This PR is a 2nd attempt at https://github.com/cardstack/boxel/pull/3025 and https://linear.app/cardstack/issue/CS-9160/look-back-at-payload-why-created-is-returning-0. 

To recap, the reason this solution is needed is bcos 
- linux doesn't support birthtime 
- efs variant of the nfsv4 file system doesn't support extended attributes

**New Solution**

We introduce a RealmFileMeta table 
```
 CREATE TABLE IF NOT EXISTS realm_file_meta (
   realm_url TEXT NOT NULL,
   file_path TEXT NOT NULL,
   created_at INTEGER NOT NULL,
   PRIMARY KEY ( realm_url, file_path ) 
);
```

This is a db solution to track created times of files that are created and deleted (via `persistFileMeta`). We don't expect created times to change if they already exists

This solution doesn't affect mtimes solution since that data is accessed directly from the file system

**How to Test**

- probably sort grid by created date
- create a new card
- the file should be at the top of the grid


tintinthong:
Oh wait. Might have overlooked something. Pause on review first

tintinthong:
So much flakiness in these test. Dont know why CI test fails but no test fails locally

tintinthong:
There is playground-panel tests which fail bcos it expects the latest (the last index) spec that was created to be displayed but it shows the first one. Its bcos search is sorted by createdDate that actually works now. I fixed it by just doing a re-order

can preview compound field instance
can update the instance chooser when selected declaration changes
can create new field instance (has preexisting Spec)

packages/postgres/migrations/1758580000000_create-realm-file-meta.js
```diff
@@ -0,0 +1,22 @@
+exports.up = (pgm) => {
+  pgm.createTable('realm_file_meta', {
+    realm_url: { type: 'varchar', notNull: true },
+    file_path: { type: 'varchar', notNull: true },
+    // store as epoch seconds (integer) to align with existing conventions
+    created_at: {
+  type: 'integer',
```
habdelra: perhaps this needs to be a big int?

habdelra: or wait--if this is seconds then integer is probably fine

lukemelia (approved):

packages/host/tests/integration/realm-test.ts
habdelra, line 1: do we have a delete test too?

tintinthong: added here https://github.com/cardstack/boxel/pull/3270/commits/ad82184777e3ef84fad96b1fa18285963b8bbf3e

tintinthong: thx to this comment, I know some assertions for when a card is created was missing in card-endpoints too.

habdelra:
consider merging upstream changes to get less flaky tests

habdelra (approved):

jurgenwerk (approved):
Our convention is to only keep the last schema file–you can delete old ones

tintinthong:
> Our convention is to only keep the last schema file–you can delete old ones

i have updated the script to remove previous schemas

backspace:
> i have updated the script to remove previous schemas

thank you 🎉

PR #3272, opened 2025-09-22
Add JWT verification failure logging

habdelra:
This PR adds JWT verification failure logging, and makes it very loud. The next time we encounter a scenario where JWT's fail to verify after a period that is too short (less than 7 days) use the grafana logs to inspect the realm server for this log message to get more insight on why the JWTs are prematurely failing to verify


backspace (approved):

jurgenwerk (approved):
👍

PR #3276, opened 2025-09-23
Fix: Cannot read properties of undefined (reading 'id') when publishing realm

FadhlanR:
There was a bug that @backspace encountered when publishing a realm. The issue occurs because the `publish-realm` endpoint does not remove the realm instance from `server.realms`, which allows multiple instances of the same realm to exist in `server.realms` while only one record is present in the `published_realms` table. When unpublishing, only one instance and the corresponding record in the `published_realms` table are removed. As a result, if the user tries to publish the realm again, an error occurs because the instance still exists in memory but no data exists in the table.


FadhlanR:
The fix is straightforward, so I’ll merge this PR directly.

backspace:
thanks for figuring this out 🎉

PR #3277, opened 2025-09-23
server: Check realm publishability before publishing

backspace:
Currently a realm without `publishable: true` can be a source when publishing.


packages/realm-server/tests/publish-unpublish-realm-test.ts
```diff
@@ -111,565 +112,639 @@ module(basename(__filename), function () {
...
+          .set(
+            'Authorization',
+            `Bearer ${createRealmServerJWT(
+              { user: ownerUserId, sessionRoom: 'session-room-test' },
+              realmSecretSeed,
+            )}`,
+          )
+          .send(
+            JSON.stringify({
+              sourceRealmURL: sourceRealmUrlString,
```
backspace, line 578: I recommend the “Hide whitespace” feature for review of this file because there are few changes other than the tests being nested in the source realm creation module, and the URL itself having changed in comparisons:

Image

lukemelia (approved):

FadhlanR (approved):

PR #3281, opened 2025-09-23
Fix AiAssistantApplyButton "preparing" and "ready" states

lukemelia:
Before:

image

After:

image

### DIFF
--- a/packages/host/app/components/ai-assistant/apply-button/index.gts
+++ b/packages/host/app/components/ai-assistant/apply-button/index.gts
@@ -25,58 +25,55 @@ interface Signature {
 }
 
 const AiAssistantApplyButton: TemplateOnlyComponent<Signature> = <template>
-  {{#if (eq @state 'ready')}}
-    <BoxelButton
-      @kind='primary'
-      @size='small'
-      class='apply-button'
-      {{setCssVar boxel-button-text-color='var(--boxel-dark)'}}
-      data-test-apply-state={{@state}}
-      ...attributes
-    >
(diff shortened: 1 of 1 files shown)


packages/host/app/components/ai-assistant/apply-button/index.gts
```diff
@@ -25,58 +25,55 @@ interface Signature {
...
+    class='state-indicator {{@state}}'
+    data-test-apply-state={{@state}}
+    ...attributes
+  >
+    {{#if (eq @state 'ready')}}
+      <BoxelButton
+        @kind='primary'
+        @size='small'
+        class='apply-button'
+        {{setCssVar boxel-button-text-color='var(--boxel-dark)'}}
```
burieberry: this seems unnecessary

packages/host/app/components/ai-assistant/apply-button/index.gts
```diff
@@ -25,58 +25,55 @@ interface Signature {
...
+        Working…
+      </BoxelButton>
+    {{/if}}
+  </div>
   <style scoped>
     .apply-button {
       --boxel-button-font: 600 var(--boxel-font-xs);
       padding: 3px 10px;
       min-width: inherit;
       min-height: inherit;
```
burieberry, line 67: maybe what you want is `@size='auto'` instead of `@size=small`. none of the small button properties are being used.

burieberry (approved):
Your changes fix the immediate problem, but this component can use a refactor to remove much of the css and the div block.

PR #3283, opened 2025-09-23
Avoid search returning 500 due to unknown card type

lukemelia:
- Adds negative caching for missing card definitions in query engine
- Affects realm's _search and _prerendered-search endpoints

### DIFF
--- a/packages/runtime-common/definitions-cache.ts
+++ b/packages/runtime-common/definitions-cache.ts
@@ -19,6 +19,7 @@ export class FilterRefersToNonexistentTypeError extends Error {
     if (opts?.cause !== undefined) {
       (this as any).cause = opts.cause;
     }
+    // make sure instances of this Error subclass behave like instances of the subclass should
     Object.setPrototypeOf(this, new.target.prototype);
   }
 }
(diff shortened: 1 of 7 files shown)


packages/runtime-common/definitions-cache.ts
```diff
@@ -7,16 +7,40 @@ import {
...
+  constructor(codeRef: ResolvedCodeRef, opts?: { cause?: unknown }) {
+    super(
+      `Your filter refers to a nonexistent type: import { ${codeRef.name} } from "${codeRef.module}"`,
+    );
+    this.name = 'FilterRefersToNonexistentTypeError';
+    this.codeRef = codeRef;
+    if (opts?.cause !== undefined) {
+      (this as any).cause = opts.cause;
+    }
+    Object.setPrototypeOf(this, new.target.prototype);
```
backspace, line 23: I don’t think I’ve ever seen this before, what does it do?

backspace (approved):

habdelra (approved):

packages/runtime-common/definitions-cache.ts
lukemelia: Adding a comment

PR #3289, opened 2025-09-25
Show delete button in inspector when file is empty

richardhjtan:
Screenshot 2025-09-25 at 3 22 35 PM


backspace (approved):
I tried this locally, both with-permission and without, it worked as expected 🎉

PR #3291, opened 2025-09-25
Regenerate schema (to include published_realms table which is needed for tests)

jurgenwerk:
We were experiencing errors in tests due to an outdated migration file. 

I deleted the old one and ran `pnpm make-schema`. 

image


backspace (approved):
Should we have something in CI that flags when there are two files here?

PR #3293, opened 2025-09-25
host: Update top bar arrangement

backspace:
This started as an attempt to make the host mode publish button align properly, with equal spacing:

image

I ended up rearranging the structure of the submode layout so each submode can place its items in the top bar container (named block `topBar`) instead of having multiple top-aligned containers, each positioned absolutely.

The different submodes before:

before

After:

after

This also updates the Publish button to look more like the design and copies the outline offset from the code submode new file button:

Boxel 2025-09-26 09-13-29


packages/host/app/components/operator-mode/code-submode/playground/playground-panel.gts
```diff
@@ -875,7 +875,8 @@ export default class PlaygroundPanel extends Component<Signature> {
       }
       .instance-chooser-container {
         position: sticky;
-        bottom: var(--playground-padding);
+        /* Without the top-bar-height addition, the instance chooser is cut off at the bottom */
+        bottom: calc(var(--playground-padding) + var(--top-bar-height));
```
backspace: I hate this as I don’t understand why the top bar’s height would matter here, and its height isn’t explicitly set, it just happens to be 60px from the height of its contents with padding. But without this the instance chooser was truncated:

Image

If anyone has an idea of how to better fix this I’d love to know!

backspace: I figured it out! The submode layout container was still 100% height despite now sharing viewport height with the non-absolute top bar. So I made the operator mode container flex column.

burieberry (approved):
Great improvements, thanks for fixing this. Percy changes look good.

PR #3295, opened 2025-09-25
Standardize all packages on TypeScript 5.8

ef4:


packages/host/app/components/matrix/room-message.gts
```diff
@@ -149,6 +149,7 @@ export default class RoomMessage extends Component<Signature> {
         @userMessageThisMessageIsRespondingTo={{this.userMessageThisMessageIsRespondingTo}}
         @registerScroller={{@registerScroller}}
         @isFromAssistant={{this.isFromAssistant}}
+        {{! @glint-ignore }}
```
ef4, line 152: This is working around what I think is a glint bug, I think it's OK to leave this ignore in until we can do the 2.0 upgrade.

ef4:
@lukemelia I think this is good to land, it should only effect types and the types are passing in CI. I don't think the CI failures here are real.

PR #3296, opened 2025-09-25
Update boxel-ui color components and base ColorField template

burieberry:
Changes:
- Enable theming
- Create `Swatch` component with different view styles
- Add text input field to the ColorPicker component. This enables colors in `okchl` color space to be able to use this component. The color preview icon still opens browser's default color-picker, but the value is also editable via the input field. Clicking X icon clears the input.
- Update ColorPalette component with the new ColorPicker and also add tooltips on color buttons.
- Add disabled state
- Update `base` package `ColorField` templates

color-picker

color-palette

swatch-default

swatch-round

### DIFF
--- a/packages/base/color.gts
+++ b/packages/base/color.gts
@@ -1,17 +1,21 @@
 import { Component, StringField } from './card-api';
-import { ColorPalette } from '@cardstack/boxel-ui/components';
-import { ColorPicker } from '@cardstack/boxel-ui/components';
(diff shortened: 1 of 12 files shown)


lukemelia (approved):
Can we document in the boxel-ui README when to use @layer and what layer names to use? This is fuzzy to me and maybe to others as well

richardhjtan (approved):

PR #3299, opened 2025-09-26
fix and improve avatar creator command

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-389/fix-avatar-creator-listing-missing

### DIFF
--- a/packages/catalog-realm/commands/suggest-avatar.gts
+++ b/packages/catalog-realm/commands/suggest-avatar.gts
@@ -31,15 +31,14 @@ export class SuggestAvatar extends Command<
       let result = await useAiAssistantCommand.execute({
         roomName: `Avatar Suggestions: ${name || 'Unnamed Avatar'}`,
         openRoom: true,
-        prompt:
-          'Please edit the following card with an avatar based upon params of https://getavataaars.com/. The params supported are: topType, accessoriesType, hairColor, facialHairType, clotheType, eyeType, eyebrowType, mouthType and skinColor.',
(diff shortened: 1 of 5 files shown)


packages/catalog-realm/commands/suggest-avatar.gts
```diff
@@ -31,15 +31,14 @@ export class SuggestAvatar extends Command<
...
-          'Please edit the following card with an avatar based upon params of https://getavataaars.com/. The params supported are: topType, accessoriesType, hairColor, facialHairType, clotheType, eyeType, eyebrowType, mouthType and skinColor.',
+        prompt: `Please suggest two example avatar prompts: one describing a visual style and one referencing a celebrity's look. Then ask if these examples are helpful or if different suggestions are needed.`,
       });
 
       if (result.roomId) {
         let setActiveLLMCommand = new SetActiveLLMCommand(this.commandContext);
         await setActiveLLMCommand.execute({
           roomId: result.roomId,
-          mode: 'act',
+          mode: 'ask',
```
tintinthong, line 41: wudn this just be 'act'. 'act' means the bot will auto-apply changes. 'ask' means the bot will need confirmation before it applies changes.

lucaslyl: Chuan just gave feedback about this. He mentioned it shouldn’t apply the changes immediately — instead, it should work like Remix, providing prompt example suggestions for the user to review and decide whether to accept.

tintinthong: oh right. okie dokes

tintinthong (approved):
nicee

PR #3302, opened 2025-09-26
Upgrade matrix js sdk version to 38.3.0

FadhlanR:


packages/runtime-common/matrix-client.ts
```diff
@@ -60,6 +60,7 @@ export class MatrixClient {
       if (!this.access) {
         throw new Error(`Missing matrix access token`);
       }
+      console.log('accessToken', this.access.accessToken);
```
lukemelia: Did you mean to leave this in?

patches/matrix-js-sdk@38.3.0.patch
```diff
@@ -0,0 +1,180 @@
...
+     [M_BEACON.name]: MBeaconEventContent;
+     [M_POLL_START.name]: PollStartEventContent;
+     [M_POLL_END.name]: PollEndEventContent;
++
++    "app.boxel.active_llm": any;
++    "app.boxel.llm_mode": any;
++    "app.boxel.room_skills": any;
++    "app.boxel.realm_event": any;
++
++    [t: string]: any;
```
lukemelia, line 15: Do we have to use a patch for this? Is there a way to re-open the type and add things?

package.json
```diff
@@ -70,7 +70,7 @@
     "ember-cli-htmlbars": "^6.3.0",
     "ember-resources": "catalog:",
     "ember-source": "catalog:",
-    "ember-template-imports": "^4.1.1",
+    "ember-template-imports": "^4.3.0",
```
lukemelia: Is this related to this PR?

lukemelia (approved):

patches/matrix-js-sdk@38.3.0.patch
FadhlanR: I don't see any other way for now. Actually, we can cast it where we use it, but I don't think that's any better.

PR #3303, opened 2025-09-26
Switch theming

richardhjtan:
https://github.com/user-attachments/assets/88858977-8722-4524-a4e5-32a068e39a3a


burieberry (commented):
For some of the themes, it's hard to tell when the switch is on or off because the colors remain the same. If it's easier, we can just have the default and have default coloring when off, accent or primary coloring when on. You can check the toggles in the "cookie settings" card here as an example to see which variables to use: https://tweakcn.com/editor/theme

richardhjtan:
> For some of the themes, it's hard to tell when the switch is on or off because the colors remain the same. If it's easier, we can just have the default and have default coloring when off, accent or primary coloring when on. You can check the toggles in the "cookie settings" card here as an example to see which variables to use: https://tweakcn.com/editor/theme

@burieberry good point, using default color when off can avoid the switch blend with the background for some themes. The switch should be more simpler now

burieberry (approved):
Looks good

PR #3305, opened 2025-09-27
cleanup listing isolated

tintinthong:
- add tags next to category
- make images scrollabe
- when a spec type cant be identified, it shows unknown in the spec breakdown

### DIFF
--- a/packages/catalog-realm/catalog-app/listing/listing.gts
+++ b/packages/catalog-realm/catalog-app/listing/listing.gts
@@ -289,7 +289,7 @@ class EmbeddedTemplate extends Component<typeof Listing> {
 
       <hr class='divider' />
 
-      <div class='two-col'>
+      <section class='two-col'>
         <section
           class='app-listing-categories'
           data-test-catalog-listing-embedded-categories-section
@@ -312,7 +312,7 @@ class EmbeddedTemplate extends Component<typeof Listing> {
             <p class='no-data-text'>No Tags Provided</p>
           {{/if}}
         </section>
-      </div>
+      </section>
 
       <hr class='divider' />
       <section
--- a/packages/catalog-realm/catalog-app/listing/listing.gts
+++ b/packages/catalog-realm/catalog-app/listing/listing.gts
@@ -394,7 +394,7 @@ class EmbeddedTemplate extends Component<typeof Listing> {
         width: 100%;
         height: auto;
         border-radius: var(--boxel-border-radius);
-        padding: var(--boxel-sp);
+        padding: var(--boxel-sp-sm);
         background-color: var(--boxel-100);
       }
       .info-box :deep(.markdown-content p) {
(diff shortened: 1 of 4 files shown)


packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -284,26 +289,30 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
-              <li class='categories-item'>
-                <Pill>{{category.name}}</Pill>
-              </li>
-            {{/each}}
-          </ul>
-        {{else}}
-          <p class='no-data-text'>No Categories Provided</p>
-        {{/if}}
-      </section>
+      <div class='two-col'>
```
lucaslyl: Maybe we can use <section> for the outer block (since it’s semantic block)
 and the data wrapped inside should be div
 

```
<section class="two-col">
  <div>...</div>
  <div>...</div>
 </section>
```

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -428,23 +437,27 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
-        grid-template-columns: repeat(3, 1fr);
+        display: flex;
+        flex-wrap: nowrap;
         gap: var(--boxel-sp);
         list-style: none;
-        margin-block: 0;
-        padding-inline-start: 0;
+        margin: 0;
+        padding: 0 0 var(--boxel-sp-xs) 0;
+        overflow-x: auto;
```
lucaslyl, line 448: Can you check again on this? I noticed that the images list is no longer scrollable.

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -492,6 +498,18 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
+      .two-col {
+        display: grid;
+        grid-template-columns: repeat(2, minmax(0, 1fr));
+        gap: var(--boxel-sp-xxl);
+        margin-top: var(--boxel-sp);
+        align-items: start;
+      }
+      .two-col .app-listing-categories,
+      .two-col .app-listing-tags {
+        min-width: 0;
```
lucaslyl, line 510: It looks like this block of CSS isn’t affecting the visuals — perhaps we could consider removing it.

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -284,25 +289,29 @@ class EmbeddedTemplate extends Component<typeof Listing> {
...
-              <li class='categories-item'>
-                <Pill>{{category.name}}</Pill>
-              </li>
-            {{/each}}
-          </ul>
-        {{else}}
-          <p class='no-data-text'>No Categories Provided</p>
-        {{/if}}
+      <section class='two-col'>
+        <section
```
lucaslyl, line 293: **app-listing-categories** and **app-listing-tags** can be wrapped with a <**div**> for clarity.

lucaslyl (approved):
nice

PR #3306, opened 2025-09-28
remove pauseTest in workspace delete many

tintinthong:

PR #3136, opened 2025-08-18
Multi-card delete

lukemelia:
This pull request introduces multi-card selection and bulk actions to the operator mode stack item UI, via a utility menu dropdown in CardHeader. The changes improve the user experience by allowing users to select multiple cards, perform bulk deletes, and manage selections through a dropdown menu.

* Introduced a reusable `utilityMenu` dropdown to `CardHeader`, supporting dynamic menu items and integration into operator mode stack items for selection management.
* Added bulk selection actions: "Select All", "Deselect All", and "Delete N items" to the new utility menu, with corresponding logic for confirming and performing bulk deletes, including error handling and a confirmation modal.
* Update CardHeader usage component

https://github.com/user-attachments/assets/98b14898-a2b7-48cb-bd59-e88a2e0701e6

PR #3310, opened 2025-09-29
host: Add lint rule against committing pauseTest

backspace:
This is a possibility for preventing the need for #3306. `eslint-plugin-ember` already [checks for `pauseTest`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/no-pause-test.js) but it wouldn’t have caught this unusual import path:

```
import { pauseTest } from 'ember-testing/lib/helpers/pause_test';
```

I renamed `packages/eslint-plugin-window-mock` to `packages/eslint-plugin-boxel-monorepo` to let us collect rules for the monorepo that aren’t rules for userland code, as `packages/eslint-plugin-boxel` is oriented toward.

You can see the rule producing an error [here](https://github.com/cardstack/boxel/actions/runs/18092215670/job/51475257089?pr=3310#step:15:76), there are a lot of other errors because I added `gts` files for checking, #3307 is meant to address those separately.

But is this too niche? I’ve never seen an import like this. Maybe any import from `ember-testing` could be banned, and if it‘s really needed, the usual exception can be used.


habdelra:
i thought `pauseTest()` was a global you could invoke. you don't need to import it right?

backspace:
> i thought `pauseTest()` was a global you could invoke. you don't need to import it right?

It’s `this.pauseTest()` since the general deglobalification of Ember tests, so this rule should catch both approaches.

lukemelia:
I'm not opposed, but I've also never seen that particular import

backspace:
> I'm not opposed, but I've also never seen that particular import

It was in [a PR you opened](https://github.com/cardstack/boxel/pull/3136/files#diff-27e0edd9c92c690861eabf22aa2e06f16c2db0701021cebfc369a90c980180e3R4), maybe an LLM artefact?

I’ll close this for now, I doesn’t seem like this `pauseTest` was the problem, but I think the package-renaming might still be useful in the future.

PR #3311, opened 2025-09-29
Switching to host submode should hide the ai assistant

FadhlanR:
https://github.com/user-attachments/assets/1c91a39c-5104-4e5f-ad33-af01cdd0a5b3


backspace (approved):
I tried this out in the preview deployment, works great, and the code is nice and concise 🎉

packages/host/app/services/ai-assistant-panel-service.ts
```diff
@@ -88,8 +88,14 @@ export default class AiAssistantPanelService extends Service {
     return this.commandModuleResource.module as typeof CommandModule;
   }
 
+  get hideAiAssistant() {
```
lukemelia: I prefer to name this in a pay that more clearly communicates it is a boolean and not an action. e.g. `isAiAssistantHidden`

lukemelia (approved):

PR #3312, opened 2025-09-29
Add caching for all text file downloads

IanCal:
For discussion - I added caching of text files (through the matrix service) that stops the repeated download of files for every message. We do this for card defs, so I just moved it for other files.

Does this cause issues with the way we're dealing with command defs and skills? This might clash with the re-caching side? 

It may be that this has a risk of side effects but solves a bigger problem right now and we do this while we re-work the downloading/caching?

### DIFF
--- a/packages/host/app/lib/file-def-manager.ts
+++ b/packages/host/app/lib/file-def-manager.ts
@@ -354,6 +354,13 @@ export default class FileDefManagerImpl
   }
 
   async downloadContentAsText(url: string): Promise<string> {
+    const cachedEntry = this.downloadCache.get(url);
+    if (
+      cachedEntry &&
+      Date.now() - cachedEntry.timestamp < CACHE_EXPIRATION_MS
+    ) {
+      return cachedEntry.content;
+    }
     const response = await fetch(url, {
       headers: {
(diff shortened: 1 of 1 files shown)


packages/host/app/lib/file-def-manager.ts
```diff
@@ -362,7 +369,21 @@ export default class FileDefManagerImpl
...
+    // Update cache
+    this.downloadCache.set(url, {
+      content,
+      timestamp: Date.now(),
+    });
+    // Clean up cache if it gets too large
+    if (this.downloadCache.size > 100) {
+      this.cleanupCache();
+    }
+    return content;
```
FadhlanR, line 386: I noticed that the current implementation caches the result after the download completes, but if multiple requests for the same URL happen simultaneously, they'll all start downloading in parallel instead of sharing the same promise. We need to cache the promise itself, not just the result. It looks good to me to move the downloadCache to this function.

FadhlanR: We need an option in this function to ignore the cache, because re-caching also uses this function but requires the content from the Matrix storage instead of the cache.

lukemelia:
Took these ideas and built on them in https://github.com/cardstack/boxel/pull/3321

PR #635, opened 2023-09-13
host: Add test report and failure checks

backspace:
This adds another reporter for `host` and a Github Action to surface its output as a “check” so test failures are more immediately discoverable, you can see an example failure [here](https://github.com/cardstack/boxel/runs/16762778852) when I added a deliberately-failing assertion.

boxel@a0319e5 2023-09-13 12-23-26

The action posts a [comment](https://github.com/cardstack/boxel/pull/635#issuecomment-1717978795) summarising the result of the tests. Subsequent pushes only update the comment vs posting a new one, but this can be turned off altogether if we find it annoying.

PR #3315, opened 2025-09-29
ci: Update host merging to ignore old job reports

backspace:
I noticed this when I had to rerun jobs many times, 10k+ runs is a lot:

boxel 2025-09-29 16-08-33

With the current implementation, the step that downloads the artifacts that contain host test reports gets them all, even from failed jobs that were rerun. The job report merging should only use the latest report from each job. This accomplishes that by fetching all artifact metadata and selecting only the newest one from each shard job.


backspace:
This has actually been happening [for a long time](https://github.com/cardstack/boxel/pull/2401#issuecomment-2787726786), maybe even since [the beginning](https://github.com/cardstack/boxel/pull/635#issuecomment-1717978795) of having test reports. My theory about it being reruns was incorrect, as the updated script to only download the most recent artifact didn’t change anything.

The XML test report has the total number of tests for each shard despite only running a subset of them:

```
<testsuite name="Testem Tests" tests="155" skipped="0" failures="0" timestamp="Wed Oct 01 2025 10:36:17 GMT+0000 (Coordinated Universal Time)" time="382.008">
    <testcase classname="Chrome 140.0" name="Acceptance | AI Assistant tests: attaches a card in a conversation multiple times" time="42.145"/>
```

I think that semi-explains the inflated numbers, although the math doesn’t quite work… but regardless, I haven’t found a way to get the reports to only show the count for the shard, nor an option for the merge tool to handle this case. I think it’s fine to leave; confusing, but the test failure count and link still work, which are what I look at.

PR #3332, opened 2025-09-29
Update typography & spacing theming variables

burieberry:
- move variables from `getStyleConversions()` to boxel card-container
- add ability to change typography options for themes
- add boxel css variables for line-height and font-weight

### DIFF
--- a/packages/boxel-ui/addon/src/components/card-container/index.gts
+++ b/packages/boxel-ui/addon/src/components/card-container/index.gts
@@ -8,6 +8,7 @@ interface Signature {
   Args: {
     cssImports?: string[];
     displayBoundaries?: boolean;
+    isThemed?: boolean;
     tag?: keyof HTMLElementTagNameMap;
   };
   Blocks: {
@@ -22,6 +23,7 @@ const CardContainer: TemplateOnlyComponent<Signature> = <template>
       class={{cn
         'boxel-card-container'
         boxel-card-container--boundaries=@displayBoundaries
+        boxel-card-container--themed=@isThemed
       }}
       data-test-boxel-card-container
       ...attributes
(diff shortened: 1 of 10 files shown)


packages/boxel-ui/addon/src/components/card-container/index.gts
```diff
@@ -66,6 +71,46 @@ const CardContainer: TemplateOnlyComponent<Signature> = <template>
     :global(.boxel-card-container--boundaries.hide-boundaries) {
       box-shadow: none;
     }
+
+    :global(.boxel-card-container) {
+      --theme-spacing: calc(var(--spacing) * 4);
+      /* setting boxel base css variable overrides, with boxel defaults as fallback */
+      --boxel-fontSize: var(--typescale-body, var(--_boxel-fontSize));
```
lukemelia: why is this one camelCase and everything else is dash-case?

lukemelia (approved):

packages/boxel-ui/addon/src/components/card-container/index.gts
burieberry: there's already a `--boxel-font-size` variable so I can name it `--boxel-font-size-base` if you want. Also the text-wrapping annoyed me but I can get over it.

lukemelia: If there's already --boxel-font-size then having --boxel-fontSize that is different is even more confusing. Let's find something else to name it.

burieberry: How about `--boxel-font-size-base: var(--typescale-body, var(--_boxel-font-size))`?

burieberry: or I can move line 88 up here and directly say `--boxel-font-size: var(--typescale-body, var(--_boxel-font-size))`

lukemelia: The latter seems better to me

PR #3335, opened 2025-09-30
Apply cardContainerModifier to prerendered search cards automatically

lukemelia:


tintinthong (approved):
LGTM. Catalog cards are working as expected too

PR #3338, opened 2025-09-30
Add study hub listing with catalog realm wait

backspace:
This is #3243 with a readiness check on the catalog realm.


richardhjtan:
Fixed a new lint error caused by the latest Typescript update, and now we can see that the Host Test Shard 14th has passed.

lukemelia (approved):

lucaslyl (approved):
LGTM

PR #3339, opened 2025-09-30
ci: Fix realm-waiting and exclude experiments

backspace:
`host` tests in CI use `pnpm start:skip-experiments` but that eventually results in a call to `pnpm start:development` which was actually still starting the experiments realm. This adds a flag that lets `pnpm start:development` skip starting the experiments realm.

---

Another problem is that `pnpm start:skip-experiments` was only waiting for two realms before proceeding: `localhost:4201/base/` and `localhost:4202/test/`, but was actually starting up six realms, you can see the partial indexing completion in some recent host test shard realm server logs:

Search Results (completed indexing) 2025-10-01 09-31-37

`localhost:4201/experiments/` will now no longer start but this adds the other missing three realms, as it’s my understanding that tests shouldn’t start until all realms are indexed. Some questions:
* should the catalog realm be part of the list of realms?
* is it true that every realm should be fully indexed before tests proceed?


packages/host/scripts/test-wait-for-servers.sh
```diff
@@ -1,6 +1,23 @@
...
+SKILLS_REALM_READY="$SKILLS_REALM$READY_PATH"
+TEST_REALM_READY="$TEST_REALM$READY_PATH"
+
+SYNAPSE_URL="http://localhost:8008"
+SMTP_4_DEV_URL="http://localhost:5001"
+
 NODE_NO_WARNINGS=1 start-server-and-test \
   'pnpm run wait' \
-  'http-get://localhost:4201/base/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http-get://localhost:4202/test/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http://localhost:8008|http://localhost:5001' \
+  "$BASE_REALM_READY|$CATALOG_REALM_READY|$NODE_TEST_REALM_READY|$SKILLS_REALM_READY|$TEST_REALM_READY|$SYNAPSE_URL|$SMTP_4_DEV_URL" \
```
backspace, line 22: I extracted these variables because the line would have looked like this otherwise:

```
'http-get://localhost:4201/base/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http-get://localhost:4201/catalog/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http-get://localhost:4202/node-test/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http-get://localhost:4201/skills/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http-get://localhost:4202/test/_readiness-check?acceptHeader=application%2Fvnd.api%2Bjson|http://localhost:8008|http://localhost:5001'
```

I tried to make it multiline but it didn’t work. Instead, using variables with names for each realm/server makes it more clear what we’re waiting for. If people think this is helpful/acceptable I can apply it to other files.

FadhlanR (approved):

habdelra (approved):

lukemelia (approved):

PR #3341, opened 2025-10-01
Create claimed_domains_for_sites table, add domain availability endpoint

jurgenwerk:
- Adds a new database table to support claiming custom site names (before publishing a realm)
- Adds an endpoint to check the subdomain availability `_check-site-name-availability`

These features will support claiming custom site names for published realms:

image


packages/host/config/schema/1759459363190_schema.sql
```diff
@@ -55,6 +55,14 @@
    PRIMARY KEY ( url, realm_url ) 
 );
 
+ CREATE TABLE IF NOT EXISTS claimed_domains_for_sites (
+   id DEFAULT (hex(randomblob(16))) NOT NULL,
+   hostname TEXT NOT NULL,
```
jurgenwerk, line 60: Using `hostname` for the following reason:
1. Being able to perform a join with published_realms on `source_realm_url` so that we will be able to reference the published realm
2. We can use this for another purpose - when users bring their own domains (planned future feature)

FadhlanR:
How should we handle the case where a user has already claimed a site name but reloads the page before publishing the realm? Should they be required to repeat the claiming step? Would it be better to include source_realm_url in the new table so that we can provide an endpoint to fetch the claimed site name by user ID and source realm URL?

jurgenwerk:
@FadhlanR that's a good point, I'll add user id and source realm url to the table

packages/realm-server/tests/site-name-availability-test.ts
```diff
@@ -0,0 +1,247 @@
...
+    });
+
+    async function callHandler(subdomain?: string) {
+      const ctx: any = {
+        query: subdomain !== undefined ? { subdomain } : {},
+        status: undefined,
+        body: undefined,
+        req: {
+          headers: { host: 'localhost:4200' },
+          url: '/check-site-name-availability',
```
backspace, line 44: Is this lacking the `_` because it’s wired directly to the handler? Does this approach bypass needing to run the full realm server just to test the handler?

packages/realm-server/tests/site-name-availability-test.ts
```diff
@@ -0,0 +1,247 @@
...
+      assert.strictEqual(
+        ctx.status,
+        400,
+        'Should return 400 for missing subdomain',
+      );
+      assert.ok(
+        ctx.body.includes('subdomain query parameter is required'),
+        'Should have error message about missing subdomain',
+      );
+    });
```
backspace, line 65: IMO this is a 422; 400 according to MDN:

> The reason for a 400 response is typically due to malformed request syntax, invalid request message framing, or deceptive request routing.

422:

> The HTTP 422 Unprocessable Content [client error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#client_error_responses) status code indicates that the server understood the content type of the request content, and the syntax of the request content was correct, but it was unable to process the contained instructions.

packages/realm-server/handlers/handle-check-site-name-availability.ts
```diff
@@ -0,0 +1,82 @@
...
+  const nodeEnv = process.env.NODE_ENV;
+
+  if (nodeEnv === 'production') {
+    return 'boxel.site';
+  } else if (nodeEnv === 'staging') {
+    return 'staging.boxel.build';
+  } else {
+    return 'boxel.dev.localhost';
+  }
+}
```
backspace: I’m unsure about this approach, in all other cases I’m aware of, things like this are passed to the realm server as flags when starting up. I think that’s preferable to having what’s essentially config scattered around the application behind conditionals.

`VALID_PUBLISHED_REALM_DOMAINS` exists for enumerating what domains publishing is allowed at, maybe that could be split into `PUBLISHED_SUBDOMAIN_REALM_DOMAIN` (`boxel.site`) and `PUBLISHED_SUBDIRECTORY_REALM_DOMAIN` (`boxel.space`). Or something like that?

packages/realm-server/lib/user-subdomain-validation.ts
```diff
@@ -0,0 +1,163 @@
...
+// Patterns that should be blocked
+const RESERVED_PATTERNS = [
+  /^[a-z]+-\d+$/, // service-123
+  /^[a-z]+-v\d+$/, // api-v2, service-v1
+  /^v\d+$/, // v1, v2, v3
+  /^\d+$/, // pure numbers
+  /^[a-z]+-api$/, // anything-api
+  /^[a-z]+-admin$/, // anything-admin
+  /^[a-z]+-test$/, // anything-test
+];
```
backspace, line 99: Do we need to protect against punycode/homoglyph attacks? For instance, this punycode _looks_ like “support” but would bypass this filtering:

Image

This could be revisited in the future though. I like full Unicode support because it lets people have diacritics and emoji, etc… not sure what the answer is.

backspace (changes requested):
I tried this locally and it all worked great with many variations:

Image

I do think the config thing is a blocker, though I’m open to being convinced otherwise!

packages/realm-server/tests/site-name-availability-test.ts
jurgenwerk: Yes that's right, it tests the handler directly, to make the tests quicker. This circumvents the jwt checker though, that's why I also added tests in [server-endpoints-test.ts‎](https://github.com/cardstack/boxel/pull/3341/files#diff-59859df7e8cd8f84857c077cd8f0ce4d43ebafe4af4298514ca5a60cd771391c)

packages/realm-server/tests/site-name-availability-test.ts
jurgenwerk: Makes sense, will adjust this.

packages/realm-server/handlers/handle-check-site-name-availability.ts
jurgenwerk: Good idea, will explore this direction

jurgenwerk: I split `VALID_PUBLISHED_REALM_DOMAINS` into `PUBLISHED_REALM_BOXEL_SPACE_DOMAIN` and `PUBLISHED_REALM_BOXEL_SITE_DOMAIN`

packages/realm-server/lib/user-subdomain-validation.ts
jurgenwerk: Added some basic protection against this kind of attacks

packages/realm-server/server.ts
```diff
@@ -185,7 +193,7 @@ export class RealmServer {
           realmsRootPath: this.realmsRootPath,
           getMatrixRegistrationSecret: this.getMatrixRegistrationSecret,
           createAndMountRealm: this.createAndMountRealm,
-          validPublishedRealmDomains: this.validPublishedRealmDomains,
```
backspace, line 188: An update will be needed in this file so [`host` application config](https://github.com/cardstack/boxel/pull/3343/files#diff-14c49ddb85659961b6177a036fc15d7166291f81169c9a1dd0dfead082ef68e5) can know about the correct host for subdirectory published realms, which is why the Matrix test from that PR is failing.

jurgenwerk: My idea in the most recent commits is to get rid of `config.defaultPublishedRealmDomain` and use `publishedRealmBoxelSpaceDomain` for that case. That way we can only keep track of two env vars, `publishedRealmBoxelSpaceDomain` and `publishedRealmBoxelSiteDomain`. Are you ok with this?

jurgenwerk:
@backspace I think this is ready for another review, the failed tests looks like they are unrelated to this PR.

packages/realm-server/server.ts
backspace: yes, although I think I (and maybe others?) will be confused about `space` vs `site`, I know those are the official names, I just struggle to differentiate them and have been thinking of them as subdirectory and subdomain. But maybe it’s just me!

backspace:
> @backspace I think this is ready for another review, the failed tests looks like they are unrelated to this PR.

👍🏻 I’ll return to this in a couple of hours

packages/realm-server/server.ts
jurgenwerk: I chose that nomenclature because the subdirectory realm also comes with a subdomain, so I thought naming these by their features could be a little clearer. But I agree it's confusing, possibly either way :S I think if I add some comments it will help. Will go and do that

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -133,7 +133,11 @@ export default class PublishRealmModal extends Component<Signature> {
   }
 
   private getDefaultPublishedRealmDomain(): string {
-    // TODO: adjust this after we add the feature for publishing with a custom boxel site name
+    // publishedRealmBoxelSpaceDomain is the domain that is used to form urls like "mike.boxel.space/game-mechanics"
+    // which are used to create Boxel Spaces (we will also have Boxel Sites, which is a different published realm)
+
+    // TODO: since we currently only have Boxel Spaces, we can default to that domain. When we add Boxel Sites,
+    // adjust this component to know which published realm domain to use.
```
backspace, line 140: 👍🏻 that’s what I’m working on at the moment 😀

backspace:
Does the Matrix test fail for you locally? I doesn’t seem directly related to this but I’m surprised it would fail both retries.

jurgenwerk:
Looks like it works. It's a super flaky test.

.github/workflows/build-host.yml
```diff
@@ -30,14 +30,16 @@ jobs:
...
+            echo "PUBLISHED_REALM_BOXEL_SPACE_DOMAIN=boxel.space" >> $GITHUB_ENV
+            echo "PUBLISHED_REALM_BOXEL_SITE_DOMAIN=boxel.site" >> $GITHUB_ENV
           elif [ "$INPUT_ENVIRONMENT" = "staging" ]; then
             echo "RESOLVED_BASE_REALM_URL=https://realms-staging.stack.cards/base/" >> $GITHUB_ENV
             echo "RESOLVED_SKILLS_REALM_URL=https://realms-staging.stack.cards/skills/" >> $GITHUB_ENV
             echo "MATRIX_URL=https://matrix-staging.stack.cards" >> $GITHUB_ENV
             echo "MATRIX_SERVER_NAME=stack.cards" >> $GITHUB_ENV
             echo "EXPERIMENTAL_AI_ENABLED=true" >> $GITHUB_ENV
-            echo "DEFAULT_PUBLISHED_REALM_DOMAIN=staging.boxel.build" >> $GITHUB_ENV
+            echo "PUBLISHED_REALM_BOXEL_SPACE_DOMAIN=staging.boxel.build" >> $GITHUB_ENV
```
backspace: Maybe it’s my mistake but I think this should be `staging.boxel.dev`:

Image

packages/realm-server/lib/user-subdomain-validation.ts
```diff
@@ -0,0 +1,187 @@
...
+    };
+  }
+
+  // Check for non-ASCII characters (homoglyph attack protection)
+  if (!/^[\x20-\x7E]*$/.test(subdomain)) {
+    return {
+      valid: false,
+      error: 'Subdomain can only contain printable ASCII characters',
+    };
+  }
```
backspace: Isn’t this redundant with the `validSubdomainRegex` already excluding these?

backspace (approved):

jurgenwerk:
I'm weirded out by this [failing percy test](https://percy.io/Cardstack-1/web/-cardstack-host/builds/43668197/changed/2317198942?browser=firefox&browser_ids=69%2C70&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=1280&widths=1280):

image

backspace:
> I'm weirded out by this [failing percy test](https://percy.io/Cardstack-1/web/-cardstack-host/builds/43668197/changed/2317198942?browser=firefox&browser_ids=69%2C70&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=1280&widths=1280):
> 
> image

ya that’s pretty drastic, my guess is that opening the card failed on `main` or something… a mess

.github/workflows/build-host.yml
jurgenwerk: Fixed 👌

packages/realm-server/lib/user-subdomain-validation.ts
jurgenwerk: Right!

PR #3343, opened 2025-10-01
server: Restore publishability check before publishing

backspace:
This restores the publishability check to `POST /_publish-realm` with a Matrix test to exercise it, you can see the test failing [here](https://boxel-matrix-playwright-reports.stack.cards/server/restore-publishability-check-cs-9468/20251001_134736Z/index.html#?testId=66a8561acd7006dadeb0-98bb66392ea58d902f77) when I brought the check back without fixing the permissions.

### DIFF
--- a/packages/realm-server/server.ts
+++ b/packages/realm-server/server.ts
@@ -293,8 +293,14 @@ export class RealmServer {
       /(<meta name="@cardstack\/host\/config\/environment" content=")([^"].*)(">)/,
       (_match, g1, g2, g3) => {
         let config = JSON.parse(decodeURIComponent(g2));
+
+        if (config.defaultPublishedRealmDomain === 'localhost:4201') {
+          // if this is the default, this needs to be the realm server’s host
+          // to work in Matrix tests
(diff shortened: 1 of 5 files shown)


packages/realm-server/server.ts
```diff
@@ -293,6 +293,13 @@ export class RealmServer {
       /(<meta name="@cardstack\/host\/config\/environment" content=")([^"].*)(">)/,
       (_match, g1, g2, g3) => {
         let config = JSON.parse(decodeURIComponent(g2));
+
+        if (config.defaultPublishedRealmDomain === 'localhost:4201') {
```
backspace, line 297: Without this the Matrix test was publishing the realm to `localhost:4201`. Just overriding without checking whether it’s the default means staging and production domains will be lost.

Another approach to this would be to have this be part of realm server configuration, if that seems better, I’m happy to update to that!

jurgenwerk:
@codex can you review this PR?

jurgenwerk (approved):

PR #3345, opened 2025-10-01
AI Bot Locking - Avoid re-processing already handled events

lukemelia:
- Added a new migration (packages/postgres/migrations/1759000000000_add-completed-at-to-ai-bot-event-processing.js)
  introducing completed_at TIMESTAMPTZ
  - Hardened the lock helpers (packages/ai-bot/lib/queries.ts:11) so that releaseLock timestamps the row instead of deleting it.
  - Wrapped the event handling pipeline in a try…finally to ensure every exit path releases the lock (packages/ai-bot/
  main.ts:289)
  - Added tests covering the new locking behaviour (packages/
  ai-bot/tests/locking-test.ts)


packages/ai-bot/lib/queries.ts
```diff
@@ -8,18 +8,20 @@ import {
...
-
 export async function acquireLock(
   pgAdapter: PgAdapter,
   eventId: string,
   aiBotInstanceId: string,
 ): Promise<boolean> {
+  let existingLockRows = await query(pgAdapter, [
+    `SELECT 1 FROM ai_bot_event_processing WHERE event_id_being_processed = `,
+    param(eventId),
+  ]);
```
backspace, line 19: This doesn’t need `WHERE completed_at IS NULL`?

lukemelia: No, because we want to skip the event if it another process is working on it but has not completed it OR if any process worked on it and completed it.

packages/ai-bot/lib/queries.ts
```diff
@@ -8,18 +8,20 @@ import {
...
-    param(eventId),
-  ]);
-}
-
 export async function acquireLock(
   pgAdapter: PgAdapter,
   eventId: string,
   aiBotInstanceId: string,
 ): Promise<boolean> {
+  let existingLockRows = await query(pgAdapter, [
```
IanCal, line 16: This is a two step process:

Check if being processed
(time passes)
Set as being processed 

Could this be collapsed into one query? Could we use the `ON CONFLICT` to return something where there's a clash? Right now with two steps you can have the following process

Bot A -> checks if event E is being processed, no
Bot B -> same

Bot A -> writes that it is being processed
Bot B -> same

Both think they're the only one and start running.

packages/postgres/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "es2020",
     "allowJs": true,
-    "module": "CommonJS",
+    "module": "NodeNext",
```
IanCal, line 5: Is this an intended change as part of this pr?

packages/ai-bot/lib/queries.ts
backspace: ah understood, thanks, will review now

packages/host/config/schema/1759000000000_schema.sql
```diff
@@ -0,0 +1,88 @@
...
+   write BOOLEAN NOT NULL,
+   realm_owner BOOLEAN DEFAULT false NOT NULL,
+   PRIMARY KEY ( realm_url, username ) 
+);
+
+ CREATE TABLE IF NOT EXISTS realm_versions (
+   realm_url TEXT NOT NULL,
+   current_version INTEGER NOT NULL,
+   PRIMARY KEY ( realm_url ) 
+);
```
backspace, line 88: Does the other schema file need to be deleted? I think we should automate this and/or lint against there being two.

backspace (approved):

packages/postgres/tsconfig.json
lukemelia: Yes, pnpm make-schema doesn't run without it

packages/ai-bot/lib/queries.ts
lukemelia: Thanks, I'll tackle this separately.

lukemelia: https://github.com/cardstack/boxel/pull/3351

PR #3348, opened 2025-10-02
Ensure user record exists for default development users

jurgenwerk:
After landing https://github.com/cardstack/boxel/pull/3260, developers who use our default development user started experiencing the error attached below. This is because our development user setup lacked inserting a record to the users table, which is needed for the bulk realms login. This PR adds that to the `register-test-user` script.

image

### DIFF
--- a/packages/host/app/services/realm-server.ts
+++ b/packages/host/app/services/realm-server.ts
@@ -20,6 +20,8 @@ import { RealmAuthClient } from '@cardstack/runtime-common/realm-auth-client';
 
 import ENV from '@cardstack/host/config/environment';
 
+import config from '@cardstack/host/config/environment';
+
 import RealmService from './realm';
 
 import type { ExtendedClient } from './matrix-sdk-loader';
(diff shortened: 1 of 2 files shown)


packages/host/app/services/realm-server.ts
```diff
@@ -197,8 +199,19 @@ export default class RealmServerService extends Service {
     });
 
     if (!response.ok) {
+      let responseText = await response.text();
+
+      // Temporary development instruction to help with user setup
+      let isDevelopment = config.environment === 'development';
+      if (isDevelopment && responseText.includes('User in JWT not found')) {
+        console.error(
+          '\x1b[1m\x1b[31m%s\x1b[0m',
```
jurgenwerk, line 208: Prints an emphasized error message. This is just temporary until all devs run `pnpm register-all` again.

Image

backspace (approved):

lukemelia (approved):

PR #3349, opened 2025-10-02
Add meta info to serialized instance from /render route

habdelra:
This PR adds meta info to the serialized card instance. The info being added is really external to the card and hence the prerenderer has no way of knowing it. moreover, deriving this additional meta requires access to the filesystem/DB which is not permitted in the prerenderer's security context. So the idea is that we patch this additional meta into the prerender result from within the worker after we receive it from the prerenderer


habdelra:
@codex pls review

lukemelia (approved):

PR #3351, opened 2025-10-02
Collapse SQL queries in acquireLock function to a single one

lukemelia:


IanCal (approved):
LGTM - I was wrong about there being a risk of double processing things (the on conflict do nothing / returning handles it entirely) on the last PR but I think this is cleaner and removes a query.

PR #3355, opened 2025-10-03
update readme to configure proxy endpoint locally

tintinthong:


lucaslyl (approved):

jurgenwerk (approved):

FadhlanR (approved):
Thanks for this PR. I missed it when move the config to db.

PR #3356, opened 2025-10-03
Implement open site button

FadhlanR:
https://github.com/user-attachments/assets/8f79046a-2a16-4af6-9fd0-2217d8221ac9


packages/host/tests/acceptance/host-submode-test.gts
```diff
@@ -395,6 +410,101 @@ module('Acceptance | host submode', function (hooks) {
...
+          assert.strictEqual(target, '_blank', 'Open in a new tab');
+          return null;
+        };
+        await click(
+          '[data-test-published-realm-item="https://another-domain.com/realm/Person/1"] [data-test-open-site-button]',
+        );
+        window.open = originalWindowOpen;
+
+        getService('network').resetState();
+      });
```
backspace, line 531: Instead of using mocks I’d suggest adding to the `publish-realm.spec` Playwright test, as that’s more full-stack-ish… [the PR isn’t merged yet](https://github.com/cardstack/boxel/pull/3343/files#diff-79929f4b4fa0b4adc06b19a1b8ddcc69f6bb57b4be8b5b71e6bfccac0e267d31) as I’m unsure it‘s been reviewed…? But you can see there that I have it use the Open Site button in the modal, and then the test finds the new tab and asserts on it. I think your test could operate similarly

packages/matrix/tests/publish-realm.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    await page.locator('[data-test-open-site-button]').click({
+      modifiers: ['Shift'],
+    });
+
+    let popup = await popupPromise;
+    expect(popup).toBeNull();
+
+    await expect(page.locator('[data-test-open-site-popover]')).toBeVisible();
+
+    let newTabPromise1 = page.waitForEvent('popup');
```
backspace: Could `newTabPromise` just be reused here? Or if another variable is needed, maybe they should have distinct names.

Same for `newTab` below.

packages/host/tests/acceptance/host-submode-test.gts
backspace: Did you see my comment here? I think the Matrix test is a better place for this because it doesn’t need to mock `window`, it’s more like a real browser. The only thing I’d add is an assertion that the Open Site button doesn’t exist when a realm hasn’t been published.

backspace (approved):
I tried this locally and it worked as expected 🎉 I do wonder about how people are supposed to know about this shift-clicking but that’s not your concern. This will become more useful when we can publish on custom subdomains.

backspace:
oh I didn’t see that your Matrix test is failing, two elements with the same selector 🤔 so that should be addressed before merging

packages/matrix/tests/publish-realm.spec.ts
FadhlanR: You're right, I updated it to use the same variable.

FadhlanR:
> I tried this locally and it worked as expected 🎉 I do wonder about how people are supposed to know about this shift-clicking but that’s not your concern. This will become more useful when we can publish on custom subdomains.

I added tooltip so there is an info to use Shift+Click to open the options.

PR #3365, opened 2025-10-07
Persist session rooms in DB instead of matrix accountData

lukemelia:


packages/realm-server/node-realm.ts
```diff
@@ -247,6 +257,24 @@ export class NodeAdapter implements RealmAdapter {
...
       return;
     }
+    let roomsFromDB = await fetchAllSessionRooms(dbAdapter, realmUrl);
+
+    // verify that roomFromDB and dmRooms are the same - TODO: remove when we cutover to DB version only
+    for (let userId of Object.keys(roomsFromDB)) {
+      let roomId = roomsFromDB[userId];
+      if (Object.keys(dmRooms).length !== Object.keys(roomsFromDB).length) {
+        console.log({ dmRooms, roomsFromDB });
+        throw new Error(
```
IanCal: We might not want to throw an error here, though definitely logging it clearly - the process for adding rooms to the account data right now does three async steps:

Get the account data, maybe create a dm room, write updated account data

If there are two concurrent requests, iiuc these could interleave and there would be a race condition

req A -> get state = a = {} 
req B -> get state = b = {} 
req A -> create room, add to dict a {user1: room1} 
req A -> write updated state {user1: room1} 
req B -> create room, add to dict b {user2: room2} 
req B -> write updated state {user2: room2}

 Result = lost the {user1: room1} mapping?

I may be wrong about how async and nodes single thread interact though. If not this PR fixes the problem by not having a single map we're updating, so inconsistencies might be good.

lukemelia: True. I intend to drop the accountData code and the comparison code before merging. The side by side approach is to establish that functionality has not changed as tests run.

lukemelia:
@codex review

packages/realm-server/handlers/handle-realm-auth.ts
```diff
@@ -30,26 +30,24 @@ export default function handleRealmAuth({
...
+        user.matrixUserId,
+      );
       sessions[realm] = createJWT(
         {
           user: matrixUserId,
           realm: realm,
           permissions,
-          sessionRoom: sessionRoomId,
+          sessionRoom: sessionRoom!,
         },
```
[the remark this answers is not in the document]
lukemelia: Fixed.

packages/postgres/scripts/convert-to-sqlite.ts
```diff
@@ -107,33 +107,50 @@ function createColumns(
...
+              column.push('DEFAULT', 'CURRENT_TIMESTAMP');
+              break;
+            }
+          }
+
+          throw new Error(
+            `Don't know how to serialize default value constraint for expression type '${constraint.expr.type}'`,
+          );
+        }
+        default: {
```
habdelra, line 149: nice--thanks!

habdelra (approved):

FadhlanR (approved):

PR #3366, opened 2025-10-07
Include Missing Spec to Avatar Listing

lucaslyl:


tintinthong:
@lucaslyl can you add 3 instances of avatars that can be shown straight away  after remix?

lucaslyl:
> @lucaslyl can you add 3 instances of avatars that can be shown straight away after remix?

image
ya, I’ve already added the example with the extra three instances. After you remix, you should be able to see those instances (like screenshot).

tintinthong (approved):

PR #3370, opened 2025-10-07
host: Add subdomain claim validation

backspace:
This uses the `GET /_check-site-name-availability` realm server endpoint in the publish realm modal to validate prospective custom subdomains. Since claiming doesn’t yet exist, it stops there.

screencast 2025-10-09 17-43-42

I decided to use the existing `InputGroup` accessory/validation vs adding yet another custom solution to exactly match the design:

Publish modal flow - Publish to - Zeplin 2025-10-09 17-45-45

The `domain-option` containers now use a grid to have consistent spacing and alignment.

I created a `host-submode` directory within `operator-mode` and moved existing components in there. However, I was unable to move `publish-realm-modal` into it without it being interpreted as a whole new file, so I’ll do that after PR approval 😬


packages/host/app/services/realm-server.ts
```diff
@@ -503,6 +509,32 @@ export default class RealmServerService extends Service {
...
+    let url = new URL(`${this.url.href}_check-site-name-availability`);
+    url.searchParams.set('subdomain', subdomain);
+
+    let response = await this.network.fetch(url.href, {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        Authorization: `Bearer ${this.token}`,
+      },
+    });
```
backspace, line 549: I tried `authedFetch` but the token header wasn’t included… what is that function for then?

lukemelia: Hmm, not sure for realm-server-level endpoints. I think we should make authedFetch work

lukemelia (approved):

jurgenwerk:
I just realized that after merging https://github.com/cardstack/boxel/pull/3402, the validation endpoint got renamed, which breaks the implementation of this PR. I'll make changes to this PR (since @backspace is absent)

jurgenwerk (approved):

packages/host/app/services/realm-server.ts
jurgenwerk: `this.realmServer.authedFetch` works

jurgenwerk: Adjusted this in https://github.com/cardstack/boxel/pull/3370/commits/a337312f3d519bd7571371cb8c89f9e857772041

jurgenwerk (approved):

PR #3371, opened 2025-10-07
wip time machine

tintinthong:


packages/experiments-realm/components/lightbox-carousel.gts
richardhjtan, line 1: Is it intentional to add the cards to the experiment realm?

packages/catalog-realm/components/polaroid.gts
```diff
@@ -0,0 +1,95 @@
...
+    caption?: string;
+  };
+}
+
+export default class Polaroid extends GlimmerComponent<PolaroidSignature> {
+  <template>
+    <div class='polaroid'>
+      <div class='polaroid-photo'>
+        {{#if (has-block 'content')}}
+          {{yield to='content'}}
```
lucaslyl, line 19: I think the yield pattern should add Block: [...] to the PolaroidSignature interface, right?

packages/catalog-realm/components/polaroid.gts
```diff
@@ -0,0 +1,95 @@
...
+        {{else if @base64}}
+          <img
+            src='data:image/*;base64={{@base64}}'
+            alt={{@caption}}
+            class='generated-image'
+          />
+        {{/if}}
+        {{#if (has-block 'loading')}}
+          {{yield to='loading'}}
```
lucaslyl, line 28: same for this

lucaslyl (commented):
Overall, it’s cool! 
Just need to fix the yield block types, then it’s okay to merge already.

packages/catalog-realm/time-machine/polaroid-image.gts
```diff
@@ -0,0 +1,41 @@
...
+  <template>
+    <Polaroid @caption={{this.caption}} @base64={{@model.image.data.base64}} />
+  </template>
+}
+
+export class PolaroidImage extends CardDef {
+  static displayName = 'Polaroid Image';
+
+  @field caption = contains(StringField);
+  @field image = linksTo(() => ImageCard);
```
lucaslyl: How about if we PolaroidImage just extend the ImageCard ,so we dont need to handle 2 instances always

PR #3372, opened 2025-10-08
Integrate prerender server in worker (node context) indexing

habdelra:
This PR integrates the prerender server into our worker (node context) to facilitate headless chrome indexing in our running app in the dev env. As part of this I also fixed a bunch of fundamental issues that blocked the headless chrome from working in our indexer including session issues and /render route model hook loading. 

We still have outstanding work to port the server indexing tests, and we need to investigate all the indexing errors when indexing catalog realm and experiments realm to ensure we are not seeing any new types of issues.. But at the very least this will allow us to try out the new headless chrome indexing locally. Use the env var `USE_HEADLESS_CHROME_INDEXING=true` when running `start:all` to enable the feature flag that uses headless chrome for indexing.


packages/host/app/services/loader-service.ts
```diff
@@ -72,7 +75,15 @@ export default class LoaderService extends Service {
...
-      if (this.isIndexing) {
+      if (
+        this.isIndexing &&
+        // we are selective about adding this header because external sites,
+        // like esm.run will reject our custom header because of CORS. This hack
+        // is not sustainable, once we have more than 1 realm server or custom
+        // realm CNAMEs this won't work anymore...
+        (req.url.startsWith(this.realmServer.url.href) ||
+          req.url.startsWith(baseRealm.url))
+      ) {
```
habdelra: FYI: we need a better solution here....

packages/host/app/services/network.ts
```diff
@@ -61,7 +63,15 @@ export default class NetworkService extends Service {
...
-        if (this.loaderService.isIndexing) {
+        if (
+          this.loaderService.isIndexing &&
+          // we are selective about adding this header because external sites,
+          // like esm.run will reject our custom header because of CORS. This
+          // hack is not sustainable, once we have more than 1 realm server or
+          // custom realm CNAMEs this won't work anymore...
+          (req.url.startsWith(this.realmServer.url.href) ||
+            req.url.startsWith(baseRealm.url))
+        ) {
```
habdelra: ditto

packages/host/app/router.ts
```diff
@@ -10,7 +10,7 @@ Router.map(function () {
   this.route('host-freestyle', { path: '/_freestyle' });
   this.route('indexer', { path: '/indexer/:id' });
   this.route('render-error', { path: '/render-error/:reason' });
-  this.route('render', { path: '/render/:id' }, function () {
+  this.route('render', { path: '/render/:id/:nonce' }, function () {
```
habdelra, line 13: We add a nonce, otherwise repeated visits to the same instance will not trigger the model hook to re-run in the render route (e.g. auto save)

habdelra:
@codex review

lukemelia (approved):

PR #3373, opened 2025-10-08
host: Update Monaco to remove patch

backspace:


backspace:
The patch is [still needed](https://github.com/microsoft/monaco-editor/issues/4702#issuecomment-3382083776) sadly, it doesn’t seem worth updating, although the removal of the `window.monaco` global is something we should be aware of, I think the approach here is salvageable.

PR #3378, opened 2025-10-08
Fix playground dropdown menu ui

burieberry:
Fixes:
- dropdown trigger border-radius
- cropped dropdown after-items menu
- scroll in main dropdown items menu
- selected item hover background
- dropdown item overflow
- remove usage of !important keywords
- mismatched widths of selected and unselected dropdown items
- extra padding in attach-button dropdown
- keep items full width if selection state is not shown
- created tickets for menu ui issues (cs-9507), dropdown/select theme coloring fixes (cs-9513), and better way to maintain and provide theme variables to wormhole components (cs-9509)

Before:
before

After:
instance-chooser

### DIFF
--- a/packages/boxel-ui/addon/src/components/select/index.gts
+++ b/packages/boxel-ui/addon/src/components/select/index.gts
@@ -258,7 +258,7 @@ export default class BoxelSelect<ItemT> extends Component<Signature<ItemT>> {
         );
         --select-placeholder-color: var(
           --boxel-select-placeholder-color,
-          var(--foreground, var(--boxel-450))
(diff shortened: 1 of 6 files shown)


backspace:
In the preview deployment I still have overflow 😳

blog-category gts in buck 2025-10-09 16-07-39

burieberry:
> In the preview deployment I still have overflow 😳
> 

Ah a last minute change broke it again. It should be good now. It was a boxel-ui fix so not sure if preview will update.

backspace (approved):
ya `boxel-ui` builds as part of the preview, so the fix is there, looks good 🎉

Image

PR #3380, opened 2025-10-09
Add GET /_boxel-site-hostname

jurgenwerk:
Will be used to show this domain, when the realm is not published yet but the domain has been claimed:

image


packages/realm-server/tests/get-boxel-site-hostname-test.ts
```diff
@@ -0,0 +1,236 @@
...
+    hooks.beforeEach(async function () {
+      prepareTestDB();
+      dbAdapter = new PgAdapter({ autoMigrate: true });
+      handler = handleGetBoxelSiteHostnameRequest({
+        dbAdapter,
+        domainsForPublishedRealms: { boxelSite: boxelSiteDomain },
+      } as any);
+
+      // Clean up any existing claims to ensure clean state
+      await query(dbAdapter, ['DELETE FROM claimed_domains_for_sites']);
```
backspace: Doesn’t `prepareTestDB` produce an empty database, is this needed?

packages/realm-server/tests/server-endpoints-test.ts
```diff
@@ -1679,6 +1682,99 @@ module(basename(__filename), function () {
...
+            response.body.data.attributes.subdomain,
+            'my-site',
+            'Should return correct subdomain',
+          );
+          assert.strictEqual(
+            response.body.data.attributes.sourceRealmURL,
+            sourceRealmURL,
+            'Should return correct source realm URL',
+          );
+        });
```
backspace, line 1745: `assert.deepEqual(response.body.data.attributes, {…` would be more concise IMO, this is a matter of taste of course

backspace (approved):

packages/realm-server/tests/get-boxel-site-hostname-test.ts
jurgenwerk: Yep, it's not needed. I removed it

PR #3382, opened 2025-10-09
Workspace header bug

burieberry:
Fixes:
- keep header without background, let the submode/workspace container set the background
- rename `top-bar` class, used with `:deep` selector (it collided with a card template)
- reduce padding on workspace chooser

Before (when opening from code-mode):
bef

After:
after

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode.gts
+++ b/packages/host/app/components/operator-mode/code-submode.gts
@@ -821,6 +821,7 @@ export default class CodeSubmode extends Component<Signature> {
 
     <style scoped>
       :global(:root) {
+        --code-submode-background: #74707d;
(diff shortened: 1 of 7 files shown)


richardhjtan:
Screenshot 2025-10-13 at 9 56 30 PM

Something that has been catching my eye when we focus on the header - the logo and text aren’t vertically centered. Perhaps we can ticket up if it is not intended

richardhjtan (approved):
Looks good to me

burieberry:
> Something that has been catching my eye when we focus on the header - the logo and text aren’t vertically centered. Perhaps we can ticket up if it is not intended

Good point. We should also use an image file for the logo because the font always loads late. I'll create a ticket

PR #3386, opened 2025-10-10
Fix `_catalog-realms` endpoint to exclude published realms

FadhlanR:
Previously, we treated a realm with publicly readable permission as a catalog realm. However, this criterion is no longer correct, since published realms are also publicly readable, but we don’t want to include them as catalog realms. In this PR, I updated the query used to retrieve catalog realms.


tintinthong (approved):
yes tq so much

backspace (approved):

richardhjtan (approved):

PR #3387, opened 2025-10-10
Apply theme to beat maker

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-399/update-beat-maker-with-chris-changes


packages/catalog-realm/beat-maker/beat-maker.gts
```diff
@@ -1573,20 +1598,24 @@ class BeatMakerIsolated extends Component<typeof BeatMakerCard> {
...
-        background: #374151;
-        border: 1px solid #4b5563;
-        color: white;
-        padding: 0.375rem 0.75rem;
-        border-radius: 6px;
+        background: var(--card, #374151);
+        border: 1px solid var(--border, #4b5563);
+        color: var(--card-foreground, #ffffff);
+        padding: var(--spacing, 0.375rem) calc(var(--spacing, 0.25rem) * 3);
+        border-radius: var(--radius-sm, 6px);
```
burieberry: Shadcn standard variables only offer `--radius`. But if you use any boxel border radius variables, it will be calculated based on theme radius. You can write it as a fallback if your theme has `--radius-sm`: `border-radius: var(--radius-sm, var(--boxel-border-radius-sm);`

tintinthong (approved):

PR #3389, opened 2025-10-10
host: Remove realm subscription in host mode

backspace:
This prevents spurious errors like this:

Ramped 2025-10-10 14-07-48

Maybe we’ll want subscription in some form in the future but there’s no facility for it at the moment.

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -61,6 +61,7 @@ import { type CardSaveSubscriber } from './card-service';
 import EnvironmentService from './environment-service';
 
 import type CardService from './card-service';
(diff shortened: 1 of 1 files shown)


lukemelia (approved):
It seems like it might be better to not call `subscribeToRealm` if in host mode, but this is fine if it is the best place.

jurgenwerk (approved):

backspace:
> It seems like it might be better to not call `subscribeToRealm` if in host mode, but this is fine if it is the best place.

bm3 2025-10-20 11-23-35

These are the three places it’s called, it seems cleaner to me to make the call a no-op vs adding the check amid all this other code, but I’m open to being wrong!

PR #3397, opened 2025-10-13
Remove base64 entries of polaroid image

tintinthong:


richardhjtan (approved):

tintinthong:
I do not think base64 is the thing that is tanking the full run indexing for catalog. I think its more related to loading a link inside a computed. My browser ran out of memory locally while trying to debug on a single card

The timeouts are just a coincidence -- other cards are also timing out -- we should handle the Javascript Heap error

tintinthong:
after eliminating the other errors. I do actually now believe that its these image cards. Its the ones causing javascript heap limit errors

PR #3399, opened 2025-10-13
Create test to ensure catalog index card renders with no errors

tintinthong:
Very typical is just bad linking of catalog cards. This test ensures the index card renders with no errors


richardhjtan (commented):
Maybe we can say `catalog-realm-test`?

richardhjtan (approved):

PR #3404, opened 2025-10-13
Add a system card concept

IanCal:
The system card defines which LLMs a user can pick between, and will be the place for more things to be added.

There are default system cards (one for each env). Users can set one system card to be the one use for them in their account data.

Notes:
* There's a host command to set a system card in the users account data
* System cards have an addition in the three dot menu to call this host command so you can set a 
* The list of models in the dropdown is anything used in the room already *plus* anything in the users system card
* Default model is the first one in the list
* Default code model is still hard coded, this will be managed later with shortcuts 
* Model configs added with descriptions pulled once from openrouter
* System card and model config are intentionally bare, these will be expanded more later as part of the shortcuts work
* LLM is still set by ID in the rooms. This will have to change but is out of scope here
* The flag for whether tools are usable by the model isn't used by the ai bot but is one of the most important parts to implement there as right now new models typically don't work properly until the ai bot resets and re-pulls this info.


IanCal:
Some of the tests I'm very confused about.

The scrolling test works on main but not this branch, there is one additional message seen as read, and so the scrolling goes to a different point and fails the test. I've not yet been able to see any reason why this should be the case from the changes, and is reliably failing on CI and locally as the scrolling is to a slightly different point.

IanCal:
Scrolling tests fixed, now there are some catalog app listing tests failing, I'll look at those.

packages/ai-bot/main.ts
```diff
@@ -152,7 +152,7 @@ class Assistant {
   }
 
   // TODO: This function is used to avoid a thinking model of gpt-5.
-  // Remove this function after we have LLM environment.
+  // Remove this function after we have system card.
```
lukemelia: Doesn't this PR introduce a system card? Does this comment need updating?

packages/base/system-card.gts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export class SystemCard extends CardDef {
+  static displayName = 'System Card';
+
+  @field modelConfigurations = linksToMany(ModelConfiguration, {
+    description: 'List of available model configurations for this system',
+  });
+
+  [getCardMenuItems](params: GetCardMenuItemParams): MenuItemOptions[] {
```
lukemelia, line 29: Cool to see this API being useful!

packages/host/tests/helpers/mock-matrix/_client.ts
```diff
@@ -551,6 +562,8 @@ export class MockClient implements ExtendedClient {
   private eventHandlerType(type: string) {
     switch (type) {
       case APP_BOXEL_REALMS_EVENT_TYPE:
+      case APP_BOXEL_SYSTEM_CARD_EVENT_TYPE:
+      case 'm.direct':
```
lukemelia, line 566: Is m.direct related to this PR?

lukemelia (approved):
Is there a script to generate the ModelConfiguration cards and/or documentation about what to do when a new model becomes available?

IanCal:
> Is there a script to generate the ModelConfiguration cards and/or documentation about what to do when a new model becomes available?

No, that's a good point. I'll add docs now in this PR, and add a ticket for a script to make new ones to not block the work further. Since I'm taking the descriptions from openrouter, we could have a script that just takes in the openrouter model id and generates a basic model card.

packages/base/system-card.gts
IanCal: it was really nice, I didn't want to have a custom UI just to add it and this felt like a more natural place for it to live.

packages/ai-bot/main.ts
IanCal: I'll update it, this part still needs to exist but a followup part of system card work will remove it (when we can pass over more settings than just the model id we can have reasoning effort passed over rather than hard coded)

packages/host/tests/helpers/mock-matrix/_client.ts
IanCal: Tests failed with an error without it, it's to do with setting account data triggering an event in the mocks now and there's an event that has the DM rooms in it iirc. So it's not an event added in this PR but a change in the mocks meant it had to be handled.

PR #3418, opened 2025-10-15
Ensure user joins the session room

FadhlanR:


packages/host/app/services/realm-server.ts
```diff
@@ -248,10 +248,37 @@ export default class RealmServerService extends Service {
...
+    let directRooms = await this.client.getAccountDataFromServer('m.direct');
+    let userId = this.client.getUserId();
+    if (
+      userId &&
+      (!directRooms?.[userId] || !directRooms[userId].includes(sessionRoom))
+    ) {
+      await this.client.setAccountData('m.direct', {
+        [userId]: [...(directRooms?.[userId] ?? []), sessionRoom],
+      });
+    }
```
lukemelia: Is this part necessary? What happens if we skip it?

FadhlanR: In sliding sync, we distinguish between AI rooms and auth rooms (session rooms) based on whether a room is a DM or not. Unfortunately, the is_dm flag we set when creating a room isn’t enough to make that filter work — we also need to add the room to the m.direct account data so it can be identified as a DM room.

PR #3419, opened 2025-10-15
Improved timeout logic to better handle unsuccessful fetch of instance and adding perf logging

habdelra:
This PR ads perf logging to the prerender server as well as to fix an issue where unsuccessful fetch of instance in the render route would only be captured by render timeout. Now if there is a non-200 HTTP response (e.g. instance using a broken template) we capture that error immediately.


habdelra:
@codex review

lukemelia (approved):

PR #3423, opened 2025-10-16
Fix gc store error for cyclic dependencies

tintinthong:
Garbage collection breaks when 2 friends link to each other but neither are saved

Screenshot 2025-10-16 at 09 21 32


tintinthong:
@codex review

habdelra (approved):

PR #3426, opened 2025-10-16
Restore nested open directories when root is reopened

FadhlanR:
https://github.com/user-attachments/assets/48033aa6-a867-4091-89cb-47550d52ced3

### DIFF
--- a/packages/host/app/components/editor/directory.gts
+++ b/packages/host/app/components/editor/directory.gts
@@ -154,27 +154,15 @@ export default class Directory extends Component<Args> {
 
   @action
   private selectDirectory(entryPath: LocalPath) {
-    // Ensure entryPath ends with '/' for directories
-    const dirPath = entryPath.endsWith('/') ? entryPath : `${entryPath}/`;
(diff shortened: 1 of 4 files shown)


habdelra:
@codex review

packages/host/app/components/editor/directory.gts
```diff
@@ -187,8 +175,13 @@ export default class Directory extends Component<Args> {
...
-      ? this.args.openDirs.find((item) => item.startsWith(entryPath))
-      : this.openDirs.find((item) => item.startsWith(entryPath));
+    let dirPath = this.normalizeDirPath(entryPath);
+    let openDirs = this.args.openDirs ?? this.openDirs;
+
+    return openDirs.includes(dirPath);
+  }
+
+  private normalizeDirPath(entryPath: LocalPath): LocalPath {
+    return entryPath.endsWith('/') ? entryPath : `${entryPath}/`;
```
habdelra: this doesn't use `this` so it can be a pure function, and probably something you want to share with operator-mode-state-service, since it does the same thing.

jurgenwerk (approved):

PR #3427, opened 2025-10-16
Optimization: when logging into realms in bulk, do not authenticate to published realms

jurgenwerk:
Previously, when logging into multiple realms, it tried to get jwts for all realms with public read, which included published realms. I think this is not a good idea since the amount of published realms will grow, resulting in choking the login process. So now we only login to own realms and catalog realms.


FadhlanR (approved):

habdelra:
@codex review

habdelra (approved):

PR #3429, opened 2025-10-16
Upgrade to @glimmer/component 2.0

ef4:
The only relevant-to-us change in 2.0 is improvements to the typescript types that make it compatible with newer ember+glint.

I added an override because there are some third-party packages pulling in 1.0 and that makes typescript sad. All the runtime behavior inside ember apps is compatible so this is safe.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
       "@types/eslint": "8.56.5",
       "@embroider/util": "1.13.1",
       "@glimmer/tracking>@glimmer/validator": "0.84.3",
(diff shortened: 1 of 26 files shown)


ef4:
This is ready to merge. 

The upgrade caused one subtle issue, which is that we were relying on a specific ordering between two destructors, when that isn't really guaranteed. The upgrade changed the order. I refactored so that we aren't order-sensitive anymore.

jurgenwerk (approved):

PR #3430, opened 2025-10-16
Parallel global setup for matrix tests

IanCal:
Notes:

The idea is to start one isolated realm server and one synapse server for all tests, and have modifications happen within realms created per user. Users are created for each test that needs them with unique ids and passwords.

This seems like a large PR, and it sort of is, but the core changes are smaller they just have wider effects

Guide to the changes:

1. Tests have been updated so that they do not alter the test realm
2. All tests get fresh users to keep isolation
3. There is a global setup (global.setup.ts) which starts the realm server and synapse server and tears them down. Since playwright works with different worker processes, information about ports/tokens/etc need to be passed via env vars. 
4. We can't pass the running IsolatedRealmServer into the workers, so there is an sql executor which changes the db directly rather than communicating with the isolated realm server. 
5. Some of the flaky tests should have been fixed after I worked on lot of timing issues, I've left their ticket number in the test name in case they start flaking again.

Notes for tests:

1. Tests should not change the test realm
2. If you want to make edits, create a new realm for your user
3. You can now run tests in parallel, with --workers 4, with this I can run the test suite in around 5 minutes
4. You can run and rerun tests without restarting the realm server (although right now if you change the application you'll need to rerun this.
6. You should be able to start the tests with `pnpm playwright test --workers 4 --ui` and get a nice UI and the realm server/etc will remain up and running.
7. High worker counts start hitting limits, these may be avoided in a future pr but I see more issues past 8 concurrent workers and the time saving is smaller.

The next steps for this are to make the tests more isolated. The base realm and prerender server can happily be run just for the tests, and with the right overrides synapse can be run on a different port entirely.


packages/matrix/tests/messages.spec.ts
```diff
@@ -1,63 +1,48 @@
...
-
   test(`it can send a message in a room`, async ({ page }) => {
-    await login(page, 'user1', 'pass', { url: appURL });
+    // TODO: this tests multiple different behaviours beyond just sending a message.
+    // There are tests for:
+    // - sending messages with a button
+    // - sending messages with Enter key
+    // - handling whitespace only messages (not sending them)
+    // - persisting messages across reloads
+    // - ensuring room state doesn't leak between users
```
lukemelia: Is this idea of this TODO that we should split up the test?

lukemelia (approved):

packages/matrix/tests/registration-without-token.spec.ts
```diff
@@ -1,74 +0,0 @@
...
-    await page.bringToFront();
-
-    await expect(page.locator('[data-test-email-validated]')).toContainText(
-      'Success! Your email has been validated',
-    );
-
-    await setupPayment('@user1:localhost', realmServer, page);
-    await assertLoggedIn(page);
-  });
-});
```
backspace, line 74: Did you try not skipping this vs deleting it? It seems like something we should keep around

backspace:
I’ll return to properly review this tomorrow, I haven’t been able to get it working:

Playwright Test 2025-11-05 16-53-55

It’s confusing because the CI script didn’t change so I’d expect it to work. But it’s impressive to see it down to three shards in Actions!

The `README.md` should probably be updated with your suggestion of how to run the tests. But does it still work to run them via the buttons in VS Code?

packages/matrix/tests/messages.spec.ts
IanCal: Yes, I should either actually split it up or create a ticket, and reading it I should phrase it as a thing to do if it's a todo.

backspace:
> Playwright Test 2025-11-05 16-53-55

I was able to get the tests running so I can proceed with review 😌

backspace:
Something weird, host mode tests aren’t running for me locally:

Playwright Test 2025-11-06 11-09-50

They‘re also not in the [report](https://boxel-matrix-playwright-reports.stack.cards/parallel-global-setup/20251105_132722Z/index.html) from the latest Actions run.

But the file doesn’t show them as skipped?!

IanCal:
@backspace the host mode ones are all skipped at the top level, they're all skipped in main as well (https://github.com/cardstack/boxel/blob/main/packages/matrix/tests/host-mode.spec.ts) 

VS code extension didn't work due to path differences and firing off ts-node for me but I'm fixing that now.

backspace:
> @backspace the host mode ones are all skipped at the top level, they're all skipped in main as well

ohhh yes I was thinking of `publish-realm-spec` that I’ve been focusing on lately, thanks.

packages/matrix/tests/auth-rooms.spec.ts
```diff
@@ -1,75 +1,61 @@
...
+        for (let realmUser of realmUsers) {
+          if (!realmRoomsByUser.has(realmUser) && joinedMembers[realmUser]) {
+            realmRoomsByUser.set(realmUser, roomId);
+          }
+        }
+      }
 
-    let roomIdToMembers = new Map<string, any>();
+      expect(realmRoomsByUser.size).toBe(realmUsers.length);
+    }).toPass();
```
backspace, line 51: I haven’t seen this before, do we need a mechanism for followup if/once it does show to be stable?

packages/matrix/tests/commands.spec.ts
```diff
@@ -1,153 +1,134 @@
...
-                    },
-                  },
-                },
-              },
-            },
-          },
-          required: ['attributes', 'description'],
-        },
-      },
-    });
```
backspace, line 125: This is a significant removal, is the idea that comparing the card id is sufficient?

packages/matrix/tests/create-realm.spec.ts
```diff
@@ -1,53 +1,21 @@
...
-    });
-    await registerRealmUsers(synapse);
-    realmServer = await startRealmServer();
-    await registerUser(synapse, 'user1', 'pass');
-  });
-
-  test.afterEach(async () => {
-    await realmServer?.stop();
-    await synapseStop(synapse.synapseId);
-  });
```
backspace, line 40: I appreciate that so much of this boilerplate is extracted/superfluous now, its accumulation has been bothering me

backspace (approved):
This is great, huge improvements!

PR #3433, opened 2025-10-17
Refactor /render route model loading and all node tests for headless chrome indexing ported

habdelra:
In this PR I made a very good refactor that solved async bugs in our prerender. Specifically we were not waiting for the right moment to begin tracking promises related to field loading. The idea is that the rendering our our template is what pulls on the linked fields to begin their lazy loading. However, previously in our `/render` route, we were waiting for the linked fields to finish loading _before_ we actually rendered the template. We were just doing a simple wait for the store to settle in the model hook—this was simply too early.  this meant that oftentimes the `store.loaded()` promise would resolve before we ever started tracking the link loading promises because the template didn't yet have an opportunity to pull on the linked fields. In fact, there is no point in time in the model hook of the `/render` route when we even know what all linked fields should be loaded. 

this led to a lot of flakiness that became ever more present in the headless chrome node indexing tests as i got farther along. after codex start making crazy hacks to work around the fact that the instance was not settling, I took a step back and rethink this.

 This PR establishes a deferred in our `/render` route model hook that only begins the wait on the linked fields _after_ we have rendered the instance's template, at the point that all the lazy loading for the linked fields has kicked off and we’ve begun tracking those inflight requests. There is a lot of rather paranoid prerender capture code that stemmed from this flakiness that could probably be cleaned up now. I'll try to slim things  down in a subsequent PR.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -280,6 +280,9 @@ export function instanceOf(instance: BaseDef, clazz: typeof BaseDef): boolean {
 class Logger {
   private promises: Promise<any>[] = [];
 
+  // TODO this doesn't look like it's used anymore. in the past this was used to
+  // keep track of async when eagerly running computes after a property had been set.
+  // consider removing this.
   log(promise: Promise<any>) {
     this.promises.push(promise);
     // make an effort to resolve the promise at the time it is logged
@@ -3047,9 +3050,9 @@ export function setCardAsSavedForTest(instance: CardDef, id?: string): void {
   instance[isSavedInstance] = true;
 }
 
-export async function searchDoc<CardT extends BaseDefConstructor>(
+export function searchDoc<CardT extends BaseDefConstructor>(
   instance: InstanceType<CardT>,
-): Promise<Record<string, any>> {
+): Record<string, any> {
   return getQueryableValue(instance.constructor, instance) as Record<
     string,
     any
--- a/packages/host/app/routes/render.ts
+++ b/packages/host/app/routes/render.ts
@@ -1,8 +1,9 @@
+import type Controller from '@ember/controller';
 import { action } from '@ember/object';
 import Route from '@ember/routing/route';
 import RouterService from '@ember/routing/router-service';
 import Transition from '@ember/routing/transition';
-import { join } from '@ember/runloop';
+import { join, scheduleOnce } from '@ember/runloop';
 import { service } from '@ember/service';
 
 import { TrackedMap } from 'tracked-built-ins';
@@ -18,6 +19,7 @@ import {
   parseRenderRouteOptions,
   serializeRenderRouteOptions,
 } from '@cardstack/runtime-common';
+import { Deferred } from '@cardstack/runtime-common/deferred';
 import { serializableError } from '@cardstack/runtime-common/error';
 
 import type { CardDef } from 'https://cardstack.com/base/card-api';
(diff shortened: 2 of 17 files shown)


packages/realm-server/prerender/utils.ts
```diff
@@ -397,8 +438,14 @@ export async function captureResult(
...
           alive,
-          id: resolvedElement.dataset.prerenderId ?? undefined,
-          nonce: resolvedElement.dataset.prerenderNonce ?? undefined,
+          id:
+            resolvedElement.dataset.prerenderId ??
+            errorElement?.dataset.dataPrerenderId ??
+            undefined,
+          nonce:
+            resolvedElement.dataset.prerenderNonce ??
+            errorElement?.dataset.dataPrerenderNonce ??
```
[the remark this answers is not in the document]
habdelra: fixed

habdelra:
@codex review

packages/realm-server/tests/prerendering-test.ts
```diff
@@ -492,11 +530,7 @@ module(basename(__filename), function () {
       test('searchDoc', function (assert) {
         assert.strictEqual(result.searchDoc?.name, 'Maple');
         assert.strictEqual(result.searchDoc?._cardType, 'Cat');
-        // This assertion seems flaky in CI is there some kind of race condition
-        // here?. we do have coverage for this in host tests, but it would be
-        // nice to see this in server tests too...
-
-        // assert.strictEqual(result.searchDoc?.owner.name, 'Hassan');
```
habdelra, line 499: the source of the flakiness was indeed the /render route link loading. this is now fixed and this assertion has been revived

packages/realm-server/tests/headless-chrome-indexing-test.ts
```diff
@@ -1477,19 +1646,6 @@ module(basename(__filename), function () {
...
-          instancesIndexed: 0,
-          instanceErrors: 2,
-          moduleErrors: 2,
-          modulesIndexed: 0,
-          definitionErrors: 0,
-          definitionsIndexed: 0,
-          totalIndexEntries: 20,
-        },
-        'indexed correct number of files',
-      );
```
habdelra, line 1492: in general i'm starting to move away from these types of assertions. they are too brittle, and often making assertions around counts that have nothing to do with the tests. there are other more robust ways to assert that an instance was indexed (like make a change to it and search for that change--which we do in a lot of places that we are also making this kind of assertion)

packages/host/app/routes/render.ts
```diff
@@ -137,17 +180,52 @@ export default class RenderRoute extends Route<Model> {
...
+    // the isolated format), and then the store will start tracking the
+    // in-flight link requests. after the store settles the prerendered output
+    // will be ready to capture. this readyDeferred will let us know when the
+    // prerendered output is ready for capture.
+    let readyDeferred = new Deferred<void>();
+    let modelState: ModelState = {
+      state,
+      readyDeferred,
+      isReady: false,
+    };
```
habdelra, line 199: this is the meat of the new refactor and what is making our prerender stable now

packages/host/app/routes/render/meta.ts
```diff
@@ -36,7 +36,8 @@ export default class RenderMetaRoute extends Route<Model> {
 
   async model(_: unknown, transition: Transition) {
     let api = await this.cardService.getAPI();
-    let parentModel = this.modelFor('render') as ParentModel;
+    let parentModel = this.modelFor('render') as ParentModel | undefined;
+    await parentModel?.readyPromise;
```
habdelra, line 40: waiting for the parent's model hook is not enough--it's only after that that we'll start to get the linked fields resolved, which is what this ready promise will guarantee.

packages/base/card-api.gts
```diff
@@ -3047,9 +3050,9 @@ export function setCardAsSavedForTest(instance: CardDef, id?: string): void {
   instance[isSavedInstance] = true;
 }
 
-export async function searchDoc<CardT extends BaseDefConstructor>(
+export function searchDoc<CardT extends BaseDefConstructor>(
```
habdelra, line 3053: this is actually not async, it never was....

lukemelia (approved):

PR #3438, opened 2025-10-20
Implement fetching claimed site name and unclaim button

FadhlanR:
https://github.com/user-attachments/assets/dfd5c8fa-4f74-48b6-8001-fe9728abc9c7

### DIFF
--- a/packages/host/app/components/operator-mode/publish-realm-modal.gts
+++ b/packages/host/app/components/operator-mode/publish-realm-modal.gts
@@ -1,7 +1,7 @@
 import { fn } from '@ember/helper';
 import { on } from '@ember/modifier';
-import type Owner from '@ember/owner';
 import { action } from '@ember/object';
+import type Owner from '@ember/owner';
 import { service } from '@ember/service';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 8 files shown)


packages/host/app/services/realm-server.ts
```diff
@@ -558,6 +565,75 @@ export default class RealmServerService extends Service {
     return (await response.json()) as SubdomainAvailabilityResult;
   }
 
+  async fetchClaimedSiteHostname(
```
jurgenwerk: We ended up with a naming dichotomy that is confusing–the endpoints are named in a style like "claimed-domains" but the type they are returning in the json api response is "claimed-site-hostname" (this incosistency was already there before your PR, but now it became more apparent), and now these terms are both used in this component, and I think we should decide which one to use. I suggest changing the type of this resource from `claimed-site-hostname` to `claimed-domain` (in the endpoints), and adjust names like `fetchClaimedSiteHostname` to `fetchClaimedDomain`, and similar.

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -351,6 +466,11 @@ export default class PublishRealmModal extends Component<Signature> {
     },
   );
 
+  @action
+  handleUnclaimSiteName() {
```
jurgenwerk: You can use {{perform ...}} helper. I think the way it's done here it will cast the action call into a promise, effectively losing the benefits of an ember concurrency task

jurgenwerk (changes requested):

jurgenwerk (approved):

PR #3440, opened 2025-10-20
Update schema

jurgenwerk:
Excludes `claimed_domains_for_sites` and `session_rooms` since these are not needed for in-browser indexing in tests.


backspace (approved):

lukemelia (approved):
Thanks!

PR #3445, opened 2025-10-20
Headless chrome indexer error audit results fixes

habdelra:
This PR fixes indexing issues discovered during the headless chrome indexer error audit:

1. Indexing jobs need more lenient timeouts. When realms have no indexing errors or recoverable indexing errors the indexing is very fast, faster than our current indexer. However, when there are render errors that are not recoverable, the indexing is much much slower. in general the catalog realm needs much more time to index from a blank slate. Generally i'm seeing it take about 6 minutes to index from a blank slate. The experiments realm (2.5min), base realm (20s), and skills realm (5s) all index faster than before tho.
2. The last known good html that appears in the error docs is getting confused for actual DOM when we try to render the serialized form of these errors. we strip this out from our errors before rendering them in the DOM so the browser doesn't get confused (this HTML is only meant for human consumption on errors--the prerenderer doesn't need it)
3. the NotLoadedValue sentinel that we put in our deserialized state bucket for instances was leaking out when we try to render a card while the links are still inflight. I was able to clean this up for all teh field types except for LinksToMany. there is a deeper component stability issue here that we cant really address until the .value refactor. details have been left in the comments. however, this means that for LinksToMany, the NotLoaded sentinel will occasionally be exposed to card developers. I cleaned up a bunch of cards in our catalog realm where that happens. Codex was very very adamant that we don't try to go deeper with a fix here. it brought up lots of very good points about a super painful refactor that leaves us with lots of duplicated state management barring a .value refactor.
4. Our glimmer scoped styles is actually causing issues for us due to the way it manipulates the DOM out from underneath glimmer. when we reuse a page and the render route is trying to tear down the DOM, glimmer gets mad that one of the DOM nodes that its trying to remove is no longer there (the <style> tag embedded in the card). I tried a bunch of approaches to this, but none were stable. so instead we detect this error and when we see it we clear the store cache and try again--this solves the issue.

Most of the rest of the issues are actually related to downloading modules from external sites and CORS violations due to our own custom headers. I broke that out into a separate ticket that i'll followup with shortly https://linear.app/cardstack/issue/CS-9562/remove-the-need-to-use-x-boxel-building-index-header.

 Beyond that there are a handful of other errors that we don't see in the current indexer that I'm still tracking down. One nice thing tho is that there are also a bunch of errors in our current indexer around rendering cycles that are actually solved in the headless chrome indexer.


habdelra:
@codex review

packages/realm-server/prerender/prerenderer.ts
```diff
@@ -31,6 +31,13 @@ import {
 
 const log = logger('prerenderer');
 const boxelHostURL = process.env.BOXEL_HOST_URL ?? 'http://localhost:4200';
+const CLEAR_CACHE_RETRY_SIGNATURES: readonly (readonly string[])[] = [
+  // this is a side effect of glimmer scoped styles moving a DOM node that
+  // glimmer is tracking. when we go to teardown the component glimmer gets mad
+  // that a node it was tracking is no longer there. performing a new prerender
+  // capture with a cleared store/loader cache will workaround this issue.
+  [`Failed to execute 'removeChild' on 'Node'`, 'NotFoundError'],
+];
```
backspace, line 40: I’m surprised the `style` element is even present, I thought the AST transform would [remove it from the component altogether](https://github.com/cardstack/glimmer-scoped-css/blob/19299bdf0032cea47d04aecf711e4121ec34e522/glimmer-scoped-css/src/ast-transform.ts#L86)

backspace (approved):

packages/realm-server/prerender/prerenderer.ts
habdelra: agreed, I wonder if it’s the AST removing it that is what glimmer is mad about. i did confirm tho if I change the AST to leave the style tag in the card DOM it error goes away.

backspace: Maybe there’s a way to have the transform happen earlier in the chain, before Glimmer has a hold on any elements at all. The way it runs on `boxel-ui`, for instance, is static. But I’m definitely out of my depth here.

PR #3448, opened 2025-10-21
host: Restore registration

backspace:
When `realmServer.authenticateToAllAccessibleRealms` was added to `matrixService.start` as part of logging into multiple realms with one call, that introduced a race condition where authentication might fail because the necessary record hadn’t yet been inserted into `users`, so `_realm-auth` would fail:

Boxel 2025-10-21 09-31-47

This extracts Matrix client setup so the `users` row can be inserted before realm login happens in `this.start`.

Without this, I had ≈25% success rate in registering on staging, with this deployed I never encountered a failure, though of course it’s difficult to prove an absence.


packages/host/app/services/matrix-service.ts
```diff
@@ -435,7 +435,7 @@ export default class MatrixService extends Service {
   ) {
     displayName = displayName.trim();
     this._isInitializingNewUser = true;
-    this.start({ auth });
+    await this.start({ auth, skipRealmAuthentication: true });
```
jurgenwerk: Is it possible to somehow get the the user id before calling `this.start()`? That way we could create the user before `start`ing.

packages/host/app/services/matrix-service.ts
```diff
@@ -504,9 +506,10 @@ export default class MatrixService extends Service {
     opts: {
       auth?: MatrixSDK.LoginResponse;
       refreshRoutes?: true;
+      skipRealmAuthentication?: boolean;
     } = {},
   ) {
-    let { auth, refreshRoutes } = opts;
+    let { auth, refreshRoutes, skipRealmAuthentication } = opts;
```
jurgenwerk: Perhaps it's worth adding a comment why we added this option?

packages/host/app/services/matrix-service.ts
backspace: Do you have a suggestion on how to do this? `createUser` [calls `login`](https://github.com/cardstack/boxel/blob/30a17f383af5ec8753b440880f4efdce6ddfa31d/packages/host/app/services/realm-server.ts#L127), which fails without a client:

Boxel 2025-10-21 14-09-37

boxel-motion 2025-10-21 14-10-03

I‘m unfamiliar with the intricacies here

packages/host/app/services/matrix-service.ts
backspace: done in 3487bb042, thanks

packages/host/app/services/matrix-service.ts
habdelra: i'm not sure of your entire flow here, but if you have the matrix username in hand, there is a function in runtime common that can synchronously derive the user id using the matrix naming rules: `userIdFromUsername()`

backspace: I do have the Matrix username:

Image

but calling `createUser` fails because `this.start` hasn’t been called. I don’t understand how to get out of this chicken-and-egg situation without something like `skipRealmAuthentication`

IanCal:
When running the matrix tests on a different branch under high contention, I'm seeing what I think is exactly this bug appearing with quite high frequency in the main user registration test. I'll try and test this fix against it

IanCal (approved):
I'm not sure I entirely follow the realm auth flow, so am happy to defer to others about this solution, however I was seeing this regularly break my tests with 2-4 parallel workers and with this change they all pass with 8+ workers. I've run it a few times now and not seen the JWT error and before it was happening in various tests every single run.

PR #3454, opened 2025-10-22
host: Add text to Boxel icon

backspace:
It was previously misaligned:

image

It also had the font changing while loading:

screencast 2025-10-22 13-57-49

This fixes both of those:

screencast 2025-10-22 13-55-59

The spacing of the word from the image is also increased to be more like the design:

V1 Dashboard - Generative Software (Work in Progress) - Zeplin 2025-10-22 13-57-14


packages/host/app/components/operator-mode/submode-layout.gts
```diff
@@ -633,7 +640,10 @@ export default class SubmodeLayout extends Component<Signature> {
       .workspace-button--dark {
         --icon-bg-opacity: 1;
         --icon-color: var(--boxel-dark);
+        --icon-bg-color: var(--boxel-highlight);
+        --boxel-icon-button-width: 40px;
```
backspace, line 644: Another way to accomplish this would be to use different icons based on the state, vs truncating the text from this one via CSS

burieberry (approved):

PR #3455, opened 2025-10-22
Prevent copy card to appear with 2 stacks when you can't have write access to the index card

tintinthong:
Screenshot 2025-10-22 at 20 25 38

The copy button shudn appear here when you dont have write access to the catalog

Since the refactor of the viewCard api, we lost the stack context hence viewing the listing card by clicking from catalog will open the listing card on the stack to the right. Tho, I think somehow the experience of 2 stacks is "not bad" so I decided to fix the issue where the copy button appears when a person doesn't have write access to the index card in the realm

Screenshot 2025-10-22 at 20 30 42


tintinthong:
flaky-test related to this change 

can delete a card from the index card stack item

tintinthong:
@codex review

backspace (approved):
I tried this out in the preview deployment and the button didn’t show as expected 🎉

PR #3444, opened 2025-10-20
Add runtime-resolvable @cardstack/catalog realm "package path"

lukemelia:
- expose resolvedCatalogRealmURL and wire the virtual network to resolve @cardstack/catalog/…
- teach realm-server start scripts, wait scripts, and mock helpers to respect the resolved catalog URL (with portable trailing-slash handling)
- replace hard-coded localhost catalog URLs in host acceptance/integration tests and AI bot fixtures with config-driven values

PR #3456, opened 2025-10-22
host: Add resolved catalog realm to builds

backspace:
This was missing from #3444. In general I think we have confusion about which things are configured via environment variables during a `host` build vs which are [overwritten when served by the realm server](https://github.com/cardstack/boxel/blob/9d4e0d4b163f1032d3bd3388686cb8a746492341/packages/realm-server/server.ts#L323-L327). This variable is [in the overwritable section](https://github.com/cardstack/boxel/blob/9d4e0d4b163f1032d3bd3388686cb8a746492341/packages/host/config/environment.js#L52-L54) but isn’t actually overwritten.


tintinthong (approved):
It looks good !

PR #3457, opened 2025-10-22
Move source code and transpiled code out of index and into read thru cache in realm

habdelra:
This PR is a prerequisite for removing the X-Boxel-Building-Index request header. First we need to move the source code and transpiled code out of the index and into a read thru cache in the realm.


packages/host/app/components/operator-mode/syntax-error-display.gts
```diff
@@ -20,15 +20,20 @@ export default class SyntaxErrorDisplay extends Component<Signature> {
...
+      return maybeCardError;
     }
     delete json!.deps; // definitely json exists at this point
-    return JSON.stringify(json, null, 2);
+    const stackDepth = 3;
+    // prevent super deep stacks
+    return (json!.stack ?? JSON.stringify(json, null, 2))
+      .split('\n')
+      .slice(0, stackDepth)
+      .join('\n');
```
habdelra: we've moved where we are transpiling gts files, and the stack traces are enormous now because of that. generally only the last few frames are useful for debugging so i'm concatting the stack trace.

lukemelia (approved):

PR #3461, opened 2025-10-23
Publish with custom site name

FadhlanR:
This PR ensures that publishing a realm with a custom site name works as expected. In addition, it fixes the margin issue we found by removing the background color from each domain option to match the design. It also implements an error message to display when publishing fails.

https://github.com/user-attachments/assets/9db6cac1-1b20-4661-bc20-429178e5b87a

https://github.com/user-attachments/assets/2f6ee85a-2fb2-443c-a066-b5517037dee2

### DIFF
--- a/packages/host/app/components/operator-mode/host-submode.gts
+++ b/packages/host/app/components/operator-mode/host-submode.gts
@@ -11,7 +11,6 @@ import onClickOutside from 'ember-click-outside/modifiers/on-click-outside';
 
 import { restartableTask } from 'ember-concurrency';
 import perform from 'ember-concurrency/helpers/perform';
-import window from 'ember-window-mock';
 
 import { BoxelButton, Tooltip } from '@cardstack/boxel-ui/components';
 import { PublishSiteIcon } from '@cardstack/boxel-ui/icons';
@@ -158,23 +157,33 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
     await this.realm.unpublish(this.realmURL, publishedRealmURL);
   });
 
+  get defaultPublishedSiteURL(): string | undefined {
+    let defaultURL = this.defaultPublishedRealmURL;
+    if (defaultURL) {
+      return this.getFullURL(defaultURL);
+    }
+    return undefined;
+  }
+
   @action
   handleOpenSiteButtonClick(event: MouseEvent) {
-    event.preventDefault();
-    event.stopPropagation();
-
     if (event.shiftKey) {
+      event.preventDefault();
+      event.stopPropagation();
       this.isOpenSitePopoverOpen = !this.isOpenSitePopoverOpen;
       return;
     }
 
-    this.isOpenSitePopoverOpen = false;
-    let defaultURL = this.defaultPublishedRealmURL;
-    if (defaultURL) {
-      window.open(this.getFullURL(defaultURL), '_blank');
-    } else {
+    // If there's no default URL, prevent navigation and show popover
+    if (!this.defaultPublishedRealmURL) {
+      event.preventDefault();
+      event.stopPropagation();
       this.isOpenSitePopoverOpen = true;
+      return;
     }
+
+    // Otherwise, let the anchor navigate naturally
+    this.isOpenSitePopoverOpen = false;
   }
 
   private parsePublishedAt(value: unknown) {
@@ -240,9 +249,13 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
             <Tooltip class='open-site-tooltip'>
               <:trigger>
                 <BoxelButton
+                  @as='anchor'
                   @kind='secondary'
                   @size='tall'
+                  @href={{this.defaultPublishedSiteURL}}
                   class='open-site-button'
+                  target='_blank'
+                  rel='noopener noreferrer'
                   {{on 'click' this.handleOpenSiteButtonClick}}
                   data-test-open-site-button
                 >
--- a/packages/host/app/components/operator-mode/host-submode.gts
+++ b/packages/host/app/components/operator-mode/host-submode.gts
@@ -30,6 +30,8 @@ import HostModeContent from '../host-mode/content';
 
 import SubmodeLayout from './submode-layout';
 
+import type { PublishError } from './publish-realm-modal';
+
 interface HostSubmodeSignature {
   Element: HTMLElement;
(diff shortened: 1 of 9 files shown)


backspace:
I was trying out your branch via the preview, I’d already claimed a subdomain and it correctly detected that, but when I unclaimed it and tried to claim it again, this happened:

minicatalog 2025-10-23 09-39-31

I’ll look into it further, I don’t think it relates to your PR here, just making note. Will proceed with review now!

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -43,14 +43,20 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
   @tracked isPublishRealmModalOpen = false;
   @tracked isPublishingRealmPopoverOpen = false;
   @tracked isOpenSitePopoverOpen = false;
+  @tracked publishError: string | null = null;
+  @tracked publishErrors: Map<string, string> = new Map();
```
backspace: Is it possible to combine `publishError` into `publishErrors` somehow? It seems clumsy to me to have two of these passed around everywhere and reset.

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -112,7 +118,27 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
...
+          const error = result as PromiseRejectedResult;
+          const errorMessage =
+            error.reason?.message || 'Failed to publish to this domain';
+          this.publishErrors.set(url, errorMessage);
+        }
+      });
+
+      // Trigger reactivity by creating a new Map instance
+      this.publishErrors = new Map(this.publishErrors);
+    }
```
backspace, line 137: Since this is an Ember Concurrency task, it seems possible to encapsulate the errors into the task’s `last` run, as described [here](https://ember-concurrency.com/docs/derived-state). Then the error state can be detected with `handlePublish.last.isError` and the error(s) accessed with `handlePublish.last.error`. This seems more elegant to me.

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -468,6 +481,13 @@ export default class PublishRealmModal extends Component<Signature> {
     window.open(this.generatedUrl, '_blank');
   }
 
+  @action
+  handleOpenCustomSubdomainSite() {
+    if (this.claimedDomainPublishedUrl) {
+      window.open(this.claimedDomainPublishedUrl + 'index', '_blank');
+    }
+  }
```
backspace, line 484: I failed to notice this before but I’d prefer these buttons be true links vs using `window.open`. That lets you hover over the button and see the URL. I don’t see any value in doing this via redirect. (Same for the above.)

packages/host/app/components/operator-mode/publish-realm-modal.gts
backspace: Let’s rename this now that it’s not the only site?

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -712,19 +820,43 @@ export default class PublishRealmModal extends Component<Signature> {
...
+                @kind='text-only'
+                @size='extra-small'
+                class='clear-error-button'
+                {{on 'click' this.clearPublishError}}
+                data-test-clear-publish-error
+              >
+                <IconX width='12' height='12' />
+              </BoxelButton>
+            </div>
+          {{/if}}
```
backspace: How does this get rendered? I didn’t see it in your videos. It doesn’t seem to be covered by a test.

packages/host/tests/acceptance/host-submode-test.gts
```diff
@@ -1012,6 +1012,298 @@ module('Acceptance | host submode', function (hooks) {
...
+
+          window.open = (url?: URL | string, target?: string) => {
+            assert.strictEqual(
+              url,
+              'http://my-custom-site.localhost:4201/index',
+              'Open custom subdomain site URL',
+            );
+            assert.strictEqual(target, '_blank', 'Open in a new tab');
+            return null;
+          };
```
backspace: Another benefit of using true links (styled as buttons) is that you don’t need to mock `window.open`, you can just check the value of `a.href`.

backspace (changes requested):
I tried this out and it worked great, I have some suggestions for implementation improvement though.

I’m not sure if this is related to your PR but notice the alignment of the Claim button here, could it be grid-related?

Image

It’s meant to be aligned with the field:

Image

backspace:
I do think the button alignment is a problem with this PR because this is how it looks on staging:

oct2 2025-10-23 11-01-45

(The state appearing valid despite the field being empty is a bug though, I’ll open an issue.)

packages/host/app/components/operator-mode/publish-realm-modal.gts
FadhlanR: I removed this part, this is not needed.

FadhlanR:
> I do think the button alignment is a problem with this PR because this is how it looks on staging:
> 
> oct2 2025-10-23 11-01-45
> (The state appearing valid despite the field being empty is a bug though, I’ll open an issue.)

Really appreciate your feedback. I fixed the position of the “Claim Site” button, refactored the error handling using `task.last.error` so it looks cleaner, and changed the “Open Site” button to be a link.

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -111,32 +112,78 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
...
 
+  @action
+  handlePublish(publishedRealmURLs: string[]) {
+    let promise = this.handlePublishTask.perform(publishedRealmURLs);
+    // Catch the error so it doesn't bubble
+    // to the browser (so error reporters like Bugsnag will see it).
+    // TODO: remove this after this issue (https://github.com/machty/ember-concurrency/issues/40)
+    // is resolved.
+    promise.catch((_error) => {});
+  }
```
backspace, line 150: I forgot about or didn’t know about this, unpleasant 😞 but I think the error encapsulation is still worth it

backspace (approved):
thanks for the updates, I tried it again via the preview, excellent 🎉

habdelra:
@codex review

PR #3462, opened 2025-10-23
Make card items in cards grid clickable in host mode

FadhlanR:
https://github.com/user-attachments/assets/498a6835-e009-44cc-b01f-06e6f401b21e

### DIFF
--- a/packages/base/components/card-list.gts
+++ b/packages/base/components/card-list.gts
@@ -1,12 +1,26 @@
+import { fn } from '@ember/helper';
+import { on } from '@ember/modifier';
+import { action } from '@ember/object';
 import Component from '@glimmer/component';
+
+import { consume } from 'ember-provide-consume-context';
 
 import { LoadingIndicator } from '@cardstack/boxel-ui/components';
 
 import { cn, eq } from '@cardstack/boxel-ui/helpers';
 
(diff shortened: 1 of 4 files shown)


backspace (approved):
I tried this out locally and it worked well!

It’s not a problem with your PR but I do wonder why there needs to be special handling for host mode, do card authors need to think about it when developing cards? It seems to me that The Correct Way should be the same when authoring cards regardless of where they’ll be used. I think host mode would be the place most people are interacting with cards.

FadhlanR:
> I tried this out locally and it worked well!
> 
> It’s not a problem with your PR but I do wonder why there needs to be special handling for host mode, do card authors need to think about it when developing cards? It seems to me that The Correct Way should be the same when authoring cards regardless of where they’ll be used. I think host mode would be the place most people are interacting with cards.

Based on the discussion with @lukemelia, we want card authors to decide whether they want to implement the `viewCard` function or other functions from `cardCrudFunctions` in their cards. We’re not using the same approach as in the *interact* submode, because in that case, overlays are applied to all rendered cards and `linksTo`/`linksToMany` cards. The issue with that approach is that card authors cannot opt out — the overlay is always applied, regardless of whether they want to use it or not. CMIIW @lukemelia

lukemelia (approved):

PR #3469, opened 2025-10-24
Button fixes and refactor

burieberry:
### `ContextButton` component (new):  
- Open to other naming options
- Goal: to provide a standard component for use in system menus and reduce inconsistencies in styling
- Defaults to `...` menu, offers options for `context-menu-vertical`, `add`, `edit`, `delete`, `close` (most common use cases) or can provide custom icon
- Accepts size arguments described in button components below
- Provides color schemes for most common use cases
A couple of examples on hover:
context-button
context-ghost
context-destructive

### `Button` component:
 - Update button `loading` state styling: buttons should maintain current style with loading icon in place (instead of disabling the button) - this is a more accessible, recommended pattern.
 - Introduce `rectangular` button argument
rectangular
- Update `muted` & `text-only` button hover styles since they were not visible on certain backgrounds and now `text-only` works for both dark and light backgrounds as is
- Introduce standardized button size variables for sizes `extra-small`, `small`, `base`, `tall`, and `touch` size variants

### `IconButton` component:
- Introduce the same `size` variants as button component
- Adjust border-radius per size variant
- Standardize icon sizes (16px or 20px) -- this is tricky because some icons have padding, some don't. Most lucide icons seem to have 2px padding, so I selected the sizes accordingly

### `Menu` component:
- Update the hover styling on `danger` menu-item and adjust `Dropdown` component to not force-override it via `!important` keyword. For most places `important` keyword is used in dropdown, we can instead update the menu component, but keeping that work separate.

### `Alert` component:
- Fix the color-clash of primary button on warning background
warning-before
warning-after

## Other fixes:
- Button fixes on various components (many of them now using ContextButton component)
- Positioning on past sessions menu when chat sidebar is wide
- Dropdown hover state fixes

Before: cropped/weirdly colored icon and menu position
before-menu

After:
after-position
hover-3

hover-1

hover-2

dropdown-danger

dropdown-before-danger-hover

after-danger-hover


lukemelia (approved):
Glad to see AddButton lose its special status

PR #3470, opened 2025-10-24
Field Configuration API

lukemelia:
_Note: we're going to build enumField on top of this._

# Per-Instance Field Configuration (without exposing parent instance)

## Objective

Provide a consistent, general mechanism for per-instance field configuration that:

- Lets a parent card compute configuration for a field (e.g., options, palettes, constraints).
- Keeps the parent instance private (never passed into templates or field components).
- Offers a safe, reactive API that gracefully handles NotLoaded relationships.
- Applies to any field type (independent of enumField).

## API Summary (single, consistent key: configuration)

- Field definition (FieldDef):
  - `static configuration?: ConfigurationInput<any>`
- Field usage (`contains` / `containsMany` / `linksTo` / `linksToMany`):
  - `contains(Field, { configuration?: ConfigurationInput<ParentInstance> })`
  - `containsMany(Field, { configuration?: ConfigurationInput<ParentInstance> })`
  - `linksTo(Card, { configuration?: ConfigurationInput<ParentInstance> })`
  - `linksToMany(Card, { configuration?: ConfigurationInput<ParentInstance> })`

Where:

- `ConfigurationInput<T> = FieldConfiguration | ((self: Readonly<T>) => FieldConfiguration | undefined)`
- `FieldConfiguration` is field-specific (documented by each field type).
- Both configuration sources (FieldDef and contains) are computed and merged. There is no choose-one precedence; instead, the effective config is the merge of all available sources.

## Function signature and NotLoaded handling

- Configuration functions have the same signature in both places (FieldDef and contains): `(self: Readonly<T>) => FieldConfiguration | undefined`.
- The runtime handles reactivity and NotLoaded for you:
  - Before invocation, call `entangleWithCardTracking(self)`.
  - Invoke the function inside a try/catch; if accessing `self.someLink` throws NotLoaded, treat this fragment as `undefined` for the current render and rely on re-render when links load.
  - No special context object or `read()` wrapper is required.

## FieldConfiguration shape

- Each field type defines and validates its own schema. Suggested pattern is to namespace by capability:
  - Validation: `{ validate?: { min?: number; max?: number } }`
  - Presentation: `{ presentation?: { palette?: string[] } }`
  - Options (select-like fields): `{ options?: (primitive | { value; label?; icon? })[] }`

The runtime treats `FieldConfiguration` as an opaque object; the field class interprets it.

## Resolution and Merge (runtime)

Introduce `resolveFieldConfiguration(field, instance)` that:

- Calls `entangleWithCardTracking(instance)` to ensure reactive recomputation.
- Computes zero or more configuration fragments:
  - `fromFieldDef` by evaluating `field.card.constructor.configuration` if present (static POGO or function form with context).
  - `fromFieldUsage` by evaluating `field.configuration` (stored by contains/containsMany/linksTo/linksToMany) if present.
- Merges available fragments to produce the effective configuration.

Merge rules:

- Deterministic order: `fromFieldDef` → `fromFieldUsage`.
- Shallow merge by default (object spread semantics). For nested objects, a one-level deep merge is acceptable, but avoid surprising deep behaviors.
- Arrays are replaced rather than concatenated (field authors can decide to concatenate explicitly inside their configuration provider if desired).
- `undefined` values are treated as “absent” and do not overwrite existing values.
- If a configuration function throws due to NotLoaded, catch and treat its fragment as `undefined` for this render; allow re-render to recompute later.

Caching:

- Optionally cache by `(instance, fieldName)` in a WeakMap.
- Invalidate on `notifyCardTracking(instance)` to recompute on next access.

## Consumption

- Editors and atoms: do not receive the parent instance. The runtime computes the merged configuration and passes it directly to format components as `@configuration`.
  - All field/card format components (isolated, embedded, fitted, edit, atom) can read `@configuration` directly.
- Helpers: optional. The preferred pattern is to use the injected `@configuration` arg instead of helpers. If a helper is still desired for specific cases, it should consult `resolveFieldConfiguration` and entangle with tracking.
- Validation: field `validate()` is based on the same effective configuration. When configuration is partially `undefined` because inputs aren’t loaded yet:
  - Allow `null`/empty assignments.
  - For concrete values, validate against known configuration; if unknown, either optimistically accept (and rely on later interactions to revalidate) or fail fast with a clear “configuration still loading” error. Choose a consistent policy per field type.

## Examples

1. FieldDef-level configuration (static or function), merged

```ts
class ColorPaletteField extends FieldDef {
  static configuration: ConfigurationInput<Card> = function() {
    return {
      presentation: { palette: self.theme?.palette },
    };
  };
}

class Card extends CardDef {
  @field theme = linksTo(Theme);
  @field color = contains(ColorPaletteField);
}
```

2. Parent-level configuration merged with FieldDef

```ts
class Card extends CardDef {
  @field theme = linksTo(Theme);
  @field color = contains(ColorPaletteField, {
    configuration: { presentation: { palette: ['red', 'blue'] } },
  });
  // Effective configuration merges:
  //   ColorPaletteField.configuration(ctx or POGO) → contains(...).configuration
}
```

3. Accessing the merged configuration in templates via argument

```gts
class ColorPaletteField extends FieldDef {
  static edit = class Edit extends Component<typeof this> {
    <template>
      {{#each (@configuration.presentation.palette ??) as |color|}}
        <span class='swatch' style={{concat 'background:' color}} />
      {{/each}}
    </template>
  };
}
```

4. Validation based on merged configuration

```ts
class ColorPaletteField extends FieldDef {
  validate(instance: BaseDef, value: string | null) {
    let cfg = resolveFieldConfiguration(this, instance);
    let palette = cfg?.presentation?.palette ?? [];
    if (value == null) return value;
    if (!palette.includes(value))
      throw new Error('color not allowed by palette');
    return value;
  }
}
```

## Backward compatibility

- Existing fields continue to work with no configuration.
- Existing `contains(Field)` calls continue to work; the `configuration` option is optional and additive.
- No template changes required. Editors/helpers adopt the resolver internally.

## Implementation plan (incremental)

- Types and hooks
  - Define `ConfigurationInput<T>`: `FieldConfiguration | ((self: Readonly<T>) => FieldConfiguration | undefined)`.
  - On `FieldDef`, add optional `static configuration?: ConfigurationInput<any>`.
- On `contains/containsMany/linksTo/linksToMany` options, add optional `configuration?: ConfigurationInput<Parent>`; store on the field descriptor.

- Runtime
  - Implement `resolveFieldConfiguration(field, instance)` that entangles the instance, evaluates function or POGO inputs with NotLoaded tolerance, and merges fragments.
  - Pass the resulting merged configuration into all format components as `@configuration` (and optionally `@fieldConfig` as an alias). No helper required for common usage.

- Consumers
  - Update field editors/atoms to rely on the injected `@configuration` arg.
  - If any helper surfaces configuration data, ensure it uses the same resolver and entangles with tracking.

- Reactivity and caching
  - Cache per `(instance, fieldName)`; invalidate on `notifyCardTracking(instance)`.

- Validation
  - Update fields that validate against configuration to consult the resolver.

- Tests
  - Unit tests: merging behavior (objects vs arrays), NotLoaded tolerance, caching/invalidations.
  - Integration tests: FieldDef+contains merged configuration; CardDef+linksTo merged config; reactivity when linked data loads; validation against merged configuration.

## Performance considerations

- Avoid synchronous loading in configuration functions; throwing NotLoaded will be caught and treated as `undefined` for the current render.
- Keep configuration schemas small and flat; shallow merges are predictable and fast.
- Memoize resolved configuration sensibly; avoid excessive recomputation.

## Merge helpers (recommended)

To keep merging predictable and consistent across fields, provide internal utilities with the following semantics:

- `shallowMerge<T extends object>(a: T | undefined, b: T | undefined): T | undefined`
  - Returns a new object with one-level shallow merge of `a` then `b`.
  - `undefined` inputs are treated as absent (return the other operand).
  - `null` in `b` overwrites.
  - Arrays are replaced (not concatenated).

- `mergeConfigurations<A extends object, B extends object>(...fragments: (A | B | undefined)[]): (A & B) | undefined`
  - Folds left-to-right using `shallowMerge` to implement the deterministic order.

TypeScript sketch:

```ts
function isObject(v: unknown): v is Record<string, unknown> {
  return !!v && typeof v === 'object' && !Array.isArray(v);
}

export function shallowMerge<T extends Record<string, any>>(
  a?: T,
  b?: T,
): T | undefined {
  if (!a && !b) return undefined;
  if (!a) return b;
  if (!b) return a;
  let out: Record<string, any> = { ...a };
  for (let [k, v] of Object.entries(b)) {
    if (v === undefined) continue; // ignore undefined
    // arrays and primitives replace; objects shallow-merge one level
    if (isObject(v) && isObject(out[k])) {
      out[k] = { ...(out[k] as Record<string, unknown>), ...v };
    } else {
      out[k] = v; // replace (includes null and arrays)
    }
  }
  return out as T;
}

export function mergeConfigurations<T extends object>(
  ...fragments: (T | undefined)[]
): T | undefined {
  return fragments.reduce<T | undefined>(
    (acc, next) => shallowMerge(acc, next),
    undefined,
  );
}
```

Notes:

- If a field requires deeper or custom semantics (e.g., array concatenation), it should do so inside its configuration function and still return a plain POGO for merging.
- Keep helpers internal; only field authors and runtime use them.

## Security and safety

- Parent instance is never exposed to templates or field components.
- Consider dev-mode guards against mutation of `self` inside configuration functions.
- Treat configuration as immutable by convention.

## Relation to `enumField`

- `enumField` can adopt this mechanism by defining `static configuration?: ConfigurationInput<Parent>` that returns `{ options }` (either as a POGO or via link access on `self` that may throw NotLoaded; runtime tolerates this).
- Parent authors can add per-usage configuration via `contains(PriorityField, { configuration })` and the two are merged.

## Consuming `@configuration` in custom field components

Any field format component (embedded, fitted, atom, edit) receives the merged configuration as an argument named `@configuration`. Read it directly from `this.args.configuration` and render defensively in case configuration is temporarily unavailable (e.g., while links load).

Example: simple palette consumer in an edit template

```gts
import { FieldDef, Component } from 'https://cardstack.com/base/card-api';

// A field whose configuration provides a list of colors to render/select
export class ColorField extends FieldDef {
  static edit = class Edit extends Component<typeof this> {
    // Normalize to an array to keep template simple
    get palette(): string[] {
      return ((this.args.configuration as any)?.presentation?.palette ??
        []) as string[];
    }

    <template>
      {{#if this.palette.length}}
        <ul class='color-palette' data-test-color-palette>
          {{#each this.palette as |color|}}
            <li
              class='swatch'
              style={{concat 'background:' color}}
              data-test-swatch={{color}}
            />
          {{/each}}
        </ul>
      {{else}}
        <em data-test-no-palette>palette loading or not configured</em>
      {{/if}}
    </template>
  };
}
```

Example: options for a select-like field (primitive or rich options)

```gts
import { FieldDef, Component } from 'https://cardstack.com/base/card-api';

type Option = string | { value: string; label?: string };

export class SelectField extends FieldDef {
  static edit = class Edit extends Component<typeof this> {
    get options(): { value: string; label: string }[] {
      let raw = (this.args.configuration as any)?.options as
        | Option[]
        | undefined;
      if (!raw) return [];
      return raw.map((o) =>
        typeof o === 'string'
          ? { value: o, label: o }
          : { value: o.value, label: o.label ?? o.value },
      );
    }

    <template>
      <select data-test-select>
        {{#each this.options as |opt|}}
          <option value={{opt.value}}>{{opt.label}}</option>
        {{/each}}
      </select>
    </template>
  };
}
```

Notes:

- Reactivity: configuration recomputes automatically when its inputs change. If the configuration function reads links, it may be undefined until those links load; guard the template accordingly.
- Privacy: the parent instance is not provided to components—only `@configuration`.
- Types: configuration is intentionally opaque; field authors define the shape and should validate/normalize locally inside the component or field logic.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -416,7 +416,6 @@ class ContainsMany<FieldT extends FieldDefConstructor>
   readonly description: string | undefined;
   readonly isUsed: undefined | true;
   readonly isPolymorphic: undefined | true;
-  // Optional per-usage configuration for this field
   configuration: ConfigurationInput<any> | undefined;
   constructor({
     cardThunk,
@@ -731,7 +730,6 @@ class Contains<CardT extends FieldDefConstructor> implements Field<CardT, any> {
   readonly description: string | undefined;
   readonly isUsed: undefined | true;
   readonly isPolymorphic: undefined | true;
-  // Optional per-usage configuration for this field
   configuration: ConfigurationInput<any> | undefined;
   constructor({
     cardThunk,
@@ -934,6 +932,7 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
   readonly description: string | undefined;
   readonly isUsed: undefined | true;
   readonly isPolymorphic: undefined | true;
+  readonly configuration?: ConfigurationInput<any>;
   constructor({
     cardThunk,
     computeVia,
@@ -1299,6 +1298,7 @@ class LinksToMany<FieldT extends CardDefConstructor>
   readonly description: string | undefined;
   readonly isUsed: undefined | true;
   readonly isPolymorphic: undefined | true;
+  readonly configuration?: ConfigurationInput<any>;
   constructor({
     cardThunk,
     computeVia,
@@ -1875,7 +1875,6 @@ export function containsMany<FieldT extends FieldDefConstructor>(
         description,
         isUsed,
       });
-      // Save per-usage configuration on field descriptor when provided
       (instance as any).configuration = options?.configuration;
       return makeDescriptor(instance);
     },
@@ -1897,7 +1896,6 @@ export function contains<FieldT extends FieldDefConstructor>(
         description,
         isUsed,
       });
-      // Save per-usage configuration on field descriptor when provided
       (instance as any).configuration = options?.configuration;
       return makeDescriptor(instance);
     },
@@ -1912,15 +1910,15 @@ export function linksTo<CardT extends CardDefConstructor>(
   return {
     setupField(fieldName: string) {
       let { computeVia, description, isUsed } = options ?? {};
-      return makeDescriptor(
-        new LinksTo({
-          cardThunk: cardThunk(cardOrThunk),
-          computeVia,
-          name: fieldName,
-          description,
-          isUsed,
-        }),
-      );
+      let instance = new LinksTo({
+        cardThunk: cardThunk(cardOrThunk),
+        computeVia,
+        name: fieldName,
+        description,
+        isUsed,
+      });
+      (instance as any).configuration = options?.configuration;
+      return makeDescriptor(instance);
     },
   } as any;
 }
@@ -1933,15 +1931,15 @@ export function linksToMany<CardT extends CardDefConstructor>(
   return {
     setupField(fieldName: string) {
       let { computeVia, description, isUsed } = options ?? {};
-      return makeDescriptor(
-        new LinksToMany({
-          cardThunk: cardThunk(cardOrThunk),
-          computeVia,
-          name: fieldName,
-          description,
-          isUsed,
-        }),
-      );
+      let instance = new LinksToMany({
+        cardThunk: cardThunk(cardOrThunk),
+        computeVia,
+        name: fieldName,
+        description,
+        isUsed,
+      });
+      (instance as any).configuration = options?.configuration;
+      return makeDescriptor(instance);
     },
   } as any;
 }
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -191,6 +191,12 @@ export type FieldsTypeFor<T extends BaseDef> = {
 };
 export { formats, type Format };
 export type FieldType = 'contains' | 'containsMany' | 'linksTo' | 'linksToMany';
+// Opaque configuration passed to field format components and validators
+export type FieldConfiguration = Record<string, any>;
+// Configuration may be provided as a static object or a function of the parent instance
+export type ConfigurationInput<T> =
+  | FieldConfiguration
+  | ((self: Readonly<T>) => FieldConfiguration | undefined);
 export type FieldFormats = {
   ['fieldDef']: Format;
   ['cardDef']: Format;
@@ -206,6 +212,8 @@ interface Options {
   // in which case we need to tell the runtime that a card is
   // explicitly being used.
   isUsed?: true;
+  // Optional: per-usage configuration provider merged with FieldDef-level configuration
+  configuration?: ConfigurationInput<any>;
 }
 
 export interface CardContext<T extends CardDef = CardDef> {
@@ -332,6 +340,8 @@ export interface Field<
   fieldType: FieldType;
   computeVia: undefined | (() => unknown);
   description: undefined | string;
+  // Optional per-usage configuration stored on the field descriptor
+  configuration?: ConfigurationInput<any>;
   // there exists cards that we only ever run in the host without
   // the isolated renderer (RoomField), which means that we cannot
   // use the rendering mechanism to tell if a card is used or not,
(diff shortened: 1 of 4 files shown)


burieberry (approved):
Sounds good to me

habdelra (approved):
I like this approach. it seems like the right direction

lukemelia:
> Forward configuration when rendering card-valued fields

Addressed this 👍

PR #3473, opened 2025-10-24
ci: Add job to exercise migrations

backspace:
The goal is to catch things like #3467. You can see [here](https://github.com/cardstack/boxel/actions/runs/18775561514/job/53569336180) that the job failed when I reverted the change in #3467:

```
> PGDATABASE=boxel ./scripts/ensure-db-exists.sh && PGPORT=5435 PGDATABASE=boxel PGUSER=postgres node-pg-migrate --migrations-table migrations down 3

Database boxel exists
> Migrating files:
> - 1759459363190_create-realm-file-meta
> - 1759420000000_create-session-rooms
> - 1759232510123_create-claimed-domains-for-sites
### MIGRATION 1759459363190_create-realm-file-meta (DOWN) ###
DROP TABLE "realm_file_meta";
DELETE FROM "public"."migrations" WHERE name='1759459363190_create-realm-file-meta';


### MIGRATION 1759420000000_create-session-rooms (DOWN) ###
DROP TABLE IF EXISTS "session_rooms";
DELETE FROM "public"."migrations" WHERE name='1759420000000_create-session-rooms';


### MIGRATION 1759232510123_create-claimed-domains-for-sites (DOWN) ###
DROP INDEX "claimed_domains_for_sites_removed_at_index";
DROP INDEX "claimed_domains_for_sites_claimed_domains_for_sites_hostname_unique_index_index";
DROP INDEX "claimed_domains_for_sites_user_id_index";
DROP INDEX "claimed_domains_for_sites_source_realm_url_index";
DROP TABLE "claimed_domains_for_sites";
DELETE FROM "public"."migrations" WHERE name='1759232510123_create-claimed-domains-for-sites';


Error executing:
DROP INDEX "claimed_domains_for_sites_claimed_domains_for_sites_hostname_unique_index_index";
error: index "claimed_domains_for_sites_claimed_domains_for_sites_hostname_un" does not exist
```

When I restored that fix but added a comment within the migration to trigger the job, it [ran and passed](https://github.com/cardstack/boxel/actions/runs/18775808745/job/53570081999?pr=3473).


packages/postgres/scripts/ensure-db-exists.sh
```diff
@@ -1,5 +1,27 @@
...
+  done
+
+  echo "Timed out waiting for Postgres to become ready after $MAX_ATTEMPTS seconds" >&2
+  return 1
+}
+
+if ! wait_for_postgres; then
+  exit 1
+fi
+
```
backspace, line 24: I’m confused why this would be needed as other packages don’t have anything similar but without it I was seeing the job fail like [this](https://github.com/cardstack/boxel/actions/runs/18774342477/job/53565411587):

```
psql: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
```

Maybe other things were waiting in a different way?

.github/workflows/ci.yaml
```diff
@@ -131,6 +134,103 @@ jobs:
...
+          done <<< "$ALL_MIGRATIONS"
+
+          if [[ "$EARLIEST_INDEX" -eq 0 ]]; then
+            echo "Unable to locate changed migration $EARLIEST_BASENAME in migration list" >&2
+            exit 1
+          fi
+
+          DOWN_COUNT=$((TOTAL - EARLIEST_INDEX + 1))
+          echo "down_count=$DOWN_COUNT" >> "$GITHUB_OUTPUT"
+          echo "Will migrate down $DOWN_COUNT migration(s) to cover earliest changed migration"
```
lukemelia: Should put some of this in a script?

lukemelia (approved):

.github/workflows/ci.yaml
backspace: 👍🏻 it’s in `packages/postgres/scripts/determine-changed-migrations.sh` now, it requires some environment variable setup to make work, but certainly the job is easier to scan.

PR #3474, opened 2025-10-24
ci: Reduce Percy instability

backspace:
The top source of Percy instability is flaky tests and this only partially addresses that, but it’s better than nothing!

At Codex’s suggestion, this includes some D-Bus setup and a `dbus-run-session` wrapper around the test command. I know nothing about this but it’s meant to address the occasional failures we’ve seen like this and indeed since adding that I haven’t gotten it:

```
not ok 106 Chrome - [undefined ms] - error
    ---
        message: >
            Error: Browser timeout exceeded: 10s
            Error while executing test: Integration | card-copy: copy button appears when right and left stacks are index cards and there are mulitple selections on one side
            Stderr: 
             [5233:5267:1024/061509.940519:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
            [5233:5267:1024/061511.148588:ERROR:dbus/bus.cc:408] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")
```

This adds synchonous syntax highlighting calls to Monaco editors in the test waiter with the aim of preventing spurious diffs like this:

Build #17986 - Cardstack - Percy | Visual testing as a service 2025-10-24 11-24-58

It also hides this box shadow from Percy as it doesn’t render reliably:

Build #17982 - Cardstack - Percy | Visual testing as a service 2025-10-24 11-26-04

It adds another Poppins weight to ensure is loaded, as that’s the one used in these instances where the font is wrong:

Build #17969 - Cardstack - Percy | Visual testing as a service 2025-10-24 12-27-08


.github/workflows/ci-host.yaml
```diff
@@ -66,15 +66,25 @@ jobs:
...
+      - name: Install D-Bus helpers
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y dbus-x11 upower
+          sudo service dbus restart
+          sudo service upower restart
+
       - name: host test suite (shard ${{ matrix.shardIndex }})
-        run: pnpm test-with-percy
+        run: dbus-run-session -- pnpm test-with-percy
```
jurgenwerk, line 78: I haven't seen this before, and wanted to find out more:

>  Most CI runners (like GitHub Actions, GitLab CI, or CircleCI) use minimal, headless Linux environments. These lack a proper D-Bus session bus — which desktop applications often require — and they don’t have access to power management services like UPower.

(just posting this in case other reviewers find it interesting)

jurgenwerk (approved):

PR #3475, opened 2025-10-24
Eco 396 chips editor

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-396/chips-editor-component

image


tintinthong (approved):
Nice. This may evolve into different things with the new enum api. But, I think this is easy-enough to use as a component

PR #3479, opened 2025-10-27
Fix: reodering linksToMany items is broken

FadhlanR:
**Problem**

Dragging cards in a links-to-many or contains-many field reuses the same card instances in new positions. When card templates read data directly from `@model` (instead of staying on the boxed `<@fields>` path), Ember thinks the rendered DOM is still valid—so things like `author.thumbnailURL` stay stuck in their original slots even though the items moved.

**Solution**

After every reorder we now wrap each child box with a nonce-backed key and iterate those decorated entries. That forces Ember to refresh the child component inputs whenever the order changes, so even templates that use the raw `@model` values pick up the new position and re-render correctly.

Before:

https://github.com/user-attachments/assets/a81bc96f-fdf0-4c7c-9da8-588f8ed10832

After:

https://github.com/user-attachments/assets/b15cf276-9dcf-41f0-8218-d64683cae46f

### DIFF
--- a/packages/base/contains-many-component.gts
+++ b/packages/base/contains-many-component.gts
@@ -1,4 +1,5 @@
 import GlimmerComponent from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
 import { on } from '@ember/modifier';
 import { fn } from '@ember/helper';
 import {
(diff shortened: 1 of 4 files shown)


lukemelia (commented):
Could this be accomplished by setting a natural key like `id`?

FadhlanR:
> Could this be accomplished by setting a natural key like `id`?

Do you mean card id?

habdelra:
@codex review

lukemelia:
>> >> Could this be accomplished by setting a natural key like id?

>> Do you mean card id?

Yes

FadhlanR:
> > > > > Could this be accomplished by setting a natural key like id?
> 
> > > Do you mean card id?
> 
> Yes

It's not working because card id stays the same across reorders, so Ember would keep the existing component instance wired to that id no matter where it moves.

lukemelia (approved):

PR #3480, opened 2025-10-27
Assistant: Restore AI prompt input in case it wasn't sent, handle unknown message send errors

jurgenwerk:
This PR introduces 2 features, which try to address the frustration when something goes wrong when trying to send an AI message. These potential problems could include network connection problems, broken card serialization issues, and exceptions with uploading skills and commands. Previously, if these problems occurred, message sending would quietly fail, and the prompt the user wrote got lost. 

To communicate this better to the user, and to prevent prompt loss, there are now 2 new features:

1. AI Message drafts - if user writes a prompt but doesn't send it, and navigates away from the room, then the message will be persisted in the input field when user comes back, or refreshes the page. The prompt will also be preserved in case the message fails to be sent.

2. Generic error message when message sending fails due to reasons listed above

image


packages/host/app/components/matrix/room.gts
```diff
@@ -866,14 +887,13 @@ export default class Room extends Component<Signature> {
       myLastMessage.body,
       myLastMessage!.attachedCardIds || [],
       myLastMessage.attachedFiles,
-      true,
       myLastMessage.clientGeneratedId,
     );
```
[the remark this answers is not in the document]
jurgenwerk: I'll check this

packages/host/app/services/local-persistence-service.ts
```diff
@@ -41,6 +44,48 @@ export default class LocalPersistenceService extends Service {
...
+
+  setMessageDraft(roomId: string, message: string | undefined) {
+    let drafts = this.readMessageDrafts();
+    if (message && message.length > 0) {
+      drafts[roomId] = message;
+    } else {
+      delete drafts[roomId];
+    }
+    this.writeMessageDrafts(drafts);
+  }
```
lukemelia, line 70: do we want to expire drafts at some point or keep them around indefinitely?

packages/host/app/services/matrix-service.ts
```diff
@@ -995,6 +1012,7 @@ export default class MatrixService extends Service {
     clientGeneratedId = uuidv4(),
     context?: BoxelContext,
   ): Promise<void> {
+    // throw new Error('Failed to send message, deliberately');
```
lukemelia: Is this supposed to be left in?

lukemelia (approved):

packages/host/app/services/matrix-service.ts
jurgenwerk: Oops, forgot to remove it. Fixed

packages/host/app/services/local-persistence-service.ts
jurgenwerk: Yeah, would be good to delete older ones to prevent localstorage from growing too much. Addressed in https://github.com/cardstack/boxel/pull/3480/commits/ffbd9d56370e4ff5d762e7be7f42aead2863912c

lukemelia (approved):

PR #3492, opened 2025-10-29
Date related cards from base realm: do not crash when date string value is invalid

jurgenwerk:
Passing an invalid string value into our date related cards via code mode will cause an app exception and make the app unresponsive. This PR makes the date parsing part more robust, preventing bad values from crashing the app. I'm not sure if this requires a test since I didn't find any other dedicated tests for our base fields/cards. 

image

### DIFF
--- a/packages/base/date.gts
+++ b/packages/base/date.gts
@@ -1,5 +1,5 @@
 import { Component, primitive, FieldDef } from './card-api';
-import { parse } from 'date-fns';
+import { isValid, parse } from 'date-fns';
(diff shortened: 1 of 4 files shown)


backspace:
There are indeed [date/datetime tests](https://github.com/cardstack/boxel/blob/0d727373c2b646f9b8210a6f008adca45ddfb1c9/packages/host/tests/integration/components/card-basics-test.gts#L3529)

lukemelia (approved):

habdelra (approved):

PR #3498, opened 2025-10-29
Better client-side email validation behavior

burieberry:
- Validate email address on blur
- Error clears when typing again
- Provide more helpful error messages
- If field is not marked as `required`, can remove inputted value without validation error
~- If new value is invalid, the input is not submitted, and pre-existing value remains unchanged. **--> Would appreciate feedback about this. The other option is committing `null` to the `onChange` function if new value is invalid, which would override the previous value. There's also the issue of what to do if there was a pre-existing value and that value happens to be invalid. Trying to set null here throws ember invalidation error.**~
See update in comments below.

email-input


burieberry:
@codex

packages/boxel-ui/addon/src/components/input/email/index.gts
```diff
@@ -0,0 +1,116 @@
...
+  @action onBlur(): void {
+    this.hasBlurred = true;
+    this.debouncedInput.flush();
+    this.handleValidation(this.inputValue);
+  }
+
+  override willDestroy(): void {
+    this.debouncedInput.cancel();
+    super.willDestroy();
+  }
```
burieberry: This was recommended by codex. I'm not sure if it is really needed.

packages/boxel-ui/test-app/tests/integration/components/email-input-test.gts
```diff
@@ -0,0 +1,151 @@
...
+  settled,
+  triggerEvent,
+  typeIn,
+} from '@ember/test-helpers';
+
+import { EmailInput } from '@cardstack/boxel-ui/components';
+
+async function waitForDebounce() {
+  await new Promise<void>((resolve) => setTimeout(resolve, 350));
+  await settled();
```
backspace: Maybe we should have our own debounce that wraps the lodash one, because if we wire it into the Ember test waiter system, we won’t need things like this.

packages/boxel-ui/test-app/tests/integration/components/email-input-test.gts
```diff
@@ -0,0 +1,151 @@
...
+    await triggerEvent('[data-test-boxel-email-input]', 'blur');
+
+    assert
+      .dom('[data-test-boxel-email-input]')
+      .hasAttribute('data-test-boxel-input-validation-state', 'invalid');
+    assert
+      .dom('[data-test-boxel-input-error-message]')
+      .hasText('Enter an email address');
+    assert.strictEqual(value, null, 'required empty input not committed');
+  });
```
backspace, line 165: Great tests, reading through them gives me a tour of the API.

packages/boxel-ui/addon/src/components/email-input/usage.gts
```diff
@@ -0,0 +1,73 @@
+import { fn } from '@ember/helper';
+import { action } from '@ember/object';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import FreestyleUsage from 'ember-freestyle/components/freestyle/usage';
+
+import EmailInput from './index.gts';
+
+export default class EmailInputUsage extends Component {
```
backspace, line 9: Maybe you changed the name at some point but I think it should be one or the other:

Image

packages/boxel-ui/addon/src/components/input/email/index.gts
backspace: I’ve seen this a lot lately, IMO the `super` calls are almost never needed

backspace (changes requested):
on the question of what to do with invalid values, what if consumers could include a second parameter for `onChange` that says whether the value is validated or not? And the value is always sent regardless of whether it’s valid. Then consumers can decide whether to use the value or ignore it.

I feel like there must be best practices in this though 🤔

packages/boxel-ui/addon/src/components/email-input/usage.gts
burieberry: I think the first one is autogenerated by combining the directory names. I can put it outside the `input` directory or fix the generated name.

backspace: oh hmm… maybe it’s fine then, I misunderstood, sorry

it would be good to have the correct import to use shown here (and everywhere in Freestyle) but that’s not a problem with your PR

packages/boxel-ui/test-app/tests/integration/components/email-input-test.gts
burieberry: Ended up using `debounce` from `'@ember/runloop'` instead of the `lodash` one. This one works with the built-in test helpers. https://guides.emberjs.com/release/testing/testing-components/#toc_waiting-on-asynchronous-behavior

burieberry:
Main updates:

1- Added `validation` and `event` parameters to the onChange function:
- value is set on input entry
- `value`, `validation` with error message, `event` params are available to the end consumer
- error UI is only surfaced on blur

2- Updated the `EmailField` in `base/email.gts`. Added some basic data massaging so that when an invalid value like `{}` is entered via the code editor, the card does not crash. We will still need to handle server-side validation, though.

backspace (approved):
nice, I’m glad there was an existing waiter-compatible debounce to use 🎉

PR #3505, opened 2025-10-30
add query option to linksTo/linksToMany

lukemelia:
https://github.com/user-attachments/assets/d4849a3b-1884-4f1c-9b03-c10e7891bb55

# Query-Backed Relationship Fields

Objective: introduce a new syntax for `linksTo` and `linksToMany` fields that define the field in terms of a query that is run against a realm's search index and the cards returned from the query become the cards in the field. It is similar to a computed field, but instead of defining the field with a TypeScript or JavaScript function, we define it with a query.

## Goals

- Provide a declarative mechanism for relationship fields whose contents are derived from search queries.
- Avoid bespoke imperative resolvers while still allowing fields to react to card data and cross-realm content.
- Ensure query-backed fields participate correctly in card serialization (JSON:API), invalidation, and live updates.
- Establish guardrails that keep indexes consistent and keep query execution predictable under load.

## Non-Goals (Initial Release)

- Indexing or searching on the derived relationship content.
- Waiting for query execution during prerender to unblock static HTML generation.
- Defining new filter/query operators beyond what is covered in `docs/search.md`.
- Supporting imperative mutation of the query output (all results are read-only projections).
- Federated multi-realm query execution.

## Terminology

- **Query field**: a `linksTo` or `linksToMany` relationship whose value is resolved by running a search query.
- **Host realm**: the realm where the card instance lives when the query field resolves (may differ from the realm that defines the card).
- **Result card**: a card returned by the search query.
- **Query context**: runtime values (`$this`, `$card`, `$thisRealm`, etc.) that can be interpolated into the query definition.

## Supported Field Shapes

Query-backed relationships are only supported on the relationship helpers that reference other cards:

- `linksTo(CardType, { query: ... })`
- `linksToMany(CardType, { query: ... })`

Other helper signatures (e.g. `containsMany`) remain unsupported for the initial implementation.

## Declaration Syntax

A query definition mirrors the `SearchIndex#search` API (see `docs/search.md`). Example:

```ts
@field shirts = linksToMany(Shirt, {
  query: {
    filter: {
      // implicit clause merged during execution: on: { module: Shirt.module, name: 'Shirt' }
      eq: { size: '$this.profile.shirtSize' },
    },
    realm: '$thisRealm',
    sort: [
      {
        by: 'updatedAt',
        direction: 'desc',
      },
    ],
    page: { size: 12 },
  },
});

@field profile = linksTo(Profile, {
  query: {
    filter: {
      eq: { primary: true },
    },
    // `linksTo` takes the first matching card (post-sort) or null when no results.
  },
});
```

### Implicit `on` Clause

During execution we merge an implicit `on` filter targeting the referenced card type. Implementations should not require the author to restate the type module and name if it matches the field type declaration.

The `realm` option takes a single string and must resolve to one realm URL after interpolation. Supplying an array or multiple realms is invalid and fails validation.

### Query Object Properties

| Property | Supported | Notes                                                                                                    |
| -------- | --------- | -------------------------------------------------------------------------------------------------------- |
| `filter` | ✅        | Follows `docs/search.md`. Supports interpolation tokens.                                                 |
| `sort`   | ✅        | Matches search spec. Required `on` entries can be inferred when omitted if they refer to the field type. |
| `page`   | ✅        | Only `size` is honored for `linksToMany`; `number` defaults to 0. For `linksTo`, `size` is forced to 1.  |
| `realm`  | ✅        | Determines search index scope. Defaults to `$thisRealm`.                                                 |
| `limit`  | ❌        | Use `page.size` instead.                                                                                 |
| `cursor` | ❌        | Not supported; queries are re-run fresh.                                                                 |

## Query Context and Interpolation

We support simple string interpolation tokens for values:

- `$this.<path>`: resolved against the card instance that owns the field.
- `$thisRealm`: resolves to the host realm URL.

Interpolations are resolved against the serialized JSON:API attributes of the owning card (plus its `id`) on the server. Client-side recomputation uses the in-memory card instance so local edits can trigger fresh searches before persistence. During compilation we verify each path segment we can statically reason about—walking through local `contains` / `containsMany` fields (requiring numeric indexes for `containsMany`) and ensuring their descendants exist. Paths may continue through relationship fields; we allow those references even though interpolation still relies on the owning card’s serialized attributes at runtime. When an interpolation produces `undefined` we drop that value. If the removal empties an `eq`, `contains`, `range`, `any`, or `every` predicate we skip executing the query entirely, leaving the relationship empty and omitting `links.search`. For every other array or object we simply prune the empty node and continue evaluating the rest of the query tree.

## Execution Semantics

1. The card loader evaluates the query definition during card resolution.
2. Interpolation tokens are replaced with actual values using the in-memory representation of the card.
3. The query is executed against the search index of the single specified realm. We do not support federated execution across multiple realms.
4. The response is normalized into card references:
   - `linksTo` returns the first card from the result list (or `null` when no matches).
   - `linksToMany` returns the entire result list (minus duplicate IDs). The `page.size` caps the number of cards returned by the realm.
5. The resolved cards are materialized into the JSON:API document returned by the server, including `included` resources for dependent cards.
6. Each response includes a `links.search` URL that captures the fully interpolated query; this URL is the canonical record of what executed.

The query executes server-side. Resolved card IDs are cached in the field payload to avoid re-running queries during the same request. Subsequent requests use invalidation rules (see below) to determine whether the query must be re-run.

## Serialization

- Query field results appear in the relationship data exactly like explicitly linked cards.
- When we successfully execute a query we populate `links.search` with the interpolated query URL and include `data` so the client can reuse the same results. If normalization short-circuits before execution (because an interpolation cleared the predicate), we skip the search entirely, omit `links.search`, and leave the relationship empty without `data`.
- Each response includes a `links.search` URL for every query-backed relationship whose query executed, even when the result set is empty. That URL is the canonical representation of the last query and is sufficient for seeding the client cache.
- Clients must be able to losslessly convert `links.search` back into the normalized query object so they can compare it against newly interpolated queries. Runtime helpers provide this round-trip conversion.
- `included` resources contain the full JSON:API representation of result cards so clients do not need to issue follow-up requests.
- The client can detect read-only query fields via the schema (`linksTo` / `linksToMany` declarations with a `query` option), so no additional JSON:API metadata is required for Phase 4.
- During hydration we reuse the server-provided `links.search` + relationship `data` (when present) instead of automatically re-running the query. Client refreshes only occur when the interpolated signature changes or when an invalidation triggers a refresh.

## Caching, Invalidation, and Seeds

- Query-backed fields are backed by a `SearchResource`. All previous `LiveQuery` references in client runtime docs are replaced with this resource.
- The resource is created per field instance (no global dedupe by signature). It skips a fetch when the normalized query string and realm list match the previous values but otherwise issues a fresh search for each live field.
- `CardStore#createSearchResource` constructs resources. The parent argument is the actual card instance (not an ID), and destroying the parent tears down the resource.
- GC must call `destroy` on every instance it releases; cache/index cleanup remains in the store.
- `getSearch` accepts an optional `seed` argument shaped `{ cards: CardDef[], searchURL?: string, meta?: QueryResultsMeta, errors?: ErrorEntry[] }`. Seed data is deserialized into the store before `getSearch` runs and is authoritative for the initial render (no network request). Seeds apply once per resource; subsequent refreshes ignore them.
- When tracked local fields change, autotracking reruns the getter and rebuilds the query. If the newly built search URL differs from the cached signature, the resource refreshes; otherwise it reuses the cached payload.
- Realm invalidations trigger an immediate refresh for each live query-backed field targeting that realm; we do not mark them stale and wait for the next access.
- Resources expose `instances` plus optional `meta`/`errors` for UI consumption; reuse existing loading/error metadata if already available.

## Live Updates

- Realm clients already receive invalidation events. When a source realm reports invalidations and there are query fields that target that realm, those fields are marked dirty so Glimmer's auto-tracking invalidates naturally on the next Ember run loop turn (no additional sequencing guarantees required).
- We do **not** stream incremental diff updates for the query results in the initial release.
- Invalidation flow:
  - Collect query-field metadata (field id, source realm, canonical search URL) when queries execute so we can quickly look up affected fields.
  - On an invalidation event, each live query-backed field whose `realm` matches the source realm immediately refreshes its search (one fetch per field/query).
  - Applies to both `linksTo` and `linksToMany` because the resource exposes both `instances` (array) and `instance` (first item/null).

## Authorization

- Search results are filtered according to the requesting principal's permissions. This is the same behavior as direct search calls.
- Because each query targets exactly one realm, credential propagation matches the standard single-realm search flow—there is no cross-realm fallback path.
- If the realm rejects the request, the field resolves to an empty relationship and we log a warning so operators can diagnose access issues.

## Error Handling

- Compilation only validates that the head segment of each `$this.*` interpolation exists on the card. Nested fields and relationship dereferences are evaluated at runtime and, when unresolved, simply cause the affected predicate (and sometimes the entire query) to be skipped without throwing.
- When normalization aborts because a predicate emptied out, we short-circuit the field to an empty relationship, omit `links.search`, and do not emit a warning.
- Runtime search errors (network issues, realm offline, or authorization failures) add entries under `relationship.meta.errors`. We log these at debug level; an additional warning is emitted only when every failing realm response was unauthorized.

## Interaction with Prerendering

- During prerender we do **not** block on query execution. The prerendered HTML excludes query-backed relationship content.
- Hydration reuses the server-populated query relationship data (when present) and only re-runs the query if the computed signature changes or an invalidation has fired.
- The two-phase render would add all cards in the batch to the indexer and then render for HTML capture.

### Schema & Compilation

- Extend the card field decorator typings to accept `query` for `linksTo` / `linksToMany`.
- Validate that `query` conforms to the search schema, enforcing implicit `on` merges.
- Statically analyze interpolation tokens to ensure referenced fields exist on the card.

### Server Runtime

- Enhance the field resolver pipeline to detect query-backed relationships and call the search index.
- Add a query execution helper that handles interpolation, realm resolution, deduping, and pagination semantics without adding extra global pagination.
- Extend card JSON:API assembly to attach result cards to `included`, populate `links.search` with the interpolated query URL, ensure the URL can be converted back into the normalized query object, and record dependency metadata for invalidation.
- Implement error and authorization fallbacks as described.

### Client Runtime

- No changes required for consumers reading the JSON:API response.
- `CardStore#createSearchResource` builds `SearchResource` instances. The parent argument is the actual card instance; destroying the parent tears down the resource.
- `SearchResource` exposes `instances`, plus any existing `meta`/`errors` and loading status we already surface. The field runtime uses `resource.instances` to render the relationship.
- Signature comparison uses the canonical search URL derived from the interpolated query. When the signature changes, the resource refreshes; otherwise it reuses cached data.
- `getSearch` accepts an optional `seed` argument `{ cards: CardDef[], searchURL: string }`. Seed records are deserialized before `getSearch` runs and power the initial render without a network request. Seeds are ignored after the resource becomes stale.
- Optional: expose richer metadata later if the schema alone proves insufficient for UX needs.

### Invalidation & Refresh

- Search resources subscribe to the realm invalidation stream; each records `{ cardId, fieldName, realmHref, searchURL }` and marks itself stale when the realm matches the invalidation event.
- Refreshes dedupe per `{ realmHref, searchURL }`, and results fan out to every field using the same resource.
- The query-field runtime reads the post-refresh payload via the resource, updates the shared state cache (signature, relationship data, `links.search`), and clears the stale mark so repeated reads within a tick reuse the hydrated results.
- Applies to both `linksTo` and `linksToMany` because the resource exposes both `instances` (array) and `instance` (first item/null).

### Testing

- Unit tests for interpolation resolution and validation failures.
- Integration tests that ensure linksTo/linksToMany query fields hydrate correctly and respect sorting/pagination.
- Regression tests ensuring query-backed fields are excluded from prerendered HTML yet appear post-hydration.

## Future Enhancements

- Make query results searchable.
- Implement eager invalidations with explicit opt-in syntax.
- Provide tooling that previews the live query during card authoring.
- Support cursor-based pagination for dynamic, infinite-scroll experiences.
- Reintroduce federated execution across multiple realms with clear ergonomics.
- Prerendered HTML capture always includes query results and/or allow authors to opt in/out of blocking prerender for specific query fields
- Visualize field query in the Code Mode Schema Editor

## FAQ

1. **Are query-based relationships searchable?**
   - Not initially. Including them in search docs requires dependency-aware invalidation.
2. **Do we cache/assemble query results on the server when requesting a card and include them in the JSON API response?**
   - Yes. This avoids N+1 client fetches.
3. **Do the query results show in prerendered HTML?**
   - Yes.
4. **Do we re-run queries on client upon realm invalidations?**
   - Yes. We leverage the same invalidation flow used by prerendered searches.


packages/base/card-api.gts
```diff
@@ -982,6 +1011,12 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
     visited: Set<string>,
     opts?: SerializeOpts,
   ) {
+    if (this.queryDefinition) {
+      let docMeta = (doc.data.meta = doc.data.meta ?? {});
+      let queryMeta = (docMeta as any).queryFields ?? {};
+      (docMeta as any).queryFields = queryMeta;
+      queryMeta[this.name] = JSON.parse(JSON.stringify(this.queryDefinition));
+    }
```
lukemelia: Do we need this?

packages/base/card-api.gts
```diff
@@ -1223,8 +1258,11 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
       format: Format | undefined,
       defaultFormat: Format,
       isComputed: boolean,
+      isQueryBacked: boolean,
```
lukemelia: Let's just have isComputed here

packages/base/card-api.gts
lukemelia:
```suggestion
    let isComputed = !!this.computeVia || !!this.queryDefinition;
```

packages/base/card-api.gts
```diff
@@ -1223,8 +1258,11 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
...
     ) {
-      return (format ?? defaultFormat) === 'edit' && !isComputed;
+      return (
+        (format ?? defaultFormat) === 'edit' && !isComputed && !isQueryBacked
```
lukemelia:
```suggestion
        (format ?? defaultFormat) === 'edit' && !isComputed
```

packages/base/card-api.gts
```diff
@@ -1256,7 +1294,12 @@ class LinksTo<CardT extends CardDefConstructor> implements Field<CardT> {
         <CardCrudFunctionsConsumer as |cardCrudFunctions|>
           <DefaultFormatsConsumer as |defaultFormats|>
             {{#if
-              (shouldRenderEditor @format defaultFormats.cardDef isComputed)
+              (shouldRenderEditor
+                @format
+                defaultFormats.cardDef
+                isComputed
+                (if linksToField.queryDefinition true false)
```
lukemelia:
```suggestion
                isComputed
```

packages/base/card-api.gts
```diff
@@ -1438,6 +1482,12 @@ class LinksToMany<FieldT extends CardDefConstructor>
     visited: Set<string>,
     opts?: SerializeOpts,
   ) {
+    if (this.queryDefinition) {
+      let docMeta = (doc.data.meta = doc.data.meta ?? {});
+      let queryMeta = (docMeta as any).queryFields ?? {};
+      (docMeta as any).queryFields = queryMeta;
+      queryMeta[this.name] = JSON.parse(JSON.stringify(this.queryDefinition));
+    }
```
lukemelia: Do we need this?

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,304 @@
...
+      let referencedField = getField(ownerClass, head, { untracked: true });
+      if (!referencedField) {
+        let ownerName = ownerClass.name ?? 'Card';
+        throw new Error(
+          `query field "${fieldName}" references unknown path "${token}" on ${ownerName}`,
+        );
+      }
+    }
+  }
+}
```
lukemelia, line 211: Let's make this fully validate by walking down into additional "." segments using getField.

packages/base/card-api.gts
```diff
@@ -1846,7 +1899,7 @@ export const field = function (
   key: string | symbol,
   { initializer }: { initializer(): any },
 ) {
-  let descriptor = initializer().setupField(key);
+  let descriptor = initializer().setupField(key, target as BaseDef);
```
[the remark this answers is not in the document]
lukemelia: Consider separate validation to ensure @field is being used only inside BaseDef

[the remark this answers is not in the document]
lukemelia: Otherwise ignore this complaint

packages/base/card-api.gts
```diff
@@ -1866,7 +1916,7 @@ export function containsMany<FieldT extends FieldDefConstructor>(
   options?: Options,
 ): BaseInstanceType<FieldT>[] {
   return {
-    setupField(fieldName: string) {
+    setupField(fieldName: string, _ownerPrototype?: BaseDef) {
```
lukemelia:
```suggestion
    setupField(fieldName: string, _ownerPrototype: BaseDef) {
```

packages/base/card-api.gts
```diff
@@ -1887,7 +1937,7 @@ export function contains<FieldT extends FieldDefConstructor>(
   options?: Options,
 ): BaseInstanceType<FieldT> {
   return {
-    setupField(fieldName: string) {
+    setupField(fieldName: string, _ownerPrototype?: BaseDef) {
```
lukemelia:
```suggestion
    setupField(fieldName: string, _ownerPrototype: BaseDef) {
```

packages/base/card-api.gts
```diff
@@ -1905,17 +1955,21 @@ contains[fieldType] = 'contains' as FieldType;
 
 export function linksTo<CardT extends CardDefConstructor>(
   cardOrThunk: CardT | (() => CardT),
-  options?: Options,
+  options?: RelationshipOptions,
 ): BaseInstanceType<CardT> {
   return {
-    setupField(fieldName: string) {
-      let { computeVia, description, isUsed } = options ?? {};
+    setupField(fieldName: string, ownerPrototype?: BaseDef) {
```
lukemelia:
```suggestion
    setupField(fieldName: string, ownerPrototype: BaseDef) {
```

packages/base/card-api.gts
```diff
@@ -1926,17 +1980,21 @@ linksTo[fieldType] = 'linksTo' as FieldType;
 
 export function linksToMany<CardT extends CardDefConstructor>(
   cardOrThunk: CardT | (() => CardT),
-  options?: Options,
+  options?: RelationshipOptions,
 ): BaseInstanceType<CardT>[] {
   return {
-    setupField(fieldName: string) {
-      let { computeVia, description, isUsed } = options ?? {};
+    setupField(fieldName: string, ownerPrototype?: BaseDef) {
```
lukemelia:
```suggestion
    setupField(fieldName: string, ownerPrototype: BaseDef) {
```

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,304 @@
...
+
+export function seedQueryFieldState(
+  instance: CardDef,
+  resource: LooseCardResource,
+): void {
+  let queryFieldsMeta =
+    ((resource.meta as any)?.queryFields as Record<string, unknown>) ?? {};
```
lukemelia: Let's introspect on the card's fields to know this instead of needing meta.queryFields in the JSON payload

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,304 @@
...
+        try {
+          let parsed = parseQuery(queryString);
+          assertQuery(parsed);
+          normalizedQuery = normalizeQueryForSignature(parsed as Query);
+          signature = querySignature(normalizedQuery);
+        } catch {
+          normalizedQuery = undefined;
+          signature = undefined;
+        }
+      }
```
lukemelia: Avoid needing to do this by ensuring that the server generates a consistently ordered search URL when serializing to JSON API links.search node and then generate a search URL in the same way from the runtime query to compare

packages/runtime-common/query.ts
```diff
@@ -472,3 +490,44 @@ export const parseQuery = (queryString: string) => {
...
+
+  return cloned;
+}
+
+export function querySignature(query: Query | undefined): string | undefined {
+  if (!query) {
+    return undefined;
+  }
+  return JSON.stringify(sortKeysDeep(query));
+}
```
lukemelia: we're going to get rid of this in favor of using the URL as the staleness key

packages/host/app/services/query-field-coordinator.ts
```diff
@@ -0,0 +1,558 @@
...
+      );
+    }
+    let deduped = this.#dedupeResults(json.data);
+    return {
+      cards: deduped,
+      included: [
+        ...deduped,
+        ...((json.included as CardResource[] | undefined) ?? []),
+      ],
+    };
```
lukemelia: Use the store's search method for this. That will give us back card instances that are already loaded into the store

packages/host/app/services/query-field-coordinator.ts
```diff
@@ -0,0 +1,558 @@
...
+        }
+        let otherState = api.getQueryFieldState(payload.instance, fieldName);
+        if (otherState?.relationship) {
+          relationships[fieldName] = cloneRelationship(
+            otherState.relationship,
+          )!;
+        }
+      }
+    }
+
```
lukemelia: Update applyResults to accept a collection of card instances instead of a payload. Figure out how to get them into the field's "value" safely

lukemelia: careful not to trigger autosave but we want to trigger card tracking

packages/base/links-to-many-component.gts
```diff
@@ -456,8 +456,9 @@ function shouldRenderEditor(
   format: Format | undefined,
   defaultFormat: Format,
   isComputed: boolean,
+  isQueryBacked: boolean,
```
lukemelia: Just use isComputed and define it based on queryDefinition

packages/host/app/services/query-field-coordinator.ts
```diff
@@ -0,0 +1,558 @@
...
+type PendingMap = Map<string, Promise<void>>;
+type FieldHandle = { instance: CardDef; fieldName: string };
+type FieldRealmEntry = { realm?: string; handle?: FieldHandle };
+type RealmRegistration = {
+  unsubscribe: () => void;
+  fields: Set<FieldHandle>;
+};
+
+export default class QueryFieldCoordinatorService
+  extends Service
```
lukemelia: Consider replacing this by introducing "live queries" to Store service

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,304 @@
...
+  queryFieldCoordinator = coordinator;
+}
+
+export function notifyQueryFieldAccess(instance: CardDef, field: Field) {
+  if (!field.queryDefinition || !queryFieldCoordinator) {
+    return;
+  }
+  if (isQueryFieldEvaluationInProgress(instance, field.name)) {
+    return;
+  }
```
lukemelia: this is likely wrong, because we would want to cancel and restart if the inputs changed

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,304 @@
...
+export interface QueryFieldCoordinator {
+  handleQueryFieldAccess(payload: QueryFieldAccessPayload): void;
+}
+
+let queryFieldCoordinator: QueryFieldCoordinator | undefined;
+
+export function registerQueryFieldCoordinator(
+  coordinator: QueryFieldCoordinator | undefined,
+) {
+  queryFieldCoordinator = coordinator;
```
lukemelia: This will all be replaced with a new Store interface that offers live query bits

packages/host/app/components/operator-mode/card-schema-editor.gts
```diff
@@ -546,6 +568,15 @@ export default class CardSchemaEditor extends Component<Signature> {
...
   }
 
+  @action
+  canManipulateField(field: FieldOfType) {
+    return (
+      this.args.allowFieldManipulation &&
+      !field.isComputed &&
+      !field.isQueryField
+    );
+  }
```
lukemelia: Is this correct? Do we want to limit "manipulation" for query fields?

packages/host/tests/helpers/base-realm.ts
```diff
@@ -93,6 +93,7 @@ let serializeCard: (typeof CardAPIModule)['serializeCard'];
 let isSaved: (typeof CardAPIModule)['isSaved'];
 let relationshipMeta: (typeof CardAPIModule)['relationshipMeta'];
 let getQueryableValue: (typeof CardAPIModule)['getQueryableValue'];
+let getQueryFieldState: (typeof CardAPIModule)['getQueryFieldState'];
```
lukemelia: I don't think we need this anymore

packages/host/tests/helpers/index.gts
```diff
@@ -831,9 +860,50 @@ export async function saveCard(
...
+) {
+  if (!id) {
+    return;
+  }
+  let url = new URL(id);
+  let registry = getTestRealmRegistry();
+  let matching = [...registry.values()].find(({ realm }) =>
+    realm.paths.inRealm(url),
+  );
+  if (!matching) {
```
lukemelia, line 1038: Do we need this registry thing?

packages/host/tests/integration/components/serialization-test.gts
```diff
@@ -3647,6 +3656,467 @@ module('Integration | serialization', function (hooks) {
...
+    );
+
+    assert.strictEqual(
+      getQueryFieldState(queryCard, 'favorite')?.relationship?.links?.self,
+      `${testRealmURL}Person/backup`,
+      'query field state reflects refreshed relationship data',
+    );
+  });
+
+  test('store.refreshQueryField forces a refresh with the latest results', async function (assert) {
```
lukemelia: Consider waiting for someone to ask for this feature

packages/runtime-common/index-structure.ts
```diff
@@ -57,6 +58,7 @@ export interface FieldDefinition {
   isComputed: boolean;
   fieldOrCard: CodeRef;
   serializerName?: SerializerName;
+  query?: Query;
```
lukemelia: make sure this is used

packages/runtime-common/index.ts
```diff
@@ -428,10 +400,10 @@ export interface Store {
   patch<T extends CardDef>(
     id: string,
     patchData: PatchData,
-    opts?: { doNotPersist?: boolean; clientRequestId?: string },
   ): Promise<T | CardErrorJSONAPI | undefined>;
   search(query: Query, realmURL?: URL): Promise<CardDef[]>;
   getSaveState(id: string): AutoSaveState | undefined;
+  refreshQueryField(card: CardDef, fieldName: string): Promise<void>;
```
lukemelia: remove until need is established

packages/runtime-common/query-field-utils.ts
```diff
@@ -0,0 +1,292 @@
...
+    throw new Error(
+      `query field "${fieldName}" only supports a single realm but received multiple entries`,
+    );
+  }
+  let interpolatedRealm = interpolateNode(realmsList, 'realm');
+  if (interpolatedRealm !== undefined) {
+    realmsList = interpolatedRealm;
+  }
+  delete queryAny.realm;
+  delete queryAny.realms;
```
lukemelia: delete this line

packages/runtime-common/resource-types.ts
```diff
@@ -70,6 +71,11 @@ export interface CardResource<Identity extends Unsaved = Saved> {
   };
 }
 
+export type LooseCardResource = Omit<CardResource, 'id' | 'type'> & {
+  type?: 'card';
+  id?: string;
+};
+
```
lukemelia, line 78: doesn't this exist already?

packages/runtime-common/query-field-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+    if (current == null) {
+      return undefined;
+    }
+    if (Array.isArray(current)) {
+      let index = Number(segment);
+      if (!Number.isInteger(index)) {
+        return undefined;
+      }
+      current = current[index];
+      continue;
```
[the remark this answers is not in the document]
lukemelia: Ignoring...

packages/host/tests/helpers/base-realm.ts
lukemelia: using it in serialization test currently

packages/base/query-field-support.ts
lukemelia: Review this with @ef4

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,614 @@
...
+  }
+  return {
+    type: field.fieldType,
+    isPrimitive: false,
+    isComputed: !!field.computeVia,
+    fieldOrCard: ref,
+  };
+}
+
+function syncQueryFieldStateFromResource(
```
lukemelia: Let's try to just use this resource directly

packages/host/app/lib/gc-card-store.ts
```diff
@@ -530,6 +564,40 @@ export default class CardStoreWithGarbageCollection implements CardStore {
...
+        | {
+            cards: CardDef[];
+            searchURL?: string;
+            meta?: QueryResultsMeta;
+            errors?: ErrorEntry[];
+          }
+        | undefined;
+    },
+  ) {
+    if (!this.#storeHooks?.getSearchResource) {
```
lukemelia, line 572: Is this a legit case that can happen?

packages/host/app/resources/search.ts
```diff
@@ -1,46 +1,49 @@
...
-
 const waiter = buildWaiter('search-resource:search-waiter');
-
 export interface Args {
   named: {
     query: Query | undefined;
     realms: string[] | undefined;
     isLive: boolean;
+    isAutoSaved?: boolean;
+    storeService?: StoreService;
```
lukemelia, line 42: Why this and not using @service store?

packages/host/app/resources/search.ts
```diff
@@ -82,30 +93,66 @@ export class SearchResource extends Resource<Args> {
...
-    this.loaded = this.search.perform(query);
-
+    if (seed && !this.#seedApplied) {
+      this.applySeed(seed);
+      this.#seedApplied = true;
+      this.loaded = Promise.resolve();
+      this.#log.info(
+        `apply seed for search resource (one-time); count=${seed.cards.length}; searchURL=${seed.searchURL}`,
+      );
+    }
```
lukemelia, line 126: Move this to the beginning of the method, before conditional check. Set previousQuery from seed data's search URL

packages/host/app/resources/search.ts
```diff
@@ -82,30 +93,66 @@ export class SearchResource extends Resource<Args> {
...
+    // Only kick off a fetch if we didn't just seed; seeding is the initial state.
+    // Defer to afterRender to avoid mutating tracked state during render.
+    if (!seed || this.#seedApplied) {
```
lukemelia: This guard will not be needed

packages/host/app/services/message-service.ts
```diff
@@ -11,22 +11,36 @@ export default class MessageService extends Service {
     new Map();
   @service declare private network: NetworkService;
 
+  constructor() {
+    super(...arguments);
+    this.register();
```
lukemelia: This is also called in store constructor and operatormode component. Rethink this.

habdelra:
+11,000 lines of code? really? is there a bad merge with the headless chrome indexing refactor that added back a bunch of stuff?

mfa.sh
habdelra, line 1: did you mean to commit this?

.env
habdelra, line 1: did you mean to commit this?

packages/host/app/components/card-prerender.gts
```diff
@@ -314,22 +319,44 @@ export default class CardPrerender extends Component {
       };
       if (shouldClearCache) {
         initialRenderOptions.clearCache = true;
-        this.loaderService.resetLoader({
-          clearFetchCache: true,
-          reason: 'module-prerender clearCache',
-        });
-        this.store.resetCache();
       } else {
```
habdelra, line 322: why?

packages/host/app/routes/module.ts
```diff
@@ -125,294 +152,338 @@ export default class ModuleRoute extends Route<Model> {
     options?: string;
   }) {
     let parsedOptions = parseRenderRouteOptions(options);
-    // Make it easy for Puppeteer to do regular Ember transitions
-    (globalThis as any).boxelTransitionTo = (
-      ...args: Parameters<RouterService['transitionTo']>
-    ) => {
-      this.router.transitionTo(...args);
```
habdelra, line 135: why did you do this? this is the machinery that lets the prerenderer reuse chrome pages

packages/host/app/components/card-prerender.gts
```diff
@@ -63,6 +72,9 @@ export default class CardPrerender extends Component {
   #renderErrorPayload: string | undefined;
   #cardTypeTracker = new RenderCardTypeTracker();
   #currentContext: CardRenderContext | undefined;
+  #moduleTypesCache: ModuleTypesCache = new WeakMap() as ModuleTypesCache;
+  #moduleLastStoreResetKey: string | undefined;
+  #moduleAuthGuard = createAuthErrorGuard();
```
habdelra, line 77: why do we need this? card-prerender is for host rendering only i don't think we run into auth errors in our host tests that would require this

packages/host/app/components/card-prerender.gts
```diff
@@ -63,6 +72,9 @@ export default class CardPrerender extends Component {
   #renderErrorPayload: string | undefined;
   #cardTypeTracker = new RenderCardTypeTracker();
   #currentContext: CardRenderContext | undefined;
+  #moduleTypesCache: ModuleTypesCache = new WeakMap() as ModuleTypesCache;
```
habdelra, line 75: i mean this is just for host tests--how much does this really matter? my goal for this component is to keep it as lean as possible since it's not for the real app, it's just to support indexing in tests. the more logic we add here the more our test burden and the more the possibility that we see bugs in test indexing specifically that don't exist in real app indexing. when it is at all possible also consider sharing logic between this and the real /render and /module routes, so that we reduce the likelihood of testing only logic here.

packages/runtime-common/realm.ts
```diff
@@ -2606,8 +2612,13 @@ export class Realm {
       throw e;
     }
 
+    let useWorkInProgressIndex = Boolean(
+      request.headers.get('X-Boxel-Building-Index'),
+    );
+
```
habdelra: this is dead now. i'm concerned about your upstream merges if you need this.

packages/host/app/routes/module.ts
```diff
@@ -125,294 +152,338 @@ export default class ModuleRoute extends Route<Model> {
...
+    status: 'ready' as const,
+    deps,
+    lastModified,
+    createdAt,
+    isShimmed,
+    definitions,
+  };
+}
+
+async function makeDefinition(
```
habdelra, line 341: what happened that required this to be public API?

packages/host/app/routes/module.ts
```diff
@@ -125,294 +152,338 @@ export default class ModuleRoute extends Route<Model> {
...
+      {
+        id,
+        nonce,
+        renderOptions: parsedOptions,
+      },
+      this.#moduleModelContext(),
+    );
+  }
+
+  #moduleModelContext(): ModuleModelContext {
```
habdelra, line 168: tell me more about the motivation for using a context here. is there a reason why you'd do this for /module routes but not /render routes?

packages/host/app/services/render-service.ts
```diff
@@ -85,6 +85,20 @@ export class CardStoreWithErrors implements CardStore {
   async loaded() {
     await Promise.allSettled(this.#inFlight);
   }
+  getSearchResource<T extends CardDef = CardDef>(
```
habdelra, line 88: why was this necessary? is there an interface that this service is fulfilling that required this?

lukemelia:
@habdelra wrote:

> +11,000 lines of code? really? is there a bad merge with the headless chrome indexing refactor that added back a bunch of stuff?

I inadvertently committed a bunch of docs. Fixed now. down to ~+4K

mfa.sh
lukemelia: Fixed

.env
lukemelia: Fixed

packages/runtime-common/realm.ts
lukemelia: Fixed

packages/host/app/routes/module.ts
lukemelia: We ran into a timing problem when the module route was being used inside the test process (i.e. while the app was also routing). I think this was due to routing triggering card loading triggering server-side search for query fields triggering module definition lookup. We initially worked around this by directly invoking the `model` method on the route instead of using `recognizeAndLoad` and that avoided the problem. However, it read very hacky, so I extracted the logic that the model method performed to functions that can be used from the card-prerender component or from the module route. The render path doesn't have this problem since it is not triggered from the same code paths that lookup definitions.

packages/host/app/routes/module.ts
lukemelia: It moved from being a private method on the route class to being a non-exported module scope function. No real change in API privacy level.

packages/host/app/services/render-service.ts
lukemelia: Yes, the `CardStore` interface

packages/host/app/routes/module.ts
lukemelia: It is moved down into buildModuleModel

packages/host/app/components/card-prerender.gts
lukemelia: It is now in one place: `buildModuleModel`

packages/runtime-common/query-field-utils.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+    if (typeof current === 'object' && segment in current) {
+      current = (current as any)[segment];
+      continue;
+    }
+    if (
+      typeof current === 'object' &&
+      current !== null &&
```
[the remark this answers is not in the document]
lukemelia: Updated

packages/base/query-field-support.ts
```diff
@@ -0,0 +1,337 @@
...
+    () => args()?.query,
+    () => {
+      let realm = args()?.realmHref;
+      return realm ? [realm] : undefined;
+    },
+    {
+      isLive: true,
+      seed: seedRecords
+        ? {
+            cards: seedRecords ?? undefined,
```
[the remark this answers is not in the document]
lukemelia: Updated

habdelra (approved):

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,16 +1221,48 @@ module(basename(__filename), function () {
               return this.author?.firstName + '-poo';
             }
           })
-          static embedded = class Embedded extends Component<typeof this> {
-            <template><@fields.firstName/> (<@fields.nickName/>)</template>
-          }
-          static fitted = class Fitted extends Component<typeof this> {
-            <template><@fields.firstName/> (<@fields.nickName/>)</template>
-          }
```
habdelra, line 1229: why did we do this? this breaks the resulting search in this test because these items won't be in the search doc since they are not longer in these templates

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,16 +1221,48 @@ module(basename(__filename), function () {
...
+          JSON.stringify(postIndexEntry, null, 2),
+        );
+        let postModule = await realm.realmIndexQueryEngine.module(
+          new URL(`${testRealm}post`),
+        );
+        assert.strictEqual(
+          postModule?.type,
+          'module',
+          'post module is in the index after recreation',
+        );
```
habdelra, line 1265: I'm very confused why we added these assertions here. the whole idea is that we want to see the realm.write invalidate everything correclty
but this added code intentionally invalidates this one particular file and we are not able to tell if realm.write invalidated things correctly
what was th emotivation behind this change?
theres also a lot of very low level assertions made in this test now tha tI don't agree with. it seems like not a good idea to be directly qyuerying the DB to see if this module exists
rather we should use the API 's that live at this abstration layer
moreover this test is about the ability to delete an index entry and see that readding it repairs the instances that depend on it. now i'm not really sure what tthis test is trying to prove

PR #3512, opened 2025-10-31
Eco 418 number field configuration

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-418/number-field-configuration


burieberry (commented):
We'll need basic test coverage for all these

packages/boxel-ui/addon/src/components/input-group/index.gts
```diff
@@ -120,6 +124,10 @@ export default class InputGroup extends Component<Signature> {
               @onBlur={{@onBlur}}
               autocomplete={{@autocomplete}}
               inputmode={{@inputmode}}
+              type={{@type}}
+              min={{@min}}
+              max={{@max}}
+              step={{@step}}
```
burieberry, line 130: This is fine, but all input properties can already be passed into this component via the `...attributes`

burieberry: Ignore this comment, realized InputGroup does not pass the `...attributes` to the input element but to the outside container only.

packages/catalog-realm/fields/number.gts
```diff
@@ -0,0 +1,2581 @@
...
+      return readConfig(this.args.configuration, {
+        min: 0,
+        max: 1000,
+        step: 10,
+        allowNegative: false,
+        decimals: 0,
+        currency: '$',
+        showInputs: true,
+      });
+    }
```
burieberry, line 2025: I found this strange that it defaults to currency. I recommend naming it `unit` or `label` instead of `currency`, so that it's more multi-purpose and default should be `''`. Or maybe even use `prefix` and `suffix` they way you did in some other fields, and place it before or after the number based on that.

packages/catalog-realm/fields/number.gts
```diff
@@ -0,0 +1,2581 @@
...
+
+    <template>
+      <span class='dual-range-atom'>{{this.displayValue}}</span>
+      <style scoped>
+        .dual-range-atom {
+          font-weight: var(--boxel-font-weight-semibold);
+          color: var(--primary, var(--boxel-dark));
+        }
+      </style>
```
burieberry, line 2066: let's remove the styling here and let the user style it how ever they want. This looks inconsistent with the rest of the embedded views.

packages/catalog-realm/fields/number.gts
```diff
@@ -0,0 +1,2581 @@
...
+          font-weight: var(--boxel-font-weight-semibold);
+          text-transform: uppercase;
+          letter-spacing: var(--boxel-lsp);
+          color: var(--muted-foreground, var(--boxel-450));
+        }
+        .dual-range__value-amount {
+          font-size: var(--boxel-font-size-lg);
+          font-weight: var(--boxel-font-weight-semibold);
+          color: var(--primary, var(--boxel-dark));
+        }
```
burieberry, line 1893: Let's remove these.

packages/catalog-realm/fields/number.gts
```diff
@@ -0,0 +1,2581 @@
...
+          align-items: center;
+        }
+        .dual-range__track {
+          position: absolute;
+          width: 100%;
+          height: var(--boxel-sp-xxs);
+          border-radius: var(--boxel-border-radius-xl);
+          background: linear-gradient(
+            to right,
+            var(--track-bg, var(--boxel-200)) var(--range-start, 0%),
```
burieberry, line 1907: use `--border` for `--track-bg`, and `--primary` for `--active-bg` for a better theming experience.

burieberry:
I started reviewing these. Here are a couple of comments so far:
- Would be nice to have each of these fields in a separate file for legibility.
- Would be good to have each be a default export for consistency with other fields and so people can name their imports as they wish.
- The specs for these don't seem to exist when adding them to a card via the code mode UI
- Something seems to be broken even after fixing the merge conflict. I couldn't use the preview card you made. The edit mode crashes.
- Need consistent empty states. Below is what it currently looks like. It's fine for rating field and progress circle. Progress bar can also display 0%. For the rest either leave it blank or use the –. 

empty-state

packages/catalog-realm/fields/number.gts
```diff
@@ -0,0 +1,2581 @@
...
+// PIN / OTP Input ----------------------------------------------------------
+// Simple PIN input: type digit → go to next, backspace → go to previous
+
+type PinConfig = {
+  length: number;
+  label?: string;
+  description?: string;
+};
+
+class PinNumberFieldEdit extends Component<typeof PinNumberField> {
```
burieberry, line 755: Should this be named `PinField` since 'number' is duplicative?

lucaslyl: @burieberry Nice catch! Really appreciate your suggestions in this PR.

This PR version is just a demo to gather feedback. Our team had a demo session with Chris two days ago, and we’ve been planning to come up with a new strategy for how to display the number fields using a presentation-type method (with a variant to choose). I’ll also exclude the pin number fields from the number type — those should be string fields instead (along with some other advanced fields). The same goes for dual range and currency fields — since those are advanced fields, they’ll be handled separately instead of being grouped with number fields.

I’ll make sure to address your other comments when I create the new PR with the updated strategy, and I’ll ping you for another review then.

PR #3526, opened 2025-11-03
Use headless chrome for testing realm server

habdelra:
Moving our testing realm server to use the new headless chrome indexing. This should add stability for starting the realm server for tests. This realm server is read only so it should be pretty low stakes.


habdelra:
this actulaly needs some bugfixes in https://github.com/cardstack/boxel/actions/runs/19048818924?pr=3517. closing

PR #3529, opened 2025-11-04
Mechanism for alternate home page for published realm

FadhlanR:
This PR introduces a SiteConfig card definition with a home relationship so each realm can declare its host-mode entry point. It also adds a shared home-page-resolver service that host mode and host submode both use to load either the configured home card (from site.json) or the legacy index card, keeping their behavior consistent.

https://github.com/user-attachments/assets/57a8611e-d52f-40cc-80c9-37c784076f53


FadhlanR:
I’ll do further investigation to check whether the flaky tests are related to this PR or not. Overall, this PR is ready for review.

backspace:
I haven’t managed to exercise this locally, my `site.json` is erroring:

site json in minicatalog 2025-11-06 16-36-44

I’ll try again tomorrow.

I have some confusion about it being `host` that’s deciding what to render for a request to `/`. If `site.json` is present, should the realm itself be resolving this and serving the right card?

FadhlanR:
> I haven’t managed to exercise this locally, my `site.json` is erroring:
> 
> site json in minicatalog 2025-11-06 16-36-44
> I’ll try again tomorrow.
> 
> I have some confusion about it being `host` that’s deciding what to render for a request to `/`. If `site.json` is present, should the realm itself be resolving this and serving the right card?

I think this logic belongs to how we render the index HTML in specific contexts, host mode and host submode, while the realm itself doesn’t know (and shouldn’t need to know) about those modes. So we let the realm keep serving whatever card is requested, and apply the host-mode override inside the host app. But I open for any other approach.

backspace:
I still can’t get this working locally, I’m sure it’s a problem with my environment and not your code though 😖

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -132,6 +144,24 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
...
+        this.operatorModeStateService.hostModePrimaryCard ?? undefined;
+      let normalizedRealm = realmURL.endsWith('/') ? realmURL : `${realmURL}/`;
+      let isRealmIndex = homePageCardId === `${normalizedRealm}index`;
+      let shouldUseHome =
+        !currentPrimary ||
+        (!isRealmIndex && currentPrimary === `${normalizedRealm}index`);
+      if (shouldUseHome && currentPrimary !== homePageCardId) {
+        this.operatorModeStateService.setHostModePrimaryCard(homePageCardId);
+      }
+    }
```
lukemelia, line 171: This is confusing and needs at least a comment explaining what it is doing

packages/host/app/components/operator-mode/host-submode.gts
```diff
@@ -239,14 +269,21 @@ export default class HostSubmode extends Component<HostSubmodeSignature> {
...
       <:default as |layout|>
-        <HostModeContent
-          @primaryCardId={{this.operatorModeStateService.hostModePrimaryCard}}
-          @stackItemCardIds={{this.operatorModeStateService.hostModeStack}}
-          @removeCardFromStack={{this.removeCardFromStack}}
-          @openInteractSubmode={{fn layout.updateSubmode 'interact'}}
-          @viewCard={{this.viewCard}}
-          class='host-submode-content'
-        />
+        {{#if this.ensureHomePageCardTask.isRunning}}
```
lukemelia, line 281: This is find for Host Mode Preview, but ideally we will avoid spinners before the homepage renders when viewing actual host mode. OK to save that optimization for a separate PR.

packages/host/app/routes/index.gts
```diff
@@ -51,11 +54,79 @@ export default class Card extends Route {
   @service private declare operatorModeStateService: OperatorModeStateService;
   @service private declare router: RouterService;
   @service private declare store: StoreService;
+  @service private declare homePageResolver: HomePageResolver;
+  @service declare fastboot: {
```
lukemelia: We're working on eliminating fastboot, so this new usage should be checked by @habdelra

packages/host/app/services/home-page-resolver.ts
```diff
@@ -0,0 +1,166 @@
...
+
+    let promise = this.inflight.get(realmURL);
+    if (!promise) {
+      promise = this.loadHomePage(realmURL).finally(() => {
+        this.inflight.delete(realmURL);
+      });
+      this.inflight.set(realmURL, promise);
+    }
+
+    let cardId = await promise;
```
lukemelia:
```suggestion
    let loadingHomePagePromise = this.inflight.get(realmURL);
    if (!loadingHomePagePromise) {
      loadingHomePagePromise = this.loadHomePage(realmURL).finally(() => {
        this.inflight.delete(realmURL);
      });
      this.inflight.set(realmURL, loadingHomePagePromise);
    }

    let cardId = await loadingHomePagePromise;
```

lukemelia (approved):

packages/host/app/routes/index.gts
habdelra: confirm--we should not be doing this in headless chrome.

packages/host/app/routes/index.gts
```diff
@@ -51,11 +54,79 @@ export default class Card extends Route {
...
+      transition.abort();
+      this.router.replaceWith('index', targetPath, {
+        queryParams: transition.to?.queryParams ?? {},
+      });
+    }
+  }
+
+  private currentURL(transition: Transition): string | undefined {
+    if (this.fastboot.isFastBoot) {
+      let { protocol, host, path } = this.fastboot.request ?? {};
```
habdelra: this will not work once we remove the feature flag for headless chrome. help me to understand why we use fastboot here. is there an indexing concern?

packages/host/app/routes/index.gts
```diff
@@ -51,11 +54,79 @@ export default class Card extends Route {
...
+      let normalizedProtocol = protocol ?? 'http:';
+      if (!normalizedProtocol.endsWith(':')) {
+        normalizedProtocol = `${normalizedProtocol}:`;
+      }
+      return `${normalizedProtocol}//${host}${path}`;
+    }
+
+    return `${this.hostModeService.hostModeOrigin}/${
+      transition.to?.params?.path ?? ''
+    }`;
```
habdelra, line 107: during indexing we use puppeteer to render your app in the same way that a user would experience your app. so this non-fastboot branch in your code will the what we use during indexing when using headless chrome indexing.

habdelra (commented):
I don't understand why we are using fastboot as that is something that is only seen by indexing. please explain more. Also, a lot of our published realm config is actaully held in the ember environment config that is established in packages/realm-server/server.ts#retrieveIndexHTML(). Have you explored using this for the published realm configuration?

packages/host/app/routes/index.gts
FadhlanR: Yes, there was an indexing error when I wasn’t retrieving url from FastBoot. However, you're right — I don’t need it anymore. I’ve removed it, and everything works well now. All code related fasboot has been removed.

PR #3535, opened 2025-11-04
realm: Add private dependency check endpoint

backspace:
This adds `${realmURL}_has-private-dependencies`, which returns a report on whether a realm has private external dependencies that will result in host mode rendering problems, which are detected by starting with the realm’s instances and traversing the dependency graph to its leaf nodes.

The easiest way to review this is to read the tests.

The next step will be consuming this in the host submode UI to warn when publishing a realm, allowing users to manually fix private dependency problems. The longer-term goal is to bundle dependencies automatically, and/or allow users to select and bundle what they want published.


packages/runtime-common/publishability.ts
```diff
@@ -0,0 +1,214 @@
...
+        let dependencyRealmURL = ensureTrailingSlash(dependencyEntry.realmUrl);
+        if (dependency !== canonicalDependencyUrl) {
+          dependency = canonicalDependencyUrl;
+        }
+
+        if (dependencyRealmURL === normalizedRealmURL) {
+          if (!realmResources.has(dependency)) {
+            continue;
+          }
+          let subchains = await collectChains(dependency, updatedAncestry);
```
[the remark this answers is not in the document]
backspace: thanks, fixed

packages/realm-server/tests/realm-endpoints/publishability-test.ts
```diff
@@ -0,0 +1,707 @@
...
+    );
+    assert.true(
+      violation.externalDependencies.every(
+        (dep: any) => dep.realmURL === privateRealmURL,
+      ),
+      'dependencies report the private realm URL',
+    );
+  });
+
+  test('traces transitive dependencies through public realms', async function (assert) {
```
lukemelia: I'm not convinced we need to do this. Stopping at a dep in a public realm seems like a pretty darn good optimization.

packages/runtime-common/realm.ts
```diff
@@ -440,6 +446,11 @@ export class Realm {
         SupportedMimeType.CardTypeSummary,
         this.fetchCardTypeSummary.bind(this),
       )
+      .get(
+        '/_has-private-dependencies',
```
lukemelia: Might be more clear as `_private-dependencies-check` or  `_dependencies-check` or  `_publishability-check`

packages/runtime-common/realm.ts
```diff
@@ -2762,6 +2773,139 @@ export class Realm {
...
+        } else {
+          visibility = 'private';
+        }
+      }
+
+      visibilityCache.set(realmUrl, visibility);
+      return visibility;
+    };
+
+    while (queue.length > 0) {
```
lukemelia: Does this loop assume that all realms are queryable in this database? If so, we will have a problem with federation of realm servers.

lukemelia (approved):
Please make sure @habdelra gets to review this.

jurgenwerk (approved):

packages/runtime-common/realm.ts
backspace: hmm indeed… is there a way to check for dependencies just using API endpoints? I’ll think more about this

backspace: I have some draft code that addresses this, it involves a new `_resource-index` endpoint to cross server boundaries, but my question is: is this getting too involved for a temporary solution, would it be better to instead work on the bundling of external dependencies vs just warning about them?

lukemelia: Hard to say. Will we need to know whether there are private deps in order to know whether to bundle the private deps?

backspace: ya it does seem like this work would be applicable to the bundling step too, so I went ahead with it 🤓

packages/runtime-common/realm.ts
backspace: ya I’ve cycled through a few names but chose `GET _publishability` finally, thanks

packages/realm-server/tests/realm-endpoints/publishability-test.ts
backspace: ya… I had some convoluted scenario in mind but it really doesn’t make sense, at least with the current setup, so removed now, thanks

backspace:
Adding handling for cross-server realm dependencies has introduced problems because it’s trying to check _everything_, not just dependencies that come from realms. I naïvely added exclusions (icons, commands), but it’s clear to me now a different approach is needed:

```
failed to fetch remote resource index for https://packages/@glimmer/component via https://packages/@glimmer/_resource-index?url=https%3A%2F%2Fpackages%2F%40glimmer%2Fcomponent: fetch failed
failed to fetch remote resource index for https://packages/@glimmer/component via https://packages/_resource-index?url=https%3A%2F%2Fpackages%2F%40glimmer%2Fcomponent: fetch failed
```

I’m not sure where this `https://packages` pattern comes from but if it can reliably differentiate realm-originating dependencies vs npm-esque ones, that’s what I’ll do.

packages/runtime-common/scoped-css.ts
```diff
@@ -2,12 +2,14 @@ import { decodeScopedCSSRequest, isScopedCSSRequest } from 'glimmer-scoped-css';
 import jsEscapeString from 'js-string-escape';
 
 export async function maybeHandleScopedCSSRequest(req: Request) {
-  if (isScopedCSSRequest(req.url)) {
+  let { pathname } = new URL(req.url);
+
+  if (isScopedCSSRequest(pathname)) {
```
backspace, line 7: This is needed because `req.url` includes query parameters, so requests that actually were for scoped CSS were being missed.

backspace:
> Please make sure @habdelra gets to review this.

Can you look at this, @habdelra?

packages/realm-server/tests/realm-endpoints/publishability-test.ts
```diff
@@ -0,0 +1,735 @@
...
+    afterEach: async () => {
+      await closeServer(testRealmHttpServer);
+      if (pathExistsSync(tempDir.name)) {
+        removeSync(tempDir.name);
+      }
+    },
+  });
+
+  test('reports publishable realm when there are no private dependencies', async function (assert) {
+    let { url, realm } = await createRealm({
```
habdelra: in these tests, it might be clearer using the test helper `setupPermissionedRealms()` which more clearly describes the permissions for each realm. I was a little confused at first about the permissions for createRealm(), before I realized it was specifically creating private realms.

packages/realm-server/tests/realm-endpoints/publishability-test.ts
```diff
@@ -0,0 +1,735 @@
...
+    },
+    afterEach: async () => {
+      await closeServer(testRealmHttpServer);
+      if (pathExistsSync(tempDir.name)) {
+        removeSync(tempDir.name);
+      }
+    },
+  });
+
+  test('reports publishable realm when there are no private dependencies', async function (assert) {
```
habdelra: what about a test for trying to use an external private link (not a module)?

packages/realm-server/tests/realm-endpoints/dependencies-test.ts
```diff
@@ -0,0 +1,101 @@
...
+          @field title = contains(StringField);
+        }
+      `,
+    );
+    await testRealm.realmIndexUpdater.fullIndex();
+
+    let targetUrl = `${testRealm.url}resource-index-card.gts`;
+    let response = await request
+      .get(`/_resource-index?url=${encodeURIComponent(targetUrl)}`)
+      .set('Accept', SupportedMimeType.JSON);
```
habdelra: is there a reason why this API doesn't use JSON-API ? I think all our other API's do...

habdelra:
I was confused by what `/_resource_index` did based on it's name--it was pretty ambiguous. I think a better name for this endpoint would be `/_dependencies`, since that is what it is returning.

packages/realm-server/tests/realm-endpoints/dependencies-test.ts
backspace: oversight, thanks for pointing it out, fixed in f551110

packages/realm-server/tests/realm-endpoints/publishability-test.ts
backspace: I’ve updated all the tests to use that helper (or the singular one), thanks

packages/realm-server/tests/realm-endpoints/publishability-test.ts
backspace: can you go into more detail about what this would entail? Do you mean an image or the like?

backspace:
> I was confused by what `/_resource_index` did based on it's name--it was pretty ambiguous. I think a better name for this endpoint would be `/_dependencies`, since that is what it is returning.

I changed it to that, thanks

packages/realm-server/tests/realm-endpoints/publishability-test.ts
backspace: I’m going to merge this for now but I can add this test in a followup PR

PR #3536, opened 2025-11-05
Fix realm ETag handling for Accept-specific module responses

lukemelia:
Prior to this commit, the realm server uses the file’s last-modified timestamp as the ETag for every representation of a module URL. That caused a subtle cache-busting bug: a browser could fetch `Accept: application/vnd.card+source`, cache the raw TypeScript along with its ETag, and then immediately revalidate with `Accept: */*`. Because the downstream request carried the same `If-None-Match` value, the server returned `304 Not Modified` even though the `*/*` variant is supposed to be the transpiled AMD output. The loader would then reuse the browser’s stale, untranspiled copy and fail at runtime.

This change makes ETags variant-aware. We append a “module” or “source” suffix when we build the tag, which keeps the `Accept: */*` and `Accept: card-source` responses in separate validation buckets. The module fallback and the raw-source path both emit consistent cache-control headers, and new tests in the realm endpoints suite lock in the expectation that the two variants expose different ETags and only the matching tag produces a 304.

With ETags split per representation, `Vary: Accept` now works as intended: the module loader always receives compiled JavaScript, we eliminate mysterious “raw TypeScript” regressions during live editing, and browsers can still revalidate efficiently without poisoning the cache.

### DIFF
--- a/packages/runtime-common/realm.ts
+++ b/packages/runtime-common/realm.ts
@@ -148,6 +148,8 @@ export interface FileRef {
 const CACHE_HEADER = 'X-Boxel-Cache';
 const CACHE_HIT_VALUE = 'hit';
 const CACHE_MISS_VALUE = 'miss';
+const MODULE_ETAG_VARIANT = 'module';
+const SOURCE_ETAG_VARIANT = 'source';
 
 type CachedSourceFileEntry = {
   type: 'file';
@@ -186,6 +188,17 @@ type ModuleLoadResult =
       body: string;
       headers: Record<string, string>;
     };
+
+function buildEtag(
+  lastModified: number | undefined,
+  variant?: string,
+): string | undefined {
+  if (lastModified == null) {
+    return undefined;
+  }
+  let base = String(lastModified);
+  return variant ? `${base}:${variant}` : base;
+}
 
 export interface TokenClaims {
   user: string;
@@ -1454,13 +1467,12 @@ export class Realm {
       return { kind: 'shimmed', response };
     }
 
-    let etag =
-      fileRef.lastModified != null ? String(fileRef.lastModified) : undefined;
+    let etag = buildEtag(fileRef.lastModified, MODULE_ETAG_VARIANT);
     if (etag && request.headers.get('if-none-match') === etag) {
       let headers: Record<string, string> = {
         'cache-control': 'public, max-age=0',
-        etag,
       };
+      headers.etag = etag;
       if (fileRef.lastModified != null) {
         headers['last-modified'] = formatRFC7231(fileRef.lastModified * 1000);
       }
@@ -1536,12 +1548,11 @@ export class Realm {
     requestContext: RequestContext,
     options?: {
       defaultHeaders?: Record<string, string>;
+      etagVariant?: string;
     },
   ): Promise<ResponseWithNodeStream> {
-    if (
-      ref.lastModified != null &&
-      request.headers.get('if-none-match') === String(ref.lastModified)
-    ) {
+    let etag = buildEtag(ref.lastModified, options?.etagVariant);
+    if (etag && request.headers.get('if-none-match') === etag) {
       return createResponse({
         body: null,
         init: { status: 304 },
@@ -1555,7 +1566,7 @@ export class Realm {
       ...(Symbol.for('shimmed-module') in ref
         ? { 'X-Boxel-Shimmed-Module': 'true' }
         : {}),
-      etag: String(ref.lastModified),
+      ...(etag ? { etag } : {}),
       'cache-control': 'public, max-age=0', // instructs the browser to check with server before using cache
     };
     if (createdFromDb != null) {
@@ -1750,6 +1761,7 @@ export class Realm {
                 ...cached.defaultHeaders,
                 [CACHE_HEADER]: CACHE_HIT_VALUE,
               },
+              etagVariant: SOURCE_ETAG_VARIANT,
             },
           );
         } finally {
@@ -1803,6 +1815,7 @@ export class Realm {
       if (bypassCache) {
         return await this.serveLocalFile(request, handle, requestContext, {
           defaultHeaders,
+          etagVariant: SOURCE_ETAG_VARIANT,
         });
       } else {
         let cachedRef = await this.materializeFileRef(handle);
@@ -1814,6 +1827,7 @@ export class Realm {
         });
         return await this.serveLocalFile(request, cachedRef, requestContext, {
           defaultHeaders,
+          etagVariant: SOURCE_ETAG_VARIANT,
         });
       }
     } finally {
--- a/packages/realm-server/tests/realm-endpoints-test.ts
+++ b/packages/realm-server/tests/realm-endpoints-test.ts
@@ -267,6 +267,124 @@ module(basename(__filename), function () {
       );
     });
 
+    const transpileTestCardSource = `
+      import {
+        linksToMany,
+        field,
+        Component,
+        FieldDef,
+      } from 'https://cardstack.com/base/card-api';
+      import { Country } from './country';
+
+      export class TranspileTestField extends FieldDef {
+        static displayName = 'Trips';
+        @field countriesVisited = linksToMany(Country);
+
+        static embedded = class Embedded extends Component<typeof this> {
+          <template>
+            <address data-test-trips-card>
+              <@fields.countriesVisited />
+            </address>
+          </template>
+        };
+      }
+    `;
+
+    test('serves transpiled .gts modules when Accept is */*', async function (assert) {
+      let modulePath = 'transpile-test.gts';
+      let authHeader = `Bearer ${createJWT(testRealm, 'user', ['read', 'write'])}`;
+
+      await testRealm.write(modulePath, transpileTestCardSource);
+
+      let response = await request
+        .get(`/${modulePath}`)
+        .set('Accept', SupportedMimeType.All)
+        .set('Authorization', authHeader);
+
+      assert.strictEqual(response.status, 200, 'module request succeeds');
+      assert.strictEqual(
+        response.headers['content-type'],
+        'text/javascript',
+        'transpiled module advertises javascript content type',
+      );
+      assert.ok(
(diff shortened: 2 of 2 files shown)


packages/runtime-common/realm.ts
```diff
@@ -1454,13 +1467,14 @@ export class Realm {
...
 
-    let etag =
-      fileRef.lastModified != null ? String(fileRef.lastModified) : undefined;
+    let etag = buildEtag(fileRef.lastModified, MODULE_ETAG_VARIANT);
     if (etag && request.headers.get('if-none-match') === etag) {
       let headers: Record<string, string> = {
         'cache-control': 'public, max-age=0',
-        etag,
       };
+      if (etag) {
```
IanCal: this if isn't required here, 1471 has  `if(etag &&`  so we could set it in the headers directly

packages/runtime-common/realm.ts
```diff
@@ -187,6 +189,17 @@ type ModuleLoadResult =
...
 
+function buildEtag(
+  lastModified: number | undefined,
+  variant?: string,
+): string | undefined {
+  if (lastModified == null) {
+    return undefined;
+  }
+  let base = String(lastModified);
+  return variant ? `${base}:${variant}` : base;
```
IanCal, line 200: If lastModified is undefined we're going to have problems with the same (strong?) etag regardless of content, perhaps if lastModified is undefined we shouldn't have any etag.

packages/runtime-common/realm.ts
```diff
@@ -148,6 +148,8 @@ export interface FileRef {
 const CACHE_HEADER = 'X-Boxel-Cache';
 const CACHE_HIT_VALUE = 'hit';
 const CACHE_MISS_VALUE = 'miss';
+const MODULE_ETAG_VARIANT = 'module';
+const SOURCE_ETAG_VARIANT = 'source';
```
IanCal, line 152: Could we just use the accept headers rather than explicit module vs source?

lukemelia: Too verbose and ugly IMO

packages/runtime-common/realm.ts
lukemelia: True. Are there any circumstances where last modified could be undefined?

IanCal:
I tried this locally and can click around and not get transpiled code in codemode, but I'm not sure I follow quite what we're doing with headers overall and the requests.

I think our accept headers and responses are wrong. We request a module with accept headers `*/*` but we don't actually want to accept anything - we want text/javascript, accept `*/*` surely means it would be OK for the server to respond with the source code. We request the card source with "application/vnd.card+source" but then we return "text/plain; charset=utf-8", which isn't what the browser said it wanted. I'm concerned that a valid thing for a cache to do would be to return the source instead of the transpiled module.

Loading code mode with caching enabled I see multiple requests for my card.

request gts with source, with an if-none-match, 304 not modified is returned - misses the etag on the response

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match

> When the condition fails for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) methods, the server must return a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304) and any of the following header fields that would have been sent in a 200 response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.

Request for the module (.gts in the url), does not include an if-none-match header in the request. Response has an etag.

Then we request the source code again without the .gts extension, and no if-none-match header. This is a x-boxel-cache miss, despite having just reloaded the page. 302 temporary redirect (is it temporary?)

This leads to another get request. The get request is for the .gts and card source (accept header), however the request comes with if-none-match "1756221459:module" which is not correct and the response then has the etag "1756221459:source". This shows either our loader or the browser thinks that the module etag could be a valid thing to match against. Given that we had the same URL respond with for a request for `*/*` that seems fair.

I can't show an error with this branch but I'm concerned we're doing the wrong thing with headers and etags and could be setting up a more subtle caching bug.

lukemelia:
I don't have time to respond to this because I just got tot he gym, but I agree with concern about multiple requests. Should be investigated but preexists this PR. @habdelra can explain the header logic. I'm confident it makes sense. This needs to land before sprint planning or staging will be un-demoable.

packages/runtime-common/realm.ts
habdelra: It never undefined. We actually throw during indexing if that is the case

packages/runtime-common/realm.ts
```diff
@@ -1454,13 +1467,12 @@ export class Realm {
...
 
-    let etag =
-      fileRef.lastModified != null ? String(fileRef.lastModified) : undefined;
+    let etag = buildEtag(fileRef.lastModified, MODULE_ETAG_VARIANT);
     if (etag && request.headers.get('if-none-match') === etag) {
       let headers: Record<string, string> = {
         'cache-control': 'public, max-age=0',
-        etag,
       };
+      headers.etag = etag;
```
habdelra, line 1475: Seems superfluous

habdelra:
> I tried this locally and can click around and not get transpiled code in codemode, but I'm not sure I follow quite what we're doing with headers overall and the requests.
> 
> I think our accept headers and responses are wrong. We request a module with accept headers `*/*` but we don't actually want to accept anything - we want text/javascript, accept `*/*` surely means it would be OK for the server to respond with the source code. We request the card source with "application/vnd.card+source" but then we return "text/plain; charset=utf-8", which isn't what the browser said it wanted. I'm concerned that a valid thing for a cache to do would be to return the source instead of the transpiled module.
> 
> Loading code mode with caching enabled I see multiple requests for my card.
> 
> request gts with source, with an if-none-match, 304 not modified is returned - misses the etag on the response
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match
> 
> > When the condition fails for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) methods, the server must return a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304) and any of the following header fields that would have been sent in a 200 response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.
> 
> Request for the module (.gts in the url), does not include an if-none-match header in the request. Response has an etag.
> 
> Then we request the source code again without the .gts extension, and no if-none-match header. This is a x-boxel-cache miss, despite having just reloaded the page. 302 temporary redirect (is it temporary?)
> 
> This leads to another get request. The get request is for the .gts and card source (accept header), however the request comes with if-none-match "1756221459:module" which is not correct and the response then has the etag "1756221459:source". This shows either our loader or the browser thinks that the module etag could be a valid thing to match against. Given that we had the same URL respond with for a request for `*/*` that seems fair.
> 
> I can't show an error with this branch but I'm concerned we're doing the wrong thing with headers and etags and could be setting up a more subtle caching bug.

I didn’t have time to read it all. Will later, but one super important thing to keep in mind. When you use `import foo from ‘./bar’`, to evaluate a module, JavaScript itself (via native browser implementation)will issue the fetch, and we have no control over the accept header (or any header) it will use. And that accept header is always `*/*`.

habdelra (approved):

habdelra:
> I tried this locally and can click around and not get transpiled code in codemode, but I'm not sure I follow quite what we're doing with headers overall and the requests.
> 
> I think our accept headers and responses are wrong. We request a module with accept headers `*/*` but we don't actually want to accept anything - we want text/javascript, accept `*/*` surely means it would be OK for the server to respond with the source code. We request the card source with "application/vnd.card+source" but then we return "text/plain; charset=utf-8", which isn't what the browser said it wanted. I'm concerned that a valid thing for a cache to do would be to return the source instead of the transpiled module.
> 
> Loading code mode with caching enabled I see multiple requests for my card.
> 
> request gts with source, with an if-none-match, 304 not modified is returned - misses the etag on the response
> 
> https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match
> 
> > When the condition fails for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) methods, the server must return a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304) and any of the following header fields that would have been sent in a 200 response to the same request: Cache-Control, Content-Location, Date, ETag, Expires, and Vary.
> 
> Request for the module (.gts in the url), does not include an if-none-match header in the request. Response has an etag.
> 
> Then we request the source code again without the .gts extension, and no if-none-match header. This is a x-boxel-cache miss, despite having just reloaded the page. 302 temporary redirect (is it temporary?)
> 
> This leads to another get request. The get request is for the .gts and card source (accept header), however the request comes with if-none-match "1756221459:module" which is not correct and the response then has the etag "1756221459:source". This shows either our loader or the browser thinks that the module etag could be a valid thing to match against. Given that we had the same URL respond with for a request for `*/*` that seems fair.
> 
> I can't show an error with this branch but I'm concerned we're doing the wrong thing with headers and etags and could be setting up a more subtle caching bug.

So I want to be sure you are not conflating source code with transpiled code. If there is a request with an accept for `*/*` we never should respond with source code, we respond with transpiled code. Due to the way we emulate native loading. Native loading always uses `*/*` for loads triggered via `import` statements. Not sure why, that’s just how browsers have implemented it. So that actually has special meaning to us since we made our loader compliant with the native loader: when requesting executable file with `*/*` we will treat it as a module request: always. Since imports can use node resolution (without an extension), this is why we redirect to the canonical path. It’s useful four our app, but also, that’s definitely a module aka transpiled code request. At the same time Monaco is likely updating its display of source code which is why you’ll also see requests for card+source too.

PR #3537, opened 2025-11-05
Keep track of card mutation requests coming from the AI assistant (by room)

jurgenwerk:
This is the first step towards having the self-healing card/code patching feature where we want the assistant to be able to detect card errors as a result of its own code/card patching.

This PR introduces tracking of the requests that the AI assistant made - there is now a new `aiAssistantClientRequestIdsByRoom` property in the commands service.

The idea is that when the card store detects a card error after receiving the invalidation, it will receive the client request id in the matrix realm event, and then if it can find the client request id in `aiAssistantClientRequestIdsByRoom`, it will mean *the assistant* is at fault for introducing an indexing error–in that we want to give it feedback and the error document so that it will try to correct itself (future feature)

### DIFF
--- a/packages/base/command.gts
+++ b/packages/base/command.gts
@@ -62,6 +62,7 @@ export class CopySourceResult extends CardDef {
 export class PatchCardInput extends CardDef {
   @field cardId = contains(StringField);
   @field patch = contains(JsonField);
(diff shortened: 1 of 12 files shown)


jurgenwerk:
@codex review

habdelra (approved):

PR #3542, opened 2025-11-06
linksToMany performance optimizations

ef4:
There's still more to do here, but these changes take one of our problematic examples from 62s to 25s.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -1390,8 +1390,11 @@ class LinksToMany<FieldT extends CardDefConstructor>
       if (!(globalThis as any).__lazilyLoadLinks) {
         throw new NotLoaded(instance, notLoadedRefs, this.name);
       }
-      for (let reference of notLoadedRefs) {
-        lazilyLoadLink(instance, this, reference, { value });
+      for (let entry of value) {
+        if (isNotLoadedValue(entry) && !(entry as any).loading) {
+          lazilyLoadLink(instance, this, entry.reference, { value });
+          (entry as any).loading = true;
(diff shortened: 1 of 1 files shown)


packages/base/card-api.gts
```diff
@@ -3415,7 +3418,7 @@ export class Box<T> {
       }
     });
     this.prevChildren = newChildren;
-    this.prevValues = newChildren.map((child) => child.value);
+    this.prevValues = value.slice();
```
ef4, line 3421: This one helps because `child.value` goes back through `this.value`, which ultimately re-runs the getter for our field. We can skip an extra N executions of the getter since we already ran it and have the value in local scope.

packages/base/card-api.gts
```diff
@@ -1390,8 +1390,11 @@ class LinksToMany<FieldT extends CardDefConstructor>
       if (!(globalThis as any).__lazilyLoadLinks) {
         throw new NotLoaded(instance, notLoadedRefs, this.name);
       }
-      for (let reference of notLoadedRefs) {
-        lazilyLoadLink(instance, this, reference, { value });
+      for (let entry of value) {
+        if (isNotLoadedValue(entry) && !(entry as any).loading) {
+          lazilyLoadLink(instance, this, entry.reference, { value });
+          (entry as any).loading = true;
```
ef4, line 1396: This is sticking a flag on the not-loaded markers so that we can short-circuit working on them repeatedly, earlier in the process.

There was already correct caching further down the call stack, but just finding the right spot in the cache was expensive enough to make this matter.

PR #3552, opened 2025-11-10
Make patch commands more optimistic

FadhlanR:
The scope of this PR is to make the patch commands faster by avoiding the wait for the HTTP `PATCH` request to complete. The idea is to update the local data first before calling the patch endpoint.

For patching instances and fields, we can handle this through the store. I’ve added a new option called `doNotWaitForPersist` the same option we already have in `store.add`, which allows us to modify the card instance and trigger the patch endpoint without waiting for the request to succeed.

The patch code is slightly different. Previously, we directly called the `cardService.saveCard` function to send the patch request, without modifying any local data first. To make it consistent with the other commands, I’ve added logic to check the currently open file. If the open file is the same as the one being patched, we call the patch endpoint through the file resource instead. (I’m open to other approaches if you have suggestions.)

### DIFF
--- a/packages/host/app/commands/patch-card-instance.ts
+++ b/packages/host/app/commands/patch-card-instance.ts
@@ -52,10 +52,14 @@ export default class PatchCardInstanceCommand extends HostBaseCommand<
         "Patch command can't run because it doesn't have all the fields in arguments returned by open ai",
       );
     }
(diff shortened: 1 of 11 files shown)


lukemelia (approved):

jurgenwerk:
@codex review

PR #3559, opened 2025-11-11
Persist attached files and cards

FadhlanR:
This PR fixes an issue where attached cards and files would suddenly disappear. In this update, I’ve enhanced the local-storage-service to also store attached cards and files along with draft message. This not only prevents attachments from disappearing unexpectedly but also ensures they persist after a browser reload or when a message fails to send.

https://github.com/user-attachments/assets/1552f229-111d-484d-9aa1-f001d483556c

### DIFF
--- a/packages/host/app/components/matrix/room.gts
+++ b/packages/host/app/components/matrix/room.gts
@@ -460,6 +460,7 @@ export default class Room extends Component<Signature> {
   });
   private removedAttachedCardIds = new TrackedArray<string>();
   private removedAttachedFileUrls: string[] = [];
+  private lastAutoAttachedFileUrl: string | undefined;
   private getConversationScrollability: (() => boolean) | undefined;
   private scrollConversationToBottom: (() => void) | undefined;
   private roomScrollState: WeakMap<
@@ -541,7 +542,15 @@ export default class Room extends Component<Signature> {
 
     let autoAttachedFileUrl = this.autoAttachedFileUrl;
     let manuallyAttachedFiles = this.filesToAttach;
-    let removedFileUrls = this.removedAttachedFileUrls;
+
+    let removedFileUrls: string[];
+    if (autoAttachedFileUrl !== this.lastAutoAttachedFileUrl) {
+      this.removedAttachedFileUrls.splice(0);
+      removedFileUrls = this.removedAttachedFileUrls;
+      this.lastAutoAttachedFileUrl = autoAttachedFileUrl;
+    } else {
+      removedFileUrls = this.removedAttachedFileUrls;
+    }
 
     let isManuallyAttached = manuallyAttachedFiles.some(
       (file) => file.sourceUrl === autoAttachedFileUrl,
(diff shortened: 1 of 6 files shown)


lukemelia (approved):
It would be good to get @jurgenwerk's review on this

packages/host/app/components/matrix/room.gts
```diff
@@ -538,34 +539,65 @@ export default class Room extends Component<Signature> {
       },
     });
 
-    if (!this.autoAttachedFileUrl) {
+    let autoAttachedFileUrl = this.autoAttachedFileUrl;
+    let manualFiles = this.filesToAttach;
```
jurgenwerk: manuallyAttachedFiles could be a clearer name

packages/host/app/components/matrix/room.gts
```diff
@@ -538,34 +539,65 @@ export default class Room extends Component<Signature> {
...
       });
     }
 
     return state;
   });
 
   private get autoAttachedFileUrl() {
-    return this.operatorModeStateService.state.codePath?.href;
+    let url = this.operatorModeStateService.state.codePath?.href;
+    // use a task to avoid ember/no-side-effects rule violation
```
jurgenwerk: This looks like it shouldn't be in a getter - I think we should respect that violation message. Perhaps autoAttachedFileUrl should be a property that gets set?

packages/host/app/components/matrix/room.gts
```diff
@@ -538,34 +539,65 @@ export default class Room extends Component<Signature> {
...
   private get autoAttachedFile() {
     return this.operatorModeStateService.state.submode === Submodes.Code
       ? this.autoAttachedFileResource.value
       : undefined;
   }
 
   private get removeAutoAttachedFile() {
-    return this.autoAttachedFileResource.remove;
+    return () => {
+      //this.markAttachedFileRemoved(this.autoAttachedFileUrl);
```
jurgenwerk: Probably this can be removed?

packages/host/app/components/matrix/room.gts
```diff
@@ -981,19 +1018,22 @@ export default class Room extends Component<Signature> {
...
-    if (fileIndex != undefined && fileIndex !== -1) {
-      if (this.filesToAttach !== undefined) {
-        this.filesToAttach.splice(fileIndex, 1);
-      }
+    let files = this.filesToAttach;
+    if (!files.length) {
+      return;
+    }
+
+    let next = files.filter((f) => f.sourceUrl !== file.sourceUrl);
```
jurgenwerk: Is there a better variable name than `next` ?

packages/host/app/services/matrix-service.ts
```diff
@@ -1963,6 +2001,35 @@ export default class MatrixService extends Service {
     ]);
   }
 
+  private ensureRoomDraftHydrated(roomId: string) {
```
jurgenwerk: Since we are restoring from localstorage can we simply use that nomenclature? I.e. "ensureMessageDraftRestored"?

To me hydration means more something like when you take raw html and then attach event handlers in the DOM. I think this is the traditional meaning of hydration, so it could lead to confusion when looking at these method names.

jurgenwerk:
@codex review

packages/host/app/components/matrix/room.gts
FadhlanR: Instead of using getter, I used resource. Please check it again.

PR #3561, opened 2025-11-13
Fix: Missing line-breaks for list with emojis

FadhlanR:
The goal is to detect lines beginning with an emoji and prefix them with `*` so that the markdown to html library recognizes them as list items and outputs `<li>` tags.

Before:
Screenshot 2025-11-13 at 15 43 38

After:

Screenshot 2025-11-13 at 15 41 27

Screenshot 2025-11-13 at 15 42 07

### DIFF
--- a/packages/runtime-common/marked-sync.ts
+++ b/packages/runtime-common/marked-sync.ts
@@ -2,6 +2,9 @@ import { marked } from 'marked';
 import { sanitizeHtml } from './dompurify-runtime';
 import { escapeHtml } from './helpers/html';
 
+const DECORATIVE_BULLET_PATTERN =
+  /(^|\n)(\s*)([\p{Extended_Pictographic}★•▪●❖✦✧◉◦◾◽⬢⬡☑✔✅☑️➤➔➜➡→▶])(\s+)/gu;
+
 export function markedSync(
   markdown: string,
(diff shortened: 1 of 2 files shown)


packages/runtime-common/marked-sync.ts
```diff
@@ -2,6 +2,10 @@ import { marked } from 'marked';
 import { sanitizeHtml } from './dompurify-runtime';
 import { escapeHtml } from './helpers/html';
 
+const DECORATIVE_BULLET_PATTERN =
+  // eslint-disable-next-line no-misleading-character-class -- we explicitly want to match pictographic symbols in addition to ASCII glyphs
+  /(^|\n)(\s*)([\p{Extended_Pictographic}★•▪●❖✦✧◉◦◾◽⬢⬡☑✔✅☑️➤➔➜➡→▶])(\s+)/gu;
```
backspace: it looks like `Extended_Pictographic` [includes](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AExtended_Pictographic%3A%5D&g=Emoji_Presentation&i=) some of the symbols explicitly listed like ▶ and ✅, could be pared down to just those not part of it, not a big deal though

backspace (approved):

richardhjtan (approved):

jurgenwerk (approved):

PR #3563, opened 2025-11-13
Improve naming directories for card of default export card defs

FadhlanR:
This PR adds a shared helper that reliably derives the directory name for newly created card instances, so default-exported card defs no longer end up in awkward `/default/<id>.json` folders, for example, a card adopting from `../pet` now saves under `/Pet/<id>.json` and a barrel like `../animals/index.gts` becomes `/Animals/<id>.json`.

### DIFF
--- a/packages/host/tests/acceptance/interact-submode/create-file-test.gts
+++ b/packages/host/tests/acceptance/interact-submode/create-file-test.gts
@@ -1,4 +1,4 @@
-import { click } from '@ember/test-helpers';
+import { click, waitUntil } from '@ember/test-helpers';
 
 import { getService } from '@universal-ember/test-support';
 import { module, test } from 'qunit';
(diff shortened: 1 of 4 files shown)


packages/host/tests/acceptance/interact-submode/create-file-test.gts
```diff
@@ -407,6 +408,39 @@ module('Acceptance | interact submode | create-file tests', function (hooks) {
     assert.dom(`[data-test-stack-card-index]`).exists({ count: 3 });
   });
 
+  test<TestContextWithSave>('creates default-export instances under module directories', async function (assert) {
+    removeRecentCards();
```
backspace, line 412: I see that other tests use this but I’m surprised it would be needed, I’d think that the window mock setup done in `setupApplicationTest` would accomplish this because `localStorage` would start clear with every test.

backspace (approved):
I tried this via the preview deployment, useful improvement!

PR #3567, opened 2025-11-13
Remove unused symbol

ef4:
I think this part of card-api isn't actually used.


ef4:
Oop, nevermind, I found where it's used...

PR #3570, opened 2025-11-13
host: Add publishability warning

backspace:
This uses the `/_publishability` endpoint to show a warning when publishing a realm that has external private dependencies:

Boxel 2025-11-17 15-46-41


FadhlanR (approved):
Looks great to me. A small question: should we disable the publish button until the check is finished?

backspace:
> Looks great to me. A small question: should we disable the publish button until the check is finished?

hmm… the warning doesn’t mean you can’t publish, so maybe it doesn’t matter if you haven’t waited for the warning?

In practice I think people will rarely see the check in progress, because the modal contents render even before the modal has opened. I had to add a deliberate delay to be able to see it.

habdelra (approved):

habdelra:
The fix for the prerender auth issue is in https://github.com/cardstack/boxel/pull/3704

habdelra:
Ok, the auth issue for the prerenderer has been merged.

PR #3573, opened 2025-11-17
add theme creator

tintinthong:


packages/catalog-realm/theme-generation/theme-creator.gts
```diff
@@ -0,0 +1,851 @@
...
+      return 'Generate a theme. This is a brand guide theme.';
+    }
+    if (this.moduleMatches(codeRef, 'style-reference')) {
+      return 'Generate a theme. This is a style reference theme.';
+    }
+    if (this.moduleMatches(codeRef, 'structured-theme')) {
+      return 'Generate a theme. This is a structured theme.';
+    }
+    return null;
+  }
```
richardhjtan: Can consider a getter to get the moduleMatches value and do those as switch statement, it can be cleaner

eg
```
get themeType(): 'brand-guide' | 'style-reference' | 'structured-theme' | null {
  let ref = this.codeRefSelection;
  if (this.moduleMatches(ref, 'matchvalue')) {
    return 'brand-guide';
  }
  if (this.moduleMatches(ref, 'matchvalue')) {
    return 'style-reference';
  }
  if (this.moduleMatches(ref, 'matchvalue')) {
    return 'structured-theme';
  }
  return null;
}

promptGuidanceFor(): string {
  switch (this.themeType) {
 case 'style-reference':
      return this.styleReferenceGuidance;
    case 'brand-guide':
      return this.brandGuideGuidance;
    case 'structured-theme':
      return this.structuredThemeGuidancePrompt;
    default:
      console.error('No prompt guidance for theme type:', this.themeType);
      return '';
  }
}

```

packages/catalog-realm/theme-generation/theme-creator.gts
```diff
@@ -0,0 +1,851 @@
...
+        let realmPath = realmUrl.pathname.replace(/\/+$/, '');
+        if (
+          realmPath &&
+          path.startsWith(realmPath) &&
+          (path.length === realmPath.length ||
+            path.charAt(realmPath.length) === '/')
+        ) {
+          path = path.slice(realmPath.length);
+        }
+      }
```
richardhjtan, line 295: Do we have relative path helper from realm? Or we can split it as another function to explain the function

packages/catalog-realm/components/notification-bubble.gts
```diff
@@ -2,7 +2,7 @@ import GlimmerComponent from '@glimmer/component';
 
 export interface NotificationBubbleSignature {
```
burieberry, line 3: Shouldn't the component styles here be using the theme variables too? Does the boxel-ui Alert component not work for this purpose?

packages/catalog-realm/fields/theme-code-ref.gts
```diff
@@ -0,0 +1,67 @@
+import { Component } from 'https://cardstack.com/base/card-api';
+import CodeRefField from 'https://cardstack.com/base/code-ref';
+import type { ResolvedCodeRef } from '@cardstack/runtime-common';
+import { BoxelSelect } from '@cardstack/boxel-ui/components';
+
+const THEME_CODE_REF_OPTIONS: Array<{ label: string; ref: ResolvedCodeRef }> = [
```
burieberry, line 5: How about the regular Theme Card?

packages/catalog-realm/fields/theme-code-ref.gts
```diff
@@ -0,0 +1,67 @@
...
+      name: 'default',
+    },
+  },
+  {
+    label: 'Brand Guide',
+    ref: {
+      module: 'https://cardstack.com/base/brand-guide',
+      name: 'default',
+    },
+  },
```
burieberry, line 26: maybe the BrandGuide generator should be a separate tool?

packages/catalog-realm/theme-generation/Skill/theme-design.json
```diff
@@ -0,0 +1,30 @@
...
...structions": "## Structured Theme\n- Base theme card defining cohesive tokens (palette, typography, spacing, radius, shadows, charts, sidebar) across light `:root` and dark `.dark` modes.\n- Includes plain maps for `rootVariables` and `darkModeVariables`, plus compiled `cssVariables` (with `@theme inline`) so any card can consume the theme.\n- Use OKLCH or Hex values with AA contrast and reference tokens with `var(--token, fallback)`.\n\n## Style Reference (extends Structured Theme)\n- Adds narrative context for the look and feel: `styleName`, `visualDNA`, inspirations, wallpaper imagery, and card metadata while still carrying all structured theme tokens and CSS.\n- Use it when you want both the full token set and a short story for the mood; it stays compatible with any card that reads structured theme variables.\n\n## Brand Guide (extends Style Reference)\n- Adds brand assets and semantics on top of the style reference: marks/clearspace, heading/body typography stacks, named swatches and functional palette, plus all structured theme tokens and CSS.\n- Use it to ship a complete brand kit that can theme any card while also documenting marks, fonts, and semantic colors for your org."
```
burieberry, line 20: our cards don't actually use the `@theme inline`. I think it's for tailwind

packages/catalog-realm/theme-generation/ThemeCreator/f29cce63-ba80-4afd-8a32-8f4e43d62a65.json
```diff
@@ -0,0 +1,38 @@
...
+    "meta": {
+      "adoptsFrom": {
+        "name": "ThemeCreator",
+        "module": "../theme-creator"
+      }
+    },
+    "type": "card",
+    "attributes": {
+      "realm": "http://localhost:4201/catalog/",
+      "prompt": "make something yellow",
```
burieberry: let's use a better example here

packages/catalog-realm/theme-generation/theme-templates.gts
```diff
@@ -0,0 +1,381 @@
...
+  "data": {
+    "type": "card",
+    "attributes": {
+      "cardInfo": {
+        "title": "<title describing the theme>",
+        "description": "<1-2 sentence summary>",
+        "thumbnailURL": "<https url>"
+      },
+      "rootVariables": {
+        "background": "<hex color>",
```
burieberry: it could be any color value

burieberry (commented):
I ran into the error below when trying to generate via Theme Creator. I created a new Theme Creator card because the catalog one is readonly? 

theme-creator

Other feedback:
- It would be great if we could directly use the catalog version of Theme Creator. (But if not, the buttons and fields should not appear as though they are editable. A message or button about what the user should do to be able to use this card would help too.)

- UX feedback: The dropdowns should have default options selected (for example the LLM model and realm). Ideally the user can type in the text area and press enter to generate.

- The existing theme card previews are confusing. I don't know what these are supposed to be. Maybe we should pick a larger fitted card size.

existing-cards

packages/catalog-realm/theme-generation/theme-creator.gts
tintinthong: Done

packages/catalog-realm/theme-generation/theme-creator.gts
tintinthong: Using realmPath. Thanks for the spot

packages/catalog-realm/theme-generation/ThemeCreator/f29cce63-ba80-4afd-8a32-8f4e43d62a65.json
tintinthong: DONE "Create a theme in deep crimson, with bold contrast, minimal typography, and clean UI surfaces suitable for dashboards."

packages/catalog-realm/components/notification-bubble.gts
tintinthong: Noted. I am using the Alert component now

packages/catalog-realm/fields/theme-code-ref.gts
tintinthong: Added

packages/catalog-realm/theme-generation/theme-templates.gts
tintinthong: ok I just used the word "css color"

tintinthong:
> UX feedback: The dropdowns should have default options selected (for example the LLM model and realm). Ideally the user can type in the text area and press enter to generate.

This seems unsupported api currently. There are workarounds to this but I think they make for a more confusing. We shud have a single theme creator and not rely on user to build a new one from scratch so perhaps the defaults are not that meaningful

tintinthong:
> It would be great if we could directly use the catalog version of Theme Creator. (But if not, the buttons and fields should not appear as though they are editable. A message or button about what the user should do to be able to use this card would help too.)

Super valid. We shud absolutely shift this to a remix and the catlaog as a listing. Otherwise, its just kinda unusable

packages/catalog-realm/components/paginated-cards.gts
```diff
@@ -0,0 +1,178 @@
...
+  <template>
+    {{#if this.canRenderSearch}}
+      {{#let
+        this.paginatedQuery this.searchComponent
+        as |paginatedQuery SearchComponent|
+      }}
+        {{#if paginatedQuery}}
+          <SearchComponent
+            @query={{paginatedQuery}}
+            @format='embedded'
```
burieberry, line 123: I think the reason the Existing Cards appear broken is that the embedded template is broken. It only shows a description for some. Select "fitted" here and it looks much better.

burieberry (approved):

PR #3579, opened 2025-11-17
Improved inheritance handling when looking up field descriptions

lukemelia:


ef4 (approved):

habdelra (approved):
co-pilot left a couple small suggestions, but otherwise looks good

PR #3581, opened 2025-11-18
Add “Set as Site Home” menu in the … stack menu of Site Config card instances

FadhlanR:
In this PR, I added a new menu item to the stack item menu if the card is an instance of a site config. When the menu is clicked, it will copy the content of the instance and create the site.json file under the root directory of the realm. With this new menu, the user has a way to set the alternate home page in host mode. I used the _atomic endpoint since it has the capability to create a card instance with a deterministic URL (location), but I need to tweak it a little bit to allow an update operation.

https://github.com/user-attachments/assets/d0e6bba6-43cc-42d7-bb8a-8b257cb756af

### DIFF
--- a/packages/base/site-config.gts
+++ b/packages/base/site-config.gts
@@ -1,10 +1,39 @@
 import { CardDef, Component, field, linksTo } from './card-api';
+import { getCardMenuItems } from '@cardstack/runtime-common';
+import { type GetCardMenuItemParams } from './card-menu-items';
+import { type MenuItemOptions } from '@cardstack/boxel-ui/helpers';
+import SetSiteConfigCommand from '@cardstack/boxel-host/commands/set-site-config';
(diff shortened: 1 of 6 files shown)


backspace:
The failing realm server test does seem relevant:

```
not ok 16 atomic-endpoints-test.ts > Realm-specific Endpoints: can make request to post /_atomic > validation > rejects update when resource does not exist
  ---
  message: "Promise rejected during \"rejects update when resource does not exist\": expected 404 \"Not Found\", got 201 \"Created\""
  severity: failed
  stack: |
    Error: expected 404 "Not Found", got 201 "Created"
        at Object.<anonymous> (/home/runner/work/boxel/boxel/packages/realm-server/tests/atomic-endpoints-test.ts:899:14)
```

backspace (approved):

PR #3586, opened 2025-11-19
Field showcase card

richardhjtan:
### What is changing
- introduce a field showcase, inspired by Storybook, to consolidate catalog field examples
- add edit theme field to try theming on the spot

https://www.loom.com/share/a37c8b85be074892b4985b8a33b22329


burieberry (commented):
Can we use boxel-ui components in the template and import lucide icons from boxel-icons for those that exist there?

tintinthong (approved):

PR #3589, opened 2025-11-20
Show tool call description in preparing state of apply button

FadhlanR:
Before:

https://github.com/user-attachments/assets/19e8354d-7bf5-44ce-aaaa-5e3623cf7504

After:

https://github.com/user-attachments/assets/f23cd114-64b9-4edd-9a69-2a26fd116809


jurgenwerk (approved):

packages/host/app/components/matrix/room-message-command.gts
```diff
@@ -169,7 +169,9 @@ export default class RoomMessageCommand extends Component<Signature> {
       ...attributes
     >
       {{#if @isStreaming}}
-        <PreparingRoomMessageCommand />
+        <PreparingRoomMessageCommand
+          @commandDescription={{@messageCommand.description}}
+        />
       {{else}}
         <CodeBlock
```
burieberry: Can the `CodeBlock` component be used instead of duplicating the css?

PR #3590, opened 2025-11-20
Highlights should only show on "/personal" realm

FadhlanR:
https://github.com/user-attachments/assets/32330a64-540d-4936-8d5e-c7e15fbc2b0f


burieberry:
This change seems to be working. Question: Do we create a /personal/ realm for the user at one point?

FadhlanR:
> This change seems to be working. Question: Do we create a /personal/ realm for the user at one point?

Yes, the /personal/ realm is created as part of the user registration process.

lukemelia (approved):

PR #3591, opened 2025-11-20
Make check box to checked if custom subdomain is already claimed

FadhlanR:
This PR makes default check box for custom subdomain to checked if the subdomain is already claimed.

Before:

https://github.com/user-attachments/assets/4e0c7251-584b-4a17-a6eb-31aadafc4d34

After:

https://github.com/user-attachments/assets/e0d6f297-6812-4fb0-b5ae-20df8a902ef3


backspace:
This is (part of?) the Matrix test failure:

Playwright Trace Viewer 2025-11-21 08-20-52

I tried this and it worked as expected, though I do wonder about this edge case: I refreshed after having claimed and then the box wasn’t checked. Is that expected? Maybe if the subdomain has been claimed but the site hasn’t been published, the box should default to checked? But I’m unsure about this… because if it’s been published and you refresh, neither box is checked.

FadhlanR:
> This is (part of?) the Matrix test failure:
> 
> Playwright Trace Viewer 2025-11-21 08-20-52
> I tried this and it worked as expected, though I do wonder about this edge case: I refreshed after having claimed and then the box wasn’t checked. Is that expected? Maybe if the subdomain has been claimed but the site hasn’t been published, the box should default to checked? But I’m unsure about this… because if it’s been published and you refresh, neither box is checked.

Thanks for the feedback. I've updated the checkbox behavior based on your feedback, now the checkbox will be checked if the subdomain already claimed even after user refreshing the page.

backspace (approved):

PR #3592, opened 2025-11-20
Fix flaky file watcher test

IanCal:
See logs here: https://github.com/cardstack/boxel/actions/runs/19534405076/job/55924903477

Although the above didn't fail, the issue is there. 

The tests were waiting for *any* event to appear and then checking that a *specific* event appeared. Here's the ordering 

```[
  '{"type":"app.boxel.realm-event","room_id":"!uZQZdQdUpjBRXewAXP:localhost","sender":"@node-test_realm:localhost","content":{"eventName":"update","added":"new-file.json"},"origin_server_ts":1763636351616,"unsigned":{"membership":"join","age":164},"event_id":"$l_PhrUdn1DjO5QNZxfB5Vezz6ELnev5mEbDFuHE_4go","user_id":"@node-test_realm:localhost","age":164}',
  '{"type":"app.boxel.realm-event","room_id":"!uZQZdQdUpjBRXewAXP:localhost","sender":"@node-test_realm:localhost","content":{"eventName":"index","indexType":"incremental-index-initiation","updatedFile":"http://127.0.0.1:4444/new-file.json"},"origin_server_ts":1763636351610,"unsigned":{"membership":"join","age":170},"event_id":"$hKUBVM7mGCgo48Uzt4t-zSnhOPxZd4HgwmFwzMhnXG4","user_id":"@node-test_realm:localhost","age":170}',
  '{"type":"m.room.message","room_id":"!uZQZdQdUpjBRXewAXP:localhost","sender":"@node-test_realm:localhost","content":{"body":"auth-response: b8ae72ea-4050-4692-bf1d-08ef48894bc4","msgtype":"m.text"},"origin_server_ts":1763636351384,"unsigned":
```

The one tested for is the update event, but the index event appeared first. Locally I see this the other way around so I assume it isn't guaranteed.

Since we wait for the first event, but then *re-fetch* the messages before asserting, there are two timing issues that have to happen together for this to fail. One is that the events occur in this order. The second is that the update event doesn't arrive before we do the final test.

I added some helpers for waiting for a specific event, there are other tests that do similar things that possibly have the same kind of issue.

### DIFF
--- a/packages/realm-server/tests/file-watcher-events-test.ts
+++ b/packages/realm-server/tests/file-watcher-events-test.ts
@@ -99,6 +99,7 @@ module(basename(__filename), function () {
     let { getMessagesSince } = setupMatrixRoom(hooks, getRealmSetup);
 
     test('file creation produces an added event', async function (assert) {
+      console.log('Starting file creation test');
       realmEventTimestampStart = Date.now();
 
       let newFilePath = join(
@@ -126,6 +127,7 @@ module(basename(__filename), function () {
 
       await waitForRealmEvent(getMessagesSince, realmEventTimestampStart);
       let messages = await getMessagesSince(realmEventTimestampStart);
+      console.log('Messages since', realmEventTimestampStart, messages);
       let updateEvent = findRealmEvent(messages, 'update', 'incremental');
 
       assert.deepEqual(updateEvent?.content, {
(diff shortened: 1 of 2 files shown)


habdelra:
I have a fix as well for this in https://github.com/cardstack/boxel/pull/3576

PR #3594, opened 2025-11-20
New definitions lookup

lukemelia:


packages/realm-server/tests/definition-lookup-test.ts
```diff
@@ -0,0 +1,258 @@
...
+      });
+      assert.strictEqual(definition?.displayName, 'Person');
+      assert.strictEqual(
+        prerenderModuleCalls,
+        2,
+        'prerenderModule was called a second time after invalidation',
+      );
+    });
+
+    test('uses public cache scope when realm is public', async function (assert) {
```
ef4, line 214: Additional tests to add:
 - [x] same server public
 - [ ] cross server public
 - [ ] same server private
 - [ ] cross server private

packages/runtime-common/definition-lookup.ts
```diff
@@ -0,0 +1,426 @@
...
+  async invalidate(realmURL: string): Promise<void> {
+    await this.#inner.invalidate(realmURL);
+  }
+
+  registerRealm(realm: LocalRealm): void {
+    this.#inner.registerRealm(realm);
+  }
+}
+
+function parseJSON<T>(value: unknown, fallback: T): T {
```
ef4: Not needed (cast the db result type once, and no manual parsing should ever be required).

ef4:
Status: we think the global lazilyLoadLinks flag might be leaking and causing an instance prerender to hit NotLoaded.

Debug logging branch is `move-definitions-debug-logging`.

habdelra:
the prerendering-test flakiness should be fixed if you merge upstream

PR #3596, opened 2025-11-20
Add head format

backspace:
This adds a new `head` format for specifying HTML elements that belong there like `meta` for Open Graph and `title`. There’s a future issue to add a proper preview but this includes a hackish placeholder because otherwise the preview would show nothing:

ramped json in minicatalog 2025-11-21 09-56-35

A proper preview would look something like this:

Meta Tags — Preview, Edit and Generate 2025-11-21 09-59-16


packages/base/field-component.gts
```diff
@@ -538,6 +538,8 @@ function defaultFieldFormats(containingFormat: Format): FieldFormats {
       return { fieldDef: 'embedded', cardDef: 'fitted' };
     case 'atom':
       return { fieldDef: 'atom', cardDef: 'atom' };
+    case 'head':
+      return { fieldDef: 'head', cardDef: 'head' };
```
backspace, line 542: Does this make sense?

habdelra:
I think this also needs a test in host/acceptance/prerender-html-test.gts. 
the way it breaks down:
1. host/acceptance/prerender-html-test.gts tests the meat of the /render route
2. realm-server/tests/prerendering-test.gts tests the capture mechanism (only possible in the realm server, the host tests have a special capture mechanism such that the routing for indexing doesn't interfere with the routing of the app since they are all running in the same ember runtime together)
3. integration/tests/realm-indexing-render-route.gts tests the in browser indexing using our "special" host capture mechanism
4.  integration/tests/realm-indexing.gts the in browser old indexing mechanism
5. realm-server/tests/headless-chrome-indexing tests the indexers ability to use the real prerendering mechanism for testing
6. realm-server/tests/-indexing tests the old indexing mechanism using fastboot

habdelra:
in your PR comment the screenshot has a card container around the head in the playground viewer. we should not include that right?

backspace:
> in your PR comment the screenshot has a card container around the head in the playground viewer. we should not include that right?

That’s the desire but I think that’s an artefact of the preview component, at least as far as I can tell… in the followup PR with indexing, the container isn’t included:

boxel_index head html — boxel — boxel realm server 2025-11-25 11-04-29

I’ll see if I can get rid of it in the preview, but it’s only a placeholder anyway.

backspace:
> I think this also needs a test in host/acceptance/prerender-html-test.gts. the way it breaks down:
> 
> 1. host/acceptance/prerender-html-test.gts tests the meat of the /render route
> 2. realm-server/tests/prerendering-test.gts tests the capture mechanism (only possible in the realm server, the host tests have a special capture mechanism such that the routing for indexing doesn't interfere with the routing of the app since they are all running in the same ember runtime together)
> 3. integration/tests/realm-indexing-render-route.gts tests the in browser indexing using our "special" host capture mechanism
> 4. integration/tests/realm-indexing.gts the in browser old indexing mechanism
> 5. realm-server/tests/headless-chrome-indexing tests the indexers ability to use the real prerendering mechanism for testing
> 6. realm-server/tests/-indexing tests the old indexing mechanism using fastboot

thanks for this, do these all apply if this isn’t about storing prerendered cards, just about adding head format support?

habdelra:
> > I think this also needs a test in host/acceptance/prerender-html-test.gts. the way it breaks down:
> > 
> > 1. host/acceptance/prerender-html-test.gts tests the meat of the /render route
> > 2. realm-server/tests/prerendering-test.gts tests the capture mechanism (only possible in the realm server, the host tests have a special capture mechanism such that the routing for indexing doesn't interfere with the routing of the app since they are all running in the same ember runtime together)
> > 3. integration/tests/realm-indexing-render-route.gts tests the in browser indexing using our "special" host capture mechanism
> > 4. integration/tests/realm-indexing.gts the in browser old indexing mechanism
> > 5. realm-server/tests/headless-chrome-indexing tests the indexers ability to use the real prerendering mechanism for testing
> > 6. realm-server/tests/-indexing tests the old indexing mechanism using fastboot
> 
> thanks for this, do these all apply if this isn’t about storing prerendered cards, just about adding head format support?

they do--like i said, you want to make sure that your capture is correct and that you are not including the card boundary. you need to have a test that is asserting that, for both host and realm server (since the capture implementation differs between the two and we are still committed to running a realm server in the browser). i would think that you would want to follow the same set of tests that we use for atom format....

habdelra:
these two comments from co-pilot are also relavant (co-pilot seemed unsure):

> - Missing column type specification for head_html. This should be head_html TEXT to match the pattern used for other HTML columns like isolated_html TEXT (line 15) and atom_html TEXT (line 20).
packages/host/config/schema/1762000000000_schema.sql:52
> - Missing column type specification for head_html in boxel_index_working table. This should be head_html TEXT to match the pattern used for other HTML columns and to be consistent with the boxel_index table definition.

It seems odd that the schema generation resulted in a different column type for head_html vs for isolated_html and atom_html. probably you should look into that.

backspace:
> these two comments from co-pilot are also relavant (co-pilot seemed unsure):
> 
> > * Missing column type specification for head_html. This should be head_html TEXT to match the pattern used for other HTML columns like isolated_html TEXT (line 15) and atom_html TEXT (line 20).
> >   packages/host/config/schema/1762000000000_schema.sql:52
> > * Missing column type specification for head_html in boxel_index_working table. This should be head_html TEXT to match the pattern used for other HTML columns and to be consistent with the boxel_index table definition.
> 
> It seems odd that the schema generation resulted in a different column type for head_html vs for isolated_html and atom_html. probably you should look into that.

hmm yes… the migration used `text` but the SQLite conversion doesn’t include that. I changed it to `varchar` like the others which results in `text` for SQLite! thanks

backspace:
> I think this also needs a test in host/acceptance/prerender-html-test.gts. the way it breaks down:
> 
> 1. host/acceptance/prerender-html-test.gts tests the meat of the /render route
> 2. realm-server/tests/prerendering-test.gts tests the capture mechanism (only possible in the realm server, the host tests have a special capture mechanism such that the routing for indexing doesn't interfere with the routing of the app since they are all running in the same ember runtime together)
> 3. integration/tests/realm-indexing-render-route.gts tests the in browser indexing using our "special" host capture mechanism
> 4. integration/tests/realm-indexing.gts the in browser old indexing mechanism
> 5. realm-server/tests/headless-chrome-indexing tests the indexers ability to use the real prerendering mechanism for testing
> 6. realm-server/tests/-indexing tests the old indexing mechanism using fastboot

I’ve added tests either here or in #3614 based on whether they were database-related:

1. [here](https://github.com/cardstack/boxel/pull/3596/files#diff-585183a109c564e6a819e313140e1130a6e67f94152d0136e2032bf910463838)
2. [here](https://github.com/cardstack/boxel/pull/3596/files#diff-fb1167db22f061ab8bcbd710d3c489c5f66287405757d60e88d352b17a089aec)
3. [here](https://github.com/cardstack/boxel/pull/3596/files#diff-b68687421ba9c53bb974ca7e1fd7db768cdaa7cbeaf10f861bb739e0b00f9ba6)
4. [here](https://github.com/cardstack/boxel/pull/3596/files#diff-fda2eb0ca2e53f54807cb49ace9dadb78e1a87ae564eae5397f639663a7525e8) (with changes to `current-run`)
5. moved [here](https://github.com/cardstack/boxel/pull/3596/files#diff-777e0d0405d10dbc761d604f260253f972f3e4a6f08c7cad199ae7c365c02a49)
6. [there](https://github.com/cardstack/boxel/pull/3614/files#diff-a065ae1210d36bdb863cbd431bc7971c7d0c7911b21ba42c5a494d4477fdecf9)

Thanks for spelling it all out @habdelra, it’s hard to know where everything is.

habdelra (approved):

PR #3597, opened 2025-11-21
Implement patch realm.json endpoint

FadhlanR:
In this PR, I introduce a new endpoint to patch the `.realm.json` file. This endpoint will be used to set the home card in both interact mode and host mode. For this reason, I added two new properties to the realm info: `interactHome` and `hostHome`.


packages/runtime-common/realm.ts
```diff
@@ -3102,6 +3118,101 @@ export class Realm {
...
+          : 'hostHome must be a string or null',
+      publishable: (val) =>
+        val === null || typeof val === 'boolean'
+          ? undefined
+          : 'publishable must be a boolean or null',
+      visibility: (val) =>
+        val === 'private' || val === 'shared' || val === 'public'
+          ? undefined
+          : "visibility must be 'private', 'shared', or 'public'",
+    };
```
backspace: This is readable but we’re adding enough validation that I wonder if we should be looking at Zod or the like

packages/runtime-common/realm.ts
```diff
@@ -3102,6 +3118,101 @@ export class Realm {
...
+    if (existingConfig?.content) {
+      try {
+        realmConfig = JSON.parse(existingConfig.content);
+      } catch (e: any) {
+        return systemError({
+          requestContext,
+          message: `Unable to parse existing realm config: ${e.message}`,
+        });
+      }
+    }
```
backspace, line 3203: If this is only for patching, how does someone create the file initially?

backspace (approved):

habdelra:
Why can't we use the HTTP POST with card-source accept header instead of this new endpoint?

FadhlanR:
> Why can't we use the HTTP POST with card-source accept header instead of this new endpoint?

Based on the discussion with @lukemelia , we want to expose an endpoint that accepts only the subset of .realm.json properties the client intends to update. We also want to prevent the client from updating certain properties. What do you think @lukemelia ?

packages/runtime-common/realm.ts
FadhlanR: This file will be created during realm creation.

FadhlanR:
@lukemelia  @habdelra  I made the following updates based on our discussion:

* Instead of using `_info` as the path, I used `_config` so it can still be used in the future when we remove the `.realm.json` approach.
* All properties can be updated by the realm owner, except for `showAsCatalog`.

packages/runtime-common/realm.ts
```diff
@@ -3109,6 +3126,65 @@ export class Realm {
...
+
+    const realmConfigPatchSchema = z.object({
+      property: z
+        .string()
+        .min(1)
+        .refine(
+          (property: any) => property !== 'showAsCatalog',
+          'showAsCatalog cannot be updated',
+        ),
```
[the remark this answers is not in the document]
FadhlanR: I don't think this is necessary, since we don’t want to update the code every time we introduce a new property.

PR #3599, opened 2025-11-21
capture testem.logs in CI

habdelra:
To help with debugging failed fetch and chrome crashing we capture testem logs

### DIFF
--- a/packages/host/testem.js
+++ b/packages/host/testem.js
@@ -4,6 +4,7 @@ const MultiReporter = require('testem-multi-reporter');
 const TapReporter = require('testem/lib/reporters/tap_reporter');
 const XunitReporter = require('testem/lib/reporters/xunit_reporter');
 const fs = require('fs');
+const path = require('path');
 
 const config = {
   test_page: 'tests/index.html?hidepassed',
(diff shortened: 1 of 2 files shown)


packages/host/testem.js
```diff
@@ -31,19 +32,27 @@ const config = {
...
 
   const reporters = [
     {
       ReporterClass: TapReporter,
       args: [false, null, { get: () => false }],
     },
+    {
+      ReporterClass: TapReporter,
+      args: [false, testemLog, { get: () => false }],
+    },
```
backspace, line 49: Is there an advantage to doing it this way vs hoisting it to happen in the CI script like [this](https://github.com/cardstack/boxel/blob/ac551ef65d5e1daa8cb3e98985d670a6e9f9ffc4/.github/workflows/ci-host.yaml#L63-L65)?

habdelra: it was Codex's call to do it that way. the rationale is:
> Doing it inside Testem has a couple of small advantages over tee-ing in the workflow:
> - packages/host/testem.js emits a clean TAP stream directly from the reporter, so the log file contains only test output. A workflow-level tee would mix in Percy/stdout noise and any other logging the script produces.
> - It travels with the test config, so anyone running the suite locally or in a different CI gets the same log artifact without having to mirror the workflow plumbing.
> - The file is created only when tests actually start; we don’t depend on the workflow wiring to have the right path and options.

PR #3600, opened 2025-11-21
enhancement to file watcher test assertion

habdelra:
I still spotted a file watcher flaky test failure even though now we are asserting on the specific event type for the file watcher tests. Codex thinks this is the issue:

> Fix ensures the update events aren’t filtered out when they happen in the same millisecond as the timestamp we record before the file change. In `packages/realm-server/tests/helpers/index.ts` the Matrix message filter now uses `origin_server_ts >= since` (with a note about clock skew), so an “updated” event emitted immediately after the timestamp is no longer skipped—listener wiring was fine; the cutoff was too strict.


backspace:
I reran the job a few times and got [a failure](https://github.com/cardstack/boxel/actions/runs/19577728741/job/56085812461?pr=3600#step:10:44), it could be another problem, but maybe a way to know better would be to print out the messages that were received so we can see whether what actually arrived was close in some way to what was expected.

habdelra:
> I reran the job a few times and got [a failure](https://github.com/cardstack/boxel/actions/runs/19577728741/job/56085812461?pr=3600#step:10:44), it could be another problem, but maybe a way to know better would be to print out the messages that were received so we can see whether what actually arrived was close in some way to what was expected.

yeah that's the same failure that i got that this PR is in response to

habdelra:
@copilot please update this PR to console.log the events that were recieved in this test so we can figure out what is going on

PR #3605, opened 2025-11-21
server: Add head tags when head_html exists

backspace:
This serves `boxel_index.head_html` within delimiter comments of `host` application’s `index.html` to show link previews and the like, here it is on Discord when this branch was deployed:

#neochamps | sportsracers - Discord 2025-11-28 13-59-35

Implementation notes and future work:
* `<title>` was causing Ember to crash because `ember-page-title` throws when multiple titles are found, so I removed it from the default head template for now
* there may be a cleaner approach than the `HEADSTART`/`HEADEND` delimiter comments, work for another issue
* published realms are inheriting `head_html` from the origin realm instead of it being re-prerendered
   * the default head template derives canonical URL from card id, but that id is different in a published realm
   * this seems like a bigger problem but I’m not that familiar with indexing
   * so for now I’ve commented out the Matrix test assertion on this


packages/realm-server/server.ts
```diff
@@ -292,7 +293,13 @@ export class RealmServer {
       }
 
       ctxt.type = 'html';
-      ctxt.body = await this.retrieveIndexHTML();
+      let indexHTML = await this.retrieveIndexHTML();
+      let headHTML = await this.retrieveHeadHTML(
+        new URL(`${ctxt.protocol}://${ctxt.host}${ctxt.originalUrl}`),
+      );
```
lukemelia, line 299: We could grab these in parallel

lukemelia (approved):

PR #3609, opened 2025-11-24
Fix test on main

tintinthong:


tintinthong:
Still some flaky test but unrelated to merged changes

PR #3611, opened 2025-11-24
Increase run time for ai generation: 10 retries in 10mins (it was 5 seconds before)

jurgenwerk:
After ai bot finishes generating an AI response, it tries to fetch the cost for that generation, and deduct the credit amount from the user's credit balance. 

I observed the endpoint currently needs around 10-30 seconds to return a result, and responds with 500 in the meanwhile. The current logic does not wait for long enough (currently only around 4-5 seconds), and it will swallow the error and not deduct users' credits. This means that users currently aren't spending any credits to use Boxel. I am not sure what the reason for the larger delay of Open Router's generation cost reporting is, but we must adapt to it, and make sure we get informed if this fails. 

This PR increases the run time for the retry/backoff logic to 10 retries in 10 minutes, and makes sure an error is reported to Sentry in case it is not able to get the cost that deducts the users' credits.

### DIFF
--- a/packages/billing/ai-billing.ts
+++ b/packages/billing/ai-billing.ts
@@ -7,6 +7,7 @@ import {
   type DBAdapter,
   MINIMUM_AI_CREDITS_TO_CONTINUE,
   logger,
+  delay,
 } from '@cardstack/runtime-common';
 import * as Sentry from '@sentry/node';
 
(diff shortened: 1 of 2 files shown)


jurgenwerk:
@codex review

packages/billing/ai-billing.ts
```diff
@@ -103,6 +105,42 @@ export async function saveUsageCost(
...
+    }
+
+    let elapsed = Date.now() - startedAt;
+    if (attempt === MAX_FETCH_ATTEMPTS || elapsed >= MAX_FETCH_RUNTIME_MS) {
+      break;
+    }
+
+    let remainingTime = MAX_FETCH_RUNTIME_MS - elapsed;
+    let sleepMs = Math.min(delayMs, remainingTime);
+    await new Promise((resolve) => setTimeout(resolve, sleepMs));
```
backspace: We have `delay` for this in `runtime-common` if you want to reduce duplication

packages/billing/ai-billing.ts
```diff
@@ -121,7 +159,6 @@ async function fetchGenerationCost(
   if (data.error && data.error.message.includes('not found')) {
     return null;
   }
-
```
backspace, line 124: why 😆

backspace (approved):

PR #3614, opened 2025-11-24
Add head format to indexing

backspace:
This updates indexing to store the `head_html` of a card’s head format.

The failing test is a known flaky one, [here](https://github.com/cardstack/boxel/actions/runs/19711084891/job/56471605783#step:12:640) it is failing on `main`.


habdelra:
we should probably have tests in the realm-server/tests/prerendering-test.ts as well, and consider tests in hosts/tests/acceptance/prerender-html-test.gts. you'll want to make sure that you are capturing the right boundary (i'm assuming you want the card container to not be part of the captured HTML, which should be asserting in tests). you might also want legacy tests for the old indexer too since prod is still not using headless chrome.

backspace:
> we should probably have tests in the realm-server/tests/prerendering-test.ts as well, and consider tests in hosts/tests/acceptance/prerender-html-test.gts. you'll want to make sure that you are capturing the right boundary (i'm assuming you want the card container to be part of the captured HTML). you might also want legacy tests for the old indexer too since prod is still not using headless chrome.

thanks, I tried to find the relevant places where tests were needed but it was a struggle. I’ll look into these. I don’t actually want the card container to be part of the HTML since it’ll be inserted into `<head>` and a container element doesn’t make sense there.

habdelra:
> > we should probably have tests in the realm-server/tests/prerendering-test.ts as well, and consider tests in hosts/tests/acceptance/prerender-html-test.gts. you'll want to make sure that you are capturing the right boundary (i'm assuming you want the card container to be part of the captured HTML). you might also want legacy tests for the old indexer too since prod is still not using headless chrome.
> 
> thanks, I tried to find the relevant places where tests were needed but it was a struggle. I’ll look into these. I don’t actually want the card container to be part of the HTML since it’ll be inserted into `<head>` and a container element doesn’t make sense there.

makes sense. its  amatter of specifying whether to use innerHTML (no card container) or outerHTML (with card container) in the capture options. but yeah, tests would be important for asserting this. I wasn't sure whether the testing should be here or in https://github.com/cardstack/boxel/pull/3596 or both...

habdelra (approved):

PR #3620, opened 2025-11-25
Add card instance correctness checking to CheckCorrectnessCommand

jurgenwerk:
This PR adds an implementation to `CheckCorrectnessCommand` where it checks whether the card instance (json) that was just patched using the assistant (either with the `PatchCardInstanceCommand`, or `PatchCodeCommand`) entered an error state (i.e. has an error document in the store). If yes, it will report the error and the LLM will try to fix it. Otherwise it will report the card as "correct".


packages/host/app/commands/check-correctness.ts
```diff
@@ -1,11 +1,33 @@
...
 import HostBaseCommand from '../lib/host-base-command';
 
+import { waitForRealmState } from './utils';
+
+import type CardService from '../services/card-service';
+import type CommandService from '../services/command-service';
+import type RealmService from '../services/realm';
+import type StoreService from '../services/store';
+
+const CARD_INDEX_TIMEOUT_MS = 15_000;
```
habdelra: is this the amount of time to wait for indexing? our prerenderer actually has this as well, which is probably the authorative source: realm-server/prerender/utils.ts `renderTimeoutMs`. perhaps we should centralize this and you can import that?

habdelra (approved):

packages/host/app/commands/check-correctness.ts
jurgenwerk: Yes, this is the amount of time the correctness check will wait for the invalidation event to arrive. So I think it maps to what you described. I adjusted this in https://github.com/cardstack/boxel/pull/3620/commits/74b74f85a431be29cd5bef34ca5731e157416bf9

lukemelia (approved):

PR #3630, opened 2025-11-27
Remove multipleImageField and centralize edit view rendering in imageField

lucaslyl:
Remove multipleImageField and allow imageField to delegate rendering for both the contains and containsMany edit views.

## pending:

- test
- field showcase update


lucaslyl:
We will investigate this task later, as we need additional discussion on how to handle the overrides of the containsMany template. cc: @tintinthong

PR #3632, opened 2025-11-28
Eco 431 number field rating gauge slider fixed

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-431/number-field-rating-gauge-slider-fixed

Uploading Screen Recording 2025-11-28 at 5.09.56 PM.mov…


tintinthong:
uhmmm what does this PR do?

lucaslyl:
> uhmmm what does this PR do?

I consolidated all the fields into a number field with presentation usage. This means that switching the presentation mode will display different embedded views. I also added an options section in the field showcase so users know which configuration options they can choose from.

tintinthong (approved):
ok u moved a lot of files

PR #3634, opened 2025-12-01
Add copy and edit for Theme Cards

tintinthong:
- Added Copy and Edit flow: copies a card in-place realm, renames it “(Copy ####)”, relinks parent using stack relationship context, and replaces the stack item in edit mode for both linksTo and linksToMany (including cross-realm when types match).
- Extended stack items to carry relationship metadata and added replaceCardOnStack to operator-mode-state-service for precise stack swaps.
- Theme menu now includes a Copy and Edit action wired to the new command.


packages/host/app/commands/copy-and-edit.ts
```diff
@@ -0,0 +1,267 @@
...
+        let itemIndex = stack.indexOf(item);
+        if (itemIndex > 0) {
+          let parentItem = stack[itemIndex - 1];
+          return {
+            parentId: parentItem.id,
+            relationshipContext: item.relationshipContext,
+          };
+        }
+      } catch {
+        // not in this stack, continue
```
jurgenwerk: To avoid catch-all, can we be more specific here in a way where we only catch the specific potential throw in `findCardInStack`?

jurgenwerk (approved):

packages/host/app/commands/copy-and-edit.ts
```diff
@@ -0,0 +1,264 @@
...
+    let targetRealm = this.realm.url(input.card.id as string);
+    if (!targetRealm) {
+      throw new Error('Could not determine realm for card copy');
+    }
+    if (!targetRealm.endsWith('/')) {
+      targetRealm = `${targetRealm}/`;
+    }
+    if (!this.realm.canWrite(targetRealm)) {
+      throw new Error(`Do not have write permissions to ${targetRealm}`);
+    }
```
burieberry, line 59: this is throwing for me for a theme card on catalog because targetRealm is still catalog. is this only meant for theme cards in writable realms?

burieberry (approved):
Looks good for same realm copy and edit.

packages/host/app/commands/copy-and-edit.ts
tintinthong: I think you highlight quite an important issue. If most themes exist in catalog and are linked from there. My current behavior is to copy into the same place that the theme is from which seems problematic if ppl just constantly link themes from catalog.  I think a more reasonable way is to TRY to copy into the realm of your interact submode. 


EDIT: Yea I think most of the time ppl want to make the copy into the workspace their looking at

tintinthong:
I have enhanced this PR to handle the case where only a single theme card on a stack is opened. A person should still be able to Copy and Edit and replace that single theme card.

PR #3635, opened 2025-12-01
Audio field with presentations

richardhjtan:
### What is changing
- add an audio field also to provide presentations
- add field showcase example
- cover integration test

https://www.loom.com/share/6ecdb0ed3bdb42fba19cd854dbacf8bd


lukemelia (approved):
I did not review in depth but the approach looks good

packages/catalog-realm/fields/components/base-audio-player.gts
```diff
@@ -0,0 +1,347 @@
...
+
+  // ¹⁸ Configuration getters
+  get presentationStyle() {
+    return (
+      (this.args as any).configuration?.presentation?.style ?? 'inline-player'
+    );
+  }
+
+  get showVolume() {
+    return (this.args as any).configuration?.presentation?.showVolume ?? false;
```
lucaslyl: Perhaps put those  controls into configuration options: { ... } like we discussed last week? So it will be consistent for fields 

example:
```
options: {
  showVolume : ....
  etc
}
```

packages/catalog-realm/fields/components/base-audio-player.gts
```diff
@@ -0,0 +1,347 @@
...
+  }
+
+  formatFileSize(bytes: number): string {
+    if (!bytes) return '0 MB';
+    const mb = (bytes / (1024 * 1024)).toFixed(1);
+    return `${mb} MB`;
+  }
+
+  // ¹⁸ Configuration getters
+  get presentationStyle() {
```
lucaslyl: Same here, we could simplify it by using presentation: string 👀

image

lucaslyl:
I’ve added a few comments in that section. The UI looks good!

packages/catalog-realm/fields/components/base-audio-player.gts
```diff
@@ -0,0 +1,347 @@
...
+    return (
+      (this.args as any).configuration?.presentation?.style ?? 'inline-player'
```
lucaslyl: one small suggestion ,we could refine some of the types in certain parts of the code(maybe for config part if possible), since I noticed there are several "as any" usages.

packages/catalog-realm/fields/components/base-audio-player.gts
richardhjtan: Agreed, updated the API configuration structure in latest comment

lucaslyl (approved):

PR #3636, opened 2025-12-01
Transform and transpile asynchronously

IanCal:
Currently we're using synchronous transformation/transpilation on the realm server and in the host. This  blocks the thread, a more pressing issue with the realm server, and there is an async version available to us. Places where this is used are already async functions.


jurgenwerk (approved):

habdelra:
the prerendering-test flakiness should be fixed if you merge upstream

PR #3657, opened 2025-12-03
Remove globalThis.__lazilyLoadLinks feature flag

habdelra:
This is the first of a pile of feature removal PR's. I initially tried to do it all in one big bang and everything broke and it was unclear why. so i'm doing this incrementally keeping the tests green as I go. 

This PR removes the `globalThis.__lazilyLoadLinks` feature flag.


packages/host/tests/integration/card-prerender-test.gts
habdelra, line 1: this tests the old card-prerender component that we'd use in fastboot for indexing. all these tests are obsolete (we still need this component, but it has a very different purpose now, and it is tested in our host indexing tests in the new way that we are using it)

packages/host/tests/integration/realm-indexing-render-route-test.gts
habdelra, line 1: this test was renamed to "realm-indexing-test.gts", and the old "realm-indexing-test.gts" is gone.

packages/realm-server/tests/headless-chrome-indexing-test.ts
habdelra, line 1: This test was renamed to indexing-test.ts and the old indexing-test.ts is gone.

jurgenwerk (approved):

PR #3659, opened 2025-12-04
Fix host home URL in published realm

FadhlanR:
**Problem**

- In host mode, the alternate home page (site config card) is stored in .realm.json as hostHome, using the source realm URL because selection happens there.
- Publishing copies .realm.json verbatim, so the published realm keeps a hostHome that points back to the source realm and tries to fetch site config from the wrong host.

**Solution**

During publishing, detect when hostHome starts with the source realm URL and rewrite it to use the published realm URL before writing .realm.json for the published realm.


backspace:
this seems to have some overlap with issue 9805, is there a possibility of solving it in a more general way?

jurgenwerk (approved):

PR #3660, opened 2025-12-04
Add module correctness checking to CheckCorrectnessCommand

jurgenwerk:
This PR adds .gts files correctness checks after AI assistant makes changes to them. The goal of this is to catch and auto correct errors in the AI-generated gts file code. These errors can be syntax errors, issues with code importing,...

It leverages our correctness checks mechanism where we already check for correctness of card instances, but this now also adds card definition checking (module). This is done using the `prerender-module` endpoint in the prerender server which is proxied through the realm server.

Here's how it works (I intentionally told the LLM to introduce an error, but this should also work when AI makes an error in generated code when vibe coding in real world scenarios):
image


jurgenwerk:
@codex review

packages/host/app/commands/check-correctness.ts
```diff
@@ -142,60 +135,105 @@ export default class CheckCorrectnessCommand extends HostBaseCommand<
...
-        `CheckCorrectnessCommand: unable to determine realm for ${cardId}`,
-      );
-      return;
+    let { moduleURL, realmURL } = moduleInfo;
+
+    await this.commandService.waitForInvalidationAfterAIAssistantRequest(
+      moduleURL.href,
+      roomId,
+      cardIndexingTimeout,
+    );
```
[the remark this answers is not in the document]
jurgenwerk: Will check

lukemelia (approved):

jurgenwerk:
@codex review

jurgenwerk:
@codex review

packages/host/app/services/command-service.ts
```diff
@@ -455,22 +494,14 @@ export default class CommandService extends Service {
...
-        );
+
         await this.store.patch(
           cardId,
           {
             attributes: payload?.attributes?.patch?.attributes,
             relationships: payload?.attributes?.patch?.relationships,
           },
-          { doNotWaitForPersist: true, clientRequestId },
+          { doNotWaitForPersist: true },
```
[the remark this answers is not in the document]
jurgenwerk: I'm checking with the team because it's not clear to me how we get to here - I thought PatchCardInstanceCommand is responsible for patching

jurgenwerk:
@habdelra could you check if things look ok to you, namely the prerender proxy, and waiting for an invalidation after card/code patching originating from AI assistant?

packages/realm-server/tests/prerender-proxy-test.ts
```diff
@@ -0,0 +1,448 @@
...
+        .send({ data: { attributes: {} } })
+        .expect(500);
+
+      assert.ok(
+        res.text.includes('Prerender proxy is not configured'),
+        'returns a useful error message when upstream is missing',
+      );
+    });
+
+    test('returns forbidden when user has no realm permissions', async function (assert) {
```
habdelra, line 243: how about a 401 test? 403 means the server knows who you are and you are not allowed. but 401 means the server doesn't know who you are (so no auth token would result in a 401)

packages/realm-server/handlers/handle-prerender-proxy.ts
```diff
@@ -0,0 +1,132 @@
...
+    headers.set('content-type', 'application/vnd.api+json');
+    headers.set('accept', 'application/vnd.api+json');
+
+    try {
+      let response = await fetch(upstream, {
+        method: 'POST',
+        headers,
+        body: forwardBody,
+      });
+      await setContextResponse(ctxt, response);
```
habdelra: instead of doing this, why don't you use the `remote-prerenderer.ts` it will do all this for you. we made a remote prerenderer when we were making the realm server in main.ts, just pass that into the realm server's constructor. it's already configured and good to go.

packages/realm-server/handlers/handle-prerender-proxy.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export default function handlePrerenderProxy({
+  path,
+  prerendererUrl,
+  dbAdapter,
+  createPrerenderAuth,
+}: {
+  path: '/prerender-card' | '/prerender-module';
+  prerendererUrl?: string;
+  timeoutMs?: number;
```
habdelra: is this being used? probably we can just let the remote-prerenderer take care of the timeouts for us

packages/realm-server/handlers/handle-prerender-proxy.ts
```diff
@@ -0,0 +1,132 @@
...
+  sendResponseForBadRequest,
+  sendResponseForForbiddenRequest,
+  sendResponseForSystemError,
+  setContextResponse,
+} from '../middleware';
+import type { RealmServerTokenClaim } from '../utils/jwt';
+
+export default function handlePrerenderProxy({
+  path,
+  prerendererUrl,
```
habdelra: pass  in the remote prerenderer instead of this

packages/realm-server/server.ts
```diff
@@ -98,6 +99,7 @@ export class RealmServer {
     getRegistrationSecret,
     enableFileWatcher,
     domainsForPublishedRealms,
+    prerendererUrl,
```
habdelra: instead of passing this in, pass in the remote prerenderer that we created in main.ts

packages/host/app/services/command-service.ts
```diff
@@ -92,61 +94,98 @@ export default class CommandService extends Service {
...
+      },
+      { timeoutMs: 5 * 60 * 1000, keepRealmSubscription: true }, // we don't wanna close the realm subscription since other places could be subscribed, like the store service
+    )
+      .then(() => {
+        let current = this.aiAssistantInvalidations.get(key);
+        current?.deferred.fulfill();
+      })
+      .catch(() => {
+        this.aiAssistantInvalidations.delete(key);
+      });
```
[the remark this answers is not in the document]
jurgenwerk: The purpose of `waitForInvalidationAfterAIAssistantRequest` is to be called *after* `trackAiAssistantCardRequest`, and in this case the deferred will always be properly set up. So the problem described is not really an issue here.

packages/realm-server/server.ts
jurgenwerk: Fixed

packages/realm-server/tests/prerender-proxy-test.ts
jurgenwerk: Added

packages/realm-server/handlers/handle-prerender-proxy.ts
jurgenwerk: Changed to use the remote prerenderer.

jurgenwerk:
@habdelra I used the remote prerenderer instead of re-implementing it. I didn't realize that already exists. I think this is now ready for another review.

habdelra (approved):

PR #3661, opened 2025-12-04
Implement index config card for alternate index card in interact submode

FadhlanR:
We implemented hostHome so users can set a card as an alternate home page in host mode. In this PR, we apply the same approach for the “home page” (index) in interact submode. I introduced a new card definition named IndexConfig; it’s a child of the SiteConfig card (we can modify it in the future if needed). There’s also a new menu item on an IndexConfig instance, “Set as interact home,” which updates the interactHome property in .realm.json and replaces the index card in stacks (if present). So, when a user opens the workspace in interact submode, the first stack item will be the card referenced by the IndexConfig home field.

https://github.com/user-attachments/assets/d566b731-de55-4cb3-912f-e244a86e4782


backspace:
What’s a good way for me to exercise this locally? I don’t have the “Set as interact home” option when I’m viewing a card when I have this branch running

Was there a discussion about terminology for this? To me it’s confusing that `SiteConfig` would be about host mode (is that right?) and `IndexConfig` is a child of it but refers to interact submode. Should the names instead refer to those (sub)modes directly? Because `index` is the usual name for visiting `/` on a site… it’s all mixing together, to me

FadhlanR:
> What’s a good way for me to exercise this locally? I don’t have the “Set as interact home” option when I’m viewing a card when I have this branch running

The "Set as interact home" menu is only available for an instance of index config card so you need to create the index config card instance first.

> Was there a discussion about terminology for this? To me it’s confusing that `SiteConfig` would be about host mode (is that right?) and `IndexConfig` is a child of it but refers to interact submode. Should the names instead refer to those (sub)modes directly? Because `index` is the usual name for visiting `/` on a site… it’s all mixing together, to me

Yes it's correct, the indexConfig is for the interact submode and the siteConfig is for the host submode/mode. Do you think it is better renaming it to HostHomeConfig and InteractHomeConfig since we name it hostHome and intractHome in realm.json file?

FadhlanR:
Please hold off on the review, we decided to implement another approach for the index card.

FadhlanR:
I'll open another PR with the new approach.

PR #3668, opened 2025-12-05
Fix download links for matrix event debug files

jurgenwerk:
Before this fix, clicking on the the debug file download button failed with a 404. This PR fixes the link so that the file can be downloaded. 

image


IanCal:
This works locally although I'm unsure as to why the change

IanCal (approved):

lukemelia (approved):

PR #3670, opened 2025-12-05
update agents.md to describe our software prereqs 

habdelra:
this makes it easier for codex running in linear to collaborate so it knows how to setup its environment

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -2,7 +2,7 @@
 
 ## Tooling prerequisites
 
-- We pin the toolchain with Volta (`.volta`), using Node.js 22.20.0 and pnpm 10.17.0. Install Volta and set `VOLTA_FEATURE_PNPM=1` so pnpm is managed automatically—avoid global installs outside Volta.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -1,5 +1,11 @@
 # Instructions for AI Agents
 
+## Tooling prerequisites
+
+- We pin the toolchain with Volta (`.volta`), using Node.js 22.20.0 and pnpm 10.17.0. Install Volta and set `VOLTA_FEATURE_PNPM=1` so pnpm is managed automatically—avoid global installs outside Volta.
```
lukemelia:
```suggestion
- We pin the toolchain with Volta (`.volta`), using the versions of Node.js and pnpm specified in package.json. Install Volta and set `VOLTA_FEATURE_PNPM=1` so pnpm is managed automatically—avoid global installs outside Volta.
```

AGENTS.md
```diff
@@ -1,5 +1,11 @@
...
+- pnpm is required for all scripts; use the pinned version above.
```
lukemelia:
```suggestion
- pnpm is required for all scripts; use the pinned version as specified above.
```

lukemelia (approved):

PR #3677, opened 2025-12-08
Allow responder responde to roomState initialPrompt

tintinthong:


tintinthong:
Deprioritising this task

PR #3684, opened 2025-12-09
Create Spec should use displayName for card or field

tintinthong:


lucaslyl (approved):
Cool, I just finished testing the spec creation from card/field, and the result is as expected.

PR #3685, opened 2025-12-09
Eco 432 color field

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-432/color-field

## Add the **showPalette** option to **ColorField** for a compact default design

Based on request: The current ColorField edit view with ColorPalette takes too much spacing. So I added a
configuration option where the default behavior uses only the ColorPicker Boxel UI component (compact).

- Default: Simple ColorPicker only (slim design)
- When `showPalette: true`: Shows ColorPalette with swatches (Boxel UI component)
- Reduced ColorPalette gaps for more compact layout
- Sorted the DEFAULT_PALETTE_COLORS array to follow the visual order of swatches
- Moved color palette component to desired position in the UI

## Result
Default: 
Screenshot 2025-12-09 at 5 22 15 PM

With the configuration option **ShowPalette** enabled:
Screenshot 2025-12-10 at 2 44 00 PM


packages/boxel-ui/addon/src/components/color-palette/index.gts
```diff
@@ -49,6 +54,12 @@ export default class ColorPalette extends Component<Signature> {
 
   <template>
     <div class='color-palette-group' ...attributes>
+      <ColorPicker
+        @color={{@color}}
+        @onChange={{@onChange}}
+        @placeholder='Custom hex color (#ff00ff)'
+        @disabled={{@disabled}}
+      />
       {{#unless @disabled}}
```
lucaslyl, line 58: Chris requested swapping the positions so the ColorPicker appears at the top of the color palette.

richardhjtan (approved):

burieberry (commented):
The rearranged color order in the palette and the uneven rows look disorienting. I asked about it in design channel. Otherwise looks good

lucaslyl:
> The rearranged color order in the palette and the uneven rows look disorienting. I asked about it in design channel. Otherwise looks good

Good advice. I agree with your point. I’ll switch the color reorder code and color palette gap back to the previous revision.

Since Chris mentioned that the palette should be customizable through field configuration, this will align better with that direction. We also have an upcoming PR that covers the advanced color field with a full set of configuration variants and  options - including the palette config:
https://linear.app/cardstack/issue/ECO-447/color-field-advanced-version

I’ll proceed with this PR first so we can minimize merge conflicts in the upcoming PR.😄

PR #3686, opened 2025-12-09
Cs 9824 remix button pushed off the tile

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-9824/remix-button-pushed-off-the-tile

## Fix:

1. Fix Remix button being cropped
2. fix ListingFittedSkeleton css bug
3. Fix button loading state alignment

## Result:

https://github.com/user-attachments/assets/f3a045a4-2f61-48b2-8159-d00db70c7e1f


packages/catalog-realm/catalog-app/components/showcase-view.gts
```diff
@@ -127,11 +124,11 @@ export default class ShowcaseView extends GlimmerComponent<ShowcaseViewArgs> {
       }
       .showcase-cards-display :deep(.cards.grid-view),
       .featured-cards-display :deep(.cards.grid-view) {
-        --grid-view-height: 380px;
+        --grid-view-height: 400px;
         grid-template-columns: repeat(2, 1fr);
       }
       .new-this-week-cards-display :deep(.cards.grid-view) {
-        --grid-view-height: 380px;
+        --grid-view-height: 400px;
```
backspace: I see that this was an existing thing, but given that this is repeated in three places, is it worth extracting to something higher-level (`:root`?)?

backspace (approved):

packages/catalog-realm/catalog-app/components/showcase-view.gts
lucaslyl: Nice catch! You're right — I totally overlooked this CSS after changing the height value.But, Since --default-grid-view-height is already 400px, setting --grid-view-height: 400px is unnecessary. The fallback covers it, so I can remove those declarations.

PR #3687, opened 2025-12-09
[DRAFT] Improve the host test performance

IanCal:
Several things seem to be really hitting the host test performance locally:

1. Loading modules (particularly card-api which can cost ~500ms per simple test)
2. Creating test realms
3. Creating matrix things

Here I've added two major things:

1. Snapshotting state (database, matrix, loader) after creating realms
2. Not reloading the base realm items again and again

The snapshots are purely in memory, so there is no caching to manage between test runs. The concept is that the first test run creates the realm/etc in the beforeeach, and this is then snapshotted and restored for all following tests that use the same beforeeach.

### DIFF
--- a/.github/workflows/ci-host.yaml
+++ b/.github/workflows/ci-host.yaml
@@ -31,8 +31,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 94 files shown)


habdelra:
Abandoned PR

PR #3689, opened 2025-12-10
Fix style host mode content

FadhlanR:


backspace:
I tried it via [this preview deployment link](https://fix-host-mode-content-style.boxel-host-preview.stack.cards/?hostModeOrigin=https://minicatalog.staging.boxel.build) and it worked

backspace (approved):
I’m surprised to learn we’ve never been removing the loading overlay element, I think that’s the correct approach, but for now this is a good fix.

PR #3690, opened 2025-12-10
Show fail indicator if correctness check report error

FadhlanR:
In the message command, the indicator shows whether the command was successfully applied or not. However, for the Check Correctness command, this behavior doesn’t quite fit. Even when the command is successfully applied, it may still report an error.

Because of this, we need different handling for the Check Correctness command. Instead of relying only on whether the command was applied successfully, we need to check whether it reports an error. If it does, we should display the failure indicator, even though the command itself was successfully applied.


jurgenwerk:
image

packages/host/app/components/matrix/room-message-command.gts
```diff
@@ -134,6 +134,22 @@ export default class RoomMessageCommand extends Component<Signature> {
     );
   }
 
+  private get correctnessFailure() {
```
jurgenwerk: perhaps a clearer name would be `didFailCorrectnessCheck`

FadhlanR:
> image

@jurgenwerk Thanks for the feedback! I introduced a new state named 'applied-with-error' and it also has the warning icon as you suggested.

Screenshot 2025-12-15 at 14 22 16

jurgenwerk (approved):

PR #3691, opened 2025-12-10
Remove ember-moment

ef4:
The previous version had a dep on an old `@embroider/addon-shim` that won't work with `@embroider/core` 4.x.


backspace:
Is it even needed? I recall adding it for file modification dates but that looks to have been replaced with a resource

ef4:
Good catch, changed to attempt removal instead...

lukemelia (approved):

PR #3699, opened 2025-12-11
Fix ContainsMany, LinksTo and LinksToMany edit component field alignments

FadhlanR:
**ContainsMany no items**
Before:

ContainsMany-before-no-items

After:

ContainsMany-after-no-item

**ContainsMany with items**
Before:

ContainsMany-before-with-items

After

ContainsMany-after-with-item

**LinksTo/LinksToMany no items**
Before:

LinksToMany-before-no-items

After:

LinksToMany-after-no-item

**LinksTo/LinksToMany with items**

Before:

LinksToMany-before-with-item

After:

LinksToMany-after-with-item

### DIFF
--- a/packages/base/contains-many-component.gts
+++ b/packages/base/contains-many-component.gts
@@ -153,7 +153,7 @@ class ContainsManyEditor extends GlimmerComponent<ContainsManyEditorSignature> {
     </PermissionsConsumer>
     <style scoped>
       .contains-many-editor {
-        --remove-icon-size: var(--boxel-icon-sm);
+        --remove-icon-size: var(--boxel-icon-med);
       }
       .contains-many-editor :deep(.compound-field.edit-format .add-new) {
         border: 1px solid var(--border, var(--boxel-border-color));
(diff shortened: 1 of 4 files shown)


burieberry (commented):
This definitely looks much better than before. Something to fix is the outline when you're tabbing over the elements. The updated outline looks a little elongated, let's make sure the outline remains square and there's a 1 or 2px offset from the svg.

Before:
outline-before

After:
outline-after

You should probably change the size of the button from `tall` to a different size. Here are the available sizes: https://boxel-ui.stack.cards/?s=Components&ss=%3CIconButton%3E

jurgenwerk (approved):

richardhjtan (approved):

FadhlanR:
> This definitely looks much better than before. Something to fix is the outline when you're tabbing over the elements. The updated outline looks a little elongated, let's make sure the outline remains square and there's a 1 or 2px offset from the svg.
> 
> Before: outline-before
> 
> After: outline-after
> 
> You should probably change the size of the button from `tall` to a different size. Here are the available sizes: https://boxel-ui.stack.cards/?s=Components&ss=%3CIconButton%3E

@burieberry Thanks for catching it! I've updated it, I'd appreciate if you could check it again.

PR #3709, opened 2025-12-13
Updated skill and skill-set documentation templates

burieberry:
- Add SkillPlus card to catalog with updated documentation template
- Add updated SkillSet template
- Updated embedded template for CommandField
- Remove skill and skill-set jsons that have been moved to boxel-skills repo

Skill Set:
skill-set

Skill Plus:
skill


lukemelia (commented):
Why aren't we using the boxel-skills repo for skills?

burieberry:
> Why aren't we using the boxel-skills repo for skills?

I replaced the ones in catalog, but I agree, I can put the skill jsons there and keep the card definitions in catalog

lukemelia (approved):

habdelra (approved):

PR #3710, opened 2025-12-15
Implement locking per room

FadhlanR:
In this PR we switch the AI bot’s locking from event-scoped to room-scoped so only one responder runs per room at a time, avoiding duplicate replies. Previously we locked on each Matrix event ID, which let near-simultaneous events in the same room race and double-respond. Now we acquire/release locks keyed by room ID, update the responder flow accordingly, add a migration making `ai_bot_event_processing` primary-keyed on `room_id` with cleanup intact, and refresh the locking tests to cover the new behavior.

### DIFF
--- a/packages/ai-bot/lib/queries.ts
+++ b/packages/ai-bot/lib/queries.ts
@@ -12,7 +12,7 @@ export async function acquireRoomLock(
   pgAdapter: PgAdapter,
   roomId: string,
   aiBotInstanceId: string,
-  eventId?: string,
+  eventId: string,
 ): Promise<boolean> {
   // Attempts to take an exclusive lock per room by upserting a row. The insert succeeds when no
   // unfinished processing exists for the room; otherwise an UPDATE runs only if the previous run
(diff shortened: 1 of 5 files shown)


packages/ai-bot/lib/queries.ts
```diff
@@ -23,16 +25,21 @@ export async function acquireLock(
...
     ...addExplicitParens(separatedByCommas(valueExpressions)),
-    `ON CONFLICT (event_id_being_processed) DO NOTHING`,
-    `RETURNING ai_bot_instance_id, event_id_being_processed`,
+    `ON CONFLICT (room_id) DO UPDATE SET`,
+    `ai_bot_instance_id = EXCLUDED.ai_bot_instance_id,`,
+    `event_id_being_processed = EXCLUDED.event_id_being_processed,`,
+    `processing_started_at = EXCLUDED.processing_started_at,`,
+    `completed_at = NULL`,
+    `WHERE ai_bot_event_processing.completed_at IS NOT NULL`,
+    `RETURNING ai_bot_instance_id, room_id, event_id_being_processed`,
```
lukemelia, line 38: Can we add a comment describing what this SQL is doing?

lukemelia (approved):

PR #3711, opened 2025-12-15
Fix correctness checks not arriving after patching (when there are cancelled tool calls in room history)

jurgenwerk:
Fixes this bug:

image


jurgenwerk:
@codex review

lukemelia (approved):

PR #3713, opened 2025-12-15
Skip failing test

lukemelia:


jurgenwerk:
Thanks, I'll take a look

jurgenwerk:
@copilot can you investigate why this test did not fail in https://github.com/cardstack/boxel/pull/3660? Can you attempt to make a PR with a fix

PR #3718, opened 2025-12-16
Fix correctness checks not arriving after patching code when there are some cancelled code patches in room history

jurgenwerk:
SImilar to https://github.com/cardstack/boxel/pull/3711 but for code patches


jurgenwerk:
Hmm, it looks like this is not needed, I can't reproduce it anymore in main.

PR #3720, opened 2025-12-16
Audio field tests, don't use external URLs

IanCal:
example.com is rightly leaving us with long pending requests which fails a lot in tests.


lukemelia (approved):

habdelra (approved):

richardhjtan (approved):
Thanks for addressing this @IanCal , I will visit this after to find possible ways to improve

PR #3724, opened 2025-12-16
Explicit null for the system card means no card set

IanCal:
This lets us turn off system cards entirely, and allows a quick way of quieting the tests.


habdelra:
Abandoned PR

PR #3738, opened 2025-12-18
Introduce index card

FadhlanR:
Previously we have the `site-config` card where user can select a card instance as alternate home page in host mode. The selected `site-config` card is stored in `hostHome` property of `.realm.json` and is retrieved by the `home-page-resolver` service to redirect user to the alternate home page if it exists. In other words, in that approach the site-config card is only the data source, while the routing happens in the host. We also wanted to provide capability where user can have an alternate home page in interact submode (alternate index card). This requires implementing another config card and improving the routing in the host.

In this PR, we implemented a new approach by introducing index card. Instead of having two different config cards for alternate home page in host mode and interact submode, with index card we can store those selected cards in linksTo fields. And also the routing happens directly in this card; we just need to add new properties: mode and submode in `cardContext`. Based on that, the index card can determine which card to be displayed to the user.

https://github.com/user-attachments/assets/c7f80dfe-c35f-4432-bd53-ac1c39b33d39


FadhlanR:
@codex

FadhlanR:
@codex review this PR

FadhlanR:
> ### 💡 Codex Review
> https://github.com/cardstack/boxel/blob/60fc95e1907c0f1300797d58a6c11b0ab92d125c/packages/host/app/routes/index.gts#L70-L74
> 
> 
> **P1 Badge Host-mode ignores existing hostHome configuration**
> In host mode the index route now simply fetches the requested path and renders that card (lines 70-74) and the former `home-page-resolver` lookup was removed in this commit, so `.realm.json` `hostHome` values are no longer consulted. Existing realms that previously set an alternate landing page via `setHostHome` will now serve the default index/card grid instead until someone manually edits the new `IndexCard` to include the same link, which breaks previously configured host home pages after deploy.
> 
> ℹ️ About Codex in GitHub

This PR tends to not backwards compatibality.

> ### 💡 Codex Review
> https://github.com/cardstack/boxel/blob/60fc95e1907c0f1300797d58a6c11b0ab92d125c/packages/host/app/routes/index.gts#L70-L74
> 
> 
> **P1 Badge Host-mode ignores existing hostHome configuration**
> In host mode the index route now simply fetches the requested path and renders that card (lines 70-74) and the former `home-page-resolver` lookup was removed in this commit, so `.realm.json` `hostHome` values are no longer consulted. Existing realms that previously set an alternate landing page via `setHostHome` will now serve the default index/card grid instead until someone manually edits the new `IndexCard` to include the same link, which breaks previously configured host home pages after deploy.
> 
> ℹ️ About Codex in GitHub

Backward compatibility is not a priority here and legacy realms will simply revert to the standard card grid.

backspace (approved):
This seems like a great improvement, I was uneasy with the front-end being responsible for this work, pushing it into the index card itself is an elegant solution 🤩

lukemelia (approved):

PR #3739, opened 2025-12-18
Limit the amount of loops of failed correctness checks to 3 attempts

jurgenwerk:
This PR introduces a safety precaution where we don't want the correctness check loop to spiral out of control. So for example if AI generates code and our correctness checks report an error related to targeted instance or module, and this loop continues for more than 3 times, we bail and report to AI that it should step sending new patches to try to fix the errors. 

This is obviously hard to test on a real case, so I relied heavily on adding tests for this.

### DIFF
--- a/packages/runtime-common/ai/prompt.ts
+++ b/packages/runtime-common/ai/prompt.ts
@@ -946,8 +946,10 @@ function describeCheckCorrectnessTarget(request?: CommandRequest) {
   if (!request) {
     return 'the requested target';
   }
-  let attributes = ((request.arguments as Record<string, any>) ?? {})
-    .attributes as Record<string, any> | undefined ?? {};
+  let attributes =
+    (((request.arguments as Record<string, any>) ?? {}).attributes as
+      | Record<string, any>
+      | undefined) ?? {};
   let targetType = attributes.targetType;
   let targetRef = attributes.targetRef;
   if (targetType && targetRef) {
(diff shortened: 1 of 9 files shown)


packages/runtime-common/ai/prompt.ts
```diff
@@ -992,6 +995,8 @@ const SEARCH_REPLACE_FIX_INSTRUCTION = `1. Propose fixes for the above errors by
 const CORRECTNESS_SUCCESS_SUMMARY_INSTRUCTION =
   'Summarize the correctness results above in one short sentence confirming that the target is now fixed. Mention any warnings if they exist.';
 
+const CORRECTNESS_FAILURE_LIMIT_INSTRUCTION = `Automated correctness fixes have already been attempted ${MAX_CORRECTNESS_FIX_ATTEMPTS} times and the target is still failing correctness checks. Stop proposing further automated patches; instead, summarize the remaining errors and ask the user how they want to proceed.`;
```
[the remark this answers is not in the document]
jurgenwerk: There is no "step sending" in that string

packages/runtime-common/ai/prompt.ts
```diff
@@ -1067,14 +1072,197 @@ function formatCorrectnessResultSummary(
...
+    } else if (attributes.fileUrl || args.fileUrl) {
+      targetType = 'file';
+    }
+  }
+  let targetEventId =
+    attributes.targetEventId ??
+    args.targetEventId ??
+    attributes.target_event_id ??
+    args.target_event_id ??
+    undefined;
```
lukemelia: Why do we have to support so many permutations?

jurgenwerk: It was for backwards compatibility reasons but it turns out we don't really need it. I made the adjustments in next commits and simplified `CheckCorrectnessInput` by getting rid of `cardId` and `fileUrl`, and only keep the `targetType` and `targetRef`.

lukemelia (approved):

PR #3740, opened 2025-12-18
Card header action icons shouldn’t be red when card is in an error state

FadhlanR:
Before:

Screenshot 2025-12-18 at 17 23 29

After:

Screenshot 2025-12-18 at 17 20 18

### DIFF
--- a/packages/host/app/components/operator-mode/card-error.gts
+++ b/packages/host/app/components/operator-mode/card-error.gts
@@ -114,6 +114,9 @@ export default class CardErrorComponent extends Component<Signature> {
         background-color: var(--boxel-100);
         box-shadow: 0 1px 0 0 rgba(0 0 0 / 15%);
       }
+      .error-header :deep(.actions) {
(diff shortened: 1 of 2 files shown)


jurgenwerk (approved):
Perhaps adding a test that tests if css is working correctly is not really needed? It feels like it's testing how css works internally, maybe a percy shot would be more fitting for this test

packages/host/app/components/operator-mode/card-error.gts
```diff
@@ -114,6 +114,9 @@ export default class CardErrorComponent extends Component<Signature> {
         background-color: var(--boxel-100);
         box-shadow: 0 1px 0 0 rgba(0 0 0 / 15%);
       }
+      .error-header :deep(.actions) {
+        color: var(--boxel-dark);
+      }
```
burieberry: We can set the color `var(--boxel-error-300)` on only the title (`.card-type-display-name`) and not on the whole header.

PR #3742, opened 2025-12-18
Fix: room locking migration

FadhlanR:
This PR updates the migration file to delete all data from ai_bot_event_processing, so it won't break the case where there is old data in that table. Thanks @IanCal for catching this issue!

### DIFF
--- a/packages/postgres/migrations/1765785229000_room-scoped-ai-bot-locks.js
+++ b/packages/postgres/migrations/1765785229000_room-scoped-ai-bot-locks.js
@@ -1,18 +1,9 @@
 exports.shorthands = undefined;
 
 exports.up = (pgm) => {
-  // Existing rows may contain Matrix event ids (e.g. $abc:domain) which cannot be
-  // converted to room ids (e.g. !room:domain) without additional context. Force a
-  // manual cleanup/backfill so we do not silently persist invalid room ids.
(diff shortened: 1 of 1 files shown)


richardhjtan (approved):

packages/postgres/migrations/1765785229000_room-scoped-ai-bot-locks.js
```diff
@@ -1,18 +1,9 @@
...
-    BEGIN
-      IF EXISTS (SELECT 1 FROM ai_bot_event_processing) THEN
-        RAISE EXCEPTION 'ai_bot_event_processing contains existing rows; truncate or backfill with actual Matrix room ids (e.g. !room:domain) before rerunning migration 1765785229000_room-scoped-ai-bot-locks';
-      END IF;
-    END;
-    $$;
-  `);
+  // Backward compatibility is not required here; clear
+  // any existing data so we can safely introduce room-scoped locks.
+  pgm.sql(`DELETE FROM ai_bot_event_processing;`);
```
richardhjtan, line 6: We encountered the error mentioned in this PR this morning, and tried this SQL, which also achieves the same result. I guess DELETE do the same
```
TRUNCATE TABLE ai_bot_event_processing;
```

PR #3747, opened 2025-12-18
Cleaning up indexing test around mending broken instances using module restored after deletion

habdelra:
This PR restores a test that lost cohesion in a recent change. specifically we want this specific test to prove that an instance that points to a deleted module can be mended if the module is restored. The test had a bunch of stuff added to it that made it impossible to prove this because an incremental index was being manually triggered. As well as a lot of assertions that really had nothing to do with this test and were asserting below this abstraction layer were being made. The additional assertions were redundant with other tests that we already have and added unnecessary noise.

### DIFF
--- a/packages/realm-server/tests/indexing-test.ts
+++ b/packages/realm-server/tests/indexing-test.ts
@@ -1221,48 +1221,16 @@ module(basename(__filename), function () {
               return this.author?.firstName + '-poo';
             }
           })
+          static embedded = class Embedded extends Component<typeof this> {
+            <template><@fields.firstName/> (<@fields.nickName/>)</template>
+          }
+          static fitted = class Fitted extends Component<typeof this> {
+            <template><@fields.firstName/> (<@fields.nickName/>)</template>
+          }
         }
       `,
       );
       {
-        assert.true(
-          await adapter.exists('post.gts'),
-          'post module file exists on disk',
-        );
-        realm.__testOnlyClearCaches();
-        await realm.realmIndexUpdater.update([new URL(`${testRealm}post.gts`)]);
-        let moduleResponse = await realm.handle(
-          new Request(`${testRealm}post`, {
-            headers: { Accept: 'application/javascript' },
-          }),
-        );
-        assert.strictEqual(
-          moduleResponse?.status,
-          200,
-          `module response status ${moduleResponse?.status}`,
-        );
-        assert.ok(
-          realm.realmIndexUpdater.stats.modulesIndexed >= 1,
-          `modulesIndexed=${realm.realmIndexUpdater.stats.modulesIndexed}`,
-        );
-        let [postIndexEntry] = (await testDbAdapter.execute(
-          `SELECT url, is_deleted, type FROM boxel_index WHERE url = '${testRealm}post.gts'`,
-        )) as { url: string; is_deleted: boolean; type: string }[];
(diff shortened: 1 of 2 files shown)


packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,48 +1221,16 @@ module(basename(__filename), function () {
...
         }
       `,
       );
       {
-        assert.true(
-          await adapter.exists('post.gts'),
-          'post module file exists on disk',
-        );
-        realm.__testOnlyClearCaches();
-        await realm.realmIndexUpdater.update([new URL(`${testRealm}post.gts`)]);
```
habdelra, line 1233: ugh--this manually triggers the incremental index what we want to see happen automatically as part of the realm.write

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,48 +1221,16 @@ module(basename(__filename), function () {
...
-        let [postIndexEntry] = (await testDbAdapter.execute(
-          `SELECT url, is_deleted, type FROM boxel_index WHERE url = '${testRealm}post.gts'`,
-        )) as { url: string; is_deleted: boolean; type: string }[];
-        assert.ok(postIndexEntry, 'post module row exists in index');
-        assert.false(postIndexEntry?.is_deleted);
-        assert.strictEqual(
-          postIndexEntry?.type,
-          'module',
-          JSON.stringify(postIndexEntry, null, 2),
-        );
```
habdelra, line 1257: this seems super unnecessary here. we have lower level unit tests that already cover this.

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,48 +1221,16 @@ module(basename(__filename), function () {
...
-        assert.strictEqual(
-          moduleResponse?.status,
-          200,
-          `module response status ${moduleResponse?.status}`,
-        );
-        assert.ok(
-          realm.realmIndexUpdater.stats.modulesIndexed >= 1,
-          `modulesIndexed=${realm.realmIndexUpdater.stats.modulesIndexed}`,
-        );
```
habdelra, line 1247: we have other tests that already prove that module can be added and realm endpoint tests that prove you can HTTP get modules

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,48 +1221,16 @@ module(basename(__filename), function () {
...
-          JSON.stringify(postIndexEntry, null, 2),
-        );
-        let postModule = await realm.realmIndexQueryEngine.module(
-          new URL(`${testRealm}post`),
-        );
-        assert.strictEqual(
-          postModule?.type,
-          'module',
-          'post module is in the index after recreation',
-        );
```
habdelra, line 1265: we have other tests that prove you can add a module to the index

packages/realm-server/tests/indexing-test.ts
```diff
@@ -1221,48 +1221,16 @@ module(basename(__filename), function () {
...
+          }
         }
       `,
       );
       {
-        assert.true(
-          await adapter.exists('post.gts'),
-          'post module file exists on disk',
-        );
-        realm.__testOnlyClearCaches();
```
habdelra, line 1232: this is only necessary if you write directly into the RealmAdapter such that you bypass the indexing

lukemelia (approved):

PR #3750, opened 2025-12-19
Fix: AI Assistant model selection is lost on refresh or submode switch

FadhlanR:
This PR removes the default LLM model requirement for submodes, eliminating the active LLM check during submode transitions. Consequently, this fixes a bug where the user's manual selection failed to persist due to being overridden by the submode's default settings.


FadhlanR:
@codex review

habdelra (approved):

PR #3751, opened 2025-12-19
Include check for error documents in publishability check

FadhlanR:
The `_publishability` endpoint now also looks at `boxel_index` error entries (type = 'error' with error_doc) and marks the realm as not publishable when any exist.

The publishability response now returns:

- type: "realm-publishability",
- a unified violations list with kind: "private-dependency" or kind: "error-document", and 
- warningTypes: ["has-private-dependencies", "has-error-card-documents"] as appropriate.

The host realm service and publish modal are updated to understand the new shape and to show a separate warning block for error-card documents. Added a realm-server test that forces an error index entry and asserts `publishable === false`, an error-document violation, and warningTypes containing "has-error-card-documents".

Screenshot 2025-12-22 at 21 12 44


lukemelia (approved):
OK to merge once green, but we will need follow-up to apply some design polish.

PR #3759, opened 2025-12-22
Fixed issue were store was not being reset for schema changes

habdelra:
This fixes a bug where the store is not resetting based on code changes, e.g. using the schema editor to add a field to a card. This fixes a bug where when we use a schema editor to add a field to a card we were not seeing it update in the playground edit format. now we are.

https://github.com/user-attachments/assets/024c5d62-49b0-4abd-a647-a053dff7d6bf

### DIFF
--- a/packages/host/tests/acceptance/code-submode/card-playground-test.gts
+++ b/packages/host/tests/acceptance/code-submode/card-playground-test.gts
@@ -13,9 +13,13 @@ import { module, test } from 'qunit';
 
 import {
   baseRealm,
+  hasExecutableExtension,
   trimJsonExtension,
   type Realm,
 } from '@cardstack/runtime-common';
+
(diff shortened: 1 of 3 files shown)


packages/host/tests/acceptance/code-submode/card-playground-test.gts
```diff
@@ -1081,6 +1085,68 @@ module('Acceptance | code-submode | card playground', function (_hooks) {
       assert.dom('[data-test-byline]').containsText('Jane Doe');
     });
 
+    test('playground edit format refreshes even if code invalidations are ignored', async function (assert) {
```
habdelra, line 1088: because of the way our indexer runs in the host app it's a little tricky testing this without getting a false positive. that's why we disable the mechanism that would otherwise clobber this test so that we can see it really does work now.

lukemelia (approved):

PR #3774, opened 2025-12-26
Add component stability cache to getContainsManyComponent()

habdelra:
This PR adds a component stability cache to the `getContainsManyComponent()` function so that we can type in a containsMany field without losing focus on each keystroke. We actually have the a component stability cache for LinksToManyComponents, but are missing this for ContainsManyComponents.

### DIFF
--- a/packages/host/tests/acceptance/interact-submode-test.gts
+++ b/packages/host/tests/acceptance/interact-submode-test.gts
@@ -3,6 +3,7 @@ import {
   click,
   fillIn,
   find,
+  typeIn,
   triggerKeyEvent,
   settled,
 } from '@ember/test-helpers';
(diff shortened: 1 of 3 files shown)


packages/host/tests/acceptance/interact-submode-test.gts
```diff
@@ -980,5 +981,183 @@ module('Acceptance | interact submode tests', function (hooks) {
...
+      inputElement = find(inputSelector) as HTMLInputElement;
+      assert.strictEqual(
+        document.activeElement,
+        inputElement,
+        'focus is preserved on the input element after typing',
+      );
+      assert.dom(inputSelector).hasValue(typedText);
+    });
+
+    test('containsMany field def preserves focus while typing', async function (assert) {
```
habdelra, line 1059: chris really wanted to make sure this pattern works too, so I added a test for it as well

lukemelia (approved):

PR #3775, opened 2025-12-29
Fix: Copy search card result to workspaces

FadhlanR:
This PR fixes a bug where the "Copy to workspace" menu item in the search card results fails to function. This occurs because the copy action requires write permissions, yet the menu item was being displayed for all users. This fix adds a permission check and ensures the correct realm URL is retrieved from the `OperatorModeStateService`.

https://github.com/user-attachments/assets/56bfccac-b1fa-4484-bdff-cc6d8d2ce63c

### DIFF
--- a/packages/host/app/components/matrix/room-message-command.gts
+++ b/packages/host/app/components/matrix/room-message-command.gts
@@ -159,21 +159,24 @@ export default class RoomMessageCommand extends Component<Signature> {
   private get moreOptionsMenuItems() {
     let menuItems =
       this.commandResultCard.card?.[getCardMenuItems]?.({
-        canEdit: this.canCopyToWorkspace,
+        canEdit: false,
         cardCrudFunctions: {},
         menuContext: 'ai-assistant',
+        menuContextParams: {
+          activeRealmURL: this.activeRealmURL,
+          canEditActiveRealm: this.canEditActiveRealm,
+        },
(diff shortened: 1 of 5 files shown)


packages/host/app/components/matrix/room-message-command.gts
```diff
@@ -157,14 +159,24 @@ export default class RoomMessageCommand extends Component<Signature> {
   private get moreOptionsMenuItems() {
     let menuItems =
       this.commandResultCard.card?.[getCardMenuItems]?.({
-        canEdit: false,
+        canEdit: this.canCopyToWorkspace,
```
lukemelia: This is weird because it is unclear what canEdit means in this context. The user cannot edit the card that the menu item is appearing on, so from that perspective, canEdit being false is correct.

FadhlanR: I see, I misunderstood the the intent of this argument. Should we introduce a new canEditCurrentRealm argument instead?

lukemelia: It's unclear what that would mean in many of the possible menuContexts. Perhaps we need a new menuContextParams, that would be typed differently for different menuContext types. The type for ai-assistant could include a canEditActiveRealm boolean.

FadhlanR:
@codex review

lukemelia (approved):

PR #3778, opened 2025-12-29
Add catalog feature flag

tintinthong:
when USE_EXTERNAL_CATALOG=true, we pull from /contents that is git clone from https://github.com/cardstack/boxel-catalog


richardhjtan (approved):
It works for me

habdelra:
please document this as well

PR #3794, opened 2026-01-05
Implement node script to grant daily credits to users

FadhlanR:
In this PR, I added a Node script that enqueues a daily credit‑grant job for users whose balance is below the minimum threshold. The job is pushed to the PG queue and picked up by the worker to find eligible users and their balances. I also updated the spending logic to consume daily credits before extra credits.


packages/runtime-common/tasks/daily-credit-grant.ts
```diff
@@ -0,0 +1,76 @@
...
+        user_id: row.user_id,
+        credit_amount: creditAmount,
+        credit_type: 'daily_credit',
+        subscription_cycle_id: null,
+      });
+      await query(
+        dbAdapter,
+        insert('credits_ledger', nameExpressions, valueExpressions),
+      );
+      grantedCount++;
```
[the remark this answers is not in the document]
habdelra: yeah good point--using a for loop for executing sql queries is definitely a code smell. this is way more efficient to do in SQL itself

packages/realm-server/daily-credit-grant.Dockerfile
habdelra, line 1: do you have infra PR to go along with this dockerfile? that will need to be deployed before this can be merged

packages/realm-server/daily-credit-grant.ts
```diff
@@ -0,0 +1,64 @@
...
+import './setup-logger'; // This should be first
+import { logger } from '@cardstack/runtime-common';
+import { PgAdapter, PgQueuePublisher } from '@cardstack/postgres';
+import yargs from 'yargs';
+import * as Sentry from '@sentry/node';
+
+const log = logger('daily-credit-grant');
+const DAILY_CREDIT_GRANT_JOB_TIMEOUT_SEC = 10 * 60;
+
+let { priority = 0, migrateDB } = yargs(process.argv.slice(2))
```
habdelra: we discussed having a priority between 0 and 10 for this. probably 4 or 5 is a good place to start. I don't think it should be configurable

packages/realm-server/daily-credit-grant.ts
```diff
@@ -0,0 +1,64 @@
...
+  .usage('Enqueue daily credit grant job')
+  .options({
+    priority: {
+      description: 'The priority of the daily credit grant job (defaults to 0)',
+      type: 'number',
+    },
+    migrateDB: {
```
habdelra: why is this an option? i think we only want to migrate the DB in a single place so we can manage it, which is part of our deployment.

packages/runtime-common/tasks/daily-credit-grant.ts
```diff
@@ -0,0 +1,76 @@
+import type * as JSONTypes from 'json-typescript';
+import type { Task } from './index';
+import { asExpressions, insert, jobIdentity, query } from '../index';
+
+export interface DailyCreditGrantArgs extends JSONTypes.Object {
+  lowCreditThreshold: number;
+}
+
+export { dailyCreditGrant };
```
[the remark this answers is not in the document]
habdelra: this is fine, module scoping means that this is statically loaded so order doens't matter

packages/realm-server/daily-credit-grant.Dockerfile
habdelra: based on discussion, it might actually be a better idea just use the worker manager instead, which means we would not need this dockerfile nor the deploy for that

.github/workflows/manual-deploy.yml
```diff
@@ -111,6 +111,17 @@ jobs:
...
+    name: Build daily credit grant Docker image
+    uses: cardstack/gh-actions/.github/workflows/docker-ecr.yml@main
+    secrets: inherit
+    with:
+      repository: "boxel-daily-credit-grant-${{ inputs.environment }}"
+      environment: ${{ inputs.environment }}
+      dockerfile: "packages/realm-server/daily-credit-grant.Dockerfile"
+      build-args: |
+        "daily_credit_grant_script=start:daily-credit-grant"
+
```
habdelra: if we use worker manager docker, then we don't need this

FadhlanR: Yes, I'll remove this and the dockerFile. And I'll make this PR focusing on implementing the task for daily boxel credit.

packages/runtime-common/tasks/daily-credit-grant.ts
```diff
@@ -0,0 +1,126 @@
...
+    reportStatus(jobInfo, 'start');
+
+    let { lowCreditThreshold } = args;
+    if (!Number.isInteger(lowCreditThreshold) || lowCreditThreshold < 0) {
+      throw new Error(
+        `daily-credit-grant requires a non-negative integer lowCreditThreshold (received ${lowCreditThreshold}).`,
+      );
+    }
+
+    // PERF: This scans all users + aggregates full ledger each run; consider a
```
jurgenwerk, line 39: When this becomes problematic we could also try fetching users and their balances in batches, for example 1000 users in a batch

jurgenwerk (approved):
Looks good to me. Perhaps as a separate PR - I think we want to add some messaging to users so they know they'll be getting free credits soon in case they have a low balance. Perhaps a simple note would suffice, this is a quick sketch:

Image

Not sure if we have a design for this, maybe we can discuss in office hours. Another area that needs a better messaging to include this new feature is here:

https://github.com/cardstack/boxel/blob/468bcaebd27eee1d1f8bc65a2f34ac4c955d8400/packages/ai-bot/main.ts#L428-L434

and here

https://github.com/cardstack/boxel/blob/468bcaebd27eee1d1f8bc65a2f34ac4c955d8400/packages/host/app/components/ai-assistant/message/index.gts#L285-L294

jurgenwerk:
i tested this locally by running `scripts/daily-credit-grant.ts` script and I am able too see `daily_credit` ledger entries, so can confirm enqueing and executing this task works 👍

packages/runtime-common/tasks/daily-credit-grant.ts
```diff
@@ -0,0 +1,126 @@
...
+    // materialized balance table/view (or denormalized user balance) to avoid
+    // re-summing credits_ledger and to make the daily eligibility check cheaper.
+    let rows = (await query(dbAdapter, [
+      `SELECT users.id as user_id,
+        COALESCE(SUM(credits_ledger.credit_amount), 0) as credit_balance,
```
IanCal: Would something like 

```
lowCreditThreshold - COALESCE(SUM(credits_ledger.credit_amount), 0)  as credit_amount
...
having credit_amount > 0
```

Give us the filtered list exactly as we want to insert later? Could we do this in one single transaction within the DB?

FadhlanR: I think it's gonna work, I'll try it. I'm not sure if we need to have one single transaction here, since we only do the insert query once.

jurgenwerk:
From our call, regarding the account info, 2 states:
1. if you are below the threshold, show when the next daily grant will be
2. if you're over the threshold, show when the last grant happened

PR #3795, opened 2026-01-05
"Add workspace" dialog should generate display name correctly

jurgenwerk:
- Generate title-cased display names for new workspaces so that it is more human readable
- Keep endpoint slugging

Here we can see that the display name is now more friendly to read (previously it was slugged): 
image

### DIFF
--- a/packages/host/app/lib/random-name.ts
+++ b/packages/host/app/lib/random-name.ts
@@ -13,9 +13,14 @@ export function generateRandomWorkspaceName(): string {
     ? [adjectives, animals, numberDictionary]
     : [adjectives, animals];
 
-  return uniqueNamesGenerator({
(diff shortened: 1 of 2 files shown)


packages/host/app/lib/random-name.ts
```diff
@@ -13,9 +13,14 @@ export function generateRandomWorkspaceName(): string {
     ? [adjectives, animals, numberDictionary]
     : [adjectives, animals];
 
-  return uniqueNamesGenerator({
+  let slug = uniqueNamesGenerator({
     dictionaries,
     separator: '-',
     style: 'lowerCase',
```
FadhlanR, line 19: Can we instead update these args: `seperator: ' '` and `style: 'capital'`?

habdelra (approved):

PR #3799, opened 2026-01-06
get card dependecies api support return for modules and card url

lucaslyl:


lukemelia:
Ae you sure we need this? You can pass a url for either an instance or a module to this endpoint now if I read the code correctly.

lucaslyl:
> Ae you sure we need this? You can pass a url for either an instance or a module to this endpoint now if I read the code correctly.

Yeah, I agree. I think it makes sense to pause this task for now until we get a bit more clarity on the flow. We can revisit it once things are clearer.

PR #3806, opened 2026-01-07
Additionally run lint checks in file correctness checks 

jurgenwerk:
This adds an additional check in correctness check - it runs the eslint linter and if there are linter errors, it tries to correct them. Here is an example:

image


backspace (approved):
that example is amusing to me 😆

PR #3807, opened 2026-01-07
Federate realm server /_prerendered-search endpoint

habdelra:
This PR creates a federated realm server /_prerendered-search endpoint to consolidate multiple realm prerendered search requests that we were previously initiating from the client.

https://github.com/user-attachments/assets/9a55a9ae-c3a5-41bb-8b82-09e4c3aa9a67


packages/host/app/commands/search-cards.ts
```diff
@@ -81,7 +81,6 @@ export class SearchCardsByQueryCommand extends HostBaseCommand<
       instances = await this.store.search(input.query, realmUrls);
     } catch (e) {
       console.error(`Error searching in realms:`, e, input.query);
-      instances = [];
     }
```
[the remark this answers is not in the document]
habdelra: i think you're high. `instances` _is_ initialized on line 79

packages/host/app/components/prerendered-card-search.gts
```diff
@@ -136,6 +137,18 @@ const normalizeRealms = (realms: string[]) => {
   });
 };
 
+function resolveCardRealmUrl(cardId: string, realms: string[]): string {
+  let cardUrl = new URL(cardId);
+  for (let realm of realms) {
+    let realmUrl = new URL(realm);
+    let realmPaths = new RealmPaths(realmUrl);
+    if (realmPaths.inRealm(cardUrl)) {
+      return realmPaths.url;
```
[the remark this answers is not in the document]
habdelra: this not a real scenario. a Realm cannot overlap with another realm

packages/runtime-common/prerendered-html-format.ts
```diff
@@ -0,0 +1,17 @@
+export const PRERENDERED_HTML_FORMATS = [
+  'embedded',
+  'fitted',
+  'atom',
+  'head',
+] as const;
```
backspace, line 6: this is a side question because it relates to my current work of the server embedding prerendered isolated format cards in `index.html`, why isn’t isolated a valid format?

packages/runtime-common/realm.ts
```diff
@@ -2966,28 +2968,68 @@ export class Realm {
     });
   }
 
-  private async searchPrerendered(
+  public async searchPrerendered(
+    cardsQuery: Query,
+    opts: {
+      htmlFormat: 'embedded' | 'fitted' | 'atom' | 'head';
```
backspace: Could `PrerenderedHtmlFormat` type be used here?

packages/runtime-common/realm.ts
```diff
@@ -2998,7 +3040,7 @@ export class Realm {
               status: '400',
               title: 'Bad Request',
               message:
-                "Must include a 'prerenderedHtmlFormat' parameter with a value of 'embedded' or 'atom' to use this endpoint",
+                "Must include a 'prerenderedHtmlFormat' parameter with a value of 'embedded', 'fitted', 'atom', or 'head' to use this endpoint",
```
backspace: Would it make sense to use the format list constant here? So it doesn’t have to be manually kept up to date if/when another format is added

backspace (approved):
I ran this locally and saw the expected requests and responses 🎉

packages/runtime-common/prerendered-html-format.ts
habdelra: it can be, its just that until now we didn't need it yet. it's totally fine to add this as a new prerender format

packages/runtime-common/realm.ts
habdelra: we could just reword it to something like "Must include a valid 'prerenderedHtmlFormat' parameter"

PR #3811, opened 2026-01-08
Cs 9803 migrate field configuration showcases

lucaslyl:
Migrate field specs from field-showcase to individual subclass specs

This is the **first pass** to migrate fields from field-showcase to each subclass spec. 

## This change:
- Creates spec for subclass override with isolated and edit templates
- Adds reusable **_FieldSpecIsolatedTemplate and FieldSpecEditTemplate_** components for subclass specs
- Removes field-showcase dependency
- Updates all field spec subclasses to use the new shared templates

**FieldSpecIsolatedTemplate /  FieldSpecEditTemplate**  include the Field Configuration Playground (dynamic fields) and hide the example section, which differs from the Base Spec template.te and  FieldSpecEditTemplate include the Field Configuration Playground (dynamic fields) and hide the example section, which differs from the Base Spec template.

## Demo:

https://github.com/user-attachments/assets/02c067be-5b9f-4d22-abf8-70d8358bed97

## Follow-up task:
https://linear.app/cardstack/issue/CS-9974/base-spec-isolatededit-and-subclass-spec-isolatededit-templates-have
In the catalog, we override the Subclass Spec isolated/edit template to display the Field Configuration Playground (dynamic fields / hide the example section). However, most of the remaining code is the same as
the Base Spec template.

We discussed whether we should move the shared isolatedTemplate and editTemplate components into the Base Spec (so both base spec and subclass spec can reuse it), or introduce a shared implementation that both Base and Subclass specs can reference. _** Currently all subclass spec is located under catalog-realms_ **


packages/catalog-realm/field-spec/audio-field-spec.gts
```diff
@@ -0,0 +1,71 @@
...
+        showVolume: true,
+        showSpeedControl: true,
+        showLoopControl: true,
+      },
+    },
+  });
+
+  static isolated =
+    FieldSpecIsolatedTemplate as unknown as typeof Spec.isolated;
+  static edit = FieldSpecEditTemplate as unknown as typeof Spec.edit;
```
lucaslyl, line 70: TypeScript can't verify the compatibility directly. I use this  double type assertion (as unknown as typeof Spec.isolated) to tell TypeScript the template is compatible with the expected type

packages/catalog-realm/components/field-renderer.gts
```diff
@@ -1,6 +1,7 @@
 import GlimmerComponent from '@glimmer/component';
```
tintinthong, line 1: why did the types get more complicated in this module??

packages/catalog-realm/field-spec/AudioFieldSpec/aad14b4f-2b08-468f-ab34-18d9890c39c3.json
```diff
@@ -0,0 +1,192 @@
...
+      "adoptsFrom": {
+        "name": "AudioFieldSpec",
+        "module": "../audio-field-spec"
+      }
+    },
+    "type": "card",
+    "attributes": {
+      "ref": {
+        "name": "AudioFieldSpec",
+        "module": "http://localhost:4201/catalog/field-spec/audio-field-spec"
```
tintinthong: why are the refs absolulte? How is this gonna work in live environment

packages/host/tests/integration/commands/create-spec-test.gts
```diff
@@ -331,4 +350,52 @@ export default class TestCommand extends Command {
       'readMe content sourced from proxy mock',
     );
   });
+
```
tintinthong: does this test work? I thought we had to copy over files for host test that depend on catalog now

packages/host/app/commands/create-specs.ts
```diff
@@ -152,6 +153,11 @@ export default class CreateSpecCommand extends HostBaseCommand<
     const title = this.getSpecTitle(declaration, codeRef.name);
     const specType = new SpecTypeGuesser(declaration).type;
 
+    const specClassToUse = await getSpecClassFromDeclaration(
```
tintinthong: lets just call this specKlass as before

packages/catalog-realm/field-spec/AudioFieldSpec/aad14b4f-2b08-468f-ab34-18d9890c39c3.json
lucaslyl: Yes, I think this is a bug in the existing create-spec command. I searched through the spec instances created recently, and they all point to an absolute path. I’ll work on changing this behavior so it uses relative paths instead.

image

packages/host/app/commands/create-specs.ts
lucaslyl: I will to rename this to ResolvedSpecKlass because this function also passes SpecKlass as an argument.

packages/host/tests/integration/commands/create-spec-test.gts
lucaslyl: ya ,it work but sometimes it flaky in the host test,  i wil investigate this

packages/runtime-common/file-serializer.ts
```diff
@@ -68,7 +68,7 @@ export default function serialize({
       definition,
       doc,
       relativeTo,
-      codeRefOpts,
+      codeRefOpts: metaCodeRefOpts,
       customFieldDefinitions,
```
lucaslyl, line 72: @tintinthong 

Here is my research.

When creating a Spec, the ref.module field (a CodeRef in attributes) was stored as an absolute URL (e.g., "http://localhost:4200/catalog-realm/blog-app") instead of a relative URL (e.g., "../blog-app/blog-app").

Root Cause:
In file-serializer.ts, processAttributes used codeRefOpts, which lacks:

- allowRelative: true
- maybeRelativeURL function

Without these, codeRefAdjustments couldn't convert absolute URLs to relative ones during serialization.

Solution:

Changed processAttributes to use metaCodeRefOpts instead of codeRefOpts. metaCodeRefOpts includes:
allowRelative: true — enables relative URL conversion
maybeRelativeURL — converts absolute URLs to relative when realmURL is available

Here is the demo after i made this changes:

https://github.com/user-attachments/assets/3fcf748b-c667-4694-bca3-3ed4237bd930

packages/host/app/commands/create-specs.ts
```diff
@@ -337,3 +345,60 @@ export default class CreateSpecCommand extends HostBaseCommand<
...
+ *
+ * @param codeRef - Code reference of the code being documented (used for naming convention)
+ * @param fallbackSpecClass - Base Spec class to use if no subclass found
+ * @param loader - Module loader for dynamically loading Spec subclasses
+ * @param targetRealm - Realm URL to use as relativeTo for loading the spec class
+ * @returns Spec subclass if found, otherwise fallback base Spec
+ */
+async function getSpecClassFromDeclaration(
+  codeRef: ResolvedCodeRef,
+  fallbackSpecClass: typeof BaseDef,
```
habdelra: wouldnt this be a typeof CardDef--Specs are Cards not fields

packages/catalog-realm/utils/is-own-field.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// When checking ColorFieldSpec:
+// ✅ Returns: 'standard', 'wheel', 'sliderRgb', etc. (directly on ColorFieldSpec)
+// ❌ Does NOT return: 'readMe', 'ref', 'title' (from Spec.prototype - inherited)
+// ❌ Does NOT return: 'id' (from CardDef.prototype - inherited)
+export function isOwnField(card: typeof BaseDef, fieldName: string): boolean {
+  return Object.keys(Object.getOwnPropertyDescriptors(card.prototype)).includes(
+    fieldName,
+  );
+}
```
habdelra: This is a dupe of the schema-editor.ts's isOwnFields() function. instead of duplicating, please create function in runtime-common that can be shared by these two

lucaslyl: Great catch. Thanks for the review and suggestion. I’ve created a shared isOwnField utility in runtime-common

packages/host/app/commands/create-specs.ts
lucaslyl: Agreed with it, the issue has been addressed

packages/catalog-realm/field-spec/AudioFieldSpec/aad14b4f-2b08-468f-ab34-18d9890c39c3.json
```diff
@@ -0,0 +1,192 @@
...
+        "playbackRate": null,
+        "waveformData": "[20,20,20,20,20,25.468943589150488,20,20.839556153871676,20,20.653483247335476,22.689638354503945,20,20,20,20,20,20,20.34625231611922,20.63454208435883,24.64886588202841,20,20,20,20,27.78972745914967,25.06101566378016,26.04226072168363,22.2606157445312,21.17618556528294,25.402236794871865,24.66342729516531,20,20,20,23.13270282269059,21.501124252599276,23.542775253807232,20.42887791061779,30.02859743139635,27.321848188657782,31.693067194560054,29.790110231741878,33.453476689575766,32.71048361202666,30.555399922998777,34.24184987510519,41.93558126323925,34.6566024338964,40.79690685531166,41.20448751605889,46.83560562190585,40.50808809228478,39.63210456917528,37.69942149315344,34.91144364570426,30.356303520049803,23.471311855280465,32.79452268924026,35.87921767873682,42.896503577207184,39.59424818000045,42.4367460347902,38.907690244496784,32.948703374080644,26.78791864475809,20,25.14337238491296,27.311346600902496,26.409369466103545,29.725711205432663,24.07767698513343,22.052080387224112,20,20,20,20,20,20,20,20]"
+      },
+      "albumCover": {
+        "url": "blob:http://localhost:4200/b3bfc6f8-82f9-4c20-901d-8b11baf138c3",
```
tintinthong, line 44: localhost here

tintinthong (approved):

packages/catalog-realm/field-spec/AudioFieldSpec/aad14b4f-2b08-468f-ab34-18d9890c39c3.json
lucaslyl: I noticed that the AudioField in audio field itself saves temporary blob URLs instead of uploading files to cloud storage (Missing cloud upload feature). I think we can create a ticket for having a cloud upload audio feature

https://linear.app/cardstack/issue/ECO-457/audio-field-missing-cloud-upload-saves-temporary-blob-urls-instead-of

PR #3812, opened 2026-01-08
Prevent displaying the input field for computed contains fields

FadhlanR:
The aim of this PR is to clarify that a computed `contains` field cannot be edited. Previously, we displayed a disabled input field, however, this did not clearly indicate that the field was not editable. In this PR, I added logic to prevent `edit` format from being used for computed fields, so the input field is no longer displayed.

Before:

Screenshot 2026-01-08 at 20 30 31

After:

Screenshot 2026-01-08 at 20 29 58


lukemelia (approved):
We should fix the typo that copilot caught

PR #3815, opened 2026-01-08
ui: Remove duplicate scopedCSS call

backspace:
This seems to have no effect but is confusing.


packages/boxel-ui/addon/rollup.config.mjs
backspace, line 17: This one still exists and is sufficient.

habdelra (approved):

PR #3825, opened 2026-01-12
fix issue where prerendering during indexing was permitting the realm writes

habdelra:
This PR fixes an issue where prerendering during indexing was permitting realm writes during a specific window of time.


lukemelia (approved):
Do we need a test?

habdelra:
> Do we need a test?

sure we can add a test. I created this because every time i started the realm server i saw a missing-link.json get modified and i was getting tired of that. we can embody something like that in a test

PR #3827, opened 2026-01-13
Create PR command

richardhjtan:
### What is changing 
- This `Create PR` command contains logic to execute GitHub operation like create branch, create PR via a package call the Octokit.
- New realm server endpoint is introduce `/_github-pr`
- To run it locally, include your github token and run `GITHUB_TOKEN=yourToken` when starting your realm server


packages/base/command.gts
```diff
@@ -338,6 +338,20 @@ export class ListingInstallResult extends CardDef {
   @field selectedCodeRef = contains(CodeRefField);
 }
 
+export class CreateListingPRInput extends CardDef {
+  @field roomId = contains(StringField);
+  @field realm = contains(StringField);
```
tintinthong: Realm Field

tintinthong (approved):

PR #3829, opened 2026-01-13
FileDef in linksTo/linksToMany

lukemelia:
This PR adds support for using `FileDef` with `linksTo` and `linksToMany` relationships by introducing a new `FileMetaResource` type that represents file metadata as a resource. Files can now be linked from cards similar to how cards link to other cards, using the new 'file-meta' resource type.

Changes:

- Extended `linksTo`/`linksToMany` to accept `FileDef` in addition to `CardDef`
- Introduced `FileMetaResource` type and 'file-meta' resource type in JSON-API serialization to represent file metadata
- Updated serialization/deserialization logic to handle file-meta resources in relationships
- Added validation requiring `FileDef` instances to have an id when used in links

image


backspace:
Is tapping the attachments in the `file-links-example` card definition meant to work? They both 404 for me because of `.json` being appended:

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-01-15 16-09-10

I’ll review with the assumption that this doesn’t matter, as it‘s just meant to demonstrate the possibility of linking vs be something used by card authors.

lukemelia:
> Is tapping the attachments in the `file-links-example` card definition meant to work? They both 404 for me because of `.json` being appended:

It works in interact mode but not yet in code mode. I'll make a separate issue for that.

packages/host/tests/acceptance/interact-submode-test.gts
```diff
@@ -510,6 +510,45 @@ module('Acceptance | interact submode tests', function (hooks) {
...
+            },
+          ],
+        ],
+      });
+
+      assert
+        .dom(
+          `[data-test-stack-card="${testRealmURL}FileLinkCard/with-file"] [data-test-card="${fileId}"]`,
+        )
+        .exists('linked file is rendered in the card');
```
backspace, line 530: For me it’s not actually rendered:

Image

nor is an image:

Image

packages/host/tests/integration/components/card-basics-test.gts
```diff
@@ -381,6 +385,472 @@ module('Integration | card-basics', function (hooks) {
...
+      assert
+        .dom('[data-test-plural-view-field="attachments"]')
+        .exists('FileDef links render via the plural view');
+      assert
+        .dom('[data-test-image-def]')
+        .exists(
+          { count: 2 },
+          'FileDef uses delegated fitted view for each item',
+        );
+    });
```
backspace, line 481: On the topic of editing, is it only possible via code submode? I see this when I try to edit in interact submode:

Image

I can’t delete links to attached files either, in interact submode

backspace (approved):
Because I’m not that familiar with the parts of the application where the support is implemented, I’m mostly focused on the UI of the example card, which surely isn’t the important thing. But I ran this locally and it does prove the successful linking of a file and files, so it does work! 🎉

packages/host/tests/acceptance/interact-submode-test.gts
lukemelia: Those are for later, when we implement subclasses of FileDef

packages/host/tests/integration/components/card-basics-test.gts
lukemelia: Editing is out of scope for FileDef. I'll create an issue to make sure it doesn't render in code mode

habdelra (approved):

PR #3830, opened 2026-01-13
Cs 9965 links spec and update spec should use get dependencies to support

lucaslyl:
## Problem:

image (1)

1. Realm.ts had two similar endpoints that caused confusion:
- `/_dependencies` - JSONAPI format, supports instances and modules
- `/_card-dependencies` - JSON array, instances only (renamed in this PR to 
  distinguish these two APIs: `/_dependencies` → `/_card-dependencies`)

2. The `/_card-dependencies` endpoint only supports instances, not modules.

## Changes

1. Renamed one of the `/_dependencies` endpoints → `/_card-dependencies` to 
   distinguish the two APIs.

2. Update listing-create(linkSpecs) and listing-update-specs to use `/_dependencies` instead 
   of `/_card-dependencies` to support both instance and module dependencies.
   - Parse JSONAPI response format (data[].attributes.dependencies)
   - Update Accept header to JSONAPI mime type
   - Extract dependencies from all entries in the JSONAPI data array

This clarifies endpoint usage and enables module dependency support.


packages/runtime-common/realm.ts
```diff
@@ -559,7 +550,7 @@ export class Realm {
         this.publishability.bind(this),
       )
       .get(
-        '/_dependencies',
+        '/_card-dependencies',
         SupportedMimeType.CardDependencies,
```
lucaslyl, line 567: @richardhjtan I will rename this to card-dependencies to better distinguish its API purpose

tintinthong (approved):

PR #3837, opened 2026-01-14
Refactor federated search to use realms list in the body of the request

habdelra:
This PR refactors the federated search endpoints to specify the realms list to search within into the body of the HTTP QUERY requests.

https://github.com/user-attachments/assets/6944ce93-cfa8-43ec-b3a1-a8a2de9f79e2


packages/runtime-common/search-utils.ts
```diff
@@ -88,6 +88,68 @@ export function parseRealmsParam(url: URL): string[] {
...
+    if (realmList.length === 0) {
+      throw new SearchRequestError(
+        'missing-realms',
+        'realms must be supplied in request body',
+      );
+    }
+    return realmList;
+  }
+
+  let realmList = parseRealmsParam(url);
```
lukemelia: Is this backward compatibility actually needed?

habdelra: no. looks like AI paranoia

packages/runtime-common/search-utils.ts
```diff
@@ -88,6 +88,68 @@ export function parseRealmsParam(url: URL): string[] {
...
+      );
+    }
+    let realmsValue = payloadRecord.realms;
+    if (!Array.isArray(realmsValue)) {
+      throw new SearchRequestError(
+        'missing-realms',
+        'realms must be an array of strings',
+      );
+    }
+    if (!realmsValue.every((realm) => typeof realm === 'string')) {
```
lukemelia: Combine these two conditionals since the errors is the same

packages/runtime-common/search-utils.ts
```diff
@@ -88,6 +88,68 @@ export function parseRealmsParam(url: URL): string[] {
...
 
+export async function parseRealmsFromRequest(
+  request: Request,
+  url: URL = new URL(request.url),
+): Promise<string[]> {
+  let method = resolveSearchRequestMethod(request);
+  if (method === 'QUERY') {
+    let payload: unknown;
+    try {
+      payload = await request.clone().json();
```
lukemelia: This seems inefficient. Anything that calls this will need to parse the request payload as JSON anyway, so why not get realms from the already parsed payload?

lukemelia (changes requested):

packages/runtime-common/query.ts
```diff
@@ -18,28 +18,34 @@ export class InvalidQueryError extends Error {
...
     size: number;
     realmVersion?: number;
   };
-  realm?: string;
 }
 
-export interface QueryWithInterpolations {
+export type Query =
+  | (QueryBase & { realm?: string; realms?: never })
+  | (QueryBase & { realms?: string[]; realm?: never });
```
lukemelia, line 33: Do we really need to support both "realm" singular and "realms" plural?

packages/runtime-common/query.ts
```diff
@@ -207,6 +222,24 @@ function assertRealm(realm: any, pointer: string[]): asserts realm is string {
...
+  realms: any,
+  pointer: string[],
+): asserts realms is string[] {
+  if (!Array.isArray(realms)) {
+    throw new InvalidQueryError(
+      `${pointer.join('/') || '/'}: realms must be an array of strings`,
+    );
+  }
+  for (let realm of realms) {
+    if (typeof realm !== 'string') {
```
lukemelia, line 235: Combine conditionals since the error thrown is the same.

packages/runtime-common/search-utils.ts
habdelra: actually it looks like there are still places in our app that are continuing to use HTTP GET for searching. and AI rightfully tried to keep that backward compatibility. I'll update all the call sites to no longer use HTTP GET for searching

packages/runtime-common/query.ts
habdelra: so for the realm based search we supported a `realm` singular. I haven't looked into why, but thats what was currently there. for federated realm-server that is not sufficient. perhaps we can dive more deeply into this.

packages/runtime-common/search-utils.ts
habdelra: fixed in [f8041cd](https://github.com/cardstack/boxel/pull/3837/commits/f8041cd1f678f6310f935da79c9444af2c4d0c97)

packages/runtime-common/search-utils.ts
habdelra: fixed in [8913b83](https://github.com/cardstack/boxel/pull/3837/commits/8913b83a6915ca0aad8c58c37aafb0b44b25706d)

packages/runtime-common/search-utils.ts
habdelra: fixed in [8913b83](https://github.com/cardstack/boxel/pull/3837/commits/8913b83a6915ca0aad8c58c37aafb0b44b25706d)

packages/runtime-common/query.ts
habdelra: fixed in [8913b83](https://github.com/cardstack/boxel/pull/3837/commits/8913b83a6915ca0aad8c58c37aafb0b44b25706d)

packages/runtime-common/query.ts
habdelra: so the query fields use "realm" singular. if we want to standardize on this we'll need to be backwards compatible with existing cards that use "realm" singular.

habdelra: fixed in [e63e4f0](https://github.com/cardstack/boxel/pull/3837/commits/e63e4f0a59da94ec0e116b3d80fc9a8f051e81ef)

habdelra:
@lukemelia can you take another look? I addressed all the issues you raised.

packages/host/tests/integration/components/serialization-test.gts
```diff
@@ -4113,15 +4113,15 @@ module('Integration | serialization', function (hooks) {
       @field title = contains(StringField);
       @field favorite = linksTo(Person, {
         query: {
-          realm: '$thisRealm',
```
lukemelia, line 4116: I may have lead you astray the the single `realm` / multiple `realms` stuff. query fields are intentionally limited to just a single realm and so the public facing API should always be singular. The search APIs are capable of multiple realms and so they should always be plural IMO.

lukemelia (changes requested):
Some more singular/plural issues. LMK if you want to discuss.

packages/host/tests/integration/components/serialization-test.gts
habdelra: ok, i'll revert that commit

habdelra:
@lukemelia ok, i reverted the change that forced the query interface to only have a plural "realms" property. now it is `realm` XOR `realms`.

lukemelia (approved):

PR #3842, opened 2026-01-14
Host app dependency upgrades for vite

ef4:
- ember-exam adds vite support at 10.x


ef4:
This is waiting behind a couple prerequisites that were uncovered.

PR #3849, opened 2026-01-15
disable generate readme and title,description inputs on spec when no write access

tintinthong:
since specs can be in any writeable or non-writeable realm, we need to adapt the UI to adapt to write permissions

when no write access and previewing format = edit 
- generate readme button (hidden)
- title <-disabled (uses boxel input)
- description <- disabled (uses boxel input)

Screenshot 2026-01-15 at 23 25 21

PS: The markdown field is already disabled as per Markdown Field

### DIFF
--- a/packages/base/spec.gts
+++ b/packages/base/spec.gts
@@ -35,7 +35,7 @@ import {
   type CommandContext,
   type ResolvedCodeRef,
 } from '@cardstack/runtime-common';
-import { eq, type MenuItemOptions } from '@cardstack/boxel-ui/helpers';
+import { eq, not, type MenuItemOptions } from '@cardstack/boxel-ui/helpers';
 import { AiBw as AiBwIcon } from '@cardstack/boxel-ui/icons';
 
 import GlimmerComponent from '@glimmer/component';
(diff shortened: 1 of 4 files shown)


burieberry (approved):
So this is necessary because we're overriding the string field template just so that we can pass in a placeholder right? 

Just make sure that the `this.realm` in create-spec command you're checking refers to the realm of the spec instance and not of the current code-mode realm because they can be different realms.

Your changes are not reflected in the preview links, but maybe they're behind. Code looks good to me.

backspace:
> Your changes are not reflected in the preview links, but maybe they're behind.

A sad limitation of the `host` previews is that they only reflect changes made in that application or its build-time dependencies (`boxel-icons`, `boxel-ui`). The base realm change would require a realm server deployment to be served to the previews. I’d love to have more-thorough preview deployments that incorporate more of the stack but it seems prohibitively complex.

tintinthong:
> So this is necessary because we're overriding the string field template just so that we can pass in a placeholder right?

Confirm. It is bcos we overwrite the input with custom templates. But the new component seems to NOT be permission-aware.

> Just make sure that the `this.realm` in create-spec command you're checking refers to the realm of the spec instance and not of the current code-mode realm because they can be different realms.

Confirm, The permission checked is actually for the realm you are writing the spec to

> Your changes are not reflected in the preview links, but maybe they're behind. Code looks good to me.

PR #3850, opened 2026-01-15
Prevent large amounts of data from getting into cards

FadhlanR:
### Purpose

This PR introduces a guard to prevent large amounts of data from being stored in cards.

### Background

Card data can currently be inserted or updated through several pathways:

* Editing card data from the **stack item edit view** in **Interact** submode
* Editing card data in the **code editor** in **Code** submode
* Patching a card via **AI commands**
* Calling the **realm endpoint** directly

Without a guard, these paths can allow large payloads to enter the system.

### Implementation

To address this, I implemented `validateWriteSize` function in both the **host** and the **realm**:

* **Host-level guard**:
  Implemented to fail early for scenarios that originate in the host.

* **Realm-level guard**:
  Added as a second line of defense to ensure protection even when requests reach the realm directly.

### Error Handling by Scenario

* **Interact submode** and **Code submode**:
  Errors are thrown directly and surfaced to the user immediately.

* **AI-based patching**:
  I extended the *check correctness* command to retrieve the card limit error from `cardService`, so we can handle errors originating from:

  * `patch-code` command
  * `patch-card-instance` command

I’m open to alternative approaches here if there’s a better way to handle validation or error propagation for AI-driven updates.

Interact Submode:

Screenshot 2026-01-19 at 15 45 39

Code Submode:

Screenshot 2026-01-19 at 15 44 25


FadhlanR:
@codex review

packages/host/config/environment.js
```diff
@@ -40,6 +40,7 @@ module.exports = function (environment) {
     cardRenderTimeout: Number(
       process.env.RENDER_TIMEOUT_MS ?? DEFAULT_CARD_RENDER_TIMEOUT_MS,
     ),
+    cardSizeLimit: Number(process.env.CARD_SIZE_LIMIT ?? 64 * 1024),
```
lukemelia: Adjust name or add comment indicating units (bytes, I assume?)

lukemelia (approved):
LGTM. I'd like to make sure @habdelra reviews this and weighs in on what a suitable limit value is.

packages/runtime-common/realm.ts
```diff
@@ -359,6 +361,8 @@ interface Options {
   fromScratchIndexPriority?: number;
 }
 
+const DEFAULT_CARD_SIZE_LIMIT = 64 * 1024;
```
habdelra: i think this is 64KB (please add units to this env var name). if so, i think that is a perfectly fine limit.

packages/host/config/environment.js
habdelra: why do we hard code `64 *1024` twice? once here and once in runtime-common. let's just have a single place for this constant

PR #3852, opened 2026-01-15
give write permissions to homepage_editor

burieberry:
This migration is to give boxel-homepage production write permissions to a new user I created


burieberry:
@codex review

PR #3855, opened 2026-01-16
Cs 10066 point the code ref of field spec to field

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10066/point-the-code-ref-of-field-spec-to-field

- Update all field-spec instance JSON files to point ref.name and ref.module to the actual field classes in fields/ directory instead of field-spec files.

Demo:
image


packages/catalog-realm/Spec/01800772-0f10-4079-a517-ea38f3ba43d5.json
```diff
@@ -4,7 +4,7 @@
     "attributes": {
       "readMe": null,
       "ref": {
-        "module": "http://localhost:4201/catalog/fields/image/components/image-presentation",
+        "module": "../fields/image/components/image-presentation",
```
lucaslyl, line 7: didnt delete these component spec because it used in one of the CardListing

packages/catalog-realm/fields/color.gts
```diff
@@ -21,7 +21,7 @@ class EditView extends Component<typeof ColorField> {
   </template>
 }
```
tintinthong, line 23: why having two exports? I think there are consequences to this when for example you auto create-specs? If you dont want the default name I wud say just remove the default export imo

lucaslyl: Thank you for the advice — I agree this can be confusing, as the current catalog-realm fields have evolved with different export patterns over time. Based on Burcu’s earlier guidance to favor default exports for consistency and flexible imports, I did a cleanup in catalog-realm so the related fields now follow the same export default pattern, giving us a single consistent format to follow going forward.

tintinthong (approved):

PR #3861, opened 2026-01-16
Fix invalidation bug where jsonb "null" was being conflated with SQL NULL

habdelra:
Issue: recovered instances (instances that were previously an error, but then repaired) kept showing errors because error_doc was stored as jsonb null (not SQL NULL), so the index still treated them as errored even after dependencies were fixed.

the fix was to store SQL NULL for cleared JSON fields so recovered instances/modules no longer retain error state (expression.ts). Added regression coverage to confirm error_doc is SQL NULL after a recovery path (indexing-test.ts).

### DIFF
--- a/packages/realm-server/tests/indexing-test.ts
+++ b/packages/realm-server/tests/indexing-test.ts
@@ -1377,7 +1377,10 @@ module(basename(__filename), function () {
         `SELECT error_doc IS NULL AS is_sql_null
          FROM boxel_index
          WHERE realm_url = '${testRealm}'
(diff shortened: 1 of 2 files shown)


backspace (approved):

backspace:
hmm the modified test has a failed assertion 😞

habdelra:
> hmm the modified test has a failed assertion 😞

fixed now

jurgenwerk (approved):

lukemelia (approved):

PR #3866, opened 2026-01-19
Introduce MarkdownDef

lukemelia:
- includes an example of usage in MarkdownSkill

- [x] Based against https://github.com/cardstack/boxel/pull/3865 is merged

### DIFF
--- a/packages/base/markdown-file-def.gts
+++ b/packages/base/markdown-file-def.gts
@@ -1,0 +1,136 @@
+import { byteStreamToUint8Array } from '@cardstack/runtime-common';
+import { StringField, contains, field } from './card-api';
+import MarkdownFilePreview from './markdown-file-preview';
+import {
+  FileContentMismatchError,
+  FileDef,
+  type ByteStream,
+  type SerializedFile,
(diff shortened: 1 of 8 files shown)


packages/base/markdown-file-def.gts
```diff
@@ -0,0 +1,136 @@
...
+
+function fileNameWithoutExtension(name: string): string {
+  return name.replace(/\.[^/.]+$/, '');
+}
+
+function normalizeMarkdown(markdown: string): string {
+  return markdown.replace(/\r\n/g, '\n').replace(/\r/g, '\n');
+}
+
+function stripMarkdown(text: string): string {
```
habdelra, line 34: this is interesting--do we also use this function when generating the search doc field value for a card that uses a MarkDownDoc as its field?

lukemelia: Not currently. Interesting idea.

habdelra: yeah i'm wondering if the special markdown text characters might interfere with the querying against a search doc generated from an md file.

PR #3873, opened 2026-01-20
Provide basic menu items for FileDef instances

lukemelia:
- Includes a BREAKING rename of getCardMenuItems symbol to getMenuItems


habdelra:
where are the tests that actually clicks on these new items to assert they do the thing they say they do?

lukemelia:
> where are the tests that actually clicks on these new items to assert they do the thing they say they do?

The only one accessible so far is Copy File URL in Interact because the code mode preview panel is not yet set up for files. And testing that one is tricky. So the win for this PR is that menu items that *don't work* for files are no longer showing for files in Interact mode and there is a place to add default file menu items over time and for FileDef subclasses to expose menu items.

packages/base/file-api.gts
```diff
@@ -148,3 +161,63 @@ export function createFileDef({
...
+        await new SwitchSubmodeCommand(params.commandContext).execute({
+          submode: 'code',
+          codePath: fileDefInstanceId
+            ? new URL(fileDefInstanceId).href
+            : undefined,
+        });
+      },
+      icon: CodeIcon,
+    });
+  }
```
habdelra, line 221: what about testing clicking on these menu items, open in interact mode and open in code mode?

lukemelia: You can't see them yet, because we are not yet rendering code mode preview for files.

PR #3878, opened 2026-01-21
Fix: host mode does not indicate 404s well

FadhlanR:
This PR fixes an issue where the “card not found” error is not displayed because the `#host-loading` element is never removed, causing the error to be covered by the loading spinner. The operator mode and host mode components work correctly because they are rendered above the `#host-loading` element. However, keeping `#host-loading` in the DOM indefinitely is unexpected.

In this PR, I leverage Ember route substates, introduce a global `loading` template, and remove the `#host-loading` element from `index.html`. As a result, the loading spinner only appears during the initial loading of the Boxel app.

### DIFF
--- a/packages/host/app/index.html
+++ b/packages/host/app/index.html
@@ -32,71 +32,9 @@
     <meta data-boxel-head-start />
     <meta data-boxel-head-end />
 
-    <style scoped>
-      #host-loading {
-        background-color: #686283;
-        display: flex;
-        align-items: center;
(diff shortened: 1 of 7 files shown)


habdelra (approved):
Thanks for this PR--this issue was driving me nuts. this loader is also visible in the prerendered output which makes debugging really annoying.

jurgenwerk (approved):

PR #3881, opened 2026-01-21
Create bot registration endpoint

tintinthong:
# Bot registration endpoint

Short summary of the bot registration updates in this PR, focused on the three endpoints:

### `POST /_bot-registration`
- Registers a bot for the authenticated user.
- Once registered, the bot filters out events in rooms it is invited to.
- Requires a realm server JWT.
- `matrixUserId` must match the authenticated user.
- Returns `201` with the created registration 

### `GET /_bot-registrations`
- Lists bot registrations for the authenticated user only. Only gets his/her own bot registrations
- Requires a realm server JWT.

### `DELETE /_bot-registration`
- Unregisters a bot by id for the authenticated user.
- Requires a realm server JWT.
- Returns `204` on success.


tintinthong:
@codex review

packages/bot-runner/main.ts
```diff
@@ -45,6 +46,34 @@ const botPassword = process.env.BOT_RUNNER_PASSWORD || 'password';
...
+      }
+      users.set(row.matrix_user_id, createdAtMs);
+    }
+    return users;
+  }
+
+  let registeredUsers = new Map<string, number>();
+  try {
+    registeredUsers = await getRegisteredUsers();
+  } catch (error) {
```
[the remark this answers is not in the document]
tintinthong: 👍 good catch but I solved the implementation already. I just need to query the db. No cache is needed. an event will be ignored if the registration was not there at the time the event comes in

packages/postgres/migrations/1769200000000_create-bot-registrations.js
```diff
@@ -0,0 +1,22 @@
...
+    unique: ['user_id', 'name'],
+  });
+  pgm.addConstraint('bot_registrations', 'bot_registrations_user_id_fkey', {
+    foreignKeys: {
+      columns: 'user_id',
+      references: 'users(id)',
+      onDelete: 'CASCADE',
+    },
+  });
+};
```
tintinthong, line 7: this is the schema. Im not sure if we want matrix_user_id here but given the users table already has it maybe I don't need redundancy? Or we can just keep this table be simple and only have matrix_user_id without the user_id?

Also, I added name that shud be unique per user so ppl can name their bots registrations -- this is optional

@lukemelia

packages/bot-runner/main.ts
```diff
@@ -45,6 +46,34 @@ const botPassword = process.env.BOT_RUNNER_PASSWORD || 'password';
...
+    log.error('failed to load bot registrations', error);
+    process.exit(1);
+  }
```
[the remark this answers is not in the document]
tintinthong: This is not relevant anymore

packages/postgres/migrations/1769200000000_create-bot-registrations.js
tintinthong: after some thought. im just going to exclude `name` and use `username` as the field expressed in the ticket which wud refer to matrixUserId

packages/realm-server/routes.ts
```diff
@@ -236,6 +241,21 @@ export function createRoutes(args: CreateRoutesArgs) {
...
+  router.get(
+    '/_bot-registrations',
+    jwtMiddleware(args.realmSecretSeed),
+    handleBotRegistrationsRequest(args),
+  );
+  router.delete(
+    '/_bot-registration',
+    jwtMiddleware(args.realmSecretSeed),
+    handleBotUnregistrationRequest(args),
+  );
```
tintinthong, line 258: im handling delete payload inside the body. Although there are some patterns to include it inside the url. Fine with either altho botRegistration might be leaked in logs if url

packages/base/command.gts
```diff
@@ -240,6 +240,19 @@ export class CreateAIAssistantRoomResult extends CardDef {
   @field roomId = contains(StringField);
 }
 
+export class RegisterBotInput extends CardDef {
+  @field username = contains(StringField);
+  @field name = contains(StringField);
```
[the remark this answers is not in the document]
richardhjtan: Seems like it, @tintinthong do we want to clean this up?

richardhjtan (approved):

docs/generic-bot-runner.md
richardhjtan, line 1: @tintinthong perhaps we can add some description about the membership handler + timeline handler

PR #3882, opened 2026-01-21
Script to rename CardDef and CardInfo fields in gts files

FadhlanR:
In this PR, I also included the results of the renaming script.


scripts/rename-card-fields-gts.js
```diff
@@ -0,0 +1,112 @@
...
+  // CardInfo inline objects
+  [/cardInfo:\s*\{\s*title\s*:/g, 'cardInfo: { name:'],
+  [/cardInfo:\s*\{\s*description\s*:/g, 'cardInfo: { summary:'],
+  [/cardInfo:\s*\{\s*thumbnailURL\s*:/g, 'cardInfo: { cardThumbnailURL:'],
+
+  // Templates
+  [/@model\.title\b/g, '@model.cardTitle'],
+  [/@model\.description\b/g, '@model.cardDescription'],
+  [/@model\.thumbnailURL\b/g, '@model.cardThumbnailURL'],
+];
```
backspace: Would this update this template?

Image

I think you may need to make a codemod-like-tool vs using string replacement

backspace:
This card was missed by the original script, any ideas why?

image

scripts/rename-card-fields-gts.js
FadhlanR: Ah I see, I'll use the codemod tool

FadhlanR:
> This card was missed by the original script, any ideas why?
> 
> 
> 
> image
> 
> 

Isn't it correct? We rename cardInfo.description to cardInfo.summary

backspace:
> Isn't it correct? We rename cardInfo.description to cardInfo.summary

wow yes you’re right, sorry about that 😵

FadhlanR:
I updated this to use codemod-tool instead of string replacement. I’ve also run the script in staging against my realms, and it works. Please check it again.

lukemelia (approved):

PR #3892, opened 2026-01-22
Improve handling of CardJson accept header when non-card paths are specified

lukemelia:


packages/runtime-common/error.ts
```diff
@@ -376,6 +376,21 @@ export function methodNotAllowed(
...
+  return responseWithError(
+    new CardError(
+      `request for ${request.url} can't be fulfilled for accept header ${request.headers.get('accept')}`,
+      {
+        status: 415,
+      },
+    ),
+    requestContext,
+  );
+}
```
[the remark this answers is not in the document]
lukemelia: Addressed

habdelra:
I noticed we got rid of the public writable tests--is that because that's an unrealistic scenario?

packages/host/tests/helpers/index.gts
```diff
@@ -819,12 +819,12 @@ async function setupTestRealm({
...
   virtualNetwork.mount(realm.maybeHandle);
-  if (startMatrix) {
-    await mockMatrixUtils.start();
-  }
   await adapter.ready;
   await worker.run();
   await realm.start();
+  if (startMatrix) {
+    await mockMatrixUtils.start();
+  }
```
habdelra, line 827: what was the reason for moving this? was it causing test flakiness?

lukemelia:
>I noticed we got rid of the public writable tests--is that because that's an unrealistic scenario?

The only thing in that module was a test for a non-card URL, and I reorganized to put all the non-card URLs together under one module.

packages/host/tests/helpers/index.gts
lukemelia: I think AI thought it would help. I doubt it matters one way or the other.

PR #3894, opened 2026-01-22
Add FileDef GC coverage

lukemelia:
- FileDef instances already live in `CardStoreWithGarbageCollection`, but GC and reference counting only consider cards.
- Requirement: file-meta instances should participate in GC and reference counting.

This PR creates the following behavior:

- Keep FileDef instances alive when they have active references.
- GC FileDef instances when their references drop to zero.

### DIFF
--- a/packages/host/app/lib/gc-card-store.ts
+++ b/packages/host/app/lib/gc-card-store.ts
@@ -130,7 +130,7 @@ export default class CardStoreWithGarbageCollection implements CardStore {
   #nonTrackedInstances = new Map<string, StoredInstance>();
   #nonTrackedInstanceErrors = new Map<string, CardErrorJSONAPI>();
 
-  #cards = new TrackedMap<string, StoredInstance>();
+  #instances = new TrackedMap<string, StoredInstance>();
   #instanceErrors = new TrackedMap<string, CardErrorJSONAPI>();
   #gcCandidates: Set<LocalId> = new Set();
   #referenceCount: ReferenceCount;
@@ -298,7 +298,7 @@ export default class CardStoreWithGarbageCollection implements CardStore {
   }
 
   reset() {
-    this.#cards.clear();
+    this.#instances.clear();
     this.#clearEphemeralErrors(this.#instanceErrors);
     this.#nonTrackedInstances.clear();
     this.#clearEphemeralErrors(this.#nonTrackedInstanceErrors);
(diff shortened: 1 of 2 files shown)


packages/host/app/lib/gc-card-store.ts
```diff
@@ -367,49 +373,68 @@ export default class CardStoreWithGarbageCollection implements CardStore {
...
-          // brand the instance to make it easier for debugging
-          (instance as unknown as any)[
-            Symbol.for('__instance_detached_from_store')
-          ] = true;
-          this.delete(localId);
-          if (instance.id) {
-            this.delete(instance.id);
+      if (isFileDefInstance(instance)) {
+        let fileId = instance.id;
+        if (!fileId) {
```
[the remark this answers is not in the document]
lukemelia: FileDef instances *always* have IDs. I removed the check.

packages/host/app/lib/gc-card-store.ts
```diff
@@ -127,11 +127,11 @@ export default class CardStoreWithGarbageCollection implements CardStore {
   // importantly these properties are not tracked so that we are able
   // to deserialize an instance without glimmer rendering the inner workings of
   // the deserialization process.
-  #nonTrackedCards = new Map<string, StoredInstance>();
-  #nonTrackedCardErrors = new Map<string, CardErrorJSONAPI>();
+  #nonTrackedInstances = new Map<string, StoredInstance>();
+  #nonTrackedInstanceErrors = new Map<string, CardErrorJSONAPI>();
 
   #cards = new TrackedMap<string, StoredInstance>();
```
habdelra: to be consistent we should probably rename this to `instances`

packages/host/app/lib/gc-card-store.ts
```diff
@@ -367,49 +373,65 @@ export default class CardStoreWithGarbageCollection implements CardStore {
...
+            (instance as unknown as any)[
+              Symbol.for('__instance_detached_from_store')
+            ] = true;
+            this.delete(fileId);
+          } else {
+            this.#gcCandidates.add(fileId);
           }
         } else {
-          this.#gcCandidates.add(localId);
+          this.#gcCandidates.delete(fileId);
```
habdelra: these 2 branches are pretty similar is there the ability for consolidation here?

packages/host/app/lib/gc-card-store.ts
habdelra: i noticed u didn't change this--was that an  oversight?

PR #3898, opened 2026-01-26
Refactor listing create

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-9964/listing-create-should-support-coderef

- Require codeRef/targetrealm parameter for ListingCreateCommand
- Add "Create listing" action to detail panel (requires canWrite permission)
- Support card and field listings via codeRef
- Update AI prompts (autoPatchName/autoPatchSummary) to use codeRef
- Make openCardId optional: when provided, used as example; when omitted, system searches by codeRef using SearchCardsByTypeAndTitleCommand and links the example for the listing
- Update menu items to pass codeRef while keeping openCardId optional

Demo (create listing for field):

https://github.com/user-attachments/assets/18e52b26-dee9-4f5d-ad9e-a840bfa8b5a9


tintinthong (changes requested):
Looks as I expected. There is no test covereage for createListing. Particularly, when does creatateListing in LHS appear when does it not appear (not talking about execution of command)

lucaslyl:
> Looks as I expected. There is no test covereage for createListing. Particularly, when does creatateListing in LHS appear when does it not appear (not talking about execution of command)

Done add the inspector test for the create listing

The "Create listing" action appears when the selected declaration is an exported card or field definition (has an exportName) and the user has write permission to the realm (canWrite). This inspector tests verify: the button appears for exported card and field definitions (with write permission), does not appear for non-exported card definitions, and does not appear when the user lacks write permission. Both conditions must be true for the action to appear.

tintinthong (approved):

PR #3905, opened 2026-01-27
Fix catalog app/gaming hub search to use cardTitle instead of title

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10111/catalog-app-search-gaming-hub-search-is-not-functioning

fix catalog search: use cardTitle instead of title

The catalog app/ gaming hub search filter was using the 'title' field, which doesn't exist on Listing cards—updated to use 'cardTitle', which is the standard field for title searches. Also updated search.md to document this pattern.

### DIFF
--- a/docs/search.md
+++ b/docs/search.md
@@ -90,7 +90,7 @@ let { data: matching } = await indexer.search({
       name: 'Post',
     },
     every: [
-      { eq: { title: 'Card 1' } },
+      { eq: { cardTitle: 'Card 1' } },
       { not: { eq: { 'author.firstName': 'Cardy' } } },
     ],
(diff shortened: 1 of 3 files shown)


tintinthong:
anywhere suffers from this same issue?

lucaslyl:
> anywhere suffers from this same issue?

So far, no. In experiments-realm and catalog-realm, only a few cards are query searched  by cardTitle: catalog-realm: **catalog-app/gaming-hub**, for experiments-realm: **simple-search-card**

the rest search by name or other specific fields.

tintinthong (approved):

PR #3908, opened 2026-01-27
Introduce picker component

FadhlanR:
This new component is a modified version of the multi-select component with a few behavioral differences:

* The trigger component supports a customizable label.
* A search input is available for filtering option items.
* The **Select All** option and selected options always remain at the top and are not affected by search results.
* A divider is displayed between selected and unselected options.

This component will be used for the **realm picker** and **type picker** in the card chooser.

https://github.com/user-attachments/assets/546712e1-8cb8-4ba3-bd47-df57c8a238ee


packages/boxel-ui/addon/src/components/picker/selected-item.gts
```diff
@@ -0,0 +1,188 @@
...
+            type='button'
+            class='picker-selected-item__remove-button'
+            {{on 'click' (fn this.remove @option)}}
+            aria-label='Remove item'
+          >
+            <IconX class='picker-selected-item__icon--remove' />
+          </button>
+        </:iconRight>
+      </Pill>
+    </div>
```
burieberry, line 131: So I think the selected item template html/css should look very similar to that of the linksToMany compact editor: https://github.com/cardstack/boxel/blob/06f550e6a24a208670a3de09f2c5f1279eb070bf/packages/base/links-to-many-component.gts#L378-L390

packages/boxel-ui/addon/src/components/picker/selected-item.gts
```diff
@@ -0,0 +1,188 @@
...
+      .picker-selected-item {
+        all: unset;
+      }
+
+      .picker-selected-item__pill {
+        display: flex;
+        align-items: center;
+        gap: var(--boxel-sp-5xs);
+        padding: 1px 8px 3px 4px;
+        border-radius: 5px;
```
burieberry: --boxel-border-radius-xs or --boxel-border-radius-sm

packages/boxel-ui/addon/src/components/picker/selected-item.gts
```diff
@@ -0,0 +1,188 @@
...
+        border: solid 1px #d9d9d9;
+        background-color: #d9d9d9;
```
burieberry: --boxel-300 seems to be the closest color

burieberry (commented):
Here are a couple of suggestions for improvement:

Functionality: 
- When a selection is made, the default selection ("All" Realms or Types) should no longer be selected as per the design:
refining-selection-facet

Styling: 
- the default state needs to be updated to mimic the design
- The X button on pills looks really small and not aligned in the middle. You can use a similar X button as the linksToMany component pills.
- We need to adjust the colors for hover, selected, selected hover states. Currently they seem to have uneven padding and different coloring:
picker-selected-items-dropdown
Screenshot 2026-01-27 at 11 43 46 AM
Screenshot 2026-01-27 at 11 43 29 AM

FadhlanR:
@burieberry Thanks for the feedback! I made a few updates based on your suggestions:

* Made the `select-all` option required in the picker options.
* Set the `select-all` option as selected by default.
* Removed the “remove” button from the selected `select-all` option.
* Automatically deselect `select-all` when the user selects any other `non–select-all` option.
* Automatically select `select-all` (and deselect other options) when the user selects all `non–select-all` options.
* Deselect all non–select-all options when the user selects `select-all`.
* Followed the use of "remove" button from `linksToManyCompactEditor`.
* Updated some css.

https://github.com/user-attachments/assets/7c39fd2e-9c69-40d0-b852-cd4b1cf15e58

FadhlanR:
The only left is the icon for `select-all` icon, I'm asking for it in the design channel, but we can add it in the next PR.

burieberry:
Now the selection process looks much better. The styles are still off, though. I still see the blue highlights I posted screenshots of, and the paddings and margins are uneven for the dropdown items and the selected pills. Also opening/closing of the dropdown menu should not cause a shift in the page. 

If styling is not going to be handled here, there should be a new ticket made. Here's the zeplin link for the next reviewer for reference: https://app.zeplin.io/project/67eda3cfaa0e10139932a389/screen/67f441fff878385c6cc2771e

FadhlanR:
I’ve made the following styling updates:
- Removed the blue highlights. These persisted because I hadn't disabled the linting rule for require-scoped-style.
- Fixed the dropdown menu so that opening/closing it no longer causes a layout shift.
- Updated the padding and margins to match the provided design specifications.

Screenshot 2026-02-02 at 14 21 51



https://github.com/user-attachments/assets/c36ce194-2a97-46a7-8d50-064da9dbad5e

backspace (approved):

PR #3916, opened 2026-01-28
Only lint gts/ts files for correctness checks

jurgenwerk:
Otherwise, the linter will try to lint files like .json and it will error out with a message about missing config, like:

```
"lintIssues": [
        "line 0:0 No matching configuration found for 9adeec41-7dab-4dca-b3d8-5be42235f5b3.json."
      ]
```


FadhlanR (approved):
It would be good to have a test for this

PR #3917, opened 2026-01-28
Update custom domain publishing approach

backspace:
More requirements have emerged so I’m adapting the implementation of custom domain publishing, which is a workaround until we have full support for dynamic creation of CNAMES and the like.

1. Instead of hardcoding usernames who can publish to overridden custom domains, anyone with write permissions for a realm can publish it.
2. `PUBLISHED_REALM_DOMAIN_OVERRIDES` is an environment variable that lets us add to the hardcoded mapping without requiring a full deployment.
3. The publish modal UI now is now aware of overrides, which necessitated threading `PUBLISHED_REALM_DOMAIN_OVERRIDES` through from the realm server’s `index.html`-serving function. e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-01-29 14-01-35
1. Claiming and overriding the custom subdomain when publishing is no longer needed.

### DIFF
--- a/packages/realm-server/handlers/handle-publish-realm.ts
+++ b/packages/realm-server/handlers/handle-publish-realm.ts
@@ -1,6 +1,5 @@
 import type Koa from 'koa';
 import {
-  ensureTrailingSlash,
   fetchUserPermissions,
   query,
   SupportedMimeType,
@@ -11,6 +10,7 @@ import {
   asExpressions,
   param,
   PUBLISHED_DIRECTORY_NAME,
+  ensureTrailingSlash,
   type DBAdapter,
   type PublishedRealmTable,
   fetchRealmPermissions,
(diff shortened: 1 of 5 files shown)


packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -33,55 +35,50 @@ import { passwordFromSeed } from '@cardstack/runtime-common/matrix-client';
...
 
   let publishedURL: URL;
   try {
     publishedURL = new URL(publishedRealmURL);
   } catch {
-    return publishedRealmURL;
+    return { applied: false, publishedRealmURL };
+  }
+  if (publishedURL.host.toLowerCase() !== overrideDomain.toLowerCase()) {
+    return { applied: false, publishedRealmURL };
```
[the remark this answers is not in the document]
backspace: I’ve deployed this again after addressing this and was able to publish the rewards document

packages/runtime-common/constants.ts
```diff
@@ -86,4 +86,54 @@ export const DEFAULT_PERMISSIONS = Object.freeze([
...
+// Workaround to override published realm URLs to support custom domains. Remove in CS-9061.
+export const PUBLISHED_REALM_DOMAIN_OVERRIDES: Record<string, string> = {
+  // staging
+  'https://realms-staging.stack.cards/buck/load-testing/':
+    'custombuck.stack.cards',
+
+  // production
+  'https://app.boxel.ai/bucktest/20251216/': 'custombuck.boxel.ai',
+  'https://app.boxel.ai/official/cardstack-reward/': 'tealpaper.cardstack.com',
+  'https://app.boxel.ai/official/boxel-whitepaper/': 'whitepaper.boxel.ai',
```
lukemelia: Do we want to mix and match this constant with the new environment variable or just rely on the new env var?

lukemelia (approved):

packages/runtime-common/constants.ts
backspace: hah well I figured they’re both annoying in their own ways but I guess we might as well go all the way toward the more-dynamic approach. I’ll move these into the SSM parameters

PR #3920, opened 2026-01-29
capitalize action item labels via css

lucaslyl:
Fix: set the root css for menu/action item to text-transform: capitalize; 

image

### DIFF
--- a/packages/host/app/components/operator-mode/definition-container/base.gts
+++ b/packages/host/app/components/operator-mode/definition-container/base.gts
@@ -209,6 +209,7 @@ const Active: TemplateOnlyComponent<ActiveSignature> = <template>
       justify-content: flex-start;
       gap: var(--boxel-sp-xxxs);
       align-self: flex-start;
+      text-transform: capitalize;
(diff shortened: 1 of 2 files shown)


packages/host/app/components/operator-mode/definition-container/base.gts
```diff
@@ -209,6 +209,7 @@ const Active: TemplateOnlyComponent<ActiveSignature> = <template>
       justify-content: flex-start;
       gap: var(--boxel-sp-xxxs);
       align-self: flex-start;
+      text-transform: capitalize;
```
richardhjtan, line 212: @lucaslyl is this something from the code mode inspector view?

lucaslyl: yup, this is part of the inspector panel

Image

richardhjtan (approved):

richardhjtan:
The PR looks good, if the percy snapshot difference is intended due to the changes kindly accept them

PR #3924, opened 2026-01-29
server: Remove SSR for `is_deleted IS TRUE`

backspace:
This is another consequence of using the index as a source of truth vs the filesystem.

You can see the test fail [here](https://github.com/cardstack/boxel/actions/runs/21486704726/job/61898425074?pr=3924#step:10:47) before I added the fix.

### DIFF
--- a/packages/realm-server/tests/server-endpoints/index-responses-test.ts
+++ b/packages/realm-server/tests/server-endpoints/index-responses-test.ts
@@ -2,6 +2,7 @@ import { module, test } from 'qunit';
 import { join, basename } from 'path';
 import { systemInitiatedPriority } from '@cardstack/runtime-common';
 import { setupServerEndpointsTest, testRealm2URL } from './helpers';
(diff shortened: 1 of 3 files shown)


packages/realm-server/tests/server-endpoints/index-responses-test.ts
```diff
@@ -237,6 +238,92 @@ module(`server-endpoints/${basename(__filename)}`, function () {
...
+
+            return true;
+          },
+          {
+            timeout: 5000,
+            interval: 200,
+            timeoutMessage:
+              'Timed out waiting for deleted index entries to be tombstoned',
+          },
+        );
```
backspace, line 288: I don’t love polling the database to make sure the deletion has been recorded but is there a better way?

lukemelia (approved):
It's not new in this PR, but host mode serving should not be pulling from boxel_index_working, only boxel_index

PR #3926, opened 2026-01-29
Simplify ElementTracker

ef4:
This is a prerequisite for work on trying to use view transitions for the stack animations, because we want to compose more behavior into the `cardComponentModifier`, and that's easier to do if the ElementTracker's modifier is a function, and the types are more generic.

We also noticed while doing this that the `observers` `Map` was being recreated many times unnecessarily, which is fixed by this PR.


lukemelia (approved):

ef4:
Closing this, because further exploration has made us unsure if this is a good path forward.

PR #3932, opened 2026-01-30
fix remix failed on player-progress

lucaslyl:
Add the missing spec.


tintinthong:
is there a small improvement we can make to the error message?

lucaslyl:
> is there a small improvement we can make to the error message?

Ya I’ll create a separate Linear task to track this improvement independently :: https://linear.app/cardstack/issue/ECO-466/display-a-error-message-ui-whenever-a-remix-run-fails

richardhjtan (approved):
Able to remix now

PR #3933, opened 2026-01-30
Fix: `ember-page-title` errors when head format template contains `<title>`

FadhlanR:
Previously, the default `<title>Boxel</title>` lived outside the injectable head zone in `index.html`, while the realm-server and host mode inject head HTML between `data-boxel-head-start` and `data-boxel-head-end`. Enabling the title in the head template would have added a second title, so it stayed commented out. This PR moves the default title into that injectable zone and re-enables it in the head template so injected head replaces the default instead of duplicating it. For custom head templates that omit a title, the default "Boxel" is added automatically so the document always has exactly one `<title>`.


lukemelia:
@codex please review

lukemelia:
Is there still a conflict between ember-page-title updating the title and dynamically updated "head" formats updating the page title? If so, which one wins?

FadhlanR:
> Is there still a conflict between ember-page-title updating the title and dynamically updated "head" formats updating the page title? If so, which one wins?

The title that is set by `ember-page-title` win because it is executed in the index template after the `updateHeadTemplate` is executed in the index route. Which one do we expect to win?

FadhlanR:
@lukemelia I've added logic so the ember page title no longer overwrites the head format template.

lukemelia (approved):
Rebase and LGTM

PR #3935, opened 2026-01-30
Add endpoint and button to download realm

backspace:
screencast 2026-01-30 23-11-49


packages/host/app/components/operator-mode/code-submode/left-panel-toggle.gts
```diff
@@ -6,12 +6,24 @@ import Component from '@glimmer/component';
 import FileCheck from '@cardstack/boxel-icons/file-check';
 import FolderTree from '@cardstack/boxel-icons/folder-tree';
 
+import { Button as BoxelButton } from '@cardstack/boxel-ui/components';
 import { cn, not } from '@cardstack/boxel-ui/helpers';
+import { Download } from '@cardstack/boxel-ui/icons';
 
 import RealmDropdown from '@cardstack/host/components/realm-dropdown';
+
+// These were inline but caused the template to have spurious Glint errors
```
backspace, line 15: You can see the errors [here](https://github.com/cardstack/boxel/actions/runs/21538452365/job/62068624109#step:15:32). I tried for a while to come up with a better way, including splitting the template, moving the functions outside the class… this was the only way I found that worked.

lukemelia:
@codex review this. in particular, I'm curious if the implementation in this PR requires the zip file for the full realm to be assembled in memory on the realm server. If so, could it be adjusted to stream instead?

lukemelia:
@codex Does the implementation in this PR mean that the entire zip file will be assembled in memory on the realm server? If so, could it be adjusted to stream as it adds/compresses instead?

lukemelia:
The answer to my question:

The server-side implementation already streams the zip — it does not assemble it in memory. Here's the key mechanism in packages/realm-server/handlers/handle-download-realm.ts:                                    
                                                                                                            
  - Line 149: An archiver stream is created: archiver('zip', { zlib: { level: 9 } })                        
  - Line 161: ctxt.respond = false disables Koa's default response handling, giving manual control of the   
  response stream                                                                                           
  - Line 163: archive.pipe(ctxt.res) pipes the archive directly to the HTTP response — compressed chunks are
   written to the socket as they're produced                                                                
  - Line 166: archive.directory(realmPath, false) reads files from disk incrementally and feeds them into   
  the archive stream                                                                                        
                                                                                                            
  So archiver reads files from the filesystem, compresses them, and writes chunks directly to the HTTP      
  response. The full zip is never materialized in a single buffer on the server.                            
                                                                                                            
  However, the client side does buffer the entire response in memory. In                                    
  packages/host/app/components/operator-mode/code-submode/left-panel-toggle.gts:139:                        
                                                                                                            
  let blob = await response.blob();                                                                         
                                                                                                            
  This collects the entire streamed response into a Blob in browser memory before triggering the download   
  via a temporary <a> link. For very large realms this could be a concern in the browser, but it's a common 
  pattern and browser Blob handling is generally efficient. An alternative would be to use the              
  https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream piped to a                                
  https://github.com/nickersgmh/streamsaver.js-style approach, or more simply, to just navigate the browser 
  directly to the download URL (via window.location or an <a> tag with the href set to the endpoint) and let
   the browser handle the download natively — which would avoid holding it all in JS memory. The auth token 
  would need to be handled differently in that case (e.g., via a short-lived query parameter or cookie      
  rather than an Authorization header).

backspace:
@lukemelia do you think the browser memory aspect is worth changing?

jurgenwerk (approved):
Nice!

lukemelia:
> @lukemelia do you think the browser memory aspect is worth changing?

Yes but can be a separate PR.

backspace:
> Yes but can be a separate PR.

👍🏻 I opened 10145 for that.

PR #3936, opened 2026-01-31
Change host mode default response to HTML

backspace:
This should make embeds work on more platforms, which don’t specify the `Accept: text/html` header we were expecting to serve host mode responses.

(#3947 has `server.ts` extractions related to this work.)

When I deployed this to production, Whatsapp:

IMG_3493

[OpenGraph](https://www.opengraph.xyz/url/https%3A%2F%2Fwhitepaper.boxel.ai):

image

[OrcaScan validator](https://orcascan.com/tools/open-graph-validator?url=https%3A%2F%2Fwhitepaper.boxel.ai) warns of missing tags, which has its own issue (10140) but mostly works:

Open Graph Validator 2026-01-30 22-32-52


backspace:
I didn’t mean to request review on this yet!

lukemelia (approved):
This will likely get more complicated when we build routing, but this seems good for now.

PR #3941, opened 2026-02-03
Adjust Card Chooser search bar to include Realm picker

FadhlanR:
https://github.com/user-attachments/assets/8e7ae64f-dbf9-4634-9393-8f376b29b940

### DIFF
--- a/packages/boxel-ui/addon/src/components/picker/index.gts
+++ b/packages/boxel-ui/addon/src/components/picker/index.gts
@@ -51,10 +51,6 @@ export default class Picker extends Component<PickerSignature> {
     scheduleOnce('afterRender', this, this.ensureDefaultSelection);
   }
 
-  get renderInPlace() {
-    return this.args.renderInPlace ?? true;
-  }
-
   private validateSelectAllOption() {
     const hasSelectAll = this.args.options.some(
       (option) => option.type === 'select-all',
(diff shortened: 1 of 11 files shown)


packages/boxel-ui/addon/src/components/picker/index.gts
```diff
@@ -51,10 +52,6 @@ export default class Picker extends Component<PickerSignature> {
     scheduleOnce('afterRender', this, this.ensureDefaultSelection);
   }
 
-  get renderInPlace() {
-    return this.args.renderInPlace ?? true;
-  }
```
backspace, line 56: I don’t know all the places this is used so maybe Codex’s comment is overly imaginative, but it does seem sensible to me to provide a default for this vs leaving it undefined if the consumer doesn}t pass anything in

packages/host/app/components/realm-picker/index.gts
```diff
@@ -0,0 +1,101 @@
...
+      const name = info?.name ?? this.realmDisplayNameFromURL(realmURL);
+      const icon = info?.iconURL ?? undefined;
+      options.push({
+        id: realmURL,
+        icon,
+        name,
+        type: 'option',
+      });
+    }
+    console.log(options);
```
backspace: Did you mean to leave this in?

packages/host/app/components/realm-picker/index.gts
```diff
@@ -0,0 +1,101 @@
...
+        display: flex;
+        align-items: center;
+        gap: var(--boxel-sp-xs);
+        padding: 0 var(--boxel-sp-sm);
+        color: var(--boxel-450);
+        font-size: var(--boxel-font-size-sm);
+      }
+      .realm-picker__loading-label {
+        font-weight: 500;
+      }
```
backspace, line 98: These are probably an LLMism but IMO since we’re using scoped CSS there‘s no reason to use BEM-style class names.

backspace (approved):
I tried this via the preview deployment, looks good. Maybe it’s called for in the design but I do find it a bit awkward that the checked realms disappear immediately upon click:

Image

I don’t know where to find discussion of this but I think the best practice is to keep things stable until the cursor has left the area, because if I changed my mind about including a realm, I now have to scroll up to find it to unclick it

backspace:
I see the stability change, thanks, hopefully I didn’t just dream this up as a best practice 🙃

PR #3942, opened 2026-02-03
Add last_known_good_deps so error state cards have prerendered CSS

backspace:
The staging card that has a Checkly check for scoped CSS has been mysteriously spontaneously breaking. The card prerenders in host mode but not its scoped CSS (the text should be pink):

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-09 11-14-27

boxel-staging static site custombuck stack cards html | Check session 2026-02-09 11-44-53

This is because `boxel_index.deps` is the source for scoped CSS and it’s truncated when the card has an error. This PR adds a new `boxel_index.last_known_good_deps` that preserves a fallback in case a new version of a card has an error but a prior one worked.

It’s an open question what `host` should do when taking over if a card is in an error state. If this PR is merged it would look like this, but it probably makes sense to detect the error state and act differently:
e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-09 11-53-02


packages/host/config/schema/1770139812965_schema.sql
```diff
@@ -1,6 +1,13 @@
 -- This is auto-generated by packages/realm-server/scripts/convert-to-sqlite.ts
 -- Please don't directly modify this file
 
+ CREATE TABLE IF NOT EXISTS bot_registrations (
+   id NOT NULL,
+   username TEXT NOT NULL,
+   created_at NOT NULL,
+   PRIMARY KEY ( id ) 
+);
+
```
backspace, line 10: I’m confused by the presence of this, maybe updating the schema was missed in another PR?

lukemelia, line 10: @tintinthong can you confirm what's up here?

lukemelia (approved):

packages/host/config/schema/1770139812965_schema.sql
tintinthong: I confirm that these additions for `bot_registrations` are correct but it had already been covered in the other schema.sql file. 

`schema/1770616671686_schema.sql`

Currently, inside `main` there are 2 of these files however this file you just changed doesn't include `bot_commands` so I'm leaning towards deleting this one

backspace: boxel 2026-02-10 20-51-43

Mine didn’t include it because it didn’t exist when I submitted this for review. I think we need a better system here.

PR #3944, opened 2026-02-03
Change realm download to not use in-memory blob

backspace:
This changes the endpoint to support either a header or token so the UI can download via a link and not need to contain the entire realm in memory.

screencast 2026-02-03 19-47-31

I also added a Playwright test since this uses the endpoint in a slightly different way.

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode/left-panel-toggle.gts
+++ b/packages/host/app/components/operator-mode/code-submode/left-panel-toggle.gts
@@ -12,7 +12,7 @@ import { Download } from '@cardstack/boxel-ui/icons';
 
 import RealmDropdown from '@cardstack/host/components/realm-dropdown';
 
-// These were inline but caused the template to have spurious Glint errors
+// This was inline but caused the template to have spurious Glint errors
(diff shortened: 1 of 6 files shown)


lukemelia (approved):

packages/host/app/components/operator-mode/code-submode/left-panel-toggle.gts
```diff
@@ -110,42 +105,18 @@ export default class CodeSubmodeLeftPanelToggle extends Component<Signature> {
   private get downloadRealmURL() {
     let downloadURL = new URL('/_download-realm', this.args.realmURL);
     downloadURL.searchParams.set('realm', this.args.realmURL);
+    // Include token for authenticated streaming download (browser navigates directly)
+    let token = this.realm.token(this.args.realmURL);
```
lukemelia, line 117: Another approach could be to attach a signature obtained by hashing the token plus the URL, so that token only authorizes downloading the particular URL you are sending. Might be good in case we need to use this approach for anything else.

backspace: 👍🏻 that’s in bd465d6, thanks

PR #3952, opened 2026-02-05
Throw friendly error when listing install fails

lucaslyl:
linear: https://linear.app/cardstack/issue/ECO-466/display-a-error-message-ui-whenever-a-remix-run-fails

## Result:
Before:
image

After:

https://github.com/user-attachments/assets/4ae09af2-7101-44a6-ada2-459c9dea8791

### DIFF
--- a/packages/host/app/commands/listing-install.ts
+++ b/packages/host/app/commands/listing-install.ts
@@ -155,11 +155,12 @@ export default class ListingInstallCommand extends HostBaseCommand<
     if (!Array.isArray(atomicResults)) {
       let detail = (results as { errors?: Array<{ detail?: string }> })
         .errors?.[0]?.detail;
-      throw new Error(
-        detail
-          ? `Please make sure your listing has all required specs linked. ${detail}`
-          : 'Please make sure your listing has all required specs linked',
-      );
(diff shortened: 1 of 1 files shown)


packages/host/app/commands/listing-install.ts
```diff
@@ -150,7 +150,17 @@ export default class ListingInstallCommand extends HostBaseCommand<
...
-    let atomicResults: AtomicOperationResult[] = results['atomic:results'];
+    let atomicResults: AtomicOperationResult[] | undefined =
+      results['atomic:results'];
+    if (!Array.isArray(atomicResults)) {
+      let detail = (results as { errors?: Array<{ detail?: string }> })
+        .errors?.[0]?.detail;
+      throw new Error(
+        detail
+          ? `Please make sure your listing has all required specs linked. ${detail}`
+          : 'Please make sure your listing has all required specs linked',
```
tintinthong: - check condition filter is of non-existent type 
  - if true, return suggestion to "click on update specs" and/or "make sure all specs are linked" 
  - if false, fallback to detail

tintinthong (changes requested):
added commment

tintinthong (approved):

PR #3965, opened 2026-02-09
Add JSON:API sparse fieldsets support to search endpoint

lukemelia:
## Summary

- Adds `fields` parameter to the `Query` type following the [JSON:API sparse fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets) spec
- Server-side filtering strips attributes from both `card` and `file-meta` resources while preserving `id`, `type`, `links`, and `meta`
- Client-side store skips identity map cache for sparse instances to avoid polluting it with partial data
- First consumer will be `IndexedFileTree` (ChooseFile modal) which only needs `file.id` — requesting `fields: { 'file-meta': [] }` reduces payload ~90%

## Test plan

- [x] New tests: sparse fieldsets with empty fields array returns resources with no attributes (file-meta and card)
- [x] New tests: sparse fieldsets with specific fields returns only those attributes (file-meta and card)
- [x] New tests: invalid fields value returns 400
- [x] New tests: queries without fields return all attributes (backward compat for both types)
- [x] Existing search tests continue to pass

### DIFF
--- a/packages/host/app/resources/search-data.ts
+++ b/packages/host/app/resources/search-data.ts
@@ -1,0 +1,204 @@
+import { registerDestructor } from '@ember/destroyable';
+import type Owner from '@ember/owner';
+import { setOwner } from '@ember/owner';
+import { service } from '@ember/service';
+import { buildWaiter } from '@ember/test-waiters';
+import { tracked, cached } from '@glimmer/tracking';
+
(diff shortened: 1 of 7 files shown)


lukemelia:
Going ahead with this merge.

PR #3967, opened 2026-02-10
Filter out empty messages from LLM prompt construction

IanCal:
Empty messages (with no content and no tool calls) could be sent to the
LLM, which is wasteful and potentially causes issues. This change:

- Skips assistant messages with empty body and no tool calls during
  history construction (matching the existing guard on user messages)
- Adds a safety filter (hasMessageContent) on the final messages array
  to catch any empty messages from any source, while preserving
  assistant messages with tool calls and tool response messages

CS-10116

https://claude.ai/code/session_01JMzG85yR6uQUXVfnUTegXz

### DIFF
--- a/packages/ai-bot/tests/prompt-construction-test.ts
+++ b/packages/ai-bot/tests/prompt-construction-test.ts
@@ -5153,6 +5153,268 @@ new
     );
   });
 
+
+  test('excludes assistant messages with empty body and no tool calls', async () => {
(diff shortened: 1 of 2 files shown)


IanCal:
@claude iterate to fix the linting and test issues. One of the matrix tests is failing because there is an expect for one attached file that should be checking the count is 2.

jurgenwerk (approved):

PR #3970, opened 2026-02-10
Fix cards entering a broken state when reindexing a subdomain based published realm

jurgenwerk:
if you have a published realm with a subdomain (i.e. `http://hi.localhost:4201/`) and then reindex the realm, it will put the containing cards into an error state with this error (I borrowed this screenshot from @backspace, but I was struggling with the same issue locally):

image


backspace (approved):
Thanks for figuring this out!

lukemelia (approved):

PR #3983, opened 2026-02-11
server: Fix filesystem realm path when downloading

backspace:
This builds on #3979 so downloading the special `boxel-homepage` realm works.

With this branch deployed to production, the top download has contents, the bottom is from `main`:

Downloads 2026-02-11 19-47-08


lukemelia (approved):
Is there a test missing here?

backspace:
> Is there a test missing here?

IMO if there were to be a test it would be in the parent branch, but can you think of a way to test it? The `boxel-homepage` realm is given special handling by being passed in as an argument when the realm server starts up. So testing that I think would mean starting a realm server in a different way, seems prohibitive. The existing tests for this still work.

PR #3984, opened 2026-02-11
Remove module indexing

habdelra:
This PR removes module indexing. This PR is meant to be backwards compatible so that if there are still module index entries we can ignore them safely. 

After this PR gets deployed to staging and production, I'll create a followup PR to remove all the module index entries and streamline the queries since we can safely assume those wont exist.


packages/runtime-common/index-writer.ts
```diff
@@ -389,6 +348,11 @@ export class Batch {
...
         };
         break;
+      case 'module-error':
+        // TODO: Remove module rows from the index in a follow-up PR.
+        throw new Error(
+          'module index entries are no longer supported in the search index',
+        );
       default:
         throw new Error(`Unsupported index entry type`);
     }
```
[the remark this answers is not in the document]
habdelra: this is  support backwards compatability with  existing index that still has module errors within it. we will remove this in a subsequent PR after his PR has been pushed to all deployed envs

packages/realm-server/server.ts
```diff
@@ -541,6 +565,36 @@ export class RealmServer {
...
+
+    let moduleExtensions = ['.gts', '.ts', '.js'];
+    for (let extension of moduleExtensions) {
+      if (existsSync(join(realm.dir, `${localPath}${extension}`))) {
+        return true;
+      }
+      if (existsSync(join(realm.dir, localPath, `index${extension}`))) {
+        return true;
+      }
+    }
```
lukemelia: This is up to 6 IO ops on a hot path (serving a card). Should we consider a cache, or consider doing this in the case of a database miss only?

packages/realm-server/tests/indexing-test.ts
```diff
@@ -2104,23 +2364,34 @@ module(basename(__filename), function () {
...
+          );
+          assert.strictEqual(stats.instancesIndexed, 0, 'no instances indexed');
+        } else {
+          assert.true(
+            indexingFailedBeforeStatsWereProduced,
+            `indexing failed before stats were produced: ${JSON.stringify(
+              stats,
+            )}`,
+          );
+        }
```
lukemelia: Seems weird to have a conditional like this in the test. Shouldn't we know what's going to happen?

packages/runtime-common/cache/module-cache-invalidation.ts
```diff
@@ -0,0 +1,86 @@
+import type { LocalPath, RealmPaths } from '../paths';
+import { executableExtensions } from '../index';
+
+const importSpecifierExpression =
+  /(?:import|export)\s+(?:[^'"`]*?\sfrom\s+)?['"]([^'"]+)['"]|import\(\s*['"]([^'"]+)['"]\s*\)/g;
```
lukemelia: Have we always used a regexp for this kind of thing?

lukemelia (approved):

packages/runtime-common/cache/module-cache-invalidation.ts
habdelra: actually no. we do have a module syntax component in runtime common that uses an AST. let me refactor

packages/realm-server/server.ts
habdelra: fixed in https://github.com/cardstack/boxel/pull/3984/commits/730cad34f2b061d896aa4bf5f186b25ed18d2b89

packages/realm-server/tests/indexing-test.ts
habdelra: fixed in https://github.com/cardstack/boxel/pull/3984/commits/730cad34f2b061d896aa4bf5f186b25ed18d2b89

packages/runtime-common/cache/module-cache-invalidation.ts
habdelra: fixed in https://github.com/cardstack/boxel/pull/3984/commits/730cad34f2b061d896aa4bf5f186b25ed18d2b89

lukemelia (approved):

PR #3986, opened 2026-02-12
Implement the new design for card search results

FadhlanR:
https://github.com/user-attachments/assets/d90d094d-45f8-42f2-8a7f-df68954449ff


packages/host/app/components/search-sheet/section-header.gts
```diff
@@ -0,0 +1,109 @@
...
+      <div class='icon'>
+        {{#if @iconURL}}
+          <img src={{@iconURL}} alt='' class='realm-image' />
+        {{else if @icon}}
+          {{#let @icon as |IconComponent|}}
+            <IconComponent class='icon-svg' />
+          {{/let}}
+        {{/if}}
+      </div>
+      <div class='title'>{{@title}}</div>
```
burieberry, line 46: There's a RealmIcon component that you can use here and adjust the spacing between the icon and the label.

https://github.com/cardstack/boxel/blob/cd80b2b73698c2d8187894ecdab4237871d1246d/packages/host/app/components/operator-mode/definition-container/base.gts#L120-L131

burieberry (commented):
Overall looks good. I think the query needs to run on the Recents too? I can implement my changes after your PR is merged.

PR #3987, opened 2026-02-12
Add og:url tag to head template, reindex published realm to regenerate templates after publishing

jurgenwerk:
Reindexing assures the prerendered templates regenerate when we publish a realm (which gets copied from the source realm). We need to regenerate them now that we included `<meta property='og:url' content={{this.args.model.id}} />` so that the new url gets picked up (otherwise we see source realm's URLs in the published realm's template).

This correctness assurance comes at a cost though - the user needs to wait to publish a realm a little longer (depending on the realm complexity). We discussed that with the team in office hours and the conclusion is that we're okay with this tradeoff for now.

### DIFF
--- a/packages/realm-server/handlers/handle-publish-realm.ts
+++ b/packages/realm-server/handlers/handle-publish-realm.ts
@@ -205,7 +205,7 @@ export default function handlePublishRealm({
           validPublishedRealmDomains.length > 0
         ) {
           let isValidDomain = validPublishedRealmDomains.some((domain) =>
-            publishedURL.host.endsWith(domain),
+            publishedURL.hostname.endsWith(domain),
           );
           if (!isValidDomain) {
             await sendResponseForBadRequest(
--- a/packages/realm-server/handlers/handle-publish-realm.ts
+++ b/packages/realm-server/handlers/handle-publish-realm.ts
@@ -394,6 +394,12 @@ export default function handlePublishRealm({
       );
       await realm.start();
 
+      // reindexing is to ensure that prerendered templates that get copied over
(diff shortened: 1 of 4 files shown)


packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -394,6 +396,12 @@ export default function handlePublishRealm({
       );
       await realm.start();
 
+      // reindexing is to ensure that prerendered templates that get copied over
+      // to the published realm get regenerated - we want this so that the
+      // places in the templates that refer to model.id are updated to the new
+      // published realm URL (for example in the og:url meta tag).
+      await realm.fullIndex();
+
```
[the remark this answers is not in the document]
jurgenwerk: Fixed

packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -394,6 +396,12 @@ export default function handlePublishRealm({
       );
       await realm.start();
```
lukemelia, line 397: Does start already do an incremental index?

lukemelia (approved):

packages/realm-server/handlers/handle-publish-realm.ts
jurgenwerk: No, `start()` does not do an incremental index. In this case, `start()` queues a copy-index job that copies the source realm's index entries into the published realm's index at the DB level. It rewrites URLs in structured fields (url, file_alias, types, deps, pristine_doc, search_doc, etc.) but does not re-render templates (we need `fullIndex()` for that)

PR #3989, opened 2026-02-12
Filter file chooser by type for LinksTo/LinksToMany fields (CS-10080)

lukemelia:
## Summary
- When a `LinksTo` or `LinksToMany` field targets a specific `FileDef` subclass (e.g. `ImageDef`), the file chooser now filters the file tree to only show compatible files
- The modal title dynamically reflects the file type (e.g. "Choose Image" instead of "Choose a File")
- Passes the field's `CodeRef` through `chooseFile()` → `ChooseFileModal` → `IndexedFileTree` → `fileTreeFromIndex` resource, which uses it as the search filter type

## Test plan
- [x] Added acceptance test: file chooser filters by type when field is `ImageDef`
- [x] Added acceptance test: file chooser shows all files when field is `FileDef`
- [ ] Manual test: open an edit view with an ImageDef LinksTo field, click add, verify only images appear
- [ ] Manual test: open an edit view with a FileDef LinksTo field, click add, verify all files appear


packages/host/tests/acceptance/interact-submode-test.gts
```diff
@@ -612,6 +612,129 @@ module('Acceptance | interact submode tests', function (hooks) {
...
+      await waitUntil(
+        () =>
+          document.querySelector(
+            '[data-test-links-to-editor="photo"] [data-test-add-new]',
+          ),
+        {
+          timeout: 5000,
+          timeoutMessage: 'add button did not render for ImageDef field',
+        },
+      );
```
backspace: I know we have `waitUntil` everywhere now, but is there no waiter that could be added to make it unnecessary?

backspace (approved):
I tried this locally, much easier to work with now!

packages/host/tests/acceptance/interact-submode-test.gts
lukemelia: Good call. Cleaned this up!

PR #3996, opened 2026-02-13
Implement the new card chooser design in card catalog

FadhlanR:
This PR implements a new design for the card-catalog modal. I reused components from the search-sheet, which already utilizes the new design.

https://github.com/user-attachments/assets/207d80f2-c161-46d7-988e-79a667cac3ed


burieberry (commented):
Other than the chatgpt review suggestions, clicking on one card then the next causes a shift in the card due to the overlay addition. when tabbing from one card to the next, I don't see the same problem. So I recommend fixing the click css to match the focus behavior.

habdelra:
co-pilot also had this comment which it wasn't too sure about. probably worth checking out:

The "Create New" button is wired to call handleCreateCard on every keydown event (regardless of which key was pressed). This makes normal keyboard navigation (arrow keys, tabbing, etc.) trigger the create-selection unexpectedly. Consider removing the keydown handler entirely (native button Enter/Space already fires click), or change it to only act on Enter/Space and prevent double-invocation.

FadhlanR:
@burieberry @habdelra Thanks for the feedbacks. I've made updates based on your feedbacks. Please check it again.


https://github.com/user-attachments/assets/9783ff60-e366-4ecc-8302-e2d67c8628e4

habdelra (approved):

burieberry (approved):

PR #3997, opened 2026-02-13
Realm picker improvement

FadhlanR:
- [x] Auto-focus search field in Picker dropdown so users can type to filter immediately when opened
- [x] Make the default width of the realm picker's dropdown options wider
- [x] Auto-focus search input in Search Sheet so users can start typing their query immediately when opened

https://github.com/user-attachments/assets/e53c2575-4108-4dbc-a581-4f903473d572

### DIFF
--- a/packages/host/app/components/search-sheet/search-bar.gts
+++ b/packages/host/app/components/search-sheet/search-bar.gts
@@ -21,6 +21,10 @@ let elementCallback = modifier(
   },
 );
 
+const autoFocus = modifier((element: HTMLElement) => {
+  element.focus();
+});
+
 interface Signature {
   Element: HTMLElement;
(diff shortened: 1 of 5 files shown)


packages/host/app/components/search-sheet/search-bar.gts
```diff
@@ -21,6 +21,10 @@ let elementCallback = modifier(
   },
 );
 
+const autoFocus = modifier((element: HTMLElement) => {
```
jurgenwerk: This sounds like a general purpose utility, maybe we can add it to boxel-ui so that LLMs have a better chance of reusing it for future use cases

jurgenwerk (approved):

PR #4009, opened 2026-02-17
Preview pane for non-module files in Module Inspector

lukemelia:
## Summary
- Added `FilePreviewPanel` component that renders a FileDef instance with a format toggle bar (isolated/embedded/fitted/atom)
- Added a reactive `fileDefResource` in `module-inspector.gts` that loads the FileDef via `store.get(url, { type: 'file-meta' })` when an incompatible file is selected
- Updated the template decision chain to show the preview (with loading state) instead of the "No tools available" message
- Updated tests in `code-submode-test.ts` to assert `[data-test-file-preview-panel]` instead of `[data-test-file-incompatibility-message]`

https://github.com/user-attachments/assets/60b7b65c-3c8f-4863-a3e6-6e7b1f4c4f6a

## Test plan
- [x] Open a `.png` file in CodeMode → should see image preview with format toggle
- [ ] Open a `.json` non-card file → should see FileDef preview
- [ ] Open a `.gts` module → should see existing Schema/Playground/Spec (unchanged)
- [ ] Open a card instance `.json` → should see existing CardRendererPanel (unchanged)
- [ ] Toggle between formats (isolated/embedded/fitted/atom) → should re-render
- [ ] Verify existing code-submode tests pass


backspace:
The test plan suggests opening a non-card JSON file but this one won’t render, what’s expected?

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-18 08-16-03

backspace (approved):
I tried this in the preview deployment, success

Image

Since we control the rendering (right?) why not add a bit of padding beneath the filename? It looks weird right up against the bottom edge

packages/host/app/components/operator-mode/code-submode/module-inspector.gts
```diff
@@ -158,6 +160,53 @@ export default class ModuleInspector extends Component<ModuleInspectorSignature>
...
+          state.value = result as unknown as FileDef;
+          state.error = undefined;
+        }
+      } catch (e) {
+        state.error = e;
+        state.value = undefined;
+      } finally {
+        state.isLoading = false;
+      }
+    })();
```
habdelra, line 194: similar comment as your other PR, do we need a test waiter here so that we don't need to wait for isLoading to be false (whatever the DOM equivalent of that is) in our tests? otherwise will we have test flakiness around this?

lukemelia: We already have waiter coverage at the roots of asynchrony involved here. I removed the `waitFor`s.

lukemelia:
> The test plan suggests opening a non-card JSON file but this one won’t render, what’s expected?

I'll capture this in a separate issue

PR #4011, opened 2026-02-17
File Inspector shows FileDef type and inheritance for non-module files

lukemelia:
## Summary
- When a non-module file (e.g., `.png`, `.md`, `.css`) is open in CodeMode, the File Inspector (detail panel) now shows a **File Inheritance** view instead of the generic "Details" section
- Displays the file as a "File Instance" with an "Adopts From" chain pointing to its FileDef type (e.g., `PngDef` for `.png`, `MarkdownDef` for `.md`)
- Clicking the FileDef type navigates to its `.gts` definition file
- Also includes earlier commits: preview pane for non-module files, and unification of CardRendererPanel/FilePreviewPanel into PreviewPanel

image

## Test plan
- [x] Open a `.png` file in CodeMode → File Inspector shows "File Inheritance" with "File Instance .PNG" and "Adopts From → PngDef (File Definition .GTS)"
- [x] Open a `.md` file → shows "File Instance .MD" and "Adopts From → MarkdownDef"
- [ ] Open a `.css` file → shows appropriate FileDef inheritance
- [x] Click on the FileDef type container → navigates to the `.gts` definition file
- [x] Open a `.json` card instance → still shows "Card Inheritance" as before
- [x] Open a `.gts` module file → still shows "In This File" and card/field inheritance as before
- [ ] Delete action still works on non-module files from the File Instance panel

Closes CS-10198

### DIFF
--- a/packages/host/app/components/operator-mode/detail-panel.gts
+++ b/packages/host/app/components/operator-mode/detail-panel.gts
@@ -11,6 +11,8 @@ import { use, resource } from 'ember-resources';
 
 import startCase from 'lodash/startCase';
 
+import { TrackedObject } from 'tracked-built-ins';
+
 import {
   LoadingIndicator,
   ContextButton,
@@ -31,6 +33,7 @@ import {
   isFieldDef,
   isFileDef,
   isBaseDef,
+  isCardErrorJSONAPI,
   internalKeyFor,
   type ResolvedCodeRef,
   type CardErrorJSONAPI,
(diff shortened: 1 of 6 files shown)


packages/host/app/components/operator-mode/detail-panel.gts
```diff
@@ -122,6 +127,50 @@ export default class DetailPanel extends Component<Signature> {
...
+        }
+      } catch (e) {
+        state.error = e;
+        state.value = undefined;
+      } finally {
+        state.isLoading = false;
+      }
+    })();
+    return state;
+  });
```
habdelra, line 163: i'm surprised we didn't need test waiters here (and it doesn't look like you are waiting for isLoading to be set to false in the tests). is there flakiness in the resulting test for this because of that?

PR #4013, opened 2026-02-18
Allow opening linked filedef in the stack

jurgenwerk:
This PR adds the ability to click on a FileDef, which opens as a card in the stack.

Here is an example with a `@field featuredImage = linksTo(ImageDef);`:

https://github.com/user-attachments/assets/75c52d7e-f2b7-4119-9ea5-f2660a303184

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes stack item/state semantics and store reference wiring to support `file-meta` reads, which can affect operator mode navigation, persistence, and realm subscription lifecycles.
> 
> **Overview**
> Enables opening `FileDef`/`file-meta` targets in the operator-mode stack (e.g. clicking a `linksTo(FileDef)` now opens the file as a stack item).
> 
> Introduces a `StackItem.type` (`'card' | 'file'`) and threads it through `viewCard`/`openCardInInteractMode`, stack serialization/deserialization, and rendering so file targets always open in `isolated` and cannot be edited.
> 
> Updates `CardResource`/`getCard` and `StoreService.addReference`/`wireUpNewReference` to support typed reads (`card` vs `file-meta`) and fixes realm subscription cleanup when references drop to zero. Adds/updates tests covering linked-file stack opening and reference/subscription balancing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11b49d6358ec83aae044c3598562e2f32bf8ce57. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/host/app/components/operator-mode/operator-mode-overlays.gts
```diff
@@ -353,6 +359,16 @@ export default class OperatorModeOverlays extends Overlays {
...
 
+  private isFileMetaTarget(
+    renderedCard: StackItemRenderedCardForOverlayActions,
+  ): boolean {
+    let cardDefOrId = renderedCard.cardDefOrId;
+    if (typeof cardDefOrId === 'string') {
+      return hasExtension(cardDefOrId) && !cardDefOrId.endsWith('.json');
+    }
+    return isFileDefInstance(cardDefOrId);
+  }
```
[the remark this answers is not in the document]
jurgenwerk: Fixed

packages/host/app/services/store.ts
```diff
@@ -259,7 +260,7 @@ export default class StoreService extends Service implements StoreInterface {
       this.subscribeToRealm(new URL(id));
       // intentionally not awaiting this. we keep track of the promise in
       // this.newReferencePromises
-      this.wireUpNewReference(id);
+      this.wireUpNewReference(id, readType);
```
habdelra, line 263: could we add a test for file def reference counting? there are a bunch of reference count balancing tests in the store-test.gts, could we update this to include file def's in the mix too?

packages/host/app/resources/card-resource.ts
```diff
@@ -50,14 +55,42 @@ export class CardResource extends Resource<Args> {
...
+    return this.readType === 'file-meta' ? 'card' : 'file-meta';
+  }
+
+  private peekForType(type: StoreReadType): unknown {
+    if (!this.#id) {
+      return undefined;
+    }
+    return type === 'file-meta'
+      ? (this.store.peek(this.#id, { type: 'file-meta' }) as unknown)
+      : (this.store.peek(this.#id) as unknown);
```
habdelra, line 68: be aware there is a very subtle nuance around peeking. if the server state of the instance is in error and the local state is not, then the peek will give u the local state. unsure if this effects what you are trying to do. there is a `peekLive()` that will give you the server state regardless of whether or not it is in error.

packages/host/app/resources/card-resource.ts
```diff
@@ -67,7 +100,12 @@ export class CardResource extends Resource<Args> {
     if (!this.#id) {
       return undefined;
     }
-    let maybeError = this.store.peekError(this.#id);
+    if (this.readType === 'file-meta' && this.peekForType('file-meta')) {
+      return undefined;
```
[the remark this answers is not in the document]
habdelra: i think this comment is trying to convey the subtleties of peek returning local state when the server state is in error (as i was mentioning in another comment). consider `peekLive()` if you need to know the current server state of the instance.

packages/host/app/lib/read-type.ts
```diff
@@ -0,0 +1,15 @@
...
+    return false;
+  }
+  return hasExtension(id) && !id.endsWith('.json');
+}
+
+export function inferStoreReadType(
+  id: string | null | undefined,
+): StoreReadType {
+  return isFileMetaId(id) ? 'file-meta' : 'card';
+}
```
lukemelia: This is an AI anti-pattern that I have had to fight . We should be using more explicit type specification, not inferring stuff from URLs

jurgenwerk: Note from our call: 
- consider adding a subprotocol or url annotation/prefix in urls in operator mode state that differentiate between card or file
- default should be card

packages/host/app/components/operator-mode/interact-submode.gts
```diff
@@ -251,6 +258,23 @@ export default class InteractSubmode extends Component {
...
+      typeof cardOrURL === 'object' &&
+      !(cardOrURL instanceof URL)
+    ) {
+      return isFileDefInstance(cardOrURL as CardDef) ? 'file-meta' : 'card';
+    }
+    let fileMetaInstanceOrError =
+      this.store.peek(cardId, { type: 'file-meta' }) ??
+      this.store.peekError(cardId, { type: 'file-meta' });
+    return fileMetaInstanceOrError ? 'file-meta' : 'card';
+  }
```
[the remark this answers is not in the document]
jurgenwerk: Is this still an issue after new commits?

packages/host/app/resources/card-resource.ts
jurgenwerk: `peekLive` doesn't exist anymore, did you mean `getWithoutCache`? But I'm unsure if this is really needed here, I think the stack items already handle showing cases where card is in error in server state. What I added (that is related to store and card resource) is just store read type selection (`card`, or `file-meta`) depending on which type we want to render in the stack item.

packages/host/app/services/store.ts
jurgenwerk: Added!

packages/host/app/lib/read-type.ts
jurgenwerk: This has been refactored in a way where `StackItem` now has an additional argument `type`, which can be `card` or `file`. This gets serialized in the operator mode state in the host app's URL.

lukemelia (approved):

habdelra (approved):

PR #4018, opened 2026-02-18
Track and invalidate instance and FileDef relationship dependencies in the index

habdelra:
This PR introduces the ability for the index to track instance and FileDef dependencies transitively (which includes their respective module dependencies). Additionally now that these dependencies are being tracked they also participate in our invalidation, so that consumers of links that are updated will be invalidated as well. This means that the previous work to update the meta.ts prerender route to capture module deps of relationships is no longer necessary (a test for CSS via relationships was added to prove this). Also I've refactored the index-runner and split it apart into more cohesive modules as it was growing too large.


packages/runtime-common/index-runner/dependency-resolver.ts
```diff
@@ -0,0 +1,886 @@
...
+      let href = queue.shift()!;
+      ordered.push(href);
+      for (let dependent of edges.get(href) ?? []) {
+        let next = (indegree.get(dependent) ?? 0) - 1;
+        indegree.set(dependent, next);
+        if (next === 0) {
+          queue.push(dependent);
+        }
+      }
+      queue.sort((a, b) => (order.get(a) ?? 0) - (order.get(b) ?? 0));
```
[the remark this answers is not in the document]
habdelra: fixed

packages/runtime-common/index-writer.ts
```diff
@@ -724,10 +734,104 @@ export class Batch {
...
+        selectedRows.push(working);
+        continue;
+      }
+      if (production) {
+        selectedRows.push(production);
+      }
+    }
+
+    return selectedRows;
+  }
```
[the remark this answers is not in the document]
habdelra: fixed

habdelra:
> ### 💡 Codex Review
> https://github.com/cardstack/boxel/blob/e444a5ec8b8ac8943210a39b8f01ba7ac0427d59/packages/runtime-common/index-runner.ts#L161-L162
> 
> 
> **P1 Badge Order from-scratch visits using dependency graph**
> Transitive relationship expansion now depends on reading already-indexed rows (`collectTransitiveRelationshipDeps`), so consumers must be visited after their direct relationship targets to capture second-degree deps. Incremental mode does this via `orderInvalidationsByDependencies`, but from-scratch mode still iterates only `sortInvalidations`, so a cold/full rebuild can process a consumer before its target (for example, `a.json` before `z.json`) and persist an incomplete dependency set. That leaves missing transitive invalidations until another unrelated reindex touches the consumer.
> 
> ℹ️ About Codex in GitHub

fixed

packages/runtime-common/index-runner/dependency-resolver.ts
```diff
@@ -0,0 +1,900 @@
...
+      for (let nested of Object.values(value as Record<string, unknown>)) {
+        visit(nested);
+      }
+    };
+
+    visit(searchDoc);
+    return deps;
+  }
+
+  extractSerializedRelationshipDeps(
```
lukemelia, line 233: This doesn't have any special treatment for query fields, which means that it will treat the query results at the time of indexing as dependencies of the card. However, this is only part of the story for query field invalidation. The other part would be new or changed cards not in the result set becoming part of the result set. We don't necessarily need to try to solve those problems now but we should have a POV and try to be internally consistent.

packages/runtime-common/index-runner/dependency-resolver.ts
```diff
@@ -0,0 +1,900 @@
...
+      }
+    }
+
+    if (ordered.length !== hrefs.length) {
+      for (let href of hrefs) {
+        if (!ordered.includes(href)) {
+          ordered.push(href);
+        }
+      }
+    }
```
lukemelia, line 144: Claude sez:

```
  if (ordered.length !== hrefs.length) {
    for (let href of hrefs) {
      if (!ordered.includes(href)) {
        ordered.push(href);
      }
    }
  }
```

`ordered.includes()` is O(n) per call, making this O(n²) in the worst case. If there are cycles involving many nodes, this could be slow. Consider using a Set for O(1) lookups:

```
  if (ordered.length !== hrefs.length) {
    let orderedSet = new Set(ordered);
    for (let href of hrefs) {
      if (!orderedSet.has(href)) {
        ordered.push(href);
      }
    }
  }
```

lukemelia (approved):

packages/runtime-common/index-runner/dependency-resolver.ts
habdelra: good point

packages/runtime-common/index-runner/dependency-resolver.ts
habdelra: fixed in https://github.com/cardstack/boxel/pull/4018/commits/6a7f4ff2a46054c6bbbd1cf2a1799def69e3cfab

packages/runtime-common/index-runner/dependency-resolver.ts
habdelra: I was bouncing ideas off of codex and it suggested this approach for now:
> There is a simple path, and it avoids the full can-of-worms.

> Use a coarse synthetic dep for query fields.

> When indexing a card/file with query-backed `linksTo/linksToMany`, add a synthetic dep token (as a URL-like string), e.g. `https://<realm>/__query-fields__` (or type-scoped tokens).
On any mutation in that realm (or any mutation of relevant types), also invalidate that token.
Query-field consumers get reindexed, recompute current query results, and then recompute concrete deps/CSS from those results.
This gives internal consistency today:

> Query fields are treated as dynamic subscriptions, not just snapshot edges.
CSS remains correct after reindex because concrete result deps are rebuilt.
Tradeoff:

> Over-invalidation (more reindex work), but straightforward and predictable.

habdelra: maybe as a followup PR?

habdelra: Based on discussion we will eliminate any consideration of query fields when determining deps. This is captured in CS-10219

PR #4020, opened 2026-02-18
Support "Inherit" action for FileDefs in code mode File Inspector

lukemelia:
Works the same way as expect does for a CardDef class, as one would expect.


backspace (approved):
I exercised this in the preview deployment, worked as expected!

Image

PR #4024, opened 2026-02-19
Background Job to Run a Command in Headless Chrome

tintinthong:
## Demo

https://www.loom.com/share/2fe9a3e58a7c459ba574b2c0f747a667

## Testing Locally

1. Start `bot-runner` with `pnpm start:development`.
2. Start realm server with `pnpm start:all`.
3. Inside `packages/matrix`, run `pnpm setup-submission-bot`.
4. Open http://localhost:4200/experiments/BotRequestDemo/bot-request-demo.
5. Click **Send Show Card Bot Request** on the demo card.

### User issuing task

Example source is the experiments demo card, but the same flow is used for any `app.boxel.bot-trigger` event.

```mermaid
flowchart TD

    B["Trigger 'Send Show Card Bot Request' from UI"]
    B --> D["CreateShowCardRequestCommand.execute()"]
    D --> E["SendBotTriggerEventCommand.execute()"]
    E --> F["sendEvent('app.boxel.bot-trigger')"]
    F --> G["bot-runner receives timeline event"]
    G --> H["enqueueRunCommandJob() to jobs table"]
```

### Command runner architecture

```mermaid
flowchart TD

    A["bot-runner"]
    A -->|enqueue run-command job with command string| B["jobs table"]
    B --> C["runtime-common worker"]
    C -->|runCommand task| D["remote prerenderer"]
    D -->|POST /run-command| E["prerender manager"]
    E -->|proxy| F["prerender server app"]
    F -->|transitionTo command-runner route| G["headless Chrome tab"]
    G --> H["host command-runner route"]
    H -->|command executes + DOM status| G
    G -->|capture data-prerender + data-command-result| F
    F --> D
    D --> C
    C --> B
```

## Matrix Event Payload (`app.boxel.bot-trigger`)

```ts
const event = {
  type: 'app.boxel.bot-trigger',
  content: {
    type: 'show-card',
    realm: 'http://localhost:4201/experiments/',
    input: {
      cardId: 'http://localhost:4201/experiments/Author/jane-doe',
      format: 'isolated',
    },
  },
};
```

## Bot Trigger Data Structures

```ts
type SendBotTriggerEventInput = {
  roomId: string;
  type: string;
  realm: string;
  input: Record<string, unknown>;
};

type BotTriggerContent = {
  type: string;
  realm: string;
  input: unknown;
};
```

## Submission Bot Commands (DB)

`setup-submission-bot` now writes canonical scoped command specifiers into `bot_commands.command`.

```json
[
  {
    "name": "create-listing-pr",
    "command": "@cardstack/boxel-host/commands/create-listing-pr/default",
    "filter": {
      "type": "matrix-event",
      "event_type": "app.boxel.bot-trigger",
      "content_type": "create-listing-pr"
    }
  },
  {
    "name": "show-card",
    "command": "@cardstack/boxel-host/commands/show-card/default",
    "filter": {
      "type": "matrix-event",
      "event_type": "app.boxel.bot-trigger",
      "content_type": "show-card"
    }
  },
  {
    "name": "patch-card-instance",
    "command": "@cardstack/boxel-host/commands/patch-card-instance/default",
    "filter": {
      "type": "matrix-event",
      "event_type": "app.boxel.bot-trigger",
      "content_type": "patch-card-instance"
    }
  }
]
```

## Run Command Job Payload

`RunCommandArgs.realmURL` is derived from `event.content.realm`.

```json
{
  "realmURL": "http://localhost:4201/experiments/",
  "realmUsername": "@alice:localhost",
  "runAs": "@alice:localhost",
  "command": "@cardstack/boxel-host/commands/show-card/default",
  "commandInput": {
    "cardId": "http://localhost:4201/experiments/Author/jane-doe",
    "format": "isolated"
  }
}
```

### Command Normalization in `run-command` task

- Command stays a `string` across bot-runner, job queue, and prerender request.
- `runtime-common/tasks/run-command.ts` normalizes legacy realm-server URL forms (`/commands/<name>/<export>`) into a realm-local module specifier path when needed.
- String -> `ResolvedCodeRef` conversion happens in the host `command-runner` route when it parses `:command`.

## Host `command-runner` Route

```ts
route: /command-runner/:command/:input/:nonce

type CommandRunnerRouteParams = {
  command: string;
  input: string;
  nonce: string;
};
```

### Example (`command` and `input` as path params)

```ts
const command = '@cardstack/boxel-host/commands/show-card/default';
const input = JSON.stringify({
  cardId: 'http://localhost:4201/experiments/Author/jane-doe',
  format: 'isolated',
});
const nonce = '2';

const url = `http://localhost:4200/command-runner/${encodeURIComponent(command)}/${encodeURIComponent(input)}/${encodeURIComponent(nonce)}`;
```

```txt
http://localhost:4200/command-runner/%40cardstack%2Fboxel-host%2Fcommands%2Fshow-card%2Fdefault/%7B%22cardId%22%3A%22http%3A%2F%2Flocalhost%3A4201%2Fexperiments%2FAuthor%2Fjane-doe%22%2C%22format%22%3A%22isolated%22%7D/2
```


packages/matrix/scripts/setup-submission-bot.ts
```diff
@@ -2,12 +2,35 @@ import { registerRealmUser } from './register-realm-user-using-api';
...
+    commandURL: `${realmServerURL}/commands/create-listing-pr/default`,
+    filter: {
+      type: 'matrix-event',
+      event_type: 'app.boxel.bot-trigger',
+      content_type: 'create-listing-pr',
+    },
+  },
+  {
+    name: 'show-card',
+    commandURL: `${realmServerURL}/boxel-host/commands/show-card/default`,
```
tintinthong: perhaps this should be @boxel-host/commands which is the import specifier but in url form?

packages/bot-runner/lib/timeline-handler.ts
```diff
@@ -90,10 +132,143 @@ export function onTimelineEvent({
...
+  if (!commandUrl) {
+    return undefined;
+  }
+
+  try {
+    let url = new URL(commandUrl);
+    let path = url.pathname;
+
+    // TODO: boxel-host commands are not exposed internally as HTTP URLs; they
+    // are only available via module specifiers, so we map those URLs to code refs.
```
tintinthong: Perhaps, this can be removed if we treat the url as an import specifier

packages/host/app/commands/bot-requests/create-show-card-request.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import SendBotTriggerEventCommand from './send-bot-trigger-event';
+
+import type MatrixService from '../../services/matrix-service';
+import type StoreService from '../../services/store';
+
+export default class CreateShowCardRequestCommand extends HostBaseCommand<
+  typeof BaseCommandModule.CreateShowCardRequestInput
+> {
+  // TODO: remove this command once the temporary card menu item is removed.
```
tintinthong: This comment is not relevant

packages/base/command.gts
```diff
@@ -434,6 +434,7 @@ export class SendBotTriggerEventInput extends CardDef {
   @field roomId = contains(StringField);
   @field type = contains(StringField);
   @field input = contains(JsonField);
+  @field realm = contains(StringField);
```
tintinthong, line 437: It seems that the worker task requires a realm to be known to check for permissions. So we just constantly send a realm from UI

packages/realm-server/prerender/render-runner.ts
```diff
@@ -386,6 +393,184 @@ export class RenderRunner {
...
+          error: error.length > 0 ? error : null,
+          result: result.length > 0 ? result : null,
+        };
+      }, String(this.#nonce));
+
+      let response: RunCommandResponse = {
+        status: payload.status,
+        result: payload.result ?? undefined,
+        error: payload.error ?? undefined,
+      };
```
tintinthong, line 554: I realise payload.result is the serialized json. There is also .value which is the actual card instance. Currently, there is no meaning to this as we are not using the rendering of the command anywhere other than the command-runner route. 

So I'm not sure what is best to return here

packages/host/app/routes/command-runner.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  deactivate() {
+    (globalThis as any).__boxelRenderContext = undefined;
+  }
+
+  model(params: { nonce: string }, transition: Transition): CommandRunnerModel {
+    let model = new CommandRunState(params.nonce);
+    let queryParams = transition?.to?.queryParams ?? {};
+    let command = parseResolvedCodeRef(getQueryParam(queryParams, 'command'));
+    let commandInput = parseJson(getQueryParam(queryParams, 'input'));
```
habdelra: this is not a good pattern: you should add the command and the input as route segments. let's not use query params for our routes. so your route should look like: 
```
this.route('command-runner', { path: '/command-runner/:command/:input/:nonce' });
```

packages/host/app/routes/command-runner.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+    if (commandInput) {
+      model.commandInput = JSON.stringify(commandInput);
+    }
+
+    if (!command) {
+      model.status = 'error';
+      model.error = new Error('Missing or invalid command');
+      return model;
+    }
```
habdelra: this all goes away if you use route segments for the command and command input, they are guaranteed to be there

habdelra:
~~There are tests missing that i'm expecting to see:~~

~~you should have a host test module that tests your command-runner route directly. similar to prerender-html-test.gts, prerender-meta-test.gts, and prerender-module-test.gts. This way you can have a smaller easier place to test just the stuff that runs inside the prerender server instead of having to boot up puppeteer. That way your server tests can focus on the stuff outside of what is happening inside the command-runner route.~~

MY BAD! I just noticed acceptance-test/commands/test-.gts. this is exactly what I wanted to see :+1:

packages/host/tests/acceptance/commands-test.gts
```diff
@@ -466,6 +491,20 @@ module('Acceptance | Commands tests', function (hooks) {
...
+    };
+    let commandParam = encodeURIComponent(JSON.stringify(commandRef));
+    await visit(`/command-runner/1?command=${commandParam}`);
+
+    await waitFor('[data-prerender][data-prerender-status="ready"]');
+    assert
+      .dom('[data-test-command-runner-greeting]')
+      .includesText('Hello from command runner');
+  });
+
```
habdelra: add a test for when the command throws an error and ensure that the DOM is what you would expect for an error

packages/realm-server/tests/prerender-server-test.ts
```diff
@@ -194,6 +221,51 @@ module(basename(__filename), function () {
       assert.ok(res.body.meta?.pool?.pageId, 'has pool.pageId');
     });
 
+    test('it handles run-command request', async function (assert) {
```
habdelra: Add a test for when there is an error generated by the command and assert that the prerenderer is able to capture the error from the DOM correctly and return it in the properly serialized format for a command error

packages/host/app/templates/command-runner.gts
```diff
@@ -0,0 +1,33 @@
...
+    data-prerender-nonce={{@model.nonce}}
+    data-prerender-status={{@model.prerenderStatus}}
+  >
+    {{#if @model.error}}
+      <pre data-prerender-error>{{@model.error.message}}</pre>
+    {{else}}
+      {{#if @model.value}}
+        <CardContainer class='command-runner-result'>
+          <CardRenderer @card={{@model.value}} @format='isolated' />
+        </CardContainer>
```
habdelra, line 24: does a command always return a card?

packages/host/app/routes/command-runner.ts
```diff
@@ -0,0 +1,186 @@
...
+
+import { registerBoxelTransitionTo } from '../utils/register-boxel-transition';
+
+import type CardService from '../services/card-service';
+import type LoaderService from '../services/loader-service';
+import type RealmService from '../services/realm';
+
+class CommandRunState implements CommandInvocation<CardDefConstructor> {
+  @tracked status: CommandInvocation<CardDefConstructor>['status'] = 'pending';
+  @tracked value: CardDef | null = null;
```
habdelra: `this.value` and `this.result` are confusing for me. i think what you really mean is `resultCard`, `resultString`. can we rename this to make more sense for what they actually are. its kind of ambiguous now.

packages/realm-server/tests/prerender-server-test.ts
```diff
@@ -194,6 +221,51 @@ module(basename(__filename), function () {
...
+      );
+      assert.strictEqual(
+        res.body.data.attributes.status,
+        'ready',
+        'command status ready',
+      );
+      assert.notOk(res.body.data.attributes.error, 'no command error');
+      assert.ok(res.body.meta?.timing?.totalMs >= 0, 'has timing');
+      assert.ok(res.body.meta?.pool?.pageId, 'has pool.pageId');
+    });
```
habdelra, line 340: a command can return a card or it can return a string result. you need to make sure that you test the ability for the prerenderer to capture these 2 different types of responses from a command

packages/host/tests/acceptance/commands-test.gts
```diff
@@ -466,6 +491,20 @@ module('Acceptance | Commands tests', function (hooks) {
...
+      module: `${testRealmURL}command-runner-hello-command`,
+      name: 'default',
+    };
+    let commandParam = encodeURIComponent(JSON.stringify(commandRef));
+    await visit(`/command-runner/1?command=${commandParam}`);
+
+    await waitFor('[data-prerender][data-prerender-status="ready"]');
+    assert
+      .dom('[data-test-command-runner-greeting]')
+      .includesText('Hello from command runner');
```
habdelra: a command can return either a card or it can return a string. which one of these is this? and where is your test for the other case?

packages/realm-server/prerender/render-runner.ts
habdelra: it's up to you what you want to return here. it looks like a command can return either a card or a string. you should have optional properties for these 2 different types of results.

packages/realm-server/prerender/prerender-app.ts
```diff
@@ -337,6 +338,147 @@ export function buildPrerenderApp(options: {
     },
   });
 
+  router.post('/run-command', async (ctxt: Koa.Context) => {
```
habdelra: can we try to reuse the registerPrerenderRoute function here? you have a lot of dupe code that is covered by that function.

packages/realm-server/prerender/utils.ts
```diff
@@ -61,6 +67,25 @@ export async function transitionTo(
...
+  encodedCommand: string,
+  encodedInput?: string,
+): string {
+  let origin = page.url();
+  try {
+    origin = new URL(origin).origin;
+  } catch {
+    // best effort; fall back to raw page url
+  }
+  let url = `${origin}/command-runner/${encodeURIComponent(nonce)}?command=${encodedCommand}`;
```
habdelra: like i mentioned before, let's not use query params for the command. make it a dynamic segment of the route--this is idiomatic ember...

packages/realm-server/prerender/utils.ts
```diff
@@ -61,6 +67,25 @@ export async function transitionTo(
...
+  if (encodedInput) {
+    url += `&input=${encodedInput}`;
```
habdelra: ditto for the input

packages/host/app/templates/command-runner.gts
tintinthong: not always. but this template renders conditionally

packages/host/app/routes/command-runner.ts
tintinthong: renamed to cardResult and cardResultString

packages/realm-server/tests/prerender-server-test.ts
tintinthong: prerender only ever returns the resultCardString -- which is the serialized json

packages/realm-server/prerender/prerender-app.ts
tintinthong: refactored

packages/realm-server/prerender/utils.ts
tintinthong: Yep updated to refactor

packages/host/app/router.ts
```diff
@@ -18,6 +18,9 @@ Router.map(function () {
   this.route('module', { path: '/module/:id/:nonce/:options' });
   this.route('connect', { path: '/connect/:origin' });
   this.route('standby');
```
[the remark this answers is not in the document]
habdelra: this is an interesting note. i don't think that will work tho since fundamentally our only interface for the prerender server is the URL since we are relying on a chrome browser to process our input. it might mean tho that we need to de-reference command input--like if its super massive, perhaps the input can be a card instance whose id you provide as a parameter instead of the actual payload.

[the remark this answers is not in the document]
habdelra: codex says this:

In **Chrome (Chromium)**, the practical max URL length is about **2 MB** (roughly **2,097,152 ASCII characters**).

Two important caveats:
- The address bar (omnibox) only displays about **32 KB** of URL text on most platforms.
- Real-world limits are often lower because servers/proxies may reject long URLs.

Source: https://chromium.googlesource.com/chromium/src/+/HEAD/docs/security/url_display_guidelines/url_display_guidelines.md

packages/host/app/routes/command-runner.ts
```diff
@@ -0,0 +1,189 @@
...
+      model.commandInput = JSON.stringify(commandInput);
+    }
+
+    if (!command) {
+      model.status = 'error';
+      model.error = new Error('Missing or invalid command');
+      return model;
+    }
+
+    void this.#runCommand(model, command, commandInput);
```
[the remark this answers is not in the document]
habdelra: this is fascinating. a malicious card could be written to execute commands evilly using this route. i wonder if we should also have a shared secret that is supplied as an additional route parameter that the server needs to verify in order to run a command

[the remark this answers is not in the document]
habdelra: might be worth a team (office hours) discussion

packages/host/tests/acceptance/commands-test.gts
```diff
@@ -466,6 +492,34 @@ module('Acceptance | Commands tests', function (hooks) {
...
+      let inputParam = encodeURIComponent(JSON.stringify(null));
+      await visit(`/command-runner/${commandParam}/${inputParam}/1`);
+
+      await waitFor('[data-prerender][data-prerender-status="ready"]');
+      assert
+        .dom('[data-test-command-runner-greeting]')
+        .includesText('Hello from command runner');
+    });
+
+    test('route captures command runtime error', async function (assert) {
```
habdelra, line 530: do we also have a test on the server where we test that we can capture a command error from puppeteer and return successfully from the prerenderer.ts?

habdelra (commented):
in general i think this looks good. i had a couple questions, and probably you should go thru all the new co-pilot comments.  I think we need to talk about the security issue that codex found probably during office hours and how we want to mitigate that as well.

packages/host/tests/acceptance/commands-test.gts
tintinthong: already covered here "it captures run-command error state"

it returns run-command error state

packages/realm-server/prerender/render-runner.ts
```diff
@@ -386,6 +390,191 @@ export class RenderRunner {
...
+          localStorage.setItem('boxel-session', sessionAuth);
+          localStorage.setItem(
+            key,
+            JSON.stringify({
+              command: commandToRun,
+              input,
+              nonce: requestNonce,
+              createdAt,
+            }),
+          );
```
tintinthong, line 452: as discussed we now carry set input (as well as command)inside localStorage

packages/host/app/routes/command-runner.ts
```diff
@@ -0,0 +1,267 @@
...
+    }
+    if (!requestId || typeof requestId !== 'string') {
+      return undefined;
+    }
+    let key = `${commandRequestStorageKeyPrefix}${requestId}`;
+    let raw = window.localStorage.getItem(key);
+    if (!raw) {
+      return undefined;
+    }
+    window.localStorage.removeItem(key);
```
tintinthong, line 175: as discussed, the headless chrome then picks up the localStorage key via requestId from the url

tintinthong:
Here is documentation of the updated behaviour 


## Host `command-runner` Route

```ts
route: /command-runner/:request_id/:nonce

type CommandRunnerRouteParams = {
  request_id: string;
  nonce: string;
};
```

### Request payload handoff via localStorage

```ts
const requestId = '6f5508cf-0f10-44a8-a288-0f11f74c4f20';
const command = '@cardstack/boxel-host/commands/show-card/default';
const input = {
  cardId: 'http://localhost:4201/experiments/Author/jane-doe',
  format: 'isolated',
};
const nonce = '2';

localStorage.setItem(
  `boxel-command-request:${requestId}`,
  JSON.stringify({
    command,
    input,
    nonce,
    createdAt: Date.now(),
  }),
);

const url = `http://localhost:4200/command-runner/${encodeURIComponent(requestId)}/${encodeURIComponent(nonce)}`;
```

```txt
http://localhost:4200/command-runner/6f5508cf-0f10-44a8-a288-0f11f74c4f20/2
```

### Host-side consumption behavior

- Host route reads and removes `boxel-command-request:${request_id}` from localStorage (one-time consume).
- Route validates nonce match before command execution.
- Route rejects stale entries using `createdAt` TTL checks.

## Manual Host Simulation (No Bot Runner)

Use this when you want to test `command-runner` directly in the browser.

1. Open host in a browser:
  - `http://localhost:4200`
2. Open browser devtools console and run:

```js
const requestId = '6f5508cf-0f10-44a8-a288-0f11f74c4f20';
const nonce = '2';

localStorage.setItem(
  `boxel-command-request:${requestId}`,
  JSON.stringify({
    command: '@cardstack/boxel-host/commands/show-card/default',
    input: {
      cardId: 'http://localhost:4201/experiments/Author/jane-doe',
      format: 'isolated',
    },
    nonce,
    createdAt: Date.now(),
  }),
);
```

3. Visit this URL (or refresh if already there):
  - `http://localhost:4200/command-runner/6f5508cf-0f10-44a8-a288-0f11f74c4f20/2`

Notes:
- `request_id` in the URL must match the localStorage key suffix.
- `nonce` in the URL must match the `nonce` in stored JSON.
- The entry is consumed once; set it again before another refresh.

packages/host/app/router.ts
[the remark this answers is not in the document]
tintinthong: Solved based upon this update 

https://github.com/cardstack/boxel/pull/4024#issuecomment-3959605199

packages/realm-server/tests/command-parsing-utils-test.ts
```diff
@@ -0,0 +1,56 @@
+import { module, test } from 'qunit';
+import { basename } from 'path';
+import { runSharedTest } from '@cardstack/runtime-common/helpers';
+import commandParsingTests from '@cardstack/runtime-common/tests/command-parsing-utils-test';
+
+module(basename(__filename), function () {
+  module('command parsing utils', function () {
+    test('parseBoxelHostCommandSpecifier parses scoped command specifier', async function (assert) {
+      await runSharedTest(commandParsingTests, assert, {});
```
richardhjtan, line 9: are these tests intended with same parameters ({})? or it shall map the same test key as mentioned in `commandParsingTests`

richardhjtan (approved):

PR #4025, opened 2026-02-19
Should not be able to edit file def instance in code mode

FadhlanR:
Previously, a file was set to read-only in code mode only if the user lacked write permissions and the file was binary. In this PR, I added a check for whether the file is a `fileDef` instance, if it is, the file is now marked as read-only.


packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -594,8 +600,51 @@ export default class CodeSubmode extends Component<Signature> {
...
+        state.error = e;
+        state.value = undefined;
+      } finally {
+        state.isLoading = false;
+      }
+    })();
+    return state;
+  });
+
+  private get isFileDefInstance() {
```
jurgenwerk, line 643: What about this existing function? Can it be used here? https://github.com/cardstack/boxel/blob/8d45ff74115ce0462f300d5de389c8dfde3039c1/packages/runtime-common/code-ref.ts#L123-L124

lukemelia (approved):
Approved pending resolution of AI code review comment

PR #4026, opened 2026-02-19
Use local icons URL for test web assets build

IanCal:
Currently in CI and for local dev we hit the production icons. While this doesn't seem to save much/any time in CI it's about a 20+% speedup locally for me (I assume due to latency hitting the US from the UK, balanced against a http2 server vs local http1.1), and removes a production requirement for CI and dev.

Tests already started the server, we just never used it.

Local dev built things but wasn't setup to start the server

## Summary
- Add CORS headers to the local server for icons to be suported
- switch host default `iconsURL` to `http://localhost:4206` for local-first behavior, matching the other defaults
- remove `ICONS_URL` from `test-web-assets` and set production `ICONS_URL=https://boxel-icons.boxel.ai` in `build-host` env setup
- start icons from realm-server `start-development.sh` and `start-test-realms.sh`, and update orchestrator scripts to rely on those scripts while still waiting for `http://localhost:4206` readiness

Key things to review

* Is this override enough for the build and deployments to be pointing to the correct place
* Does this correctly work launching locally for development
* Start-all are all as far as I understand just for local work, any of these scripts actually used for staging/prod?


habdelra (approved):
looks good, could we also document in the port table of the project readme that we are using 4206 for serving icons?

PR #4028, opened 2026-02-19
Cache card-api module in host test adapter

IanCal:
## Summary
- cache `https://cardstack.com/base/card-api` once per test run in `TestRealmAdapter`
- shim each loader with the shared module so subsequent adapters avoid repeated imports
- reuse the shared card-api path in `openFile` and harden loader-shim behavior for in-flight imports


habdelra:
Abandoned PR

PR #4034, opened 2026-02-20
Cache realm setup for host acceptance/integration tests

IanCal:
This change introduces a wrapper function that can cache the sqlite database created during host tests. It will, by default, use the module name for the test, but can have an additional key added if required. 

Caching is *in-process only* there is no caching between test runs or with a browser refresh (as happens on rebuild). 

To use this in a host test, you should wrap the setup code in "withCachedRealmSetup" and add "setupRealmCacheTeardown".

Here is an example from create-file-tests in code submode:

```
  setupRealmCacheTeardown(hooks);


  hooks.beforeEach(async function () {
    ({ adapter } = await withCachedRealmSetup(async () => {
      await setupAcceptanceTestRealm({
        contents: { ...SYSTEM_CARD_FIXTURE_CONTENTS, ...filesB },
        realmURL: testRealmURL2,
        mockMatrixUtils,
      });
      return await setupAcceptanceTestRealm({
        contents: { ...SYSTEM_CARD_FIXTURE_CONTENTS, ...files },
        mockMatrixUtils,
      });
    }));

```

This will  add a small overhead on the first test (x00ms at most), however can save seconds on each subsequent test. in a run, and lowers the amount we need to hit the base realm if you run tests in parallel.

QOL improvement - parallel test running now doesn't complain if one browser would have no tests so you can run ` pnpm ember exam --path ./dist --load-balance --parallel=8` regardless of the number of modules you have. Along with the icons fix that gets my local runtime for the whole suite down to ~10m.


lukemelia (approved):

backspace (approved):
exciting 🤩

PR #4035, opened 2026-02-20
Add FileDef support for prerendered search 

jurgenwerk:
Support using 

```
<PrerenderedCardSearch
    @query={{this.query}}
    @format='embedded' // or other formats
    @realms={{this.realms}}
 >
```

where query is something like:

```
{
  filter: {
    on: { module: 'https://cardstack.com/base/file-api', name: 'FileDef' },
    eq: { url: `${this.realm}/prerendered-filedef-playground.md` },
  },
}
```

Example:
image

There is one important limitation - a single query is still effectively one target kind. It does not return a mixed list of both card instances and file-meta in the same response. I'm not sure this is a priority right now but I think we can add this later if needed.

### DIFF
--- a/packages/host/app/components/card-prerender.gts
+++ b/packages/host/app/components/card-prerender.gts
@@ -429,6 +429,7 @@ export default class CardPrerender extends Component {
     async ({
       url,
       fileData,
+      types,
       renderOptions,
     }: FileRenderArgs): Promise<FileRenderResponse> => {
       this.#nonce++;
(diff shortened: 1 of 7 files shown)


habdelra (approved):
this looks good. did u run into substantial indexing performance issues as a result of this?

lukemelia (approved):

jurgenwerk:
@habdelra I ran 2 benchmark tests locally, I was running reindexing of `experiments` realm:

1. main branch - `time=348.369380s`
2. this branch (with filedef prerendering) - `time=361.538509s`

To get more accurate data I think I'd have to run this multiple times, but judging by this one time benchmark test I would say this is not a significant performance degradation.

PR #4036, opened 2026-02-20
Pin to Node.js 24.13.1 (LTS) for boxel-icons boxel-motion boxel-ui

jurgenwerk:
Was getting these warnings when running lint (locally + CI):

image


backspace (approved):
Thanks for catching this, I don’t know how I missed them!

PR #4042, opened 2026-02-22
Track runtime dependencies transitively during prerendering instead of using index dependency resolution

habdelra:
## Summary
- Switches prerender dependency collection to a runtime-first model using the new `RuntimeDependencyTracker`, so indexed `deps` are based on what rendering actually consumed (modules + instance/file edges), rather than transitive expansion from index rows.
- Integrates tracker lifecycle into render flow (`start/reset/snapshot/stop`) and wires tracking through loader/store/card-api/code-ref paths, including file indexing.
- Preserves query-field exclusion semantics (including overlap retention: deps seen in both query and non-query contexts are kept).
- Refactors index-runner dependency responsibilities for clarity:
  - runtime deps are primary for persisted rows and error-path seeding
  - index-backed lookups are now explicit and limited to invalidation ordering + dependency error fan-out. **(Note that I created ticket CS-10225 to use runtime deps for invalidation ordering as well)**
- Hardens short-circuit prerender error paths so fallback dependency context is preserved (pre-model and invalid-response scenarios), avoiding sparse `error.deps`.
- Fixed bug around prerender timeout for file-def indexing errors

## Testing
- Targeted runtime dependency tracker tests (reset isolation, query classification, overlap retention, root exclusion, loader cache-hit)
- Added/updated focused prerender regressions for short-circuit dep-context preservation

note a lot of changes where indentation related. probably easier to review with "ignore whitespace" enabled


packages/runtime-common/dependency-tracker.ts
```diff
@@ -0,0 +1,410 @@
...
+    if (!normalized) {
+      return;
+    }
+    this.#rootCandidates.add(normalized);
+    if (rootKind === 'instance') {
+      this.#rootCandidates.add(normalized.replace(/\.json$/, ''));
+    } else {
+      this.#rootCandidates.add(normalized.replace(/\.json$/, ''));
+      this.#rootCandidates.add(`${normalized}.json`);
+    }
```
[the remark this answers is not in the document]
habdelra: fixed in https://github.com/cardstack/boxel/pull/4042/commits/bde3725828d04410f7fbebac70ee240ab15cf587

packages/runtime-common/dependency-tracker.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    let merged = {
+      ...this.#currentContext(),
+      ...context,
+    } satisfies RuntimeDependencyTrackingContext;
+    this.#contextStack.push(merged);
+
+    try {
+      let result = cb();
+      if (isPromiseLike(result)) {
```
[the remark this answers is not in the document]
habdelra: fixed in https://github.com/cardstack/boxel/pull/4042/commits/d50682d22d534b07a0b6a34d4e356ba3e5dda426

packages/runtime-common/realm-index-query-engine.ts
```diff
@@ -730,16 +730,37 @@ export class RealmIndexQueryEngine {
           );
         }
         let relationshipType = relationship.data?.type;
-        let expectsFileMeta = relationshipType === FileMetaResourceType;
+        // Legacy relationship payloads may still use "file" as the type.
+        // Treat it as file-meta so linksTo(MarkdownDef) and similar FileDef
+        // relationships resolve through the file index path.
+        let expectsFileMeta =
+          relationshipType === FileMetaResourceType ||
+          relationshipType === 'file';
```
lukemelia: Is this really something we need to maintain backward compatibility on?

lukemelia (approved):

packages/runtime-common/realm-index-query-engine.ts
habdelra: it came up in a test failure. perhaps there is an out-of-date test?

habdelra: it looks like this is out of date: packages/experiments-realm/MarkdownSkill/f9e6e5fa-4c1d-45f2-a1f5-ce8165b9411a.json

habdelra: fixed in https://github.com/cardstack/boxel/pull/4042/commits/86cefa68de07ec1cd9a33b954b82dee333a452b1

PR #4047, opened 2026-02-23
Add theme-related icon head tags when available

backspace:
See the new tests [fail here](https://github.com/cardstack/boxel/actions/runs/22317973315/job/64568752501?pr=4047#step:10:57) before the head format includes the `link` elements. The failing host test is [flaky](https://github.com/cardstack/boxel/pull/4053).

With this branch deployed to production, this favicon checker shows the icon from the Boxel brand guide:

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-23 11-59-50


packages/host/app/index.html
```diff
@@ -10,7 +10,7 @@
 
   <link integrity="" rel="stylesheet" href="{{rootURL}}assets/vendor.css" />
   <link integrity="" rel="stylesheet" href="{{rootURL}}assets/@cardstack/host.css" />
-  <link href="{{rootURL}}boxel-favicon.png" rel="shortcut icon" type="image/x-icon" />
```
backspace, line 13: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel): <p><strong>Note:</strong> The <code>shortcut</code> link type is often seen before <code>icon</code>, but this link type is non-conforming, ignored and <strong>web authors must not use it anymore</strong>.</p>

packages/base/card-api.gts
```diff
@@ -2391,7 +2391,7 @@ export class CardInfoField extends FieldDef {
   @field name = contains(StringField);
   @field summary = contains(StringField);
   @field cardThumbnailURL = contains(MaybeBase64Field);
-  @field theme = linksTo(() => Theme);
+  @field theme = linksTo(() => Theme, { isUsed: true });
```
backspace, line 2434: Currently only the isolated template’s dependencies are fully loaded before prerendering, so this is a workaround to force the theme to be loaded so its properties can be referenced in the head template.

This is why all the serialisation tests now include a null theme relationship. Undesirable? 🥴

lukemelia (approved):

PR #4050, opened 2026-02-23
Fix bug where `isUsed: true` is not being honored for relationships in compound fields

habdelra:
This fixes a bug where we are not honoring the `isUsed: true` property in the field declaration for relationships in compound fields.


backspace:
When I ran this locally with `isUsed` in the homepage’s animated grid card definition the cards still weren’t in the SSR:

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-23 14-13-16

but there are a lot of moving parts here so I may have exercised this wrong, will keep trying

lukemelia (approved):

habdelra:
> When I ran this locally with `isUsed` in the homepage’s animated grid card definition the cards still weren’t in the SSR:
> 
> e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-23 14-13-16
> but there are a lot of moving parts here so I may have exercised this wrong, will keep trying

was the `isUsed` set on the linksToMany field or the contains field that contains the FieldDef?

packages/realm-server/tests/prerendering-test.ts
```diff
@@ -1248,6 +1261,38 @@ module(basename(__filename), function () {
...
+
+              class DogProfileField extends FieldDef {
+                @field primaryOwner = linksTo(Person);
+                @field caretakers = linksToMany(Person);
+              }
+
+              export class DogProfile extends CardDef {
+                static displayName = "Dog Profile";
+                @field name = contains(StringField);
+                @field profile = contains(DogProfileField, { isUsed: true });
```
habdelra: @backspace not sure I totally love this, but the way this implementation works is that the `isUsed: true` is set on the field that contains the FieldDef (which in turn has the linksToMany). i think I will change this so it's more obvious that you put the `isUsed` on the relationship field

backspace: I put it [here](https://github.com/cardstack/boxel-home/blob/main/boxel-ai-website/animated-grid.gts#L389):
```
  @field sampleCards = linksToMany(() => SampleCard, {
    isUsed: true,
    query: {},
  });
```

habdelra: fixed in https://github.com/cardstack/boxel/pull/4050/commits/56c87381eacec3387acd2c6a54c1452363ba0e54

habdelra: > I put it [here](https://github.com/cardstack/boxel-home/blob/main/boxel-ai-website/animated-grid.gts#L389):
> 
> ```
>   @field sampleCards = linksToMany(() => SampleCard, {
>     isUsed: true,
>     query: {},
>   });
> ```

ah yeah, try that again in the latest (https://github.com/cardstack/boxel/pull/4050/commits/56c87381eacec3387acd2c6a54c1452363ba0e54)

backspace: on 2a43f4a (the commit after that), with `prerenderer=debug`:

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-23 16-17-45

for reference, it’s supposed to look like this, the reason it broke is that the sample cards in the grid were hardcoded before and now are related via the query:

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-02-23 16-18-58

habdelra: what are the steps for publishing the home page locally so that I can try to reproduce as well?

backspace: make a local realm and copy the contents of [`boxel-home`](https://github.com/cardstack/boxel-home) into it (you can skip `.realm.json` if you don’t want your realm’s display name to be changed). You could clone [this branch](https://github.com/cardstack/boxel-home/pull/37) to get the `isUsed` change.

I’m deep into magical thinking territory so I usually make a change to `index.json` to trigger a prerender, this may not be needed. Change to host submode and publish, you can just publish to the subdirectory “Your Boxel Space”, that will take a while but then you can “Open Site” and see the result.

habdelra: actually i noticed this ticket around the renderer not waiting for search to finish loading for query fields before taking pre rendered snapshot. @lukemelia could this also be why prerenderer is not capturing homepage animated cards? https://linear.app/cardstack/issue/CS-10153/use-storeinflightcardloads-to-wait-for-rendering-to-be-finished-before

habdelra: I'm gonna merge this PR, but then continue to work thru the homepage issue on ticket CS-10153

PR #4051, opened 2026-02-23
Add TextFileDef and JsonFileDef for plain text and JSON files

lukemelia:
## Summary
- Adds `TextFileDef` for `.txt`/`.text` files with plain text rendering
- Adds `JsonFileDef` for `.json` files with regex-based syntax highlighting (keys, strings, numbers, booleans, null) in a dark-themed code block
- Registers both extensions in `file-def-code-ref.ts`
- Includes acceptance tests for both file defs

## Test plan
- [x] `ember test --path dist --filter "text file def"` — 3 tests pass
- [x] `ember test --path dist --filter "json file def"` — 3 tests pass

image

image

### DIFF
--- a/packages/base/json-file-def.gts
+++ b/packages/base/json-file-def.gts
@@ -70,9 +70,9 @@ function highlightJson(json: string): string {
     /(&quot;)((?:[^&]|&(?!quot;))*)(&quot;)\s*:/g,
     '<span class="json-key">$1$2$3</span>:',
   );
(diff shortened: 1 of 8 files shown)


habdelra:
regarding the failing indexing test, i saw something like that too and addressed it in a PR that was just merged. perhaps if you merge upstream changes that will fix the issue.

habdelra (approved):

PR #4056, opened 2026-02-24
Update the prerenderer to wait for query fields to load

habdelra:
This PR updates the prerenderer to wait for the search promise from query fields to complete (as well as any downstream promises from the search results) before capturing a snapshot.

### DIFF
--- a/packages/base/query-field-support.ts
+++ b/packages/base/query-field-support.ts
@@ -16,6 +16,7 @@ import {
   getField,
   getSingularRelationship,
   identifyCard,
+  isCardInstance,
   THIS_INTERPOLATION_PREFIX,
   THIS_REALM_TOKEN,
   realmURL as realmURLSymbol,
@@ -40,6 +41,7 @@ interface QueryFieldState {
     status?: number;
   }>;
   searchResource?: StoreSearchResource;
+  renderCycleBarrier?: Promise<void>;
 }
 
 const queryFieldStates = initSharedState(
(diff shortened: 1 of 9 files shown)


lukemelia (approved):
The patching via puppeteer stuff seems pretty fragile. Did you consider an alternative approach?

habdelra:
> The patching via puppeteer stuff seems pretty fragile. Did you consider an alternative approach?

an alternative approach did not occur to me (nor codex) despite prompting. we are patching for 2 reasons:

1. to add a delay to the search response for deterministic results
2. to handle the fact that we don't support cross realm server queries: the realm server is running with a dynamic port (4445) vs the fact that the prerender server is using the app running on the default 4201 server.

we might be able to eliminate the cross realm server support if we move this to host test, but then the issue is that the thing we are trying to test is not even real in the host tests--the wait that we use  for prerendering capture in the host tests is totally different and that would just end up testing the test infrastructure and not the feature itself.

i think we do plan on supporting cross realm server searches, but we have no way of handling auth in those scenarios because we dont yet have a federated identity. but when we do we can slim down the patch drastically

backspace:
I tried this with the homepage and the animated cards aren’t there in the SSR, is this meant to address that? But it’s possible my environment is trashed because I was working on the `@cardstack/catalog` thing so everything’s a jumble

habdelra:
> I tried this with the homepage and the animated cards aren’t there in the SSR, is this meant to address that? But it’s possible my environment is trashed because I was working on the `@cardstack/catalog` thing so everything’s a jumble

i havent tested that out locally yet. i'll try it as well

habdelra:
> I tried this with the homepage and the animated cards aren’t there in the SSR, is this meant to address that? But it’s possible my environment is trashed because I was working on the `@cardstack/catalog` thing so everything’s a jumble

so it looks like there is more to it. there appears to be a rehydration issue when the seeded query is empty. I think in order to keep the PR size managable, i'll tackle that as a followup PR.

UPDATE:  I do have it working now with more of the  work that I'll include in subsequent PR: 

image

PR #4060, opened 2026-02-25
CS-10113: support AI Assistant file attachments

lukemelia:
## Summary
- add end-to-end AI Assistant file attachments from realm and local sources with pending/error/retry UX and send gating
- upload attachments to Matrix with FileDef representations, binary-safe content-hash dedupe, and stable attachment ordering
- make prompt construction model-aware (native image inputs where supported, metadata fallback for unsupported MIME) and scope tool-based file reads to same conversation


lukemelia:
Closing in favor of #4130

PR #4062, opened 2026-02-25
fix: improve error message for invalid icon imports

FadhlanR:
**Problem**

When a card definition has a static icon that is a bad named import (e.g. `import { NonExistent } from '@cardstack/boxel-icons/non-existing-path/nope'`), JavaScript silently resolves it to undefined. This caused the prerenderer to produce a cryptic `[data-prerender] has no child element to capture` error, giving developers or ai-bot no actionable information.

**Fix**

Added a guard in the render/icon route's model() hook that checks whether `cardTypeIcon()` resolved to a valid component. If not, it throws a descriptive error: `static icon is undefined — check that the import resolves to a valid icon component`. Because Ember is still intact at this point, the error propagates via the parent render route's error action hook into the render.error route, which renders a `[data-prerender-error]` element. The prerenderer's `captureResult` picks this up and surfaces the message in the error document.

**Before**
Screenshot 2026-02-26 at 13 15 32

**After**
Screenshot 2026-02-26 at 13 06 31


backspace (approved):
Is it possible to show the actual incorrect import? But regardless, this is a huge DX improvement 🎉

packages/host/app/routes/render/icon.ts
```diff
@@ -25,6 +25,12 @@ export default class RenderIconRoute extends Route<Model> {
     if (!instance) {
       throw new Error('Missing render instance');
     }
-    return { Component: cardTypeIcon(instance) };
+    let component = cardTypeIcon(instance);
+    if (!component) {
+      throw new Error(
+        `static icon is undefined — check that the import resolves to a valid icon component`,
```
habdelra: lets improve this message by including in the error the instance's constructor name. that will help who ever read the error to know which card def has the bad icon definition

PR #4064, opened 2026-02-25
Fix JSON syntax highlighting in metadata panel

lukemelia:
## Summary
- Fix a bug where `NUMBER_RE`, `BOOL_RE`, and `NULL_RE` regex patterns matched content inside already-highlighted JSON string spans (e.g., port numbers like `4201` in URLs would get green number highlighting instead of the correct orange string color)
- Use a placeholder strategy: replace keys and strings with null-byte delimited placeholders before applying number/boolean/null highlighting, then restore the placeholders with the actual highlighted HTML
- All existing regex patterns and CSS styling are preserved; only the application order/strategy is changed

## Test plan
- [ ] Verify that JSON string values containing numbers (e.g., `"http://localhost:4201/..."`) render entirely in the string color (orange `#ce9178`)
- [ ] Verify that JSON string values containing `true`, `false`, or `null` substrings render entirely in the string color
- [ ] Verify that standalone number values, booleans, and nulls outside of strings still get their respective colors (green, blue, blue)
- [ ] Verify that JSON keys still render in the key color (blue `#9cdcfe`)

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -122,6 +122,7 @@ import {
   serialize,
   serializeCard,
   serializeCardResource,
+  serializeFileDef,
   resourceFrom,
   type DeserializeOpts,
   type JSONAPIResource,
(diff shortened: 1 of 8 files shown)


lukemelia:
Dupe

PR #4066, opened 2026-02-25
Filter cards by type in Recents Section of Card Picker

burieberry:
Recents section was missing the type filter for the following cases:
- Create New card instance
- Add card to LinksTo field 
- Add card to LinksToMany field 
- Find Instances in code mode

recents


packages/host/app/components/card-search/search-content.gts
```diff
@@ -135,6 +162,31 @@ export default class SearchContent extends Component<Signature> {
...
+            return sub.type;
+          }
+        }
+      }
+      return undefined;
+    }
+    // Search-sheet mode: extract type from carddef: search key (searchForInstances)
+    return getCodeRefFromSearchKey(this.args.searchKey);
+  }
+
```
lukemelia: This seems like it might belong in a query-utils module with unit tests

lukemelia (approved):

packages/host/app/components/card-search/search-content.gts
burieberry: I have added this as `getTypeRefsFromFilter` utility function in `query-field-utils` and also expanded the logic to handle types in other filter kinds and more complex filters.

PR #4067, opened 2026-02-25
Make prerender more reliable for query fields

habdelra:
This change set improves prerender reliability for query-backed UIs by closing timing/auth/recovery gaps from card/module hydration through render-settle capture. It ensures query-driven and nested relationship content (for example hero mini-cards) is loaded before prerender is marked ready, ensures fallback federated-search calls execute with the correct auth/realm context, and aligns timeout/retry behavior across prerender layers. It also hardens failure recovery by treating unhandled rejection-style render failures as eviction-worthy so pooled pages recover cleanly. Tests were expanded and split to reproduce delayed-query/nested-load conditions and private/public fallback-search auth paths, validating that captured HTML/search docs now include expected dynamic content.

Here is a prerender of the homepage index.json isolated format from this branch:

Screenshot from 2026-02-25 10-36-27

Here is the isolated html for the homepage index.json, which includes the hero-mini-cards in the captured HTML:

Screenshot from 2026-02-25 10-35-32

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -1222,7 +1222,9 @@ class LinksTo<CardT extends LinkableDefConstructor> implements Field<CardT> {
         `linksTo field '${this.name}' cannot deserialize a list of resource ids`,
       );
     }
-    let reference = value.links?.self;
+    let resourceId =
+      value.data && 'id' in value.data ? value.data?.id : undefined;
+    let reference = value.links?.self ?? resourceId;
     if (reference == null || reference === '') {
       return null;
     }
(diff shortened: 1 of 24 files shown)


packages/host/tests/helpers/index.gts
```diff
@@ -416,6 +416,44 @@ export async function capturePrerenderResult(
   return { status: 'ready', value: container.children[0][capture]! };
 }
 
+export interface WaitForLoadedImageOptions {
+  timeout?: number;
+  timeoutMessage?: string;
+}
+
+export async function waitForLoadedImage(
```
habdelra, line 424: this was to fix flaky image tests

packages/base/card-api.gts
```diff
@@ -1728,7 +1748,13 @@ class LinksToMany<FieldT extends LinkableDefConstructor> implements Field<
             `linksToMany field '${this.name}' cannot deserialize a list of resource ids`,
           );
         }
-        let reference = value.links?.self;
+        // links.self is used to tell the consumer of this payload how to get the resource via HTTP.
+        // data.id is used to tell the consumer how to find the resource in the included bucket.
+        // Prefer data.id for resourceFrom(), and fall back to links.self when data.id is missing
+        // (the array-style linksToMany format omits data.id).
```
lukemelia: I see that you only moved this comment, but it seems like a bug if only linksToMany omits data.id

lukemelia (approved):

packages/base/card-api.gts
habdelra: I'm not sure that is the case anymore though (i think AI moved the comment to be consistent with the declaration of `resourceId`), but this comment is actually ancient. probably we can remove this note in the comment.

PR #4069, opened 2026-02-26
Add context menu with Delete option to file tree entries

lukemelia:
## Summary

- Adds a "..." context menu to each file row in the code submode file tree (CS-10078)
- Menu is triggered by hovering/clicking the three-dot `ContextButton` or right-clicking the file row
- First menu item is **Delete**, which opens the existing delete confirmation modal
- Context menu is only shown when the user has write access to the realm (read-only realms are unaffected)

## Changes

- **`directory.gts`**: Wraps each file in a `.file-row` div; adds `BoxelDropdown` with hidden `ContextButton` trigger and "Delete" `Menu` item; right-click (`contextmenu`) programmatically opens the dropdown via a `Map<LocalPath, BoxelDropdownAPI>`
- **`file-tree.gts`**: Threads new `onDeleteFile` prop through to `Directory`
- **`code-submode.gts`**: Adds `canWriteToRealm` getter and `deleteFileInTree` action; conditionally passes `@onDeleteFile` to `FileTree`
- **`file-tree-test.ts`**: 4 new acceptance tests

## Test plan

- [x] `can delete a file from file tree context menu` — full delete flow via "..." button
- [x] `can cancel delete from file tree context menu` — cancel flow preserves file
- [x] `can delete a file via right-click in file tree` — contextmenu event triggers dropdown
- [x] `file tree does not show context menu in read-only realm` — no trigger shown for read-only realms
- [x] All 22 file-tree tests pass (0 failures)

image

### DIFF
--- a/packages/host/app/components/editor/directory.gts
+++ b/packages/host/app/components/editor/directory.gts
@@ -1,4 +1,4 @@
-import { fn, concat } from '@ember/helper';
+import { fn, concat, array } from '@ember/helper';
 import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 import Component from '@glimmer/component';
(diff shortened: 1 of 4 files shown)


jurgenwerk (approved):
I added a trash icon 🗑️

image

PR #4074, opened 2026-02-26
Cs 10217 trigger command on webhook event

richardhjtan:
- Use command runner from headless chrome to run command when filter is matched
- Introduce github event card def, this can be use as data source for submission card to query the events of a PR
- Introduce process github event command to create github event card


packages/host/app/commands/save-card.ts
```diff
@@ -28,9 +28,17 @@ export default class SaveCardCommand extends HostBaseCommand<
   protected async run(
     input: BaseCommandModule.SaveCardInput,
   ): Promise<undefined> {
-    await this.store.add(input.card, {
-      realm: input.realm,
-      localDir: input.localDir,
-    });
+    if (input.doNotWaitForPersist) {
```
richardhjtan: Added doNotWaitForPersist option for save card command, this avoid timeout issue when running the command in headless chrome

packages/runtime-common/db-queries/session-room-queries.ts
```diff
@@ -55,12 +57,23 @@ export async function fetchRealmSessionRooms(
...
+    '    AND (read = true OR write = true)',
+    '  )',
+    '  OR EXISTS (',
+    '    SELECT 1 FROM realm_user_permissions',
+    '    WHERE realm_url =',
+    param(realmURL),
+    "    AND username = '*'",
+    '    AND read = true',
+    '  )',
+    ')',
```
richardhjtan, line 82: This is to fix the issue for the user with read-only access (*) to the public submission realm. Before that, users have to refresh the browser to get the latest file list after the GitHub event card is created from the command runner. 

The original query only matched rows where a user had an explicit permission entry. It never matched the wildcard username = * row, so users whose only access to a realm comes from a world-readable permission (* read = true) were silently excluded.

The submissions realm only has explicit entries for the realm owner and the submission bot — everyone else accesses it via *. So the realm server never sent Matrix events to their session rooms after a write, and the UI never received the incremental index notification needed to refresh the card list.

richardhjtan: Hello @lukemelia @IanCal, does this fix sound reasonable to you?

packages/base/command.gts
```diff
@@ -41,6 +41,7 @@ export class SaveCardInput extends CardDef {
   @field card = linksTo(CardDef);
   @field realm = contains(StringField);
   @field localDir = contains(StringField);
+  @field doNotWaitForPersist = contains(BooleanField);
```
lukemelia: Why are we doing this? It seems problematic -- i.e. what if saving fails?

packages/host/app/commands/bot-requests/create-listing-pr-request.ts
```diff
@@ -68,4 +79,47 @@ export default class CreateListingPRRequestCommand extends HostBaseCommand<
...
+  private async registerPRWebhook(): Promise<void> {
+    try {
+      // TODO: Avoid the list+find round-trip by storing the webhook ID (output
+      // by register-github-webhook script) as a realm config or env var so the
+      // command can fetch the GitHub webhook directly by ID.
+      const webhooks = await this.realmServer.listIncomingWebhooks();
+      const githubWebhook = webhooks.find(
+        (w: { verificationType: string }) =>
+          w.verificationType === 'HMAC_SHA256_HEADER',
+      );
```
lukemelia: This is not a robust way to lookup the webhook. Other webhooks could use this verification type as well couldn't they?

lukemelia: Probably should go ahead and do the above TODO now

packages/host/app/services/realm-server.ts
```diff
@@ -338,6 +338,10 @@ export default class RealmServerService extends Service {
       .map((r) => r.url);
   }
 
+  get submissionRealmURL(): string {
+    return `${this.url.href}submissions/`;
+  }
```
lukemelia: This seems awkwardly specific for the realm-server service. Can we define it somewhere else?

packages/base/commands/search-card-result.gts
```diff
@@ -37,6 +38,9 @@ interface CardListSignature {
 
 export class JsonField extends FieldDef {
   static [primitive]: Record<string, any>;
+  static [queryableValue](_value: any, _stack: BaseDef[]): null {
+    return null;
+  }
```
lukemelia: why was this necessary?

packages/catalog-realm/github-event/github-event.gts
```diff
@@ -0,0 +1,14 @@
...
+import { JsonField } from 'https://cardstack.com/base/commands/search-card-result';
+
+export class GithubEventCard extends CardDef {
+  static displayName = 'GitHub Event';
+
+  @field eventType = contains(StringField); // 'pull_request', 'check_run', etc.
+  @field action = contains(StringField); // 'opened', 'completed', etc.
+  @field prNumber = contains(NumberField);
+  @field prUrl = contains(StringField);
+  @field payload = contains(JsonField); // full raw payload
```
lukemelia, line 10: If you are going to store the full payload, should we just make the other fields computed on it?

packages/host/app/commands/bot-requests/create-listing-pr-request.ts
```diff
@@ -73,4 +79,47 @@ export default class CreateListingPRRequestCommand extends HostBaseCommand<
...
+          'No catalog realm URL found, skipping webhook command registration.',
+        );
+        return;
+      }
+
+      await this.realmServer.createWebhookCommand({
+        incomingWebhookId: githubWebhook.id,
+        command: `${catalogRealmURL}commands/process-github-event/default`,
+        filter: {
+          submissionRealmUrl: this.realmServer.submissionRealmURL,
```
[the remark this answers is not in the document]
lukemelia: I was wondering the same thing. Isn't this done by matrix/scripts/register-github-webhook.ts?

packages/host/app/commands/create-listing-pr.ts
```diff
@@ -187,7 +187,7 @@ export default class CreateListingPRCommand extends HostBaseCommand<
 
     log.debug('PR created successfully:', prResult);
 
-    // Open room and send PR status message with full details
+    // Open room and send PR status message
```
lukemelia, line 190: What kind of response do we expect from this message?

packages/host/app/commands/save-card.ts
```diff
@@ -28,9 +28,17 @@ export default class SaveCardCommand extends HostBaseCommand<
...
+      this.store.add(input.card, {
+        realm: input.realm,
+        localDir: input.localDir,
+        doNotWaitForPersist: true,
```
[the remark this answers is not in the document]
lukemelia: I also wonder this

packages/realm-server/handlers/handle-webhook-receiver.ts
```diff
@@ -70,13 +73,112 @@ export default function handleWebhookReceiverRequest({
...
+
+        // Check if PR number matches filter (for pull_request events)
+        if (
+          commandFilter.prNumber &&
+          payload.pull_request?.number !== commandFilter.prNumber
+        ) {
+          continue;
+        }
+
+        // Additional filter checks can be added here as needed
```
lukemelia: The command filter should specify a type along with its configuration and we should use the type to look up a code block somewhere that has this specific logic of comparing the payload with the filter configuration. The logic should not be in line within this handle-webhook-receiver.

packages/realm-server/handlers/handle-webhook-receiver.ts
```diff
@@ -70,13 +73,112 @@ export default function handleWebhookReceiverRequest({
...
+      ]);
+      let runAs =
+        (realmOwnerRows[0]?.username as string | undefined) ??
+        (webhook.username as string);
+
+      let commandInput = {
+        eventType: eventType ?? '',
+        submissionRealmUrl,
+        payload,
+      };
```
lukemelia: This logic of assembling the command input also should not be in line here. It needs to be abstracted in some way. The idea is that this route handler should handle any webhook in the system, not just the github webhook.

packages/realm-server/handlers/handle-webhook-receiver.ts
```diff
@@ -70,13 +73,112 @@ export default function handleWebhookReceiverRequest({
...
+
+      try {
+        await enqueueRunCommandJob(
+          {
+            realmURL: submissionRealmUrl,
```
lukemelia: This shouldn't be fixed. It should be based on something from the database.

packages/runtime-common/db-queries/session-room-queries.ts
lukemelia, line 82: Did you check other consumers of this method to see if there are any unintended consequences of this change?

lukemelia (changes requested):
This needs some work. See comments.

packages/catalog-realm/commands/process-github-event.gts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  protected async run(input: ProcessGithubEventInput): Promise<undefined> {
+    const { eventType, submissionRealmUrl, payload } = input;
+
+    let card = new GithubEventCard({
+      eventType,
+      action: payload?.action ?? null,
+      prNumber: payload?.pull_request?.number ?? null,
+      prUrl: payload?.pull_request?.html_url ?? null,
```
lucaslyl: if see the instance structure, check_run events don't have a top-level pull_request field, so payload?.pull_request?.number always returns null for
  them. The PR number is nested inside check_run.pull_requests[0].number instead. 

So you need add two fallbacks to extract prNumber for check_run events, so stored cards are queryable by PR number.

Example as below code, that works for me to query the check run/check suite event, but still needs more testing on this:
```
 prNumber:
        payload?.pull_request?.number ??
        payload?.check_run?.pull_requests?.[0]?.number ??
        payload?.check_suite?.pull_requests?.[0]?.number ??
        null,
```

packages/host/app/commands/bot-requests/create-listing-pr-request.ts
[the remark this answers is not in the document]
richardhjtan: Indeed, that was a duplicate of my first approach - I will remove this

packages/base/commands/search-card-result.gts
richardhjtan: Screenshot 2026-03-02 at 10 35 04 PM

@lukemelia  This is to fix this card error

packages/base/command.gts
richardhjtan: Agree, reverted this

packages/runtime-common/db-queries/session-room-queries.ts
richardhjtan: There are no tests about this, I have conducted a test to demonstrate different scenarios to understand and avoid unintentional behaviour

richardhjtan:
Close this PR due to the base branch PR being closed and split into several PRs.

The feedback changes and newer approach are addressed in https://github.com/cardstack/boxel/pull/4098

cc: @lukemelia @tintinthong

PR #4076, opened 2026-02-26
Update card-picker search result summary display

burieberry:
This PR gets rid of the customization around the result summary that was specified in the designs because it didn't work well in practice, as we've seen during the demo and created confusion. Results will be always be displayed in the form "[X] results across [Y] realms". The count from the Recents section does not affect the total result count.

results-1

results-2


lukemelia (approved):
Do any tests need to change?

burieberry:
> Do any tests need to change?

Tests weren't checking for the additional statement for specific realms. I can add a test in one of the Recents tests, since that one previously didn't show the total count.

PR #4081, opened 2026-02-27
Change base references to @cardstack/base

backspace:


packages/host/app/routes/module.ts
```diff
@@ -469,7 +466,11 @@ async function getTypes(
         codeRef: loadedCardRef,
         displayName: getDisplayName(loadedCard),
       });
-      if (!isEqual(loadedCardRef, baseCardRef)) {
+      // This was isEqual but it was comparing @cardstack/base/ and localhost:4201/base/ … was that valid? is this?
+      if (
+        internalKeyFor(loadedCardRef, undefined) !==
+        internalKeyFor(baseCardRef, undefined)
+      ) {
```
backspace, line 473: I’m very uncertain about this, does it reflect a fundamental problem with the approach?

PR #4091, opened 2026-03-02
Command to copy a file to another realm

FadhlanR:
In this PR, I implemented a command to copy files from one realm to another. I also added a 'Copy File' menu item to files attached in the AI Assistant.


habdelra:
there's a fundamental issue here that we've discussed around copying before that has not been addressed--specifically how you deal with relationships. A card may have relative relationships to other cads in its own realm, or relationships to other cards in other private realms, as well as relationships to cards in public realms. same for modules. A card may import modules relatively in it's own realm, import modules in other private realms, or import modules in public realms. copying should be able to trace the graph of consumed dependencies and make a decision around what to pull in, and what to leave out. only in the most simplistic cases are you ever copying a single instance. many times though you need to consider the graph of dependencies (instances, file-defs, modules) that also need to be carried over. Failure to consider this means that you'll end up with an errored instance. Probably this is worth a discussion.

habdelra:
> there's a fundamental issue here that we've discussed around copying before that has not been addressed--specifically how you deal with relationships. A card may have relative relationships to other cads in its own realm, or relationships to other cards in other private realms, as well as relationships to cards in public realms. same for modules. A card may import modules relatively in it's own realm, import modules in other private realms, or import modules in public realms. copying should be able to trace the graph of consumed dependencies and make a decision around what to pull in, and what to leave out. only in the most simplistic cases are you ever copying a single instance. many times though you need to consider the graph of dependencies (instances, file-defs, modules) that also need to be carried over. Failure to consider this means that you'll end up with an errored instance. Probably this is worth a discussion.

chris's feedback was that this is ok. we want the copy operation to be very primitive

lukemelia (approved):

PR #4095, opened 2026-03-02
optimization: automatically create a user on first server session, rather than requiring another request to create the user

jurgenwerk:
## Summary
- Users are now automatically created in the `users` table when they get a server session (`POST /_server-session`), eliminating the requirement for a separate `POST /_user` call
- Added `getOrCreateUser` (atomic INSERT ... ON CONFLICT DO NOTHING) and `updateUserRegistrationToken` to `user-queries.ts`
- `POST /_user` is now idempotent — updates the registration token for existing users instead of returning 422
- Initial daily credits are granted on first creation (either via session or explicit endpoint), with no double-granting

## Test plan
- [x] Verify existing user login flow works (user auto-created during session)
- [x] Verify new user registration flow works (POST /_user still creates + grants credits)
- [x] Verify calling POST /_user twice returns 200 both times with no duplicate credits
- [x] Verify registration token is updated on second POST /_user call
- [x] Run `realm-endpoints/user-test.ts` tests

Closes CS-10253

I also tested this manually by creating a new user and I confirm it works OK and that the matrix registration token (e.g. `dev-token`) gets written to to the users database.

### DIFF
--- a/packages/host/app/services/matrix-service.ts
+++ b/packages/host/app/services/matrix-service.ts
@@ -538,11 +538,10 @@ export default class MatrixService extends Service {
       );
     }
 
-    await this.realmServer.createUser(userId, registrationToken);
-
     await this.start({ auth });
     this.setDisplayName(displayName);
 
+    await this.realmServer.loginWithRegistrationToken(registrationToken);
     await this.realmServer.authenticateToAllAccessibleRealms();
 
     await Promise.all([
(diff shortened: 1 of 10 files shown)


habdelra:
is there a place on the client where we were making the request to create a user that that this now replaces that we can remove?

jurgenwerk:
That's right, there's some cleanup to do, adding that shortly

jurgenwerk:
I also tested this manually by creating a new user and I confirm it works OK and that the matrix registration token (e.g. dev-token) gets written to to the users database.

lukemelia (approved):

PR #4098, opened 2026-03-02
CS-10217 trigger command on webhook event

richardhjtan:
## What is changing 
- Triggers a command when a webhook event is received, enabling automated workflows (e.g., creating a GithubEventCard for every GitHub PR/review/check event).
- Introduce github event card def, this can be use as data source for submission card to query the events of a PR
- Introduce process github event command to create github event card
- Cover tests 

**Webhook command execution (realm-server)**
- handle-webhook-receiver.ts — After verifying the webhook signature, fetches matching webhook_commands rows from the DB, filters them against the incoming payload using pluggable filter handlers, and enqueues run-command jobs for each match. Falls back to the webhook-registering user if no realm owner is found.
- webhook-filter-handlers.ts (new) — Introduces a WebhookFilterHandler interface with a registry pattern. Ships two handlers:
github-event — matches by X-GitHub-Event header; assembles { eventType, realm, payload } as command input.
default — pass-through; always matches and forwards the raw payload.


packages/catalog-realm/commands/process-github-event.gts
```diff
@@ -0,0 +1,45 @@
...
+
+    let card = new GithubEventCard({
+      eventType,
+      payload,
+    });
+
+    let savedCard = await new SaveCardCommand(this.commandContext).execute({
+      card,
+      realm: submissionRealmUrl,
+    });
```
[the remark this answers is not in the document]
richardhjtan: Webhook registration is currently admin-only

habdelra:
from your description:
> realm.ts — Supports a new X-Boxel-Async-Index: true request header on card POST. When set, the realm writes the file to disk and persists file meta, then returns 201 with the assigned card ID immediately. Indexing runs fire-and-forget in the background. This prevents a deadlock where the run-command worker job can't pick up the incremental-index job it enqueues.

> store.ts — Uses the async-index path when creating new cards inside the headless Chrome render context (or when asyncIndex option is explicitly set). Calls setId() on the card instance with the server-assigned ID so downstream commands can reference it.

are we still doing this?

richardhjtan:
@habdelra planning to remove it as we shall try increasing the worker approach in local from the office hour discussion

packages/base/commands/search-card-result.gts
```diff
@@ -37,6 +38,9 @@ interface CardListSignature {
 
 export class JsonField extends FieldDef {
   static [primitive]: Record<string, any>;
+  static [queryableValue](_value: any, _stack: BaseDef[]): null {
+    return null;
+  }
```
tintinthong, line 43: do we need this?

packages/catalog-realm/commands/process-github-event.gts
```diff
@@ -0,0 +1,41 @@
+import { CardDef, field, contains } from 'https://cardstack.com/base/card-api';
+import StringField from 'https://cardstack.com/base/string';
+import { JsonField } from 'https://cardstack.com/base/commands/search-card-result';
+import { Command } from '@cardstack/runtime-common';
+import SaveCardCommand from '@cardstack/boxel-host/commands/save-card';
+import { GithubEventCard } from '../github-event/github-event';
+
+class ProcessGithubEventInput extends CardDef {
+  @field eventType = contains(StringField); // from command_filter
+  @field submissionRealmUrl = contains(StringField); // from command_filter
```
tintinthong: only one submssion realm right? field not needed. Do we need that?

tintinthong: or maybe if its a write. It shud just b called realm

packages/host/app/services/store.ts
```diff
@@ -390,6 +393,17 @@ export default class StoreService extends Service implements StoreInterface {
...
+      // downstream commands can reference them. Use async-index so the realm
+      // returns the ID immediately without waiting for indexing to complete.
+      if (!instance.id && !opts?.doNotPersist) {
+        return (await this.persistAndUpdate(instance, {
+          realm: opts?.realm,
+          localDir: opts?.localDir,
+          asyncIndex: true,
+        })) as T | CardErrorJSONAPI;
+      }
+      // Existing cards: skip saves in render context (no auto-save).
```
tintinthong: do we need this portion of code?

packages/runtime-common/db-queries/session-room-queries.ts
```diff
@@ -55,12 +63,23 @@ export async function fetchRealmSessionRooms(
   let rows = await query(dbAdapter, [
```
tintinthong, line 63: why was this fetchSessionRooms changed ahh?

tintinthong (commented):
Think we want to move all the async index stuff out of this PR

packages/base/commands/search-card-result.gts
richardhjtan: Screenshot 2026-03-02 at 10 35 04 PM

It will have this error initially before adding this

packages/host/app/services/store.ts
richardhjtan: Not needed because is part of async index work, will take it out

packages/runtime-common/db-queries/session-room-queries.ts
richardhjtan: @tintinthong  Before that, users with read access from the public realm (eg: db permission record like catalog/submission realm with * all users) will not get a notification when a new card is created

The obvious scenario is when in the code mode file tree, the webhook is triggered and the command creates a card - the user will not see the latest file added in the LHS under the submission realm. The user has to refresh to see the latest file result

tintinthong (approved):

packages/catalog-realm/github-event/github-event.gts
```diff
@@ -0,0 +1,29 @@
+import { CardDef, field, contains } from 'https://cardstack.com/base/card-api';
+import StringField from 'https://cardstack.com/base/string';
+import NumberField from 'https://cardstack.com/base/number';
+import { JsonField } from 'https://cardstack.com/base/commands/search-card-result';
+
+export class GithubEventCard extends CardDef {
```
tintinthong, line 6: is there any value in subclassing this.

richardhjtan: Might not be for now. If we found more different event cards involved, we can refactor the base event card which has similar attributes

PR #4109, opened 2026-03-04
Fix catalog env SKIP_CATALOG again

tintinthong:
The pattern is SKIP_CATALOG is either true or not set


packages/host/config/environment.js
```diff
@@ -59,7 +60,7 @@ module.exports = function (environment) {
     realmServerURL: process.env.REALM_SERVER_DOMAIN || 'http://localhost:4201/',
     resolvedBaseRealmURL:
       process.env.RESOLVED_BASE_REALM_URL || 'http://localhost:4201/base/',
-    resolvedCatalogRealmURL: process.env.SKIP_CATALOG
```
tintinthong, line 62: process.env.SKIP_CATALOG="false" will be registered as true in node

richardhjtan (approved):

packages/host/config/environment.js
```diff
@@ -77,7 +78,7 @@ module.exports = function (environment) {
     // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
     // ENV.APP.LOG_VIEW_LOOKUPS = true;
     ENV.defaultSystemCardId = process.env.DEFAULT_SYSTEM_CARD_ID;
-    if (!ENV.defaultSystemCardId && !process.env.SKIP_CATALOG) {
+    if (!ENV.defaultSystemCardId && !skipCatalog) {
```
habdelra, line 81: why don't you just test against the string value of 'true'. then u don't have to worry about the ambiguity

tintinthong: yea thats why skipCatalog variable is doing 

```
  const skipCatalog = process.env.SKIP_CATALOG === 'true';
```

PR #4115, opened 2026-03-04
Ember 6.10 upgrade

ef4:
This is extracted out of the vite branch.


ef4:
Next step here is to pick one of the remaining host test failures and dig into it. Many probably share the same few causes.

I was investigating `Integration | ai-assistant-panel | sending: displays retry button for message that failed to send` and ran out of time before getting to the bottom of it.

ef4:
After discussions with @lukemelia and followup investigations, there's only one outstanding open question here, which is figured out what is consuming cached old-wire-format templates in our CI (making the patch of ember-source required). We're going to land this and then follow up on that part. We need to do that anyway because the wire format will change again (there are already more upstream wire format changes -- and it's not guarantee stable between any particular ember versions).

PR #4121, opened 2026-03-05
Update boxel ui input type checkbox

FadhlanR:
Previously, the 'show only' checkbox in the card chooser did not use the boxel-ui input component. In this PR, I’ve updated the design of the boxel-ui checkbox to ensure it meets the requirements for use within the card chooser.

Screenshot 2026-03-05 at 15 31 54

### DIFF
--- a/packages/base/structured-theme-variables.gts
+++ b/packages/base/structured-theme-variables.gts
@@ -450,6 +450,10 @@ export default class ThemeVarField extends FieldDef {
   @field trackingNormal = contains(CSSValueField, {
     description: 'Specifies letter-spacing base value.',
   });
+  @field boxelBodyFontSize = contains(CSSValueField, {
+    description:
+      'Base body font size. Also controls checkbox and radio button dimensions.',
+  });
 
   // box-shadow variables
   @field shadow2xs = contains(CSSValueField, {
--- a/packages/base/structured-theme-variables.gts
+++ b/packages/base/structured-theme-variables.gts
@@ -450,11 +450,6 @@ export default class ThemeVarField extends FieldDef {
   @field trackingNormal = contains(CSSValueField, {
     description: 'Specifies letter-spacing base value.',
   });
-  @field boxelBodyFontSize = contains(CSSValueField, {
-    description:
-      'Base body font size. Also controls checkbox and radio button dimensions.',
-  });
-
   // box-shadow variables
   @field shadow2xs = contains(CSSValueField, {
     description: 'Smallest shadow depth.',
--- a/packages/boxel-ui/addon/src/components/input/index.gts
+++ b/packages/boxel-ui/addon/src/components/input/index.gts
@@ -75,7 +75,7 @@ export interface Signature {
     size?: 'large' | 'default';
     state?: InputValidationState;
     type?: InputType;
-    value: string | number | null | undefined;
+    value: string | number | boolean | null | undefined;
   };
   Element: HTMLInputElement | HTMLTextAreaElement | HTMLDivElement;
 }
@@ -93,6 +93,10 @@ export default class BoxelInput extends Component<Signature> {
 
   get isSearch() {
     return this.args.type === 'search';
+  }
+
+  private get isCheckbox() {
+    return this.args.type === 'checkbox';
   }
 
   private get type() {
@@ -141,6 +145,7 @@ export default class BoxelInput extends Component<Signature> {
         'input-container'
         has-validation=this.hasValidation
         is-multiline=this.isMultiline
+        is-checkbox=this.isCheckbox
       }}
     >
       {{#if (and (not @required) @optional)}}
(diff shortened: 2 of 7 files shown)


packages/boxel-ui/addon/src/components/input/index.gts
```diff
@@ -75,7 +75,7 @@ export interface Signature {
     size?: 'large' | 'default';
     state?: InputValidationState;
     type?: InputType;
-    value: string | number | null | undefined;
+    value: string | number | boolean | null | undefined;
   };
   Element: HTMLInputElement | HTMLTextAreaElement | HTMLDivElement;
```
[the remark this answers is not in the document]
backspace: While I don’t know that we use traditional form submission this does seem possibly confusing

packages/boxel-ui/addon/src/components/input/index.gts
```diff
@@ -432,6 +437,90 @@ export default class BoxelInput extends Component<Signature> {
...
+        .boxel-input[type='checkbox']:focus-visible {
+          outline: 2px solid var(--ring, var(--boxel-highlight));
+          outline-offset: 2px;
+          border-color: var(--checkbox-border-color);
+        }
+
+        .boxel-input[type='checkbox']:hover:not(:disabled):not(:checked) {
+          border-color: var(--boxel-dark);
+        }
+
```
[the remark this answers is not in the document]
backspace: this seems worth addressing

packages/boxel-ui/addon/src/components/input/usage.gts
```diff
@@ -351,6 +434,13 @@ export default class InputUsage extends Component {
         background-color: var(--sidebar);
         color: var(--sidebar-foreground);
       }
+      .checkbox-example-label {
+        display: flex;
+        align-items: center;
+        gap: var(--boxel-sp-xxs);
```
burieberry: --boxel-sp-2xs

packages/boxel-ui/addon/src/components/input/index.gts
```diff
@@ -432,6 +437,90 @@ export default class BoxelInput extends Component<Signature> {
...
+            --boxel-checkbox-checked-background-color,
+            var(--primary, var(--boxel-highlight))
+          );
+          --checkbox-checked-border-color: var(
+            --boxel-checkbox-checked-border-color,
+            var(--primary, var(--boxel-dark))
+          );
+          --checkbox-checkmark-color: var(
+            --boxel-checkbox-checkmark-color,
+            #333
```
burieberry: if checkbox background is `--primary`, the checkmark color should be `--primary-foreground` for contrast.

Image

also the required, optional, helper text placement may need some adjusting

packages/boxel-ui/addon/src/components/input/index.gts
```diff
@@ -432,6 +437,90 @@ export default class BoxelInput extends Component<Signature> {
...
+          border-color: var(--checkbox-checked-border-color);
+        }
+
+        .boxel-input[type='checkbox']:checked::after {
+          content: '';
+          position: absolute;
+          left: 5px;
+          top: 1px;
+          width: 5px;
+          height: 9px;
```
burieberry: try this in a themed card where the theme font-size (`--theme-body-font-size`) and `--spacing` is set to much larger values to see if the checkbox sizing works

FadhlanR:
Thanks for the feedback @burieberry @backspace . I addressed the feedback and I added tests to ensure the theme works with checkbox inputs.

backspace (approved):
Image

Is the contrast of the checkmark a concern of this? With the StarryNight and many other themes, I think the black checkmark probably should be white, or another high-contrast theme colour.

FadhlanR:
> Image
> Is the contrast of the checkmark a concern of this? With the StarryNight and many other themes, I think the black checkmark probably should be white, or another high-contrast theme colour.

I'm currently using --primary-foreground from the theme for the checkmark. I'm not sure if we need a dynamic contrast calculation here. What do you think, @burieberry?

packages/base/structured-theme-variables.gts
```diff
@@ -450,6 +450,10 @@ export default class ThemeVarField extends FieldDef {
   @field trackingNormal = contains(CSSValueField, {
     description: 'Specifies letter-spacing base value.',
   });
+  @field boxelBodyFontSize = contains(CSSValueField, {
+    description:
+      'Base body font size. Also controls checkbox and radio button dimensions.',
+  });
```
burieberry: bodyFontSize already exists in `typography` field in this card. please don't add a new one. Themed Boxel CardContainer uses this value to set `--boxel-body-font-size` (and it includes fallback values). Please do not add this because it already exists.

burieberry (commented):
The checkmark color is now updated. However, there's something strange going on with the grid layout and the size of the checkmark. The checkmark size is very small. We can revert to the fixed values and not do scaling.
Image
Image

FadhlanR:
@burieberry I updated the checkmark implementation to prevent it from being too small. I also hid the error icon and error message for the checkbox input because I don't think they are necessary for checkboxes. Here is the screenshot.

https://github.com/user-attachments/assets/b9638d56-f87e-478a-b64b-310772f93f9d

PR #4143, opened 2026-03-09
feat: keyboard-friendly Choose File modal

lukemelia:
## Summary

Makes the Choose File modal friendlier to keyboard-only users.

### File list area
- The file list `<nav>` is now focusable (`tabindex="0"`, `aria-label="File tree"`)
- **Auto-focus**: when the modal opens, the file list receives focus immediately via `AutoFocusModifier`
- **Type-ahead search**: while the file list (or any button inside it) has focus, typing printable characters builds a buffer and highlights + scrolls to the first visible file or folder whose name starts with the typed string (case-insensitive, 600 ms inactivity resets the buffer)
- Matched entries get a distinct visual highlight (`type-ahead-match` CSS class: boxel-200 background + highlight outline)

### Realm chooser
- The BoxelSelect/PowerSelect trigger is already natively focusable via `tabindex` — no changes needed; added a test to confirm

## Tests

New module `Acceptance | file chooser keyboard tests`:
- ✅ File list is focused on modal open
- ✅ Typing highlights first matching file (e.g. `r` → `README.txt`)
- ✅ Typing highlights first matching file with narrowed query (e.g. `t` → `test-image.png`)
- ✅ Typing highlights a matching directory (e.g. `f` → `FileLinkCard/`)
- ✅ Realm chooser trigger is keyboard focusable
- [ ] Use the preview link to try it out

## Implementation notes

- Type-ahead uses `button.getAttribute('title')` (more reliable than `textContent` which could include SVG icon text)
- Buffer accumulates across key presses and clears after 600 ms of inactivity
- Space is allowed through to child buttons (for button activation), all other printable chars are captured
- `AutoFocusModifier` uses a `#firstRun` guard so it focuses only on initial insertion (not on subsequent renders)

### DIFF
--- a/packages/host/tests/acceptance/file-chooser-test.gts
+++ b/packages/host/tests/acceptance/file-chooser-test.gts
@@ -614,7 +614,10 @@ module('Acceptance | file chooser keyboard tests', function (hooks) {
       trigger?.getAttribute('tabindex') !== null ||
       trigger?.getAttribute('role') === 'button' ||
       trigger?.tagName.toLowerCase() === 'button';
-    assert.ok(triggerIsFocusable, 'realm chooser trigger is keyboard focusable');
+    assert.ok(
+      triggerIsFocusable,
+      'realm chooser trigger is keyboard focusable',
+    );
 
     await click('[data-test-choose-file-modal-cancel-button]');
   });
--- a/packages/host/tests/acceptance/file-chooser-test.gts
+++ b/packages/host/tests/acceptance/file-chooser-test.gts
@@ -486,6 +486,57 @@ module('Acceptance | file chooser keyboard tests', function (hooks) {
     await click('[data-test-choose-file-modal-cancel-button]');
   });
 
+  test('file list area is focused when the modal reopens', async function (assert) {
+    await visitOperatorMode({
+      stacks: [
+        [
+          {
+            id: `${testRealmURL}FileLinkCard/empty`,
+            format: 'edit',
+          },
+        ],
+      ],
+    });
+
+    await click(
+      '[data-test-links-to-editor="attachment"] [data-test-add-new="attachment"]',
+    );
+
+    await waitUntil(
+      () =>
+        document.activeElement ===
+        document.querySelector('[data-test-file-tree-nav]'),
+      {
+        timeout: 5000,
+        timeoutMessage: 'file tree nav was not focused on first open',
+      },
+    );
+
+    await click('[data-test-choose-file-modal-cancel-button]');
+
+    await click(
+      '[data-test-links-to-editor="attachment"] [data-test-add-new="attachment"]',
+    );
+
+    await waitUntil(
+      () =>
+        document.activeElement ===
+        document.querySelector('[data-test-file-tree-nav]'),
+      {
+        timeout: 5000,
+        timeoutMessage: 'file tree nav was not focused on reopen',
+      },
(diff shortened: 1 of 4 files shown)


backspace:
I’ll try other ways but I’m strangely unable to see any files initially:

screencast 2026-03-10 07-09-15

lukemelia:
> I’ll try other ways but I’m strangely unable to see any files initially:

@backspace weird. It is working fine for me on the production preview link.

backspace:
> @backspace weird. It is working fine for me on the production preview link.

it works for me on the production preview too, very strange! Does it work for you on staging?

Could be a future fix, autofocus only works the first time the modal is opened:

screencast 2026-03-10 07-15-10

You can arrow-key out of view, I’d expect it to scroll:

screencast 2026-03-10 07-22-20


Bigger problem but could also be fixed in a followup, typeahead isn’t working for me:

screencast 2026-03-10 07-24-07

IMO none of these are blockers to merging but I do think they should be addressed. Now I’ll look at the code.

packages/host/tests/acceptance/file-chooser-test.gts
```diff
@@ -436,6 +439,522 @@ module('Acceptance | file chooser tests', function (hooks) {
...
+      timeout: 5000,
+      timeoutMessage: 'file tree did not load',
+    });
+
+    await focus('[data-test-file-tree-nav]');
+    await triggerEvent('[data-test-file-tree-nav]', 'keydown', { key: 'r' });
+
+    assert
+      .dom('[data-test-file="README.txt"]')
+      .hasClass('cursor', 'README.txt gets the cursor');
```
backspace, line 598: hmm strange that it wouldn’t work for me in practice, what does that mean about this assertion?

packages/host/tests/acceptance/file-chooser-test.gts
```diff
@@ -436,6 +439,522 @@ module('Acceptance | file chooser tests', function (hooks) {
...
+
+    assert
+      .dom('[data-test-file="test-image.png"]')
+      .doesNotHaveClass('cursor', 'test-image.png does not get the cursor');
+
+    await click('[data-test-choose-file-modal-cancel-button]');
```
backspace, line 604: Could this be removed, and others like it?

packages/host/tests/acceptance/file-chooser-test.gts
```diff
@@ -436,6 +439,522 @@ module('Acceptance | file chooser tests', function (hooks) {
...
+    // Second ArrowDown: cursor moves to next item
+    let firstItem = document.querySelector(
+      '[data-test-file-tree-nav] button.cursor',
+    );
+    await triggerKeyEvent('[data-test-file-tree-nav]', 'keydown', 'ArrowDown');
+    assert.notStrictEqual(
+      document.querySelector('[data-test-file-tree-nav] button.cursor'),
+      firstItem,
+      'cursor moved to a different item on second ArrowDown',
+    );
```
backspace, line 722: This could be extended to ensure the selection stays in view.

backspace (approved):
I appreciate the thoroughness of the tests, I think keyboard functionality is often overlooked in that realm.

lukemelia:
@backspace the issues you reported were due to reliance on data-test-* selectors that are stripped in prod builds. This is now fixed. Thanks for reporting!

PR #4147, opened 2026-03-09
Change realm server test ports in environment mode

backspace:
This lets realm server tests be run in parallel on different branches.


lukemelia (approved):
LGTM assuming you're comfortable with the shard conflict possibility pointed out by AI review

backspace:
ya I think handling environment mode with mise will open up opportunities for mitigating that kind of problem

PR #4154, opened 2026-03-10
Inline scroll-anchor modifier into search-content component

FadhlanR:
## Summary

- **Inlined** the `ScrollAnchor` modifier directly into `search-content.gts` and **renamed** it to `ScrollToFocusedSection`
- **Deleted** the standalone `scroll-anchor.ts` modifier file
- **Fixed** a `qunit/no-assert-logical-expression` lint error in the card-catalog test

## Problem

When toggling the "Show only" checkbox in the card catalog, the UI exhibited jumpiness caused by a conflict between DOM reordering and the browser's built-in scroll anchoring:

1. **Checking "Show only"**: The focused section gets moved to the top of the list. The browser's scroll anchoring tries to preserve the user's visual scroll position during this DOM reorder, fighting with the intended layout change and causing a visible jump.
2. **Unchecking "Show only"**: All sections are restored back into the DOM. Without explicit scroll management, the user loses their place entirely — the viewport jumps to an unpredictable position instead of staying on the section they were just looking at.

## How this fixes it

The `ScrollToFocusedSection` modifier tracks the previously focused section and uses `requestAnimationFrame` to defer scroll adjustments until *after* the DOM has re-rendered:

- **On focus (check "Show only")**: Waits one frame for the reordered sections to settle, then sets `scrollTop = 0`. This avoids the browser's scroll anchoring from shifting the position back.
- **On unfocus (uncheck "Show only")**: Remembers the previously focused section SID, waits one frame for all sections to be restored, then calls `scrollIntoView` on that section — so the user stays on the section they were viewing.

Deferring to the next animation frame is the key: it lets the DOM update complete before applying scroll corrections, preventing the browser's scroll anchoring from interfering.

## Why inline

The modifier was only used by `SearchContent`. Colocating it removes a single-use standalone file, reduces indirection, and the new name `ScrollToFocusedSection` better describes its purpose.

## Test plan

- [ ] `pnpm lint` passes in `packages/host`
- [ ] Card catalog integration tests pass
- [ ] Manually toggle "Show only" in the card catalog — no jumpiness when focusing/unfocusing a section


habdelra:
to be really clear--the deleted code in test was a lot of AI work in order to stabilize this test. I want to make sure that we are not introducing flakiness back into the test. can you make sure to prompt AI about this.

FadhlanR:
> to be really clear--the deleted code in test was a lot of AI work in order to stabilize this test. I want to make sure that we are not introducing flakiness back into the test. can you make sure to prompt AI about this.

Thanks for the heads up. It recommended using await requestAnimationFrame() to ensure the UI is fully rendered before the assertions run. Please check it again.

habdelra:
> > to be really clear--the deleted code in test was a lot of AI work in order to stabilize this test. I want to make sure that we are not introducing flakiness back into the test. can you make sure to prompt AI about this.
> 
> Thanks for the heads up. It recommended using await requestAnimationFrame() to ensure the UI is fully rendered before the assertions run. Please check it again.

it wasn't just about that--it was also about how you measure--like measuring relative to the scrollable div instead of the entire viewport

FadhlanR:
> > > to be really clear--the deleted code in test was a lot of AI work in order to stabilize this test. I want to make sure that we are not introducing flakiness back into the test. can you make sure to prompt AI about this.
> > 
> > 
> > Thanks for the heads up. It recommended using await requestAnimationFrame() to ensure the UI is fully rendered before the assertions run. Please check it again.
> 
> it wasn't just about that--it was also about how you measure--like measuring relative to the scrollable div instead of the entire viewport

The old test needed relative-to-container measurement because it was asserting the section's visual offset was preserved (a pixel delta). The new behavior is simpler: on check, the section moves to first position with scrollTop=0 (structural/absolute checks, no pixel math). On uncheck, we only verify the section is visible within the container bounds, not that it returned to an exact pixel offset.

habdelra (approved):

PR #4156, opened 2026-03-10
Start icons server in matrix test services startup

lukemelia:
## Summary
- include `start:icons` in `start:services-for-matrix-tests` startup sequence
- keep readiness check on `http://localhost:4206` aligned with services that are actually started


backspace (approved):

habdelra (approved):

IanCal:
Doesn't look like it breaks anything but probably worth removing the explicit start of the icons server in the github action for the matrix tests

PR #4158, opened 2026-03-10
Replace Volta with mise

backspace:
Volta is [explicitly unmaintained](https://github.com/volta-cli/volta/blob/main/README.md?plain=1#L21) and they recommend mise.

This is the first of a few steps:
1. removing Volta
2. bringing scripts and tasks under control with mise
3. consolidating “environment mode” with mise/helper scripts

I [deployed](https://github.com/cardstack/boxel/actions/runs/22956233589) this branch and all was well (after the first one revealed some problems).


backspace:
Actually there are icons errors after deploying, seeing if it’s related to this

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-03-10 17-12-14

backspace:
The environment variables were getting overwritten, that’s fixed now so the icons problem is gone:

e8c0a21ef0e0f5101a0d9e590435203ca7bf96af 2026-03-11 07-45-20

habdelra (approved):

richardhjtan:
In morning, we tested from staging and we got an `Internal server error`  from most of the cards that load the module from either base and catalog. Not sure if it is relevant from this branch to staging changes

Screenshot 2026-03-12 at 10 07 14 PM

backspace:
> In morning, we tested from staging and we got an `Internal server error` from most of the cards that load the module from either base and catalog. Not sure if it is relevant from this branch to staging changes
> 
> Screenshot 2026-03-12 at 10 07 14 PM

hmm worrisome, thanks for letting me know, I’m deploying again and will look more deeply into this.

backspace:
With this deployment I saw no problems, it’s hard to make any definitive conclusion but I think it’s platform instability vs anything particular to mise, which should only make a difference in setup, not once deployed.

PR #4166, opened 2026-03-11
Add OpenRouter realm and automated model sync pipeline

lukemelia:
## Summary

- **Phase 1 — OpenRouter Realm Infrastructure**: Creates a dedicated public-readable realm (`@cardstack/openrouter/`) with the `OpenRouterModel` card type (isolated/embedded/fitted format templates), environment-mapped URLs for dev/staging/production, host config, network service prefix registration, and deployment setup scripts.
- **Phase 2 — Automated Sync Command & Cron**: Adds `SyncOpenRouterModelsCommand` that fetches all models from the OpenRouter API, writes/updates them via batch atomic operations (batches of 50), and marks models no longer in the API as deprecated. Scheduled via daily cron job in worker-manager (4am ET, configurable via `OPENROUTER_SYNC_CRON_SCHEDULE`).

## Linear Issues

Closes CS-10404
Closes CS-10367
Closes CS-9564

## Test plan

- [x] Verify openrouter-realm starts in local dev (`pnpm start:all`) and is accessible at `http://localhost:4201/openrouter/`
- [x] Verify `@cardstack/openrouter/` URL prefix resolves correctly in network service
- [x] Manually trigger sync command and verify OpenRouterModel cards are created in the realm
- [x] Verify OpenRouterModel format templates render correctly (isolated, embedded, fitted)
- [ ] Verify cron job starts in worker-manager with `OPENROUTER_REALM_URL` set
- [ ] Verify cron job is skipped gracefully when `OPENROUTER_REALM_URL` is not set
- [x] Verify worker scripts include openrouter realm URL mappings in dev/staging/production


packages/postgres/migrations/1773265582000_add-openrouter-realm-permissions.js
```diff
@@ -0,0 +1,69 @@
+exports.up = (pgm) => {
+  switch (process.env.REALM_SENTRY_ENVIRONMENT) {
+    case 'staging':
+      pgm.sql(
+        `INSERT INTO realm_user_permissions (realm_url, username, read, write, realm_owner)
+         VALUES
+           ('https://realms-staging.stack.cards/openrouter/', '@openrouter_realm:stack.cards', true, true, true),
```
habdelra, line 7: maybe i'm missing it, don't we also need to update a matrix script to register this realm user?

packages/realm-server/worker-manager.ts
```diff
@@ -580,6 +592,33 @@ function startDailyCreditGrantCron() {
   return job;
 }
 
+function startOpenRouterSyncCron(): CronJob | undefined {
```
habdelra: probably we can abstract this out into a module that is responsible for running cron, and instead you'd register a sync job with it. that way this doesn't become a giant bucket of cron jobs. as we already have the daily credit cron in here too....

lukemelia: I want to push the cron scheduling into userland and the database, but for now I agree extracting a module would be good.

packages/postgres/migrations/1773265582000_add-openrouter-realm-permissions.js
lukemelia: Added, thanks!

habdelra (approved):

PR #4171, opened 2026-03-12
CS-10380: Fetch type picker options from _federated-types endpoint

FadhlanR:
https://github.com/user-attachments/assets/b26f806a-e1d4-47b2-8581-1060732efb0b

## Summary

- Fetch type picker options from the `_federated-types` endpoint instead of deriving them from search results
- Move loading indicators and infinite scroll behavior from the host `TypePicker` into the boxel-ui `Picker` component
- Split type loading into two distinct states: full overlay for initial/search fetches, bottom spinner for load-more pagination
- Implement server-side search and pagination for type picker options (page size: 25)
- Types with the same display name but different code refs are merged into a single picker option

### `packages/boxel-ui/addon/src/components/picker/index.gts`
- Added `isLoading`, `isLoadingMore`, `hasMore`, `onLoadMore` args to Picker
- Built-in `PickerAfterOptions` component with full overlay loading for initial/search and bottom spinner for load-more
- Scroll-based infinite scroll detection with immediate check via `requestAnimationFrame`
- `dropdownClass` adds `--loading` modifier for z-index layering (search input stays above overlay)

### `packages/host/app/components/type-picker/index.gts`
- Now passes `@isLoading`, `@isLoadingMore`, `@hasMore`, `@onLoadMore` directly to Picker

### `packages/host/app/components/card-search/panel.gts`
- Added `fetchTypeSummaries` restartable task calling `realmServer.fetchCardTypeSummaries()`
- Resource-based reactivity: realm URL or search key changes trigger re-fetch automatically
- Split loading into `_isLoadingTypes` (initial/search) and `_isLoadingMoreTypes` (pagination)
- `_typeCodeRefs` map tracks multiple code refs per display name for accurate search filtering

### `packages/host/app/components/card-search/search-bar.gts`
- Added `isLoadingTypes` arg, passed through to TypePicker

### Tests
- Updated type counts and added `waitFor` for paginated type options in operator-mode UI tests

## Test plan

- [ ] Open type picker — verify loading indicator shows during initial fetch
- [ ] Type in type picker search — verify loading overlay covers options area (not search input)
- [ ] Scroll to bottom of type options — verify more types load with bottom spinner
- [ ] Select a realm filter — verify type options refresh from the endpoint
- [ ] Select specific type(s) — verify card search results filter correctly
- [ ] Run `packages/host` integration tests for operator-mode UI


backspace:
I don’t know what best practice is for this kind of thing, I assume pagination is a reason but I find it confusing that I scroll to the bottom and see “WelcomeToBoxel” as the last item, then loading, then it’s still the last item, what were the new ones? And then after enough of that, “YouTube Thumbnail Composer” becomes the last item. Is it possible to trigger the load of them all instead of needing to scroll to the bottom? There may be constraints here I’m unaware of, it just seems confusing.

screencast 2026-03-16 15-27-22

FadhlanR:
@backspace Thanks for the feedback! The confusing behavior you saw was caused by two issues:

1. **Client-side deduplication** was collapsing multiple types with the same display name into one picker option. So a page of 25 raw items from the server might yield far fewer *visible* options after dedup, making it seem like nothing loaded.

2. **No server-side sorting** — the server returned types in an arbitrary order, so after the client sorted alphabetically, newly loaded items could end up scattered throughout the list rather than appended at the bottom.

In the latest commit I've fixed both:

- **Removed deduplication**: Each `code_ref` is now its own picker option (with a tooltip showing the code_ref on hover so users can distinguish same-named types). This means 25 items per page = 25 visible items, always.
- **Added server-side alphabetical sort**: Types are now sorted by `displayName` before pagination, so load-more always appends items that come later alphabetically. No more confusing reordering.

**Why we still need pagination**: The type list can be very large (hundreds of types across multiple realms). Loading them all upfront would mean a slower initial render and a large payload on first open of the type picker. Incremental pagination (25 at a time) keeps the picker responsive. With the server-side sort fix, the UX should now feel natural — scroll down, see 25 more types in alphabetical order.

*) You cannot try this in the preview because we need to deploy the server-side sorting.


https://github.com/user-attachments/assets/70760e17-8d13-4c92-82ae-285ae26a218d

lukemelia:
Minor issue: the dropdown needs some extra bottom padding so that the hover state is not flush against the bottom edge:

image

lukemelia:
Looks like types are not loading as expected from AI Assistant "+" > "Attach a Card" context

lukemelia:
Height of picker is changing when loading spinner appears. It should be stable.

lukemelia:
This should auto-choose "Author" as the type filter: 
image

lukemelia:
Bug:

1. Type "Skill" into type picker search box (list shows types matching "Skill"
2. Check "Skill" (Skill shows as checked at top of picker)
3. Clear type picker search box

Expected behavior: Show the latest paginated load of all types
Actual behavior: shows only types matching "Skill"

lukemelia (changes requested):
Left comments and discussed interactively

FadhlanR:
## Fixes in latest commits

### 1. Simplified loading spinner (picker component)
Replaced the `PickerAfterOptions` / `afterOptionsComponent` pattern with an inline loading spinner rendered directly in the Picker template after the picker-divider. The spinner only displays when the option is the last option and `isLoadingMore` is true. A minimal `PickerLoadingOverlay` is kept only for the initial full-page loading state.

### 2. Fixed types not loading from AI Assistant "Attach a Card"
**Root cause:** When "Attach a Card" is clicked, the `baseFilter` is `{ type: CardDef }`. The `baseFilterCodeRefs` getter was doing an exact match against each type summary's `code_ref`, but since all summaries are specific types (Author, Skill, etc.) — none literally match `CardDef`. This filtered out all types.

**Fix:** Skip client-side type filtering when the only refs in the baseFilter are root types (`CardDef` / `FieldDef`), since all card types inherit from them.

### 3. Auto-select type when baseFilter constrains to a specific type
When the baseFilter constrains to a specific type (e.g., Author for a `linksTo` field), the type picker now auto-selects that type instead of defaulting to "Any Type".

### 4. Root types hidden from type picker
CardDef, FieldDef, and BaseDef are now filtered out from type picker options — users should never see these internal meta types as selectable options.

### 5. "Any Type" disabled when baseFilter constrains to specific type
When the card catalog is opened from a `linksTo` field (e.g., Pet), the "Any Type" option is now visually disabled (grayed out, non-interactive). The constrained type (e.g., Pet) is auto-selected. This prevents users from switching to "Any Type" in a context where the search is already type-constrained.

For root type baseFilters (CardDef — e.g., "Attach a Card"), "Any Type" remains enabled so users can still select specific types to filter results.

### 6. Search result filtering moved to server-side query
Type filtering of search results is now done server-side via `buildSearchQuery` instead of client-side filtering. Selected type IDs are included in the query filter, which avoids SQL conflicts by stripping the baseFilter's type constraint when a type picker selection is active. This also correctly handles subtypes (e.g., BoomPet under Pet).

### 7. Search clearing bug fix — selected types preserved across pages
Directly performing the fetch task in `onTypeSearchChange` ensures clearing the type picker search reliably re-fetches all types. When selected types exist, the fetch now recursively loads additional pages until all selected types are found in the results (or no more pages remain). This prevents selected types from being lost when they fall beyond the first page (25 items).

### 8. "Any Type (N)" total count fix
The total count shown in "Any Type (N)" now uses `meta.page.total` from the API response instead of the client-side `typeFilter.options.length`. This ensures the count reflects all available types, not just the currently loaded page.

### 9. Tests added
- Type picker with baseFilter shows constrained types and disabled "Any Type"
- Auto-select constrained type when baseFilter specifies a specific type
- Attach card from AI assistant: types load, "Any Type" enabled, type selection filters results
- Clearing type picker search restores all types and preserves selection
- Selected types preserved after clearing type search when selection is beyond first page
- Type picker total count reflects API total, not loaded options count

lukemelia:
Needs space between the "Show only" text and the scrollbar, as well as the sort widget and the scrollbar:

image

lukemelia:
I expected to be able to click where it says "Search for a type" and focus the field and type, but I can't.

image

lukemelia:
Horizontal scrollbar should not appear when choosing the "strip" size:

image

lukemelia (approved):

PR #4173, opened 2026-03-12
Add option to delete a realm

jurgenwerk:
https://github.com/user-attachments/assets/7fe70b29-79c6-4f5c-9506-e1ad80db0ced

## Summary
- add a workspace chooser context menu action to delete user-owned realms and show a richer destructive confirmation
- add a realm-server delete endpoint that removes source realms, published copies, related permissions, and soft-deletes claimed domains
- add host acceptance coverage and targeted realm-server endpoint coverage for the delete flow


jurgenwerk:
I had no design for this so I vibed it. I'll show it to the design people and get feedback. In the meanwhile this is ready for review.

habdelra:
are there any odd behaviors if you have an existing chat with AI that has artifacts from a deleted realm?

jurgenwerk:
Yeah the assistant will show errors for attached cards that used to work but are now deleted. Perhaps we can adjust this to say the cards/files are not found

image

habdelra:
@IanCal this is the PR that we can use for removing a realm for teh software factory so that realms become more disposable. probably a good next step is to make a command and skill for this (and prob refactor the menu logic here to call the command)

packages/realm-server/tests/server-endpoints/delete-realm-test.ts
```diff
@@ -0,0 +1,246 @@
...
+    if (response.status !== 201) {
+      throw new Error(
+        `/_create-realm failed: ${JSON.stringify(response.body)}`,
+      );
+    }
+
+    return response.body.data.id as string;
+  }
+
+  test('DELETE /_delete-realm removes a created realm, its published copies, and related domain claims', async function (assert) {
```
habdelra, line 98: i think there is more cleanup. we should delete all items for the deleted realm in the `realm_versions`, `boxel_index`, `boxel_working_index`, and `modules` tables and add assertions our tests for that. Also probably we should delete any pending jobs in the jobs table that are tied to the deleted realm's concurrency group (indexing:<realm_url>)

habdelra: maybe also ask AI to think of any other tables that might need cleanup

backspace: maybe a way to detect this is to dump the database and look for any mentions of the should-have-been-deleted realm name

jurgenwerk:
I got the new design and implemented it: 

image

There has been more cleanup added (db tables, recent files, local storage tokens). This is now ready for another review.

packages/realm-server/handlers/realm-destruction-utils.ts
```diff
@@ -0,0 +1,184 @@
...
+    ` AND status = 'unfulfilled'`,
+  ])) as { id: number }[];
+
+  if (pendingJobs.length > 0) {
+    await query(dbAdapter, [
+      `DELETE FROM job_reservations WHERE job_id IN (${pendingJobs
+        .map(({ id }) => id)
+        .join(', ')})`,
+    ]);
+  }
```
habdelra, line 156: i think you need to clear job_reservations first so that you don't get a foreign key constraint

habdelra: or wait nevermind--thats exactly what you are doing.

packages/realm-server/handlers/realm-destruction-utils.ts
```diff
@@ -0,0 +1,184 @@
...
+  realmURL: string;
+}) {
+  let { dbAdapter, realmURL } = args;
+  let pendingJobs = (await query(dbAdapter, [
+    `SELECT id FROM jobs WHERE concurrency_group =`,
+    param(`indexing:${realmURL}`),
+    ` AND status = 'unfulfilled'`,
+  ])) as { id: number }[];
+
+  if (pendingJobs.length > 0) {
```
habdelra, line 150: so something to think about--this won't work if you are actively indexing a realm that you want to delete--maybe the realm has a timed out job and is  taking super long. you should consider first cancelling all indexing jobs for the realm. we have a realm server handler for cancelling an indexing job whose logic you can use.

habdelra: using that mechanism you can resolve the promises for the indexing in a safe manner as opposed to pulling the rug out from underneath them which might result in more waiting for timeouts at a different layer.

habdelra (approved):
I left a note about how you can cleanly stop any indexing jobs for a realm about to be deleted. but otherwise looks good.

packages/realm-server/handlers/realm-destruction-utils.ts
jurgenwerk: Thanks! Added this

PR #4176, opened 2026-03-12
Fix: Allow duplicate model IDs in AI Assistant model list

lukemelia:
The LLM select menu previously used Record<string, string> keyed by modelId, which prevented showing multiple configurations for the same model (e.g. "high" and "low" thinking variants). Switch to an array of LLMOption objects keyed by ModelConfiguration card ID, and match the active selection by modelId + reasoningEffort.


habdelra (approved):
looks good aside from a prettier nit

PR #4190, opened 2026-03-15
Cs 10383 create listing and make a PR should have a loading user experience

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10383/create-listing-should-have-a-loading-user-experience

## Summary                                                                                                                                        
                                                                                                                                               
  - Add a confirmation modal before creating a listing, allowing users to review target realm, codeRef, and select example instances             
  - Show a loading state in the modal while listing creation and background auto-patching run
  - Navigate to the listing in code mode (isolated preview) immediately after the empty listing is created                                       
  - Modal auto-dismisses after all background work (name, summary, tags, specs, etc.) finishes                                                   
  - Add isListingDef / isListingInstance to @cardstack/runtime-common for reliable listing type detection                                        
  - Change openCardId (singular) to openCardIds (plural) to support multiple example selection    
  
  New Added :
   - New CardInstancePicker component (mirrors RealmPicker/TypePicker pattern)
  - CreateListingModal refactored to own selection as PickerOption[]
  - autoLinkExample fix to respect user's explicit selection                                              
                                                                                                                                                 
 ## Architecture                                                                                                                                   
                                                                                                                                                 
  1) createListingModalPayload lives in OperatorModeStateService                                                                                
  
  - The modal is rendered at the top level in operator-mode/container.gts — it needs to be visible across submodes (code mode, interact mode). The state service is the central place that manages cross-submode UI state (similar to workspaceChooserOpened, aiAssistantOpen). Storing the payload there allows any part of the app to trigger the modal via showCreateListingModal().  (example: menu-items, details panel)                                             
                                                                                                                                               
  2) OpenCreateListingModalCommand exists as a separate command                                                                                 
  
  The "Create Listing" action can be triggered from multiple places:                                                                             
  - Detail panel                                                                                                                                         
  - Menu items (packages/base )                                                                                                    
                                                                                                                                                            
                                                                                                                                               
  ## Flow after clicking "Create"                                                                                                                   
                                                                                                                                               
  1. User clicks Create → createListing ember-concurrency task starts, button shows loading spinner + "Creating..."                              
  2. ListingCreateCommand.execute() → creates an empty listing card in the store, then kicks off backgroundWork (Promise.all of auto-patching name, summary, tags, categories, license, examples, specs) as fire-and-forget                                                                  
  3. Immediate navigation → switches to Code submode, opens the listing .json in preview with "Isolated" format selected. The listing card is  visible behind the modal and updates in real-time as background patches land.                                                                  
  4. Modal stays open → shows "Setting up your {title} listing. This may take a moment..." in the footer while backgroundWork runs             
  5. Auto-dismiss → once all background promises resolve, dismissCreateListingModal() is called and the modal closes. User sees the fully populated listing.                                                                                                                             
                                                                                                                                                                                                                                                                                  
## Demo:


https://github.com/user-attachments/assets/5b91106a-04ab-450b-8a7d-cb2020c0ecfe


lucaslyl:
@codex review

lucaslyl:
@codex review

packages/base/menu-items.ts
```diff
@@ -155,7 +155,7 @@ export function getDefaultCardMenuItems(
     });
     menuItems = [...menuItems, ...getSampleDataMenuItems(card, params)];
     menuItems.push({
-      label: `Create Listing with AI`,
+      label: `Create Listing`,
```
tintinthong, line 158: 👍

tintinthong:
Is there a recursive call somehow? I can't create-listing without it running async forever

lucaslyl:
> Is there a recursive call somehow? I can't create-listing without it running async forever

No, there is no recursive call. The flow is linear — create listing, then await background work, then dismiss modal.                                            
                                                                                                                                                                 
The issue is background work fires multiple LLM requests in parallel (autoPatchName, autoPatchSummary, etc) plus N spec creations, all with no timeout. If any one of them is slow or hangs, the modal waits forever. 

The original **await backgroundWork** was intentionally designed based on the request —it keeps the modal on screen until all the AI auto-patching finishes writing to the listing card. 

Alternative way, I think 
1) either we can remove await — dismiss the modal  immediately, listing fields fill in gradually,
2) Or add a timeout so the modal waits up to a limit then dismisses.

packages/host/app/components/operator-mode/container.gts
```diff
@@ -143,6 +144,7 @@ export default class OperatorModeContainer extends Component<Signature> {
   <template>
     <div class='operator-mode'>
       <ChooseFileModal />
+      <CreateListingModal />
```
tintinthong, line 148: seems like a very similar modal to `ChooseFileModal`

packages/host/app/services/operator-mode-state-service.ts
```diff
@@ -142,6 +148,7 @@ export default class OperatorModeStateService extends Service {
   private moduleInspectorHistory: Record<string, ModuleInspectorView>;
 
   @tracked profileSettingsOpen = false;
+  @tracked createListingModalPayload?: CreateListingModalPayload;
```
tintinthong, line 158: why was this state not INSIDE of the modal

packages/host/app/components/operator-mode/container.gts
lucaslyl: From my understanding, the main reason we can't reuse **ChooseFileModal** is that it's a host-only component that lives in **code submode**. But the listing creation gets triggered from a **menu item** in packages/base, which runs in the realm's headless Chrome process — it has no way to reach any host UI       component directly. The only bridge i can see is using the command pattern (openlistingmodal command).

The LHS detail panel "create-lisitng" dont have this issue because it was under host packages

packages/host/app/services/operator-mode-state-service.ts
lucaslyl: As for why createListingModalPayload is stored in OperatorModeStateService — the modal doesn't exist until the payload is set, so the command needs somewhere that's always alive to drop off the data.  For now i can see the OperatorModeStateService fits because it persists across all modes, unlike CreateFileModal which only lives in code submode. The payload also carries the context (targetRealm, codeRef, openCardIds) that the modal needs to know what to display, since that info has to travel from the base package through the command pattern to reach the host side.   Not sure if my understanding is precise or not

lucaslyl:
@codex feedback

packages/host/app/components/operator-mode/create-listing-modal.gts
```diff
@@ -0,0 +1,440 @@
...
+  <template>
+    {{#if this.isModalOpen}}
+      <ModalContainer
+        class='create-listing-modal'
+        @cardContainerClass='create-listing'
+        @title='Create Listing'
+        @size='small'
+        @isOpen={{this.isModalOpen}}
+        @onClose={{this.onClose}}
+        data-test-create-listing-modal
```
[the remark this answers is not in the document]
lucaslyl: I agree with this. But follow the request, we should allow users to close the modal instead of forcing them to wait for the listing creation to finish, as the background processing can take quite a long time.

tintinthong (approved):

PR #4191, opened 2026-03-16
introduce-realm-enabled-tests

tintinthong:
# Live Tests: Co-locate QUnit tests inside realm cards

I decided to call this live-test. But I'm not so commited to the name. Open to suggestions. I call it live-test bcos these are test loaded from the realm

## Exercised in external catalog 

Look at this PR https://github.com/cardstack/boxel-catalog/pull/244

Screenshot 2026-03-24 at 17 18 24

## What this does

This adds a new test mode called **live tests** that lets you write QUnit tests directly inside a realm (e.g. alongside a card definition in `catalog-realm/`), and run them against a live running realm server — without having to add them to the host test bundle.

Previously, all host tests were compiled into the Ember test bundle at build time. Live tests are different: the realm *serves* the test files over HTTP, and the host loads and executes them at runtime.

## What changed

- **`test-helper.js`** — the existing test entry point now checks for a `?liveTest` query param. If present it runs the live-test path; otherwise it runs the normal ember-exam path as before. A single check prevents double QUnit initialisation (two competing app instances).

- **`live-test.js`** — contains `loadRealmTests(application)`, which is the async function that:
  - fetches the realm index to discover `.gts` test files
  - loads each file through the realm's own module loader (not the Ember bundle)
  - calls `runTests()` on each to register QUnit modules

- **`helpers/setup-qunit.js`** — shared QUnit bootstrap (maxDepth, test waiters, qunit-dom setup, autostart flag) used by both live and normal test paths.

- **`testem-live.js`** — Testem config that drives Chrome against `tests/index.html?liveTest=true&realmURL=http://localhost:4201/catalog/`.

- **`scripts/live-test-wait-for-servers.sh`** — waits for all required services then runs `ember test --config-file testem-live.js --path ./dist`.

- **`package.json`** — adds `test:live` script.

- **`ci-host.yaml`** — adds a `live-test` CI job (non-sharded, runs with catalog realm, no `SKIP_CATALOG`).

- **`catalog-realm/sample-command-card.gts`** — example card with a co-located `runTests()` export demonstrating the pattern.

## How tests are discovered in a co-located module

A `.gts` file in the realm can export a `runTests()` function alongside its card definition:

```ts
// catalog-realm/my-card.gts
export class MyCard extends CardDef { ... }

export function runTests() {
  module('Catalog | MyCard', function (hooks) {
    test('it works', async function (assert) { ... });
  });
}
```

At runtime, `loadRealmTests` fetches the realm index, finds all `.gts` files, imports each one through the realm loader, and calls `runTests()` if it exists. Files without `runTests()` are skipped silently.

## The loader boundary

There are two separate module loaders at play:

1. **Host bundle loader** — the normal Ember/webpack compiled bundle. All standard host tests run here.
2. **Realm loader** — a second Ember app instance (`application.buildInstance(...)`) that fetches modules over HTTP from the realm server. Live test files are imported through this loader.

The realm loader needs the test helpers (`@cardstack/host/tests/helpers`, mock-matrix, etc.) to be shimmed into it explicitly — otherwise realm test files can't import them. `loadRealmTests` does this before importing any test modules.

## Things to know

- **The realm must be running** to run live tests. The test files live on the realm server — they are fetched at runtime, not compiled into the bundle. If the realm is down, no tests load.
- **Tests must be in a publicly readable realm.** The realm loader fetches modules without Matrix auth. For CI, we cover the catalog realm (`/catalog/`) which has `'*': read=true` permissions.
- **For CI**, the `live-test` job starts the catalog realm (no `SKIP_CATALOG`) and runs `pnpm test:live` against it. `sample-command-card.gts` in `catalog-realm/` is the first test covered.

## Future work

- **Immediate next thing: Move listing-create test into the gts files** 
- **Make test helpers available in Code Mode.** Right now the helpers are shimmed from within the host test bundle. The goal is to make them available inside Code Mode so realm tests can be authored and run directly in the browser without needing the host test bundle.
- **Use an iframe when running in Code Mode.** The current implementation runs live tests in the same browser context as the host test page, which is sufficient for Testem/CI. When integrating into Code Mode, the test runner should be mounted as an `<iframe>` — giving it a fully isolated JS heap, its own Ember app instance, its own SQLite WASM, and no shared state with the running host app. Testem is not involved in the Code Mode case; QUnit renders pass/fail directly inside the iframe's own UI.

## How to run locally

```sh
# Manual browser (services must be running)
open http://localhost:4200/tests?liveTest=true&realmURL=http://localhost:4201/catalog/

# Testem (services must be running, dist must be built)
cd packages/host
pnpm build
ember test --config-file testem-live.js --path ./dist
```


habdelra:
So i think your PR description is a little ambiguous, when we say tests are run against a live realm server, we mean a realm that is running in the browser correct like the current host tests? it's not actually a server in the classic sense, right?

tintinthong:
> So i think your PR description is a little ambiguous, when we say tests are run against a live realm server, we mean a realm that is running in the browser correct like the current host tests? it's not actually a server in the classic sense, right?

confirmed. Everything is made similar to host tests. The only difference is that we need to dynamically load up test specified in realm modules.

habdelra (approved):

PR #4196, opened 2026-03-17
Remove Applisting Type form listingType

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10435/default-every-listing-creation-to-cardlisting-except-fieldlisting

Remove app listing type guessing from ListingCreateCommand.  Users can manually change the listing type from cardListing to AppListing after creation.


tintinthong:
Dono what the motivation of this work is. I understand we had issues that everything was AppListing. But, now you are removing all the different listing types which makes me confused

lucaslyl:
> Dono what the motivation of this work is. I understand we had issues that everything was AppListing. But, now you are removing all the different listing types which makes me confused

I reverted the code. Now we just eliminate AppListing checking during the listing creation process.

tintinthong (approved):

PR #4201, opened 2026-03-18
Vitest for realm server tests

IanCal:
First pass at converting out realm server tests to use vitest rather than qunit. This gets us a variety of benefits with usability, and for CI supports sharding. Over time we can move to helpers it provides to clean up the tests and make use of parallelism but this is a first (codemod automated) conversion).


habdelra:
Abandoned PR

PR #4206, opened 2026-03-18
Add @cardstack/skills universal reference

backspace:
This follows the `@cardstack/catalog` pattern for skills. I deployed it to staging and was
able to get the bot to use skills, so all seemed well.

If/when this is merged, I’ll run the migration script to rewrite imports/references in cards etc.


burieberry (approved):
Haven't tried running but the code looks good to me

PR #4209, opened 2026-03-18
Move essential skill and theme files to base

burieberry:
This will be a 3 step process:
1- Copy files to base (this PR)
2- Update references in other repos
3- Remove files from catalog

I'll create a separate ticket to make the skill card improvements suggested below.

### DIFF
--- a/packages/base/Theme/boxel-brand-guide.json
+++ b/packages/base/Theme/boxel-brand-guide.json
@@ -1,0 +1,247 @@
+{
+  "data": {
+    "meta": {
+      "adoptsFrom": {
+        "name": "default",
+        "module": "https://cardstack.com/base/brand-guide"
+      }
+    },
+    "type": "card",
+    "attributes": {
+      "spacing": "0.25rem",
+      "version": "7.0",
+      "cardInfo": {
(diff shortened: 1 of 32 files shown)


habdelra (approved):

habdelra:
i think we need to also update the skills cards in the skills realm repo as well. maybe that needs to happen first in order for these tests to pass. seems like a chicken and egg problem :chicken: :egg:

burieberry:
> i think we need to also update the skills cards in the skills realm repo as well. maybe that needs to happen first in order for these tests to pass. seems like a chicken and egg problem 🐔 🥚

I haven't yet removed any of the files from catalog yet, though. Nothing should've changed for them. I can't link to the new base files before this PR is merged in.

habdelra:
> > i think we need to also update the skills cards in the skills realm repo as well. maybe that needs to happen first in order for these tests to pass. seems like a chicken and egg problem 🐔 🥚
> 
> I haven't yet removed any of the files from catalog yet, though. Nothing should've changed for them. I can't link to the new base files before this PR is merged in.

i think its' fine to merge with teh test failures--its not like we are making it any worse than it is :grinning:

PR #4226, opened 2026-03-20
ci: Add workflow for weekday PR review summaries

backspace:
It’s scheduled to run at 9am ET, regardless of DST, which in effect means running at both possible hours and ignoring the “wrong” run, because Actions cron is UTC-only.

Here it is working when I had it posting for pushes to this PR, since removed:

#🚨-alarms-draft-internal | Cardstack - Discord 2026-03-20 12-53-03

For that I had it post to the draft alarms channel but have since changed the webhook to post to `#general-internal`.

It only looks at this repository at the moment, I’m not sure how complex it would be to consider others, but we can refine as needed.

### DIFF
--- a/.github/workflows/pr-review-reminder.yml
+++ b/.github/workflows/pr-review-reminder.yml
@@ -26,10 +26,13 @@ jobs:
           set -euo pipefail
 
           # Only run at 9 AM ET — skip the trigger that doesn't match current DST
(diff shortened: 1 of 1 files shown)


habdelra:
nice!

habdelra (approved):

backspace:
This is the kind of thing that will probably need iteration, I added a commit to address Copilot’s feedback, will revisit as needed.

PR #4228, opened 2026-03-20
Update start-server-and-test to clean up dev-all

backspace:


backspace:
Claude said the new version of `start-server-and-test` supported multiple phases, which would help clean up `dev-all`, but it doesn’t! I did look at the releases on Github, no changelog, but I didn’t see any indication of the multiple phases. So this was a red herring, closing.

PR #4237, opened 2026-03-24
Use dynamic software factory harness ports

habdelra:
## Summary
- lay the groundwork for `CS-10419` by making the `software-factory` Playwright harness safe to run alongside other harness instances without port collisions
- switch the harness stack to dynamic port assignment throughout the isolated realm flow instead of assuming canonical local dev ports
- align harness port lifetime with the Playwright `testWorker`, so each worker gets one stable set of harness ports for the duration of the worker even though realm-server and worker-manager still restart between tests
- align prerender lifetime with the Playwright `testWorker`, so restarted realm stacks can reuse the same prerender process and avoid repeated standby warmup
- run the `software-factory` Playwright suite with `workers=2` as proof that multiple harnesses can execute concurrently without interfering with each other
- reduce harness log noise by treating expected standby startup misses during realm-server turnover as normal readiness behavior instead of noisy `502`/`503` errors
- trim one redundant isolated boot in `factory-bootstrap.spec.ts` by combining the bootstrap creation and idempotency assertions into a single realm run

## Why
This PR is part of the groundwork for `CS-10419`. To support multiple test harnesses running side by side, the harness cannot depend on fixed ports or on per-test port churn that makes URLs unstable inside a worker. The harness now allocates ports dynamically across the stack, keeps those URLs stable for the lifetime of each Playwright `testWorker`, reuses prerender across tests within that worker, and validates the design by running the browser suite with `workers=2`.

## Testing
- `pnpm --dir packages/software-factory exec playwright test tests/factory-bootstrap.spec.ts tests/factory-target-realm.spec.ts --workers=2 --reporter=line`
- `pnpm --dir packages/software-factory exec playwright test --workers=2 --reporter=line`


packages/realm-server/main.ts
```diff
@@ -35,6 +37,23 @@ import {
...
+
+  mkdirSync(dirname(runtimeMetadataFile), { recursive: true });
+  let tempFile = join(
+    dirname(runtimeMetadataFile),
+    `.realm-server.${process.pid}.${Date.now()}.tmp`,
+  );
+  writeFileSync(tempFile, JSON.stringify(payload, null, 2));
+  renameSync(tempFile, runtimeMetadataFile);
+}
+
```
habdelra, line 45: This is duplicated in the worker manager, can we make this a util function

packages/software-factory/src/cli/cache-realm.ts
```diff
@@ -39,6 +39,7 @@ async function main(): Promise<void> {
     );
   }
   console.log(JSON.stringify(payload, null, 2));
+  process.exit(0);
```
habdelra: will process.exit here end the process before console.log above has had a chance to log?

packages/software-factory/src/cli/serve-realm.ts
```diff
@@ -40,6 +42,7 @@ async function main(): Promise<void> {
   console.log(JSON.stringify(payload, null, 2));
 
   let cleanExit = false;
+  let keepAlive = setInterval(() => {}, 60_000);
```
habdelra: shoud this be unrefed so it doesn't block graceful text exit?

packages/software-factory/src/cli/serve-support.ts
```diff
@@ -21,14 +21,20 @@ async function main(): Promise<void> {
   writeSupportMetadata(payload);
 
   console.log(JSON.stringify(payload, null, 2));
+  let keepAlive = setInterval(() => {}, 60_000);
```
habdelra: should this be unrefed for graceful test exit?

packages/realm-server/main.ts
habdelra: Pulled this into a shared helper in [`packages/realm-server/lib/runtime-metadata-file.ts`](/home/hassan/codez/boxel/packages/realm-server/lib/runtime-metadata-file.ts) and switched both `main.ts` and `worker-manager.ts` over to it in `30d4eee62b`.

packages/software-factory/src/cli/cache-realm.ts
habdelra: Changed this in `30d4eee62b`. I removed the explicit success `process.exit(0)` and only set `process.exitCode = 1` on failure, so stdout can flush naturally after the `console.log`.

packages/software-factory/src/cli/serve-realm.ts
habdelra: Changed direction here in `30d4eee62b`. Instead of keeping an interval alive and worrying about `unref()`, `serve-realm` now waits on a signal-driven promise and shuts down cleanly on `SIGINT`/`SIGTERM`. That keeps the wrapper alive intentionally while avoiding the dummy interval entirely.

packages/software-factory/src/cli/serve-support.ts
habdelra: Same fix here in `30d4eee62b`: I removed the dummy keepalive interval and changed `serve-support` to wait on a signal-driven promise instead. So there is nothing left to `unref()`, and shutdown is now explicit and graceful.

packages/software-factory/src/harness/support-services.ts
```diff
@@ -0,0 +1,418 @@
...
+        logs = `${logs}${String(chunk)}`.slice(-20_000);
+      });
+
+      await waitUntil(
+        async () => {
+          if (child.exitCode !== null) {
+            throw new Error(
+              `host app exited early with code ${child.exitCode}\n${logs}`,
+            );
+          }
```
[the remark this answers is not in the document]
habdelra: Addressed in `30d4eee62b`. `ensureHostReady()` now treats early fallback-host exits that look like `EADDRINUSE` / `address already in use` as contention, and falls back to continuing to wait for `DEFAULT_HOST_URL` instead of failing immediately. That keeps concurrent harness startups from tripping over each other if one of them wins the host bind race.

packages/realm-server/middleware/index.ts
```diff
@@ -103,6 +109,10 @@ export function ecsMetadata(ctxt: Koa.Context, next: Koa.Next) {
 }
 
 export function fullRequestURL(ctxt: Koa.Context): URL {
+  let forwardedURL = ctxt.req.headers['x-boxel-forwarded-url'];
+  if (typeof forwardedURL === 'string' && forwardedURL.trim() !== '') {
+    return new URL(forwardedURL);
+  }
   let protocol =
     ctxt.req.headers['x-forwarded-proto'] === 'https' ? 'https' : 'http';
   return new URL(`${protocol}://${ctxt.req.headers.host}${ctxt.req.url}`);
```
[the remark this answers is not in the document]
habdelra: Addressed in `30d4eee62b`. `fullRequestURL()` now only honors `x-boxel-forwarded-url` when `BOXEL_TRUST_FORWARDED_URL=true`, the request comes from loopback, and the forwarded path/query matches the computed request. The software-factory compat proxy enables that env var explicitly for its local harness realm-server path.

packages/software-factory/tests/fixtures.ts
```diff
@@ -113,11 +135,68 @@ async function waitForPortFree(
...
+    let candidate: TestWorkerPortSet = {
+      compatRealmServerPort: blockStart,
+      realmServerPort: blockStart + 1,
+      workerManagerPort: blockStart + 2,
+      prerenderPort: blockStart + 3,
+    };
+    let ports = Object.values(candidate);
+    if (
+      (await Promise.all(ports.map((port) => isPortFree(port)))).every(Boolean)
+    ) {
```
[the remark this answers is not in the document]
habdelra: Addressed in `30d4eee62b`. I did not go as far as keeping reservation sockets open, but I did add a per-process run offset on top of the worker index when choosing the stable port block. That means concurrent Playwright runs with the same worker index no longer all probe the same block first, which materially reduces cross-run `EADDRINUSE` races while keeping the per-worker URL stability we want.

packages/postgres/pg-adapter.ts
```diff
@@ -31,6 +31,16 @@ function config() {
 
 type Config = ReturnType<typeof config>;
 
+function configuredPoolMax(): number | undefined {
+  let rawValue = process.env.PG_POOL_MAX ?? process.env.PGPOOLMAX;
```
backspace: Is there a reason to have two names for this?

backspace (approved):

packages/postgres/pg-adapter.ts
habdelra: Removed the duplicate name in `8f45d6fafa`. I checked the actual `pg` source and neither `PG_POOL_MAX` nor `PGPOOLMAX` is a built-in `pg` env var, so this was our own adapter aliasing rather than compatibility with libpq/node-postgres. I standardized on `PG_POOL_MAX` and dropped `PGPOOLMAX`.

PR #4252, opened 2026-03-26
lazily shim live-test dependencies

tintinthong:


packages/host/app/lib/externals.ts
```diff
@@ -155,64 +143,73 @@ export function shimExternals(virtualNetwork: VirtualNetwork) {
...
   );
 
-  // Always stub host test helpers here — live-test.js shimModule() on the
-  // realm loader overrides these with real implementations at test time.
+  if (inLiveTestMode) {
+    // Keep test helpers out of the normal app startup path.
+    virtualNetwork.shimAsyncModule({
+      id: '@ember/test-helpers',
+      resolve: () => import('@ember/test-helpers'),
+    });
```
habdelra, line 204: this is async, but this function is not. are you gonna have a race condition where you are expecting these modules to be available and they are not?

habdelra:
Abandoned PR

PR #4255, opened 2026-03-26
Hosted site rendering jankiness fix

burieberry:
It looks like we added a reset for tests that clears ssr-injected scoped styles at every construction, but it's running outside of tests, too. `resetState()` calls `clearInjectedScopedCSS()` which seems to be the cause of missing css in the template rendered via ssr.

### DIFF
--- a/packages/host/app/services/loader-service.ts
+++ b/packages/host/app/services/loader-service.ts
@@ -2,6 +2,7 @@ import { registerDestructor } from '@ember/destroyable';
 import type Owner from '@ember/owner';
 import Service, { service } from '@ember/service';
 
+import { isTesting } from '@embroider/macros';
 import { tracked } from '@glimmer/tracking';
 
 import type { FetcherMiddlewareHandler } from '@cardstack/runtime-common';
@@ -41,8 +42,10 @@ export default class LoaderService extends Service {
   constructor(owner: Owner) {
     super(owner);
     this.reset.register(this);
-    // this clears the fetch cache in between tests
-    this.resetState();
+    if (isTesting()) {
+      // clears the fetch cache and SSR-injected scoped styles in between tests
+      this.resetState();
+    }
     registerDestructor(this, () => this.resetState());
   }
 


packages/host/app/services/loader-service.ts
burieberry, line 131: LoaderService's `resetState()` calls `clearSessionCaches()` which does these things. Do we actually need any of them running each time at construction time outside of tests? Want to make sure this change I made is ok @habdelra

packages/host/app/services/loader-service.ts
```diff
@@ -41,8 +42,10 @@ export default class LoaderService extends Service {
   constructor(owner: Owner) {
     super(owner);
     this.reset.register(this);
-    // this clears the fetch cache in between tests
-    this.resetState();
```
burieberry, line 45: this comment seems to indicate that this code was only meant for tests?

backspace: eep, agreed

backspace (approved):
Thanks for figuring this out, I’d like to find a way we can detect this type of failure via the alerts, but I think it’s better to have it fixed

packages/host/app/services/loader-service.ts
habdelra: so the CSS in the tests builds up and is a massive cause of memory leaks. like we end up with millions of DOM elements by the end of the tests from CSS. we defintely want this logic running in the tests. but i'm ok with removing them outside of the tests. probably more long term we should think of a way to clean up the CSS that is no longer being used, it is a memory leak.

habdelra (approved):

packages/host/app/services/loader-service.ts
burieberry: is there a ticket for it? seems like it could be part of the memory use increase, the more the app is used

backspace: I thought there was a long-standing issue about this but haven’t been able to find it. But I think this will become moot if/when we move to Vite, because it’s `style-loader` Webpack plugin that inserts this CSS, and hopefully in Vite-land things would be more sensible. But Ed would probably know better about that

habdelra: yeah there is a general memory issue ticket that I've made child tickets off of https://linear.app/cardstack/issue/CS-10402/improve-browser-memory-footprint

PR #4257, opened 2026-03-26
Do not show loading screen on hosted site

burieberry:
- Do not show `Loading...` screen on hosted card
- It will remain in operator-mode as usual
- Fix for prerendered card scroll before css loads

### DIFF
--- a/packages/matrix/tests/host-mode.spec.ts
+++ b/packages/matrix/tests/host-mode.spec.ts
@@ -257,7 +257,7 @@ test.describe('Host mode', () => {
   }) => {
     await page.goto(publishedWhitePaperCardURL);
     await page.locator('[data-test-white-paper]').waitFor();
-    await page.locator('[data-test-host-mode-card-loaded').waitFor();
+    await page.locator('[data-test-host-mode-card-loaded]').waitFor();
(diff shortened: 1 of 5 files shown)


packages/matrix/tests/host-mode.spec.ts
```diff
@@ -257,11 +257,7 @@ test.describe('Host mode', () => {
   }) => {
     await page.goto(publishedWhitePaperCardURL);
     await page.locator('[data-test-white-paper]').waitFor();
-    await waitUntil(
-      async () =>
-        (await page.locator('[data-test-host-loading]').count()) === 0,
-    );
-
+    await page.locator('[data-test-host-mode-card-loaded').waitFor();
```
backspace: Adding the missing `]` here might make it pass?

burieberry: Ah good catch. I had trouble getting playwright tests to run locally

backspace: eep what was happening? I have trouble with many things but let me know if the mise changes seem related?

burieberry: not quite sure

backspace (approved):

PR #4261, opened 2026-03-27
Fix Ask/Act toggle missing default selection on new session

jurgenwerk:
When opening starting a new session in the AI assistant, the toggle is sometimes untoggled: 

image

## Summary
- Normalize `@selected` in `LLMModeToggle` so the toggle always defaults to "Ask" when the value is not explicitly `"act"`
- Prevents a blank toggle during transient room-resource states introduced by the memory leak cleanup teardown path (`f3fb792`)

## Test plan
- [ ] Open AI assistant and create a new session — toggle should show "Ask" selected
- [ ] Switch between sessions — toggle should always have a selection
- [ ] Existing Ask/Act toggle tests should pass

### DIFF
--- a/packages/host/app/components/ai-assistant/llm-mode-toggle.gts
+++ b/packages/host/app/components/ai-assistant/llm-mode-toggle.gts
@@ -20,13 +20,17 @@ interface Signature {
 }
 
 export default class LLMModeToggle extends Component<Signature> {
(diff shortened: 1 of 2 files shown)


backspace (approved):
This fixes the problem, but is there any possibility of finding out what caused it and adding a test for it?

PR #4263, opened 2026-03-27
Fix recent cards service

tintinthong:


tintinthong:
I can no longer reproduce this bug on `main`

PR #4264, opened 2026-03-27
Fix prefix card id resolution

tintinthong:
```
Seeing issues http://localhost:4201/catalog/CardListing/Spec/5ae71ae1
```

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -1433,7 +1433,7 @@ export default class StoreService extends Service implements StoreInterface {
         }
         id = remoteId;
       }
-      let url = id; // after this point we know we are dealing with a remote id, e.g. url
+      let url = cardIdToURL(id).href; // resolve prefix-form IDs (e.g. @cardstack/catalog/...) to HTTP URLs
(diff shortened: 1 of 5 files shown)


packages/host/app/services/store.ts
```diff
@@ -1433,7 +1433,7 @@ export default class StoreService extends Service implements StoreInterface {
         }
         id = remoteId;
       }
-      let url = id; // after this point we know we are dealing with a remote id, e.g. url
+      let url = cardIdToURL(id).href; // resolve prefix-form IDs (e.g. @cardstack/catalog/...) to HTTP URLs
```
habdelra, line 1436: this is fixed in https://github.com/cardstack/boxel/pull/4262

packages/runtime-common/index.ts
```diff
@@ -782,7 +783,7 @@ export function unixTime(epochTimeMs: number) {
 }
 
 export function isLocalId(id: string) {
-  return !id.startsWith('http');
+  return !id.startsWith('http') && !isRegisteredPrefix(id);
```
habdelra, line 786: this is fixed in https://github.com/cardstack/boxel/pull/4262

PR #4270, opened 2026-03-29
 Hide "Make a PR" menu item when user lacks write permission

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10358/do-not-show-make-a-pr-when-user-dont-have-the-permission

Small changes:
The change is straightforward — it adds an early return in the listing menu logic when canEdit is false, preventing the "Make a PR" option from  appearing for users without write permission.

### DIFF
--- a/packages/catalog-realm/catalog-app/listing/listing.gts
+++ b/packages/catalog-realm/catalog-app/listing/listing.gts
@@ -666,6 +666,9 @@ export class Listing extends CardDef {
     if (params.menuContext !== 'interact') {
       return;
     }
+    if (!params.canEdit) {
+      return;
+    }
     if (!this[realmURL]?.href) {
       return;
     }


lucaslyl:
@codex review

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -666,6 +666,9 @@ export class Listing extends CardDef {
     if (params.menuContext !== 'interact') {
       return;
     }
+    if (!params.canEdit) {
+      return;
+    }
```
tintinthong, line 671: This shud also be covered for the playground example correct?

lucaslyl: yes, but that is handle on another PR, https://github.com/cardstack/boxel/pull/4271

tintinthong (approved):

PR #4271, opened 2026-03-29
Make "Generate Example", "Update Specs", and "Make a PR" menu items available in all menu contexts

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10510/make-a-pr-update-specs-generateexample-should-able-to-show-in-code

1.   Make "Generate Example", "Update Specs", and "Make a PR" menu items available in all menu contexts (interact, code-mode-preview,
2.   code-mode-playground) instead of restricting them to interact mode only. Each menu item now checks canEdit permission to hide when the user lacks  write access. Also unified the guard pattern across all three methods for consistency.

Demo:

https://github.com/user-attachments/assets/2bf15697-8d66-457b-9fbc-e9ddf65ce8d1

### DIFF
--- a/packages/catalog-realm/catalog-app/listing/listing.gts
+++ b/packages/catalog-realm/catalog-app/listing/listing.gts
@@ -642,9 +642,9 @@ export class Listing extends CardDef {
     let menuItems = super
       [getMenuItems](params)
       .filter((item) => item.label?.toLowerCase() !== 'create listing');
-    const extra = this.getGenerateExampleMenuItem(params);
-    if (extra) {
-      menuItems = [...menuItems, extra];
+    const generateExample = this.getGenerateExampleMenuItem(params);
+    if (generateExample) {
+      menuItems = [...menuItems, generateExample];
     }
     const updateSpecs = this.getUpdateSpecsMenuItem(params);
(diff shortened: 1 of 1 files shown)


lucaslyl:
@codex review

packages/catalog-realm/catalog-app/listing/listing.gts
```diff
@@ -643,41 +642,35 @@ export class Listing extends CardDef {
...
-      if (updateSpecs) {
-        menuItems = [...menuItems, updateSpecs];
-      }
-      const createPRMenuItem = this.getCreatePRMenuItem(params);
-      if (createPRMenuItem) {
-        menuItems = [...menuItems, createPRMenuItem];
-      }
+    const extra = this.getGenerateExampleMenuItem(params);
+    if (extra) {
+      menuItems = [...menuItems, extra];
```
tintinthong: I think "extra" is the wrong name right

lucaslyl: yes agreed,I think we could go with a more generic name for this.

tintinthong (approved):

PR #4284, opened 2026-03-31
fix: unsubscribe file watcher before removing realm files during deletion

jurgenwerk:
## Summary
- Fixes a race condition where deleting a realm crashes the server (CS-10555)
- When a realm is deleted, `destroyMountedRealm` removes files from disk, which triggers the file watcher to fire removal events. These events try to re-index the deleted files, which calls `getRealmOwnerUsername()` → `fetchRealmPermissions()`, but the permissions have already been removed from the database. This causes an unhandled `Cannot determine realm owner` error that crashes the entire server process.
- The fix calls `realm.unsubscribe()` in both `destroyMountedRealm` and `removeMountedRealm` before any files are removed, preventing the file watcher from reacting to the deletions.

## Test plan
- [x] Added test verifying `unsubscribe` is called on the realm during deletion
- [x] All existing delete-realm tests pass (5/5)

### DIFF
--- a/packages/realm-server/handlers/realm-destruction-utils.ts
+++ b/packages/realm-server/handlers/realm-destruction-utils.ts
@@ -54,6 +54,8 @@ export async function removeMountedRealm(args: {
   let { realm, realmPath, realms, virtualNetwork } = args;
   let cleanupError: Error | undefined;
 
+  realm.unsubscribe();
+
   try {
     let allFilePaths = collectAllFilePaths(realmPath);
     if (allFilePaths.length > 0) {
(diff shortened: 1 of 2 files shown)


jurgenwerk:
@habdelra I wasn't able to reproduce this locally nor in staging or production. But Claude's idea of the missing `unsubscribe` makes sense. Since you are using linux perhaps there is a difference in how the file watcher works compared to mac or deployed environment. Could you test this branch locally and see if it fixes the bug you saw?

habdelra (approved):
Yup, this works in my local dev now--thanks!

PR #4286, opened 2026-03-31
Fix 3dot menu delete regression

tintinthong:
# File Tree Delete Dropdown Regression

## The Issue

Commit `6bea2ee2e3` refactored the file tree context menu to use a **single shared `BoxelDropdown`** across all file rows (to avoid memory leaks from per-file dropdown instances).

The new design separates the **visible trigger** (the 3-dot `ContextButton`) from the **dropdown anchor** (a hidden button):

```
[file-row]  [ContextButton .file-menu-trigger]  ← user clicks this
...
[button .file-tree-menu-anchor]                 ← ember-basic-dropdown anchors to THIS
```

The anchor is styled to be invisible:

```css
.file-tree-menu-anchor {
  position: fixed;
  width: 0;
  height: 0;
  opacity: 0;
  pointer-events: none;
  /* ⚠️ no top / left set */
}
```

`ember-basic-dropdown` positions its content by calling `getBoundingClientRect()` on the trigger element. Since the anchor has no `top`/`left`, it sits at its natural DOM position. The dropdown **renders in the DOM but appears offscreen**, hidden behind other UI.

### Why direct DOM manipulation doesn't work

An obvious fix is to set the anchor's inline styles via `anchorEl.style.top = ...` before calling `open()`. However, calling `this.dropdownApi?.actions.open(e)` sets tracked state on the BasicDropdown component, which triggers a Glimmer re-render. During that re-render, Glimmer reconciles the anchor button's DOM attributes against the template — and since the template declares no `style` attribute on the anchor, Glimmer **strips the inline styles we just set**. By the time `ember-basic-dropdown` calls `reposition()` after render, the anchor is back at its default position.

## The Fix

Use a **`@tracked` property** for the anchor's style string so that Glimmer includes it in the re-render instead of removing it:

```typescript
@tracked private anchorStyle = '';
```

In the template, bind it to the anchor button:

```hbs
<button
  class='file-tree-menu-anchor'
  style={{this.anchorStyle}}
  {{bindings}}
></button>
```

In `openFileMenu`, set the tracked property before opening the dropdown:

```typescript
@action
private openFileMenu(entryPath: LocalPath, e: MouseEvent) {
  // ...
  this.menuEntryPath = entryPath;

  // ✅ Set via tracked property — survives Glimmer re-render
  const triggerRect = (e.currentTarget as HTMLElement).getBoundingClientRect();
  this.anchorStyle = `top: ${triggerRect.top}px; left: ${triggerRect.left}px; width: ${triggerRect.width}px; height: ${triggerRect.height}px;`;

  this.dropdownApi?.actions.open(e);
}
```

Now the re-render triggered by `open()` applies `anchorStyle` to the anchor, and when `ember-basic-dropdown` calls `getBoundingClientRect()` on the anchor after render, it gets the correct coordinates of the 3-dot button.

Reset `anchorStyle` when the menu closes:

```typescript
@action
private clearFileMenu() {
  this.menuEntryPath = undefined;
  this.anchorStyle = '';
}
```

## Auto-close Timer

The menu also auto-closes after **800 ms** if the user does not hover over it. Hovering pauses the timer; moving the mouse away restarts it.

```typescript
private startCloseMenuTimer() {
  clearTimeout(this.closeMenuTimer);
  this.closeMenuTimer = setTimeout(() => {
    this.dropdownApi?.actions.close();
  }, 800);
}

private cancelCloseMenuTimer() {
  clearTimeout(this.closeMenuTimer);
}
```

The timer starts when the menu opens and is wired to the menu content via `mouseenter`/`mouseleave`:

```hbs
<Menu
  {{on 'mouseenter' this.cancelCloseMenuTimer}}
  {{on 'mouseleave' this.startCloseMenuTimer}}
  ...
/>
```

### DIFF
--- a/packages/host/app/components/editor/directory.gts
+++ b/packages/host/app/components/editor/directory.gts
@@ -7,6 +7,7 @@ import Component from '@glimmer/component';
 
 import { tracked } from '@glimmer/tracking';
 
+import onClickOutside from 'ember-click-outside/modifiers/on-click-outside';
 import { velcro } from 'ember-velcro';
 import { TrackedArray } from 'tracked-built-ins';
 
@@ -107,6 +108,7 @@ export default class Directory extends Component<Args> {
       <div
         class='file-tree-context-menu'
         {{velcro this.menuTriggerEl placement='bottom-start' strategy='fixed'}}
+        {{onClickOutside this.closeMenu exceptSelector='.file-menu-trigger'}}
       >
         <Menu
           class='file-tree-context-menu-list'
@@ -218,7 +220,6 @@ export default class Directory extends Component<Args> {
     registerDestructor(this, () => {
       this.menuEntryPath = undefined;
       this.menuTriggerEl = undefined;
-      document.removeEventListener('pointerdown', this.handleOutsideClick);
     });
   }
 
@@ -275,7 +276,6 @@ export default class Directory extends Component<Args> {
   private closeMenu() {
     this.menuEntryPath = undefined;
     this.menuTriggerEl = undefined;
-    document.removeEventListener('pointerdown', this.handleOutsideClick);
   }
 
   @action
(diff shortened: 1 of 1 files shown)


lukemelia (commented):
Any reason not to use floating-ui for this kind of thing?

jurgenwerk:
I find the auto close behavior unexpected and inconsistent with other contextual menus that we have. If I open a contextual menu I would expect it to stay open

tintinthong:
> Any reason not to use floating-ui for this kind of thing?

good point. refactored to use ember-velcro

tintinthong:
> I find the auto close behavior unexpected and inconsistent with other contextual menus that we have. If I open a contextual menu I would expect it to stay open

noted! thanks for feedback. removed any auto-closing and accounted for click outside close

lukemelia (approved):
So we have an onClickOutside modifier we can use?

PR #4296, opened 2026-04-01
Resolve module identifier with @cardstack/catalog/ prefix

tintinthong:
This affects LHS when we open a module 

Screenshot 2026-04-01 at 2 45 01 PM

It seems this is activated when opening a card that imports @cardstack/catalog/prefix from the type-chain somehow. Either LHS or RHS

### DIFF
--- a/packages/host/app/services/card-type-service.ts
+++ b/packages/host/app/services/card-type-service.ts
@@ -120,9 +120,10 @@ export default class CardTypeService extends Service {
       return cached;
     }
     let moduleIdentifier = moduleFrom(ref);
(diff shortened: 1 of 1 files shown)


richardhjtan (approved):
This fix works without error coming from the LHS inspector

habdelra (approved):
this is never ending....

tintinthong:
For future reference,

1. for instances the lhs doesn't load

Screenshot 2026-04-02 at 09 39 58

2. for modules, 

Screenshot 2026-04-02 at 09 49 00

PR #4306, opened 2026-04-02
CS-10580, CS-10582: Fix card catalog scrollbar spacing and overflow

FadhlanR:
## Summary
- Adds `padding-right: var(--boxel-sp)` to `.search-sheet-content` so that search results are not flush against the scrollbar in the card catalog (CS-10580)
- Adds `overflow-x: hidden` to `.search-sheet-content` to prevent a horizontal scrollbar from appearing on strip card size (CS-10582)

Both are fixes for PR feedback from Luke on CS-10380 (#4171).

### Before

https://github.com/user-attachments/assets/87bb7978-c7ed-4cf9-8a21-a18058691970

### After

https://github.com/user-attachments/assets/cd3d2b73-d91d-496b-9f33-ad94f866b734

### DIFF
--- a/packages/host/app/components/card-search/search-content.gts
+++ b/packages/host/app/components/card-search/search-content.gts
@@ -644,7 +644,6 @@ export default class SearchContent extends Component<Signature> {
         display: flex;
         flex-direction: column;
         flex: 1;
-        overflow-x: hidden;
         overflow-y: auto;
         overscroll-behavior: none;
         height: 100%;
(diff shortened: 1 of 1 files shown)


lukemelia (changes requested):
Fixing the layout so that the strip doesn't trigger horizontal scrolling would be better than hiding the scroll bar.

FadhlanR:
Removed `overflow-x: hidden` — the `padding-right` alone is sufficient to prevent the horizontal scrollbar in strip view. Thanks for the feedback Luke!

lukemelia (approved):

PR #4307, opened 2026-04-02
CS-10581: Fix type picker search focus in card catalog modal

FadhlanR:
## Problem

In the card catalog modal, clicking the "Search for a type" placeholder text does not focus the search input field ([CS-10581](https://linear.app/cardstack/issue/CS-10581)). This is because the picker dropdowns are rendered via ember-power-select's wormhole to `#ember-basic-dropdown-wormhole` (outside the modal DOM), and the `ember-focus-trap` modifier prevents focus from reaching elements outside its boundary.

## Root Cause

The focus trap on the modal treats only the modal's DOM subtree as focusable. When a picker dropdown opens, its content is portaled to `#ember-basic-dropdown-wormhole` in the document body — outside the focus trap boundary. The `autoFocus` modifier on the dropdown's search input tries to focus it, but the focus trap intercepts and pulls focus back into the modal.

## Solution

Use the `additionalElements` parameter on the `ember-focus-trap` modifier to include `#ember-basic-dropdown-wormhole` as part of the focus trap boundary. This way:

- **Focus trap is preserved** — keyboard users cannot Tab out to background controls (the modal uses `<dialog open>`, not `showModal()`, so there's no native focus trapping)
- **Dropdown renders normally** via wormhole with correct absolute positioning (no clipping/overflow issues)
- **Single change in one file** — no need to thread `renderInPlace` through SearchBar → TypePicker → RealmPicker → Picker

### Changes vs main

- `packages/host/app/components/card-catalog/modal.gts`: Added `additionalElements=this.focusTrapAdditionalElements` to the focus-trap modifier, with a getter that returns `[#ember-basic-dropdown-wormhole]`

### Approaches considered and rejected

1. **`renderInPlace=true`** — renders dropdown inline inside the modal DOM (stays inside focus trap), but causes dropdown clipping/overflow due to modal's constrained height. Required threading the prop through 4 component layers.
2. **Removing focus trap entirely** — fixes the focus issue but allows keyboard users to Tab out of the modal to background page controls, which is a regression.

## Test plan
- [ ] Open the card catalog modal — verify search field auto-focuses
- [ ] Click the "Search for a type" placeholder — verify input focuses and accepts typing
- [ ] Verify type picker dropdown renders with correct positioning (not clipped)
- [ ] Verify realm picker dropdown also works correctly
- [ ] Tab key stays within the modal — cannot tab to background controls
- [ ] Escape key still closes the modal

### DIFF
--- a/packages/boxel-ui/addon/src/components/multi-select/index.gts
+++ b/packages/boxel-ui/addon/src/components/multi-select/index.gts
@@ -23,6 +23,7 @@ export interface BoxelMultiSelectArgs<ItemT> extends PowerSelectArgs {
   ariaLabel?: string;
   beforeOptionsComponent?: ComponentLike<any>;
   closeOnSelect?: boolean;
+  destination?: string;
   disabled?: boolean;
   dropdownClass?: string;
   extra?: any;
@@ -74,6 +75,7 @@ export class BoxelMultiSelectBasic<ItemT> extends Component<Signature<ItemT>> {
       @placeholder={{@placeholder}}
       @disabled={{@disabled}}
       @renderInPlace={{@renderInPlace}}
+      @destination={{@destination}}
       @matchTriggerWidth={{@matchTriggerWidth}}
       @searchField={{@searchField}}
       @searchEnabled={{@searchEnabled}}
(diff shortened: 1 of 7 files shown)


lukemelia (approved):
Is a test possible?

PR #4309, opened 2026-04-02
Add BFM card references for MarkdownDef (CS-10570)

lukemelia:
## Summary
- Add Boxel Flavored Markdown card reference support: `:card[URL]` (inline, atom format) and `::card[URL]` (block, embedded format)
- New shared parsing utility (`bfm-card-references.ts`) with keyword-generic design for future `:file` support (CS-10583)
- MarkdownDef gains `cardReferenceUrls` (computed) and `linkedCards` (query-backed `linksToMany`) fields
- MarkdownTemplate renders linked cards into BFM placeholder elements via modifier + `{{in-element}}`

## Test plan
- [x] 22 unit tests for extraction, marked extensions, HTML generation, sanitization, and edge cases (code blocks, dedup, special chars)
- [x] Acceptance tests for inline/block card rendering and fallback text (require full test infrastructure)

image


burieberry (commented):
This is cool. However when there's `.json` in the URL, the card embed doesn't work. We should handle that, too.

lukemelia:
> This is cool. However when there's .json in the URL, the card embed doesn't work. We should handle that, too.

Added.

packages/runtime-common/realm-index-query-engine.ts
```diff
@@ -442,6 +442,20 @@ export class RealmIndexQueryEngine {
     return await this.#indexQueryEngine.getFile(url, opts);
   }
 
+  async loadLinksForResource(
```
habdelra: hmmm, this is making my spidey senses tingle, tell me more about how this is used. I want to make sure that it plays nice with how the store is loading links. we always lazily load links based on templates pulling on the links. i want to make sure this doesn't cause something to violate that.

lukemelia: This brings to file-def JSON-API the same thing we do for card-def JSON-API, which is to sideload any referenced linksTo/linksToMany cards in the `included` node

burieberry (approved):

packages/runtime-common/realm-index-query-engine.ts
habdelra: can you point to a callsite where this is used? i'm actually having trouble finding it

lukemelia: Good catch! It got lost. Let me get it back in there.

habdelra (approved):

PR #4311, opened 2026-04-02
Upgrade to Glint v2

lukemelia:
Migrates from Glint v1 to Glint v2 following the [official upgrade guide](https://typed-ember.gitbook.io/glint/v2-upgrade).

## Changes

- **Add `@glint/ember-tsc` 1.5.0** — the v2 replacement for the `glint` CLI
- **Upgrade `@glint/template`** from 1.3.0 → 1.7.7 (compatible with both v1 and v2)
- **Remove v1 packages**: `@glint/core`, `@glint/environment-ember-loose`, `@glint/environment-ember-template-imports`
- **Update tsconfig.json** — remove `"glint"` config objects from all 16 tsconfig files
- **Update scripts** — replace `glint` CLI with `ember-tsc` equivalents:
  - `glint` → `ember-tsc --noEmit`
  - `glint --declaration` → `ember-tsc --declaration --emitDeclarationOnly`
- **Update ambient type imports** — replace `@glint/environment-ember-loose` imports with `@glint/ember-tsc/types`
- **Remove loose mode registry augmentations** — no longer needed without ember-loose environment
- **Convert `standby.hbs` → `standby.gts`** — the last remaining .hbs template in the repo
- **Update pnpm workspace** — remove `@glint/environment-ember-loose` from catalog, add minimumReleaseAgeExclude for new Glint packages

## Notes

- This is a full v2 migration (not hybrid v1+v2)
- All `.gts`/`.gjs` files are supported; loose mode (.hbs) support is dropped
- `@glint/template` remains as a dependency (unchanged API, compatible with v2)


packages/catalog-realm/blog-app/review-blog.gts
```diff
@@ -39,6 +41,7 @@ export class ReviewBlog extends BlogApp {
       createNewButtonText: 'Category',
       cardRef: {
         name: 'BlogCategory',
+        // @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS
```
ef4, line 44: I don't think this is a gjs problem, I think this is because `moduleResolution: nodenext` follows node rules, and by node rules our files would only be interpreted as esm if we set `type: module` in package.json.

ef4 (approved):
This version is definitely better at finding problems with HTML attributes.

I left one small comment on why TS is treating everything as CJS.

packages/catalog-realm/blog-app/review-blog.gts
lukemelia: I asked Claude the same question. This is what I got; let me know what you think:

Unfortunately, "type": "module" doesn't work here. The root issue is that the host's  
  tsconfig uses module: "nodenext" and includes the realm files directly via include.   
  With nodenext + "type": "module", TypeScript enforces mandatory file extensions on all
   relative imports — but the realm files use extensionless imports designed for the    
  card runtime's custom loader.                                                         
   
  The alternatives would be:                                                            
                                                                                     
  1. Change host to module: "preserve" / moduleResolution: "bundler" — would allow      
  import.meta everywhere and not enforce extensions, but is a larger change that could
  have other effects across the whole host package                                      
  2. Separate tsconfigs with project references — realm packages get their own tsconfig
  with a module setting that allows import.meta, but that's a significant architecture  
  change
  3. Keep @ts-expect-error — pragmatic, explicit about the mismatch between the card    
  runtime's ESM semantics and TypeScript's CJS detection                                
   
  The @ts-expect-error comments are the least risky approach given the constraints. The 
  fundamental tension is that these realm files are ESM at runtime (loaded by the card
  runtime) but TypeScript sees them as CJS because nodenext respects the missing "type":
   "module" — and adding that field breaks extensionless imports.

habdelra (approved):

PR #4312, opened 2026-04-02
CS-10379: Trigger search query in initial state and when filters change

FadhlanR:
## Summary
- Add `baseFilter` to the search sheet so the query fires on open (via the existing `shouldSkipSearchQuery` logic that allows queries when `baseFilter` is present)
- Add `onFilterChange` callback chain (`panel.gts` → `search-sheet` → `submode-layout`) to expand the search sheet from compact/prompt mode to results mode when realm or type picker selection changes
- Fix `RealmPicker` to absorb the default selection internally (matching the existing `TypePicker` pattern) so Picker's `ensureDefaultSelection` doesn't trigger the parent's `onChange` and inadvertently expand the sheet on open

### Before

https://github.com/user-attachments/assets/34bb74a5-b6b5-4df8-8566-4dd786f001d3

### After

https://github.com/user-attachments/assets/f2fd94b4-0932-466f-b373-18a5168668a8


jurgenwerk:
When I select something in the dropdown, it stays floating in the bottom, not sure if this is intentional or a bug, it feels odd to have the dropdown opened in a place where its trigger is not present anymore:

image

FadhlanR:
> When I select something in the dropdown, it stays floating in the bottom, not sure if this is intentional or a bug, it feels odd to have the dropdown opened in a place where its trigger is not present anymore:
> 
> image



Good catch! This was caused by the search sheet's 0.2s CSS height transition — when the sheet expands from prompt to results mode, ember-basic-dropdown's `reposition()` fires before the transition completes, so the wormholed dropdown reads the trigger's pre-transition position.

Fixed in 66b2d524ed by adding a `transitionend` listener on the search sheet element that dispatches a `window.resize` event after the height transition finishes. This triggers ember-basic-dropdown's built-in reposition handler, so any open dropdown (TypePicker, RealmPicker) recalculates its position once the sheet has fully settled.

lukemelia (approved):

PR #4327, opened 2026-04-03
To reduce Percy consumption, only run Percy snapshots when PR is not draft anymore

jurgenwerk:
We've used all our Percy monthly quota in just 3 days. 

This is my idea of significantly reducing percy checks: run it only when the author of the PR assigns a reviewer. This usually happens when the PR is mature enough to be worthy of a  Percy run but not earlier (given the limit)

There is a drawback though - when assigning a reviewer, the CI will run again, which is not the most optimal, but maybe we can live with it?

### DIFF
--- a/.github/workflows/ci-host.yaml
+++ b/.github/workflows/ci-host.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [main]
   pull_request:
+    types: [opened, synchronize, reopened, review_requested]
     paths:
       - "packages/host/**"
       - "packages/base/**"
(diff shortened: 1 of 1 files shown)


habdelra:
will this also be smart enough to not run CI again on review assignment if we never ran host tests in the first place because CI determined that the changes didn't impact the host app (like ai-bot changes)?

lukemelia:
Will this kick off a full CI run every time a reviewer is assigned? That seems like too much. I usually request a few reviewers.

jurgenwerk:
> will this also be smart enough to not run CI again on review assignment if we never ran host tests in the first place because CI determined that the changes didn't impact the host app (like ai-bot changes)?

That's right, the path filter catches it - host tests won't be run (and percy won't be run)

> Will this kick off a full CI run every time a reviewer is assigned? That seems like too much. I usually request a few reviewers.

Yes, this was the issue I was struggling with, it's really not a good idea to run it on every added reviewer. I changed this condition so that percy runs when a PR is promoted from draft and marked as ready. Let's see how this works for us and if additional filtering is needed in the future.

backspace (approved):
Now that the review bots make their commentaries after a PR has been marked ready for review, I’ve started waiting for those and making recommended adjustments before requesting review from coworkers. So looking at Percy diffs as part of that fits my workflow well. I’d love if we could further reduce spurious diffs though, which lately seem mostly to be missing fonts and Monaco highlighting.

PR #4328, opened 2026-04-03
refactor: use inline usage cost from OpenRouter instead of generation cost API

FadhlanR:
## Summary
- OpenRouter now includes cost directly in responses via `usage.cost`, so we primarily use that instead of polling a separate endpoint
- Removed the old `saveUsageCost` flow that always polled OpenRouter's `/generation` API, and replaced it with direct `spendUsageCost` calls using the inline cost
- Simplified `CreditStrategy` interface by removing the separate `spendUsageCost` method — `saveUsageCost` now handles both inline cost extraction and fallback
- ~140 lines of billing code removed (old `saveUsageCost`, `extractGenerationIdFromResponse`)

### Why we still need the generation cost API as a fallback
OpenRouter includes `usage.cost` in the **final streaming chunk** (the one with `finish_reason`). However, if a user **cancels/stops the stream** before that final chunk arrives, the cost is never received. Without a fallback, these interrupted generations would go unbilled. The generation cost API polling (`/generation?id=`) is retained as a fallback for this case — OpenRouter still tracks the cost server-side even for interrupted streams, so we can retrieve it after the fact.

**Flow:**
1. Inline `usage.cost` available → use it directly (fast path, no extra API call)
2. No inline cost but `generationId` available → poll `/generation?id=` endpoint with backoff (fallback for cancelled streams)
3. Neither available → log warning, skip deduction

Closes CS-10506

## Test plan
- [ ] Verify AI chat generates responses and credits are deducted correctly (inline cost path)
- [ ] Verify streaming responses that are cancelled mid-way still deduct credits (fallback path)
- [ ] Verify non-streaming forwarded requests deduct credits from inline cost
- [ ] Run realm-server request-forward tests (includes both inline and fallback test cases)

### DIFF
--- a/packages/ai-bot/main.ts
+++ b/packages/ai-bot/main.ts
@@ -40,7 +40,7 @@ import {
 import type { MatrixEvent as DiscreteMatrixEvent } from 'https://cardstack.com/base/matrix-event';
 import * as Sentry from '@sentry/node';
 
-import { saveUsageCost } from '@cardstack/billing/ai-billing';
+import { spendUsageCost } from '@cardstack/billing/ai-billing';
 import { PgAdapter } from '@cardstack/postgres';
 import type { ChatCompletionMessageParam } from 'openai/resources';
 import type { OpenAIError } from 'openai/error';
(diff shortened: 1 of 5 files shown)


jurgenwerk (approved):
I tried running this locally and I saw that credits are deducted correctly

However, when I stopped AI generation in the middle of AI response, I did not see any credits being spent. I am not sure if this was how it worked before this change but something you might want to take a look at.

packages/ai-bot/main.ts
```diff
@@ -86,22 +89,41 @@ class Assistant {
...
+      let { costInUsd, generationId } = opts;
+      if (
+        typeof costInUsd === 'number' &&
+        Number.isFinite(costInUsd) &&
+        costInUsd > 0
+      ) {
+        await spendUsageCost(this.pgAdapter, matrixUserId, costInUsd);
+      } else if (generationId) {
+        log.info(
+          `No inline cost for user ${matrixUserId}, falling back to generation cost API (generationId: ${generationId})`,
```
jurgenwerk, line 110: In which case there is no inline cost?

FadhlanR:
> I tried running this locally and I saw that credits are deducted correctly
> 
> However, when I stopped AI generation in the middle of AI response, I did not see any credits being spent. I am not sure if this was how it worked before this change but something you might want to take a look at.

I also experienced the same issue on the main branch, where credits were not deducted after cancellation. However, after some investigation, I found that the issue is not caused by our code. It occurs because the cost data for canceled events takes longer to be recorded by OpenRouter. We have implemented a backoff strategy to minimize the issue, but it can still occasionally occur.

PR #4333, opened 2026-04-06
Fix BoxelSelect dropdown option content alignment

burieberry:
Before:
before

After:
after


jurgenwerk (approved):

packages/boxel-ui/addon/src/icons/types.ts
```diff
@@ -1,7 +1,10 @@
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
 import type { ComponentLike } from '@glint/template';
 
 export interface Signature {
   Element: SVGSVGElement;
 }
 
 export type Icon = ComponentLike<Signature>;
+
+export type IconComponent = TemplateOnlyComponent<Signature>;
```
burieberry, line 10: When I used type `Icon` in boxel-ui usage doc, I got a type error that prevented boxel-ui from building. This is why I added this `IconComponent` type. Let me know of your opinions

PR #4329, opened 2026-04-05
Add BFM Layer 3 markdown extensions

lukemelia:
- [x] First merge https://github.com/cardstack/boxel/pull/4309

## Summary
- Registers 5 community `marked` extensions and 1 custom extension for BFM (Boxel Flavored Markdown) Layer 3 features:
  - **GFM Alerts** (`marked-alert`): admonition syntax (`> [!NOTE]`, `> [!WARNING]`, etc.)
  - **Footnotes** (`marked-footnote`): `[^1]` reference + definition syntax
  - **Extended Tables** (`marked-extended-tables`): colspan/rowspan support
  - **Heading IDs** (`marked-gfm-heading-id`): auto-generated slug IDs on headings
  - **Math/LaTeX placeholders** (custom `bfm-math`): `$...$` and `$$...$$` emit lightweight placeholder HTML instead of bundling KaTeX (~268KB min + 1.1MB fonts)
  - **Mermaid placeholders**: ` ```mermaid ` code blocks emit `<pre class="mermaid">` for client-side rendering
- Each feature has smoke tests covering `markedSync` output and DOMPurify sanitization pass-through
- Math and Mermaid use placeholder patterns for lazy client-side hydration, avoiding bundle bloat
- **Mermaid client-side rendering (CS-10602)**: lazy-loads Mermaid.js (~2MB) on demand via a global loader bridge (same pattern as Monaco), renders SVG diagrams in-place with `securityLevel: 'strict'`

## Test plan
- [x] Unit tests pass for all new marked extensions (`marked-sync-test.ts`)
- [x] Existing BFM card reference tests updated for heading ID output
- [x] Verify DOMPurify doesn't strip new markup (alerts, math placeholders, footnotes, heading IDs)
- [x] Acceptance test verifies mermaid code blocks render as SVG diagrams
- [x] Mermaid is NOT in the initial bundle (lazy-loaded only when diagrams are present)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4334, opened 2026-04-06
Implement BFM fitted sizes, custom dimensions, and format summary

lukemelia:
- [x] First merge https://github.com/cardstack/boxel/pull/4329
## Summary
- Adds pipe separator parsing to `::card[url | specifier]` block syntax for controlling card embedding format and dimensions
- Supports all named size constants from the BFM spec (badges, strips, tiles, cards) plus BFM shorthand aliases (`strip`, `tile`, `grid-tile`)
- Supports custom dimensions: `WxH`, `w:N h:N`, `h:N` (height-only), `w:N%` (percentage width)
- Supports `isolated` format keyword for full standalone card rendering
- Markdown renderer reads format/size data attributes and applies `fitted` format with inline dimension styles
- 32 new unit tests covering all size constants, custom dimensions, pipe parsing, and edge cases

## Test plan
- [x] All 66 unit tests pass (`npx ember test --filter "bfm-card-references"`)
- [ ] Visual verification of fitted card rendering in experiments realm via BFM showcase

image


habdelra (approved):

burieberry (approved):
Looks good. What do you think about displaying the card boundaries via the card-container? It is setting the border-radius already.

Just FYI: Boxel FittedCardContainer component(https://github.com/cardstack/boxel/blob/main/packages/boxel-ui/addon/src/components/fitted-card-container/index.gts) does something similar to the markdown-bfm-card-slot container

Here are some helpers: https://github.com/cardstack/boxel/blob/main/packages/boxel-ui/addon/src/utils/fitted-formats.ts

lukemelia:
@burieberry wrote:

> Just FYI: Boxel FittedCardContainer component(https://github.com/cardstack/boxel/blob/main/packages/boxel-ui/addon/src/components/fitted-card-container/index.gts) does something similar to the markdown-bfm-card-slot container
>
> Here are some helpers: https://github.com/cardstack/boxel/blob/main/packages/boxel-ui/addon/src/utils/fitted-formats.ts

Good callout. I looked into sharing this, but the dependency directions seemed awkward. It made me wonder if we should create a new "design-tokens" package...

PR #4341, opened 2026-04-07
Cs 10652/introduce the cardthumbnail inside cardinfofield

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10652/introduce-the-cardthumbnail-inside-cardinfofield

**Fix cyclic deps:** 

1. Move FileDef, ImageDef, and NumberField into card-api; add cardThumbnail field to CardInfo
2. Consolidate definitions into card-api: FileDef, ImageDef, and NumberField have been moved into card-api.gts as their canonical home. file-api.gts and image-file-def.gts now re-export from card-api via shim modules to preserve backward compatibility for existing consumers. 
3. Update canonical module URLs: Since FileDef and ImageDef are now canonically defined in card-api, all test fixtures, query filters, and module references that previously pointed to file-api or image-file-def have been updated to point to card-api.
4. Add cardThumbnail field to CardInfo
5. 
## Demo: 
https://github.com/user-attachments/assets/5649efd3-3a0a-4b6f-bfd9-dd8f59636fb9


lucaslyl:
@codex review

packages/base/card-api.gts
```diff
@@ -2495,7 +3002,7 @@ export class CardDef extends BaseDef {
   // to use it directly now (or wait until a better image field comes along)
   @field cardThumbnailURL = contains(MaybeBase64Field, {
     computeVia: function (this: CardDef) {
-      return this.cardInfo.cardThumbnailURL;
+      return this.cardInfo.cardThumbnail?.url ?? this.cardInfo.cardThumbnailURL;
```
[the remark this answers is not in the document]
habdelra: this is a really bad suggestion. it should be ok to pull on a linksTo field in the computed. we will only lazily load it if this field is included in a template, which is exactly the behavior that we want. we should NEVER reach directly into the low level data bucket for a card. that is not a userland API. ignore this review comment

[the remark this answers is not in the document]
lucaslyl: Agreed with it. Sometimes Copilot gives feedback that feels a bit aggressive, which is not the behavior we want.

burieberry (commented):
To keep the amount of code added to the card-api to a minimum, you can place any large component or helpers/etc that doesn't cause cyclic error outside of the card-api. You'll see we're doing this with a bunch of templates in /default-templates.

lucaslyl:
@habdelra I’ve confirmed that all tests pass locally, but they seem to be flaky during CI checks (failing randomly on each rerun attempt). Do you have any idea what might be causing this

habdelra:
> @habdelra I’ve confirmed that all tests pass locally, but they seem to be flaky during CI checks (failing randomly on each rerun attempt). Do you have any idea what might be causing this

thanks for flagging this. off the top of my my head no. but claude is pretty good at figuring this out (i usually use thinking "high" for this kind of stuff). you can point out the failing tests to claude, and it might not be a bad idea for it to see it in context with the passing tests, so that it can determine if there is state leakage. you can also ask claude to add additional logging to help diagnose the flakiness in CI, and then just feedback the logs from CI back into claude for analysis. generally i find that this kind of approach solves most of the flaky issues

habdelra:
another thought based on some of the failures you are getting is that there is a memory leak. it might be worthwhile to have claude do a flame chart analysis using chrome mcp on test suites that have tests that time out at the end (that is indicative of an out of memory error). like this: 
```not ok 180 Chrome - [undefined ms] - error
    ---
        message: >
            Error: Browser timeout exceeded: 60s
            Error while executing test: Integration | serialization > base cards > BigIntegerField: can perform bigint operations with computed
            Stderr: 
             [7101:7101:0410/141631.148233:ERROR:base/memory/shared_memory_switch.cc:289] Failed global descriptor lookup: 7
            
            DevTools listening on ws://127.0.0.1:46331/devtools/browser/2a535e2f-8636-41d1-b8a8-ea974f1be0fc
            [7028:7072:0410/141735.402808:ERROR:google_apis/gcm/engine/registration_request.cc:291] Registration response error message: DEPRECATED_ENDPOINT
            Created TensorFlow Lite XNNPACK delegate for CPU.
            [7028:7072:0410/141803.881843:ERROR:google_apis/gcm/engine/registration_request.cc:291] Registration response error message: DEPRECATED_ENDPOINT
            [7028:7072:0410/141859.309882:ERROR:google_apis/gcm/engine/registration_request.cc:291] Registration response error message: DEPRECATED_ENDPOINT
            [7028:7072:0410/142041.643824:ERROR:google_apis/gcm/engine/registration_request.cc:291] Registration response error message: DEPRECATED_ENDPOINT
            [7028:7072:0410/142419.606458:ERROR:google_apis/gcm/engine/registration_request.cc:291] Registration response error message: DEPRECATED_ENDPOINT
            [0410/142840.790381:ERROR:third_party/crashpad/crashpad/util/file/file_io_posix.cc:145] open /sys/devices/system/cpu/cpu0/cpufreq/scaling_cur_freq: No such file or directory (2)
            [0410/142840.790447:ERROR:third_party/crashpad/crashpad/util/file/file_io_posix.cc:145] open /sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq: No such file or directory (2)
            [7028:7072:0410/143002.840930:ERROR:google_apis/gcm/engine/registration_request.cc:291] Registration response error message: DEPRECATED_ENDPOINT
            
            
        browser log: |
            {"type":"error","text":"Error: Browser timeout exceeded: 60s"}
            {"type":"error","text":"Error while executing test: Integration | serialization > base cards > BigIntegerField: can perform bigint operations with computed"}
            {"type":"error","text":"[7101:7101:0410/141631.148233:ERROR:base/memory/shared_memory_switch.cc:289] Failed global descriptor lookup: 7\n\nDevTools listening on ws://127.0.0.1:46331/devtools/browser/2a535e2f-8636-41d1-b8a8-ea974f1be0fc\n[7028:7072:0410
```
also make sure you have merged upstream from main to to make sure there arent any fixes that you are missing

lucaslyl:
> thanks for flagging this. off the top of my my head no. but claude is pretty good at figuring this out (i usually use thinking "high" for this kind of stuff). you can point out the failing tests to claude, and it might not be a bad idea for it to see it in context with the passing tests, so that it can determine if there is state leakage. you can also ask claude to add additional logging to help diagnose the flakiness in CI, and then just feedback the logs from CI back into claude for analysis. generally i find that this kind of approach solves most of the flaky issues

Thanks, I appreciate the suggestions. I’ll look into this following your approach and see what I find

packages/host/tests/integration/realm-test.gts
```diff
@@ -320,87 +320,50 @@ module('Integration | realm', function (hooks) {
...
+    assert.strictEqual(included.length, 1, 'one included card');
+    let hassan = included[0];
+    assert.strictEqual(
+      hassan.id,
+      'http://localhost:4202/test/hassan',
+      'included hassan id',
+    );
+    assert.strictEqual(
+      hassan.relationships['cardInfo.theme'].links.self,
+      null,
```
lucaslyl, line 364: I changed the assert pattern because this response combines cards serialized from two different paths: local test-realm cards (data) and cross-realm included cards (included). For optional null relationships like cardInfo.cardThumbnail, the local path may omit the key while the cross-realm included payload may still include it, even though both represent valid equivalent state. That made full-object deepEqual brittle and caused false failures based on JSON shape differences rather than behavior. The new assertion style checks the required semantics (IDs, links, required relationships like cardInfo.theme, included-card presence, and key metadata) without over-constraining optional relationship encoding.

packages/host/tests/helpers/index.gts
```diff
@@ -937,10 +937,11 @@ export function setupLocalIndexing(hooks: NestedHooks) {
...
     context.__cardPrerenderElement = undefined;
     context.__hasMountedCardPrerender = undefined;
-    // reference counts should balance out automatically as components are destroyed
-    store.resetCache({ preserveReferences: true });
+    // In large suites, preserved references can accumulate across tests and
+    // keep GC pressure high enough to cause late test timeouts.
+    store.resetCache();
     let renderStore = getService('render-store');
-    renderStore.resetCache({ preserveReferences: true });
+    renderStore.resetCache();
```
lucaslyl, line 944: @habdelra I fixed the intermittent 60s CI timeout by removing cross-test reference retention during test teardown. Based on Codex’s analysis, the root cause was that preserveReferences: true kept store reference counts between tests. In large serialization suites, that state accumulated over time, increasing memory/GC pressure and causing late-run browser stalls/timeouts. 

I don’t have much knowledge of the store and loader, does this fix suggestion make sense?

habdelra: that does make sense. thanks!

habdelra (approved):

PR #4347, opened 2026-04-07
Fix flaky prerender timeout test race condition

habdelra:
## Summary
- Fixes a race condition in `withTimeout` (`packages/realm-server/prerender/utils.ts`) where Puppeteer's own `TimeoutError` from `waitForFunction`/`waitForSelector` could fire before our outer `setTimeout`, bypassing the clean "Render timed-out" path
- When Puppeteer's timeout fires first, it now gets caught and converted to the same `{ timeout: true }` sentinel, ensuring consistent `unusable` status instead of flaky `error` status
- Fixes the flaky test: "it returns unusable status when command times out"

## Test plan
- [ ] CI passes on `prerender-server-test.ts` timeout test
- [ ] No regressions in other prerender tests

### DIFF
--- a/packages/realm-server/prerender/utils.ts
+++ b/packages/realm-server/prerender/utils.ts
@@ -1045,7 +1045,14 @@ export async function withTimeout<T>(
   timeoutMs = cardRenderTimeout,
 ): Promise<T | RenderError> {
   let result = await Promise.race([
-    fn(),
+    fn().catch((err) => {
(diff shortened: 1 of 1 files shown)


jurgenwerk (approved):

tintinthong (approved):

tintinthong:
All GREEN. Merging to fix CI on main. Thank you @habdelra

PR #4348, opened 2026-04-07
Fix "Invalid base URL" when creating card def inheriting from cross-realm spec

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10590/create-new-card-def-from-listing-hit-invalid-base-url

When creating a new Card Definition and selecting "Inherits From" a spec from another realm (e.g. Listing from catalog realm), the modal threw TypeError: Failed to construct 'URL': Invalid base URL.

https://github.com/user-attachments/assets/9ef60541-9ec8-4799-8f6f-53105b2fd44d

## Fix

1. Resolve spec.moduleHref through codeRefWithAbsoluteURL() before constructing the module URL, so prefix-form refs are normalized to real URLs consistently.
2. Use cardIdToURL() for spec IDs so prefix-form card IDs are resolved correctly when deriving the base URL.

## Test Added
Added a regression test in create-file-test covering creation of a card definition in workspace A that extends a card from workspace B via a prefix-form ref, ensuring the inherited definition import is generated from the resolved module URL.


lucaslyl:
@codex review

habdelra:
I would like to see a regression test for this. please include a test that fails without this fix but passes with it.

lucaslyl:
> I would like to see a regression test for this. please include a test that fails without this fix but passes with it.

I added a regression test in create-file-test for creating a card definition from a spec whose ref uses a prefix-form module ID, so we cover resolving that ref before generating the import.

habdelra (approved):

PR #4354, opened 2026-04-08
CS-10615: Reimplement boxel profile command

FadhlanR:
## Summary
- Port `boxel profile` command from standalone boxel-cli into monorepo at `packages/boxel-cli`
- Add `ProfileManager` class for CRUD operations on `~/.boxel-cli/profiles.json` (0600 perms)
- Subcommands: `list`, `add` (interactive wizard), `switch` (partial match), `remove`, `migrate` (from .env)
- 22 unit tests covering all ProfileManager operations + environment helpers

## Test plan
- [x] `pnpm --filter @cardstack/boxel-cli test` — 24 tests pass
- [x] `pnpm --filter @cardstack/boxel-cli build` — builds successfully

### DIFF
--- a/packages/boxel-cli/src/commands/profile.ts
+++ b/packages/boxel-cli/src/commands/profile.ts
@@ -1,0 +1,418 @@
+import * as readline from 'readline';
+import { Writable } from 'stream';
+import type { ProfileManager } from '../lib/profile-manager.js';
+import {
+  getProfileManager,
+  formatProfileBadge,
+  getEnvironmentFromMatrixId,
+  getEnvironmentShortLabel,
+  getUsernameFromMatrixId,
+} from '../lib/profile-manager.js';
+
+// ANSI color codes
+const FG_GREEN = '\x1b[32m';
+const FG_YELLOW = '\x1b[33m';
+const FG_CYAN = '\x1b[36m';
+const FG_MAGENTA = '\x1b[35m';
+const FG_RED = '\x1b[31m';
+const DIM = '\x1b[2m';
+const BOLD = '\x1b[1m';
+const RESET = '\x1b[0m';
+
+function prompt(question: string): Promise<string> {
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: process.stdout,
+  });
+
+  return new Promise((resolve) => {
+    rl.question(question, (answer) => {
+      rl.close();
+      resolve(answer.trim());
+    });
+  });
+}
+
+function promptPassword(question: string): Promise<string> {
+  const mutableOutput = new Writable({
+    write: (_chunk, _encoding, callback) => callback(),
+  });
+  const rl = readline.createInterface({
+    input: process.stdin,
+    output: mutableOutput,
+    terminal: true,
+  });
+
+  return new Promise((resolve) => {
+    const stdin = process.stdin;
+    if (stdin.isTTY) {
+      stdin.setRawMode(true);
+    }
+
+    process.stdout.write(question);
+    let password = '';
+
+    const onData = (char: Buffer) => {
+      const c = char.toString();
+      if (c === '\n' || c === '\r') {
+        stdin.removeListener('data', onData);
+        if (stdin.isTTY) {
+          stdin.setRawMode(false);
+        }
+        process.stdout.write('\n');
+        rl.close();
+        resolve(password);
+      } else if (c === '\u0003') {
+        // Ctrl+C
+        process.exit();
+      } else if (c === '\u007F' || c === '\b') {
+        // Backspace
+        if (password.length > 0) {
+          password = password.slice(0, -1);
+          process.stdout.write('\b \b');
+        }
+      } else {
+        password += c;
+        process.stdout.write('*');
+      }
+    };
+
+    stdin.on('data', onData);
+    stdin.resume();
+  });
+}
+
+export interface ProfileCommandOptions {
+  user?: string;
+  password?: string;
+  name?: string;
+}
+
+export async function profileCommand(
+  subcommand?: string,
+  arg?: string,
+  options?: ProfileCommandOptions,
+): Promise<void> {
+  const manager = getProfileManager();
+
+  switch (subcommand) {
+    case 'list':
+      await listProfiles(manager);
+      break;
+
+    case 'add': {
+      const password = options?.password || process.env.BOXEL_PASSWORD;
+      if (options?.user && password) {
+        await addProfileNonInteractive(
+          manager,
+          options.user,
+          password,
+          options.name,
+        );
+      } else {
+        await addProfile(manager);
+      }
+      break;
+    }
+
+    case 'switch':
+      if (!arg) {
+        console.error(
+          `${FG_RED}Error:${RESET} Please specify a profile to switch to.`,
+        );
+        console.log(`Usage: boxel profile switch <profile-id>`);
+        console.log(`\nAvailable profiles:`);
+        await listProfiles(manager);
+        process.exit(1);
+      }
+      await switchProfile(manager, arg);
+      break;
+
+    case 'remove':
+      if (!arg) {
+        console.error(
+          `${FG_RED}Error:${RESET} Please specify a profile to remove.`,
+        );
+        process.exit(1);
+      }
+      await removeProfile(manager, arg);
+      break;
+
+    case 'migrate':
+      await migrateFromEnv(manager);
+      break;
+
+    default:
+      manager.printStatus();
+      console.log(`\n${DIM}Commands:${RESET}`);
+      console.log(
+        `  ${FG_CYAN}boxel profile list${RESET}      List all profiles`,
+      );
+      console.log(
+        `  ${FG_CYAN}boxel profile add${RESET}       Add a new profile`,
+      );
+      console.log(
+        `  ${FG_CYAN}boxel profile switch${RESET}    Switch active profile`,
+      );
+      console.log(
+        `  ${FG_CYAN}boxel profile remove${RESET}    Remove a profile`,
+      );
+      console.log(
+        `  ${FG_CYAN}boxel profile migrate${RESET}   Import from .env file`,
+      );
+  }
+}
+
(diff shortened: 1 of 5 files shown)


packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+    case 'staging':
+      return 'stack.cards';
+    case 'production':
+      return 'boxel.ai';
+    default:
+      return 'unknown';
+  }
+}
+
+/**
```
habdelra, line 66: `getEnvironmentLabel` and `getEnvironmentShortLabel` are identical — they return the exact same values for all cases. Should one be removed, or were they intended to differ (e.g., short label returning abbreviated forms like "staging" / "prod")?

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+const DIM = '\x1b[2m';
+const BOLD = '\x1b[1m';
+const RESET = '\x1b[0m';
+
+export interface Profile {
+  displayName: string;
+  matrixUrl: string;
+  realmServerUrl: string;
+  password: string; // Stored in plaintext - file should have restricted permissions
+}
```
habdelra, line 14: Plaintext password storage: the comment acknowledges this, but worth flagging for the CS-10642 follow-up. When `ProfileManager` becomes the canonical auth store for the full token lifecycle (realm server tokens, per-realm JWTs, auto-acquired tokens on realm creation), plaintext credentials in a JSON file become a bigger liability. Consider planning for OS keychain integration (e.g., `keytar` or `@anthropic-ai/keychain`) as a future step — the `async` signatures on `getPassword`/`updatePassword` are well-positioned for this.

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+    const password = process.env.MATRIX_PASSWORD;
+    let realmServerUrl = process.env.REALM_SERVER_URL;
+
+    if (matrixUrl && username && password) {
+      if (!realmServerUrl) {
+        try {
+          const matrixUrlObj = new URL(matrixUrl);
+          if (matrixUrlObj.hostname.startsWith('matrix.')) {
+            realmServerUrl = `${matrixUrlObj.protocol}//app.${matrixUrlObj.hostname.slice(7)}/`;
+          } else if (matrixUrlObj.hostname.startsWith('matrix-staging.')) {
```
habdelra: This block infers `realmServerUrl` from the `matrixUrl` hostname pattern (e.g., `matrix.` → `app.`, `matrix-staging.` → `realms-staging.`). This is fragile — it breaks for non-standard deployments and encodes URL conventions that may change. The project has an established convention of never inferring realm server URL from other URLs.

Consider requiring `REALM_SERVER_URL` to be set explicitly when falling back to env vars (i.e., return `null` if it's missing rather than guessing). The `migrateFromEnv()` method already requires it explicitly — this fallback path should be consistent.

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+let _instance: ProfileManager | null = null;
+
+export function getProfileManager(configDir?: string): ProfileManager {
+  if (!_instance) {
+    _instance = new ProfileManager(configDir);
+  }
+  return _instance;
+}
+
+/**
```
habdelra, line 360: The singleton caches the first instance, but `configDir` is only used on the first call — subsequent calls silently ignore it. This could cause confusing behavior in tests or when the programmatic API (CS-10642's `BoxelAuth`) needs multiple managers for different contexts.

Consider either:
- Removing the `configDir` parameter from `getProfileManager()` (force callers who need a custom dir to use `new ProfileManager(dir)` directly)
- Or validating that the `configDir` matches if an instance already exists

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+    );
+    return matrixId;
+  }
+
+  printStatus(): void {
+    const active = this.getActiveProfile();
+    if (active) {
+      console.log(
+        `\n${BOLD}Active Profile:${RESET} ${formatProfileBadge(active.id)}`,
+      );
```
habdelra, line 325: If a profile already exists for the migrated Matrix ID, this silently returns the existing ID without updating the password or other fields. A user running `boxel profile migrate` after changing their password in `.env` would expect the stored credentials to update. Consider either updating the existing profile or logging a message that tells the user to use `boxel profile add --user ... --password ...` to update.

packages/boxel-cli/src/commands/profile.ts
```diff
@@ -0,0 +1,418 @@
...
+} from '../lib/profile-manager';
+
+// ANSI color codes
+const FG_GREEN = '\x1b[32m';
+const FG_YELLOW = '\x1b[33m';
+const FG_CYAN = '\x1b[36m';
+const FG_MAGENTA = '\x1b[35m';
+const FG_RED = '\x1b[31m';
+const DIM = '\x1b[2m';
+const BOLD = '\x1b[1m';
```
habdelra: ANSI color constants are duplicated here and in `profile-manager.ts`. Consider extracting them to a shared `utils/colors.ts` or similar — other future commands will likely need them too.

packages/boxel-cli/src/commands/profile.ts
```diff
@@ -0,0 +1,418 @@
...
+      } else if (c === '\u007F' || c === '\b') {
+        // Backspace
+        if (password.length > 0) {
+          password = password.slice(0, -1);
+          process.stdout.write('\b \b');
+        }
+      } else {
+        password += c;
+        process.stdout.write('*');
+      }
```
habdelra, line 89: If an error is thrown between `stdin.setRawMode(true)` (line 50) and the cleanup in the `onData` handler, the terminal will be left in raw mode. Consider wrapping in a try/finally, or adding a process `exit` handler that restores terminal state:

```ts
const cleanup = () => {
  stdin.removeListener('data', onData);
  if (stdin.isTTY) stdin.setRawMode(false);
};
```

Also, `stdin.resume()` at line 76 may not be paired with a `stdin.pause()` — if stdin was already flowing (e.g., piped input), this is fine, but in a TTY context the `rl.close()` in the resolve path doesn't explicitly pause it.

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -0,0 +1,367 @@
...
+
+export interface Profile {
+  displayName: string;
+  matrixUrl: string;
+  realmServerUrl: string;
+  password: string; // Stored in plaintext - file should have restricted permissions
+}
+
+export interface ProfilesConfig {
+  profiles: Record<string, Profile>;
```
habdelra, line 17: Phase 2 / CS-10642 alignment note: The `Profile` interface will need extension to support the two-tier token model. Fields like `serverToken`, `serverTokenExpiry`, `realmTokens: Record<string, { token: string, expiry: number }>` will eventually live here. The current shape is a good foundation — just flagging that `getActiveCredentials()` will evolve significantly when it needs to return (and auto-refresh) realm server tokens and per-realm JWTs.

The `async` signatures on credential methods are well-placed for this future — token refresh will require async operations.

habdelra (commented):
## Review: CS-10615 — Reimplement boxel profile command

**Overall**: Solid foundation for profile management. This is the right first step toward CS-10642's full auth lifecycle.

**Key concerns** (8 inline comments):

1. **Realm server URL inference from matrix URL** — `getActiveCredentials()` guesses `realmServerUrl` from `matrixUrl` hostname patterns when env vars are used. This is fragile and inconsistent with `migrateFromEnv()` which correctly requires `REALM_SERVER_URL` explicitly. Recommend removing the inference.

2. **Duplicate helper functions** — `getEnvironmentLabel` and `getEnvironmentShortLabel` are identical implementations.

3. **Singleton `configDir` confusion** — `getProfileManager(configDir?)` caches the first instance and silently ignores `configDir` on subsequent calls.

4. **`migrateFromEnv` no-op on existing profiles** — Silently returns without updating credentials if profile already exists.

5. **Terminal cleanup in `promptPassword`** — Raw mode may not be restored on unexpected errors.

6. **Duplicated ANSI constants** — Same color codes defined in both `profile.ts` and `profile-manager.ts`.

7. I would like to see a happy path test against a real realm server. consider leveraging testing infra from the packages/realm-server/tests so that we can spin up a real realm server that we can test against and assert that auth works against an actual realm server.

**Phase 2 alignment** (CS-10642): The `Profile` interface and `async` method signatures are well-positioned for extension with server tokens, per-realm JWTs, and auto-refresh. The `getActiveCredentials()` method is the natural evolution point for the programmatic `BoxelAuth` API. Inline comment on the interface details what fields will be needed.

habdelra:
@FadhlanR you still did not address my point about having a test against a real realm server

habdelra:
it sounds like @jurgenwerk is working on this. can you merge his branch into this once he gets to a good place and then include those tests?

jurgenwerk:
Yeah https://github.com/cardstack/boxel/pull/4368 should have the test helpers for spinning up a realm server so that the CLI can test against it. The mentioned PR is not finished yet but you can check and see if you can make it work in this branch. That part should already be functional as I am seeing my tests pass in the CI

FadhlanR:
Currently the `boxel profile` command only stores the username and password in `~/.boxel-cli/profiles.json` without doing any authentication to the realm server or realm, so I don't think a test against a real realm server is needed for this command, unless we want to change the behavior to perform authentication (e.g., store a JWT) as part of this PR.

I think we can merge this PR as-is so we can unblock other boxel CLI command PRs, including Matic's PR that introduces authentication against the realm server. We will change the auth behavior in CS-10642, the happy path test against a real realm server would fit naturally there, when the profile  command evolves to manage tokens. What do you think @habdelra?

habdelra:
that seems fair. we can tackle dealing with tokens in CS-10642

habdelra (approved):

PR #4355, opened 2026-04-08
CS-10616: Reimplement boxel workspace pull command

FadhlanR:
## Summary
- Port `boxel workspace pull` command from standalone boxel-cli into monorepo
- Register under `boxel realm pull` subcommand
- Add comprehensive test infrastructure: mock fetch router, mock Matrix/Realm servers

## Test plan
- [x] `pnpm --filter @cardstack/boxel-cli test` — 31 tests pass
- [x] `pnpm --filter @cardstack/boxel-cli build` — builds successfully


habdelra:
now that we have real realm integration don't forget to add tests for this against a running realm

jurgenwerk (approved):

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -353,11 +353,59 @@ export class ProfileManager {
...
+
+  private async resolveTokenForUrl(url: string): Promise<string> {
+    // Check if URL matches a cached realm token
+    let realmToken = this.findRealmTokenForUrl(url);
+    if (realmToken) {
+      return realmToken;
+    }
+
+    // Fetch all realm tokens and try again
+    await this.fetchAndStoreAllRealmTokens();
```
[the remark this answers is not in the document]
jurgenwerk: Agree with the codex's comment, I think this is problematic and needs fixing

packages/boxel-cli/src/lib/profile-manager.ts
```diff
@@ -353,11 +353,59 @@ export class ProfileManager {
...
+    realmToken = this.findRealmTokenForUrl(url);
+    if (realmToken) {
+      return realmToken;
+    }
+
+    // Fall back to server token for server-level endpoints
+    return this.getOrRefreshServerToken();
```
jurgenwerk: I think `resolveTokenForUrl` is confusing since the current implementation can return either a server token or a realm token. Perhaps it would be clearer to have `authedRealmFetch` and `authedRealmServerFetch` to clearly separate on which level we want to fetch, and in each one decide what we want to do on `401`.

packages/boxel-cli/src/lib/profile-manager.ts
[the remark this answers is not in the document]
FadhlanR: Fixed in 0f3051f. Split `authedFetch` into `authedRealmFetch` and `authedRealmServerFetch`. The realm-auth prefetch in `getRealmTokenForUrl` now catches errors gracefully so the 401 retry path still works when the server token is expired.

packages/boxel-cli/src/commands/realm/pull.ts
```diff
@@ -0,0 +1,205 @@
...
+            }
+          } catch (error) {
+            this.hasError = true;
+            console.error(`Error deleting local file ${relativePath}:`, error);
+          }
+        }
+      }
+    }
+
+    if (!this.options.dryRun && downloadedFiles.length > 0) {
```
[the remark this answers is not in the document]
FadhlanR: Fixed in 0f3051f. Added `deletedFiles` tracking and a separate checkpoint branch for delete-only pulls.

packages/boxel-cli/src/lib/profile-manager.ts
FadhlanR: Done in 0f3051f. Replaced `resolveTokenForUrl` + `authedFetch` with `authedRealmFetch` (for realm URL operations) and `authedRealmServerFetch` (for server-level endpoints like `_create-realm`). Each method handles its own 401 retry logic independently.

packages/boxel-cli/src/lib/realm-sync-base.ts
habdelra, line 1: Are we really reimplementing this? this looks almost exactly the same as the old implementation

Image

packages/boxel-cli/src/commands/realm/pull.ts
```diff
@@ -0,0 +1,228 @@
...
+    const downloadedFiles: string[] = [];
+    for (const [relativePath] of remoteFiles) {
+      try {
+        const localPath = path.join(this.options.localDir, relativePath);
+        await this.downloadFile(relativePath, localPath);
+        downloadedFiles.push(relativePath);
+      } catch (error) {
+        this.hasError = true;
+        console.error(`Error downloading ${relativePath}:`, error);
+      }
```
habdelra, line 71: this is downloading files seriallly? why? we should be downloading files concurrently

packages/boxel-cli/src/commands/realm/pull.ts
```diff
@@ -0,0 +1,228 @@
...
+
+        for (const relativePath of filesToDelete) {
+          try {
+            const localPath = localFiles.get(relativePath);
+            if (localPath) {
+              await this.deleteLocalFile(localPath);
+              deletedFiles.push(relativePath);
+              console.log(`  Deleted: ${relativePath}`);
+            }
+          } catch (error) {
```
habdelra, line 112: why are we doing these batch operations serially? we should be using atomic operations for this

packages/boxel-cli/src/commands/realm/pull.ts
```diff
@@ -0,0 +1,228 @@
...
+        );
+        const preDeleteCheckpoint = checkpointManager.createCheckpoint(
+          'remote',
+          deleteChanges,
+          `Pre-delete checkpoint: ${filesToDelete.size} files not on server`,
+        );
+        if (preDeleteCheckpoint) {
+          console.log(
+            `\nCheckpoint created before deletion: ${preDeleteCheckpoint.shortHash}`,
+          );
```
habdelra, line 97: where are the tests for the checkpoint stuff here?

packages/boxel-cli/src/lib/checkpoint-manager.ts
habdelra, line 1: this needs tests

habdelra:
so per our discussion with @lukemelia yes, we are copying this over and making targetted adjustments as we see them. sorry for the confusion!

PR #4356, opened 2026-04-08
CS-10617: Reimplement boxel workspace push command

FadhlanR:
## Summary
- Port `boxel realm push` command from standalone boxel-cli into monorepo
- Manifest-based incremental sync via `.boxel-sync.json` with MD5 hashing
- Options: `--delete` (remove remote-only files), `--dry-run`, `--force` (upload all)
- Protected files (`.realm.json`) are never uploaded or deleted

## Test plan
- [x] `pnpm --filter @cardstack/boxel-cli test` — 38 tests pass
- [x] `pnpm --filter @cardstack/boxel-cli build` — builds successfully (101.1 KB)


habdelra:
now that we have real realm integration don't forget to add tests for this against a running realm

packages/boxel-cli/tests/integration/realm-push.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    // Modify only one file
+    writeLocalFile(localDir, 'a.gts', 'export const a = 999;\n');
+
+    // Second push should only upload the changed file
+    await pushCommand(localDir, realmUrl, { profileManager });
+
+    let remoteFiles = await pullRemoteFiles(realmUrl);
+    expect(remoteFiles.get('a.gts')).toContain('a = 999');
+    expect(remoteFiles.get('b.gts')).toContain('b = 2');
```
habdelra: let's not verify this way--lets actually look at the server. beacuse you are verifying in the same local dir you wrote too, this doesn't demonstraibly prove that you actually mutated server state. also it's awkward that the push command relies on teh pull command for testing.

packages/boxel-cli/tests/integration/realm-push.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    let localDir = makeLocalDir();
+
+    writeLocalFile(localDir, 'file.gts', 'export const v = 1;\n');
+
+    await pushCommand(localDir, realmUrl, { profileManager });
+
+    // Push again with force (no changes) — should still succeed
+    await pushCommand(localDir, realmUrl, { force: true, profileManager });
+
+    let remoteFiles = await pullRemoteFiles(realmUrl);
```
habdelra: ditto--assert realm server state directly

packages/boxel-cli/tests/integration/realm-push.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    // Push two files
+    writeLocalFile(localDir, 'keep.gts', 'export const keep = true;\n');
+    writeLocalFile(localDir, 'remove.gts', 'export const remove = true;\n');
+    await pushCommand(localDir, realmUrl, { profileManager });
+
+    // Delete one locally, then push with --delete
+    fs.unlinkSync(path.join(localDir, 'remove.gts'));
+    await pushCommand(localDir, realmUrl, { delete: true, profileManager });
+
+    let remoteFiles = await pullRemoteFiles(realmUrl);
```
habdelra: ditto--assert realm server state directly

packages/boxel-cli/tests/integration/realm-push.test.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    writeLocalFile(localDir, 'draft.gts', 'export const draft = true;\n');
+
+    await pushCommand(localDir, realmUrl, { dryRun: true, profileManager });
+
+    // No manifest should be created
+    expect(fs.existsSync(path.join(localDir, '.boxel-sync.json'))).toBe(false);
+
+    // Remote should not have the file (only index.json from realm creation)
+    let remoteFiles = await pullRemoteFiles(realmUrl);
```
habdelra: ditto--assert realm server state directly

packages/boxel-cli/tests/integration/realm-push.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    writeLocalFile(localDir, 'card.gts', 'export const card = true;\n');
+    writeLocalFile(
+      localDir,
+      '.boxel-sync.json',
+      '{"workspaceUrl":"old","files":{}}',
+    );
+
+    await pushCommand(localDir, realmUrl, { profileManager });
+
+    let remoteFiles = await pullRemoteFiles(realmUrl);
```
habdelra: ditto--assert realm server state directly

packages/boxel-cli/tests/integration/realm-push.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    expect(fs.existsSync(path.join(localDir, '.boxel-sync.json'))).toBe(true);
+    let manifest = JSON.parse(
+      fs.readFileSync(path.join(localDir, '.boxel-sync.json'), 'utf8'),
+    );
+    expect(manifest.workspaceUrl).toBe(realmUrl);
+    expect(Object.keys(manifest.files)).toContain('card.gts');
+    expect(Object.keys(manifest.files)).toContain('data.json');
+
+    // Verify files exist on remote via pull
+    let remoteFiles = await pullRemoteFiles(realmUrl);
```
habdelra: ditto--assert realm server state directly

packages/boxel-cli/src/commands/realm/push.ts
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    if (filesToUpload.size === 0) {
+      console.log('No files to upload - everything is up to date');
+    } else {
+      console.log(`Uploading ${filesToUpload.size} file(s)...`);
+
+      for (const [relativePath, localPath] of filesToUpload) {
+        try {
+          await this.uploadFile(relativePath, localPath);
```
habdelra: this is doing the uploads serially. I thin kit might be better to use an atomic write, or something that is able to deal with changes concurrently

packages/boxel-cli/src/commands/realm/push.ts
```diff
@@ -0,0 +1,273 @@
...
+  const manifestPath = path.join(localDir, '.boxel-sync.json');
+  fs.writeFileSync(manifestPath, JSON.stringify(manifest, null, 2));
+}
+
+interface PushOptions extends SyncOptions {
+  deleteRemote?: boolean;
+  force?: boolean;
+}
+
+class RealmPusher extends RealmSyncBase {
```
habdelra, line 100: I'm super confused--i thought we were reimplementing this, but RealmSyncBase is already in the repo? which commit added that? we are we not reimplementing that too?

habdelra:
Are we really re-implementing this? i just compared your changes to what is currently in boxel-cli repo and they are the same. re-implementing is not the same as copying. i'm confused what's going on. Let's talk about this in office hours

packages/boxel-cli/src/commands/realm/push.ts
FadhlanR: RealmSyncBase is added here: https://github.com/cardstack/boxel/commit/72d1f10e46a4bbb8253d1416a2bb5460551e8281#diff-85179b7b382ff103db64d2324166436d7ef6db399991b68fea86fe9a08376a73

habdelra: why is your implementation of this so close to the existing boxel-cli?

habdelra:
> Are we really re-implementing this? i just compared your changes to what is currently in boxel-cli repo and they are the same. re-implementing is not the same as copying. i'm confused what's going on. Let's talk about this in office hours

so per our discussion with @lukemelia yes, we are copying this over and making targetted adjustments as we see them. sorry for the confusion!

FadhlanR:
Thanks for the review @habdelra! Wanted to summarize the changes that landed in response.

**Refactoring `pull` and `push`** (`14057e8cba`)

`RealmPuller`, `RealmPusher`, `RealmSyncBase`, and `CheckpointManager` were a mix of `async` outer methods wrapping synchronous internals — `fs.readFileSync`/`writeFileSync`/`mkdirSync`/`unlinkSync` and `spawnSync` for every git call — and the file-by-file work was issued serially inside `for` loops, so a pull/push of N files was always N sequential round trips. Now:

- All file-system calls go through `fs/promises`. All `spawnSync` calls in `CheckpointManager` are replaced with a promisified `spawn` wrapper.
- Independent work runs in parallel via `Promise.all`: downloads, uploads, deletes, hash computations, recursive directory scans (local *and* remote), and `CheckpointManager`'s file-sync-to-history / restore sweeps.
- Git operations against the same repo stay sequential to protect shared state — only the surrounding fs fan-out is parallelized.
- A small bug in `detectCurrentChanges()` came out of testing: `.trim()` on `git status --porcelain` output was stripping the leading space of single-line unstaged-modification entries, dropping the first character of the file name. Replaced with trailing-newline stripping.
- The internal `[init]` bootstrap commit created by `CheckpointManager.init()` is now filtered out of `getCheckpoints()` so it does not surface as a user-visible checkpoint.

**Tests and how we assert push results** (`5fb4eaaa73`)

`CheckpointManager` had no dedicated test file; added `tests/lib/checkpoint-manager.test.ts` with 28 unit tests covering init, change detection (including dotfile / `node_modules` / `.boxel-history` / `.boxel-sync.json` exclusions), checkpoint creation and classification rules, custom messages, source-tag round-trip, history retrieval, restore (including protected-file preservation), and milestones (by hash and by index).

`realm-pull.test.ts` grew from 5 → 11 scenarios. Existing scenarios now also assert `CheckpointManager` state (count, `source: 'remote'`, message format, custom pre-delete message). New flag-matrix scenarios cover `--delete --dry-run`, `--delete` with nothing to delete, idempotent re-pull, auto-created local directory, and overwrite-on-local-modification.

`realm-push.test.ts` verification is now completely rewritten in line with your feedback:

- The `pullRemoteFiles` helper is gone, along with the `pullCommand` import. Push tests no longer depend on pull working.
- Three new helpers — `listRemoteFiles`, `fetchRemoteFile`, `remoteFileExists` — query the realm endpoints directly via `profileManager.authedRealmFetch` (`Accept: application/vnd.api+json` for directory listings, `Accept: application/vnd.card+source` for file content). Every "did this push actually mutate server state?" assertion now reads from the server, not from a temp directory.
- Every scenario also asserts against the on-disk `.boxel-sync.json` manifest (`realmUrl`, file set, 32-char md5 hashes, that `.boxel-sync.json` and `.realm.json` are never listed) and against the `CheckpointManager` state (count, `source: 'local'`, message format).
- New scenarios mirror what the pull suite covers: re-push idempotency, recursive subdirectory upload, `.realm.json` protection.
- Added flag-combination scenarios for every meaningful pair/triple of `--force`, `--delete`, `--dry-run`. Each captures the manifest contents byte-for-byte before/after to verify dry-run leaves them untouched.
- Surfaced the `CheckpointManager` contract via two paired tests: `--force` re-uploads bytes to the server, but only produces a new commit when the workspace state actually differs. Re-pushing identical content with `--force` does not duplicate the checkpoint; pushing modified content with `--force` does.

**Atomic batch push + drift detection + manifest validation** (`de7cb0a4ea`)

Picking up on your "atomic write / deal with changes concurrently" hint and the two codex-bot items still open on push (remote drift and manifest-schema crash):

- `RealmSyncBase.uploadFilesAtomic` bundles every upload into a single `POST /_atomic` request. Operations use `op: add` for files new to our sync manifest and `op: update` for files we previously tracked, so the realm's batch validation surfaces concurrent creation (409) and concurrent deletion (404) as errors instead of silent clobbers. With `--force` or when no manifest exists, we defer to the actual remote state to avoid spurious conflicts; otherwise intent-based detection kicks in.
- Drift detection before upload: we now fetch `initialRemoteFiles` (reused between the access test and `--delete` sweep — one fewer round trip) and `_mtimes` in parallel, and a tracked file re-uploads when `localChanged || remoteMissing || remoteMtimeChanged`. `SyncManifest` gained an optional `remoteMtimes: Record<string, number>` so each push can persist what the server reported and the next push can notice out-of-band edits.
- Concurrent-write visibility: when drift is detected (and `--force` is not set), `console.warn` lists the paths the local push is about to overwrite so the user sees the clobber.
- On batch failure we set `hasError`, leave the old manifest untouched (retry re-attempts the same upload), and print actionable per-file hints like `<path> was created on the realm concurrently — run with --force to overwrite.` and `<path> was removed from the realm concurrently — run with --force to re-create it from your local copy.`
- `loadManifest` now validates shape via `isValidManifest` (checks `realmUrl: string`, `files: Record<string, string>`, and the optional `remoteMtimes` shape). A parseable but malformed `.boxel-sync.json` returns `null` with a warning and falls back to a full upload instead of crashing with `TypeError`.

Five new integration scenarios (push test count is now 19): a spy verifies every push sends exactly one `/_atomic` request; `re-pushes a file that was deleted on the realm out-of-band`; `re-pushes a file that was edited on the realm out-of-band` (both capture the drift warning); `recovers from a malformed .boxel-sync.json instead of crashing`; and `fails cleanly when an out-of-band create causes an atomic 409` (asserts the realm still has the rival content, the manifest was not overwritten, and the error output names the conflict).

Full suite: 62 unit + 31 integration tests passing.

habdelra (approved):
This looks good, but please make sure to address all the co-pilot issues.

PR #4366, opened 2026-04-09
host: Remove font links that are also self-hosted

backspace:
[`fonts.css`](https://github.com/cardstack/boxel/blob/861c03bbbc975c9579c5d1567bbfa0325c32b044/packages/boxel-ui/addon/src/styles/fonts.css) in `boxel-ui` already includes these.

### DIFF
--- a/packages/host/app/index.html
+++ b/packages/host/app/index.html
@@ -13,6 +13,9 @@
       rel="stylesheet"
       href="{{rootURL}}assets/@cardstack/host.css"
     />
+    <link rel="preload" href="{{rootURL}}assets/fonts/IBMPlexSansVar-Roman.woff2" as="font" type="font/woff2" crossorigin />
+    <link rel="preload" href="{{rootURL}}assets/fonts/IBMPlexMono-Regular.woff2" as="font" type="font/woff2" crossorigin />
+    <link rel="preload" href="{{rootURL}}assets/fonts/IBMPlexMono-Bold.woff2" as="font" type="font/woff2" crossorigin />
     <style>
       /* Allow prerendered cards to scroll while card-specific CSS loads */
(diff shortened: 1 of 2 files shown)


packages/host/app/index.html
```diff
@@ -13,18 +13,15 @@
...
+    <link rel="preload" href="{{rootURL}}assets/fonts/IBMPlexMono-Regular.woff2" as="font" type="font/woff2" crossorigin />
+    <link rel="preload" href="{{rootURL}}assets/fonts/IBMPlexMono-Bold.woff2" as="font" type="font/woff2" crossorigin />
     <style>
       /* Allow prerendered cards to scroll while card-specific CSS loads */
       #boxel-isolated-start + .boxel-card-container {
         overflow-y: auto;
       }
     </style>
-    <link rel="preconnect" href="https://fonts.googleapis.com" />
-    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
```
lukemelia, line 23: I would have guessed these provide performance improvements for loading fonts from Google Fonts. How do you know they are not helpful?

burieberry: Hmm I wonder if these provide benefits for fonts imported via themes, even though they're using `@import` built-into the card-container when there's a `cssImports` argument.

backspace: 👍🏻 I didn’t think about this, thanks, restored now

burieberry: The answer is, yes, those lines do somewhat help shave off some connection latency, even though it's better if `@import` is avoided altogether. Self-hosting fonts and using `@font-face` rules still seems to be the recommended way over others.

burieberry (approved):

PR #4374, opened 2026-04-09
Add string-to-content serializer for composite field migration

lukemelia:
## Summary

- Adds a generic `string-to-content` built-in serializer that normalizes plain string values to `{ content: <string> }` during deserialization, enabling migration from primitive `StringField`-based fields to composite `FieldDef`-based fields
- Extends `fieldSerializer` support to composite fields in both `Contains` and `ContainsMany` deserialization paths (previously gated behind `primitive in this.card`)
- Locks down `[deserialize]` override on composite fields with a runtime check — composite fields must use registered serializers instead of overriding deserialization directly
- Wires up `RichMarkdownField` with `static [fieldSerializer] = 'string-to-content'` so it can accept data in both old `MarkdownField` format (plain string) and new format (object with `content` property)

## Test plan

- [ ] Deserialize `RichMarkdownField` from a plain string (migration path)
- [ ] Deserialize `RichMarkdownField` from an object with `content` property (normal path)
- [ ] Deserialize `null` value for field with `string-to-content` serializer
- [ ] Verify composite field with `[deserialize]` override throws runtime error
- [ ] Verify existing serialization tests still pass

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -509,7 +509,7 @@ function cardTypeFor(
 function assertNoDeserializeOverride(cardClass: typeof BaseDef) {
   if (
     !(primitive in cardClass) &&
-    cardClass.hasOwnProperty(deserialize)
+    Object.prototype.hasOwnProperty.call(cardClass, deserialize)
   ) {
     throw new Error(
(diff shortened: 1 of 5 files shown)


habdelra (approved):
looks nice!

PR #4377, opened 2026-04-10
Submission workflow UX improvement of CI checks

richardhjtan:
This is to fix the flickering issue on the CI checks section, where it flashes back and forth from the empty state to the pipelines running state when the query gets new cards.

**Before**

https://github.com/user-attachments/assets/43b7c7dd-9009-4eef-9f75-9261904bc430

**After**

https://github.com/user-attachments/assets/92a48df0-101a-4d50-83a1-10a3976fedbe

### DIFF
--- a/packages/catalog-realm/pr-card/components/isolated/ci-section.gts
+++ b/packages/catalog-realm/pr-card/components/isolated/ci-section.gts
@@ -1,4 +1,5 @@
 import GlimmerComponent from '@glimmer/component';
+import { cached } from '@glimmer/tracking';
 import type { CiStatus, CiGroup } from '../../utils';
 
 // ── Sub-components ──────────────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


backspace (approved):

packages/catalog-realm/pr-card/components/isolated/ci-section.gts
```diff
@@ -144,31 +144,41 @@ class CiStatusLabel extends GlimmerComponent<CiStatusLabelSignature> {
 interface CiSectionSignature {
   Args: {
     ciGroups: CiGroup[];
+    isLoading?: boolean;
   };
 }
 
 export class CiSection extends GlimmerComponent<CiSectionSignature> {
+  get flatItems() {
```
habdelra: perhaps you should use @cached so we don't rebuild this everytime this.flatItems is consumed, and only if ciGroups actually changes

lucaslyl (approved):

packages/catalog-realm/pr-card/components/isolated/ci-section.gts
richardhjtan: good catch thanks @habdelra

PR #4386, opened 2026-04-13
Hide Create Listing for index cards in inspector and playground menu

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10584/shouldnt-create-listing-from-indexjson-when-inside-code-mode

1. Replaces filename-based guard in operator-mode detail panel with card-identity check (card.id === ${realmURL}index).

2. Introduces shared helpers in runtime-common:

- isRealmIndexCardId(cardId, realm)
- isRealmIndexCard(card)

3. Refactors duplicate index-card logic to use shared helper across:

- base menu items
- operator-mode detail panel
- operator-mode stack item
- operator-mode state service

4. Removes local ad-hoc isIndexCard implementations and keeps behavior consistent across code paths.

- Keeps intended behavior intact: only the actual realm root index card is excluded from Create Listing; non-root cards in index.json paths remain eligible.

Before:
image

After:
image


packages/runtime-common/code-ref.ts
```diff
@@ -138,6 +140,22 @@ export function isListingInstance(card: any): boolean {
...
+  }
+  let realmHref = typeof realm === 'string' ? realm : realm.href;
+  return cardId === `${realmHref}index`;
+}
+
+export function isRealmIndexCard(card: CardDef | undefined): boolean {
+  let cardId = typeof card?.id === 'string' ? card.id : undefined;
+  return isRealmIndexCardId(cardId, card?.[realmURL]);
+}
+
```
habdelra: seems like a weird home for this. maybe @cardstack/runtime-common/index would be better?

packages/runtime-common/code-ref.ts
```diff
@@ -138,6 +139,22 @@ export function isListingInstance(card: any): boolean {
...
 
+export function isRealmIndexCardId(
+  cardId: string | undefined,
+  realm: string | URL | undefined,
+): boolean {
+  if (!cardId || !realm) {
+    return false;
+  }
+  let realmHref = typeof realm === 'string' ? realm : realm.href;
+  return cardId === `${realmHref}index`;
```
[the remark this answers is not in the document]
habdelra: this is a legit comment, and actulaly the better way to do this is to use the ReamPaths support that we have. only specifically you'd build a file path for index using RealmPaths

packages/runtime-common/code-ref.ts
lucaslyl: Good concern! I moved these helpers out of code-ref.ts, but kept the implementation in a dedicated realm-index-card.ts module and re-exported from index.ts, since index.ts is the package entrypoint and export surface (with some shared declarations), while feature logic is easier to maintain in focused modules.

richardhjtan (approved):

PR #4400, opened 2026-04-14
Fix cross-realm linksTo MIME type handling for file/image defs

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10730/card-thumbnail-linksto-imagedeffiledef-fails-to-save-for-cross-realm

Previously, cross-realm linksTo requests for file-def and image-def could send an incorrect Accept header and fail with 415 Unsupported Media Type.(from console)
image

## Changes:

1. In loadLinks, use expectsCard as the primary signal for cross-realm links.
2. If expectsCard is true, fetch with card MIME type.
3. Otherwise, fetch with file-meta MIME type.
4. Remove the card-only isSingleCardDocument assertion from this mixed cross-realm path to avoid false failures on valid file-meta responses.

## Result:

Cross-realm file-def/image-def links resolve correctly.
Eliminates 415 errors caused by wrong media type negotiation.

https://github.com/user-attachments/assets/714bcd40-597f-4aea-9e9a-dded0a8cab59

### DIFF
--- a/packages/realm-server/tests/indexing-test.ts
+++ b/packages/realm-server/tests/indexing-test.ts
@@ -1283,6 +1283,101 @@ module(basename(__filename), function () {
           );
         });
       });
+
+      module('cross-realm FileDef linksTo', function (hooks) {
+        let consumerRealm: Realm;
+
+        setupPermissionedRealmsCached(hooks, {
+          mode: 'before',
+          realms: [
+            // provider
+            {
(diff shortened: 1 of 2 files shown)


habdelra:
Abandoned PR

habdelra:
Abandoned PR

habdelra:
Abandoned PR

PR #4401, opened 2026-04-14
Add linting for data-test attributes used outside of tests, and relevant fixes

burieberry:
we strip out the `data-test` attributes in prod, so do not use them for functionality outside of tests

- added linter in `template-lint` package
- updated `.eslintrc.js` for `host`, `boxel-ui/addon`
- remaining changes are the places where incorrect usage was caught by the linter

note: will handle `base` package separately because it turns out `pnpm lint` only lints the template there and not the js

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,4 +37,27 @@ module.exports = {
       { argsIgnorePattern: '^_', varsIgnorePattern: '^_' },
     ],
   },
+  overrides: [
+    {
+      // Disallow data-test-* CSS selectors in app code across all packages.
+      // ember-test-selectors strips these attributes in production, so selectors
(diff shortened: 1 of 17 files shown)


backspace:
I wonder if it’s possible to have a lint rule for this, non-test code should never refer to test selectors, other than the test selector specification itself.

burieberry:
> I wonder if it’s possible to have a lint rule for this, non-test code should never refer to test selectors, other than the test selector specification itself.

This was a good idea. Other incorrect usages were caught: https://github.com/cardstack/boxel/actions/runs/24466119158/job/71493545257?pr=4401

backspace:
Thanks for pivoting to that, the test failure does seem related to this, I’ll review more thoroughly when it’s passing

backspace (approved):

PR #4416, opened 2026-04-16
Cross file references and packaging upon submission

richardhjtan:
Fix cross-realm submission PR file collection and rewrite references to relative paths
image

### What is changing 
- Cross-realm auth (run-command.ts): Commands now include JWTs for all realms the user has access to (not just the target realm), fixing 401 errors when deserialising cards with cross-realm linksToMany relationships. Also injects accessibleRealms into commandInput so commands can resolve realm boundaries for any URL.
- Cross-realm path resolution (catalog.ts): ListingPathResolver.local() no longer throws on foreign realm URLs. Uses addKnownRealmURL() registration to strip the correct realm prefix, with fallback to full pathname for unregistered realms.
- Reference rewriting (collect-submission-files.ts): After collecting files, all absolute URLs pointing at submitted files are rewritten to relative paths (e.g. ../cycling-mileage-log). This makes the submission self-contained — adoptsFrom.module in card JSONs and imports in .gts files resolve correctly after the PR merges into the catalog. Example PR submission package output when cross-realm involved: https://github.com/cardstack/boxel-catalog/pull/395/changes
- Add tests

### DIFF
--- a/packages/catalog-realm/commands/collect-submission-files.ts
+++ b/packages/catalog-realm/commands/collect-submission-files.ts
@@ -18,6 +18,7 @@ import {
   CardDef,
   field,
   contains,
+  containsMany,
 } from 'https://cardstack.com/base/card-api';
 import StringField from 'https://cardstack.com/base/string';
 import GetCardCommand from '@cardstack/boxel-host/commands/get-card';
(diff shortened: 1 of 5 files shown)


habdelra (approved):

habdelra:
good catch--this looks like it was an oversight. currently the normal prerender path for indexing follows this approach as well in order to deal with dependencies in private realms that the user that owns this realm may have access too

PR #4420, opened 2026-04-16
Add Local environment option to `boxel profile add`

jurgenwerk:
When testing boxel-cli locally I needed an option to save a profile that works in a local setup. I think it would be beneficial to others too.

image


FadhlanR (approved):

habdelra:
Let's make sure to update the software-factory README.md to describe the end to end flow for running the factory. right now it's using env vars instead of relying on the boxel-cli profile

jurgenwerk:
@habdelra will add this in https://github.com/cardstack/boxel/pull/4426

PR #4421, opened 2026-04-16
CS-10677: Add `boxel run-command` CLI command

FadhlanR:
## Summary
- Add top-level `boxel run-command <specifier> --realm <url> [--input '{}'] [--json]` CLI command that executes host commands on the realm server via the `/_run-command` endpoint
- Export a programmatic `runCommand()` function for use by other CLI commands
- Unit tests covering success, error status, HTTP failures, network failures, missing profile, trailing slash handling, and malformed responses

## Test plan
- [ ] `pnpm vitest run` in `packages/boxel-cli` passes all new tests
- [ ] Manual test with a running realm server: `boxel run-command @cardstack/boxel-host/commands/get-card-type-schema/default --realm <url> --input '{"cardURL": "..."}'`


habdelra:
so the important thing here is that this command needs to use the realm server token (vs the realm specific token). my LLM is always getting these 2 types of tokens confused. how are you managing the server token?

packages/boxel-cli/src/commands/run-command.ts
```diff
@@ -0,0 +1,160 @@
...
+        realmURL: realmUrl,
+        command: commandSpecifier,
+        commandInput: options?.input ?? null,
+      },
+    },
+  };
+
+  let response: Response;
+  try {
+    response = await pm.authedRealmServerFetch(url, {
```
habdelra, line 52: how does this know to use the server token and not a realm specific token? and is there  a test that can prove that by making sure these two tokens are actually different in the real realm test. (i find that the LLM gets confused here and tries to make these 2 the same token all the time)

docs/cs-10677-run-command-plan.md
habdelra, line 1: i don't think you need to add this file to the repo

packages/boxel-cli/src/commands/run-command.ts
FadhlanR: Do you mean that the LLM itself determines which tokens should be used? If so, it is not needed, because we store the realm server token and realm-specific tokens in the profile.json file, since they are generated after the LLM executes the boxel-profile command. Additionally, there are two fetch functions in the Boxel CLI, authedRealmServerFetch and authedRealmFetch, which are used by the commands to retrieve data using one of those two tokens, depending on the command.

habdelra: no. i mean the LLM that you are using to help you code this feature up. (assuming claude code). its really bad at telling the difference between server tokens and realm tokens. for me it's introduced dozens of bugs around this that have wasted many hours for me.

habdelra: ah i see--since you are using `authedRealmServerFetch` that is how we know to chose the correct token :+1:

habdelra (approved):

PR #4422, opened 2026-04-16
Replace pullRealmFiles with BoxelCLIClient.pull()

jurgenwerk:
## Summary
- Add programmatic `pull()` method to `BoxelCLIClient` that reuses the existing `RealmPuller` (auth via active profile, concurrent downloads, checkpoints) and returns `{ files, error? }` instead of calling `process.exit()`
- Replace the 90-line HTTP implementation of `pullRealmFiles` in `realm-operations.ts` with a 3-line delegation to the new client method
- Remove 5 unit tests that tested the old HTTP internals (now boxel-cli's responsibility)

## Test plan
- [x] All 360 software-factory node tests pass
- [x] boxel-cli type checks clean
- [ ] Verify factory entrypoint still works end-to-end with `pullRealmFiles` delegation


packages/boxel-cli/src/commands/realm/pull.ts
```diff
@@ -232,3 +233,46 @@ export async function pullCommand(
     process.exit(1);
   }
 }
+
+export async function pull(
```
FadhlanR, line 203: I think we can drop the `pullCommand` function and use this function in `registerPullCommand` and move the `console.log` and `process.exit` that previously handled in pullCommand to `registerPullCommand.`

habdelra (approved):
make sure to address the co-pilot issues, but i think it looks good. it puts us in a good place to tackle https://linear.app/cardstack/issue/CS-10520/software-factory-delegates-all-boxel-api-calls-to-boxel-cli

PR #4423, opened 2026-04-16
software factory: delegate all API calls to boxel-cli, which uses profiles for auth

jurgenwerk:
This is a refactor that moves all the realm / realm server api calls related logic from software factory to boxel-cli. Previously, software factory was making requests to the realm server directly, and now, all these calls have been delegated to `BoxelCLIClient`, which handles full 2-level auth management (realm server, and realms), using boxel-cli profiles.

Obviously this is a pretty huge refactor and thus hard to review. I tested end to end correctness of this refactor by running the factory using: 

```
pnpm factory:go \
  --brief-url http://localhost:4201/software-factory/Wiki/sticky-note \
  --target-realm-url http://localhost:4201/user/my-test-realm/ \
  --debug
```
  
The factory run worked OK and produced the desired result without any problems:

image


packages/boxel-cli/src/commands/realm/pull.ts
```diff
@@ -232,3 +233,46 @@ export async function pullCommand(
     process.exit(1);
   }
 }
+
+export async function pull(
```
FadhlanR, line 203: I think we can drop the `pullCommand` function and use this function in `registerPullCommand` and move the `console.log` and `process.exit` that previously handled in `pullCommand` to `registerPullCommand`.

habdelra:
Can you also use claude code to update the phase-2-plan.md and the linear ticket CS-10520 to make sure that all client.read and client.write in the target realm for the software factory are removed, and instead replaced by local file system read and local file system write since we'll be using boxel-cli push/pull synchronization mechanisms.

habdelra (approved):
awesome to see all the code that's removed in this PR, thanks!

PR #4426, opened 2026-04-16
Remove MATRIX_* var auth from software-factory and use boxel-cli profile instead

jurgenwerk:
This PR removes the dual mode of authenticating with both MATRIX_* vars and boxel cli profile, and refactors software factory to rely on boxel-cli profile only.

Note that before executing the following scenarions, I ran `unset MATRIX_URL MATRIX_USERNAME MATRIX_PASSWORD REALM_SERVER_URL` to make sure no ENV vars influence the software factory run.

# Scenario 1: No active profile - the error will say an active profile is needed:
image

# Scenario 2: The currently active profile does not match target realm url - the error will say there is a realm server profile mismatch:
image

image

# Scenario 3: The currently active profile matches target realm url  - factory will run 🚀
image

image


habdelra (approved):
please also have claude code update the phase-2-plan.md with this detail. i think the plan mentions these env vars a few times

jurgenwerk:
@habdelra done!

PR #4428, opened 2026-04-16
Change interfaces to use Realm[Resource]Identifier types

backspace:
This updates existing types and function interfaces to use `RealmIdentifier` (realm URLs and `@cardstack/whatever`) and `RealmResourceIdentifier` (card URLs/`@cardstack/base/string`); things that were previously assumed to be URLs can now be universal identifiers. Changes like [updating `ResolvedCodeRef.module` to use `RealmResourceIdentifier`](https://github.com/cardstack/boxel/pull/4428/changes#diff-f64300a7a2ae76c7c0e89646ff65217cb920e7ef2d05947f63fd85b4121a4c64R30) cascade outward to many parts of the application, so where identifiers enter the system, they are properly typed.

This is a change across many files, but very find-and-replace-like.


lukemelia (changes requested):
Good progress.

The PR introduces ~1,260 `as RealmResourceIdentifier` casts and leaves the 72 `// @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS` comments in place. Let's add a couple of small helpers in `runtime-common` to collapse most of that.

### Proposed helpers

Add to `packages/runtime-common/card-reference-resolver.ts` alongside the type defs:

```ts
export function rri(s: string): RealmResourceIdentifier {
  return s as RealmResourceIdentifier;
}

export function ri(s: string): RealmIdentifier {
  return s as RealmIdentifier;
}

export function codeRef(
  baseUrl: string,
  relativePath: string,
  name: string,
): { module: RealmResourceIdentifier; name: string } {
  return {
    module: new URL(relativePath, baseUrl).href as RealmResourceIdentifier,
    name,
  };
}
```

### Sample conversion — `packages/catalog-realm/crm-app/crm-app.gts`

**Before** (current PR):
```ts
const DEAL_CARD_REF = {
  name: 'Deal',
  // @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS
  module: new URL('./deal', import.meta.url).href as RealmResourceIdentifier,
};

const ACCOUNT_CARD_REF = {
  name: 'Account',
  // @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS
  module: new URL('./account', import.meta.url).href as RealmResourceIdentifier,
};

const TASK_CARD_REF = {
  name: 'CRMTask',
  // @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS
  module: new URL('./task', import.meta.url).href as RealmResourceIdentifier,
};

const CONTACT_CARD_REF = {
  name: 'Contact',
  // @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS
  module: new URL('./contact', import.meta.url).href as RealmResourceIdentifier,
};
```

**After**:
```ts
// @ts-expect-error import.meta is valid ESM but TS detects .gts as CJS
const here: string = import.meta.url;

const DEAL_CARD_REF = codeRef(here, './deal', 'Deal');
const ACCOUNT_CARD_REF = codeRef(here, './account', 'Account');
const TASK_CARD_REF = codeRef(here, './task', 'CRMTask');
const CONTACT_CARD_REF = codeRef(here, './contact', 'Contact');
```

Per file: 4 `ts-expect-error` → 1, 4 casts → 0, 22 lines → 7. Aside: the PR version of this particular file references `RealmResourceIdentifier` without importing it — `codeRef` masks that bug for free since the type reference disappears.

### Why this is worth a second pass

1. **`@ts-expect-error` duplication is the ugliest part of the diff.** 72 identical comments in the PR (100 in the codebase). Hoisting `const here: string = import.meta.url` with a single suppression per file turns that into ~file-count instead of ~use-count.
2. **`as X` casts are easier to miss than `fn(x)` calls in review.** Branded-type casts are semantically meaningful — wrapping them in `rri()` / `ri()` makes boundaries greppable.
3. **Tightening source types beats sprinkling casts.** Many of the remaining `as RealmResourceIdentifier` are secondary — e.g. `this.config.taskSource.module as RealmResourceIdentifier` only needs the cast because `FilterConfig.codeRef.module` is still `string`. Branding the fixture constants (`testRealmURL: RealmIdentifier`, etc.) and interface fields makes hundreds of casts evaporate structurally. The top-3 test-fixture template patterns (`` `${testRealmURL}test-cards` ``, etc.) alone account for ~200 casts.

### `@ts-expect-error` caveat

It can't be fully hidden — the error fires on the `import.meta` keyword itself in `.gts` files (under `module: node16`), regardless of what helper it's piped into. But one suppression per file is much better than one per call site, and the hoisted `here` binding becomes a convenient handle for everything else.

### Suggested ordering

You can do these things in a follow-up PR if you prefer:

1. Brand the fixture/config source types (e.g. `testRealmURL`, `FilterConfig.codeRef.module`, similar) (biggest structural win — deletes hundreds of casts for free).
2. Add `codeRef` helper + convert the `import.meta.url` call sites (kills the `@ts-expect-error` duplication).
3. Add `rri()` / `ri()` for the remaining literal cases (small, consistency).

backspace:
I’ve implemented 1 in f53b528, unfortunately because most of the uses of `testRealmURL` are in template strings, it doesn’t “delete[] hundreds of casts for free” because the brand is lost:

image

but definitely worth doing regardless, thanks!

Continuing to the next points

backspace:
To work around that template string problem, I added helpers that just return base and test realm wrappers that concatenate the paths with the realm bases and return with the expected type, which reduces the casts:

```
  Cast reduction over the PR:
  - Before any work: ~1,238 RRI casts + ~69 RI casts
  - After step 4: 82 RRI + 16 RI casts (mostly inside runtime-common itself, which legitimately needs them)
  - Removed: ~1,200 manual casts (~91% reduction)
```

This seems like an okay compromise?

lukemelia (approved):
👍

PR #4432, opened 2026-04-17
Split out commands so we can just import host commands from realm defined commands

tintinthong:


habdelra (commented):
probably we should have dedicated tests for the new lower level commands

richardhjtan:
I haven’t been able to fully test this in the branch environment because of an existing staging issue: the resolved path is preventing the card from loading properly, which causes Remix to fail.

Screenshot 2026-04-20 at 12 19 55 AM

---
As a baseline, the Remix test case and the other newly introduced command test cases should pass in this PR

tintinthong:
> I haven’t been able to fully test this in the branch environment because of an existing staging issue: the resolved path is preventing the card from loading properly, which causes Remix to fail.
> 
> Screenshot 2026-04-20 at 12 19 55 AM
> As a baseline, the Remix test case and the other newly introduced command test cases should pass in this PR

is this on local branch or staging?

tintinthong:
> I haven’t been able to fully test this in the branch environment because of an existing staging issue: the resolved path is preventing the card from loading properly, which causes Remix to fail.
> 
> Screenshot 2026-04-20 at 12 19 55 AM
> As a baseline, the Remix test case and the other newly introduced command test cases should pass in this PR

Remix definitely works during manual testing locally

richardhjtan (approved):

richardhjtan:
Works fine after clearing the module cache locally

PR #4439, opened 2026-04-20
fix: resolve prefix-form card IDs at entry points to prevent Invalid URL crashes across catalog workflows

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-10812/resolve-prefix-form-card-ids-at-entry-points-to-prevent-invalid-url

## Issues encountered with catalog-realm (prefix-form IDs):

- **spec-preview** — completely broken when viewing any catalog card
- **Create card / Create listing** — crashed on save, card never appeared in store
- **Delete card** — crash trying to build the delete request URL
- **Remix feature** — crashed opening the remix panel for a catalog card
- **View card / click to open** — crashed navigating to any catalog card
-  And more to go...

Reference: [CS-10654 — Invalid URL issue present in multiple places](https://linear.app/cardstack/issue/CS-10654/invalid-url-issue-present-in-multiple-places)

## Root cause

There were **five distinct entry points** where prefix-form IDs entered the frontend without being resolved. No single fix covered all of them — each entry point bypassed the others.

## Fix strategy

Rather than adding guards in every affected component (dozens of call sites), the fix resolves prefix form **at the highest possible upstream point** for each entry path. Downstream components like `spec-preview` require no changes — they keep their existing `new URL(card.id)` calls, which now always receive a valid absolute URL.

**All fixes are frontend-only, in-memory. The database is never touched — it intentionally keeps prefix form for portability.**

## Fixes

| # | File | Entry point fixed | Why this location |
|---|---|---|---|
| 1 | `virtual-network.ts` | Prefix string passed as a `fetch()` URL | Every network call goes through here; intercept before `new Request()` is built |
| 2 | `store.ts` `createFromSerialized` | `resource.id` / `included[].id` from server HTTP response | Normalize before card API hydrates JSON into a card instance — `card.id` is always absolute after this |
| 3 | `store.ts` `persistAndUpdate` | `json.data.id` returned by server after save | Normalize immediately after save response; prevents false-positive `needsServerStateMerge` and wrong `api.setId` |
| 4 | `code-ref.ts` `identifyCard` | `codeRef.module` from Loader's internal identity map | Loader stores module paths as prefix internally; resolve once here so all 8+ callers automatically get absolute module paths |
| 5 | `operator-mode-state-service.ts` `deserialize` | Trail card IDs read from localStorage / URL hash | Persisted state bypasses all runtime fixes above; `normalizeTrailItem` resolves on read, drops corrupt entries |
| 6 | `routes/module.ts` `buildModuleModel` | Module `id` from browser URL query string | Route parameter enters the app before any store or Loader processing; `cardIdToURL(id)` replaces `new URL(id)` |


lucaslyl:
@copilot apply changes based on the comments in [this thread](https://github.com/cardstack/boxel/pull/4439#pullrequestreview-4137854214)

lucaslyl:
@codex comment

lucaslyl:
@backspace @habdelra  We found that prefix-form card IDs are crashing a bunch of catalog features. I fixed it by resolving the IDs at the entry points rather than adding guards everywhere. Can you review and let me know if this approach makes sense? Are there any entry points I missed?

lukemelia:
@lucaslyl This seems to be the opposite approach to what @backspace is working toward. The vision of https://linear.app/cardstack/project/cardstackbase-defea0d67319/overview is to make the prefix form first-class and keep it use for everything except for network communication itself.

backspace:
> @lucaslyl This seems to be the opposite approach to what @backspace is working toward. The vision of https://linear.app/cardstack/project/cardstackbase-defea0d67319/overview is to make the prefix form first-class and keep it use for everything except for network communication itself.

yes, my current task is to undo some steps in the wrong direction; the goal is to keep `@cardstack/whatever` unresolved until the network layer, you can see with the scattering of `resolveCardReference` everywhere that it’s a mess at the moment. I’m sorry that the `@cardstack/catalog` migration left some cases unaddressed, but this refactoring should fix these problems at a deeper level.

Maybe pointing your LLM at the detailed plan Luke linked to will help provide mitigation in the meantime? I’m at phase 2, so most of what’s merged at the point hasn’t changed anything substantial.

lucaslyl:
> > @lucaslyl This seems to be the opposite approach to what @backspace is working toward. The vision of https://linear.app/cardstack/project/cardstackbase-defea0d67319/overview is to make the prefix form first-class and keep it use for everything except for network communication itself.
> 
> yes, my current task is to undo some steps in the wrong direction; the goal is to keep `@cardstack/whatever` unresolved until the network layer, you can see with the scattering of `resolveCardReference` everywhere that it’s a mess at the moment. I’m sorry that the `@cardstack/catalog` migration left some cases unaddressed, but this refactoring should fix these problems at a deeper level.
> 
> Maybe pointing your LLM at the detailed plan Luke linked to will help provide mitigation in the meantime? I’m at phase 2, so most of what’s merged at the point hasn’t changed anything substantial.

Thanks for walking me through this. Really helpful. I'll check it out.

habdelra:
Abandoned PR

PR #4473, opened 2026-04-22
CS-10847: Add `boxel file write` command

FadhlanR:
## Summary
- Adds `boxel file write <path> --realm <url> --content <str>` CLI command (also supports `--file <path>`)
- Extracts `BoxelCLIClient.write()` inline implementation into a standalone command function at `src/commands/file/write.ts`

## Why — pattern consistency
The `pull` command already follows this pattern: a standalone exported function that both the CLI and `BoxelCLIClient` delegate to. This PR makes `write()` consistent with that approach. Every `BoxelCLIClient` method should be backed by a command function that is the single source of truth — the client is a thin delegation layer. This enables CS-10670 (tool definitions) where boxel-cli publishes tool definitions that reference these commands directly.

## Test plan
- [x] Integration tests: writes .gts and reads back via realm endpoint, writes .json card and verifies, no-profile error
- [x] Build passes
- [ ] CI green

Closes CS-10847

### DIFF
--- a/packages/boxel-cli/src/commands/file/write.ts
+++ b/packages/boxel-cli/src/commands/file/write.ts
@@ -1,6 +1,9 @@
 import type { Command } from 'commander';
 import { readFileSync } from 'fs';
-import { getProfileManager, type ProfileManager } from '../../lib/profile-manager';
+import {
+  getProfileManager,
+  type ProfileManager,
+} from '../../lib/profile-manager';
 import { FG_GREEN, FG_RED, DIM, RESET } from '../../lib/colors';
 
 export interface WriteResult {
--- a/packages/boxel-cli/src/commands/file/write.ts
+++ b/packages/boxel-cli/src/commands/file/write.ts
@@ -41,9 +41,10 @@ export async function write(
   let pm = options?.profileManager ?? getProfileManager();
   let active = pm.getActiveProfile();
   if (!active) {
-    throw new Error(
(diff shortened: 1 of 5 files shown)


lukemelia (approved):
If this is how it already works, fine to keep, but if I was designing this, I'd make it so you provide the content via STDIN

packages/boxel-cli/src/commands/file/write.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+interface WriteCliOptions {
+  realm: string;
+  content?: string;
+  file?: string;
+  json?: boolean;
+}
+
+function ensureTrailingSlash(url: string): string {
```
jurgenwerk: There was a comment in the other PR about the redundancy of this function

packages/boxel-cli/src/commands/file/write.ts
```diff
@@ -0,0 +1,144 @@
...
+  path: string,
+  content: string,
+  options?: WriteCommandOptions,
+): Promise<WriteResult> {
+  let pm = options?.profileManager ?? getProfileManager();
+  let active = pm.getActiveProfile();
+  if (!active) {
+    return {
+      ok: false,
+      error: 'No active profile. Run `boxel profile add` to create one.',
```
jurgenwerk: This error string is repeated across a bunch of opened PRs. Is there a possibility to DRY or at least having a constant.

PR #4476, opened 2026-04-22
CS-10849: Add `boxel file list` command

FadhlanR:
## Summary
- Adds `boxel file list --realm <url> [--json]` CLI command (alias: `boxel file ls`)
- Extracts `BoxelCLIClient.listFiles()` inline implementation into a standalone command function at `src/commands/file/list.ts`

## Why — pattern consistency
The `pull` command already follows this pattern: a standalone exported function that both the CLI and `BoxelCLIClient` delegate to. This PR makes `listFiles()` consistent with that approach. Every `BoxelCLIClient` method should be backed by a command function that is the single source of truth — the client is a thin delegation layer. This enables CS-10670 (tool definitions) where boxel-cli publishes tool definitions that reference these commands directly.

Note: This lists files *within* a realm (via `_mtimes`), distinct from CS-10620 (`boxel realm list`) which lists the user's realms/workspaces.

## Test plan
- [x] Integration tests: seeds files, verifies filenames appear sorted, unreachable realm error, no-profile error
- [x] Build passes
- [ ] CI green

Closes CS-10849

### DIFF
--- a/packages/boxel-cli/src/commands/file/list.ts
+++ b/packages/boxel-cli/src/commands/file/list.ts
@@ -111,9 +111,7 @@ export function registerListCommand(file: Command): void {
         for (let filename of result.filenames) {
           console.log(`${DIM}${filename}${RESET}`);
         }
-        console.log(
-          `\n${DIM}${result.filenames.length} file(s)${RESET}`,
-        );
+        console.log(`\n${DIM}${result.filenames.length} file(s)${RESET}`);
       }
     });
 }
--- a/packages/boxel-cli/src/commands/file/list.ts
+++ b/packages/boxel-cli/src/commands/file/list.ts
@@ -1,8 +1,11 @@
 import type { Command } from 'commander';
 import {
   getProfileManager,
+  NO_ACTIVE_PROFILE_ERROR,
   type ProfileManager,
 } from '../../lib/profile-manager';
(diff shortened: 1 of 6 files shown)


packages/boxel-cli/src/commands/file/list.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+interface ListFilesCliOptions {
+  realm: string;
+  json?: boolean;
+}
+
+function ensureTrailingSlash(url: string): string {
+  return url.endsWith('/') ? url : `${url}/`;
+}
```
habdelra: we should be using runtime-common for this. can you audit the boxel-cli for all the cases that we reimplemented it? i found that the agent loves to reimplement this.

packages/boxel-cli/tests/integration/file-list.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(result.filenames).toEqual(sorted);
+  });
+
+  it('returns error for an unreachable realm', async () => {
+    let result = await listFiles('http://127.0.0.1:1/', { profileManager });
+    expect(result.filenames).toEqual([]);
+    expect(result.error).toBeDefined();
+  });
+
+  it('returns error result when no active profile', async () => {
```
jurgenwerk, line 60: This test is repeated in a bunch of PRs. Is there an opportunity to DRY?

PR #4478, opened 2026-04-22
CS-10643/CS-10850: Add `boxel search` command (federated)

FadhlanR:
## Summary
- Adds top-level `boxel search --realm <url> --query <json>` for federated cross-realm search via `_federated-search`
- `--realm` is repeatable; results are merged server-side
- `BoxelCLIClient.search()` delegates to the standalone command function (matches `file/read`/`file/write` patterns)

## Why
Per review feedback, having two search commands (per-realm and federated) was confusing for the agent. This PR consolidates into a single `boxel search` that is always federated. For per-realm content access, the agent uses `boxel realm pull` + grep.

## Notable changes since the original PR
- Command moved from `commands/file/search.ts` to `commands/search.ts` (top-level, not under `boxel file`)
- Internals aligned with `file/read.ts` (`NO_ACTIVE_PROFILE_ERROR`, shared `ensureTrailingSlash`)
- Test helper gains `useRealPrerenderer: true` opt-in (only the search content test uses it; other CLI integration tests stay on the noop stub)
- New `finds the seeded SearchTarget card by title` test exercises real card indexing end-to-end
- CI: `boxel-cli-test` job now boots `mise run test-services:matrix` (host + base realm + prerenderer + worker-base + icons) so the content assertion runs in CI, mirroring the matrix-test job's pattern

## Test plan
- [x] `pnpm lint:types` and `pnpm lint:js` clean in `packages/boxel-cli`
- [x] `pnpm test:integration` — all 91 tests pass locally with the dev stack up (host on `:4200`, base realm on `:4201/base/`)
- [x] Without the dev stack up, only the search content test surfaces the friendly "host unreachable" error; the other 12 integration files still pass with the noop prerenderer
- [ ] CI green

Closes CS-10850, CS-10643

### DIFF
--- a/packages/boxel-cli/src/commands/search.ts
+++ b/packages/boxel-cli/src/commands/search.ts
@@ -1,9 +1,11 @@
 import type { Command } from 'commander';
 import {
   getProfileManager,
+  NO_ACTIVE_PROFILE_ERROR,
   type ProfileManager,
-} from '../../lib/profile-manager';
-import { FG_RED, DIM, RESET } from '../../lib/colors';
+} from '../lib/profile-manager';
+import { ensureTrailingSlash } from '@cardstack/runtime-common/paths';
+import { FG_RED, DIM, RESET } from '../lib/colors';
 
 export interface SearchResult {
   ok: boolean;
@@ -16,18 +18,16 @@ export interface SearchCommandOptions {
   profileManager?: ProfileManager;
 }
 
-function ensureTrailingSlash(url: string): string {
-  return url.endsWith('/') ? url : `${url}/`;
-}
-
 /**
- * Search a realm using the `_search` endpoint with a JSON query.
+ * Federated search across one or more realms via the `_federated-search`
+ * server endpoint.
  *
- * Sends a QUERY request with the provided query object serialized as JSON.
- * Uses the per-realm JWT via `ProfileManager.authedRealmFetch`.
+ * Sends a QUERY request with the provided query object and a `realms` array
+ * merged into the request body. Uses the server JWT via
+ * `ProfileManager.authedRealmServerFetch`.
  */
 export async function search(
-  realmUrl: string,
+  realmUrls: string | string[],
   query: Record<string, unknown>,
   options?: SearchCommandOptions,
 ): Promise<SearchResult> {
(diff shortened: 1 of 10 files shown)


packages/boxel-cli/src/commands/search.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  try {
+    let response = await pm.authedRealmFetch(searchUrl, {
+      method: 'QUERY',
+      headers: {
+        Accept: 'application/vnd.card+json',
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(query),
+    });
```
habdelra, line 58: this should specifically be a federated search. in the case where we want a realm search we should pull that realm and just use native file system tools to do the search like grep

habdelra (changes requested):
so as we discussed during planning--lets not have 2 searches. this is just confusing for the agent. lets just have a single search and have it be a federated search. for search within a realm the agent will need to pull and then just use grep or whatever it likes to use. that means we can either make this PR the federated search PR (but the command i just "search"), or we do that work in the federated search PR.

FadhlanR:
@habdelra updated to address the review:

- Single top-level `boxel search` command — no longer nested under `boxel file`.
- Always federated via `_federated-search`. `--realm` is repeatable for multi-realm queries.
- Per-realm content access pattern is `boxel realm pull` + grep, as you suggested.

Folded the federated work into this PR, so it now closes both CS-10850 and CS-10643.

Also added a CI step (`mise run test-services:matrix`) so the search content assertion runs against a real prerenderer (host + base realm + prerender service), matching the matrix-test job's orchestration. Locally with the dev stack up, all 91 integration tests pass.

packages/boxel-cli/tests/integration/search.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  cleanupProfile = testProfile.cleanup;
+  await setupTestProfile(profileManager);
+});
+
+afterAll(async () => {
+  cleanupProfile?.();
+  await stopTestRealmServer();
+});
+
+describe('federated search (integration)', () => {
```
habdelra, line 81: none of these search tests actually uses a search query. you should include a search test that actually find something using a query. you should have something in your realm that does not match the query and something in your realm that does match the query and assert that the expected result is returned

habdelra (approved):

PR #4479, opened 2026-04-22
CS-10851: Add `boxel file lint` command

FadhlanR:
## Summary
- Adds `boxel file lint <path> --realm <url> [--file <local>]` CLI command
- Extracts `BoxelCLIClient.lint()` inline implementation into a standalone command function at `src/commands/file/lint.ts`
- CLI displays errors (red) and warnings (yellow) with line:column format

## Why — pattern consistency
The `pull` command already follows this pattern: a standalone exported function that both the CLI and `BoxelCLIClient` delegate to. This PR makes `lint()` consistent with that approach. Every `BoxelCLIClient` method should be backed by a command function that is the single source of truth — the client is a thin delegation layer. This enables CS-10670 (tool definitions) where boxel-cli publishes tool definitions that reference these commands directly.

## Test plan
- [x] Integration tests: lints source via real `_lint` endpoint, no-profile error
- [x] Build passes
- [ ] CI green

Closes CS-10851

### DIFF
--- a/packages/boxel-cli/src/commands/file/index.ts
+++ b/packages/boxel-cli/src/commands/file/index.ts
@@ -1,5 +1,6 @@
 import type { Command } from 'commander';
 import { registerDeleteCommand } from './delete';
+import { registerLintCommand } from './lint';
 import { registerWriteCommand } from './write';
 
 export function registerFileCommand(program: Command): void {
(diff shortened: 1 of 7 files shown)


lukemelia (approved):
Do we need a flag about whether to auto-fix or not?

FadhlanR:
> Do we need a flag about whether to auto-fix or not?

Added

PR #4488, opened 2026-04-22
Enable parallel tool calls on OpenRouter factory agent (CS-10814)

habdelra:
## Summary

- Set `parallel_tool_calls: true` on the OpenRouter request body when tools are present, so a single assistant turn can emit multiple `tool_calls[]` — same spot in `packages/software-factory/src/factory-agent/openrouter.ts` the ticket called out.
- Without this flag, the OpenRouter → Anthropic route serialized one tool call per turn and re-sent the full \~30K-token prompt each round, producing the O(n²) context blow-up that made `--agent openrouter` runs take 40+ minutes where `--agent claude` finished in \~3.
- Extended the existing schema-boundary wire-body test to assert the flag lands on the outgoing request.

## Test plan

- [x] `pnpm lint` in `packages/software-factory` — clean (js, types, format)
- [x] `pnpm exec qunit tests/factory-agent-schema-boundary.test.ts` — 3/3 pass, including the new assertion
- [ ] **End-to-end validation still pending** (acceptance criterion #1): run a real `--agent openrouter` bootstrap on `anthropic/claude-opus-4` and confirm at least one observed turn carries `choice.message.tool_calls.length >= 2`. If the flag is silently ignored on that route, follow the ticket's fallback: document it in a comment next to the option.

Linear: CS-10814

### DIFF
--- a/packages/software-factory/src/factory-agent/openrouter.ts
+++ b/packages/software-factory/src/factory-agent/openrouter.ts
@@ -181,7 +181,12 @@ export class OpenRouterFactoryAgent implements LoopAgent {
         tool_calls: assistantToolCalls,
       });
 
-      // Execute each tool call
+      // Execute each tool call. With `parallel_tool_calls: true` a single
+      // assistant turn can carry multiple tool_calls[]; if the batch contains
+      // a terminal signal (signal_done / request_clarification) alongside
+      // other tools, we still execute the whole batch so the model's other
+      // side effects land, then return the first terminal signal observed.
+      let terminalResult: AgentRunResult | undefined;
       for (let toolCall of assistantToolCalls) {
         let toolName = toolCall.function.name;
         let tool = tools.find((t) => t.name === toolName);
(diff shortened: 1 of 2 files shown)


packages/software-factory/src/factory-agent/openrouter.ts
```diff
@@ -362,6 +362,11 @@ export class OpenRouterFactoryAgent implements LoopAgent {
 
     if (tools.length > 0) {
       body.tools = tools;
+      // Opt in to OpenAI-compatible parallel tool calls so a single assistant
+      // turn can emit multiple tool_calls[]. Without this, OpenRouter routes
+      // to Anthropic serialize 1 call/turn and re-send the full context each
+      // round, producing the O(n²) context blow-up observed in CS-10814.
+      body.parallel_tool_calls = true;
```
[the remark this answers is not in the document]
habdelra: Good catch — fixed in 51c15de. The per-call loop now remembers the first terminal signal (`signal_done` / `request_clarification`) instead of returning on it, and the `return terminalResult` happens once the whole batch has executed. Added a regression test in `factory-agent-schema-boundary.test.ts` that sends `[write_file, signal_done, write_file]` in a single assistant message and asserts both `write_file` calls run plus `status === 'done'`.

lukemelia (approved):

packages/software-factory/src/factory-agent/openrouter.ts
```diff
@@ -362,6 +372,11 @@ export class OpenRouterFactoryAgent implements LoopAgent {
...
```
jurgenwerk, line 379: I wonder if ai-bot could benefit from this too. I believe we have multi tool usage there too

jurgenwerk (approved):

packages/software-factory/src/factory-agent/openrouter.ts
habdelra: i got this idea from @lukemelia from his tool call work in the AI bot, so i''m guessing it does do this (?)

PR #4490, opened 2026-04-23
Fix boxel-cli bin warning on fresh clone/worktree

FadhlanR:
## Summary
- Move published `bin` target to `publishConfig` so pnpm doesn't warn about missing `dist/index.js` on fresh clones/worktrees
- Add `bin/boxel.js` wrapper that uses built dist when available, falls back to `ts-node` for development (no build required)
- Remove `tsx` dependency, use `ts-node` exclusively for consistency with monorepo conventions (migration to Node 24 native TS tracked in CS-10712)

## Test plan
- [ ] `rm -rf packages/boxel-cli/dist && pnpm install` — no warning
- [ ] `npm link` from `packages/boxel-cli` — `boxel --help` works without building
- [ ] `pnpm build && boxel --help` — uses built dist
- [ ] `pnpm publish:dry` — verify published package has correct bin

Fixes CS-10877

### DIFF
--- a/packages/boxel-cli/.eslintignore
+++ b/packages/boxel-cli/.eslintignore
@@ -1,4 +1,5 @@
 node_modules/
 dist/
+bin/
 *.tgz
 *.log
--- a/packages/boxel-cli/README.md
+++ b/packages/boxel-cli/README.md
@@ -49,31 +49,42 @@ pnpm build
 pnpm start
 ```
 
-### Local Development with `npm link`
+### Local Development
 
-To use your locally-built `boxel-cli` instead of the npm-installed version:
+Run directly from TypeScript source without building:
 
 ```bash
-# 1. Build the CLI
 cd packages/boxel-cli
-pnpm build
+pnpm start -- <command> [args]
 
-# 2. Create a global symlink from the package directory
-npm link
-
-# 3. Verify the global `boxel` command now points to your local build
-which boxel
-boxel --version
+# Examples:
+pnpm start -- --help
+pnpm start -- sync .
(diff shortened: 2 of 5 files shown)


habdelra (changes requested):
why are we using tsx? the original requirements for boxel-cli were to use ts-node so that we could be consistent with how scripts are executed in the rest of the monorepo.

FadhlanR:
> why are we using tsx? the original requirements for boxel-cli were to use ts-node so that we could be consistent with how scripts are executed in the rest of the monorepo.

Updated! Thanks for the feedback!

habdelra (approved):

FadhlanR:
**Re: migrating to Node 24 native TS (removing ts-node)**

We explored requiring Node 24 and using `--experimental-transform-types` to drop ts-node entirely. It mostly works, but hits a blocker:

- Node 24's type stripping treats `.ts` files as ESM, requiring **explicit `.ts` extensions on all imports**
- This includes imports from workspace dependencies like `@cardstack/runtime-common/matrix-constants` → needs `.ts` extension
- Adding `"type": "module"` to boxel-cli also requires renaming `.eslintrc.js` → `.cjs`, build output → `.cjs`, etc.
- The `@cardstack/runtime-common` package would need an `exports` map or extension updates, which ripples across the monorepo

**Deferring to [CS-10712](https://linear.app/cardstack/issue/CS-10712/migrate-off-ts-node-to-run-directly-in-node-24)** where we can coordinate the runtime-common changes alongside this migration.

PR #4520, opened 2026-04-24
CS-10860: surface revalidation-time template errors instead of timing out

habdelra:
## Summary
- Fixes the 90s render-timeout loop observed on whitepaper.boxel.ai/index.json: the render route's window-path `errorHandler` now swaps `{{outlet}}` to `render/error.gts` when an unhandled error fires, so Glimmer can commit a render even if the broken template keeps throwing on revalidation.
- Keeps the `data-prerender-status="unusable"` signal that the prerender server reads to evict the now-dead page — the fix deliberately bypasses `handleRenderError` because `processRenderError` would coerce the status to `'error'` and suppress eviction.
- Dedupes repeated errorHandler re-entries using the existing `lastRenderErrorSignature` slot so a template that keeps emitting the same rejection on every revalidation doesn't thrash the router.
- Tags errors whose stack carries Glimmer render frames (`ifHelper`, `rerender`, `revalidate`, `evaluateOuter`, `handleException`, …) with a `Error during template render (Glimmer frame X)` prefix on `error.message`, so ops can tell a template-helper bug apart from a loader/model bug without hand-parsing the stack.

## Root cause recap
In prod the whitepaper render reliably stalled with:
- one `[data-prerender]`, `data-prerender-status="loading"`, frozen 60s+
- console `Uncaught (in promise) TypeError: Cannot convert undefined or null to object` coming from `ifHelper → compile → rerender → revalidate`
- **zero `[data-prerender-error]` elements** — only reachable if no handler's DOM write stuck. The "Uncaught (in promise)" prefix says the error surfaced via `window.unhandledrejection`; the listener was racing against Glimmer rolling back every revalidation, so no handler committed a terminal state.

The fix replaces the broken outlet with `render/error.gts` before Glimmer's next revalidation, which lets the final binding commit `unusable` and the prerender server evict + surface the real TypeError within ~1s instead of 90s.

## Regression test — honest scope
`packages/realm-server/tests/prerendering-test.ts` — adds `revalidation-throws.gts` (`@tracked bork` flips after initial render; the second revalidation's `{{#if this.explode}}` calls `Object.getPrototypeOf(undefined)`). Asserts TypeError surfaces with message, non-empty stack, template-origin hint, `timedOut=false`, `evicted=true`.

**Transparency on coverage:** I tried two fixture variants to get a strict failing-before / passing-after test. Neither worked perfectly in this test environment:

1. **Committed fixture (sync throw):** passes with AND without the render.ts transition fix. The sync throw goes through `#handleUnhandledError` (RSVP/backburner), which was already doing the route transition — so this fixture exercises a path that was already working, plus the new template-origin hint. It's a regression guard on the surfacing machinery (TypeError message + stack + hint).

2. **Alternate Promise-from-getter fixture** (tested locally, not committed): the getter returns a fresh `Promise.reject(...)` on each read, which matches the prod "Uncaught (in promise)" signature more closely. **This fixture fails evict=true both with AND without the fix**, because the rejection fires continuously and the route's sync `error` action (or something else in the Glimmer/Ember async-value handling path) overwrites model.status to `'error'` before the prerender server captures. Getting this case to pass requires additional surgery beyond errorHandler — probably an early-return in the `error` action when `lastRenderErrorSignature` is already set, or a deferred status reassertion. Out of scope for this ticket.

The fix itself is justified by code analysis: pre-fix `errorHandler` made only DOM-direct and tracked-status writes with no route transition, which matches the "data-prerender-status=loading, zero error elements" prod signature (Glimmer's revalidation rolling back the direct writes without ever committing a safe outlet). The `render/error.gts` transition gives Glimmer a safe template to commit.

## Secondary goal (Hassan asked)
The offending template on prod whitepaper is not reachable from this dev box, so I did not localize it directly. Instead I delivered the enablement: in prod, the error doc on the next failure will carry `Error during template render (Glimmer frame "ifHelper" in stack): Cannot convert undefined or null to object` + the full stack, which is enough to grep the offender from the whitepaper realm's transpiled modules without another repro.

## Test plan
- [x] All 7 error-block prerendering tests pass with the fix: `throws`, `rejects`, `rsvp-rejects`, `console-error`, `console-no-error`, `revalidation-throws` (new), pre-model-throw.
- [x] `pnpm lint` green in `packages/host`.
- [x] `npx eslint tests/prerendering-test.ts` green in `packages/realm-server`.
- [ ] CI green on main.

## Follow-up
Consider a follow-up ticket: make the route's `error` action early-return when `lastRenderErrorSignature` is already set, so the async-getter-returns-rejected-Promise variant also evicts. See the transparency note above for the failing fixture.

### DIFF
--- a/packages/host/app/lib/window-error-handler.ts
+++ b/packages/host/app/lib/window-error-handler.ts
@@ -95,9 +95,49 @@ export function windowErrorHandler({
     }
   }
 
+  // CS-10860: tag errors that bubbled up through Glimmer revalidation so
+  // the prerender server / ops operator has a cheap signal that the
+  // offending code lives in a template helper (getter/if/each) rather
+  // than in loader or model wiring. The minified stack still surfaces
+  // recognizable Glimmer/Ember frames — look for any of them and prepend
+  // a hint to the message so it shows up in the error doc without having
+  // to parse the stack by hand.
+  let templateHint = describeTemplateOrigin(errorPayload.error.stack);
+  if (templateHint && typeof errorPayload.error.message === 'string') {
+    if (!errorPayload.error.message.startsWith(templateHint)) {
+      errorPayload.error.message = `${templateHint}: ${errorPayload.error.message}`;
+    }
+  }
+
   setError(JSON.stringify(errorPayload));
   setStatusToUnusable();
   event.preventDefault?.();
+}
+
+const TEMPLATE_STACK_MARKERS = [
+  'ifHelper',
+  'eachHelper',
+  'helperForImpl',
+  'evaluateOuter',
+  'handleException',
+  'rerender',
+  'renderRoots',
+  'revalidate',
+  'UpdatingVM',
+  'LowLevelVM',
+] as const;
+
+function describeTemplateOrigin(
+  stack: string | null | undefined,
+): string | undefined {
+  if (!stack) {
+    return undefined;
+  }
+  let hit = TEMPLATE_STACK_MARKERS.find((marker) => stack.includes(marker));
+  if (!hit) {
+    return undefined;
+  }
+  return `Error during template render (Glimmer frame "${hit}" in stack)`;
 }
 
 export function errorJsonApiToErrorEntry(
--- a/packages/host/app/routes/render.ts
+++ b/packages/host/app/routes/render.ts
@@ -166,8 +166,25 @@ export default class RenderRoute extends Route<Model> {
     ) {
       return;
     }
-    let context = this.#deriveErrorContext();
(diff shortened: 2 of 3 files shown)


habdelra:
Superseded by #4521 — the original hypothesis (window error handler not transitioning to render.error) turned out to be wrong: the prod throw doesn't fire any JS-level event at all (window.error / unhandledrejection / RSVP.on(error) / console.error all stay quiet). The new PR adds a DOM desync detector that catches the actual signal — model.status='ready' but DOM=loading — plus defensive timer-stub error threading for a different swallowed-error class.

PR #4526, opened 2026-04-27
Revert "CS-10948: discard in-flight prerender result if invalidate ran concurrently"

habdelra:
Reverts cardstack/boxel#4515

I think this PR is causing the software factory to timeout during search tests in main branch. Let's confirm the software factory tests pass in this PR before merging.


jurgenwerk (approved):

habdelra:
This PR does not seem to be the one that introduced the issue, it looks like it is still here. I'm gonna revert the 2 PR's that I merged today in a separate PR

PR #4527, opened 2026-04-27
Revert PRs #4515 and #4521 to investigate SF realm-search timeouts

habdelra:
## Summary

Reverts both PRs merged this morning to bisect the source of the software-factory `realm-search` timeouts now appearing on `main`:

- Revert #4521 — "Surface render errors invisible to JS event handlers (CS-10860)"
- Revert #4515 — "CS-10948: discard in-flight prerender result if invalidate ran concurrently"

## Why

Two software-factory Playwright tests started failing on `main` after this morning's merges:

1. `tests/factory-tool-executor.spec.ts:574 › realm-search with seeded fixture data › search by type returns matching cards and excludes non-matching types`
2. `tests/factory-tool-executor.spec.ts:670 › realm-search on a private realm › search with owner token succeeds, search without token fails`

Both fail with `ToolTimeoutError: Tool "realm-search" timed out after 60000ms`. CI total runtime also rose from ~20m to 50m+/timing out, suggesting the slowdown is broader than just these two tests.

Local instrumentation showed `realm-search` returns 2 cards in ~30ms but then hangs in `loadLinks → lookupDefinition` for the user-defined type module (`software-factory/darkfactory.gts`). The first prerender call took 52s in one local run; in another it never returned within the test budget.

The two PRs being reverted both touch the realm-server's prerender / definition-lookup path:

- #4515 changed `CachingDefinitionLookup` to discard in-flight prerender results when an invalidation runs concurrently (added per-realm/global generation counters + dropped in-flight map entries on invalidate).
- #4521 added a render-desync detector and a couple of prerender / page-pool changes to surface render errors invisible to JS event handlers.

This PR reverts both so we can confirm whether one (or both) is the cause and let CI green up while the underlying issue is investigated.

## What's NOT in this PR

- Root-cause analysis for why the prerender of a single 756-line `.gts` module takes 50+ s. That's a separate investigation — could be queue head-of-line blocking, a regression in prerender perf, or something else.
- Any forward-fix to either CS-10948 or CS-10860. Those tickets remain open and should be re-landed once the perf regression is understood.

## Conflicts resolved

Both reverts had merge conflicts because of subsequent main commits:

- `packages/realm-server/tests/prerendering-test.ts` — removed the `desync-repro.gts/json` fixture entries added by #4521.
- `packages/realm-server/tests/definition-lookup-test.ts` — removed the new persist-after-invalidate test cases added by #4515; kept the simpler v1/v2 displayName asserts that pre-existed.

Conflicts only because #4428 ("Realm[Resource]Identifier types") landed between the two reverted PRs and the revert.

## Test plan

- [ ] `Software Factory Tests` CI job passes on this PR (the two failing tests above pass, and total runtime is back down to ~20 minutes).
- [ ] No new lint errors introduced (the pre-existing `runtime-common/helpers/ai.ts` `lint:types` errors on main are not from these reverts).
- [ ] Any other CI jobs that were green on main pre-#4515/#4521 are still green.


habdelra:
:heavy_check_mark:  confirming the software factory tests pass now.

PR #4535, opened 2026-04-27
Extract AMG state via direct HTTP API

lukemelia:
> **Stacked on [PR #4533](https://github.com/cardstack/boxel/pull/4533)**, which is stacked on [PR #4529](https://github.com/cardstack/boxel/pull/4529). Review/merge those first.
> **Draft until I've run the script against staging AMG and the round-trip verification passes.**

## Summary

Phase 3 starter — pull AMG (AWS Managed Grafana) state into the package's source-of-truth trees.

**Approach pivot from the original ticket** ([CS-10922](https://linear.app/cardstack/issue/CS-10922/pull-amg-dashboardsdatasourcesalerts-via-direct-http-api)): instead of installing the archived Grizzly tool, this PR uses direct HTTP API extraction (`curl` + `jq`) reading the existing `GRAFANA_API_KEY` from SSM. Zero external tool dependencies, captures live state including UI edits since last `terraform apply`, and outputs files in the exact target shapes — no double-reformat.

## What's in this PR (so far)

- **`scripts/migrations/extract-amg.sh`** — bash one-shot that takes `staging` or `production` and writes:
  - `grafanactl/resources/folders/*.json` — App Platform `Folder` kind manifests
  - `grafanactl/resources/dashboards/boxel-status/*.json` — App Platform `Dashboard` kind manifests, dashboard JSON in `spec`
  - `provisioning/datasources/*.json` — Grafana provisioning `apiVersion: 1` format
  - `provisioning/alerting/*.json` — Grafana provisioning format, grouped by `folderUID + ruleGroup`

The actual extracted output (JSON files) is not in this PR yet — see "Next steps" below.

## Notable choices

- **No secrets committed.** Data source `secureJsonData` is intentionally absent from the script's output (Grafana's API never returns secret fields). Staging/production delivery wires them via `${ENV_VAR}` in a follow-up infra ticket.
- **Dashboard manifest shape is a best guess** — App Platform wrapper (`apiVersion: dashboard.grafana.app/v1beta1`, dashboard JSON in `spec`). If `grafanactl resources push` rejects the format during round-trip, the only thing that changes is the script's `jq` reshape (~3 lines).
- **Vendored Synapse note.** Sibling `synapse.NOTE.md` flags `synapse.json` as upstream-vendored (since JSON has no comments).
- **Once-only tool.** After Phase 3 lands and AMG is deprecated, the script becomes archival.

## Next steps (require your AWS creds)

```bash
# 1. Pull this branch
[...]

PR #4533, opened 2026-04-27
Loki container + data source via Grafana provisioning

lukemelia:
> **Stacked on [PR #4529](https://github.com/cardstack/boxel/pull/4529)** — review/merge that one first. ✅  This PR's diff against base shows only the CS-10918 changes.

## Summary

Adds a local Loki container and a `Loki` Grafana data source provisioned via Grafana's built-in file provisioning (not via grafanactl).

**Discovery during implementation:** `grafanactl` does **not** manage data sources. `grafanactl resources list` only enumerates App Platform kinds (dashboards, folders, playlists, snapshots, preferences, etc.) — there's no `datasources` kind, and `pull` returns nothing for them. Data sources still live behind Grafana's legacy `/api/datasources` HTTP API and are best handled via Grafana's built-in file-based provisioning, which natively supports `${ENV_VAR}` substitution at startup.

## Architecture: two trees, two delivery mechanisms

| Tree                   | Tool                              | Resource kinds                  | Apply mechanism                                  |
| ---------------------- | --------------------------------- | ------------------------------- | ------------------------------------------------ |
| `grafanactl/resources/` | `grafanactl resources push`        | Dashboards, folders             | API push (CI on merge)                           |
| `provisioning/`        | Grafana file provisioning          | Data sources, alert rules       | Files mounted at `/etc/grafana/provisioning/`, read at startup |

For staging/production, the `provisioning/` tree gets delivered to the ECS Grafana container (via image bake, S3-init, or EFS mount — TBD in a separate infra ticket). Container restart picks up changes.

## Changes

- **`provisioning/datasources/loki.yaml`** — data source manifest:
  - `uid: loki` so dashboards reference it stably across envs
  - `url: ${LOKI_URL}` substituted by Grafana itself at startup
  - `editable: false` so UI edits are rejected (config-as-code discipline)
- **`docker-compose.yml`**:
  - New `loki` service on port 3100 (Loki 3.4.4)
  - Grafana provisioning mounts switched from `./grafanactl/resources/datasources` to `./provisioning/datasources` and `./provisioning/alerting`
  - `LOKI_URL=http://loki:3100` on the grafana service env
  - Alloy included as a **commented-out** service (scraping config is a follow-up; Loki + the data source are useful even with no log shipper)
- **`grafanactl/resources/datasources/.gitkeep`** removed — grafanactl doesn't manage data sources, so the dir was misleading. Renamed in git as `provisioning/alerting/.gitkeep`.
[...]

PR #4529, opened 2026-04-27
Setup observability package 

lukemelia:
## Summary

Phase 2 of the [AMG → self-host Grafana migration](https://linear.app/cardstack/project/replace-dashboard-with-our-own-grafana-1b7e3de21dbf) — three tickets bundled because they're tightly coupled and small individually:

- **CS-10914** — package skeleton at `packages/observability/`
- **CS-10912** — local `docker-compose.yml` for Grafana, `apply.sh` wrapper, `local` grafanactl context
- **CS-10913** — `staging` + `production` grafanactl contexts, `grafanactl-env.sh` SSM helper, `pull.sh`

## What's in the package

```
packages/observability/
  README.md                                # project context + workflow sketch
  package.json                             # @cardstack/observability, private: true
  docker-compose.yml                       # local Grafana 12.4.3 on port 3001
  grafanactl/
    config.yaml                            # local + staging + production contexts (no tokens)
    resources/
      folders/.gitkeep
      datasources/.gitkeep
      dashboards/.gitkeep
      alerts/.gitkeep
  scripts/
    apply.sh                               # ./scripts/apply.sh --env <local|staging|production>
    pull.sh                                # ./scripts/pull.sh  --env <name> --path <dir>
    grafanactl-env.sh                      # sourceable helper that exports GRAFANA_TOKEN from SSM
  templates/env-vars.env.example
```

Plus `.gitignore` entries for `templates/env-vars.env` and runtime `*.local.yaml` / `*.local.json`.

## docker-compose details (CS-10912)

- Image: `grafana/grafana:12.4.3` — matches the production pin from [cardstack/infra#663](https://github.com/cardstack/infra/pull/663). grafanactl requires Grafana 12+.
- Port `3001:3000` — avoids traefik (80, 4230), realm-server stack (4201/4210/4212/4213/4221/4222/4231/4232), and the common host port 3000.
- Volumes: provisioning mounted RO from `./grafanactl/resources/`; named `grafana_data` for the SQLite DB so user-created panels survive restarts.
- `extra_hosts: host.docker.internal:host-gateway` so the container can reach the host's `boxel-pg` for the local Postgres data source (lands in CS-10918+/Phase 3).
- Auth: admin/admin, anonymous disabled.
- Commented-out Loki + Alloy block reserved for Phase 2.5 (CS-10918).

## grafanactl contexts (CS-10912 local + CS-10913 staging/production)

[...]

PR #4538, opened 2026-04-27
CI — apply observability config to grafana staging on merge

lukemelia:
> **Stacked on PR #4535** ([cs-10922-amg-export](https://github.com/cardstack/boxel/pull/4535)) → which is stacked on PR #4533 → which is stacked on PR #4529. Review/merge those first. Draft until CS-10922 merges (this workflow only does meaningful work once dashboard JSON exists in the package).
>
> **Companion infra PR: [cardstack/infra#665](https://github.com/cardstack/infra/pull/665)** — creates the dedicated IAM role this workflow assumes.

## Summary

Adds the GitHub Actions workflow that runs \`grafanactl resources push --context staging\` whenever \`main\` gets a push touching \`packages/observability/**\`. Mirrors the repo's existing terraform-apply-on-merge pattern.

This is the CI half of the dashboard-as-code feedback loop: PR review (CS-10933) → merge to main → automatic apply to staging Grafana.

## File added

\`.github/workflows/observability-apply-staging.yml\` — single \`apply\` job:

1. Checkout
2. Install grafanactl from GitHub releases (\`curl\` + \`tar\`, version pinned via env var, **SHA-256 verified against published checksums.txt** before \`sudo install\` — no unverified binary lands in PATH)
3. Configure AWS credentials via OIDC (\`aws-actions/configure-aws-credentials\`) using the dedicated \`boxel-observability-apply\` role (created in [cardstack/infra#665](https://github.com/cardstack/infra/pull/665))
4. \`aws ssm get-parameter --name /staging/grafana/grafanactl_token\` → exported as \`GRAFANA_TOKEN\` with \`::add-mask::\` so it can't leak in logs
5. \`./scripts/apply.sh --env staging\` (the wrapper from CS-10913, which renders the per-call config tempfile and invokes \`grafanactl resources push\`)
6. Step Summary writes target + grafanactl version + status

## IAM

The workflow assumes \`arn:aws:iam::680542703984:role/boxel-observability-apply\`. That role is created in **[cardstack/infra#665](https://github.com/cardstack/infra/pull/665)** with exactly one permission: \`ssm:GetParameter\` on \`/staging/grafana/grafanactl_token\`. Nothing else.

The dedicated role was a deliberate choice over extending the broader \`boxel-host\` role with SSM permissions: blast radius for an observability-apply-staging.yml compromise should be one parameter, not boxel-host's S3/CloudFront permissions.

## Trigger shape

\`\`\`yaml
on:
  push:
    branches: [main]
    paths:
      - "packages/observability/**"
      - ".github/workflows/observability-apply-staging.yml"
  workflow_dispatch:
\`\`\`

\`workflow_dispatch\` for manual reruns; the workflow file itself is on the paths list so workflow edits self-trigger.

\`\`\`yaml
concurrency:
  group: observability-apply-staging
  cancel-in-progress: false
\`\`\`

Two pushes-in-flight serialize cleanly rather than canceling the in-flight apply mid-push.

## Server URL — single source of truth

Server URL is **deliberately not** set as a workflow env var. The committed \`packages/observability/grafanactl/config.yaml\` is the single source of truth for per-context server URLs, rendered to a tempfile by \`scripts/render-config.sh\` at apply time. Setting \`GRAFANA_SERVER\` in the workflow would silently override the config — if the staging URL ever changes (e.g., Phase 6 cutover renames to \`dashboard-staging.stack.cards\`), the workflow would silently keep hitting the old hostname.

## Apply order

1. Merge + apply [cardstack/infra#665](https://github.com/cardstack/infra/pull/665) → \`boxel-observability-apply\` role exists in the staging account.
2. Land the boxel-side stack: #4529 → #4533 → #4535 → this PR.
3. First push to main touching \`packages/observability/**\` triggers the workflow. Or smoke-test via \`workflow_dispatch\` from the Actions UI.
4. If step 1 hasn't happened: the \`Configure AWS credentials\` step fails with an OIDC error — visible, contained, no silent over-permissioning.

## Test plan

Once CS-10922 merges and brings real dashboards into \`packages/observability/grafanactl/resources/\`:

- [ ] Manually \`workflow_dispatch\` once to confirm OIDC + checksum verify + SSM + apply succeed end-to-end against staging.
- [ ] Touch any file under \`packages/observability/**\` in a follow-up PR; confirm the workflow runs on merge to main.
- [ ] Touch any file outside \`packages/observability/**\`; confirm the workflow does NOT run.
- [ ] Mid-flight: kick off two consecutive merges; confirm the second blocks until the first finishes (concurrency).
- [ ] Pre-flight: confirm \`boxel-observability-apply\` role exists — otherwise the credentials step 401s.
- [ ] Tamper test (optional): bump \`GRAFANACTL_VERSION\` to a value with a mismatched/missing checksum; confirm the install step fails before \`grafanactl --version\` runs.

## Linear

- Closes [CS-10932](https://linear.app/cardstack/issue/CS-10932/github-actions-workflow-apply-to-self-host-staging-on-merge)
- Will benefit from [CS-10959](https://linear.app/cardstack/issue/CS-10959/add-lint-scripts-for-packagesobservability) (lint step before push) once that ticket lands.

### DIFF
--- a/.github/actions/install-grafanactl/action.yml
+++ b/.github/actions/install-grafanactl/action.yml
@@ -1,0 +1,34 @@
+name: Install grafanactl
+description: |
+  Download and install the pinned grafanactl release on the Ubuntu runner,
+  verifying the SHA-256 checksum from the release's grafanactl_checksums.txt
+  before placing the binary on PATH. Used by the observability-apply-staging,
+  observability-apply-production, and observability-diff workflows.
+
+inputs:
+  version:
+    description: grafanactl release version (without the leading 'v', e.g. "0.1.10")
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      env:
+        GRAFANACTL_VERSION: ${{ inputs.version }}
+      run: |
+        set -euo pipefail
+        TARBALL="grafanactl_Linux_x86_64.tar.gz"
+        CHECKSUMS="grafanactl_checksums.txt"
+        RELEASE_URL="https://github.com/grafana/grafanactl/releases/download/v${GRAFANACTL_VERSION}"
+        TMPDIR="$(mktemp -d)"
+        trap 'rm -rf "$TMPDIR"' EXIT
+
+        curl -fsSL "${RELEASE_URL}/${TARBALL}"   -o "${TMPDIR}/${TARBALL}"
+        curl -fsSL "${RELEASE_URL}/${CHECKSUMS}" -o "${TMPDIR}/${CHECKSUMS}"
+
+        ( cd "$TMPDIR" && grep " ${TARBALL}\$" "${CHECKSUMS}" | sha256sum -c - )
+
+        tar -xzf "${TMPDIR}/${TARBALL}" -C "$TMPDIR" grafanactl
+        sudo install -m 0755 "${TMPDIR}/grafanactl" /usr/local/bin/grafanactl
+        grafanactl --version
--- a/.github/workflows/observability-apply-staging.yml
+++ b/.github/workflows/observability-apply-staging.yml
@@ -1,0 +1,104 @@
+name: Observability — apply to staging
+
+# Push to main with changes under packages/observability/** triggers a
+# `grafanactl resources push` against the staging self-host Grafana.
+# Mirrors the existing terraform-apply-on-merge pattern.
+#
+# Production has its own workflow (observability-apply-production.yml from
+# CS-10936) gated on manual dispatch.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "packages/observability/**"
+      - ".github/workflows/observability-apply-staging.yml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  id-token: write
+
+concurrency:
+  # Never run two applies in parallel against the same environment, but
+  # don't cancel an in-flight apply when a new commit lands — let it finish
+  # cleanly so we don't leave Grafana in a half-pushed state.
+  group: observability-apply-staging
+  cancel-in-progress: false
+
+jobs:
+  apply:
+    name: Apply to staging
+    runs-on: ubuntu-latest
+
+    env:
+      AWS_REGION: us-east-1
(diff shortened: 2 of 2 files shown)


lukemelia:
Updated to point at a dedicated, narrowly-scoped role: \`arn:aws:iam::680542703984:role/boxel-observability-apply\`.

The role itself is created in **[cardstack/infra#665](https://github.com/cardstack/infra/pull/665)** — option (b) from the original PR description. Trust policy: GitHub OIDC, repo \`cardstack/boxel\`. Inline policy: exactly \`ssm:GetParameter\` on \`/staging/grafana/grafanactl_token\`. Nothing else.

Apply order:
1. Merge + apply cardstack/infra#665 (creates the role).
2. Merge this PR after #4535 (so dashboards exist for the workflow to actually push).
3. First push to main touching \`packages/observability/**\` triggers the workflow against the new role. If we want a smoke test before relying on natural triggers, \`workflow_dispatch\` from the Actions UI works too.

If step 1 hasn't happened by the time this PR or anything stacked on it merges, the workflow will fail at the \`Configure AWS credentials\` step with an OIDC error — visible, contained, no silent over-permissioning.

lukemelia:
Addressed all three Copilot suggestions in commit 417eaa92:

1. **PR description stale.** Body now reflects the dedicated \`boxel-observability-apply\` role (matching the workflow); removed the placeholder \`boxel-host\` reference.

2. **\`GRAFANA_SERVER\` workflow override risk.** Dropped the env-var declaration and the \`echo "GRAFANA_SERVER=…" >> \$GITHUB_ENV\` line. Server URL now flows exclusively from the committed \`grafanactl/config.yaml\` via \`render-config.sh\`. Step Summary updated to reference the config file as source of truth rather than echoing a hardcoded value.

3. **Tarball integrity check.** Install step now downloads \`checksums.txt\` alongside the tarball, runs \`sha256sum -c\` from a tempdir, and \`sudo install\`s only after verification. Tamper-test path noted in the PR's test plan.

PR #4540, opened 2026-04-27
Don't clobber parent error doc in render subroutes

habdelra:
## Summary
Fixes the flaky `indexing-test.ts > error recovery and deletion > can incrementally index instance that depends on deleted card source` test (seen on main runs [25003516731](https://github.com/cardstack/boxel/actions/runs/25003516731) and [24851372853](https://github.com/cardstack/boxel/actions/runs/24851372853), and on PR #4530's CI).

## What the test sees when it fails
The expected error doc for `post-1.json` after `post.gts` is deleted:
- `id: <realm>/post` (the deleted module URL)
- `isCardError: true`
- `message: missing file <realm>/post`
- `status: 404`
- `title: Link Not Found`

The actual doc on the failing runs:
- `id: <realm>/post-1.json` (the instance URL)
- `isCardError: undefined`
- `message: Missing render instance`
- `status: 500`
- `title: Missing render instance`

## Root cause
When a card source is deleted and a dependent instance is reindexed, the parent `render` route's `model()` correctly throws a `CardError(404, "missing file <module>")` during hydration, and its `error` action serializes that into a proper `"Link Not Found"` doc.

But during the prerender's multi-pass cardRender flow, the child routes (`render.html`, `render.icon`) sometimes race ahead and run their own `model()` while the parent is still settling. Their fallback was `throw new Error('Missing render instance')`, which the parent's error action then serialized as a *fresh* 500 doc — *replacing* the proper 404 in `renderErrorState` and the DOM. That's exactly what the test catches.

## Fix
Apply the same pattern `render/meta.ts` already uses (from commit d9f6190284): `await renderModel?.readyPromise` so the parent's disposition settles first, then on missing instance just `transition.abort()` and return undefined instead of throwing. This lets the parent's already-correct error doc stand.

That commit updated `meta` but left `html` and `icon` throwing — this finishes what it started.

## Test plan
- [ ] CI passes the previously-flaky `indexing-test.ts > error recovery and deletion > can incrementally index instance that depends on deleted card source` test
- [ ] Existing host acceptance test `prerender-html-test.gts > can handle not found prerender url` (which exercises the same error path via single-route navigation) still passes — parent's error wins; render.html's `model()` never runs in that happy/sad path
- [ ] No other test referenced `Missing render instance`; lint passes (`pnpm lint:js`, `pnpm lint:hbs`)

### DIFF
--- a/packages/host/app/routes/render/html.ts
+++ b/packages/host/app/routes/render/html.ts
@@ -1,5 +1,6 @@
 import Route from '@ember/routing/route';
 import type RouterService from '@ember/routing/router-service';
+import type Transition from '@ember/routing/transition';
 import { service } from '@ember/service';
 
 import { isValidFormat } from '@cardstack/runtime-common';
@@ -14,31 +15,41 @@ import { getClass, getTypes } from './meta';
 
 import type { Model as ParentModel } from '../render';
 
-export interface Model {
-  instance: CardDef;
-  format: Format;
-  Component: BoxComponent;
-}
+export type Model =
+  | {
+      instance: CardDef;
+      format: Format;
+      Component: BoxComponent;
+    }
+  | undefined;
 
 export default class RenderHtmlRoute extends Route<Model> {
   @service declare router: RouterService;
 
-  async model({
-    format,
-    ancestor_level,
-  }: {
-    format: string;
-    ancestor_level: string;
-  }) {
+  async model(
+    {
+      format,
+      ancestor_level,
+    }: {
+      format: string;
+      ancestor_level: string;
+    },
+    transition: Transition,
+  ) {
     let parentModel = this.modelFor('render') as ParentModel | undefined;
     // the global use below is to support in-browser rendering, where we actually don't have the
     // ability to lookup the parent route using RouterService.recognizeAndLoad()
     let renderModel =
       parentModel ??
       ((globalThis as any).__renderModel as ParentModel | undefined);
+    await renderModel?.readyPromise;
     let instance: CardDef | undefined = renderModel?.instance;
     if (!instance) {
-      throw new Error('Missing render instance');
+      // the lack of an instance is dealt with in the parent route — throwing
+      // here would clobber the parent's error doc (e.g. "Link Not Found" 404)
+      // with a generic 500 "Missing render instance"
+      transition.abort();
+      return;
     }
 
     if (!isValidFormat(format)) {
--- a/packages/host/app/routes/render/html.ts
+++ b/packages/host/app/routes/render/html.ts
@@ -42,12 +42,19 @@ export default class RenderHtmlRoute extends Route<Model> {
     let renderModel =
       parentModel ??
       ((globalThis as any).__renderModel as ParentModel | undefined);
-    await renderModel?.readyPromise;
     let instance: CardDef | undefined = renderModel?.instance;
     if (!instance) {
       // the lack of an instance is dealt with in the parent route — throwing
       // here would clobber the parent's error doc (e.g. "Link Not Found" 404)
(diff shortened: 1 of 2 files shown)


packages/host/app/routes/render/html.ts
```diff
@@ -14,31 +15,41 @@ import { getClass, getTypes } from './meta';
...
+    },
+    transition: Transition,
+  ) {
     let parentModel = this.modelFor('render') as ParentModel | undefined;
     // the global use below is to support in-browser rendering, where we actually don't have the
     // ability to lookup the parent route using RouterService.recognizeAndLoad()
     let renderModel =
       parentModel ??
       ((globalThis as any).__renderModel as ParentModel | undefined);
+    await renderModel?.readyPromise;
```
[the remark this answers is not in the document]
habdelra: Good catch — you're right that this would invert ordering with `card-prerender.gts`'s flow (`recognizeAndLoad` → `renderCardComponent` → `waitForLinkedData` → `#ensureRenderReady`). The RAF-based `#startReadyWatchdog` settles `readyPromise` after a couple of frames regardless of whether the manual render pass has fired the lazy link fetches, so awaiting `readyPromise` in the model hook would let the watchdog complete settlement before the render — `capturedDeps` would miss runtime deps and incremental indexing could miss invalidations.

I dropped the `await renderModel?.readyPromise` from both `render.html` and `render.icon` in 062e8aea70. The actual fix the test needed was just converting the throw to `transition.abort(); return;` so the parent's already-written error doc isn't clobbered — the await was unnecessary for that and actively harmful here.

Left a comment in the code explaining the asymmetry vs. `render.meta` (which legitimately awaits `readyPromise` because it runs **after** the html/icon passes have already triggered the lazy fetches).

packages/host/app/routes/render/html.ts
```diff
@@ -14,22 +15,27 @@ import { getClass, getTypes } from './meta';
...
-  format: Format;
-  Component: BoxComponent;
-}
+export type Model =
+  | {
+      instance: CardDef;
+      format: Format;
+      Component: BoxComponent;
+    }
+  | undefined;
```
lukemelia: It seems unusual to explicitly type Model as possibly being undefined.

packages/host/app/routes/render/icon.ts
```diff
@@ -9,12 +10,10 @@ import type {
 
 import type { Model as ParentModel } from '../render';
 
-export interface Model {
-  Component: CardOrFieldTypeIcon;
-}
+export type Model = { Component: CardOrFieldTypeIcon } | undefined;
```
lukemelia: Same thing here

lukemelia (approved):
Left some comments -- up to you what to do with them.

packages/host/app/routes/render/html.ts
habdelra: Agreed, that was an artifact of doing the abort inside `model()` — Ember's `Route<Model>` makes the model hook's return type covariant with the type parameter, so widening the override to `Promise<Model | undefined>` wasn't allowed and the only way to satisfy TS was to widen `Model` itself. Restructured in df0996080a to move the parent-instance check + `transition.abort()` into `beforeModel`. By the time `model()` runs, the abort has already happened on the bad-parent path, so `Model` is back to a clean interface and `model()` returns `Promise<Model>`.

packages/host/app/routes/render/icon.ts
habdelra: Same restructure applied here in df0996080a — abort moved to `beforeModel`, `Model` is back to a clean `interface { Component: CardOrFieldTypeIcon }`, `model()` returns `Promise<Model>`.

PR #4542, opened 2026-04-27
CI — apply observability config to production (manual)

lukemelia:
> **Stacked on PR #4538** ([cs-10932-staging-apply-workflow](https://github.com/cardstack/boxel/pull/4538)) → which is stacked on the rest of the PR chain. Review/merge those first.
>
> **Companion infra PR: [cardstack/infra#666](https://github.com/cardstack/infra/pull/666)** — creates the dedicated production IAM role this workflow assumes.
>
> **Draft until staging (PR #4538) is verified working post-merge.** The whole point of doing prod second is to soak staging first.

## Summary

Adds the production observability-apply workflow + wires it into the existing \`Manual Deploy [boxel]\` orchestration. Two paths to a production observability apply:

- **Standalone**: \`workflow_dispatch\` of \`observability-apply-production.yml\` — for dashboard-only changes when you don't want to run a full boxel deploy.
- **As part of a production boxel deploy**: \`manual-deploy.yml\` calls this workflow at the end (after realm-server stabilizes), so dashboards reflect the just-deployed code in the same orchestrated run.

Either way, no extra approval — the act of dispatching is approval enough (operator is already trusted to dispatch a prod deploy).

## Files added / changed

**New: \`.github/workflows/observability-apply-production.yml\`** — single \`apply\` job with these guardrails (deliberately stronger than the staging workflow):

| Guardrail                        | Staging (#4538)                              | Production (this PR)                                                  |
| -------------------------------- | -------------------------------------------- | --------------------------------------------------------------------- |
| Trigger                          | push to main on observability paths + manual | manual-dispatch + reusable via \`workflow_call\` from manual-deploy.yml |
| Branch enforcement               | n/a (auto-trigger)                           | refuses dispatch from anything other than \`refs/heads/main\`         |
| Pre-flight hostname check        | none                                         | asserts \`config.yaml\`'s production server URL matches expected      |
| GitHub environment               | none                                         | \`observability-production\` (badge + URL link, no required reviewers) |
| Install integrity check          | SHA-256 verified                             | SHA-256 verified (same)                                               |
| Concurrency                      | serialize, don't cancel                      | serialize, don't cancel (same)                                        |

**Modified: \`.github/workflows/manual-deploy.yml\`** — adds an \`apply-observability\` job that:
- Runs only when \`inputs.environment == 'production'\` (staging applies are already covered by the auto-trigger in PR #4538)
- Depends on \`post-deploy-realm-server\` — observability runs at the end so dashboards reflect just-deployed code
[...]

PR #4544, opened 2026-04-27
CS-10933: PR diff comment workflow + scripts/diff.sh

lukemelia:
> **Stacked on PR #4542** ([cs-10936-production-apply-workflow](https://github.com/cardstack/boxel/pull/4542)) → which is stacked on the rest of the chain. Review/merge those first.
>
> **Draft until staging apply (#4538) is verified working post-merge** — the diff workflow pulls from staging Grafana, so it needs the staging side ready first.

## Summary

When a PR touches \`packages/observability/**\`, post a sticky comment showing the diff between the committed \`grafanactl/resources/\` state and the live state of staging Grafana. Mirrors the existing terraform-plan-comment pattern that the repo uses for infra changes.

This was [Path B in CS-10933's design](https://linear.app/cardstack/issue/CS-10933) — grafanactl has no native \`resources diff\` subcommand (verified via \`grafanactl resources --help\`), so we implement the equivalent: pull live state to a tempdir, \`git diff --no-index\` against the committed tree.

## Files added

- **\`packages/observability/scripts/diff.sh\`** — pulls live grafanactl resources to a tempdir, diffs against committed \`./grafanactl/resources/\`. Empty output (and exit 0) = committed equals live.
- **\`.github/workflows/observability-diff.yml\`** — PR trigger, paths filter on \`packages/observability/**\`, reuses the staging \`boxel-observability-apply\` role, posts a sticky comment with the diff (or "no changes" if empty), truncates at 60KB.

## Scope of the diff

The diff is limited to **grafanactl-managed kinds**: dashboards and folders. The \`provisioning/\` tree (data sources, alert rules) is delivered to Grafana via file mount at startup, not API push — so there's no "live state" to diff against from the outside. PR review of those YAMLs is the only check for that tree.

This is documented inline in \`diff.sh\` so future maintainers don't try to extend the diff to provisioning/ without realizing the architecture mismatch.

## Sticky comment

- Marker: \`<!-- observability-diff -->\` at the top of the comment body.
- On each push: list comments, find one starting with marker, update if exists else create.
- Concurrency group is per-PR with \`cancel-in-progress: true\` — older runs cancel when a new push lands, so the comment always reflects the latest commit.
- Diff is also printed to the run log inside a collapsible \`::group::\` so reviewers can read it without expanding the comment.

## Truncation

GitHub PR comments cap at ~64KB. We truncate at 60KB and append a link to the run details for the full diff. Most observability PRs will be well under this — the synapse dashboard is the only resource large enough to risk overflow, and it changes rarely (vendored from upstream Synapse).

## IAM

Reuses \`arn:aws:iam::680542703984:role/boxel-observability-apply\` (the staging apply role from cardstack/infra#665). The role has \`ssm:GetParameter\` on \`/staging/grafana/grafanactl_token\` only — same scope a diff needs.

Note: the staging Grafana service-account token has Admin role, so technically the diff workflow could push instead of pull (it's just a question of which subcommand we invoke). The boundary between "diff" and "apply" is enforced by the workflow YAML, not by the token's permissions. A separate read-only Grafana service account would be tighter; deferred as a possible follow-up if it becomes valuable.

## Test plan

After staging-apply (#4538) is verified working post-merge:

- [ ] PR touching \`packages/observability/grafanactl/resources/dashboards/\`: comment appears showing the dashboard diff (or "No changes" if the committed state matches what was last applied to staging).
- [ ] Subsequent push to the same PR: existing comment updates in place (no new comment).
- [ ] PR touching only \`packages/observability/scripts/\`: workflow runs but the diff is empty (no resource changes) — comment says "No changes detected".
- [ ] PR touching only \`provisioning/\`: workflow runs, diff is empty (provisioning tree isn't grafanactl-managed). Comment notes the limitation… actually the comment just says "no dashboard/folder changes" which is true. The reviewer is responsible for diffing provisioning/ via the regular GitHub file diff.
- [ ] Forks: TBD — \`pull-requests: write\` permission has limited scope on fork PRs. cardstack/boxel rarely takes fork contributions; if it becomes an issue we add a fork-detection branch.
- [ ] Long diff (Synapse-dashboard scale): truncates at 60KB with a link to the run.

## Linear

- Closes [CS-10933](https://linear.app/cardstack/issue/CS-10933/pr-workflow-post-dashboard-diff-as-a-pr-comment)
- Will benefit from [CS-10959](https://linear.app/cardstack/issue/CS-10959/add-lint-scripts-for-packagesobservability) (lint step before diff) once that ticket lands.

### DIFF
--- a/.github/workflows/observability-diff.yml
+++ b/.github/workflows/observability-diff.yml
@@ -1,0 +1,175 @@
+name: Observability — diff against staging
+
+# Runs on PRs touching packages/observability/** and posts a sticky comment
+# showing the diff between the PR's committed dashboards/folders and the
+# live state of the staging Grafana. Mirrors the existing terraform-plan
+# comment workflow.
+#
+# CS-10933 — implements the "Path B" diff (pull-to-tempdir + git-diff)
+# because grafanactl has no native `resources diff` subcommand. See
+# packages/observability/scripts/diff.sh for the rendering logic.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - "packages/observability/**"
+      - ".github/workflows/observability-diff.yml"
+
+permissions:
+  contents: read
+  id-token: write
+  pull-requests: write
+
+concurrency:
+  # Cancel an in-flight diff when a new push lands on the same PR — diff is
+  # cheap and we always want the comment to reflect the latest commit.
+  group: observability-diff-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  diff:
+    name: Diff against staging
+    runs-on: ubuntu-latest
+
+    env:
+      AWS_REGION: us-east-1
+      # Reuses the apply role from cardstack/infra#665. Permission is
+      # `ssm:GetParameter` on /staging/grafana/grafanactl_token only —
+      # which is what the diff workflow needs (read the token, then use
+      # it to pull dashboards/folders from staging Grafana).
+      AWS_ROLE_ARN: arn:aws:iam::680542703984:role/boxel-observability-apply
+      GRAFANACTL_VERSION: "0.7.0"
+      DIFF_MARKER: "<!-- observability-diff -->"
+      MAX_COMMENT_BYTES: 60000
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install grafanactl
+        # Same SHA-256 verified install as the apply workflows.
+        run: |
+          set -euo pipefail
+          TARBALL="grafanactl_${GRAFANACTL_VERSION}_linux_amd64.tar.gz"
+          RELEASE_URL="https://github.com/grafana/grafanactl/releases/download/v${GRAFANACTL_VERSION}"
+          TMPDIR="$(mktemp -d)"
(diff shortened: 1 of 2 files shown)


habdelra (approved):

lukemelia:
Addressed all 5 Copilot suggestions in commit 30cc0410:

1. **Fork-PR guard.** Added \`if: github.event.pull_request.head.repo.full_name == github.repository\` at the job level. Workflow no longer runs on fork PRs (would have failed at OIDC anyway, but explicit guard is the right pattern; matches other privileged workflows in the repo).

2. **Pin \`actions/github-script\` to SHA** (60a0d83 = v7.0.1). The step has \`pull-requests: write\` and runs after assuming a cloud role — defense-in-depth pinning matters more here than for unprivileged actions.

3. **Truncate by bytes, not UTF-16 code units.** Switched to \`Buffer.byteLength(diff, 'utf8')\` for the threshold check, and walk back to a UTF-8 character boundary when slicing so the truncated comment never lands mid-codepoint.

4. **Paginate \`listComments\`.** Switched to \`github.paginate(github.rest.issues.listComments, ...)\`. Marker-comment lookup is now reliable on PRs with >100 comments.

5. **\`diff.sh\` exit-code handling.** Wrapped \`git diff --no-index\` with \`set +e\` and explicit exit-code check: 0/1 = success (no diff / diff found, expected), 2+ = real error and propagate. Previous \`|| true\` would have masked unreadable-paths or git internal errors as success.

PR #4550, opened 2026-04-28
allow collecting binary file for submission

lucaslyl:
Previously, listing PR creation only had a text-file collection flow. That worked for normal source files, but not for binary assets like thumbnails.

There were two problems with the old behavior. First, we did not have a proper way to read and carry binary file contents through the submission flow. Second, binary content should not be stored inside the PR card payload, because that data can quickly grow the card and contribute to card size limit failures.

This change adds a dedicated ReadBinaryFile command so binary assets can be fetched as base64, and updates the listing PR flow to treat binary files separately from text files.

Text files continue to be collected through the existing PR card content flow. Binary files are now passed separately into addContentsToCommit and written directly to GitHub using base64 blob encoding.

This gives us a proper path for including binary submission assets in generated PRs without storing that binary payload inside PrCardFieldContent.

## Note:
i update the 'collection-submission-file' commands to https://github.com/cardstack/boxel/pull/4550

Screenshot 2026-04-28 at 2 25 29 PM

### DIFF
--- a/packages/bot-runner/lib/command-runner.ts
+++ b/packages/bot-runner/lib/command-runner.ts
@@ -1,4 +1,5 @@
 import {
+  BINARY_FILE_EXTENSIONS,
   isBotCommandFilter,
   logger,
   param,
@@ -147,10 +148,18 @@ export class CommandRunner {
           throw new Error(errorMessage);
         }
 
-        // Extract allFileContents from the result
+        // Extract allFileContents from the result, separating binary files
+        // so they bypass the PrCard 512KB size limit.
         let allFileContents = extractFileContents(filesResult.cardResultString);
+        let binaryFileContents = allFileContents.filter((f) =>
+          isBinaryFilename(f.filename),
+        );
+        let textFileContents = allFileContents.filter(
+          (f) => !isBinaryFilename(f.filename),
+        );
         log.info('pr-listing-create: files collected', {
           fileCount: allFileContents.length,
+          binaryCount: binaryFileContents.length,
         });
 
         // Step 1.5: Lint & auto-fix collected files
@@ -313,7 +322,7 @@ export class CommandRunner {
               branchName,
               submittedBy: runAs,
               prSummary,
-              allFileContents,
+              allFileContents: textFileContents,
             },
           });
 
@@ -346,6 +355,10 @@ export class CommandRunner {
           await this.createListingPRHandler.addContentsToCommit(
             eventContent,
             prCardResult,
+            binaryFileContents.map((f) => ({
+              path: f.filename,
+              content: f.contents,
+            })),
           );
           let prResult = await this.createListingPRHandler.openCreateListingPR(
             eventContent,
@@ -489,6 +502,13 @@ function getCardUrl(cardResultString?: string | null): string | null {
   } catch {
     return null;
   }
+}
+
+
+
+function isBinaryFilename(filename: string): boolean {
+  let ext = filename.slice(filename.lastIndexOf('.')).toLowerCase();
+  return BINARY_FILE_EXTENSIONS.has(ext);
 }
 
 function extractFileContents(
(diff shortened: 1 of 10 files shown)


packages/catalog-realm/commands/collect-submission-files.ts
richardhjtan, line 1: This should be move to `boxel-catalog` repository, this folder is deprecated

richardhjtan: defer the comment, noticed the changes in this PR: https://github.com/cardstack/boxel-catalog/pull/494/changes

lucaslyl: yep, that is in another pr in boxel-catalog

habdelra:
I think the software factory failure is related to an issue that @FadhlanR is working on right now

packages/runtime-common/file-def-code-ref.ts
```diff
@@ -7,6 +7,25 @@ import {
...
+  '.bmp',
+  '.tiff',
+  '.tif',
+  '.woff',
+  '.woff2',
+  '.ttf',
+  '.otf',
+  '.eot',
+  '.pdf',
+]);
```
habdelra: do we have a better way to maintain this? i'm pretty sure that something like this already exist for the supported file-def files (it might be split across different files though). let's not reinvent this as it might diverge from teh truth

packages/bot-runner/lib/command-runner.ts
```diff
@@ -147,10 +148,18 @@ export class CommandRunner {
           throw new Error(errorMessage);
         }
 
-        // Extract allFileContents from the result
+        // Extract allFileContents from the result, separating binary files
+        // so they bypass the PrCard 512KB size limit.
         let allFileContents = extractFileContents(filesResult.cardResultString);
+        let binaryFileContents = allFileContents.filter((f) =>
+          isBinaryFilename(f.filename),
+        );
```
habdelra, line 156: i think you can actually tell a binary file also from just looking inside of it. we use that here:

```
packages/host/app/resources/file.ts:439:                                                                                          
                                                                                                                                          
  function isBinary(content: string) {                                                                                                    
    return /\ufffd/.test(content);                                                                                                        
  }                                                                                                                                       
                                                                                                                                          
  It checks for the U+FFFD replacement character, which JS inserts when you decode bytes that aren't valid UTF-8 as a string. So it's     
  content-sniffing (not extension-based), and it relies on the file having already been decoded as a string. Used via the isBinary getter 
  at line 395 and consumed in code-semantics-service.ts:163.                                                                              
                             
  The realm-server middleware (packages/realm-server/middleware/index.ts:156) also has an isBinary check, but that one looks at the       
  request Content-Type header rather than the bytes.
                                                      
```

lucaslyl: Thanks for the feedback! When I looked closely, in this collect submission workflow the binary file content is already collected as base64, so the \ufffd replacement character that `isBinary` relies on would never appear - making it unable to distinguish binary from text files in this context.

packages/runtime-common/file-def-code-ref.ts
lucaslyl: Agreeed with this. Actually, I can see that `FILEDEF_CODE_REF_BY_EXTENSION` exists, but it's incomplete for our purpose since other binary file checks like font files and others like .ico and .pdf . So I've moved isBinaryFilename into infer-content-type.ts and derived it from inferContentType instead — mime-types already knows about all these formats, so there's nothing to maintain manually. Here is the commit, happy to receive your feedback! https://github.com/cardstack/boxel/pull/4550/changes/aea5e72faf98191d7c526733ea83f2d816652d97

habdelra (approved):

PR #4554, opened 2026-04-28
CS-10623: Reimplement `boxel realm watch` command

FadhlanR:
## Summary
- Port `boxel watch` from the standalone [`cardstack/boxel-cli`](https://github.com/cardstack/boxel-cli/blob/main/src/commands/watch.ts) into `packages/boxel-cli/src/commands/realm/watch.ts`
- **Namespaced under `realm` group** (was top-level `boxel watch` in the legacy CLI). The Claude Code plugin's skill copy needs to use `boxel realm watch`.
- Options: `-i, --interval <seconds>` (default 30), `-d, --debounce <seconds>` (default 5). Silencing uses the program-level `-q, --quiet` (`boxel -q realm watch …`), introduced in `lib/cli-log.ts` on main; the watch command does not declare its own `-q`.
- Single realm at the CLI surface (`<realm-url> <local-dir>`). The internal `watchRealms()` API takes an array of specs to keep room for a future multi-realm CLI; today's CLI passes a single spec and the resolved authenticator is shared, so multi-realm callers must use realms that share a profile / secret seed.

## `boxel realm stop` — out of scope, deferred to [CS-10624](https://linear.app/cardstack/issue/CS-10624)

This PR ships only the single-terminal Ctrl+C path. SIGINT cleanly tears down the timers, releases the lock, and exits — same shape as the legacy CLI, which also ran `watch` in the foreground.

`boxel realm stop` (CS-10624) solves a different problem and is still wanted once `boxel realm track` ([CS-10622](https://linear.app/cardstack/issue/CS-10622)) lands: a cross-terminal kill switch that finds every running watch/track process (the legacy walked a process registry and SIGINTed each registered PID) and shuts them all down at once. Out of scope here.

What this PR *does* add for safety is a **concurrency guard**, not a stop mechanism: `.boxel-watch.lock` is written into the local-dir while a watch is active, so a second `boxel realm watch` against the same dir refuses to start (and overwrites a stale lock from a non-running pid). The lock records `{ pid, startedAt, realmUrl }` so CS-10624 can later use it as one discovery source.

Cross-command coordination (pull/push/sync warning when a watch is active against the same dir) is also out of scope here — separate ticket if/when wanted.

## Test plan
- [x] `pnpm --filter @cardstack/boxel-cli test:integration` passes (12 watch cases: add/modify/delete/burst/loop/abort/error paths/poll-error-doesn't-delete/pending-modify→delete-supersedes/lock-blocks-second-watch/stale-lock-overwrite)
- [x] `pnpm --filter @cardstack/boxel-cli build` succeeds
- [x] `boxel realm watch --help` documents `-i`, `-d` (silencing is via the program-level `boxel -q …`)
- [x] Manual: `boxel realm watch <staging-url>`, edit a card via Boxel web UI, confirm local pull within ~30s and a `[remote]` checkpoint is created.

## Depends on
- [CS-10625](https://linear.app/cardstack/issue/CS-10625) for checkpoint creation on detected changes — already merged when this implementation landed; uses `CheckpointManager` directly.


habdelra:
One thing we need to be careful for anytime we wire up subscription logic is teardown. this is one of the most common avenues for memory leaks, as failing to teardown properly means that you retain references to things that are now out of context. these can be really hard to spot, can you do me a favor and ask claude to "audit for proper teardown of all subscriptions to ensure that we have no memory leaks"--thanks!

lukemelia (approved):

FadhlanR:
@habdelra — audit done, one real race found and fixed in 95c7cdae14.

**Surfaces & status**

| Surface | Status | Notes |
|---|---|---|
| Per-watcher `debounceTimer` | ✅ | Cleared on every reschedule and in `shutdown()`. |
| Poll-loop `timeoutId` (self-rescheduling) | ✅ | Both the scheduler entry and the inner setTimeout body check `stopped` before doing work. |
| `process.on('SIGINT'/'SIGTERM')` | ✅ | Registered only when no `signal` is supplied; removed in `cleanup()` via `process.off`. |
| External `AbortSignal` `'abort'` listener | ✅ | `{ once: true }` auto-removes after firing; an already-aborted signal short-circuits before registration. |
| `RealmAuthenticator` (ProfileManager / SeedAuthenticator) | ✅ | Stateless `fetch` wrappers — no Matrix client, no token-refresh interval, no websockets, no listeners. |
| `RealmSyncBase` / `sync-manifest` | ✅ | Pure async I/O — no `chokidar` / `fs.watch`; `pLimit(10)` only holds in-flight promises. |
| `CheckpointManager` | ✅ | Spawns `git`; all `stdout`/`stderr`/`close` listeners are scoped to the single `git()` call and torn down when the child exits. |
| Filesystem watch lock | ✅ | Acquired pre-init, rolled back on init failure, released best-effort in `cleanup()`. |
| **In-flight `poll()` re-arming `debounceTimer` post-shutdown** | ⚠️ → ✅ | **Fixed in this commit.** |

**The race**

In `tickAll`, the per-watcher branch is `await w.poll()` then `w.scheduleFlush(...)` if there are changes. If SIGINT or `signal.abort` fires while `poll()` is awaiting an HTTP fetch:

1. `cleanup()` runs synchronously through `stopped = true` → `clearTimeout(timeoutId)` → `w.shutdown()` (clears any **existing** `debounceTimer`) → release lock → `resolve()`.
2. `watchRealms()` returns.
3. **Then** the in-flight `poll()` resolves. Its continuation calls `w.scheduleFlush(...)`, arming a brand-new `debounceTimer` that nothing clears.
4. After `debounceMs` the timer fires `flushPending()` — downloads, manifest write, checkpoint — **after the lock is released and the user thinks the watch is stopped**.

Exactly the failure mode you flagged: a reference held past its lifetime, doing work that escapes teardown.

**Fix**

Watcher-local `isShutdown` flag, set at the top of `shutdown()` and checked at the top of `scheduleFlush()`. Order matters: the flag is set **before** clearing the timer so a concurrent `scheduleFlush()` racing the in-flight poll observes shutdown state. `flushPending()` and `poll()` are deliberately not gated — direct callers (tests, programmatic single-shot use) still drive them; only the *scheduled* path needs the gate.

**New tests**

- **Abort mid-poll race repro** — uses the `authenticator` test hook to gate the 3rd `_mtimes` call, aborts while the poll is blocked, releases the gate, then asserts no extra checkpoint and no content change. Fails without the gate, passes with it.
- **SIGINT handler cleanup** — runs `watchRealms()` without a `signal` so the SIGINT/SIGTERM branch registers, snapshots existing listeners, invokes the registered handler directly (avoids tripping vitest's own SIGINT handling), and asserts the listener set returns to baseline.

PR #4556, opened 2026-04-28
Software factory: fix issue scheduler skipping backlog items after blocker completes

jurgenwerk:
As I was exercising the software factory post https://github.com/cardstack/boxel/pull/4492 merge, specifically I was running the `software-factory/Wiki/kanban-board` example, I discovered a bug where the kanban card got created (status = done), but the kanban board was stuck (status = backlog) after the software factory run finished. This fixes the bug so that the scheduler continues working through the backlog. 

## Summary

The scheduler had two ways of writing the same issue's ID, and they didn't match.                                                                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                                           
When it loaded an issue from the realm, it stored its own ID as a full URL: http://localhost:4201/user/my-test-realm/Issues/kanban-card.                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                                                           
But when it read another issue's blockedBy (the list of issues that must finish first), it stored those IDs as relative paths: ../Issues/kanban-card.                                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                                                                           
Then when checking "is this blocker done?", it tried to look up ../Issues/kanban-card in a map keyed by the full URL. No match. The scheduler concluded "blocker status unknown, treat as not done", and skipped the dependent issue. Forever.                                                                                                                                                                                                           
                                                                             
The fix: when reading blockedBy, resolve the relative path against the parent issue's URL so both sides use the same full URL. Now the lookup matches.

### DIFF
--- a/packages/software-factory/src/issue-scheduler.ts
+++ b/packages/software-factory/src/issue-scheduler.ts
@@ -362,17 +362,30 @@ export class RealmIssueStore implements IssueStore {
 // ---------------------------------------------------------------------------
 
 /**
- * Extract card IDs from a Boxel linksToMany relationship.
+ * Extract card IDs from a Boxel linksToMany relationship, resolved
+ * against the parent card's URL so the result matches `card.id` from
+ * the realm search index.
  *
  * Boxel encodes linksToMany with dotted keys:
  *   "blockedBy.0": { links: { self: "../Issues/abc" } }
  *   "blockedBy.1": { links: { self: "../Issues/def" } }
  *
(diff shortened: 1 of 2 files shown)


jurgenwerk:
Test failure unrelated to this PR (it's coming from upstream, I think Fadhlan is working on it)

backspace (approved):

habdelra (approved):

PR #4580, opened 2026-04-30
host: keep StackItem on isolated/edit format toggle

christse:
Toggling between isolated and edit on a stack item used to clone the StackItem and replace it via replaceItemInStack. The new identity caused Glimmer's {{#each}} to destroy + remount the entire stack-item subtree -> CardRenderer remounted, scroll position lost, prefersWideFormat re-evaluated against a momentarily-undefined card, width snapped.

Mutate item.format in place via @tracked instead. The StackItem identity is preserved, the {{#each}} keeps its component instance, and CardDefs whose `static isolated === static edit` re-render as a pure prop change rather than a remount.

Demo: https://screen.studio/share/XOqnzIbP

### DIFF
--- a/packages/host/app/components/operator-mode/stack-item.gts
+++ b/packages/host/app/components/operator-mode/stack-item.gts
@@ -546,13 +546,11 @@ export default class OperatorModeStackItem extends Component<Signature> {
     if (this.card) {
       request?.fulfill(this.card.id);
     }
-    this.operatorModeStateService.replaceItemInStack(
-      item,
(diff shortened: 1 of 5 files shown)


jurgenwerk:
@copilot apply changes based on the comments in [this thread](https://github.com/cardstack/boxel/pull/4580#pullrequestreview-4204730259)

habdelra (approved):

jurgenwerk (approved):

PR #4616, opened 2026-05-01
Host pill format chooser

christse:
## Demo

https://screen.studio/share/vCKDCvxI

## Summary

New format selector for the code-mode preview + playground: a single morphing pill with drag-to-select, hover preview, and compact mode for narrow panels. Also wires the synthetic "Form" pill (auto-generated standard view) all the way through to the renderer.

## Highlights

- **PillFormatChooser** — single morphing capsule, drag commits live across buttons, synthetic-click suppression so drag-to-different-format actually sticks. Compact mode (icon-only, instant tooltips) flips on automatically when the parent narrows; flips back when it grows.
- **"Form" pill renders the auto-generated standard view** — same mechanism interact mode's "Toggle Standard View" uses (`baseCardRef` as the renderer's `@codeRef`). Previously the pill rendered blank.
- **Form pill = editing-mode header** — `@onFinishEditing` now treats `'form'` like `'edit'`, so the green CardHeader bar + X-to-finish-editing show up just like interact mode.
- **Centering shell rework** — `.card-renderer-format-chooser` is now a full-width flex shell with `pointer-events: none`; the dark capsule + radius lives on the chooser itself. Required so compact-mode detection has an honest available-width signal (a `width: max-content` parent was creating a stuck-compact loop).

## Test plan

- [ ] Code-mode preview, card with custom edit template: chooser shows Form pill; selecting it renders the auto-generated standard view (not blank), with green editing header.
- [ ] Drag across the chooser; releasing on a different button keeps that format.
- [ ] Narrow the preview panel until the chooser flips to compact icon-only; hover any button → format-name tooltip appears instantly above. Widen the panel → chooser restores icon+text without sticking.
- [ ] Drag in compact mode → tooltip follows the active button.
- [ ] Existing format-chooser tests still pass (the new chooser keeps the `data-test-format-chooser="<fmt>"` selectors).

## Known follow-ups

- Tests outlined earlier in the design discussion — at minimum the drag-suppress-trailing-click integration test and the form→standard-view acceptance test.


backspace:
> Narrow the preview panel until the chooser flips to compact icon-only; hover any button → format-name tooltip appears instantly above. Widen the panel → chooser restores icon+text without sticking.

This is true but when I make it narrow enough for this to happen, Markdown format is lost and the chooser isn’t scrollable; I’m not sure it would make sense for it to be scrollable, but:

CleanShot 2026-05-11 at 08 55 01

backspace (approved):
I think the problem I mentioned deserves some thought but I don’t think it’s blocking as it’s an edge case

PR #4626, opened 2026-05-01
host: expand isolated view to full width with floating top bar

christse:
Demo: https://screen.studio/share/zXdc2bHg

Adds an "expand" toggle on the top stack-item's CardHeader that morphs the card to full width below a floating top bar. The bar gains glass- morphism while expanded and the CardHeader portals into a center pill in the bar (replacing the inline header). View-transition pins the bar across the morph so the expand/collapse animates as a single shape.

Positioning is driven by a single CSS variable, --stack-padding-top (item-height + 2 × spacing), shared by:
  * .submode-layout-top-bar height (locked, so portaled content can't push it taller)
  * .operator-mode-stack padding-top when an item is expanded (forced full value, overriding stack-medium/small-padding-top reductions that are correct for the buried-stack visual but not when expanded)
  * .item.expanded inset: 0 of .inner (which sits at y=stack-padding-top)

Result: card top sits flush at bar bottom — pixel-identical to host-mode's in-flow card top. No calc(100vh - X), no JS measurement, no !important.

Buried siblings fade when a top card is expanded; neighbor-stack-trigger arrows hide; cards-grid (Workspace "All Cards") opts out of expand since its child tile overlays would overlap. Per-card expand intent persists on operatorModeStateService (keyed by stack-item id) so it survives buries and restores when the card pops back to the top.

### DIFF
--- a/packages/boxel-ui/addon/src/components/card-header/index.gts
+++ b/packages/boxel-ui/addon/src/components/card-header/index.gts
@@ -320,7 +320,6 @@ export default class CardHeader extends Component<Signature> {
           flex-shrink: 1;
           min-width: 0;
           text-align: center;
-          padding-inline: 30px;
         }
         .card-type-display-name-text {
           font: 700 var(--boxel-font-sm);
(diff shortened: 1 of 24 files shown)


lukemelia (commented):
I thought that we were only going to show the Maximize icon in the header after you had expanded, and were going to have a menu item for to initiate expanding.

burieberry:
> I thought that we were only going to show the Maximize icon in the header after you had expanded, and were going to have a menu item for to initiate expanding.

That is how it is.

burieberry:
Small screen UI top bar without the expanded card:
without-bar

Collapse/restore is via the dropdown in small screen UI and buttons collapse further:
with-bar

lukemelia:
> That is how it is.

Sorry -- preview URL was cached in my browser!

lukemelia (approved):

PR #4539, opened 2026-04-27
Add more RRI handling and casts

backspace:
This updates many places that use `URL` to accept `RRI` to reduce assumptions that card/module identifiers are URLs, another step toward delaying resolution of `@cardstack/*` identifiers into URLs until the network layer.

Some other widening pieces split off for followup:
* `relativeTo` (CS-10966)
* use of `startsWith` to determine realm membership (CS-10982)

PR #4628, opened 2026-05-01
Update RealmPaths#inRealm to handle both formats

backspace:
This was split off #4539, although it’s so small it could have been included!

Substring-prefix checks like `card.id.startsWith(realm)` only work when both sides are in the same form (URL vs registered prefix). After RRI migration, IDs may be prefix-form while realm URLs are URL-form (or vice versa), silently breaking those checks.

`inRealm` resolves both sides to the same form comparisons work.


packages/host/app/services/realm.ts
```diff
@@ -951,6 +951,29 @@ export default class RealmService extends Service {
...
+  //
+  // `realm` may be in URL form (`http://...`) or as a registered realm-root
+  // identifier/prefix (for example `@cardstack/base/`, including the trailing
+  // slash). `resource` may be in URL form or registered-prefix form.
+  // Internally both sides are resolved to URL form via `cardIdToURL` before
+  // comparison, so a URL-form realm correctly matches a prefix-form resource
+  // (and vice versa) as long as the prefix is registered in `prefixMappings`.
+  // Returns `false` if either side fails to resolve, including when a
+  // non-URL-like string does not match a registered prefix exactly.
+  contains(realm: string | URL, resource: string | URL): boolean {
```
lukemelia: Is there a named type we should use here?

packages/host/app/services/realm.ts
```diff
@@ -951,6 +951,29 @@ export default class RealmService extends Service {
...
+    let realmHref: string;
+    let resourceHref: string;
+    try {
+      realmHref = realm instanceof URL ? realm.href : cardIdToURL(realm).href;
+      resourceHref =
+        resource instanceof URL ? resource.href : cardIdToURL(resource).href;
+    } catch {
+      return false;
+    }
+    return new RealmPaths(new URL(realmHref)).inRealm(rri(resourceHref));
```
lukemelia: Isn't there a util that can do this with with RRIs?

lukemelia (approved):

packages/host/app/services/realm.ts
backspace: 👍🏻 in 50bf89b this now accepts RI and RRI

packages/host/app/services/realm.ts
backspace: you’re right, as of bd942a3 `inRealm` now has proper typing to be able to take this over, `RealmService.contains` no longer needs to exist, thanks!

PR #4641, opened 2026-05-04
Re-land CS-10948: discard in-flight prerender result if invalidate ran concurrently

lukemelia:
Re-lands [#4515](https://github.com/cardstack/boxel/pull/4515) (CS-10948 — discard in-flight prerender result if invalidate ran concurrently) by reverting the [#4527](https://github.com/cardstack/boxel/pull/4527) revert. The race the original PR closes still exists in `main` today.

## Why the original was reverted, and why that reasoning no longer applies

[#4527](https://github.com/cardstack/boxel/pull/4527) reverted both this and [#4521](https://github.com/cardstack/boxel/pull/4521) (CS-10860) "to bisect" two failing SF Playwright tests (`realm-search` 60s tool timeouts). The revert PR was explicit that root cause was unknown.

Within ~2.5 hours of the revert, [#4528](https://github.com/cardstack/boxel/pull/4528) ("SF harness: preserve cloned modules cache to avoid cold prerender per test") root-caused those failures. The cause was unrelated to either reverted PR — `rewriteClonedRealmServerUrls` in the SF harness covered seven URL-bearing tables but had **left out the `modules` table**, so every per-test `lookupDefinition` was a structural cache miss against template-baked URLs and triggered a 45–55 s cold prerender that exceeded the 60 s tool timeout. Plus `realm-server/main.ts` was unconditionally calling `clearAllModules()` on every boot, wiping anything that did get rewritten. Both were fixed in [#4528](https://github.com/cardstack/boxel/pull/4528).

CS-10860 was re-landed cleanly the next day in [#4543](https://github.com/cardstack/boxel/pull/4543) ("Resurrect desync detector"), which described the original revert as having been done "out of an abundance of caution while we sorted out the signal." CS-10948 was the only one that didn't get picked back up — and three days later [CS-10952](https://linear.app/cardstack/issue/CS-10952) and [CS-10953](https://linear.app/cardstack/issue/CS-10953) were filed assuming the generation counters were still in the code. They aren't, today, on `main`. This PR puts them back.

## What's in this PR

`git revert` of the revert merge `5cac64e152`. Auto-merge resolved cleanly against the four intervening PRs that touched the same files:

- [#4564](https://github.com/cardstack/boxel/pull/4564) — clamp error_doc on persistence
- [#4539](https://github.com/cardstack/boxel/pull/4539) — RRI types/casts
- [#4620](https://github.com/cardstack/boxel/pull/4620) (CS-11002) — modules.timing_diagnostics column

[...]

PR #4644, opened 2026-05-04
CS-10952: cross-process invalidation broadcast for CachingDefinitionLookup

lukemelia:
First sub-issue under the [CS-10950](https://linear.app/cardstack/issue/CS-10950) umbrella ("Cross-process coordination for CachingDefinitionLookup"). Closes the invalidation half. CS-10953 (cross-process populate coalescing via advisory-lock + NOTIFY) stacks on top.

## Why

CS-10948 added in-process generation counters (`#moduleGenerations`, `#realmGenerations`, `#globalGeneration`) to discard an in-flight prerender result if a concurrent invalidate/clearRealmCache/clearAllModules ran during the prerender. That fixes the persist-after-invalidate race within one realm-server process.

Across N processes the protection doesn't span instances. If process A is mid-prerender for `foo.gts` and process B calls `invalidate('foo.gts')`, B bumps its own counters and deletes the row, but A's counters are untouched — A's persist-time generation check sees no mismatch and the UPSERT re-inserts the row B just deleted.

This PR broadcasts every invalidation over a Postgres NOTIFY channel so peer processes' counters stay in lockstep with the DB.

### `runtime-common/definition-lookup.ts`

- Adds an exported `MODULE_CACHE_INVALIDATED_CHANNEL = 'module_cache_invalidated'` constant.
- Promotes `bumpModuleGeneration` and `bumpRealmGeneration` from `private` to public (so the listener can replay them); adds a `bumpGlobalGeneration()` helper for symmetry.
- Emits `pg_notify` after each invalidation path's DELETE:
  - `invalidate(moduleURL)` → one notify per URL in `uniqueInvalidations` (target + dependency-graph fan-out): `module:<resolvedRealmURL>:<moduleURL>`
  - `clearRealmCache(url)` → single `realm:<resolvedRealmURL>` notify
  - `clearAllModules()` → single `global` notify
- Best-effort try/catch with console.warn (mirroring `Realm.#notifyFileChange`). The local instance's counters are already bumped synchronously before the DELETE, so a notify failure is a bounded cross-instance staleness window — peers self-heal on their next prerender.
- Sqlite/in-memory `dbAdapter.kind` short-circuits before `pg_notify`.

### `realm-server/lib/module-cache-invalidation-listener.ts` (new)

`ModuleCacheInvalidationListener` follows `RealmFileChangesListener`'s shape exactly:
- Dedicated LISTEN connection via `PgAdapter.listen` (Pool LISTEN is unreliable per [node-postgres#1543](https://github.com/brianc/node-postgres/issues/1543)).
- `WorkLoop` for predictable shutdown.
- 60s safety poll (no DB-side state to poll — the dispatch is in the payload — so the wake-loop just sleeps until shutdown).
- Parses the three payload formats and dispatches to `bumpModuleGeneration` / `bumpRealmGeneration` / `bumpGlobalGeneration` on the attached `CachingDefinitionLookup`.

Self-notify is idempotent: the emitter already bumped synchronously before its DELETE, and a second monotonic bump from receiving its own NOTIFY is observationally equivalent (counters are only used for snapshot equality).

### `realm-server/main.ts`

Constructs and `start()`s alongside `fileChangesListener`; adds `.shutDown()` to the existing shutdown `Promise.all`.

### A small structural divergence from the ticket

The CS-10952 spec suggested `packages/runtime-common/definition-lookup-coordination.ts`. `runtime-common` doesn't depend on `@cardstack/postgres` (and shouldn't — `@cardstack/postgres` already depends on `runtime-common` via `DBAdapter`, so the dep would be circular). The listener needs `PgAdapter.listen` and `WorkLoop`, both from `@cardstack/postgres`, so it lives in `realm-server/lib/` next to `RealmFileChangesListener`. Behavior is identical; only the file home differs. CS-10953's sibling listener will land in the same spot. The channel constant is exported from `runtime-common` so other packages can import it.

## A small spec divergence on the NOTIFY transaction

The spec says: *"`NOTIFY ... '<payload>'` inside the same transaction as each invalidation path's DELETE, so peers only see the signal on commit."*

Each existing invalidation path (`invalidate` / `clearRealmCache` / `clearAllModules`) is a single autocommit DELETE — no `BEGIN`/`COMMIT` wrapper. The existing `Realm.#notifyFileChange` precedent emits `pg_notify` as a separate query *after* the autocommit. Same observable semantics under non-crash conditions: peer sees the notify after the DELETE has committed, because both autocommit in order. Mirroring `#notifyFileChange` keeps the two cross-instance pathways using the same idiom and avoids new tx-management code. The narrow "DELETE commits, then process crashes before pg_notify" window is identical to what the file-changes path already accepts.

## Behavior

**N=1 (today's production):** inert. The emitter bumps synchronously before its DELETE; the listener replays a second monotonic bump that any in-flight snapshot would have already mismatched on the first bump.

**N>1:** immediate correctness win. Peer processes' in-flight prerenders observe the invalidation via NOTIFY before persist; their generation checks catch the mismatch and discard via `readFromDatabaseCache` instead of re-inserting. The only remaining window is bounded by NOTIFY latency (typically sub-millisecond on healthy Postgres) and self-heals on the next prerender of the same key.

## Test plan

- [ ] CI Realm Server suite green
- [ ] CI Software Factory job green
- [ ] CI Host suite green

New tests in `realm-server/tests/module-cache-invalidation-listener-test.ts`:

- **Payload parsing** units (the four shapes + malformed/empty cases).
- **Dispatch** units: `handleNotification` with each payload kind invokes the right bump on a stub lookup with the right args.
- **End-to-end with PgAdapter** (same pattern as `realm-file-changes-listener-test.ts`): two `CachingDefinitionLookup` instances share one DB; A's `invalidate(url)` / `clearRealmCache(url)` / `clearAllModules()` produces the right bump on B's recorder via the listener.
- **Self-notify idempotence**: emitter sees ≥2 bumps for the same URL (1 synchronous + 1 listener echo).
- Manual `pg_notify` sends through the channel constant and the listener picks them up — proves wire compatibility for any future emitter outside `CachingDefinitionLookup`.

## What this does NOT do (out of scope)

- Cross-process populate coalescing (the prerender-coalescing half of the work). Tracked as [CS-10953](https://linear.app/cardstack/issue/CS-10953) — stacks on this PR.
- Strict-consistency cross-process invalidation closure beyond what NOTIFY latency provides; the sub-millisecond window remains.

## Related

- [CS-10950](https://linear.app/cardstack/issue/CS-10950) — umbrella
- [CS-10948](https://linear.app/cardstack/issue/CS-10948) (#4641) — in-process generation discard, the foundation this builds on
- [CS-10892](https://linear.app/cardstack/issue/CS-10892) — `realm_file_changes` NOTIFY pattern this mirrors
- [CS-10953](https://linear.app/cardstack/issue/CS-10953) — sibling sub-issue, stacks on this

### DIFF
--- a/packages/realm-server/lib/module-cache-invalidation-listener.ts
+++ b/packages/realm-server/lib/module-cache-invalidation-listener.ts
@@ -1,0 +1,162 @@
+import {
+  logger,
+  MODULE_CACHE_INVALIDATED_CHANNEL,
+  type CachingDefinitionLookup,
+} from '@cardstack/runtime-common';
+import type { PgAdapter } from '@cardstack/postgres';
+import { WorkLoop } from '@cardstack/postgres';
+
+const log = logger('realm-server:module-cache-invalidation-listener');
+const DEFAULT_POLL_INTERVAL_MS = 60_000;
+
+// Cross-instance module-cache invalidation broadcast (CS-10952). Peer
+// realm-server processes emit `NOTIFY module_cache_invalidated, '<payload>'`
+// from CachingDefinitionLookup.invalidate / clearRealmCache / clearAllModules
+// after their DELETE commits; this listener parses the payload and replays
+// the appropriate generation bump on the locally-attached
+// CachingDefinitionLookup so its in-flight prerenders observe the
+// invalidation at persist time and discard stale results instead of
+// re-inserting the row a peer just deleted.
+//
+// Mirrors RealmFileChangesListener exactly: dedicated LISTEN connection
+// (PgAdapter.listen uses a fresh Client to dodge pool-LISTEN reliability
+// issues — see node-postgres#1543), WorkLoop for predictable shutdown, 60s
+// safety poll. There's nothing to poll from the DB side — the entire
+// dispatch is in the payload — so the wake-loop just sleeps until shutdown.
+//
+// Self-notify is harmless: the emitting process bumps its counter
+// synchronously before its DELETE, so the listener's bump on receiving its
+// own NOTIFY is a second bump on a counter that's only used for snapshot
+// equality. Idempotent.
+export interface ModuleCacheInvalidationListenerDeps {
+  dbAdapter: PgAdapter;
+  definitionLookup: CachingDefinitionLookup;
+  // Optional for tests.
+  pollIntervalMs?: number;
+}
+
+export type ParsedModuleCacheInvalidation =
+  | { kind: 'module'; resolvedRealmURL: string; moduleURL: string }
+  | { kind: 'realm'; resolvedRealmURL: string }
+  | { kind: 'global' };
+
+export class ModuleCacheInvalidationListener {
+  #deps: ModuleCacheInvalidationListenerDeps;
+  #loop: WorkLoop;
+  #started = false;
+
+  constructor(deps: ModuleCacheInvalidationListenerDeps) {
+    this.#deps = deps;
+    this.#loop = new WorkLoop(
+      'module-cache-invalidation',
+      deps.pollIntervalMs ?? DEFAULT_POLL_INTERVAL_MS,
+    );
+  }
+
+  start(): void {
+    if (this.#started) {
+      return;
+    }
+    this.#started = true;
+    this.#loop.run(async (loop) => {
+      await this.#deps.dbAdapter.listen(
+        MODULE_CACHE_INVALIDATED_CHANNEL,
+        (notification: { payload?: string }) => {
+          this.#handleNotification(notification.payload);
+        },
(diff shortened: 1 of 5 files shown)


habdelra:
do we need to deal with sqlite inadvertently encountering a pg pub/sub for this in the host tests for in-browser realms?

lukemelia:
> do we need to deal with sqlite inadvertently encountering a pg pub/sub for this in the host tests for in-browser realms?

sqlite notify is a no-op. It doesn't exist.

habdelra (approved):

PR #4672, opened 2026-05-05
Move cross-instance NOTIFY behind DBAdapter.notify

lukemelia:
## Summary

`Realm.#notifyFileChange` was issuing raw `SELECT pg_notify(...)` SQL against `this.#dbAdapter`. In the host/browser context the adapter is SQLite, which has no `pg_notify` function — every write logged `SQLITE_ERROR: no such function: pg_notify`. The host runs a single in-process realm with no peers to notify, so issuing the SQL was also semantically a no-op even when it didn't error.

This pushes the cross-instance broadcast capability down into the existing `DBAdapter` abstraction:

- Add `notify(channel, payload): Promise<void>` to the `DBAdapter` interface, documented as best-effort cache-coherency.
- `PgAdapter` implements via `SELECT pg_notify($1, $2)`.
- `SQLiteAdapter` implements as a no-op (no pub/sub primitive, no peers).
- `Realm.#notifyFileChange` calls `this.#dbAdapter.notify(...)` and no longer embeds pg-specific SQL.
- Update the realm-server listener test to drive notifications via the new method (also gives `PgAdapter.notify` direct CI coverage).
- Add `notify` to the inline `DBAdapter` mocks in bot-runner / realm-server / runtime-common tests.

Surfaced as flaky test-log noise after PR #4660 (CS-10892: realm_file_changes NOTIFY channel) widened the call site without gating SQLite. Original diagnosis and the simpler `kind === 'sqlite'` guard fix were Hassan's, on `cs-11036-indexer-resume-progress`; this PR replaces that guard with the adapter-level abstraction.

## Test plan

- [x] Host shard 16 runs cleanly with no `no such function: pg_notify` lines in the test log.
- [x] Realm-server `realm-file-changes-listener-test.ts` still passes end-to-end against real Postgres (now exercising `PgAdapter.notify`).
- [x] Full realm-server, bot-runner, and runtime-common test suites green (mock DBAdapter shapes updated).

### DIFF
--- a/packages/bot-runner/tests/bot-runner-test.ts
+++ b/packages/bot-runner/tests/bot-runner-test.ts
@@ -112,6 +112,7 @@ module('timeline handler', () => {
 
   dbAdapter = {
     kind: 'pg',
+    notify: async () => {},
     isClosed: false,
     execute: async (sql: string, opts?: ExecuteOptions) => {
       if (sql.includes('FROM bot_registrations br')) {
--- a/packages/bot-runner/tests/command-runner-test.ts
+++ b/packages/bot-runner/tests/command-runner-test.ts
@@ -84,6 +84,7 @@ module('command runner', () => {
     ]);
     let dbAdapter = {
       kind: 'pg',
+      notify: async () => {},
       isClosed: false,
       execute: async (sql: string, opts?: ExecuteOptions) => {
         if (sql.includes('FROM bot_commands WHERE bot_id =')) {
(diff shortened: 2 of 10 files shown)


backspace:
This is unrelated to your PR, I’ll look more at how it escaped the colour-ignoring changes:

CleanShot 2026-05-06 at 06 49 54@2x

backspace (approved):

PR #4683, opened 2026-05-06
CS-11010: ETag/304 for application/vnd.card+json

habdelra:
Closes [CS-11010](https://linear.app/cardstack/issue/CS-11010). Companion to #4667 ([CS-11037](https://linear.app/cardstack/issue/CS-11037)).

## How this fits with #4667

#4667 lets a `_federated-search` caller send `include: []` to skip the server-side side-loading of relationships entirely. The big win there is the cards-grid prerender path: 87% of staging render-timeouts trace to that one query shape, each timeout burning the full 90 s budget, because the server fan-outs N×M sequential link queries even when the consumer reads only `instance.id`/`instance.constructor`.

But once a caller opts out of side-loading, any *other* relationship it later wants has to be fetched individually as `GET /<card>` with `Accept: application/vnd.card+json`. **Today every one of those individual GETs is served with `cache-control: no-store, no-cache, must-revalidate` and a fresh full-body response.** A page that does N follow-up link fetches pays the realm's full ~100 ms JSON-assembly cost on every page view, every repeat visit, even when nothing changed.

This PR closes that loop:

1. #4667: `_federated-search` returns `data[]` only, no `included[]`, no `loadLinks` work.
2. **This PR**: any follow-up `GET /<card>` for a linked card gets an ETag and short-circuits to 304 on a warm cache.

So the cost shifted off `_federated-search` in #4667 doesn't reappear as N unconditional GETs on the very next render — it lands in the browser cache and stays there until the underlying card actually changes.

## Scope difference vs. how #4667's description framed CS-11010

#4667's "Companion work" section described CS-11010 as making `vnd.card+json` ETag-aware **on published realms**. After looking at it, I went broader: this PR applies to **all realms**. Reasoning is in the [ticket comment](https://linear.app/cardstack/issue/CS-11010#comment) — modules and source files are already ETagged in every context (interactive, auth-gated, mutating realms included), so cards should follow the same convention. There's no realm class where instance data has a fundamentally different cacheability story than its module/source siblings.

## Behavior

- **GET** peeks at the index row first (one cheap row read), computes the ETag from the row's `indexed_at`, and short-circuits to 304 before the expensive `loadLinks` fan-out. Cache-hit cost drops from ~100 ms+ to ~5 ms.
- **PATCH** (both the no-op short-circuit path and the post-write path) emits the ETag too, so a caller using the PATCH response body can revalidate against it without an extra round-trip GET.
- **ETag format**: `"<indexed_at>:card"`. Matches the existing `<base>:<variant>` convention from `MODULE_ETAG_VARIANT` (`<lastModified>:module`) and `SOURCE_ETAG_VARIANT` (`<contentHash>:source`) in `runtime-common/realm.ts`.
- **Cache-Control**: `public, max-age=0, must-revalidate` for world-readable realms, `private, max-age=0, must-revalidate` for auth-gated ones, mirroring the source/module file path. `private` keeps a shared cache (CDN, intermediate proxy) from serving one user's response to another.

## Why `indexed_at` is the right ETag base

The `indexed_at` column on a card's `boxel_index` row is set on every reindex. The realm's dependency-invalidation chain (the `deps` array on each row) means **any** change that affects the assembled card+json document — including a write to a transitively-linked card — triggers a reindex and bumps `indexed_at` on the dependent rows. So `indexed_at` is a complete fingerprint for the assembled `data + included[]` document.

Two alternatives I considered and rejected:

- **`last_modified` alone** (the file's mtime). Untouched by dep-only reindexes, so a 304 here would serve stale `included[]` data after a linked card was updated. Wrong.
- **md5 of the assembled JSON body** (matching the literal "lastModified + size + hash" pattern from the ticket). Correct, but you can't compute it without first paying the full assembly cost we're trying to skip — defeats the purpose.

`indexed_at` is the cheap, correct middle: one column read on the primary row, computed before `loadLinks` runs.

## Wire path

```
GET /<card> (Accept: vnd.card+json)
  ↓
Realm.getCard (runtime-common/realm.ts)
  ↓
realmIndexQueryEngine.instance(url, { includeErrors: true })  ← cheap peek
  ↓
buildEtag(instance.indexedAt, 'card')  → "<n>:card"
  ↓
if If-None-Match matches → return 304 with ETag + Cache-Control + last-modified  ← skip loadLinks
  ↓
otherwise: realmIndexQueryEngine.cardDocument(url, { loadLinks: true })  ← full assembly
  ↓
return 200 with same ETag + Cache-Control + body
```

`SearchResultDoc` (in `realm-index-query-engine.ts`) gains an `indexedAt: number | null` field so the PATCH path can emit the ETag from a single `cardDocument` call without a redundant `instance` lookup.

## Test plan

- [x] `pnpm lint` clean (`runtime-common`, includes `lint:js` + `lint:types`).
- [x] Type-checks pass.
- [ ] CI runs the new ETag tests in `packages/realm-server/tests/card-endpoints-test.ts`. Local runs of the full card-endpoints suite were blocked by an unrelated environmental flake on this machine ("No standby page available for prerender" — Puppeteer/Chrome can't open new tabs, almost certainly an Ubuntu 24.04 + AppArmor-restricted-userns interaction). The new test cases themselves are scoped to request/response contracts and don't depend on prerender behavior; CI should exercise them cleanly.

### New test cases

GET, public-readable realm:
- `returns an ETag and public cache-control on a 200 response`
- `returns 304 when If-None-Match matches the current ETag` — verifies status, ETag echo, cache-control passthrough, and empty body
- `returns 200 when If-None-Match does not match`

GET, auth-gated realm:
- `200 with permission` (extended) — asserts `private` cache-control and ETag presence

PATCH, public-writable realm:
- `PATCH response carries an ETag and writes invalidate the previous one` — verifies (a) post-write ETag differs from pre-write, (b) the old ETag no longer 304s after the write, (c) the new ETag does 304 a follow-up GET
- `no-op PATCH response carries an ETag matching the existing one` — confirms the no-op short-circuit doesn't invent a new ETag

### DIFF
--- a/packages/realm-server/handlers/handle-publish-realm.ts
+++ b/packages/realm-server/handlers/handle-publish-realm.ts
@@ -542,6 +542,17 @@ export default function handlePublishRealm({
       await publishedRealm.fullIndex(userInitiatedPriority, {
         clearLastModified: true,
       });
+
+      // The source realm's `RealmInfo.lastPublishedAt` map is built
+      // from `realm_registry` rows joined on `source_url = sourceRealmURL`,
+      // so publishing this derivative just changed it. Without
+      // invalidating the cache, the source's `getRealmInfo()` keeps
+      // returning the pre-publish snapshot — and the card+json ETag,
+      // which folds a hash of that snapshot in, would still match a
+      // stale `If-None-Match` and serve a 304 with the old
+      // `meta.realmInfo.lastPublishedAt`. (CS-11010)
+      sourceRealm.invalidateCachedRealmInfo();
+
       let publishedPermissions = await fetchRealmPermissions(
         dbAdapter,
         new URL(publishedRealmURL),
--- a/packages/realm-server/handlers/handle-unpublish-realm.ts
+++ b/packages/realm-server/handlers/handle-unpublish-realm.ts
@@ -147,6 +147,24 @@ export default function handleUnpublishRealm({
         await removeRealmPermissions(dbAdapter, new URL(publishedRealmURL));
       });
 
+      // Removing this derivative just changed the source realm's
+      // `RealmInfo.lastPublishedAt` map (rows where `source_url =
+      // sourceRealmURL`). Without invalidating the source's cached
+      // realm info, its card+json ETag (which folds a hash of the
+      // realm info in) would keep matching pre-unpublish If-None-Match
+      // headers and serve a 304 with stale `meta.realmInfo`. (CS-11010)
+      let sourceRealmURL = publishedRealmInfo.source_realm_url;
+      if (sourceRealmURL) {
+        try {
+          let sourceRealm = await reconciler.lookupOrMount(sourceRealmURL);
+          sourceRealm?.invalidateCachedRealmInfo();
+        } catch (err) {
+          log.warn(
+            `Could not invalidate source realm cached realm-info for ${sourceRealmURL} after unpublish: ${err}`,
+          );
+        }
+      }
+
       // Permissions for the published realm were removed inside the
       // write lock above, so fetchRealmPermissions(publishedRealmURL)
       // would return nothing useful for X-Boxel-Realm-Public-Readable.
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -891,6 +891,14 @@ module(basename(__filename), function () {
             'public, max-age=0, must-revalidate',
             'world-readable realm advertises public cache-control',
           );
+          // The X-Boxel-Etag-Suppressed signal only fires when the
+          // foreign-deps guard rejects the ETag; a card with purely
+          // local deps must NOT carry it, otherwise ops dashboards
+          // can't tell normal from suppressed traffic.
+          assert.notOk(
+            response.get('X-Boxel-Etag-Suppressed'),
+            'no suppression signal on a card with only local deps',
+          );
         });
 
         test('returns 304 when If-None-Match matches the current ETag', async function (assert) {
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -883,7 +883,7 @@ module(basename(__filename), function () {
           let etag = response.get('etag') ?? '';
           assert.ok(etag, 'response carries an ETag');
           assert.true(
-            /^\d+(?:-[0-9a-f]+)?:card$/.test(etag),
+            /^"\d+(?:-[0-9a-f]+)?:card"$/.test(etag),
             `ETag matches "<indexed_at>(-<realmInfoHash>)?:card" pattern (got ${etag})`,
           );
           assert.strictEqual(
@@ -2517,7 +2517,7 @@ module(basename(__filename), function () {
           let patchEtag = patchResponse.get('etag') ?? '';
           assert.ok(patchEtag, 'PATCH response carries an ETag');
           assert.true(
-            /^\d+(?:-[0-9a-f]+)?:card$/.test(patchEtag),
+            /^"\d+(?:-[0-9a-f]+)?:card"$/.test(patchEtag),
             `PATCH ETag matches "<indexed_at>(-<realmInfoHash>)?:card" pattern (got ${patchEtag})`,
           );
           assert.notStrictEqual(
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -883,8 +883,8 @@ module(basename(__filename), function () {
           let etag = response.get('etag') ?? '';
           assert.ok(etag, 'response carries an ETag');
           assert.true(
-            /^\d+:card$/.test(etag),
-            `ETag matches "<indexed_at>:card" pattern (got ${etag})`,
+            /^\d+(?:-[0-9a-f]+)?:card$/.test(etag),
+            `ETag matches "<indexed_at>(-<realmInfoHash>)?:card" pattern (got ${etag})`,
           );
           assert.strictEqual(
             response.get('cache-control'),
@@ -2517,8 +2517,8 @@ module(basename(__filename), function () {
           let patchEtag = patchResponse.get('etag') ?? '';
           assert.ok(patchEtag, 'PATCH response carries an ETag');
           assert.true(
-            /^\d+:card$/.test(patchEtag),
-            `PATCH ETag matches "<indexed_at>:card" pattern (got ${patchEtag})`,
+            /^\d+(?:-[0-9a-f]+)?:card$/.test(patchEtag),
+            `PATCH ETag matches "<indexed_at>(-<realmInfoHash>)?:card" pattern (got ${patchEtag})`,
           );
           assert.notStrictEqual(
             patchEtag,
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -873,6 +873,73 @@ module(basename(__filename), function () {
             'deeply nested linksToMany returns first match',
           );
         });
+
+        test('returns an ETag and public cache-control on a 200 response', async function (assert) {
+          let response = await request
+            .get('/person-1')
+            .set('Accept', 'application/vnd.card+json');
+
+          assert.strictEqual(response.status, 200, 'HTTP 200 status');
+          let etag = response.get('etag') ?? '';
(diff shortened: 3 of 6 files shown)


habdelra:
@codex you're right — pushed [`0ba0b0f`](https://github.com/cardstack/boxel/pull/4683/commits/0ba0b0fedc) which folds an md5 of the cached `getRealmInfo()` into the ETag base, so the format is now `<indexed_at>-<realmInfoHash>:card`.

Mechanics: `getRealmInfo()` populates a sibling `#cachedRealmInfoHash` whenever it (re)populates `#cachedRealmInfo`, and a tiny `invalidateCachedRealmInfo()` helper clears both fields atomically — replacing every `#cachedRealmInfo = null` site (`fullIndex`, the `.realm.json` file-watcher path, and `patchRealmConfig`) so the hash can never go out of sync with the cached info. The card+json handlers `await this.getRealmInfo()` up front to prime the cache before computing the ETag.

Added a regression test exactly for the case you called out: capture pre-PATCH ETag → `PATCH /_config` with a new `name` → re-`GET` with `If-None-Match: <old etag>` must return 200 with the new `meta.realmInfo.name`, not 304.

On the other half of your concern — cross-realm linked resources loaded at request time — I think that case is already covered, but checking my reasoning: `attachRealmInfo()` only stamps THIS realm's info on resources whose `meta.realmURL` matches this realm; it doesn't touch foreign-realm `included[]` entries, so foreign-realm `_config` changes don't mutate our response body. And instance-level changes on a foreign linked card propagate to this card's `indexed_at` via the deps graph (relationship URLs are extracted as deps in `relationship-dependency-extractor`, and the invalidation cascade bumps this row's `indexed_at` when any transitively-included row changes). If you spot a case that path doesn't cover I'd genuinely like to know.

habdelra:
Good catch — pushed [`c846c86`](https://github.com/cardstack/boxel/pull/4683/commits/c846c86569). Confirmed the cross-realm dep gap by tracing `index-writer.ts:1042-1044`:

```sql
WHERE deps @> [...]
  AND i.realm_url = $thisRealm   -- "probably need to reevaluate this condition when we get to cross realm invalidation"
```

So a write to a foreign-realm card never bumps the consuming realm's `indexed_at`, while `loadLinks` HTTP-fetches the foreign card fresh on every request — which is exactly the mismatch you flagged.

Implementation: `hasForeignRealmDeps(deps, this.url)` walks the index entry's `deps` array and returns true if any entry is a foreign card-instance URL. The check explicitly ignores module-graph deps:
- non-`http(s)://` strings (registered prefixes, relative paths)
- `https://cardstack.com/base/...`

If left as-is those would mean every card has a "foreign" dep (every card adopts from `card-api`) and the guard would suppress ETags universally — defeating the optimization for the exact cards the parent ticket cares about. Only foreign card *instance* URLs trip the guard, and the ones in the deps array come from relationship URL extraction — that's the case where cross-realm invalidation actually leaves us holding a stale `included[]`.

`SearchResultDoc` now also carries `deps: string[] | null` so the PATCH no-op + post-write paths can run the same guard without an extra `instance()` query. Both 304 short-circuit AND the response-header 200/PATCH paths now respect `hasForeignDeps`.

I didn't add the integration test you described in this commit — a multi-realm fixture is heavy enough that I'd rather get the guard reviewed first and add the test separately if reviewers want it. Happy to wire one through `runTestRealmServerWithRealms` if requested. The guard logic is small enough to read directly: `realm.ts` `hasForeignRealmDeps` + `isForeignRealmDep` helpers, ~30 lines.

Followup work flagged: when cross-realm dep cascading actually lands (the comment in `index-writer.ts.calculateInvalidations` is the marker), this guard can be removed.

habdelra:
Pushed [`fd83351`](https://github.com/cardstack/boxel/pull/4683/commits/fd83351278) addressing review findings from a fresh independent review I ran on the branch plus what's landed from @codex and Copilot. Summary of changes since `c846c86`:

**Correctness**
- **Foreign-dep classification was broken in production.** `dep.startsWith('http')` early-returned for prefix-form deps like `@cardstack/foreign-realm/foo.json`, which is exactly how `addRealmMapping` (used everywhere in prod) stores them. Guard silently failed for the cards it was supposed to protect. New `isForeignRealmDep` resolves the prefix first, then trips on `.json` instance deps that resolve outside this realm. Module/scoped-CSS deps continue to pass.
- **Source realm's `#cachedRealmInfo` is now invalidated on publish/unpublish.** `RealmInfo.lastPublishedAt` is built from `realm_registry` rows joined on `source_url`, so (un)publishing a derivative bumps it without otherwise touching the source. Without the hook, the source's card+json ETag would have 304'd against pre-publish validators forever.
- **`querySourceRealmPublications` now `ORDER BY url`.** Without it, `JSON.stringify` of `Object.fromEntries(rows)` was non-deterministic across instances/restarts — every cache-prime race would have flipped the realm-info hash and invalidated every browser cache.

**RFC compliance**
- ETag values are now quoted per RFC 9110 §8.8.3 (`"<indexed_at>-<realmInfoHash>:card"`).
- `If-None-Match` parsing now handles `*`, comma-separated lists, and weak (`W/`) validators (using the same matcher pattern as `server.ts` for the index-HTML route).

**Race fix**
- The response-side `hasForeignDeps` flag is now derived from `cardDocument()`'s result, not from the early peek. Closes a window where a write between the peek and the assembly could emit an ETag tied to a snapshot whose deps were already foreign.

**Layout**
- The `.json` 302-redirect now happens before any cache check, so the canonical URL is the only one that ever serves an ETag — intermediaries can't end up caching a 304 bound to the `.json` form.
- The conditional-GET fast path's `instance()` peek now only runs when `If-None-Match` is present. On cache-miss requests we go straight to `cardDocument()` (which does its own instance lookup), saving the redundant round-trip @copilot called out.

**Comments**
- `CARD_JSON_ETAG_VARIANT` now documents both inputs to the base.
- `cardJsonCacheControl` now acknowledges `must-revalidate` is added *on top of* the source/module pattern, not the same.

Tests updated to match the now-quoted format. The earlier CI red on `Realm Server Tests (2, 6)` was the foreign-dep classifier marking base-realm module deps as foreign in the test environment (where base maps to `localhost:4201`, not `cardstack.com/base/`); the new `.json`-suffix discriminator should fix that.

Still open: integration test for cross-realm dep ETag-suppression. Happy to add via `runTestRealmServerWithRealms` if reviewers want it before merge — flagging now in case there are nicer existing fixtures I should reuse.

lukemelia (approved):
I'm not crazy about quietly skipping etags when there is a cross realm dependency. It seems like it could lead to unexpectedly uneven performance experiences. I guess we can fix it when we add cross-realm invalidation?

habdelra:
Thanks for the review @lukemelia. Good call on the unevenness concern — I agree it's a real foot-gun: two cards in the same realm could have very different perf profiles (one 304s, the other re-assembles every time) and that surface area only widens as more realms link cross-realm. The suppression is meant as a *correctness* floor while we still have the realm-bounded `calculateInvalidations` (`packages/runtime-common/index-writer.ts:1042-1044`) — without it, foreign-card writes serve stale 304s. Removing the guard the moment cross-realm invalidation lands is the right plan.

Two ways I'd consider making it less *quiet* in the meantime, if it's worth a small follow-up:

1. **Surface it on the response** — emit `X-Boxel-Etag-Suppressed: foreign-deps` on the card+json response when the guard fires. Doesn't change client behavior, but ops/Grafana can see the rate at which cards are being penalized and prioritize cross-realm invalidation work accordingly.
2. **Log at debug** — fine-grained logger entry per request that hits the guard. Cheaper but only useful interactively.

Happy to do (1) as a tiny follow-up to this PR if you'd like — it's ~5 lines and gives us a measurable signal that drives the case for cross-realm invalidation. Or fold it into the cross-realm-invalidation work and remove both at the same time.

habdelra:
Done — pushed [`dacb2c8`](https://github.com/cardstack/boxel/pull/4683/commits/dacb2c8788). Card+json responses whose ETag the foreign-deps guard suppresses now carry `X-Boxel-Etag-Suppressed: foreign-deps`. Doesn't change client behavior, but `count by (status_code) (rate({app="realm-server"} | json | __X_Boxel_Etag_Suppressed__ = "foreign-deps"))` (or equivalent at the response-header layer) will give us the rate the guard fires per realm, so it's measurable evidence for prioritizing cross-realm invalidation. Both the suppression and the header come out the moment that lands.

Applied to all three response sites that participate in the guard (PATCH no-op, PATCH post-write, GET 200) plus a regression assertion that the header is *absent* on a card with only local deps so dashboards can tell normal from suppressed traffic.

PR #4704, opened 2026-05-07
mise/dev-all: kill reparented service processes on cleanup

lukemelia:
## Summary

`mise dev-all` was leaving service processes running after Ctrl-C / exit, most visibly the worker-manager holding port 4210 — which then blocks the next `mise dev-all` from starting.

Root cause: mise's task supervisor reparents long-running task scripts (`mise-tasks/services/*` and the `ts-node --transpileOnly worker` child) to init, so the existing PPID-based `kill_tree` can't reach them from dev-all's trap.

Fix: a `pkill` sweep scoped to absolute paths under this checkout's `$REPO_ROOT` so other worktrees aren't disturbed. SIGTERM first, 2s grace window, then SIGKILL for anything that didn't respond.

Patterns matched:
- `$REPO_ROOT/mise-tasks/services/` (realm-server, worker, prerender, prerender-mgr, etc.)
- `$REPO_ROOT/packages/realm-server/node_modules.*--transpileOnly worker`

## ⚠️ Cleanup before merge ✅ 

The PR also adds `tee -a /tmp/dev-all-cleanup.log` lines around each step so the cleanup sequence is observable. These are **WIP debugging output** — should be stripped before merge unless we want a permanent log file. Specifically, the `[dev-all cleanup] ...` echo statements and the `pgrep -lf ...` enumeration lines.

## Test plan

- [ ] `mise dev-all` → wait for stack to come up
- [ ] Ctrl-C → verify in another shell that no `mise-tasks/services/*` processes remain (`pgrep -lf "$REPO_ROOT/mise-tasks/services/"` returns nothing)
- [ ] Verify port 4210 is free: `lsof -nP -i :4210` returns nothing
- [ ] `mise dev-all` again — should start cleanly without "address already in use"
- [ ] Verify other worktrees' running services are untouched (the `$REPO_ROOT`-scoped patterns shouldn't match them)

### DIFF
--- a/mise-tasks/dev-all
+++ b/mise-tasks/dev-all
@@ -6,7 +6,10 @@
 # MISE dir is packages/realm-server, so ../.. is repo root
 PATH="$(pwd)/../../node_modules/.bin:$(pwd)/node_modules/.bin:$PATH"
 
-. "$(cd "$(dirname "$0")" && pwd)/lib/dev-common.sh"
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+. "$SCRIPT_DIR/lib/dev-common.sh"
 
 # Enable job control so backgrounded subprocesses run in their own process
 # group. Without this, Ctrl-C is delivered to the whole foreground group;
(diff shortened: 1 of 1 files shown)


lukemelia:
@backspace I'm not sure if this is the best approach to this problem. but it was driving me nuts.

lukemelia:
@codex review this

backspace:
I opened 11084 about this and had Claude looking at it independently, because ^C on the task was leaving a Node process with runaway CPU.

lukemelia:
> I opened 11084 about this and had Claude looking at it independently, because ^C on the task was leaving a Node process with runaway CPU.

That explains my lack of battery life yesterday.

backspace (approved):
I ran this locally and confirmed that it fixed the dangling process I was looking at in 11084. Claude found a couple of other smaller problems as part of that that I’ll include in a followup PR.

backspace:
> > I opened 11084 about this and had Claude looking at it independently, because ^C on the task was leaving a Node process with runaway CPU.
> 
> That explains my lack of battery life yesterday.

yes it was dramatic for me because environment mode means there’s no port conflict, so by the end of my workday I noticed four dangling processes 🙃

PR #4707, opened 2026-05-07
Fix SF /_atomic mock for ?waitForIndex=true; rename test fixture username

habdelra:
## Summary

- **Fix `204 No Content` failure on shard 1/3 of SF Node tests.** The `?waitForIndex=true` opt-in landed in 9444b82262 makes the factory entrypoint append a query string to the `/_atomic` POST URL when syncing the seed batch. The integration mock checked `request.url === '/testuser/personal/_atomic'` (exact match), so the new URL fell through to the generic POST handler and got `204 No Content`. The sync client expects `201` → reported `Atomic upload failed: 204 No Content`, factory:go exited 1, and `pnpm test:node` failed at `factory-entrypoint integration > factory:go --debug prints a structured JSON summary`. Switched the handler to `startsWith('/testuser/personal/_atomic')`, consistent with the other handlers already in this mock.
- **Rename test fixture username from `hassan` to `testuser`** across 5 SF test files (34 occurrences). The fixture had a real teammate's name baked into Matrix user IDs, account-data paths, and realm URLs; `testuser` is neutral.
- This was failing every shard 1 run on `main` since the merge — not flake. Failure example: https://github.com/cardstack/boxel/actions/runs/25493597169/job/74808011882

## Test plan

- [x] `pnpm test:node` in `packages/software-factory` passes locally (440 pass, 0 fail) before and after the rename
- [ ] Software Factory CI shard 1/3 turns green

### DIFF
--- a/packages/software-factory/tests/factory-agent-claude-code.test.ts
+++ b/packages/software-factory/tests/factory-agent-claude-code.test.ts
@@ -49,7 +49,7 @@ function makeContext(): AgentContext {
     },
     knowledge: [],
     skills: [],
-    targetRealm: 'https://realms.example.test/hassan/personal/',
+    targetRealm: 'https://realms.example.test/testuser/personal/',
     // System-prompt rendering requires this — see requireDarkfactoryModuleUrl.
     darkfactoryModuleUrl:
       'https://realms.example.test/software-factory/darkfactory',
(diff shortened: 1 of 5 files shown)


jurgenwerk (approved):
Thanks!

PR #4715, opened 2026-05-07
Fix flaky matrix isolated realm startup timeout

habdelra:
## Summary
- wait for the isolated matrix test harness worker manager to be actually ready before starting the realm server
- make worker startup timeout tunable for the matrix test harness and use a less aggressive default there
- replace the previous opaque realm-server startup timeout with structured diagnostics that capture child-process state, metadata, and recent startup logs

## Context
This PR is aimed at the flaky failure seen in CI here:
https://github.com/cardstack/boxel/actions/runs/25498226454/job/74824197030

The actionable failure from that run was in `Matrix Client Tests (3, 3)`:
- `timed-out waiting for realm server to start. Stopping server`
- thrown from `packages/matrix/helpers/isolated-realm-server.ts`
- surfaced while running `pnpm playwright test --shard 3/3`

That CI run also contained unrelated `Boxel CLI Tests` failures. This PR is only addressing the matrix shard startup flake.

## Failure Mode
The matrix Playwright suite brings up an isolated support stack in `global.setup.ts`:
1. start prerender
2. start worker manager
3. start isolated realm server
4. wait for the realm server's IPC `ready` message

Before this change, the helper launched the worker manager and then immediately launched the realm server. If the worker manager was still in the middle of starting its child worker, the realm server could stall in its own startup path while waiting for the worker manager to become ready. When that happened, the harness would sit on a long timeout and eventually throw a very generic startup error.

That made the flake hard to distinguish between:
- worker manager exited early
- worker manager HTTP came up but workers were not ready yet
- realm server exited before sending `ready`
- realm server was still alive but stalled in startup

## Root Cause Hypothesis
The working hypothesis is that the matrix helper had a race between "worker manager process spawned" and "worker manager actually ready to serve the realm server startup dependency".

The existing harness treated those as equivalent, but they are not equivalent:
- the worker manager's HTTP server can be listening before its worker is actually ready
- the realm server startup path explicitly waits on worker-manager readiness before it can finish startup and emit `ready`
- under CI load, that worker startup can occasionally be slow enough to trip the harness into an opaque timeout path

### `packages/matrix/helpers/isolated-realm-server.ts`
The matrix isolated-realm helper now:
- waits for the worker manager readiness endpoint before spawning the realm server
- watches for worker-manager early exit and fails fast if it dies before becoming ready
- watches for realm-server early exit and fails fast if it dies before emitting `ready`
- uses a bounded realm-server startup timeout instead of the previous very long opaque wait
- stops both child processes when startup fails, instead of leaving failure diagnosis to the outer timeout path
- captures startup diagnostics into the thrown error, including:
  - worker-manager child state
  - realm-server child state
  - runtime metadata files written by the worker manager and realm server
  - tail of recent worker-manager stdout/stderr
  - tail of recent realm-server stdout/stderr
  - timeout values, realm path, DB name, and worker-manager port

### `packages/realm-server/worker-manager.ts`
The worker manager now:
- reads `TEST_HARNESS_WORKER_START_TIMEOUT_MS`
- uses that value instead of a hard-coded 30 second worker-start timeout
- includes the effective timeout in the timeout error message

## Why This Should Reduce the Flake
This change removes the helper's assumption that the worker manager is usable immediately after spawn.

Instead of racing the realm server against worker-manager warmup, the helper now gates startup on the worker manager's readiness check. That should absorb the CI-only slow-start case where the worker process is simply late rather than actually broken.

If the underlying problem is something else and the flake persists, the new diagnostics should make the next failure actionable by showing whether the worker manager or realm server exited, whether either process reached its metadata-writing phase, and what each process last logged before failure.

## Diagnostics Added For Future Failures
If startup fails after this PR, the error message should now include structured JSON with:
- `workerManagerState`
- `realmServerState`
- `workerManagerMetadata`
- `realmServerMetadata`
- `workerManagerOutputTail`
- `realmServerOutputTail`
- timeout settings and harness context

That should let us answer questions that the old failure could not answer, such as:
- Did the worker manager ever bind its HTTP port?
- Did the realm server ever reach its HTTP listener setup?
- Did the worker manager die because its child worker never came up?
- Did the realm server remain alive but stall before IPC `ready`?

## Validation
I validated the touched slice locally as follows:
- editor diagnostics were clean for the modified files
- `pnpm --dir packages/matrix lint` completed earlier in the main checkout during the investigation
- `pnpm --dir packages/realm-server lint` completed earlier in the main checkout during the investigation
- focused module-load validation succeeded for the matrix helper in the main checkout toolchain during development
- focused ESLint validation on the touched realm-server file completed without file-level errors

I did **not** run the full matrix Playwright shard locally. The worktree setup available here did not have a clean fully-linked dependency/tooling environment for a faithful end-to-end rerun from the new worktree itself, so I kept validation focused on the touched slice and the changed files.

## Risk / Tradeoffs
- The matrix helper is now stricter about surfacing early child-process exits. That is intentional: an early exit is a more actionable failure than a later generic timeout.
- The worker-manager startup timeout is now configurable for test-harness use, but still defaults to the previous behavior when the new env var is not set.
- The diagnostics make failure messages larger, but only on startup failure paths where that extra context is useful.

## Follow-up
If CI still flakes after this lands, the next failure should tell us whether the remaining issue is:
- slow worker startup beyond the configured timeout,
- worker-manager process failure,
- realm-server process failure, or
- a later startup stall after both processes are alive.

That should be enough to decide whether the next fix belongs in worker startup, realm-server startup, or the matrix harness itself.


backspace:
I guess this can be closed, since #4717 supersedes it?

habdelra:
[Claude Code] Confirmed — closing as superseded by #4717, which has already merged. #4717 carried the same matrix flake fix on a clean branch off `main`; this branch was based on `cs-11082-bench-realm-gate` and contained ~70 unrelated files.

PR #4719, opened 2026-05-07
Add fix for prerenderer cache-clearing on publish

backspace:
Publishing a realm has been inconsistent in whether the changes actually show up, which has indeed been about caching, but across multiple layers. I’m out of my depth here, but here’s Claude’s explanation (edited after PR feedback about unnecessary work):

### Bug

  After republish, the published URL kept serving pre-republish HTML. In production:
  `nyuitp2026.boxel.site` served the old wordmark for ~37h after the user pushed the new img form.

### Root cause
  The `Realm` class holds an in-memory `#sourceCache` keyed by path. After a republish:

  1. Publish handler does the FS swap and enqueues a reindex.
  2. The index worker fetches source over HTTP into the realm-server.
  3. `getSourceOrRedirect` returns pre-swap bytes from `#sourceCache` (plus a stale content-hash etag).
  4. The worker writes stale HTML into `boxel_index.isolated_html`.

  The design assumed the `NodeAdapter` file-watcher would invalidate `#sourceCache`, but
  `ENABLE_FILE_WATCHER` is unset in staging/production, so it never fires.

### Fix
  One commit. New `Realm.clearLocalCaches()` method; `handle-publish-realm` calls it after the FS swap,
   before the reindex enqueue. Makes the invalidation synchronous — no race, no file-watcher
  dependency.


habdelra:
are there tests that we can add that demonstrate these new capabilities--like the prerenderer's new dispose affinity endpoint?

packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -535,6 +536,41 @@ export default function handlePublishRealm({
...
+                affinityValue: publishedRealmURL,
+              });
+            } catch (e) {
+              log.warn(
+                `disposeAffinity failed for ${publishedRealmURL}: ${
+                  e instanceof Error ? e.message : String(e)
+                } — continuing with publish; stale Loader cache may persist until LRU rotation`,
+              );
+            }
+          }
```
habdelra: this seems like it might be overkill. there is a `clearCache: true` option that the /render route for the prerenderer that was created specifically for this purpose: to clear the loader cache for the affinity. this is how we can use the same tab to handle both code changes and instance changes and always use the most recent code. When we do this we destroy the browser context for the realm in teh prerenderer which is bascially like throwing out the baby with the bath water. I'd be more interested to know why this option is not working for you

habdelra: also, hopefully the etag should be working for the module updates. if the etag wasn't working i could see how disposing of the affinity would yield a working result, as that would destroy the browser context and thus force the browser to refetch the modules even if the etag didn't change.

backspace: thanks for looking at this, Claude agreed with your assessment, we went through several iterations of trying to diagnose the problem, and that did end up not being needed, as I’ve deployed this branch since removing it and publishing still worked.

There are now realm server tests for the new behaviours, too. Can you look again?

packages/runtime-common/realm.ts
```diff
@@ -1339,6 +1339,23 @@ export class Realm {
...
+  // async-event race against the immediately-enqueued reindex; this
+  // method makes the invalidation synchronous from the publish
+  // handler's vantage point. Different from `__testOnlyClearCaches`
+  // in that it does NOT reset the transpile counter (which is
+  // test-only diagnostic state, unrelated to byte-correctness).
+  clearLocalCaches(): void {
+    this.#sourceCache.clear();
+    this.#dropAllModuleCacheEntries();
+  }
+
```
habdelra, line 1358: you might want to coordinate with @lukemelia here. this touches on the work to move caching out of memory in the realm server in preparation for horizontally scaled realms. if we had 2 realm servers that needed cache clearing i'm not sure how you would coordinate this.

habdelra (approved):

packages/runtime-common/realm.ts
backspace: thanks, we talked it over and came up with an approach, I’m going to merge this and implement a multi-realm server fix in CS-11153.

PR #4650, opened 2026-05-04
Add issue tracker cards and project-backed kanban board to software-factory

burieberry:
- Introduce generic kanbanBoard card and no-frills version of issue-tracker, currently in software-factory
- Added sample cards in experiments-realm for demo

## Summary
- split up card definitions for issue tracker and related cards: issue, comment, knowledge article, agent-profile 
- add generic kanban-board card with customizable columns and polymorphic cards
- extend kanban-board for IssueTrackerBoard with columns and enum options related to issue-tracker statuses
- use the new boxel-ui drag-and-drop kanban board in the project-backed issue tracker board
- update the software-factory realm wiring and dependencies to expose the new issue tracker experience
- add board creation into software-factory operations and tests
- test coverage and updates

Next Steps:
-  Move issue-tracker out of software-factory
-  More test coverage on functionality
-  UI polish

PR #4735, opened 2026-05-08
Add live tests for issue tracker cards

burieberry:
This PR builds on #4650

- Add live-card tests for software-factory issue-tracker cards
- Clean up of experiments files
- Add missing openCard action

---
Note: The file to review for this PR: `packages/software-factory/realms/issue-tracker.test.gts`
Everything else is from https://github.com/cardstack/boxel/pull/4650 that's awaiting approval


habdelra:
so these tests don't actually exercise the DOM of the issue tracker cards, shouldn't we be doing that in the tests? We can add a new playwright spec tests in software factory that do that

burieberry:
> so these tests don't actually exercise the DOM of the issue tracker cards, shouldn't we be doing that in the tests? We can add a new playwright spec tests in software factory that do that

This file is this for this PR specifically: realms/issue-tracker.test.gts

Everything else is from #4650 that's awaiting approval

habdelra:
> > so these tests don't actually exercise the DOM of the issue tracker cards, shouldn't we be doing that in the tests? We can add a new playwright spec tests in software factory that do that
> 
> This file is this for this PR specifically: realms/issue-tracker.test.gts
> 
> Everything else is from #4650 that's awaiting approval

one thing I i do that that more clear is used stacked PR, swhere the merge is into the parent PR--this it's more clear to see what specifically is different in the child PR. then right before you actually go to do the merge, you can use github UI to change the parent back to main.

burieberry:
> > > so these tests don't actually exercise the DOM of the issue tracker cards, shouldn't we be doing that in the tests? We can add a new playwright spec tests in software factory that do that
> > 
> > 
> > This file is this for this PR specifically: realms/issue-tracker.test.gts
> > Everything else is from #4650 that's awaiting approval
> 
> one thing I i do that that more clear is used stacked PR, swhere the merge is into the parent PR--this it's more clear to see what specifically is different in the child PR. then right before you actually go to do the merge, you can use github UI to change the parent back to main.

Good idea, makes sense. Thanks

packages/software-factory/realm/issue-tracker.test.gts
```diff
@@ -0,0 +1,372 @@
...
+
+import { setupApplicationTest } from '@cardstack/host/tests/helpers/setup';
+import {
+  setupLocalIndexing,
+  setupOnSave,
+  testRealmURL,
+  setupAcceptanceTestRealm,
+  SYSTEM_CARD_FIXTURE_CONTENTS,
+  visitOperatorMode,
+} from '@cardstack/host/tests/helpers';
```
habdelra, line 18: This is the first instance of using the test realm in the software factory. up to this point we've been testing with an _actual_ realm with a full boxel stack (via the test-harness). so this isn't really as faithful in terms of a real realm as the playwright specs are. but i think it's ok, as you'll be able to use ember test helpers in this test which are much more powerful and concise for DOM testing than playwright. but probably we need to be clear around the rules for when we want to actually use a real live complete boxel environment for testing vs uses an in-browser realm like this tests does. if you want to spell out the rules for when to use in-browser realms vs real node realms that would be cool. 

also i noticed that we don't have @cardstack/host in our package.json's devDependencies. probably we need that. i think it happens to be transitively included right now which is why its working, but we should actually spell it out in the package.json

burieberry: Added missing dependencies, updated readme about tests, and resolved relevant copilot catches

habdelra (approved):

PR #4745, opened 2026-05-09
Grafana: wrap Operator Action button customCode in async IIFE

lukemelia:
## Summary
The Business Forms plugin (\`volkovlabs-form-panel\`) compiles a Custom Button's \`customCode\` via \`new Function("context", body)\` — a regular, non-async function. The plugin source confirms it:

\`\`\`js
_e=(...e)=>{try{return new Function(...e)}catch(e){return console.error("Code Error",e),()=>{}}};
\`\`\`

Every Operator Action button on Indexing / Realms / Users currently has a top-level \`await fetch(...)\` in its customCode, which throws \`SyntaxError: await is only valid in async function\` at compile time. The plugin catches that, logs \`Code Error\` to the console, and returns a no-op — so the button silently does nothing. No network request fires. (Reproduced live on local Grafana with Chrome DevTools open.)

This wraps each affected button's \`try { ... await fetch ... } catch (err) {...}\` in an \`(async () => { ... })()\` IIFE. Outer closure stays sync; inner IIFE provides the async context that legalises \`await\`.

## Affected buttons
- \`indexing.json\` — Reindex ALL realms (\`btn_reindex_all\`)
- \`realms.json\` — Reindex this realm (\`btn_reindex_realm\`)
- \`users.json\` — Add 1k / 10k / 100k credits, Add custom amount (4 buttons)

The Grant Permission button on the Realms dashboard (\`btn_grant\`, added in #4744) has the same bug; it's fixed in the same PR alongside the panel introduction.

## Why this wasn't caught earlier
The bug landed in dd94c93210 when these panels were converted from Grafana-native action buttons to Business Forms. The conversion was visually verified (panel renders, queries run, blast-radius numbers display) but the button-click → fetch path was never exercised end-to-end. Staging compounded this by missing the plugin install entirely (fixed by [cardstack/infra#689](https://github.com/cardstack/infra/pull/689)), so the not-found error masked the await failure.

## Test plan
- [ ] Apply locally; click "Reindex this realm" on Realms dashboard; confirm dialog → POST fires (visible in DevTools Network) → notify success/error
- [ ] Same for Reindex ALL realms on Indexing dashboard
- [ ] Same for Add 1k / 10k / 100k / custom credit buttons on Users dashboard
- [ ] Browser console no longer shows \`Code Error\` after click


habdelra:
shouldn't we have a .catch() on that iife so that we dont swallow async errors?

lukemelia:
Good catch — pushed [ac4176e304](https://github.com/cardstack/boxel/commit/ac4176e304) attaching a defensive `.catch` to all 6 IIFEs (reindex-all, reindex-realm, +1k/+10k/+100k/custom credit). The catch `console.error`s and moves on rather than calling `notifyError` again — if `notifyError` is what tripped the inner catch in the first place, re-entering the same broken path defeats the safety net.

PR #4746, opened 2026-05-09
realm-server: add /_grafana-upsert-realm-user-permission endpoint

lukemelia:
## Summary
The Realms-dashboard "Grant Permission" Operator Action panel POSTs to \`/_grafana-upsert-realm-user-permission?realm=...&user=...&read=...&write=...\` with the same \`Authorization: Bearer \${grafana_secret}\` header used by the other grafana endpoints. The realm-server side was deliberately deferred (README phase-3.5 note) and never landed — the button currently authenticates fine and gets a clean 404 because the route doesn't exist.

This adds the missing handler + route registration. Same auth pattern (\`grafanaAuthorization\`), same dual GET/POST registration via \`registerGrafanaEndpoint\` for the CS-10987 cutover window, same upsert path the realm-server's own \`/_realm-permissions\` endpoints use (\`insertPermissions\` from runtime-common).

### Behavior
- Required query params: \`realm\` (URL), \`user\` (Matrix user id).
- \`read\` / \`write\` are \`'true'\` / \`'false'\` flags from the dashboard.
- Read-only and read+write are accepted.
- Write-without-read is rejected (write requires read).
- Both-false is rejected (use the existing delete flow to revoke).

## Depends on
- The dashboard side (#4744 Realms Grant Permission as Business Forms) needs to land alongside this for the operator workflow to be complete; either order works for the merge.
- Local-dev: restart the realm-server to pick up the new route.

## Test plan
- [ ] Test suite added in \`maintenance-endpoints-test.ts\` (read-only grant, read+write grant, idempotent re-grant, write-without-read rejection, both-false rejection, missing realm, non-URL realm, missing auth)
- [ ] Manual: open Realms dashboard, pick a user from the Matrix-username dropdown, choose a permission level, click Grant — confirm dialog → 200 → notify success → refresh shows the new row in the Realm Permissions table

### DIFF
--- a/packages/realm-server/handlers/handle-upsert-realm-user-permission.ts
+++ b/packages/realm-server/handlers/handle-upsert-realm-user-permission.ts
@@ -1,0 +1,71 @@
+import type Koa from 'koa';
+import {
+  insertPermissions,
+  type RealmAction,
+  SupportedMimeType,
+} from '@cardstack/runtime-common';
+import { sendResponseForBadRequest, setContextResponse } from '../middleware';
+import type { CreateRoutesArgs } from '../routes';
+
+export default function handleUpsertRealmUserPermission({
+  dbAdapter,
+}: CreateRoutesArgs): (ctxt: Koa.Context, next: Koa.Next) => Promise<void> {
+  return async function (ctxt: Koa.Context, _next: Koa.Next) {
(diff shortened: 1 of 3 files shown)


habdelra (approved):

habdelra (approved):
don't forget to fix lint

PR #4766, opened 2026-05-11
Add TEST_FILES env var to realm-server tests/index.ts

lukemelia:
## Summary

- Adds a `TEST_FILES` env var to `packages/realm-server/tests/index.ts` that limits which test files are parsed and executed. Comma-separated list of paths relative to `tests/`, with or without `./` prefix or `.ts` suffix.
- Default behavior is unchanged: when `TEST_FILES` is unset, every file in `ALL_TEST_FILES` is loaded — the same 102 files that were previously imported at the top of the file.
- The new array is `diff`-confirmed equal to the prior import list path-for-path.

## Why

[CS-10009](https://linear.app/cardstack/issue/CS-10009/consolidate-realm-server-test-realms) will migrate ~40 test files between three new realm fixtures (blank / simple / realistic). Per-file before/after measurement needs a clean signal — `TEST_MODULES` already filters which modules *run*, but every file still gets parsed and required, so peak-RSS deltas get masked by the all-files startup baseline. `TEST_FILES` gives a cleanly attributable per-file wall time and memory footprint, which we want in `main` first so subsequent PRs can produce comparable measurements against the same baseline.

## Usage

```
TEST_FILES=sanitize-head-html-test pnpm test
TEST_FILES=realm-endpoints/invalidate-urls-test,server-endpoints/queue-status-test pnpm test
```

## Test plan

- [x] `TEST_FILES=sanitize-head-html-test pnpm test` → only the 27 sanitize-head-html tests load and run; output contains no references to any other test file.
- [x] Default path (`TEST_FILES` unset, `TEST_MODULES=sanitize-head-html-test.ts`) → all 102 files still load; filter applies as before; 27 tests pass.
- [x] Path list is byte-for-byte equal to the prior `import './…';` block (verified with `diff`).

### DIFF
--- a/packages/realm-server/tests/index.ts
+++ b/packages/realm-server/tests/index.ts
@@ -279,7 +279,6 @@ if (testFilesEnv) {
 }
 
 for (const file of filesToLoad) {
-  // eslint-disable-next-line @typescript-eslint/no-var-requires
   require(file);
 }
 
--- a/packages/realm-server/tests/index.ts
+++ b/packages/realm-server/tests/index.ts
@@ -156,108 +156,142 @@ QUnit.done(() => {
 
 import 'decorator-transforms/globals';
 import '../setup-logger'; // This should be first
-import './atomic-endpoints-test';
-import './auth-client-test';
-import './billing-test';
-import './card-dependencies-endpoint-test';
-import './card-endpoints-test';
-import './card-source-endpoints-test';
-import './definition-lookup-test';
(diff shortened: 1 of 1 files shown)


habdelra:
can we introduce a description for this in the README or a skill so claude code knows how to use it

backspace (approved):
I agree with Hassan’s comment but otherwise 👍🏻

PR #4771, opened 2026-05-11
Remove transitional command/parameter rename code

backspace:
#4627 renamed commands and their parameters to not assume realm and resource URLs. #4730 restored the old imports for the transitional period before the [`boxel-skills` PR](https://github.com/cardstack/boxel-skills/pull/72) was deployed in staging and production. Now that everywhere has the correct imports and parameter names, we don’t need these shims and rewrite scripts.


packages/host/memory-baseline.json
```diff
@@ -330,9 +321,6 @@
     "Integration | commands | invalidate-realm-identifiers": {
       "delta_mb": -16
     },
-    "Integration | commands | invalidate-realm-urls": {
-      "delta_mb": -9.9
-    },
```
backspace, line 335: Does this make sense to include in this PR, @lukemelia? I imagine renamed/deleted modules might linger forever.

habdelra (approved):

PR #4790, opened 2026-05-12
CS-10009 PR 2: migrate realm-endpoints/ tests to explicit fixture

lukemelia:
Second slice of [CS-10009](https://linear.app/cardstack/issue/CS-10009/consolidate-realm-server-test-realms). Stack: PR 1 (#4772) landed the fixture plumbing + 2 representative migrations; this PR migrates the remaining 9 callers under `packages/realm-server/tests/realm-endpoints/`.

## Per-file decisions

| File | Fixture | Rationale |
|---|---|---|
| `cancel-indexing-job-test.ts` | `blank` | Job-queue ops only — no card content |
| `dependencies-test.ts` | `blank` | Writes its own `dependencies-card.gts` |
| `directory-test.ts` | **`realistic`** | Asserts on `dir/bar.txt`, `dir/foo.txt`, `dir/subdir/` from `tests/cards/dir/` |
| `info-test.ts` (4 modules) | `blank` | `_info` returns `testRealmInfo`-shape; doesn't read cards |
| `lint-test.ts` | `blank` | POST source bodies, no realm reads |
| `mtimes-test.ts` | `blank` | `mtimes(testRealmPath)` round-trip; works on any fixture |
| `permissions-test.ts` | `blank` (was `fileSystem: {}`) | `_permissions` endpoints |
| `search-test.ts` (4 modules) | `simple` | `_search` queries for `Person`/`Mango` |
| `user-test.ts` (2 modules) | `blank` | `_user` endpoints don't read card content |

Skipped (per Linear plan: callers that pass their own `fileSystem` inline are left alone):
- `invalidate-urls-test.ts` — already migrated to `simple` in PR 1.
- `markdown-test.ts`, `publishability-test.ts`, `reindex-test.ts` — each test slot defines its content inline.

## Drive-by: dead `copySync(cards)` blocks

Four files (`directory-test`, `info-test`, `permissions-test`, `user-test`) had a `hooks.beforeEach(() => { dir = dirSync(); copySync(cards, dir.name); })` block whose result was never read — the local `dir` was immediately overwritten by `args.dir` from `onRealmSetup`, so the kitchen-sink copy was dropped on the floor. Removed in those four files (alongside the now-unused `dirSync` / `copySync` / `join` imports).

For `directory-test`, the realm's actual content STILL needs to be the kitchen sink — that's now made explicit via `fixture: 'realistic'` rather than relying on the implicit default.

## Test plan

- [ ] CI runs every migrated test against the named fixture and they pass.
- [ ] CI runs every other realm-server test file unchanged.

### DIFF
--- a/packages/realm-server/tests/realm-endpoints/cancel-indexing-job-test.ts
+++ b/packages/realm-server/tests/realm-endpoints/cancel-indexing-job-test.ts
@@ -36,6 +36,7 @@ module(`realm-endpoints/${basename(__filename)}`, function () {
       }
 
       setupPermissionedRealmCached(hooks, {
+        fixture: 'blank',
         permissions: {
           writer: ['read', 'write'],
           reader: ['read'],
--- a/packages/realm-server/tests/realm-endpoints/dependencies-test.ts
+++ b/packages/realm-server/tests/realm-endpoints/dependencies-test.ts
@@ -29,6 +29,7 @@ module(`realm-endpoints/${basename(__filename)}`, function (hooks) {
   });
 
   setupPermissionedRealmCached(hooks, {
+    fixture: 'blank',
     permissions: {
       '*': ['read'],
     },
--- a/packages/realm-server/tests/realm-endpoints/directory-test.ts
+++ b/packages/realm-server/tests/realm-endpoints/directory-test.ts
@@ -1,8 +1,6 @@
 import { module, test } from 'qunit';
 import type { Test, SuperTest } from 'supertest';
-import { join, basename } from 'path';
-import { dirSync, type DirResult } from 'tmp';
-import { copySync } from 'fs-extra';
+import { basename } from 'path';
 import type { Realm } from '@cardstack/runtime-common';
 import {
   setupPermissionedRealmCached,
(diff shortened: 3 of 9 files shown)


lukemelia:
## Per-file benchmarks

Single run each via `packages/realm-server/scripts/measure-test-file.sh <file> 1` on local hardware (Apple Silicon, host/synapse/realm dev stack running), comparing `main` (HEAD `e32387deff`) vs this branch.

| File | New fixture | Before (s) | After (s) | Δ | Status before → after |
|---|---|---|---|---|---|
| `cancel-indexing-job-test` | `blank` | 454.73 | 418.31 | **-36.4 (-8.0%)** | FAIL → FAIL † |
| `dependencies-test` | `blank` | 207.28 | 166.06 | **-41.2 (-19.9%)** | FAIL → ✅ OK |
| `directory-test` | `realistic` (kept) | 328.60 | 331.94 | +3.3 (+1.0%) | FAIL → FAIL † |
| `info-test` | `blank` | 505.06 | 419.69 | **-85.4 (-16.9%)** | FAIL → FAIL † |
| `lint-test` | `blank` | 702.65 | 662.25 | **-40.4 (-5.8%)** | FAIL → ✅ OK |
| `mtimes-test` | `blank` | 166.13 | 122.67 | **-43.5 (-26.2%)** | FAIL → ✅ OK |
| `permissions-test` | `blank` (was `fileSystem: {}`) | 385.86 | 383.35 | -2.5 (-0.6%) | OK → OK |
| `search-test` | `simple` | 1398.83 | 1375.58 | -23.3 (-1.7%) | FAIL → FAIL † |
| `user-test` | `blank` | 397.14 | 354.74 | **-42.4 (-10.7%)** | FAIL → ✅ OK |
| **Total** | | **4546.28** | **4234.59** | **-311.7 (-6.9%)** | |

† All FAIL→FAIL cases are local-env-specific: realistic-fixture template builds occasionally exceed qunit's 60s `testTimeout` on local hardware, the offending realm holds port 4444, and subsequent realms cascade into `EADDRINUSE`. The original code had the same property (it's not introduced by this PR — the same files FAIL on `main` too). CI hardware doesn't exhibit it.

### Reliability gains

Four files flip from local-FAIL to local-OK: `dependencies`, `lint`, `mtimes`, `user`. That's the more interesting signal than the wall-time deltas — moving these off the heavy realistic fixture means their template builds complete inside the qunit 60s timeout window even on local hardware. The migration removes a real flakiness floor.

### Expected non-movers

- `directory-test` is kept on `realistic` (it asserts on `dir/bar.txt` and `dir/subdir/` from `tests/cards/dir/`), so a flat wall time is the right answer.
- `permissions-test` was already `fileSystem: {}` (semantically the same realm shape as `blank`), so the swap is a no-op for content.
- `search-test` is `simple` but only saves 1.7% — the test runs many search queries against the indexed data, so test-work dominates over realm boot here.

### Caveats

- Single runs only (each file is multi-minute locally); medians would tighten variance but not change the direction.
- Local hardware shows much wider variance than CI; treat these as order-of-magnitude / sign-of-delta signal, not as production-ready microbenchmarks.

habdelra (approved):

PR #4825, opened 2026-05-14
Lockfile update

ef4:
Attempting to regenerate lockfile so that transitive deps can float to compatible versions.


lukemelia (commented):
All the boxel-icon updates seem unrelated to the PR description?

PR #4828, opened 2026-05-14
Auth SW: ask page for token on miss before falling through

richardhjtan:
## Summary

When the auth service worker intercepts an image / asset GET to a known realm host but has no token in its in-memory map, ask the controlling page for one via `MessageChannel` and retry with `Authorization` if a token comes back. Falls through to the existing unauthed fetch when no token is available, so current behavior is preserved for non-realm hosts and for users who genuinely have no session for the realm.

Part of a broader investigation into broken-image symptoms on staging and production where users see broken `` icons and a manual refresh fixes them. This PR addresses one of the failure modes — image requests that 401 because the SW's token map is stale relative to localStorage. Full investigation report shared separately.

Linear: [CS-11144](https://linear.app/cardstack/issue/CS-11144)

## What this fixes

The SW caches realm tokens in memory and is updated by the host page via `postMessage`. If the cache is missing a token at the moment an `` GET is intercepted, the SW used to forward the request unauthenticated and the realm-server returned 401. After this PR, the SW asks the page for the token, reads from the authoritative `SessionLocalStorageKey` in localStorage, and retries with auth.

This is a freshness / resilience fix on the existing auth path — it does **not** change which realms a user has access to.

## Implementation notes

- **Origin gating** — fallback only fires when the request origin is one we've ever held a realm token for, so unrelated cross-origin asset requests don't pay any round-trip cost.
- **Single-flight** — concurrent misses for the same URL share one `MessageChannel` round-trip.
- **Short timeout (200 ms)** — if the page doesn't reply, fall through. Worst-case impact on page load is bounded.
- **No new auth surface** — only reads tokens the page already has. No new IAM, no new endpoints.

## Files

- `packages/host/public/auth-service-worker.js` — fetch handler now does on-miss MessageChannel lookup.
- `packages/host/app/utils/auth-service-worker-registration.ts` — listens for the SW's token request and replies from localStorage.
- `packages/host/tests/unit/auth-service-worker-test.ts` — extended to cover the new path: positive case, single-flight, fall-through.

## Test plan

- [x] `pnpm lint` / `pnpm lint:types` green.
- [ ] Manual repro on staging: open a card that embeds an image from a realm where the token is in localStorage but the SW's map hasn't received it yet. Before → broken-image icon. After → image loads on first paint.
- [ ] Re-run the staging log scan ~7 days after deploy; image-401 count from this failure mode should drop to ~0.

## Out of scope

Other failure modes from the same investigation (URL-construction bugs on the indexer side, prerender base-URL handling, content-negotiation edge cases) are tracked separately and will land on their own branches.

### DIFF
--- a/packages/host/app/utils/auth-service-worker-registration.ts
+++ b/packages/host/app/utils/auth-service-worker-registration.ts
@@ -32,6 +32,23 @@ export async function registerAuthServiceWorker(): Promise<void> {
     }
   });
 
+  // Respond to on-miss token lookups from the SW. The SW asks here when it
+  // intercepts a GET to a known realm host but has no token in its in-memory
+  // map (SW activation race, post-upload window before per-realm sync lands,
+  // etc.). localStorage is the authoritative source of currently-valid
+  // tokens — the SW's map is a derived cache.
+  navigator.serviceWorker.addEventListener('message', (event) => {
+    if (!event.data || event.data.type !== 'request-realm-token') {
+      return;
+    }
+    let port = event.ports?.[0];
+    if (!port) {
+      return;
+    }
+    let { realmURL, token } = resolveTokenForRequestURL(event.data.requestURL);
+    port.postMessage({ realmURL, token });
+  });
+
   try {
     await navigator.serviceWorker.register('/auth-service-worker.js', {
       scope: '/',
@@ -120,4 +137,34 @@ function readTokensFromStorage(): Record<string, string> | undefined {
     // ignore parse errors
   }
   return undefined;
+}
+
+// Find the longest realm-URL prefix in localStorage that matches the given
+// request URL. Returns `undefined` for both fields when nothing matches —
+// the SW will then preserve its existing pass-through behavior for that
+// request.
+function resolveTokenForRequestURL(requestURL: string | undefined): {
+  realmURL?: string;
+  token?: string;
+} {
+  if (!requestURL) {
+    return {};
+  }
+  let tokens = readTokensFromStorage();
+  if (!tokens) {
+    return {};
+  }
+  let bestRealmURL: string | undefined;
+  for (let realmURL of Object.keys(tokens)) {
+    if (
+      requestURL.startsWith(realmURL) &&
+      (!bestRealmURL || realmURL.length > bestRealmURL.length)
+    ) {
+      bestRealmURL = realmURL;
+    }
+  }
+  if (!bestRealmURL) {
+    return {};
+  }
+  return { realmURL: bestRealmURL, token: tokens[bestRealmURL] };
 }
--- a/packages/host/app/utils/auth-service-worker-registration.ts
+++ b/packages/host/app/utils/auth-service-worker-registration.ts
@@ -10,7 +10,18 @@ import window from 'ember-window-mock';
 
 import { SessionLocalStorageKey } from './local-storage-keys';
 
-function isServiceWorkerSupported(): boolean {
+// Structural so tests can stub without the full service surface.
+export interface AuthServiceWorkerDeps {
+  realmService: {
+    realmOf(input: URL): string | undefined;
+    reauthenticate(realmURL: string): Promise<string | undefined>;
+  };
+  matrixService: {
+    readonly isLoggedIn: boolean;
+  };
+}
+
+export function isServiceWorkerSupported(): boolean {
   return (
     !isTesting() &&
     typeof navigator !== 'undefined' &&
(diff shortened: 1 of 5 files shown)


packages/host/public/auth-service-worker.js
```diff
@@ -81,31 +90,120 @@ self.addEventListener('fetch', (event) => {
...
+  // No token in the map. Only attempt the on-miss client fallback when the
+  // request is to a host we've ever held a realm token for — that keeps the
+  // round-trip cost off every unrelated cross-origin asset request.
+  let requestOrigin;
+  try {
+    requestOrigin = new URL(url).origin;
+  } catch {
+    return;
+  }
+  if (!realmHosts.has(requestOrigin)) {
```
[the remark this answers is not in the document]
richardhjtan: Fixed in fe889a4 — gate now only short-circuits when `realmHosts.size > 0`, so cold start falls through to the on-miss lookup. Covered by the "asks the client at cold start when realmHosts is empty" test.

packages/host/public/auth-service-worker.js
```diff
@@ -81,31 +90,120 @@ self.addEventListener('fetch', (event) => {
...
+          recordRealmHost(reply.realmURL);
+          resolve(reply.token);
+        } else {
+          resolve(undefined);
+        }
+      };
+      // Ask the first window client. If multiple are open, any one of them
+      // can answer from the shared localStorage / session state.
+      clientList[0].postMessage({ type: 'request-realm-token', requestURL }, [
+        channel.port2,
```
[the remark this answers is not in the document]
richardhjtan: Fixed in fe889a4 — added `pickClientToAsk(initiatingClientId)` which prefers `self.clients.get(event.clientId)` and only falls back to the first window if the initiating client is gone.

packages/host/public/auth-service-worker.js
```diff
@@ -81,31 +90,120 @@ self.addEventListener('fetch', (event) => {
...
+      ]);
```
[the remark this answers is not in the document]
richardhjtan: Fixed in fe889a4 — added "times out and falls through when the client never replies" test that uses a never-resolving `clientTokenLookup` and asserts `fallthrough-fetch`.

packages/host/app/utils/auth-service-worker-registration.ts
```diff
@@ -121,3 +138,33 @@ function readTokensFromStorage(): Record<string, string> | undefined {
   }
   return undefined;
 }
+
+// Find the longest realm-URL prefix in localStorage that matches the given
+// request URL. Returns `undefined` for both fields when nothing matches —
+// the SW will then preserve its existing pass-through behavior for that
+// request.
+function resolveTokenForRequestURL(requestURL: string | undefined): {
```
habdelra, line 199: what if you need to get a new token here? since this is sync I don't think you can do that--but more deeply, this is a callback for an event which needs a sync function. handling an async mechansim to get a new token from the server will be a little tricky to wire thru properly....

packages/host/public/auth-service-worker.js
```diff
@@ -81,31 +90,142 @@ self.addEventListener('fetch', (event) => {
...
+      }, TOKEN_REQUEST_TIMEOUT_MS);
+      channel.port1.onmessage = (event) => {
+        if (settled) return;
+        settled = true;
+        clearTimeout(timer);
+        let reply = event.data;
+        if (reply && reply.realmURL && reply.token) {
+          realmTokens.set(reply.realmURL, reply.token);
+          recordRealmHost(reply.realmURL);
+          resolve(reply.token);
```
habdelra, line 154: it looks like you have the basic machinery to deal with an async token request already. probably just need to plumb asynchronicity in the DOM side if you wanna handle having the DOM request a new token from teh server.

packages/host/app/utils/auth-service-worker-registration.ts
richardhjtan: Yeah good point. Added changes that make the handler async and call realmService.reauthenticate on a localStorage miss. The wiring turned out simpler than expected since reauthenticate already persists + syncs the new token back to the SW as a side effect.

packages/host/public/auth-service-worker.js
richardhjtan: Added a small two-phase thing: 
- page posts {type:'pending'} first to extend the SW's timeout to 3s
- then posts the real token when reauthenticate resolves.

That way, the common cases still resolve fast, and only the actual refresh path waits longer.

backspace (approved):

PR #4829, opened 2026-05-14
Prerender: install <base href> for the card's directory

richardhjtan:
## Summary

Set a `<base href>` to the card's containing directory at the start of the prerender's `model()` hook, so relative `` / `<a href>` in a card template resolves against the card's realm instead of the synthetic `/render/<encoded-card>/<seq>/{cardRender:true}/html/...` URL the prerender browser is sitting at. Remove it on `deactivate()`.

This is one of several broken-image failure modes seen on staging and production. Full investigation report shared separately.

Linear: [CS-11146](https://linear.app/cardstack/issue/CS-11146)

## What this fixes

Card templates that use a relative image src (e.g. ``) used to issue requests like `/render/.../html/<format>/green-mango.png` during prerender, which 404. The 404 itself was logged repeatedly, and the broken `src` was baked into the captured HTML so end users also saw a broken icon. Native browser URL resolution against the new `<base>` does the right thing — no template rewrites needed.

## Implementation notes

- `<base>` is inserted as the first child of `<head>`, so it precedes anything that would resolve against the document base.
- Reused across format passes for the same card (`isolated` / `fitted` / `embedded`); updated rather than re-created if the card id changes.
- Removed on route deactivate so subsequent non-render routes are not affected.
- Same-origin app assets in the host's `index.html` are root-absolute (`/foo`), so they ignore the `<base>` and continue to work.

## Files

- `packages/host/app/routes/render.ts` — `#installPrerenderBaseHref` / `#removePrerenderBaseHref`, wired into `model()` and `deactivate()`.

## Test plan

- [ ] `pnpm lint:types` green.
- [ ] Manual repro on staging: trigger a re-index of a card that uses a relative `` (e.g. an `experiments/Friend*.json` card with `green-mango.png`). Inspect the captured HTML in `boxel_index.html_format` for the resolved URL.
- [ ] 7 days post-deploy, re-run the log scan for `/render/.../html/...` 404s. Count should drop to ~0.

## Out of scope

Other failure modes from the same investigation are tracked separately ([CS-11144](https://linear.app/cardstack/issue/CS-11144) shipped on PR #4828, [CS-11145](https://linear.app/cardstack/issue/CS-11145), [CS-11147](https://linear.app/cardstack/issue/CS-11147)).

### DIFF
--- a/packages/host/app/routes/render.ts
+++ b/packages/host/app/routes/render.ts
@@ -167,6 +167,7 @@ export default class RenderRoute extends Route<Model> {
     (globalThis as any).__waitForRenderLoadStability = undefined;
     window.removeEventListener('boxel-render-error', this.handleRenderError);
     this.#detachWindowErrorListeners();
+    this.#removePrerenderBaseHref();
     this.lastStoreResetKey = undefined;
     this.renderBaseParams = undefined;
     this.lastRenderErrorSignature = undefined;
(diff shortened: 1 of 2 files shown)


richardhjtan:
Would defer this for now, until we find that there are several usages needed to load imageUrl with a relative path

More comment info in the linear ticket: https://linear.app/cardstack/issue/CS-11146/prerender-html-resolves-relative-image-urls-against-synthetic-render#comment-28433f17

PR #4833, opened 2026-05-14
CS-11152: fix catalog remix — resolve scoped prefixes in fetcher

richardhjtan:
## Summary
- Catalog remix and other `authedFetch` calls against scoped card ids (`@cardstack/catalog/...`) were failing with 404 on `https://app.boxel.ai/@cardstack/catalog/<rest>`.
- Root cause: [`fetcher.ts`](../blob/cs-11152-fail-to-remix/packages/runtime-common/fetcher.ts) wraps string inputs in `new Request(string)` before any prefix resolution. `new Request` joins the string against `document.baseURI`, producing the broken URL. `VirtualNetwork.fetch` resolves scoped strings, but it's inside the chain — every `authedFetch` caller passes through `fetcher` first.
- Fix: resolve registered scoped prefixes (`isRegisteredPrefix` / `resolveCardReference`) before the `new Request` step in `fetcher`, mirroring `VirtualNetwork.fetch`.

Trigger from the user's stack: `CardOrFieldListingAdapter.remix` → `RemixCommand` → `ListingInstallCommand` → `FetchCardJsonCommand({ cardIdentifier: sourceCard.id })` where `sourceCard.id` is already in scoped form because the indexer rewrites instance URLs via `unresolveResourceInstanceURLs` (card-indexer.ts:81) for portability.

Linear: https://linear.app/cardstack/issue/CS-11152/fail-to-remix

## Test plan
- [x] New unit test `Unit | fetcher > resolves a registered scoped prefix before constructing the Request` passes (`pnpm test:ember --filter "Unit | fetcher"` in `packages/host`).
- [x] Deployed reproduction (Remix on a catalog listing — Mortgage Calculator / Wine Cellar Card Definition) no longer 404s.
- [ ] Existing fetcher unit tests still pass (no regression in URL/redirect/retry/middleware paths).
- [ ] Existing `auth-error-guard-test` and `index-query-engine-test` paths still pass (they construct `fetcher(virtualNetwork.fetch, ...)`).

Before Fix - Remix failed 
Screenshot 2026-05-15 at 12 34 45 AM
After Fix
Screenshot 2026-05-15 at 1 00 24 AM

### DIFF
--- a/packages/host/tests/unit/fetcher-test.ts
+++ b/packages/host/tests/unit/fetcher-test.ts
@@ -2,6 +2,8 @@ import { module, test } from 'qunit';
 
 import {
   fetcher,
+  registerCardReferencePrefix,
+  unregisterCardReferencePrefix,
   type FetcherMiddlewareHandler,
 } from '@cardstack/runtime-common';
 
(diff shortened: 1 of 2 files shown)


lukemelia (commented):
@backspace please check whether this approach is compatible with your cardstack/base approach

backspace (approved):
> @backspace please check whether this approach is compatible with your cardstack/base approach

yes this is fine for now

PR #4840, opened 2026-05-14
CS-11125: per-realm advisory lock on data-plane write paths

lukemelia:
_This PR is best reviewed with Hide Whitespace enabled._

## Summary
- Threads the per-realm `withWriteLock` (now on `DBAdapter`, via #4839) through the data-plane mutations in `packages/runtime-common/realm.ts`, so two replicas behind a load balancer can no longer race on the same realm URL.
- Pre-CS-11125 the lock from CS-10891 only wrapped realm-lifecycle handlers (create / publish / unpublish / delete). `POST` / `PATCH` / `DELETE <realm>/<path>` and `POST <realm>/_atomic` — the routes real client traffic actually hits — did not take it. Single-replica deploys were fine because the Node event loop was the implicit serializer; the moment a second replica appears, last-writer-wins on EFS and interleaved `boxel_index` rows.

Stacks on #4839.

Linear: [CS-11125](https://linear.app/cardstack/issue/CS-11125).

## What's in
- `write` / `writeMany` / `delete` / `deleteAll` wrap their inner work in `dbAdapter.withWriteLock(this.url, ...)`. Inner methods are renamed `_batchWriteUnlocked` / `_deleteUnlocked` / `_deleteAllUnlocked` so handlers that need the lock around a wider critical section reuse the inner without re-entering the lock (a second `pg_advisory_xact_lock` on the same key would block on a different pinned pool connection).
- `handleAtomicOperations` takes the lock at the handler boundary so the `add` / `update` precheck is in the same critical section as the `writeMany` — fixes the `/_atomic` TOCTOU spelled out in the ticket.
- `patchCardInstance` takes the lock around the `indexEntry` read + merge + write, so two concurrent `PATCH`es from different replicas can't both read the same `original` and clobber each other's merge.

Lock granularity stays per-realm — concurrent writes to different realms remain fully parallel. SQLite (host) is a passthrough, so this is a no-op there.

## Regression test
`packages/realm-server/tests/data-plane-write-lock-test.ts`:
- Two concurrent PATCHes against the same card with non-overlapping attribute changes — both must persist. Without the lock the second writer computes its merge from pre-first state and drops the first writer's field.
- Two concurrent `/_atomic` `add` ops on the same href — must yield 201 + 409, not 201 + 201. Without the lock both pass the precheck and both write.

## Compatibility
- Single-replica deploys: the lock is a no-op without contention — pure overhead is one `pg_advisory_xact_lock` round-trip per data-plane mutation.
- Lands before CS-10899 (atomicity follow-up) and the cache-invalidation tickets that assume writes are already serialized.

## Test plan
- [x] `realm-advisory-locks-test` (7/7) + `realm-cleanup-transaction-test` (3/3) pass after the rebase on #4839.
- [x] `pnpm exec tsc -p packages/runtime-common --noEmit` and the realm-server / postgres / host projects — zero new type errors vs. main.
- [ ] `pnpm test:wait-for-servers` against a running stack to drive the new `data-plane-write-lock-test` end-to-end. (Requires Synapse + base realm-server up — not run locally, deferred to CI.)
- [ ] Eyeballed `patchCardInstance` / `handleAtomicOperations` re-indentation diffs (heavy because everything is wrapped in a closure) — please review the structural changes carefully.

### DIFF
--- a/packages/realm-server/tests/data-plane-write-lock-test.ts
+++ b/packages/realm-server/tests/data-plane-write-lock-test.ts
@@ -1,0 +1,210 @@
+import { module, test } from 'qunit';
+import type { Test, SuperTest } from 'supertest';
+import { basename } from 'path';
+import type { Server } from 'http';
+import type { DirResult } from 'tmp';
+import type { Realm } from '@cardstack/runtime-common';
+import {
+  SupportedMimeType,
+  type LooseSingleCardDocument,
+} from '@cardstack/runtime-common';
+import {
+  setupPermissionedRealmCached,
+  createJWT,
+  type RealmRequest,
+  withRealmPath,
+} from './helpers';
+import '@cardstack/runtime-common/helpers/code-equality-assertion';
+
+// CS-11125 regression coverage: per-realm advisory lock on the data-plane
+// write paths. These tests reach the realm-server HTTP API the same way a
+// real client would, and rely on Node's event-loop interleaving + the
+// Postgres advisory lock to serialize critical sections that pre-CS-11125
+// raced. Without the lock applied to PATCH and /_atomic, the assertions
+// below detect the lost update / TOCTOU directly.
+module(basename(__filename), function () {
+  module('CS-11125: data-plane write serialization', function () {
+    let realmURL = new URL('http://127.0.0.1:4444/test/');
+    let testRealm: Realm;
+    let request: RealmRequest;
+
+    function onRealmSetup(args: {
+      testRealm: Realm;
+      testRealmHttpServer: Server;
+      request: SuperTest<Test>;
+      dir: DirResult;
+    }) {
+      testRealm = args.testRealm;
+      request = withRealmPath(args.request, realmURL);
+    }
+
+    module('concurrent PATCH on same card', function (hooks) {
+      setupPermissionedRealmCached(hooks, {
+        fixture: 'simple',
+        realmURL,
+        permissions: {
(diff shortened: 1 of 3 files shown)


lukemelia:
Addressed both Copilot review comments and the Lint failure:

* **Lint** ([prettier on `realm.ts:4111`](https://github.com/cardstack/boxel/actions/runs/25886894286/job/76080742988)) — collapsed the `buildCardJsonEtag(...)` call back onto a single line. Unrelated to advisory locks, blocking merge.
* **[r3243846068](https://github.com/cardstack/boxel/pull/4840#discussion_r3243846068)** — `Realm.clearLocalCaches()` deletion was an unintentional rebase casualty. Restored at the original CS-11043 site in commit e8f6a07.
* **[r3243846106](https://github.com/cardstack/boxel/pull/4840#discussion_r3243846106)** — added `?waitForIndex=true` to the two `/_atomic` POSTs in `data-plane-write-lock-test.ts`. `/_atomic` returns on durability, not on index completion, so the immediate GET could race the indexer. The PATCH test in the same file uses `writeMany` which defaults to `waitForIndex: true`, so only `/_atomic` needed the explicit opt-in.

Follow-up [CS-11156](https://linear.app/cardstack/issue/CS-11156) (cross-replica cache invalidation via NOTIFY broadcast) is open as draft PR #4842, stacked on this branch.

habdelra (approved):

PR #4847, opened 2026-05-15
fix: pin vite to 127.0.0.1 behind same-port dispatcher

backspace:
Hi, I'm Claude — Buck asked me to investigate why `pnpm start` in `packages/host` was reporting "Vite appears to be running" but `https://localhost:4200` wouldn't connect. Here's what I found and a one-line-ish fix.

## Problem

The same-port redirect dispatcher introduced in 6c2d7f8318 binds public port 4200 on `127.0.0.1`, peeks the first byte off each connection, then forwards TLS bytes to vite at an internal loopback port via `net.connect(internalPort, '127.0.0.1')` (scripts/vite-with-traefik.js:95).

But `runVite` was invoking `vite --port <p>` with no `--host`, so vite default-binds to `localhost`. On macOS Sonoma+ / Node 17+, `dns.lookup('localhost')` returns `::1` first, so vite ends up listening on `[::1]:<p>` only.

The dispatcher's IPv4 upstream `connect` then fails immediately, its `upstream.on('error', () => socket.destroy())` handler tears down the client socket mid-TLS-handshake, and the browser sees `ERR_CONNECTION_CLOSED` — even though the dispatcher cheerfully logs `Listening on ... → vite at 127.0.0.1:<p>`.

Plain `http://localhost:4200` still works because the 308 response is written directly to the accepted socket without touching upstream, which is why testem and curl-against-http didn't catch this.

Reproduced locally:
- `lsof`: dispatcher on `IPv4 127.0.0.1:4200`, vite on `IPv6 [::1]:53460`.
- `curl -k https://localhost:4200/` → `SSL_ERROR_SYSCALL`.
- Chrome → `net::ERR_CONNECTION_CLOSED`.

## Fix

Pin both ends of the pipe to the same loopback family. Pass `host: '127.0.0.1'` to vite when it runs behind the dispatcher so it explicitly binds the IPv4 loopback the dispatcher connects on. The Traefik / env-mode path is unaffected — that one still passes `--host` (all-interfaces).

After the fix:
- `curl https://localhost:4200/` → `200, http_version 2` ✅
- `curl http://localhost:4200/` → `308 → https://localhost:4200/` ✅
- Chrome loads `https://localhost:4200/` cleanly ✅

## Test plan

- [ ] In a shell where `REALM_SERVER_TLS_CERT_FILE` is set (i.e. the mkcert cert exists), `pnpm start` from `packages/host` and visit `https://localhost:4200` in a browser — should load the host app over HTTP/2, not `ERR_CONNECTION_CLOSED`.
- [ ] `curl http://localhost:4200/` returns a 308 to `https://localhost:4200/`.
- [ ] Env-mode (`BOXEL_ENVIRONMENT=...`) still binds vite on all interfaces so Traefik in Docker can reach it via `host.docker.internal`.

### DIFF
--- a/packages/host/scripts/vite-with-traefik.js
+++ b/packages/host/scripts/vite-with-traefik.js
@@ -21,7 +21,7 @@ const { spawn } = require('child_process');
 const path = require('path');
 const net = require('net');
 
-function runVite({ subcommand, port, allHosts, extraEnv, nodeMemory }) {
+function runVite({ subcommand, port, allHosts, host, extraEnv, nodeMemory }) {
   const args = ['vite'];
   if (subcommand) args.push(subcommand);
   args.push('--port', String(port), '--strictPort');
@@ -30,6 +30,8 @@ function runVite({ subcommand, port, allHosts, extraEnv, nodeMemory }) {
     // vite via host.docker.internal:<port>. Vite's default is 127.0.0.1
     // only, which is unreachable from inside the container.
     args.push('--host');
+  } else if (host) {
+    args.push('--host', host);
   }
   const env = { ...process.env, ...(extraEnv || {}) };
   if (nodeMemory) {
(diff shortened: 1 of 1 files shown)


habdelra (approved):

backspace:
I’ll keep this unmerged for now until it actually works for the whole stack.

habdelra:
might be worth having claude look at the user permissions table to make sure teh migration script updated it correctly. a mistake there could cause a 401 error where we wouldn't expect it.

backspace:
that does seem a likely culprit, it’s `https://localhost:4201/base/` in the table but in a working environment it’s `https://cardstack.com/base/`

backspace:
> that does seem a likely culprit, it’s `https://localhost:4201/base/` in the table but in a working environment it’s `https://cardstack.com/base/`

okay, this is almost certainly because of the ongoing chaos I experience as I work on `@cardstack/base`. Usually I use “environment mode” to bypass that so I can have clean parallel environments but I didn’t do that when exercising your PR because I didn’t want any conflicts or confusion from that. I fixed added the required permissions rows and it’s working locally. Now I’ll look over the code of the PR!

PR #4865, opened 2026-05-18
CS-11170: Include FileDefs in CardsGrid

lukemelia:
Linear: [CS-11170](https://linear.app/cardstack/issue/CS-11170/implement-include-filedefs-in-cardsgrid)
Project: [Include FileDefs in CardsGrid](https://linear.app/cardstack/project/include-filedefs-in-cardsgrid-302762ce6b78)

## Summary

Surfaces FileDef instances alongside CardDef instances in the CardsGrid sidebar. Implements the six-step plan from the Linear project description. The sidebar now shows two top-level groups — **All Cards** and **All Files** — each expandable to specific subtypes. Clicking a file opens it as an isolated stack item in interact-mode; editing remains in code-mode (a new "Open in Code Mode" affordance is added to the file kebab).

## What landed

- **Index pipeline** — `realm_meta.value` is now a `{ instances, files }` object instead of a bare array. `index-writer` runs the per-type aggregation via a new `#fetchTypeSummary` helper. `normalizeRealmMetaValue` tolerates the legacy array shape so this is a hot rollout, no DB migration, no required reindex (the next index run upgrades each realm in place).
- **File display names** — `FileExtractResponse` carries `displayNames` walked from the resolved FileDef class chain (new `getDisplayNames` helper in `FileDefAttributesExtractor`, mirroring the card-side meta route). `file-indexer` persists those to `boxel_index.display_names`.
- **`_types` endpoint** — `makeCardTypeSummaryDoc` emits a flat JSON:API list with a `kind: 'instance' | 'file'` discriminator. `FederatedCardTypeSummaryEntry` and host-side typings updated to match.
- **CardsGrid sidebar** — `cards-grid.gts` partitions `_types` results by `kind`, builds `fileTypeFilters`, and adds an `allFilesFilter` group that's hidden when the realm has no files. Base `FileDef` is excluded from the leaf list to avoid duplicating the group label.
- **File clicks → interact-mode** — Extracted `knownFileMetaUrls` to `host/app/lib/known-file-meta-urls.ts` (back-compat re-export from `prerendered-card-search.gts`). `detectStackItemTypeForTarget` consults the registry as a fallback so file URLs clicked from a freshly-loaded CardsGrid resolve to `'file'` stack items even before the file-meta resource has landed in the store. Added an "Open in Code Mode" entry to the interact-context file kebab menu.

## Test plan

- [x] Type-clean across `runtime-common`, `host`, `base`, `realm-server` for touched files.
- [x] Existing `realm-server/tests/types-endpoint-test.ts` updated to require the `kind` discriminator on every response entry.
- [x] New `host/tests/unit/index-writer-test.ts` case asserts file rows partition into `realm_meta.value.files`, collapse duplicates, and don't bleed into the instances arm.
- [x] New `normalize-realm-meta-value-test` covers the legacy array, partitioned object, partial shape, null/undefined, and unrecognized JSONB.
- [ ] **Manual:** boot a local realm with a markdown file, hit `/_types` and verify each entry has `attributes.kind`. Open CardsGrid in the realm and verify the All Files group appears, expanded clicks build a file-typed query, and clicking a result opens a file stack item with no edit button but with "Open in Code Mode" available.
- [ ] **Manual:** boot against an unmigrated realm (no reindex yet — `realm_meta.value` is still an array) and verify CardsGrid still renders All Cards correctly, no All Files group, no errors.

## Deferred (per project plan)

- Polish for the full set of FileDef isolated subtypes — MVP ships with `TextFileDef` / `JsonFileDef` / `MarkdownDef` polished; the rest land behind a follow-up.
- Binary-file (image/PDF) fallback isolated component — out of scope for this PR.

https://github.com/user-attachments/assets/35edeb03-4401-49ff-a150-3ee5f6b9b9e1

### DIFF
--- a/packages/base/components/card-list.gts
+++ b/packages/base/components/card-list.gts
@@ -65,6 +65,18 @@ export default class CardList extends Component<Signature> {
     }
   }
 
+  // Render the filename fallback only when the prerender pipeline produced
+  // no HTML AND the row is not an error. Error rows have their own
+  // dedicated error component built by PrerenderedCard's constructor, which
+  // also has empty `html`; treating them like file fallbacks would swap a
+  // helpful "rendering error" affordance for a bare filename.
+  shouldRenderFallback(card: {
+    hasHtml?: boolean;
+    isError?: boolean;
+  }): boolean {
+    return card.hasHtml === false && !card.isError;
+  }
+
   <template>
     <ul
       class={{cn
@@ -94,7 +106,7 @@ export default class CardList extends Component<Signature> {
                   'boxel-card-list-item'
                   instance-error=card.isError
                   clickable=(if this.cardCrudFunctions.viewCard true false)
-                  fallback=(eq card.hasHtml false)
+                  fallback=(this.shouldRenderFallback card)
                 }}
                 data-test-instance-error={{card.isError}}
                 data-test-cards-grid-item={{removeFileExtension card.url}}
@@ -104,12 +116,14 @@ export default class CardList extends Component<Signature> {
                 tabindex={{if this.cardCrudFunctions.viewCard '0'}}
                 {{on 'click' (fn this.handleCardClick card.url)}}
               >
-                {{#if (eq card.hasHtml false)}}
+                {{#if (this.shouldRenderFallback card)}}
                   {{! CS-11171: file rows whose prerender produced no HTML
                       (currently `.gts`/`.ts` FileDef rows) — render a name
                       so the row is at least visible and the click handler on
                       this `<li>` can still route the user into interact-mode
-                      (and from there into Code Mode via the kebab menu). }}
+                      (and from there into Code Mode via the kebab menu).
+                      Error rows are excluded so PrerenderedCard's dedicated
+                      error component still gets rendered for them. }}
                   <div class='card-fallback' data-test-card-fallback>
                     <FileIcon class='card-fallback__icon' role='presentation' />
                     <div class='card-fallback__name'>
--- a/packages/base/components/card-list.gts
+++ b/packages/base/components/card-list.gts
@@ -7,6 +7,8 @@ import { consume } from 'ember-provide-consume-context';
 
 import { LoadingIndicator } from '@cardstack/boxel-ui/components';
 
+import FileIcon from '@cardstack/boxel-icons/file';
+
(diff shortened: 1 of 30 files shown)


packages/host/tests/integration/components/operator-mode-ui-test.gts
```diff
@@ -1087,7 +1087,7 @@ module('Integration | operator-mode | ui', function (hooks) {
     assert
       .dom('[data-test-boxel-picker-option-row="select-all"]')
       .containsText(
-        'Any Type (14)',
+        'Any Type (16)',
```
burieberry, line 1090: A `const` for these may make it easier to update for next time

packages/base/components/card-list.gts
```diff
@@ -48,6 +50,33 @@ export default class CardList extends Component<Signature> {
...
+  fileNameFromUrl(url: string): string {
+    try {
+      let pathname = new URL(url).pathname;
+      let segment = pathname.split('/').filter(Boolean).pop();
+      return segment ?? url;
+    } catch {
+      let segments = url.split('/').filter(Boolean);
+      return segments[segments.length - 1] ?? url;
+    }
+  }
```
burieberry, line 66: I think there are some other places in host that does something like this for filename. would be good to refactor and maybe integrate `@runtime-common/paths`

burieberry (approved):

packages/base/components/card-list.gts
lukemelia: Good point — filed [CS-11177](https://linear.app/cardstack/issue/CS-11177/consolidate-urlfilename-helpers-and-reuse-runtime-commonpaths) to add a shared `fileNameFromUrl` helper to `@runtime-common/paths` and migrate the ~16 duplicate call sites. Keeping it out of this PR so the scope stays focused on the FileDef plumbing.

packages/host/tests/integration/components/operator-mode-ui-test.gts
lukemelia: Done — extracted `ALL_REALM_TYPE_COUNT` at the top of the module and rewired all five occurrences (three `Any Type (16)` assertions + two bare `15` counts derived as `ALL_REALM_TYPE_COUNT - 1`). Also updated a stale "(15 types)" comment in the pagination test.

PR #4882, opened 2026-05-19
ci: Remove caching from host preview deployments

backspace:
This has confused me too:

CleanShot 2026-05-19 at 08 15 41@2x

but there’s no reason to cache preview deployments, we can afford to refetch every time to guarantee we’re always seeing the latest version.

Confirmation: I made an arbitrary `host` change, saw it in the preview, removed it, and opened a preview link without a shift-refresh, and the change was gone.


burieberry (approved):
Nice. I was certainly fooled by this before, too.

PR #4885, opened 2026-05-19
CS-11175: job-scoped resolved-doc search cache during prerender

habdelra:
## Summary

Layers a `Map<key, LinkableCollectionDocument>` above CS-11121's in-flight Map so **sequential** same-`(jobId, consumingRealm, query)` `store.search` calls hit a synchronous cache instead of paying another `_federated-search` round-trip. Host-side mirror of CS-11115 Phase 2's `JobScopedSearchCache` on the realm-server.

### Lookup order in `fetchSearchDoc`

```
fetchSearchDoc(query, realms)
  ├─ if (sameRealm && jobId && prerender) return searchCache.get(key)   ← this PR
  ├─ if (prerender) coalesce via inflightSearch.get(inflightKey)         ← CS-11121
  ├─ fetch _federated-search
  ├─ on settle: inflightSearch.delete(key)
  ├─ if (cacheKey) searchCache.set(cacheKey, doc)                        ← this PR
  └─ return doc
```

### Cache eligibility

- `__boxelRenderContext === true`
- `__boxelJobId` is set (only the prerender server's `prerenderVisitAttempt` stamps this)
- `realms.length === 1 && realms[0] === __boxelConsumingRealm` — same-realm only. Cross-realm reads bypass because peer realm-servers swap on their own job cadence and a stale cross-realm hit could freeze a value.

### Why the resolved-doc pattern (not promise-storage)

Entries store the resolved doc, not the in-flight promise — same reasoning as the realm-server's [`job-scoped-search-cache.ts:82`](https://github.com/cardstack/boxel/blob/main/packages/realm-server/job-scoped-search-cache.ts#L82): a slow first populate must not tail-latency-block waiters past their render-timeout window. The in-flight dedup belongs to CS-11121; this layer answers from already-resolved bytes once the first call completes.

### Clear paths

- Render-route `deactivate()` (per-visit, primary).
- `resetState()` / `resetCache()` (parity with other in-flight maps).
- Belt-and-braces job-id-change clear at fetch-entry, in case a prerender tab is reused across jobs without either firing.

### Stacking

Base is `cs-11121-store-side-in-flight-search-dedup-during-prerender-host-spa`. Re-target to `main` before merging so this PR lands cleanly without re-merging CS-11121's commits.

Linear: [CS-11175](https://linear.app/cardstack/issue/CS-11175)

## Test plan

- [ ] CI: `Unit | Utility | searchCacheKey` (7 cases — jobId/realm/query/pagination distinctness, query-order invariance, JSON-array encoding rules out delimiter collision).
- [ ] CI: `Integration | search resource > job-scoped resolved-doc search cache` (8 cases):
  - [ ] Sequential same-key store.search calls hit the cache (1 fetch total).
  - [ ] Outside prerender, every call re-fetches.
  - [ ] Prerender with no `__boxelJobId` bypasses cache.
  - [ ] Cross-realm reads bypass.
  - [ ] jobId change drops the cache (defensive clear).
  - [ ] `clearSearchCache()` drops entries.
  - [ ] Cache hit short-circuits before the in-flight Map is consulted.
  - [ ] `consumingRealm` mismatch on a single-realm search bypasses.
- [ ] Manual bench against `/ambitious/prianha` once stacked on top of CS-11121 — distinct `_federated-search` HTTP count should drop further than CS-11121 alone (sequential repeats inside a cohort render now hit the cache).

### DIFF
--- a/packages/host/app/routes/render.ts
+++ b/packages/host/app/routes/render.ts
@@ -162,13 +162,15 @@ export default class RenderRoute extends Route<Model> {
     // visit. The window is small (typically <1s per search) but the
     // cost of an explicit clear is also small.
     this.store.clearInFlightSearch();
-    // Drop the resolved-doc search cache. Entries are scoped to a
-    // single indexing job; the next visit's `__boxelJobId` will be a
-    // different value, so cached entries become unreachable on key
-    // mismatch — but clearing here also reclaims memory promptly
-    // rather than waiting for the job-id-change defensive clear
-    // inside `fetchSearchDoc` to fire on the next request.
-    this.store.clearSearchCache();
+    // The resolved-doc search cache is INTENTIONALLY NOT cleared
+    // here. A single indexing job renders many cards in the same
+    // prerender tab — each card navigation activates and deactivates
+    // this route, but all those visits share one `__boxelJobId` and
+    // a stable view of the consuming realm's `boxel_index`. Cached
+    // entries from earlier renders in the job are the entire point;
+    // dropping them per-render would defeat the cache. Cross-job
+    // invalidation is handled by `fetchSearchDoc`'s entry-time
+    // jobId-change clear (and by `resetState` on harder resets).
     (globalThis as any).__renderModel = undefined;
     (globalThis as any).__docsInFlight = undefined;
     (globalThis as any).__boxelRenderStage = undefined;
--- a/packages/host/app/routes/render.ts
+++ b/packages/host/app/routes/render.ts
@@ -162,6 +162,13 @@ export default class RenderRoute extends Route<Model> {
     // visit. The window is small (typically <1s per search) but the
     // cost of an explicit clear is also small.
     this.store.clearInFlightSearch();
+    // Drop the resolved-doc search cache. Entries are scoped to a
+    // single indexing job; the next visit's `__boxelJobId` will be a
+    // different value, so cached entries become unreachable on key
+    // mismatch — but clearing here also reclaims memory promptly
+    // rather than waiting for the job-id-change defensive clear
+    // inside `fetchSearchDoc` to fire on the next request.
+    this.store.clearSearchCache();
     (globalThis as any).__renderModel = undefined;
     (globalThis as any).__docsInFlight = undefined;
     (globalThis as any).__boxelRenderStage = undefined;
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -91,6 +91,7 @@ import {
   duringPrerenderHeaders,
   jobIdHeader,
 } from '../lib/prerender-fetch-headers';
+import { searchCacheKey } from '../lib/search-cache-key';
 import { searchInFlightKey } from '../lib/search-in-flight-key';
 import { errorJsonApiToErrorEntry } from '../lib/window-error-handler';
 import { getSearch } from '../resources/search';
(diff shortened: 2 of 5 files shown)


packages/host/app/services/store.ts
```diff
@@ -955,40 +987,92 @@ export default class StoreService extends Service implements StoreInterface {
...
-      );
-      this.inflightSearch.set(key, pending);
-      return await pending;
+    } else {
+      doc = await this.fetchSearchDocUncoalesced(query, realms);
+    }
+
+    if (cacheKey !== undefined) {
+      this.searchCache.set(cacheKey, doc);
     }
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] 

Good catch — real race. Fixed in `6b69df6` with a `searchCacheGeneration` counter that mirrors the in-flight Map's identity check:

- Every path that empties the cache (`clearSearchCache`, `resetState`, `resetCache`, the entry-time jobId-change clear) now bumps the counter.
- `fetchSearchDoc` captures the generation *after* its own entry-time clear, and on resolve only writes to the cache if the generation is still that snapshot value.
- The caller still receives the resolved doc — only the cache write is suppressed.

New behavior test asserts the suppression: `clearSearchCache` fired while a fetch is parked causes the next same-key call to re-fetch (i.e., the resolved-but-stale doc didn't leak into the post-clear cache).

packages/host/app/routes/render.ts
```diff
@@ -162,6 +162,13 @@ export default class RenderRoute extends Route<Model> {
     // visit. The window is small (typically <1s per search) but the
     // cost of an explicit clear is also small.
     this.store.clearInFlightSearch();
+    // Drop the resolved-doc search cache. Entries are scoped to a
+    // single indexing job; the next visit's `__boxelJobId` will be a
+    // different value, so cached entries become unreachable on key
+    // mismatch — but clearing here also reclaims memory promptly
+    // rather than waiting for the job-id-change defensive clear
+    // inside `fetchSearchDoc` to fire on the next request.
+    this.store.clearSearchCache();
```
habdelra: an indexing job spans mulitple renders--why are we dropping this?!?!?! it defeats the whole purpose

habdelra: [Claude Code 🤖] 

You're completely right — apologies. A single job spans many card renders in the same tab; each navigation activates+deactivates the render route but all those visits share one `__boxelJobId`. Clearing the cache per-render defeated the entire point.

Fixed in `a3287bc`:

1. Dropped the `this.store.clearSearchCache()` call from `render.ts:deactivate`. The render route now leaves the search cache alone across visits.
2. Cross-job invalidation is handled at fetch-entry by the `__boxelJobId !== this.searchCacheJobId` check — when the next job's first fetch arrives carrying a new jobId, the cache is dropped before that fetch consults it. `resetState` / `resetCache` continue to clear on harder service resets.
3. Tightened docs on the `searchCache` field and the `clearSearchCache()` method so the per-render-clear bug doesn't re-emerge. The field comment now says "Lifetime: the entire indexing job. … The render route's `deactivate` deliberately does NOT clear this cache."
4. Added a behavior test exercising the moral equivalent of "deactivate between renders": clearing the in-flight Map between two same-key searches under the same jobId still hits the resolved-doc cache (1 fetch total across both calls).

lukemelia (approved):

PR #4886, opened 2026-05-19
Fix Ctrl+E edit shortcut on non-QWERTY layouts (CS-11092)

lukemelia:
## Summary
- `ember-keyboard`'s `'ctrl+KeyE'` listener form matches `KeyboardEvent.code` (physical position), so Dvorak users — whose physical KeyE keycap produces `.` — were getting the shortcut on Ctrl+. and silently missing it on the key that actually types `e`. Reported in [CS-11092](https://linear.app/cardstack/issue/CS-11092/edit-keyboard-shortcut-ignores-keyboard-layout).
- Switched the listener to `'ctrl+e'`, which `ember-keyboard` matches against `event.key`. The browser already resolves that against the user's active layout, so QWERTY / Dvorak / AZERTY / etc. all fire the shortcut on whichever physical key produces `e` on their layout.
- Updated the existing tests to pass `'e'` (the character the user types) instead of `'KeyE'` for semantic clarity, and added a regression test that dispatches raw `KeyboardEvent`s with mismatched `key`/`code` to lock in the layout-aware behavior (`triggerKeyEvent` can't decouple the two).

## Test plan
- [x] Existing tests at `interact-submode-test.gts` pass with the new listener string.
- [x] New "Ctrl+E respects keyboard layout (CS-11092)" test asserts:
  - dispatching `{ key: 'e', code: 'Period', ctrlKey: true }` (a Dvorak press of the key that produces `e`) toggles into edit mode, and
  - dispatching `{ key: '.', code: 'KeyE', ctrlKey: true }` (a Dvorak press of the QWERTY-E keycap) does **not** toggle — the pre-fix bug.
- [ ] Manual verification on a Dvorak software layout (System Settings → Keyboard → Input Sources → Dvorak on macOS): Ctrl+E in interact-mode toggles edit; Ctrl+. does nothing.
- [ ] Confirm no regression for QWERTY users: Ctrl+E still toggles, Ctrl+Shift+E and Cmd+E still no-op.

### DIFF
--- a/packages/host/app/components/operator-mode/interact-submode.gts
+++ b/packages/host/app/components/operator-mode/interact-submode.gts
@@ -852,9 +852,10 @@ export default class InteractSubmode extends Component {
         }}
         style={{this.backgroundImageStyle}}
         {{onKeyMod 'Escape' this.handleEscape}}
-        {{! Bind Ctrl+E on every platform — Mac users get Ctrl+E too,
-           because Cmd+E is taken by browsers ("Use Selection for Find"). }}
-        {{onKeyMod 'ctrl+KeyE' this.handleToggleEdit}}
+        {{! Ctrl+E (not Cmd+E — taken by browsers' "Use Selection for Find").
+           Lowercase 'e' matches event.key, so Dvorak/AZERTY users get the
+           shortcut on whatever key produces 'e' on their layout. }}
+        {{onKeyMod 'ctrl+e' this.handleToggleEdit}}
       >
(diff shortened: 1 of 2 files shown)


habdelra (approved):
wow

backspace:
I tried this in the preview deployment and it does fix the problem for me 🎉 but the test failure is relevant: <pre>Must provide a `key` to `triggerKeyEvent` that starts with an uppercase character but you passed `e`.</pre>

PR #4888, opened 2026-05-19
indexer: clear has_error / error_doc on tombstone + write-boundary invariant

habdelra:
## Summary

- **`tombstoneEntries`** now lists `has_error` (= `false`) and `error_doc` (= `null`) in its upsert column set, so the ON CONFLICT SET clause explicitly clears any stale error state when marking a URL deleted.
- **`updateEntry`** rejects any `instance-error` / `file-error` entry whose `error.message` is empty / missing, so a future upstream regression that hands over an unusable error fails fast into the worker's per-job reservation cap instead of writing a black-hole row.

## The bug

A row in `boxel_index` / `boxel_index_working` that ever held `has_error = true` carried that flag (and whatever `error_doc` it had — including the literal `jsonb null` shape) **forever**. Two interacting facts:

1. The table's primary key is `(url, realm_url, type)` — no `realm_version` — so a tombstone always collides with the prior row for the same URL.
2. The tombstone upsert's column list did **not** include `has_error` or `error_doc`. The ON CONFLICT SET clause generated by `upsertMultipleRows` only updates listed columns; any column not in the list keeps its previous value.

So once a row reached the `has_error = true, error_doc = jsonb null` shape (for any reason — a transient empty-error payload from an upstream construction site, a deleted file whose `tryToVisit` 404 was silently logged, an early write that pre-dated a later normalization fix), every subsequent reindex re-tombstoned the URL but never cleaned up the stale error fields. The row was invisible to the UI and to DB triage because nothing in `error_doc` named the underlying failure.

Observed on staging in `ctse/personal` for `FancyBrandGuide/c28b185c-...json` — same URL, same null shape, four separate reindex passes in a row, including one user-triggered priority-10 reindex that I watched land in real time.

## Confidence the fix is right

The fix is two lines of explicit defaults on a tombstone upsert. The behavior of the upsert helper is symmetrical for any column added to its list: `INSERT ... ON CONFLICT DO UPDATE SET col = EXCLUDED.col`. So adding `has_error` (with value `false`) and `error_doc` (with value `null`) means a tombstone always lands with cleared error fields, regardless of what was there before.

The write-boundary invariant is a guard against a class of regression rather than a fix for a specific bug, but it cooperates with the post-fatal-child finalize path: a throw at `updateEntry` reaches the worker's per-job reservation cap and the job rejects, rather than the silent thrash the cap fix was designed to terminate.

## Test plan

- [ ] `pnpm test` in `packages/realm-server` exercises the two new tests in `indexing-test.ts`:
  - `batch invalidation clears has_error and error_doc when tombstoning a previously-errored row` — plants an errored row, runs a tombstone batch, asserts `is_deleted = true`, `has_error = false`, `error_doc = null`.
  - `updateEntry refuses to write an instance-error entry with no error.message` — covers `undefined`, `{}`, missing-message, and empty-string-message shapes.
- [ ] After deploy, watch staging `ctse/personal`'s `boxel_index_working` for the FancyBrandGuide row. On the next reindex it should clear to `has_error = false, error_doc = null` (since the file is still missing — it becomes a clean tombstone).

### DIFF
--- a/packages/realm-server/tests/indexing-test.ts
+++ b/packages/realm-server/tests/indexing-test.ts
@@ -1436,6 +1436,110 @@ module(basename(__filename), function () {
         );
       });
 
+      test('batch invalidation clears has_error and error_doc when tombstoning a previously-errored row', async function (assert) {
+        // The primary key is `(url, realm_url, type)` — no `realm_version` —
+        // so a tombstone upsert always collides with the prior row for the
+        // same URL. Any column NOT in the tombstone upsert's SET list keeps
+        // its previous value. Before this guard, `has_error` and `error_doc`
+        // were not in that list, so an errored row stayed errored across
+        // every subsequent reindex even after the file was deleted —
+        // producing a "has_error = true, error_doc = jsonb null" shape
+        // that propagates forever and is invisible to UI / DB triage.
+        let mangoURL = new URL(`${testRealm}mango.json`);
+
+        // 1. Plant an error row in boxel_index by writing an
+        //    instance-error entry and committing the batch.
+        let errorBatch = await new IndexWriter(testDbAdapter).createBatch(
+          new URL(realm.url),
+        );
+        await errorBatch.updateEntry(mangoURL, {
+          type: 'instance-error',
+          error: {
+            id: mangoURL.href,
+            status: 500,
+            title: 'synthetic test error',
+            message: 'synthetic test error to verify tombstone clears it',
+            additionalErrors: null,
+          },
+        });
+        await errorBatch.done();
+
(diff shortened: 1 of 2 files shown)


habdelra:
the lint error is fixed in https://github.com/cardstack/boxel/pull/4887

habdelra:
oof, lots of errors like this:

 `[realm-server-error] mount fail url=https://localhost:4205/test/ kind=bootstrap pinned=true duration_ms=1 reason=this.isWorldReadable is not a function
failed to mount realm for request https://localhost:4205/test/module.gts: this.isWorldReadable is not a function`.

 I think these are upstream issues

lukemelia (approved):

PR #4889, opened 2026-05-19
CS-11178: invalidate cached RealmInfo on _permissions PATCH

lukemelia:
## Summary
- Follow-up to [CS-11126](https://linear.app/cardstack/issue/CS-11126) and PR #4868. CS-11126 dropped `Realm.visibility()` / `isWorldReadable()` per-instance caches so a permissions PATCH on replica A took effect on replica B without restart. But `RealmInfo.visibility` is **separately** memoized into `#cachedRealmInfo` by `parseRealmInfo()`, consumed by every card-document response via `attachRealmInfo()` and folded into the card-JSON ETag hash. `patchRealmPermissions` wrote `realm_permissions` without invalidating that cache — same multi-replica staleness pattern, surviving one layer up.
- Fix: `patchRealmPermissions` now calls `clearRealmIndexCachesAndBroadcast()` after `insertPermissions` succeeds, reusing the existing `realm_index_updated` NOTIFY channel whose listener already drops `#cachedRealmInfo` on every peer.
- Trade-off: this over-invalidates `#inFlightSearch` (cleared as part of the same broadcast). Permission PATCHes are admin-rare, so at-most-one-extra DB round-trip per concurrent `searchCards` is negligible. If we ever need to invalidate permissions-derived state without touching index-derived state, we'd introduce a dedicated `realm_permissions_changed` channel — YAGNI until then. Channel docstrings updated to acknowledge the broadened trigger set.

## Linear
[CS-11178](https://linear.app/cardstack/issue/CS-11178/patchrealmpermissions-leaves-cached-realminfovisibility-stale-across)

## Stacks on
This PR is based on #4887 (`fix-isworldreadable-removed-from-startup`), which removes a dangling `await this.isWorldReadable()` left over from CS-11126's merge. Lint:types fails without #4887, so this PR cannot land first. Will rebase onto main once #4887 merges, or this can merge naturally afterward.

## Test plan
- [x] `pnpm --filter @cardstack/realm-server lint:types`
- [x] `pnpm --filter @cardstack/runtime-common lint:types`
- [x] `pnpm --filter @cardstack/realm-server test-module` with `TEST_MODULE=realm-endpoints/permissions-test.ts` (12/12 pass, including the new `PATCH /_permissions invalidates cached RealmInfo and broadcasts to peers`)
- [ ] CI green

## New test
`PATCH /_permissions invalidates cached RealmInfo and broadcasts to peers` exercises both halves end-to-end:

1. Primes `#cachedRealmInfo` via `getRealmInfo()`, PATCHes a `*: read` grant, asserts the next `getRealmInfo()` reports `'public'` — proves the local cache invalidated. Without the fix, `getRealmInfo()` returns the memoized `'shared'` value.
2. Subscribes to `realm_index_updated` before the PATCH and asserts a NOTIFY with this realm's URL lands — proves the broadcast half ran. Without this, a refactor that swaps `clearRealmIndexCachesAndBroadcast()` for the bare `clearRealmIndexCaches()` would silently break cross-replica fan-out and only the local assertion would still pass.

Dispatch on the receiver side is covered separately by `realm-index-updated-listener-test.ts`.

### DIFF
--- a/packages/realm-server/tests/realm-endpoints/permissions-test.ts
+++ b/packages/realm-server/tests/realm-endpoints/permissions-test.ts
@@ -12,6 +12,7 @@ import {
   testRealmHref,
   testRealmURL,
   createJWT,
+  waitUntil,
 } from '../helpers';
 import '@cardstack/runtime-common/helpers/code-equality-assertion';
 import type { PgAdapter } from '@cardstack/postgres';
@@ -399,12 +400,18 @@ module(`realm-endpoints/${basename(__filename)}`, function () {
       // Dispatch on the receiver side is covered separately in
       // `realm-index-updated-listener-test.ts`.
       test('PATCH /_permissions invalidates cached RealmInfo and broadcasts to peers', async function (assert) {
-        let notifyPayloads: string[] = [];
+        // Only count NOTIFYs whose payload matches this realm — the
+        // `realm_index_updated` channel is shared across every realm
+        // mounted in this process, and the test fixture's own boot
+        // sequence (or any racing index swap) can fan out unrelated
+        // payloads on the same listener. Filtering by `testRealm.url`
+        // keeps the assertion specific to the PATCH under test.
+        let sawNotifyForThisRealm = false;
         let subscription = await dbAdapter.subscribe(
           REALM_INDEX_UPDATED_CHANNEL,
           (notification) => {
-            if (notification.payload) {
-              notifyPayloads.push(notification.payload);
+            if (notification.payload === testRealm.url) {
+              sawNotifyForThisRealm = true;
             }
           },
         );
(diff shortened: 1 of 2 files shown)


lukemelia:
Addressed both Copilot suggestions in 7a3b8396ee:

1. **Filtered the NOTIFY subscriber** to only count payloads matching `testRealm.url`, so unrelated `realm_index_updated` traffic (fixture boot, racing index swaps) can't introduce flakes. The assertion now reads a single `sawNotifyForThisRealm` boolean instead of `deepEqual`-ing the full received list.
2. **Replaced the hand-rolled polling loop with `waitUntil()`** from `tests/helpers/index.ts` — same shape as every other "wait for an async signal" site in the suite, with a `timeoutMessage` so a missed NOTIFY surfaces a clear timeout error rather than a confusing `assert.ok(false)`.

Also rebased onto `main` now that #4887 has merged, so this PR is targeting `main` directly. Re-ran the test module locally: 12/12 pass including the new test.

The earlier `Matrix Client Tests (1, 3)` failure (`download-realm.spec.ts:57` — `readableStreamImpl._read: Test ended.`) is an unrelated Playwright cleanup race — that spec doesn't touch `_permissions`, and the parent PR #4887 (identical infra minus this PR's two-file change) passed the same shard. Should re-run cleanly on the new commit.

habdelra (approved):

PR #4891, opened 2026-05-19
JobScopedSearchCache: store serialized response bytes

habdelra:
## Summary

Refactors `JobScopedSearchCache` to store the serialized JSON response body (`string`) instead of the parsed document.

- `getOrPopulate` loses its generic and is now fixed to `Promise<string>`.
- Both handlers (`handle-search.ts` and `handle-search-prerendered.ts`) move their `JSON.stringify(..., null, 2)` into the `populate` thunk — cache hits ship the cached bytes directly via `new Response(body, ...)`.
- `CachedEntry.result` tightens from `unknown` → `string`; header comment updated.

This removes the per-hit `JSON.stringify` on the search hot path. With ~358 federated-search calls per piranha-class from-scratch reindex, that's hundreds of large-doc stringify passes saved across one job. It also lays the value-type groundwork for the planned DB-backed cache swap — the migration becomes "swap the storage backend" rather than "redesign the marshalling layer."

PR1 of 2 for [CS-11176](https://linear.app/cardstack/issue/CS-11176/refactor-jobscopedsearchcache-to-string-blob-storage-job-id-based). PR2 (job-id-based ETag + If-None-Match) stacks on this.

## Test plan

- [x] Existing `job-scoped-search-cache-test.ts` suite passes after switching fixtures from parsed-doc objects to JSON strings
- [x] Added explicit test: hit returns the originally-serialized bytes (`call-1`) even when populate would have returned `call-2` on a re-run — catches accidental re-serialization
- [x] Targeted prettier + eslint clean on changed files
- [ ] CI green

### DIFF
--- a/packages/realm-server/handlers/handle-search-prerendered.ts
+++ b/packages/realm-server/handlers/handle-search-prerendered.ts
@@ -56,11 +56,15 @@ export default function handleSearchPrerendered(opts?: {
       cardUrls: parsed.cardUrls,
       renderType: parsed.renderType,
     };
-    let runSearch = () =>
-      searchPrerenderedRealms(
-        realmList.map((realmURL) => realmByURL.get(realmURL)),
-        parsed.cardsQuery,
-        searchOpts,
+    let runSearch = async () =>
+      JSON.stringify(
+        await searchPrerenderedRealms(
+          realmList.map((realmURL) => realmByURL.get(realmURL)),
+          parsed.cardsQuery,
+          searchOpts,
+        ),
+        null,
+        2,
       );
 
     // Symmetric to `_federated-search`'s gating. Cache is consulted
@@ -92,7 +96,7 @@ export default function handleSearchPrerendered(opts?: {
       : null;
     let cacheable = searchCache && jobId && consumingRealm;
 
-    let combined = cacheable
+    let body = cacheable
       ? await searchCache!.getOrPopulate({
           jobId: jobId!,
           realms: realmList,
@@ -104,7 +108,7 @@ export default function handleSearchPrerendered(opts?: {
 
     await setContextResponse(
       ctxt,
-      new Response(JSON.stringify(combined, null, 2), {
+      new Response(body, {
         headers: { 'content-type': SupportedMimeType.CardJson },
       }),
     );
--- a/packages/realm-server/job-scoped-search-cache.ts
+++ b/packages/realm-server/job-scoped-search-cache.ts
@@ -30,14 +30,19 @@ const DEFAULT_TTL_MS = 10 * 60 * 1000;
 const DEFAULT_MAX_ENTRIES = 5000;
 
 type CachedEntry = {
-  // Result is stored opaquely so both `_federated-search`'s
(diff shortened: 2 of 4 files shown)


packages/realm-server/handlers/handle-search-prerendered.ts
```diff
@@ -92,7 +96,7 @@ export default function handleSearchPrerendered(opts?: {
       : null;
     let cacheable = searchCache && jobId && consumingRealm;
 
-    let combined = cacheable
+    let body = cacheable
       ? await searchCache!.getOrPopulate({
```
habdelra, line 100: I'm surprised that the type signature for searchCache didn't have to change. we are storing a fundamentally different thing in it, how come the type didn't change?

packages/realm-server/job-scoped-search-cache.ts
```diff
@@ -30,14 +30,19 @@ const DEFAULT_TTL_MS = 10 * 60 * 1000;
...
+  // `JSON.stringify(..., null, 2)`, so the cache holds the exact bytes
+  // shipped to the next caller — no re-stringify on hit, no parsed
+  // object graph kept alive for the TTL window. Both
+  // `_federated-search` (`LinkableCollectionDocument`) and
+  // `_federated-search-prerendered` (`PrerenderedCardCollectionDocument`)
+  // share this single instance: inner-key canonicalisation already
+  // includes the endpoint-distinguishing params (htmlFormat /
+  // cardUrls / renderType pass through `opts`), so two endpoints'
+  // entries cannot collide on a key they don't both fully share.
+  result: string;
```
habdelra, line 45: ha, never mind--i see the type changed here

packages/realm-server/handlers/handle-search-prerendered.ts
habdelra: [Claude Code 🤖] The handler-facing seam (the `JobScopedSearchCache` class as an object) is unchanged — the type that actually narrowed is `getOrPopulate`'s generic-vs-fixed return shape (`<T>` → `string`) and `CachedEntry.result: unknown` → `string` inside the class. The handler just holds a reference to the class instance, so its `opts?: { searchCache?: JobScopedSearchCache }` signature stays the same.

What changed in the handler is the assignment site: `let body = ...` is now `string` instead of `combined: LinkableCollectionDocument`, and the populate thunk wraps `runSearch` with `JSON.stringify(..., null, 2)` so the cache stores bytes. The class's *value-type* tightening flows through to the call site by inference, no signature edit needed at the boundary.

packages/realm-server/job-scoped-search-cache.ts
habdelra: [Claude Code 🤖] Confirmed — `CachedEntry.result: unknown` → `string` on the class, and `getOrPopulate<T>` lost its generic and is fixed to `Promise<string>`. The handler boundary stays unchanged because it consumes the class as an object reference, not a value type.

lukemelia (approved):

PR #4911, opened 2026-05-21
Smooth memory baseline over last 5 builds

lukemelia:
## Summary

Single-build memory measurements are flaky enough that one noisy run can either trip the PR check on an unrelated branch or, worse, lock in a runaway baseline once it lands on main. Switch to a rolling-window mean over the last 5 builds so a one-build spike no longer dominates.

- `update-memory-baseline.mjs`: each module now retains a `samples` array (newest last, capped at 5). `delta_mb` is the mean of those samples. Modules absent from the current reports keep their prior entries unchanged (existing missing-shard behavior is preserved).
- `check-memory-baseline.mjs`: comparison computes the mean from `samples` when present, falling back to `delta_mb` so the script keeps working against v1 baselines until the first main run upgrades them. Adds a "Recent samples" column so reviewers can see whether a flagged value is a one-off outlier or a sustained trend.
- Baseline schema bumps to version 2 and gains `samplesWindow: 5`. The existing `packages/host/memory-baseline.json` is left as-is — the first post-merge main run will populate `samples` naturally (the prior `delta_mb` becomes a one-sample history that the new reading appends onto). Full smoothing is reached after 5 main runs.

## Why this matters

Recent flakes that this change would have softened:
- PR #4908 saw `Integration | ai-assistant-panel | file-attachment` jump 7.0 → 102.2 MB on a diff that didn't touch that area. A 5-build mean would absorb the spike instead of either failing the PR or, post-merge, pinning the baseline at 102 MB.
- The current main (`838df0e620` "Change CardDef.id to be RRI") tripped a +56 MB / +177% failure on `Acceptance | host submode`. Smoothing would let two more boring runs pull the baseline back down without manual intervention.

## Test plan

- [x] Smoke test locally: 6 simulated runs against a v1 baseline → migration produces `samples = [old_delta_mb]` on first run, grows to 5 then rolls.
- [x] Smoke test: PR-style check against v2 baseline reports "Baseline (mean of last 5)" header + recent samples column.
- [x] Smoke test: PR-style check against legacy v1 baseline (no `samples`, no `samplesWindow`) still works; falls back to `delta_mb`, header reads "Baseline".
- [x] `pnpm --filter @cardstack/host run lint` clean.
- [ ] Confirm in CI that the first main run after merge produces a v2 baseline diff (samples populated for every module).

### DIFF
--- a/packages/host/scripts/check-memory-baseline.mjs
+++ b/packages/host/scripts/check-memory-baseline.mjs
@@ -44,6 +44,23 @@ const HARD_ABSOLUTE_MB = 50;
 
 const escapeMd = (s) => String(s).replace(/\|/g, '\\|');
 
+// Compare against the rolling-window mean of recent baseline samples so a
+// single noisy build can't trip — or anchor — the budget. Fall back to the
+// pre-rolling-window `delta_mb` shape so this script keeps working against
+// older baseline files until the next main run upgrades them in place.
+const baselineDelta = (entry) => {
+  if (Array.isArray(entry?.samples) && entry.samples.length > 0) {
+    const sum = entry.samples.reduce((a, b) => a + b, 0);
+    return sum / entry.samples.length;
+  }
+  return entry?.delta_mb;
+};
+
+const fmtSamples = (entry) =>
+  Array.isArray(entry?.samples) && entry.samples.length > 0
+    ? `[${entry.samples.map((s) => s.toFixed(1)).join(', ')}]`
+    : null;
+
 const warnings = [];
 const failures = [];
 const newModules = [];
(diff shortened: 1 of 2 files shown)


backspace (approved):

habdelra:
why do we even try to capture a memory baseline autoamtically? this can still drift slowly overtime in a bad direction, and nothing will get flagged as the cumulative effect of that builds up. the way i'm approaching baselines is to measure it once and hold the thresholds constant so there is no drift. and if we need to readjust its a conscious effort.

lukemelia:
@habdelra good point. however, these baselines are per test module, so measuring "once" would be an ongoing project, since test modules get added, removed, added to, and trimmed down

PR #4912, opened 2026-05-21
feat(skills): include all skills from boxel-skills (CS-11195)

FadhlanR:
## Summary

- Remove `ALLOWLIST` from `packages/boxel-cli/scripts/build-skills.ts` so cards in the pinned `cardstack/boxel-skills` tag are automatically packaged. Refactor the script around a pure `computeEmissionPlan(cards)` so the dedup logic is unit-testable, and add `updatePluginReadme()` which keeps a fenced auto-gen block in `plugin/README.md` in sync with the emission set.
- Curate via an explicit `EXCLUDED_IDS` set: the `boxel-environment` aggregator, all 9 of its references, and the two host-only top-level `env-*` leaves (`env-creating-and-editing-cards`, `env-sim-boxel-environment-guide`) are dropped. They teach the boxel **host AI assistant's** tool-call JSON shape (`{name, payload: {description, attributes}}`) and reference host-only commands (`switch-submode_dd88`, `set-active-llm_1887`, `patch-fields_3e67`) that don't exist in Claude Code / Codex. The exclusion is explicit (no `env-*` prefix rule), so future env-* upstream additions still ship by default.
- Manifest-based stale-dir sweep (addresses Codex P2 on `build-skills.ts`): each run records emitted IDs to `packages/boxel-cli/scripts/.boxel-skills-manifest.json`; on the next run, `plugin/skills/<id>` dirs that dropped out of the emission plan are deleted. CLI-authored skill dirs are never written by the script and so are never in the manifest — the sweep can't touch them.
- Soften the previous throw-on-unknown-`kind` to a warn-and-skip so a future additive card type upstream does not brick the build.
- Final emit set at `v0.0.22`: 7 upstream skills — `boxel-design`, `boxel-development`, `boxel-ui-guidelines`, `catalog-listing`, `dev-bfm-syntax`, `dev-file-def`, `dev-markdown-format`.
- New: `packages/boxel-cli/tests/scripts/build-skills.test.ts` covering the emission-plan dedup, sort, unsupported-kind handling, description fallback chain, `isExcluded` policy (including a regression test that future env-* IDs are NOT silently dropped), and the `loadManifest` / `computeStaleIds` sweep helpers.

Closes CS-11195.

## Test plan

- [x] `pnpm build:skills` in `packages/boxel-cli` exits 0 and prints `Done. Emitted 7 skill(s) from boxel-skills@v0.0.22.`
- [x] Second `pnpm build:skills` run is idempotent — no `removed …` lines from the stale-dir sweep.
- [x] `pnpm lint` clean (both `lint:types` and `lint:js`).
- [x] `pnpm exec vitest run tests/scripts/build-skills.test.ts` — 17 unit tests pass.
- [ ] CI: `boxel-cli-pr-title.yml` accepts the `feat(skills):` prefix; on merge, `boxel-cli-publish.yml` regenerates skills and bumps `plugin.json` to a minor version.
- [ ] Smoke: `claude --plugin-dir packages/boxel-cli/plugin` and type `/boxel-cli:` to confirm the new skill IDs (e.g. `dev-bfm-syntax`, `dev-file-def`, `catalog-listing`) tab-complete and that `boxel-environment` does NOT appear.


jurgenwerk (approved):
Is the "env" prefix needed? There's a bunch of skills like

env-assistant-persona.md
env-calling-commands,...

Wondering if "env" is necessary

FadhlanR:
@jurgenwerk — addressed in d585dcf. You were right that the `env-` prefixed ones felt off. The audit showed they all live under the `boxel-environment` aggregator upstream and teach the boxel **host AI assistant's** runtime — its tool-call JSON shape (`{name, payload: {description, attributes}}`), host commands like `switch-submode_dd88` / `set-active-llm_1887` / `patch-fields_3e67`, and chat `roomId` context. None of that exists when boxel-cli is consumed by Claude Code or Codex, so shipping those skills would just mislead the assistant.

The commit drops the `boxel-environment` aggregator, all 9 of its references, and the two host-only top-level leaves (`env-creating-and-editing-cards`, `env-sim-boxel-environment-guide`). Final emit set is 7 upstream skills.

The exclusion is an explicit `Set` in `build-skills.ts`, **not** a `env-*` prefix rule — so if someone adds a genuinely CLI-friendly `env-*` skill upstream later, it'll ship by default.

Also picked up @chatgpt-codex-connector's P2 in the same commit: added a manifest at `scripts/.boxel-skills-manifest.json` so future `pnpm build:skills` runs delete `plugin/skills/<id>` dirs that drop out of the emission plan.

Skipping the Copilot inline typo / markdown comments — those are in generated files and any fix belongs upstream in `cardstack/boxel-skills`.

PR #4914, opened 2026-05-21
feat: add self contained test-running ability to boxel-cli

jurgenwerk:
## Summary

`boxel test` is now self-contained and **local-only by default** — no realm-server, no Vite dev server, no push step between writing code and running tests. The factory's agent loop tightens from `write → realm push → boxel test` to just `write → boxel test`, saving a tool-call turn per validation cycle.

### What ships

- **Local-mode execution** (`boxel test` with no args): the CLI starts an in-process HTTP server that hosts both the test page (from `bundled-test-harness/`) and the workspace's realm modules. `.gts` / `.ts` files are transpiled on demand via `runtime-common`'s `transpileJS`. Workspace cards are read from cwd; base + skills realm sources are vendored under `bundled-realms/`.
- **Bundled test harness**: `pnpm build:test-harness` copies the host's dev-mode `dist/` into `bundled-test-harness/`, dropping only sourcemaps. ~60 MB unpacked / ~15 MB compressed. See "Why we ship the whole host dist" below for the rationale.
- **Single merged server**: the asset server (test page + host bundle chunks) and the workspace transpiling server are one `http.createServer` listening on a random localhost port. Same origin for everything chromium fetches; module mounts at `/workspace/`, `/base/`, `/skills/`, the host bundle assets at `/assets/...`, and the test-page HTML at `/`.
- **Remote-realm fallback**: `boxel test --realm <url>` opts back into the older remote-realm flow for testing cards already on a published realm.

### Before vs after

| Aspect | Pre-CS-11164 | After this PR |
|---|---|---|
| Where harness comes from | Sibling `packages/host/dist/` (monorepo layout required) | `bundled-test-harness/` shipped in the npm tarball |
| `--realm` flag on `boxel test` | Required | Optional (defaults to local mode) |
| Realm-server needed at test time | Yes | No (in local mode) |
| `boxel realm push` before each test iteration | Required | Eliminated in local mode |
| Workable outside the monorepo | No | Yes |
| Agent tool-call turns per test iteration | 2 (push + test) | 1 (test) |

### Installation footprint

| Step | Lands on disk |
|---|---|
| `npm install -g @cardstack/boxel-cli` | ~75 MB CLI (CLI bundle + bundled-test-harness + bundled-realms + bundled-types) |
| User runs `boxel test` for the first time | + ~150 MB chromium (via the separate `npx playwright install chromium` step the CLI's error message prompts; not auto-installed) |
| Total for a `boxel test` user | ~225 MB |

### Why we ship the whole host dist (and not a pruned slim subset)

Spent meaningful time on this during the branch and rejected three different slim approaches. Recording the reasoning so future-us doesn't repeat it:

**Attempt 1 — captured manifest of "what chromium actually fetches"**. Ran a card test with a network-request recorder, committed the resulting list of paths, had `build-test-harness.ts` copy only those files. Got the harness from 60 MB → 27 MB. **Rejected** because:
- The committed manifest entries are content-hashed (`runtime-common-Czdr1nwr.js`). Any host content change re-hashes filenames, the committed paths 404, and the slim bundle silently misses chunks.
- "Regenerate the manifest before publishing" is a manual step. Per-branch policy, automation-or-bust: tooling can't depend on humans remembering.
- CI-side automation (capture-on-publish, Vite-manifest static graph, etc.) all add real moving parts for a single-digit-MB win in absolute terms.

**Attempt 2 — bisecting minimizer over the captured set**. Wrote a script that removed each captured entry, re-ran the test, kept the removal only if tests stayed green. ~232 entries. After 15 minutes of running and ~30 KEEP/REMOVED decisions, observed a **~7% success rate** — almost everything tried as "obviously UI" (submode-switcher, modal-container, AI-assistant chunks, code-mode commands, etc.) was actually needed at runtime. Investigated why: it's **Embroider's `@embroider/virtual/compat-modules` auto-glob**. The host's `app/app.ts` does `Resolver.withModules(compatModules)`, which materializes every `.gts` under `app/services/`, `app/components/`, `app/commands/`, `app/routes/`. The Ember resolver eagerly imports them at app boot — not because the test code uses them, but because the resolver walks the keyspace and pre-touches each entry. Removing any auto-globbed chunk fails the boot chain even though the test never executes that code. **Rejected** because the only way past the floor is to refactor the host (make `monaco-service`, `ai-assistant-panel-service`, etc. truly lazy / dynamic-import) — multi-day Embroider work that the ticket explicitly flagged as the bulk of the effort, and out of scope for what this PR is doing.

**Attempt 3 — runtime fetch from GitHub Release**. Split the harness into its own `host-test-harness-v*` release, downloaded on first `boxel test` per CLI version, cached under `~/.cache/boxel-cli/`. Got the CLI tarball down to ~3 MB. **Rejected** because:
- The host build still has to happen *somewhere* (the harness release workflow). Same coupling, just relocated.
- Adds a fetcher + cache + sha256-verify + pin file + a new publish workflow + a "bump the pin" PR dance — meaningful infrastructure for a 60→3 MB tarball win when the user pays the difference under `~/.cache/` anyway.
- First-run failure modes (offline, GH down, sha256 mismatch) are new surfaces with no equivalent in the bundled approach.

So we ship the full ~60 MB. The size context, honestly: a slim cut would save ~30 MB on disk for `boxel test` users. They're already paying 150 MB for chromium one-time, and the CLI install dominates only briefly until chromium downloads, after which it's ~25% of the total. Not nothing, but not worth the staleness/maintenance/infra cost the slim approaches require — and the only real way to a 5–10 MB harness is the host-side lazy-loading refactor anyway.

### Files & flow

- New: `packages/boxel-cli/scripts/build-test-harness.ts` — copies `packages/host/dist/` minus sourcemaps into `bundled-test-harness/`. Refuses to proceed if `tests/index.html` is missing (catches accidental production-mode host builds).
- New: `packages/boxel-cli/scripts/build-realms.ts` — vendors `packages/base/` and `packages/skills-realm/contents/` into `bundled-realms/`. Skills realm is optional (its contents are gitignored; CI doesn't have it; runtime never reaches `/skills/` from a card test).
- Changed: `packages/boxel-cli/src/commands/test.ts` — adds `runTestsLocally`; `--realm` becomes opt-in; the test-page server gains realm-mount routing and on-demand transpile.
- Changed: `packages/boxel-cli/scripts/build.ts` — copies `content_tag_bg.wasm` next to `dist/index.js` post-build (esbuild inlines content-tag's wasm loader; the loader's `readFileSync` resolves against the CLI's dist dir).
- Changed: `.github/workflows/boxel-cli-publish.yml` — adds `pnpm --filter @cardstack/host build` before the CLI build (the bundled-test-harness step reads host's dev dist).
- Changed: `.github/workflows/ci.yaml` — `boxel-cli-build` job now depends on `test-web-assets` so the cached host dist is restored before the CLI build runs.
- Changed: `packages/boxel-cli/package.json` — `@playwright/test` moves to `dependencies` (was `devDependencies`) so the headless-browser driver ships with a published install. `bundled-test-harness/`, `bundled-realms/` added to `files`. Chromium itself is *not* auto-downloaded on install; users run `npx playwright install chromium` once.
- Changed: runbook + `software-factory-operations` skill — rewritten around the local-mode default.

### Other architecture decisions

- **Local server transpiles workspace cards in-process** rather than pushing them to a realm-server first. Removes a tool-call turn per agent validation cycle.
- **One HTTP server, not two**: the workspace module mounts share the same origin as the test page so chromium needs no CORS preflights and the loader's URL-mapping works out of the box.
- **Realm-server still required for `boxel realm pull`** — the user syncs once from a realm-server (local, staging, or prod), then tests entirely locally. Validates the agent-token-efficiency goal: one command per validation cycle after the initial sync.

[CS-11164 in Linear](https://linear.app/cardstack/issue/CS-11164/bundle-a-self-contained-qunit-test-harness-into-boxel-cli)

## Test plan

- [x] `pnpm build` produces ~60 MB `bundled-test-harness/`, ~1.7 MB `bundled-realms/`, ~2.4 MB CLI bundle.
- [x] `pnpm lint` (js + types) clean.
- [x] Local smoke: `cd /tmp/cs-11164-check && boxel test` from a workspace pulled via `boxel realm pull` (no monorepo on PATH, no realm-server running, no Vite) → 11 passed, 0 failed (~3 s).
- [x] Published-install smoke: `pnpm pack` + `npm install -g <tarball>`, `which boxel` resolves to the npm-installed binary, `boxel test` from the same workspace → same result. Tarball ~18 MB.
- [x] `--debug` shows the merged-server origin handling correctly and only emits chromium console + page errors.
- [x] `BOXEL_TEST_HARNESS_DIR=<path>` env override and `--host-dist-dir <path>` flag still resolve to the right dist.
- [ ] CI green.


jurgenwerk:
My idea was to try to slim down the build needed for the harness but doing this looks like a significant endeavor in the embroider territory, so I opted into simply bundling the development host build into the cli (the description has info on the CLI size impact) that has the test runner and related support.

This is hard because a card test imports Application, which imports compat-modules, which has 138 top-level static import statements for every file under app/. Removing any one of those chunks causes the ES module graph to abort at app.ts evaluation — before Ember boots, before QUnit loads, before any test code runs. 

Claude recommends we could do this:
> Don't use compat-modules. Replace it in a test-only app-test.ts with a manually-curated registry: { 
  'services/loader-service': LoaderService, 'services/realm-info-service': RealmInfoService, ... } — only what the test path actually looks up at runtime
  
I could try this, but perhaps bundling the host development dist into boxel cli for purposes of running the tests isn't such a big deal? On the other hand it's a lot of bloat, one other option I could think of is to publish the build as a package and let the boxel-cli download (and cache it) it only when running `boxel test`.

packages/boxel-cli/scripts/build-test-harness.ts
```diff
@@ -0,0 +1,93 @@
...
+ * (CS-11164).
+ *
+ * Copies the dev-mode host build wholesale, dropping only sourcemaps.
+ * Tried a manifest-driven slim cut (~27 MB instead of ~60 MB), but
+ * the manifest needed manual regen whenever the host's chunks
+ * re-hashed, and we don't want correctness to depend on human
+ * upkeep — see `feedback_no_manual_maintenance` in memory. The full
+ * dist is ~60 MB unpacked and ~15 MB on the wire; next to the
+ * 150 MB Playwright chromium install every `boxel test` user
+ * already needs, it's noise.
```
habdelra, line 14: that's an interesting approach. like we chatted about the other day, it would be interesting if in rthe future we could spearate this into its own npm package (boxel-cli-test-support), that would prompt the user to install if you try to run the test validations.

habdelra (approved):

packages/boxel-cli/scripts/build-test-harness.ts
jurgenwerk: Made a new ticket for that: https://linear.app/cardstack/issue/CS-11258/split-boxel-cli-test-support-into-its-own-npm-package

PR #4938, opened 2026-05-22
Pin file-api as a runtime dep of file extracts (flaky test fix)

habdelra:
## Problem

`prerendering-test.ts > prerender - non-mutating tests > runner behavior > file prerender returns extracted metadata` intermittently fails on the assertion:

```
deps include base file-api module
  actual  : false
  expected: true
```

Most recent failure: https://github.com/cardstack/boxel/actions/runs/26289527361/job/77386270791

The file-extract pipeline only natively imports `${baseRealm.url}card-api` (because `baseFileRef.module` resolves there — that's where the `FileDef` class actually lives). The public `${baseRealm.url}file-api` URL — which is what the indexer keys on when invalidating file extracts — landed in `deps` only because matrix-service's `fileAPIModule = importResource(() => '…/file-api')` happened to call `loader.import('file-api')` synchronously inside the active `withRuntimeDependencyTrackingContext` window.

Whether that import landed inside the window was a microtask-scheduling race. It was reliable on a fresh page but flaky after a `BrowserManager.restartBrowser()` (e.g., from a preceding test like `concurrent restart coalescing`), where the matrix-service resource fires either too early (before the route starts the session) or too late (after `extract()` resolves).

## Prior art

PR #4875 attempted to fix this by pinning the imports in `RenderRoute.#buildModel`, but that ran the imports for **every** render — card prerenders too — and was reverted because it introduced cross-test instability.

## Solution

Pin `loader.import('${baseRealm.url}file-api')` **only inside** the file-extract route's `withRuntimeDependencyTrackingContext`. This makes the dep deterministic for file extracts and is invisible to the card-prerender path, so it can't reintroduce the destabilisation that closed #4875.

Adds a `console.warn` diagnostic that fires if `file-api` is still missing from the snapshot after the explicit import — insurance in case the tracker session is ever closed early or the URL is rewritten.

## Files

- `packages/host/app/routes/render/file-extract.ts`

## Test plan
- [ ] CI passes on this branch (specifically the Realm Server Tests shards)
- [ ] No card-prerender regressions in `prerendering-test.ts` (the broader test module)

### DIFF
--- a/packages/host/app/routes/render/file-extract.ts
+++ b/packages/host/app/routes/render/file-extract.ts
@@ -7,6 +7,7 @@ import { isTesting } from '@embroider/macros';
 
 import {
   baseFileRef,
+  baseRealm,
   formattedError,
   snapshotRuntimeDependencies,
   withRuntimeDependencyTrackingContext,
@@ -97,6 +98,7 @@ export default class RenderFileExtractRoute extends Route<Model> {
       contentSize,
       buildError: this.#buildError.bind(this),
     });
+    let fileApiURL = `${baseRealm.url}file-api`;
     let result: FileDefExtractResult;
     try {
       result = await withRuntimeDependencyTrackingContext(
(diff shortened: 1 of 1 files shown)


habdelra:
Closing — moving the same fix into #4934 directly. The failure is deterministic on that branch (4/4 recent RS shard 1 runs failed on `prerendering-test.ts > file prerender returns extracted metadata`), not a transient flake, so it makes sense to land the fix in the same PR rather than as a standalone.

PR #4948, opened 2026-05-25
fix setsid: command not found on macs when running mise run dev-all

jurgenwerk:
Closes [CS-11260](https://linear.app/cardstack/issue/CS-11260/mise-run-dev-all-prints-setsid-command-not-found-on-macos-cleanup).

## Summary

`mise run dev-all` on macOS prints:

```
/.../mise-tasks/lib/dev-common.sh: line 123: setsid: command not found
```

…on every invocation. `setsid` is part of Linux's util-linux package and isn't in the macOS base system. The dev stack still comes up, but the `spawn_cleanup_guardian` line fails, so the detached safety-net process that should fire if dev-all dies without its bash trap (SIGKILL, terminal session collapse) never starts. Subsequent `mise run dev-all` then hits `EADDRINUSE` on the leftover ports.

## Fix

Fall back to `nohup` when `setsid` isn't on PATH. `nohup` is POSIX-standard (`/usr/bin/nohup` on macOS) and gives us the property we actually need here — SIGHUP immunity so the guardian survives the controlling shell exiting. The "new session" semantics that `setsid` adds aren't actually required for this guardian.

The outer `</dev/null >/dev/null 2>&1` on the nohup branch suppresses nohup's pre-script "appending output to nohup.out" warning; the inner `exec >>'$_scg_log' 2>&1` then routes the guardian's actual output to the audit log file the same way as before. Linux/CI behavior is unchanged — `setsid` is still preferred when available.

## Test plan

- [x] On macOS: `mise run dev-all` no longer prints `setsid: command not found`.
- [x] On macOS: a `sh -c …` guardian process is spawned and writes its initial "watching dev-all pid …" line to `$BOXEL_DEV_ALL_GUARDIAN_LOG` (default `/tmp/boxel-dev-all-guardian.log`).
- [x] On macOS: killing the parent pid causes the guardian to wake up at next poll (≤1s) and run the cleanup branch.
- [ ] On Linux/CI: `setsid` path is still taken (`command -v setsid` succeeds → first branch).
- [ ] Manual verification that the trap path (normal Ctrl-C) still cleans up — the guardian is only a safety net for the trap-can't-fire scenarios.

### DIFF
--- a/mise-tasks/lib/dev-common.sh
+++ b/mise-tasks/lib/dev-common.sh
@@ -105,10 +105,20 @@ kill_from_pidfile() {
 # matters when the trap is denied a chance to fire.
 #
 # Implementation notes:
-#   - `setsid` puts the guardian in its own session so it doesn't get
-#     SIGHUP'd when dev-all's session dies.
+#   - `setsid` (Linux) puts the guardian in its own session so it
+#     doesn't get SIGHUP'd when dev-all's session dies. macOS / BSD
+#     userland doesn't ship setsid; `nohup` is the POSIX-standard
+#     equivalent for this purpose — it makes the child ignore SIGHUP
+#     so the guardian survives the same terminal-collapse path the
+#     setsid version protects against. We don't actually need a new
+#     session; we only need SIGHUP immunity.
 #   - stdin/stdout/stderr are redirected away from the terminal so the
-#     guardian can survive after the user's shell exits.
+#     guardian can survive after the user's shell exits. The outer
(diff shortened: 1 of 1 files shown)


backspace:
I have something similar to this [in this PR](https://github.com/cardstack/boxel/pull/4903/changes#diff-c150be3494a356a07e18a764179b5814cf2ba077b1229bfc0b2abea4f379584c)

lukemelia:
Should we close in favor of @backspace's already merged PR?

jurgenwerk:
Closing in favor of https://github.com/cardstack/boxel/pull/4903

PR #4960, opened 2026-05-25
Plant linksTo error sentinels in the data bucket on fetch failure

habdelra:
Stacked on #4957 (CS-11217), which is stacked on #4953 (CS-11215). Merge base is `worktree-cs-11217-prerender-sentinel-scan`.

## What

`lazilyLoadLink`'s failure path emits structured `LinkErrorValue` / `LinkNotFoundValue` sentinels (HTTP 404 → `link-not-found`, every other failure → `link-error`). `errorDoc` is the `SerializedError` shape the legacy `boxel-render-error` payload used.

Both singular and plural failures land the sentinel on a per-instance side-channel and leave the data bucket itself at its legacy shape:

```ts
(instance as any)[field.name] = null;             // legacy null-write through
                                                  // the setter — every
                                                  // bucket-driven consumer
                                                  // (serialize, queryable,
                                                  // field-support `getter`,
                                                  // peekAtField, autoSave
                                                  // notify chain) sees what
                                                  // they always saw on
                                                  // failure
setLinkSentinel(instance, field.name, sentinel);  // structured failure on
                                                  // the new per-instance
                                                  // side-channel
```

`field-support.ts` exposes the side-channel:

```ts
export function setLinkSentinel(instance, fieldName, sentinel): void
export function getLinkSentinel(instance, fieldName): LinkSentinel | undefined
export function clearLinkSentinel(instance, fieldName): void
```

CS-11224 (Relationship API), CS-11219 / CS-11220 (phantom value), and CS-11225 / CS-11226 (placeholder render) consume from here.

### Why the side-channel and not the bucket

Two reasons.

For the singular case, putting the sentinel directly into the bucket changes what every existing bucket consumer (`peekAtField`, the field-support `getter`, `relationshipMeta`, `LinksTo.serialize` / `queryableValue`, the searchDoc loop, and the autoSave path's `notifySubscribers` → `onInstanceUpdated` → `doAutoSave` → `drainAutoSaveQueue` chain) sees during the broken card's render. The autoSave chain in particular reads `store.peek(id)` at `drainAutoSaveQueue` time and decides save-or-skip on whether the result is a `CardDef` or an `error_doc` — and the bucket's contribution to that decision races the realm's `error_doc` publish in the operator-mode editor flow.

For the plural case, a `linksToMany` field's bucket value is a `WatchedArray` (defined in `packages/base/watched-array.ts`) — a `Proxy`-wrapped array whose `set` trap fires on any index assignment, slice mutation, or length change. Writing a sentinel as an array element (`value[index] = sentinel`) lets the renderer iterate into a slot that isn't a `CardDef`, which trips a template render error inside the `friends`-array sub-component. That render failure severs Glimmer's tracking chain back into `code-submode.gts`'s `cardError` getter, so the error-state surface in operator-mode stops re-rendering when the realm's `error_doc` arrives.

The side-channel keeps the bucket shape byte-for-byte identical to `main` for both cases, so both races resolve the way they always have, while still recording the structured failure where new consumers can find it.

### Out of scope

- The legacy `boxel-render-error` dispatch in `card-api.gts` stays. Both host listeners (`card-prerender.gts`, `routes/render.ts`) stay. The dispatch is still the active producer of indexer-side `error_doc`s. CS-11229 retires it.
- `scanForBrokenLinks` reads only from the bucket. The side-channel read is deferred to CS-11229 — surfacing it now would route a second `#processRenderError` on top of the dispatch.
- Singular `LinksTo.getter` / `LinksTo.serialize` / `LinksTo.queryableValue` / `relationshipMeta` are at parent (`#4957`). The bucket never holds a sentinel in this PR for either arity. CS-11224 owns the structured Relationship API replacement.

## Test plan

- [x] 14/14 in `packages/host/tests/acceptance/code-submode-test.ts` "editor tests" pass locally.
- [x] 15/15 in `packages/host/tests/integration/field-support-predicates-test.gts` pass locally.
- [x] `pnpm lint:types` (host package) clean for touched files.
- [x] Unit coverage in `packages/host/tests/integration/field-support-predicates-test.gts`:
  - `setLinkSentinel` / `getLinkSentinel` / `clearLinkSentinel` round-trip on the side-channel without disturbing the bucket
  - `linksToMany` getter does not retrigger load on a terminal-sentinel array element
- [x] Bucket-side scanner tests at parent cover singular + plural sentinel-in-array shapes.
- [ ] End-to-end indexer-side `error_doc` capture flows through the legacy dispatch — `realm-indexing-test.gts` "can recover from indexing a card with a broken link" exercises it; relying on CI.

Linear: CS-11218.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -1690,6 +1690,9 @@ class LinksToMany<FieldT extends LinkableDefConstructor> implements Field<
             `the linksToMany field '${this.name}' contains a primitive card '${instance.name}'`,
           );
         }
+        if (isNotLoadedValue(instance)) {
+          return { id: instance.reference };
+        }
         if (isLinkError(instance) || isLinkNotFound(instance)) {
           // Terminal sentinels are searchDoc-invisible (matches legacy null
           // bucket behaviour). Round-trip handling is CS-11223.
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -165,7 +165,6 @@ import {
   isCardOrField,
   isLinkError,
   isLinkNotFound,
-  isNonPresentLink,
   isNotLoadedValue,
   notifyCardTracking,
   peekAtField,
@@ -1226,8 +1225,11 @@ class LinksTo<CardT extends LinkableDefConstructor> implements Field<CardT> {
     if (instance == null) {
       return null;
     }
-    if (isNonPresentLink(instance)) {
-      return { id: instance.reference };
+    if (isLinkError(instance) || isLinkNotFound(instance)) {
+      // Terminal sentinels surface in searchDoc the same way the legacy
+      // null bucket value did — as an absent field. `CS-11223` will revisit
+      // serialization round-trip; for now, broken links are searchDoc-invisible.
+      return null;
     }
     return this.card[queryableValue](instance, stack);
   }
@@ -1245,14 +1247,7 @@ class LinksTo<CardT extends LinkableDefConstructor> implements Field<CardT> {
     let relationshipType = isFileDef(this.card)
       ? FileMetaResourceType
       : CardResourceType;
-    if (isNonPresentLink(value)) {
-      // Terminal Error/NotFound sentinels serialize the same way a
-      // NotLoadedValue does — the persisted relationship keeps the
-      // reference so a reload deserializes back to NotLoaded and the lazy
-      // loader gets a fresh chance to resolve it (or re-plant the
-      // sentinel if the underlying failure is still present). The
-      // structured errorDoc is in-memory diagnostic state, not part of
-      // the saved card document.
+    if (isNotLoadedValue(value)) {
       return {
         relationships: {
           [this.name]: {
@@ -1260,6 +1255,22 @@ class LinksTo<CardT extends LinkableDefConstructor> implements Field<CardT> {
               self: makeRelativeURL(value.reference, opts),
             },
             data: { type: relationshipType, id: value.reference },
+          },
+        },
+      };
+    }
+    if (isLinkError(value) || isLinkNotFound(value)) {
+      // Terminal Error/NotFound sentinels serialize to a null link —
+      // matching the legacy bucket-was-null behaviour. The structured
+      // errorDoc is in-memory diagnostic state, not part of the saved
+      // card document. Preserving the broken reference in the persisted
+      // relationship is what `CS-11223` (serialization round-trip) is
+      // about; until then, save drops the reference and reload reads as
+      // an empty link.
+      return {
+        relationships: {
+          [this.name]: {
+            links: { self: null },
           },
         },
       };
@@ -1711,8 +1722,10 @@ class LinksToMany<FieldT extends LinkableDefConstructor> implements Field<
             `the linksToMany field '${this.name}' contains a primitive card '${instance.name}'`,
           );
         }
-        if (isNonPresentLink(instance)) {
-          return { id: instance.reference };
+        if (isLinkError(instance) || isLinkNotFound(instance)) {
+          // Terminal sentinels are searchDoc-invisible (matches legacy null
+          // bucket behaviour). Round-trip handling is CS-11223.
+          return null;
         }
         return this.card[queryableValue](instance, stack);
       })
@@ -1770,16 +1783,22 @@ class LinksToMany<FieldT extends LinkableDefConstructor> implements Field<
         };
         return;
       }
-      if (isNonPresentLink(value)) {
-        // Terminal Error/NotFound sentinels (planted by lazilyLoadLink on
-        // fetch failure) serialize the same way NotLoadedValue does: keep
-        // the reference so reload deserializes back to NotLoaded and the
-        // lazy loader retries. The structured errorDoc is in-memory only.
+      if (isNotLoadedValue(value)) {
         relationships[`${this.name}\.${i}`] = {
           links: {
             self: makeRelativeURL(value.reference, opts),
           },
           data: { type: relationshipType, id: value.reference },
+        };
+        return;
+      }
+      if (isLinkError(value) || isLinkNotFound(value)) {
+        // Terminal sentinels in a linksToMany array serialize as a null
+        // slot — same behaviour the legacy `null` bucket write produced.
+        // Round-trip preservation of broken-link references is CS-11223.
+        relationships[`${this.name}\.${i}`] = {
+          links: { self: null },
+          data: null,
         };
         return;
       }
@@ -2309,7 +2328,7 @@ export class BaseDef {
                 }) ?? null,
             ];
           }
-          if (isNonPresentLink(rawValue)) {
+          if (isNotLoadedValue(rawValue)) {
             let normalizedId = rawValue.reference;
             if (value[relativeTo]) {
               normalizedId = resolveCardReference(
@@ -2318,6 +2337,12 @@ export class BaseDef {
               );
             }
             return [fieldName, { id: makeAbsoluteURL(rawValue.reference) }];
+          }
+          if (isLinkError(rawValue) || isLinkNotFound(rawValue)) {
+            // Terminal sentinels emit `null` in searchDoc — matching the
+            // legacy null-bucket behaviour. `CS-11223` will revisit when
+            // broken-link references should round-trip.
+            return [fieldName, null];
           }
           // Reuse the value we already peeked above instead of re-reading
           // through the descriptor — for computed fields the descriptor
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -172,6 +172,7 @@ import {
   realmContext,
   relationshipMeta,
   setFieldDescription,
+  setLinkSentinel,
   setRealmContextOnField,
   type ComputePassSnapshot,
   type LinkErrorValue,
@@ -1199,17 +1200,10 @@ class LinksTo<CardT extends LinkableDefConstructor> implements Field<CardT> {
       return value;
     }
 
-    let maybeSentinel = deserialized.get(this.name);
-    if (isNotLoadedValue(maybeSentinel)) {
-      lazilyLoadLink(instance, this, maybeSentinel.reference);
+    let maybeNotLoaded = deserialized.get(this.name);
+    if (isNotLoadedValue(maybeNotLoaded)) {
+      lazilyLoadLink(instance, this, maybeNotLoaded.reference);
       return undefined;
-    }
-    if (isLinkError(maybeSentinel) || isLinkNotFound(maybeSentinel)) {
-      // Terminal sentinels — the fetch already resolved into a structured
-      // failure. Returning null here matches the public contract that a
-      // broken linksTo reads as empty, and we deliberately do not call
-      // lazilyLoadLink so a permanently-broken link does not spin.
-      return null as unknown as BaseInstanceType<CardT>;
     }
     let value = getter(instance, this);
     trackRuntimeRelationshipDependency(value, this.card);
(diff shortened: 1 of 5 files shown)


packages/base/card-api.gts
```diff
@@ -3389,18 +3409,44 @@ function lazilyLoadLink(
...
+          let notLoadedRef = resolveCardReference(
+            item.reference,
+            instance.id ?? instance[relativeTo],
+          );
+          if (reference === notLoadedRef) {
+            value[index] = sentinel;
+          }
+        }
+        notifyCardTracking(instance);
+        notifySubscribers(instance, field.name, value);
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed — `value` here is the WatchedArray sourced from `LinksToMany.emptyValue` (the linksToMany getter passes it through as `pluralArgs.value`), so `value[index] = sentinel` already schedules the `notifySubscribers` + `notifyCardTracking` microtask via the proxy `set` trap. Dropped the explicit pair in 8fc33c2 (well, in ffeaef4 — the dedup commit landed before the fileRender fix). Now matches the per-index success branch right above, which writes without an explicit notify for the same reason.

habdelra:
[Claude Code 🤖] @chatgpt-codex-connector good catch on the fileRender pass. Fixed in 8fc33c2 — the same `#brokenLinkPayload()` scan + throw runs after the isolated render in the fileRender try block, and the existing catch transforms the thrown JSON payload into a `file-error` the same way cardRender transforms it into `instance-error`.

habdelra:
[Claude Code 🤖] Converting to draft — the sentinel-write producer change is causing a host-tests cascade that I can't sort out with CI iteration alone.

## What's red on `a902c38`

- 5 Host Tests shards: 2, 6, 11, 16, 19. Down from 8 shards (1, 2, 3, 4, 6, 13, 16, 18) on the initial push, so each revert helped, but not enough.
- Realm Performance Benchmark: `Issues/sticky-note` returns 500 with `Cannot read properties of undefined (reading 'startsWith')` — server-side stack not in the GH logs.
- Realm Server Tests (4, 6).

## Iteration timeline

1. `6267430` — initial: remove `boxel-render-error` dispatch + listeners, plant `link-error` / `link-not-found` sentinels, teach singular getter + serialize + queryable + relationshipMeta about sentinels. 8 host shards red.
2. `ffeaef4` — drop redundant notify pair (Copilot's catch — the WatchedArray proxy already schedules them).
3. `8fc33c2` — add the `#brokenLinkPayload` scan to the fileRender pass (Codex's catch).
4. `f0c449b` — revert the fileRender scan as a debug move; no change in failing-shard set.
5. `9326699` — restore the legacy `boxel-render-error` dispatch in `card-api.gts` and the host listeners in `card-prerender.gts` + `routes/render.ts` alongside the new sentinel write. 4 shards turned green (1, 4, 13, 18), but 5 still red (2, 3, 6, 16, 11).
6. `a902c38` — realign sentinel `serialize` / `queryableValue` / linksToMany serialize / `CardDef.[queryableValue]` searchDoc loop to match the legacy null-bucket shape (broken-link becomes `links: { self: null }`, `null` in searchDoc). One more shard turned green; 5 still red.

## What the failure looks like

Shard 6 / "allows fixing broken cards" — the test edits Monaco with a fixed link and expects the file content saved is the user's input verbatim. With my changes, the file ends up re-serialized through `card-api` (verbose `cardInfo`, defaulted `address.shippingInfo`, relativised URLs). That happens when `canSyncWithStore` returns true, which happens when `store.peek(<url>)` returns a `CardDef` instance instead of an error doc.

So the question is: why is the in-browser store holding a `CardDef` instance for the broken card after my changes, when the legacy behaviour ended up with the realm's `error_doc` in the store? The indexer **does** still emit the error (the indexing summary in shard 6's log reports `instanceErrors: 1, fileErrors: 4`), and `current reference count for ... is negative: -1` is showing up in the browser console — that's a store invariant violation pointing somewhere into the load lifecycle.

## Things I haven't fully ruled out

- The `LinksTo.getter` early-return for terminal sentinels (`null as unknown as BaseInstanceType<CardT>`) might be giving in-browser render paths a "successfully-loaded with null pet" answer where the legacy path stayed in an in-flight state long enough for the realm's `error_doc` to win the race. Reverting the getter change would crash other code because the bucket holds a sentinel object, so this needs a more careful redesign.
- `notifySubscribers(instance, field.name, sentinel)` may be passing the sentinel into a change subscriber (`store.ts:1730` wires `onInstanceUpdated` through it). Legacy passed `null`. A subscriber that branches on `value.id` or similar could behave differently.
- The Performance Benchmark failure on `Issues/sticky-note` has a healthy card erroring with `Cannot read properties of undefined (reading 'startsWith')` — server-side stack not visible in CI logs. I'd want a local repro before guessing.

## Next moves I'd suggest

I'm out of cheap CI iterations here. The cleanest path is probably one of:

1. Split this PR further: land just the bucket-sentinel **type/predicate plumbing** + a no-op producer (bucket still gets null) so consumers can be migrated incrementally. The actual sentinel emit becomes its own ticket once `CS-11225` (placeholder render) and `CS-11229` (host consumer migration) are sequenced.
2. Spend a session locally repro'ing the cascade with diagnostic logging on the realm's index path — the reference-count-negative invariant violation is a real bug surfacing here, not noise.
3. Treat the failing shards as expected and update test expectations as part of this PR (would expand scope significantly).

Holding off on more CI iterations until you weigh in.

habdelra:
[Claude Code 🤖] Re-pushed at `1f0029e` with the producer split described in the updated PR description. Holding draft until CI is green.

packages/base/field-support.ts
```diff
@@ -68,6 +68,21 @@ const fieldOverrides = initSharedState(
...
 );
+// Per-instance side-channel for terminal linksTo failures planted by
+// `lazilyLoadLink`'s catch block. The data bucket stays at its legacy
+// null on fetch failure — every existing consumer of the bucket (serialize,
+// queryableValue, the field-support `getter`, `peekAtField`, the in-browser
+// autoSave path that branches on bucket-derived signals, …) sees the same
+// shape it did on `main`, so the cascade triggered by sentinel-in-bucket
+// (Integration | code submode | editor tests) goes away. The structured
+// failure lives here instead, keyed by (instance, fieldName). `scanForBrokenLinks`
+// reads from this map; CS-11223 will revisit whether a bucket-storage
```
habdelra, line 79: make sure to update CS-11223 to have it revisit this chunk of code

habdelra: [Claude Code 🤖] Adding a comment on CS-11223 now to flag the `linkSentinels` side-channel as the chunk to revisit when the round-trip story comes through — if/when bucket-storage becomes reachable post-CS-11229, this is the producer that needs to be migrated off the side-channel.

packages/base/field-support.ts
```diff
@@ -496,10 +552,19 @@ export interface BrokenLinkFinding {
...
+// declaration for why the bucket is unchanged for the singular case.
+// This scanner deliberately reads ONLY from the bucket for now: the
+// legacy `boxel-render-error` dispatch is still the active producer of
+// indexer-side error_docs, and surfacing the side-channel here too
+// would route a second error through `#processRenderError`, racing the
+// dispatch path and destabilising other in-flight realm-index
+// transitions. CS-11217's scanner is therefore intentionally dormant
+// for the singular case in this transitional PR; downstream tickets
+// (CS-11229 and friends) will migrate consumers off the dispatch and
+// re-enable the side-channel read here.
```
habdelra, line 567: make sure to update CS-11229 to revisit this chunk of code

habdelra: [Claude Code 🤖] My CS-11229 comment now lists this exact chunk (the `scanForBrokenLinks` bucket-only read) as the chunk to flip when the dispatch comes out — point 2 in the comment over there.

packages/base/field-support.ts
```diff
@@ -68,6 +68,21 @@ const fieldOverrides = initSharedState(
...
+// autoSave path that branches on bucket-derived signals, …) sees the same
+// shape it did on `main`, so the cascade triggered by sentinel-in-bucket
+// (Integration | code submode | editor tests) goes away. The structured
+// failure lives here instead, keyed by (instance, fieldName). `scanForBrokenLinks`
+// reads from this map; CS-11223 will revisit whether a bucket-storage
+// approach is reachable once the consumer set has been migrated to the
+// public Relationship API (CS-11224).
+const linkSentinels = initSharedState(
+  'linkSentinels',
+  () => new WeakMap<BaseDef, Map<string, LinkErrorValue | LinkNotFoundValue>>(),
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] 

Real concern, but no side-channel consumer in this PR observes the loss yet — `scanForBrokenLinks` reads only from the bucket, and the only test exercising `getLinkSentinel` is the singular round-trip. Per-index sentinel state for `linksToMany` belongs with the Relationship API redesign (CS-11224), which is the ticket that introduces the first real consumer and will pick the storage shape (Map<index, sentinel> or similar). Deferring there so the storage decision lives with the consumer that constrains it.

packages/base/card-api.gts
```diff
@@ -3389,14 +3399,23 @@ function lazilyLoadLink(
...
+        ? { type: 'link-not-found', reference, errorDoc }
+        : { type: 'link-error', reference, errorDoc };
+      // Two writes: the null-write through the setter keeps every
+      // bucket-driven consumer (serialize, queryable, the field-support
+      // getter, peekAtField, the autoSave notify chain) on a value
+      // shape they already understand; `setLinkSentinel` records the
+      // structured failure on a per-instance side-channel for consumers
+      // that need the typed link-error/link-not-found form.
+      (instance as any)[field.name] = null;
+      setLinkSentinel(instance, field.name, sentinel);
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] 

Real concern, but only observable once a consumer reads from the side-channel — no production consumer exists in this PR. Singular-success clearing is one line; plural-success clearing is structurally wrong (a sibling slot recovering would erase a still-valid sentinel for another broken index in the same field), so it has to wait for the per-index storage shape that CS-11224 chooses. Bundling both into CS-11224 so the lifecycle (plant + clear) lives next to the consumer that depends on it.

habdelra:
[Claude Code 🤖] FYI — design correction landed at the project level (see [CS-11218 comment](https://linear.app/cardstack/issue/CS-11218) and the [project-level comment](https://linear.app/cardstack/project/48112016-cd4b-4246-8688-4fd0374ca754)). This PR is compatible with the corrected design: the sentinel is hidden from userland via the `linkSentinels` side-channel, and the data bucket stays at `null` so existing consumers are undisturbed. No phantom anywhere.

One divergence from the reframed ticket text: under the corrected design the getter is supposed to return `undefined` for broken-link states (matching today's not-loaded getter behavior) — this PR keeps the field at literal `null` (legacy bucket write). Both keep the sentinel hidden; the difference is whether userland's `card.brokenLink === null` discriminates NotSet from BrokenLink.

- Current PR: `card.brokenLink === null` is true for NotSet *and* BrokenLink (indistinguishable at the access boundary; use `getRelationship` to discriminate).
- Reframed ticket: `card.brokenLink === null` only for NotSet; `=== undefined` for BrokenLink (matches not-loaded behavior).

Not blocking — either is valid under the "hidden sentinel" rule. Calling it out so the PR author can either keep the current `null` semantics and update the ticket acceptance, or pivot the getter return.

habdelra:
[Claude Code 🤖] Closing — design correction makes the consumer-side reads in this PR out of order against the dependency graph.

The PR has two halves:

- **Producer side** — `lazilyLoadLink`'s catch block writes `LinkErrorValue` / `LinkNotFoundValue` via `setLinkSentinel` (per-instance side-channel) and stops writing `field = null`. The CustomEvent dispatch is intentionally kept during the transition. This half is on the right track.
- **Consumer side** — `LinksToMany.queryableValue` and `LinksToMany.serialize` add `isLinkError(value) / isLinkNotFound(value)` checks against the raw bucket value, and `scanForBrokenLinks` walks the data bucket directly. Under the corrected design these reads must go through `getRelationship` — the bucket / side-channel is read only by the Relationship API; every other consumer switches on `getRelationship(...).kind`.

The work is still needed. After CS-11224 (`getRelationship` v1) lands, the producer side will be re-filed essentially as-is, and the consumer-side reads in `queryableValue` / `serialize` / `scanForBrokenLinks` will be re-implemented to route through `getRelationship`.

PR #4963, opened 2026-05-26
Fix Monaco workers failing in cross-origin deployments 

jurgenwerk:
image

## Summary

In deployed environments the host bundle (e.g. `boxel-host.stack.cards`) is served from a different origin than the realm-server page that boots it (`realms.stack.cards`). Vite's `?worker` import emits a Worker constructor pointed at an absolute cross-origin URL, and `new Worker(crossOriginUrl)` is forbidden by the browser. As a result Monaco's JSON, TS, CSS, and HTML language services fail to spin up — silently degrading code-mode (no JSON diagnostics on `.json` card-instance files, no TS IntelliSense on `.gts`, no formatting) and surfacing a `TypeError: Cannot read properties of undefined (reading 'toUrl')` from Monaco's same-thread fallback path. Dev/CI is on `localhost` for both origins so the bug only manifests on staging/prod.

## Fix

`packages/host/app/services/monaco-service.ts`:

- Switch the 5 Monaco worker imports from `?worker` (Worker constructor) to `?worker&url` (URL string).
- Route them through a new `makeMonacoWorker(url)` helper:
  - Same-origin: `new Worker(url)` as before — keeps dev/CI behavior identical.
  - Cross-origin: wrap the URL in a same-origin `Blob` that immediately `importScripts(...)` the real worker code. Workers spawned from a Blob URL adopt the page's origin, and `importScripts` is allowed to fetch cross-origin without CORS, so the bundled worker code still loads.

No vite-config or realm-server changes required — the existing `experimental.renderBuiltUrl` still resolves the worker URL against `__boxelAssetsURL`, and the helper handles the cross-origin case from that.

Linear: https://linear.app/cardstack/issue/CS-11095/monaco-workers-fail-in-deployments-due-to-cross-origin-host-bundle

## Test plan

- [ ] `pnpm run lint:types` clean (verified locally).
- [ ] Deploy to staging and open a `.json` card-instance file in code mode.
- [ ] Browser console: no `Failed to construct 'Worker'` warnings, no `FileAccessImpl.toUri` TypeErrors.
- [ ] JSON diagnostics fire on a `.json` card instance, TS IntelliSense fires on a `.gts` file, formatting works.
- [ ] Dev/CI behavior unchanged (same-origin path — verify host tests still pass).


jurgenwerk:
After deploying this branch to staging- I confirm this is fixed and the error is gone.

habdelra (approved):

PR #4966, opened 2026-05-26
CS-11264: route _realm-auth through reconciler.lookupOrMount

lukemelia:
## Summary

- After Phase 3 lazy mount ([CS-10894](https://linear.app/cardstack/issue/CS-10894)), `_realm-auth` was the missing caller for non-pinned realms. The handler iterated `realms[]` directly to resolve each permission row's URL — but non-pinned (source/published) realms aren't in `realms[]` until something has driven a request-path mount. After a realm-server restart, an owner's first `boxel-cli` push/publish/sync therefore failed with `No realm token available` even though `realm_user_permissions` was intact. See [CS-11264](https://linear.app/cardstack/issue/CS-11264).
- Switch the handler to validate URLs against the registry (`reconciler.knownByUrl` + a single batched probe against `realm_registry` for URLs not yet reflected) and issue a JWT directly — **without** mounting. `fetchUserPermissions(onlyOwnRealms: false)` returns every `'*'`-readable realm, so routing each row through `reconciler.lookupOrMount` would cold-mount the entire accessible public-readable set on the first post-restart `boxel realm list` / host login. A JWT does not need a mounted realm; the mount happens later, on-demand, when the holder hits a per-realm endpoint and `findOrMountRealm` / `lookupOrMount` runs there.
- Session room is now resolved **once** per request (it is keyed by `matrixUserId` in the DB, not by realm). The create branch uses the realm-server's matrix client to call `createDM`, mirroring how `_server-session` creates session rooms — the caller has already exchanged an OpenID token via `_server-session` to get the JWT they're presenting here, so the server client is already logged in.
- Regression test in `realm-auth-test.ts` evicts the test realm from both `realms[]` and `reconciler.mounted`, asserts (via preconditions) that the row is still in `reconciler.knownByUrl`, hits `/_realm-auth`, asserts a JWT is issued, and asserts `reconciler.mounted` is still empty for the URL afterwards — proving the handler does NOT cold-mount.

## Test plan

- [x] `pnpm lint:types` clean
- [ ] CI: `pnpm test --filter "realm auth handler"` — both existing and new regression test pass (local run was contended on test-pg / synapse; leaving the green check to CI)
- [ ] CI: `pnpm test --filter "Phase 3 lazy mount"` — 7/7 still pass
- [ ] Manual verification on staging once deployed:
  - restart `boxel-realm-server-staging`, then run the boxel-home PR-preview workflow against an existing non-pinned realm (e.g. `boxel-home-pr-58`) — should succeed without first touching the realm via the request path
  - separately confirm a logged-in user with broad public-realm access can call `/_realm-auth` (host login or `boxel realm list`) immediately after restart without observable mount delay

### DIFF
--- a/packages/realm-server/handlers/handle-realm-auth.ts
+++ b/packages/realm-server/handlers/handle-realm-auth.ts
@@ -4,6 +4,7 @@ import type { CreateRoutesArgs } from '../routes';
 import {
   SupportedMimeType,
   fetchUserPermissions,
+  type Realm,
 } from '@cardstack/runtime-common';
 import type { RealmServerTokenClaim } from 'utils/jwt';
 import { getUserByMatrixUserId } from '@cardstack/billing/billing-queries';
@@ -14,11 +15,10 @@ import * as Sentry from '@sentry/node';
 export default function handleRealmAuth({
   dbAdapter,
   realmSecretSeed,
-  realms,
+  reconciler,
   serverURL,
 }: CreateRoutesArgs): (ctxt: Koa.Context, next: Koa.Next) => Promise<void> {
   return async function (ctxt: Koa.Context, _next: Koa.Next) {
-    let allRealms = realms;
     let token = ctxt.state.token as RealmServerTokenClaim;
     let { user: matrixUserId } = token;
     let user = await getUserByMatrixUserId(dbAdapter, matrixUserId);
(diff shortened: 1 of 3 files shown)


packages/realm-server/handlers/handle-realm-auth.ts
```diff
@@ -37,21 +43,100 @@ export default function handleRealmAuth({
...
+    // each row through reconciler.lookupOrMount would cold-mount every
+    // public-readable realm on the server on the first post-restart
+    // _realm-auth call (boxel realm list, host login). Phase 3's lazy-
+    // mount contract is "mount on first per-realm request" — the per-
+    // realm JWT itself is fine to issue from registry presence alone;
+    // the mount happens later, when the holder actually hits a realm
+    // endpoint and findOrMountRealm/lookupOrMount runs there.
+    //
+    // Mirrors multiRealmAuthorization's three-step lookup order
+    // (CS-11238):
```
backspace, line 60: Do you want these phase and issue references?

backspace (approved):
I deployed this branch to staging and the [`boxel-home` CI job passed](https://github.com/cardstack/boxel-home/actions/runs/26400057717/job/77929515594), its earlier failure is what prompted this issue, so the fix works!

packages/realm-server/handlers/handle-realm-auth.ts
lukemelia: No, but I want to wait for another CI run less!

PR #4970, opened 2026-05-26
CS-11271: route _grafana-reindex through reconciler.lookupOrMount

lukemelia:
## Summary

- Same anti-pattern [CS-11264](https://linear.app/cardstack/issue/CS-11264) just fixed in `_realm-auth`. After Phase 3 lazy mount ([CS-10894](https://linear.app/cardstack/issue/CS-10894)), non-pinned realms are absent from `realms[]` after a realm-server restart. `handle-reindex.ts:39` did `realms.find(...)` to decide "does this realm exist on this server", so an operator-triggered `/_grafana-reindex` for a non-pinned realm returned `400 realm <url> does not exist on this server` until something else happened to lazy-mount the realm. See [CS-11271](https://linear.app/cardstack/issue/CS-11271).
- Route through `reconciler.lookupOrMount(realmURL)`. Pinned realms keep the O(1) `mounted` fast-path; non-pinned realms pay one cold-mount cost on first hit. Mount failures bubble as 500 via `sendResponseForSystemError` (same shape as the already-present catch around `reindex()`). `undefined` (URL absent from `realm_registry`) preserves the existing 400 + error message.
- Regression test in `maintenance-endpoints-test.ts` evicts the created realm from `realms[]`, hits `/_grafana-reindex`, and asserts the from-scratch-index job lands with `clearLastModified: true`. Confirmed it fails pre-fix.

## Rebase note

This branch adds `testingOnlyEvictRealmFromRealmsList(url)` to `RealmServer` — identical to the shim CS-11264 (#4966) introduces. Whichever branch lands first, the other will hit a trivial duplicate-add conflict on `server.ts`; resolution is a no-op (drop the duplicate).

## Test plan

- [x] `pnpm test --filter "can reindex a realm via grafana endpoint"` — both existing test and new regression test pass
- [x] Confirmed regression test fails when handler is reverted to `realms.find(...)`
- [x] `pnpm lint:types` clean
- [ ] Manual verification on staging once deployed: restart boxel-realm-server-staging, then trigger Grafana "Reindex realm" against a non-pinned source realm — should enqueue the job instead of 400ing

### DIFF
--- a/packages/realm-server/handlers/handle-reindex.ts
+++ b/packages/realm-server/handlers/handle-reindex.ts
@@ -22,7 +22,7 @@ export default function handleReindex({
   serverURL,
   dbAdapter,
   definitionLookup,
-  realms,
+  reconciler,
 }: CreateRoutesArgs): (ctxt: Koa.Context, next: Koa.Next) => Promise<void> {
   return async function (ctxt: Koa.Context, _next: Koa.Next) {
     let realmPath = ctxt.URL.searchParams.get('realm')?.replace(/\/$/, '');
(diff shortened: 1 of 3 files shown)


backspace (approved):

habdelra (approved):

habdelra:
can we audit our realm server handlers for more isntances where they reach for `realms` passed into the `handleXXX()` function? Perhaps the actual issue is that we should stop passing in `realms` into the `handleXXX()` functions--this is probably always wrong.

lukemelia:
> can we audit our realm server handlers for more isntances where they reach for realms passed into the handleXXX() function? Perhaps the actual issue is that we should stop passing in realms into the handleXXX() functions--this is probably always wrong.

Yes, that's how I found this case actually.

habdelra:
> > can we audit our realm server handlers for more isntances where they reach for realms passed into the handleXXX() function? Perhaps the actual issue is that we should stop passing in realms into the handleXXX() functions--this is probably always wrong.
> 
> Yes, that's how I found this case actually.

do you know of a case where the realms that we pass into the handlers has a legitimate use?

PR #4973, opened 2026-05-26
Add realm config templates

backspace:
This adds templates and other changes for managing host routing.

The card chooser won’t let you choose from another realm (`lockConsumingRealm`/`lockSelectedRealms`):

CleanShot 2026-06-08 at 15 09 09@2x

There‘s a warning when there’s a duplicate path or invalid segment:

CleanShot 2026-06-08 at 15 08 08@2x

The interact template for the realm config shows the routes and what cards they route to:

CleanShot 2026-06-08 at 15 10 36@2x

There’s an option in the realm’s menu in the chooser:

CleanShot 2026-06-09 at 14 30 31@2x


lukemelia (approved):
When testing in the preview, the Link Card button didn't seem to limit the card choice to the current realm. If that's expected, then everything else looks good.

backspace:
> When testing in the preview, the Link Card button didn't seem to limit the card choice to the current realm. If that's expected, then everything else looks good.

Did the preview even show the custom editor? I’d expect it not to because it involves `base` changes.

PR #4998, opened 2026-05-27
Surface settled-state and pending test-waiter names in host timeout diagnostics

habdelra:
## Background and Goal

A host acceptance test (`code submode | editor tests > when the user lacks write permissions: the editor is read-only`) intermittently times out at the 140s QUnit limit with `in-flight fetches at rejection time: <none>`. With no outstanding network activity, the awaited `settled()` was blocked on something other than the network — a never-resolving test waiter, a lingering run loop, a pending timer, or an unfinished render — but the existing `[test-timeout]` diagnostic only reports fetch state, so it cannot say which. The host app registers many distinct named waiters (store, search, indexing, command, monaco, operator-mode, and more); any one stuck produces this exact symptom.

This is a diagnostics-only change: no production or test-waiter behavior is touched. The root cause is not yet reproducible, and pinning it down requires knowing which gate is still open when the timeout fires.

## Where to start

`packages/host/tests/helpers/setup.ts` — `summarizeSettledState()` and its use inside `logRejectionDiagnostics()`. That single function feeds all three diagnostic paths (`[test-timeout]`, `[test-unhandled-rejection]`, `[test-qunit-uncaught]`), so each now prints `getSettledState()` metrics plus, when a waiter is pending, the waiter name and its captured begin-async origin from `getPendingWaiterState()`.

## Key decisions

- Report **named** waiters via `@ember/test-waiters` `getPendingWaiterState()` rather than just the boolean `hasPendingWaiters` — the name is the whole point, distinguishing a monaco-rendering stall from a store/indexing/search stall.
- Wrapped in try/catch and reused across all rejection paths so diagnostics can never mask the original failure.
- No waiter behavior changed: the existing begin/track pairings are intentionally left as-is until a real failure names the actual culprit.

### DIFF
--- a/packages/host/tests/helpers/setup.ts
+++ b/packages/host/tests/helpers/setup.ts
@@ -1,7 +1,10 @@
 /* eslint-disable @cardstack/host/wrapped-setup-helpers-only */
 // This is the one place we allow these to be used directly.
 
-import { settled } from '@ember/test-helpers';
+import { getSettledState, settled } from '@ember/test-helpers';
+
+import { getPendingWaiterState } from '@ember/test-waiters';
+import type { TestWaiterDebugInfo } from '@ember/test-waiters';
 
 import {
   setupApplicationTest as emberSetupApplicationTest,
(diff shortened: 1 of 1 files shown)


backspace (approved):
This is a great idea 🎉

lukemelia (approved):

PR #4999, opened 2026-05-27
Adorn hover bar: anchor left when it fits, clamp to right corner radius when it doesn't

lukemelia:
## Summary
The Adorn hover type-label tab now uses a two-phase anchor:

- While the label's natural width fits, it hugs the card's **top-left** corner with a 4px bleed that lines up with the selection stroke.
- When it doesn't fit, its **right edge** pins to the start of the card's top-right corner radius and the extra width spills off the card's **left** edge — so long card-type names never collide with the rounded corner.
- The label stays inside the visible stack-item frame: it flips below the card when there isn't room above, and truncates with an ellipsis when it can't fit at natural width without escaping the panel into surrounding chrome (operator-mode sidebar, dialog title bar). Within the panel, intra-panel overlap with sibling cards is allowed.

Resolves CS-11280.

## Implementation
- A `trackLabelOverflow` ember-modifier on the label computes position directly from the rendered card's rect and the visible stack-item frame's rect, then writes `left`, `top`, `max-width`, and a `data-side` attribute on the label inline.
- The boundary is `.stack-item-content` — the visible top-level frame for the panel the card is rendered on. Sibling cards / columns within that frame don't constrain the label; the chrome around it does.
- Anchor decision: short labels use `left = card.left − 4`. Long labels (natural width > `card.width − corner-radius + 4`) anchor their right edge to `card.right − (corner-radius − 4)` and grow leftward, with a 4px hysteresis so sub-pixel wobble can't flip the placement on every layout pass.
- Width: when the label's natural width fits the boundary, `max-width: max-content` so the browser sizes to the true intrinsic width (writing back the integer-rounded `scrollWidth` would shave a sub-pixel remainder and trip `text-overflow: ellipsis` even with room to spare). Only when the label actually has to clamp against the boundary do we write a measured pixel value, which is the case where ellipsis is the desired outcome.
- Side: prefer above; flip below when there's not enough room above the card inside the boundary. The `[data-side="bottom"]` attribute drives a CSS rule that mirrors the `clip-path` polygon vertically so the slope still points toward the card.
- The label is wrapped around `BoxelDropdown` with an inline-flex `<span>` so the dropdown's trigger and its portal-origin counts as a single flex item of the label — without it the label's natural width grows by one flex-gap when the menu opens (the open-state wormhole-origin becomes a flex item where the closed-state placeholder is `display: none`), shifting the label on every menu open.
- `autoUpdate` from `@floating-ui/dom` re-fires the position update on scroll, resize, and ancestor mutations; the placement math is direct (floating-ui's `flip` + `shift` + `size` middleware aren't a clean fit for the right-anchored-with-truncation pattern, and `limitShift` doesn't actually cap shift distance the way it reads).

## Tests
- Integration test `overlay-menu-items-test.gts` asserts both anchor phases (short → left-anchored at `card.left − 4`, long → right edge clamped to the corner-radius point and overflowing past `card.left`) and that the label stays inside the containing-card frame either way.

## Demo card
- `packages/experiments-realm/adorn-overflow-demo.gts` renders a 3-column kanban with deliberately long task-card type names so the long-label / flip-below / boundary-clamp behavior can be reproduced by hand.

## Test plan
- [x] Integration tests pass.
- [x] Manually verified on a card with a long type name that the tab grows rightward up to the corner radius and then spills off the left as the name lengthens further.
- [x] Manually verified that opening the 3-dot menu does not shift the label horizontally.
- [x] Manually verified that a label near the top of a panel flips below the card instead of intruding on the panel chrome.

https://github.com/user-attachments/assets/21be36c6-af99-4d6b-9b75-18ee7bdf72c3

### DIFF
--- a/packages/experiments-realm/AdornOverflowDemo/1.json
+++ b/packages/experiments-realm/AdornOverflowDemo/1.json
@@ -1,0 +1,56 @@
+{
+  "data": {
+    "type": "card",
+    "attributes": {
+      "cardInfo": {
+        "name": "Adorn Overflow Demo",
+        "notes": null,
+        "summary": "Kanban-style layout that reproduces leftward clipping of the Adorn hover label when a card sits flush against an overflow:hidden container.",
+        "cardThumbnailURL": null
+      }
+    },
+    "relationships": {
+      "backlog.0": {
+        "links": {
+          "self": "../ProjectInitiativeTask/1"
+        }
+      },
+      "backlog.1": {
+        "links": {
+          "self": "../ProjectInitiativeTask/2"
+        }
+      },
+      "inProgress.0": {
+        "links": {
+          "self": "../ProjectInitiativeTask/3"
+        }
+      },
+      "inProgress.1": {
+        "links": {
+          "self": "../ProjectInitiativeTask/4"
+        }
+      },
+      "done.0": {
+        "links": {
+          "self": "../ProjectInitiativeTask/5"
+        }
+      },
+      "done.1": {
+        "links": {
+          "self": "../ProjectInitiativeTask/6"
+        }
+      },
+      "cardInfo.theme": {
+        "links": {
+          "self": null
+        }
+      }
+    },
+    "meta": {
+      "adoptsFrom": {
+        "module": "../adorn-overflow-demo",
+        "name": "AdornOverflowDemo"
+      }
+    }
+  }
+}
--- a/packages/experiments-realm/ProjectInitiativeTask/1.json
+++ b/packages/experiments-realm/ProjectInitiativeTask/1.json
@@ -1,0 +1,14 @@
+{
+  "data": {
+    "type": "card",
+    "attributes": {
+      "title": "Refactor Auth Middleware"
+    },
+    "meta": {
+      "adoptsFrom": {
+        "module": "../adorn-overflow-demo",
+        "name": "ProjectInitiativeTask"
+      }
+    }
+  }
+}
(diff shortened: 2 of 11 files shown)


habdelra:
is there a scenario where the tile you are adorning is already on the left side of the isolated view, in that case wouldn't the text get cropped? it feels like you are relying on the geometry of cards grid specifically to prevent that from happening--but that's really just one particular way of rendering tiles such that you always have space on the left. Perhaps something like the kanban board view is not so generous with its left side. its not too hard to conceive of a card that has no space on the left for the adorn title to overflow.

lukemelia:
@habdelra thanks for pushing me on that. I've totally revised it to use floating-ui with middleware to handle these scenarios gracefully. Here's an updated video:


https://github.com/user-attachments/assets/4675e613-3a5b-4d80-80da-eca77713963d

habdelra (approved):

backspace:
> - Position is written as an inline `left` value (rather than `right: calc(...)`) so it doesn't shift when velcro's `autoUpdate` re-fires on DOM mutations — for example, when the 3-dot menu dropdown opens.

Is this true? Or am I misunderstanding?

CleanShot 2026-05-28 at 16 02 54

lukemelia:
@backspace the description was out-of-date (now updated) but the behavior you noted was a re-regression (now fixed). Thanks!

backspace (approved):
pending lint fix of course

PR #5041, opened 2026-05-29
dashboard: Fix reindexing of published realms

backspace:
CleanShot 2026-05-29 at 12 49 44

The redacted URL is a published realm, which doesn’t live under the realm server domain.

With this branch deployed, reindexing succeeded:

CleanShot 2026-05-29 at 14 49 24@2x


habdelra (approved):
i think just a lint problem

PR #5048, opened 2026-05-30
Redesign broken-link placeholder as a quiet box with a reveal overlay

habdelra:
## Background and Goal

A broken `linksTo` / `linksToMany` is read mostly by card *consumers*, who usually can't act on it. The old `BrokenLinkTemplate` rendered the URL, error message, stack, and additional errors inline at every footprint — visual noise in the reading flow.

This adopts the quieter treatment already shipped for markdown-embedded card references: a card-sized box with a faint diagonal cross and a centered link-off + type-name chip. The placeholder visual deliberately **matches the markdown broken-link design** (`packages/base/default-templates/markdown.gts`, the `.markdown-bfm-broken*` rules), so a broken reference looks the same whether it's in a markdown doc or a `linksTo` field. The box is **identical for `not-found` and `error`** — what went wrong surfaces only when the reader opens the reveal overlay.

## Demo

The reveal overlay opening at each edge of the card. Positioning is deliberately **sensitive to the card boundary** — the overlay stays inside the card (it's clipped by the card, keeps a margin from the edge, and flips above when there's no room below) rather than spilling out over the page.

https://github.com/user-attachments/assets/fbc0f54a-9e1c-4fe0-b9b2-ea528e2adebc

## Where to start

- `packages/base/default-templates/broken-link-template.gts` — the component. The placeholder box mirrors the `.markdown-bfm-broken*` rules in `markdown.gts`; below it is the reveal overlay.
- `packages/host/tests/integration/components/broken-link-template-test.gts` — asserts the box is identical across states, the detail is DOM-resident but hidden by default, and the reveal opens/closes.

The dispatch sites (`field-component.gts`, `links-to-editor.gts`, `links-to-many-component.gts`) are unchanged — the component contract is the same.

## Key decisions

- **The reveal is pure CSS that just shows DOM already on the page** — no JS, no re-render. The URL, message, stack, and additional errors are always in the DOM (`display: none` until opened), so a reader or AI consumer can recover the failure even while it's hidden. A hidden checkbox toggled by `<label>` triggers it: an "i" indicator opens it (right-of-label for the inline atom) and a close button in a fixed header dismisses it.
- **The overlay is bounded to the card, not the viewport.** It's an absolutely-positioned, card-clipped element (not a top-layer popover) anchored to the trigger, so it scrolls with the card, keeps a margin from the card edge, flips above when there's no room below, clips its scrollbar to the rounded frame, and stacks in front of every placeholder in the card. No card-boundary CSS is changed.


packages/base/default-templates/broken-link-template.gts
```diff
@@ -111,183 +138,338 @@ export default class BrokenLinkTemplate extends GlimmerComponent<{
...
+      .overlay {
+        display: none;
+        /* Absolutely positioned (NOT a top-layer popover) so the overlay is a
+           normal descendant: the card slot is its containing block and the
+           card's own overflow clips it, keeping it inside the card boundary
+           rather than spilling out over the page. */
+        position: absolute;
+        width: max-content;
+        max-width: min(20rem, 80cqw);
+        max-height: min(18rem, 80cqh);
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch, and confirmed — the placeholder renders outside the `.field-component-card` containers, so `cqw`/`cqh` were falling back to the viewport.

Fixed in 0bb2f36: the width now caps relative to the containing block (`min(20rem, 100%)`) and the height uses a fixed cap, so there is no viewport fallback. The card’s own `overflow: hidden` remains the hard bound, and the inner panel scrolls past the height cap.

packages/base/default-templates/broken-link-template.gts
```diff
@@ -109,301 +152,646 @@ export default class BrokenLinkTemplate extends GlimmerComponent<{
...
-        {{#if this.statusLabel}}
-          <div class='status' data-test-broken-link-status>
-            {{this.statusLabel}}
+            <path
+              class='warn-icon-tri'
+              d='m21.73 18-8-14a2 2 0 0 0-3.48 0l-8 14A2 2 0 0 0 4 21h16a2 2 0 0 0 1.73-3Z'
+            />
+            <path class='warn-icon-mark' d='M12 9.2v4.3' />
+            <circle class='warn-icon-dot' cx='12' cy='16.9' r='1.05' />
+          </svg>
```
habdelra: this SVG appears twice, can we just make a component in this gts file for that svg and reuse it?

habdelra: [Claude Code 🤖] Done in 3de0374 — pulled the svg into a `WarningIcon` component in this file, parameterized by `@size`, reused by both the box trigger (17) and the overlay title (16). Its `.warn-icon*` styles moved into the component’s own `<style scoped>` block so the styling travels with it (the parent `<style scoped>` no longer reaches the svg once it lives in a child component). Verified both render as the solid amber triangle, and the contract test stays green.

FadhlanR (approved):

PR #5049, opened 2026-05-31
Make silent host-test timeouts diagnosable and stop one timeout cascading into sibling failures

habdelra:
## Background and Goal

Host-test shards intermittently fail with opaque `settled()` timeouts, "editor is read-only"/disabled-field assertion failures, and cascades of unrelated downstream tests. This PR (a) makes those failures self-explaining with targeted, test-gated diagnostics, and (b) lands two test-infrastructure fixes for concrete races the diagnostics pinned down. It stacks on #5050, which carries the `store.peek` render-corruption fix.

## Where to start

- `packages/host/tests/helpers/setup.ts` — the silent-timeout dump (fired only on a failed >60s test) now reports, in addition to in-flight/recent fetches: settled-state metrics + pending test-waiter names, a **DOM snapshot of `#ember-testing` sampled while the test was still mounted** (what actually rendered in place of the awaited element, instead of the post-teardown empty container), **event-loop-lag samples** (runner CPU/GC contention vs. a genuine code hang), per-realm **auth state** (`loggedIn`/`canRead`/`canWrite`), and a **cascade flag** that names the earlier test whose render-corrupting global error a later timeout is a downstream cascade of.
- `packages/host/app/services/matrix-service.ts`, `app/routes/index.gts`, `app/components/matrix/login.gts` — `[login-diag]` reports which `isLoggedIn` precondition (`authPresent` / `clientLoggedIn` / `postLoginCompleted`) is unmet when operator-mode renders the login form, and `[start-phase]` markers name which `start()` / `loadState` await is in flight when a cold-boot login hang occurs.
- `packages/host/tests/helpers/mock-matrix/_utils.ts` — `getRoomIds()` returns `[]` when the mock SDK is torn down, so a realm's late `broadcastRealmEvent` becomes a no-op instead of a QUnit global failure that reds an unrelated sibling test.
- `packages/host/app/services/realm.ts` — a `realm:session-login` test waiter held across `RealmResource`'s login task, so `await settled()` waits for the realm session JWT to be minted; without it a card editor renders with disabled/read-only fields during the anonymous→logged-in window.

## Key decisions

- Everything here is additive and test-gated: the dump prints only on a failed >60s test; the lag sampler is a raw `setInterval` (invisible to `settled()`) holding ≤12 samples; the `[login-diag]` / `[start-phase]` markers are gated by `isTesting()`; and `buildWaiter` (the realm session waiter) is inert outside tests. No production code path changes behavior.
- Stacks on #5050 for the `store.peek` backtracking fix rather than carrying a divergent copy here — #5050 is the single source of truth for that change.

### DIFF
--- a/packages/host/app/components/matrix/login.gts
+++ b/packages/host/app/components/matrix/login.gts
@@ -1,8 +1,10 @@
 import { fn } from '@ember/helper';
 import { on } from '@ember/modifier';
 import { action } from '@ember/object';
+import type Owner from '@ember/owner';
 import type RouterService from '@ember/routing/router-service';
 import { service } from '@ember/service';
+import { isTesting } from '@embroider/macros';
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 
@@ -152,6 +154,19 @@ export default class Login extends Component<Signature> {
   @tracked private password: string | undefined;
   @service declare private matrixService: MatrixService;
   @service declare router: RouterService;
+
+  constructor(owner: Owner, args: Signature['Args']) {
+    super(owner, args);
+    if (isTesting()) {
+      // Names which isLoggedIn precondition is unmet when the login UI mounts
+      // in a test (the intermittent cold-boot login-screen flake). Test-gated;
+      // no production effect.
+      console.warn(
+        `[login-diag] login UI mounted: ` +
+          JSON.stringify(this.matrixService.loginReadinessDebug),
+      );
+    }
+  }
 
   private get isLoginButtonDisabled() {
     return (
--- a/packages/host/app/lib/gc-card-store.ts
+++ b/packages/host/app/lib/gc-card-store.ts
@@ -11,6 +11,7 @@ import {
   localId as localIdSymbol,
   loadCardDocument,
   loadFileMetaDocument,
+  rri,
   trackRuntimeFileDependency,
   trackRuntimeInstanceDependency,
   logger,
@@ -72,6 +73,10 @@ type StoreHooks = {
     },
   ): StoreSearchResource<T>;
 };
+
+function isCardOrFileInstance(item: unknown): item is StoredInstance {
+  return isCardInstance(item) || isFileDefInstance(item);
+}
 
 // we use this 2 way mapping between local ID and remote ID because if we end up
 // trying to search thru all the entries in a single direction Map to find the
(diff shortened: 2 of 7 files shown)


habdelra:
[Claude Code 🤖] A few diagnostic findings from the CI runs I observed, in case they're useful for this PR's goals:

**1. The cascade mechanism (directly relevant to "one flake → several red tests").** A render-time *global* failure — a thrown error during render, e.g. the backtracking re-render assertion fixed in #5050, or the mock-matrix `serverState` failure your broadcast-guard targets — leaves Ember's renderer **unrecoverable for the rest of the page load** ("Attempted to rerender, but the Ember application has had an unrecoverable error occur during render"). Every subsequent test in that shard then fails as an opaque `settled()` timeout or `Failed to fetch`. So a single flaky render error reds a *string* of unrelated downstream tests. Your broadcast-guard removes one source of these global failures; the `store.peek` mutation in #5050 was another. Worth knowing there may be more — the class is "anything that throws synchronously during render."

**2. Your `sessionWaiter` (`8be79dc5e2`) is load-bearing — keep it.** While it was briefly absent from this branch (my mistake, now restored), CI run 26724455935 went red on **four shards, all with the realm-permission/`canWrite` race signature** — the editor/menu rendering before the realm session JWT is minted:
- shard 5 — `workspace-chooser-delete`: `[data-test-boxel-menu-item-text="Delete Workspace"]` had `disabled`
- shard 6 — `workspace-chooser > delete menu item: opens the delete modal`
- shard 9 — `file def navigation`: `[data-test-realm-indicator-not-writable]` shown when it shouldn't be
- shard 11 — `create-file tests: create local instance in readonly remote realm`

All four show `[login-diag]` with the session still settling. With the `sessionWaiter` present, `await settled()` waits for the JWT, so these settle. (One residual gap you noted still stands: the waiter only covers *login in flight* — if a render reads `canWrite` before `loginTask` starts, it can still race.)

**3. A cascade-attribution diagnostic you may want to keep.** Commit `2d7bb7a315` on this branch adds (to `tests/helpers/setup.ts`) a remembered "first render-corrupting global error of the run"; on a *later* test's timeout dump it names the originating test and flags the failure as a likely cascade rather than an independent failure (gated to the backtracking / "unrecoverable" / "Attempted to rerender" signatures so a benign earlier rejection is never blamed). It complements your DOM-snapshot + event-loop-lag dump — it answers "which test actually broke the app" for a cascade. Lift it from `2d7bb7a315` if useful, or drop it.

backspace (approved):

PR #5053, opened 2026-06-01
Surface an actionable error for missing boxel-icons imports 

jurgenwerk:
## Problem

A card importing a non-existent icon — e.g. `import PersonCircle from '@cardstack/boxel-icons/person-circle'` when no such icon exists — resolves to a CDN URL whose key is missing. The S3-backed production CDN returns **403** for missing keys (the bucket withholds `s3:ListBucket` from the public principal, so absent keys 403 instead of 404); the local `http-server` returns **404**.

The browser host loader (`loader-service.ts`) already rewrites a failed icon fetch to a fallback `error-404.js` module, so UI cards render a broken-icon glyph and never error. But the **indexing worker** builds its own loader middleware stack (`worker.ts`) with no such fallback, so the failure reached the shared `Loader.fetchModule` chokepoint and the raw S3 XML landed in `boxel_index.error_doc.message`:

```
unable to fetch https://boxel-icons.boxel.ai/.../person-circle.js:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message></Error>
```

Users saw "AccessDenied" and reasonably read it as a permissions/outage problem, when the real fix is to correct the import.

## Change

Translate the failure at the **loader layer** (per the ticket), so both the indexer and any non-fallback user-facing import benefit:

- `runtime-common/error.ts` — new `iconNotFoundMessage(moduleHref, status)`. Matches status `403`/`404` plus the environment-independent path segment `/@cardstack/boxel-icons/v1/icons/<name>.js` (the import map produces this shape in every environment, so no host config is needed inside runtime-common). Returns `undefined` to pass through when it isn't a missing-icon fetch.
- `runtime-common/loader.ts` — `fetchModule` consults it before `CardError.fromFetchResponse` and throws the actionable message instead of the XML:

  > Icon "person-circle" was not found in @cardstack/boxel-icons. Check the import path against the available icons.

The optional Levenshtein "did you mean …?" suggestion is intentionally out of scope here.

## Tests

- `runtime-common/tests/icon-not-found-message-test.ts` (+ `realm-server` shim, registered in `tests/index.ts`) — unit coverage of the helper: 403/404 translation, status/path passthrough, URL-normalization and empty-name edge cases.
- `host/tests/unit/loader-icon-error-test.ts` — a bare `Loader` (no fallback middleware, standing in for the worker path) confirms a 403/404 icon fetch throws the actionable message, not the XML.

## Verification

- `ember-tsc --noEmit` clean on runtime-common, host, realm-server; eslint + prettier clean.
- Helper URL-parsing logic validated against all edge cases.


habdelra:
The test should just live in host package--the loader never executes in the realm-server, so there is no need for an isomorphic test.

lukemelia (approved):
Keep in mind that we can also choose to adjust the behavior of the cloudfront/S3 bucket.

PR #5055, opened 2026-06-01
feat: promote AudioDef into @cardstack/base with full format support

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-11309/introduce-audio-filedef-hierarchy-to-packagesbase

## Context

Audio in the catalog used to be stored as a blob: URL on the field itself, which doesn't survive a page reload — the `<audio>` element 404s the moment the blob is GC'd, and downstream surfaces (the AI assistant included) just see `blob:http://localhost…` errors with no recovery path.

The fix is to back audio with a proper persisted `FileDef` — specifically`AudioDef`, the audio-flavored `FileDef` subclass already living in`@cardstack/base`.. To keep review surface area small, that migration is split into **two PRs**:

1. **Current PR** — promote `AudioDef` into `@cardstack/base` as a first-class `FileDef` subclass with all four format slots filled in, matching `ImageDef` / `MarkdownDef` / `TextFileDef`. No catalog consumer changes.
2. **Follow-up PR - https://github.com/cardstack/boxel-catalog/pull/588 ** — rewrite `packages/catalog/.../fields/audio.gts` on top of the now-richer `AudioDef`, removing the blob: storage path and the temporary back-compat shims.

### `packages/base`

- New default templates colocated with the existing `image-def-*` set:
  - `default-templates/audio-def-isolated.gts` — full `<audio controls>`
    with filename and formatted duration
  - `default-templates/audio-def-embedded.gts` — compact header + player
  - `default-templates/audio-def-fitted.gts` — icon + name + duration with
    portrait / landscape / very-short `@container fitted-card`
    breakpoints, mirroring `text-file-def.gts`
  - `default-templates/audio-def-atom.gts` — inline icon + name
- `audio-file-def.gts` wires the four slots, plus a `markdown` override
  that emits `[name](url)` via `markdownLink` (HTML5 has no native
  markdown syntax for audio)
- `AudioDef` stays in its own module — `ImageDef` only lives in
  `card-api.gts` because of the `CardInfoField.cardThumbnail`
  `linksTo(ImageDef)` cycle; no such cycle exists for audio, so we don't
  need the `image-file-def.gts` re-export shim pattern here
- `Mp3Def` / `WavDef` / `M4aDef` / `OggDef` / `FlacDef` all
  `extends AudioDef` and inherit the new format slots automatically


lucaslyl:
@codex review

lukemelia (commented):
How is duration getting populated? I don't see any `extractAttributes` implementations

lucaslyl:
> How is duration getting populated? I don't see any `extractAttributes` implementations

Good catch — I missed wiring up the extractAttributes layer entirely. Addressed in 621bfcc.

Each leaf audio def (Mp3Def / WavDef / FlacDef / OggDef / M4aDef) now overrides extractAttributes to parse its codec's metadata into the duration field, mirroring the image-def per-format pattern (jpg/png/etc each parse their own header bytes). Each one throws FileContentMismatchError when the bytes don't actually match the declared format, so the indexer falls back to AudioDef and the file still indexes without duration. Added a {format}-meta-extractor.ts per codec, plus acceptance tests per format under tests/acceptance/audio-def/ covering extract / mismatch fallback / isolated render / index + file-meta.

Related fixes while I was in here:

- AudioDef.markdown now emits <audio controls> via a new markdownAudio helper, so the Spec preview renders a real player instead of a bare link — symmetric with how ImageDef.markdown emits markdownImage.

lukemelia (dismissed):
There are some memory optimizations I'd like to see here. I'll open a PR against this one shortly.

lukemelia:
Once #5076, and #5077 are merged into this branch, it will be ready to merge. #5075 against main is a related piece of this memory work.

lucaslyl:
> Once #5076, and #5077 are merged into this branch, it will be ready to merge. #5075 against main is a related piece of this memory work.

Got it! Thanks for the heads up — will merge #5076 / #5077 in once they land.

lukemelia (approved):

lukemelia:
Should be good to go now. The key idea to remember with FileDef's `extractAttributes` is to avoid reading the entire contents of a file into memory, especially if the file might be large.

PR #5060, opened 2026-06-01
Keep h2 sessions live and tear down wedged ones

habdelra:
## Background and Goal

Host acceptance tests intermittently time out with no assertion failure — the diagnostic dump shows a single in-flight base-realm fetch (e.g. `QUERY https://localhost:4201/base/_info`) and an open `fetcher` test waiter, so `settled()` never returns.

The realm server speaks HTTP/2 to the browser on the dev/CI TLS path (production is plain HTTP behind Traefik, so this path is dev/CI only). Under load an h2 session can wedge: a request is queued but the stream neither completes nor errors. The browser's `fetch` then hangs forever — and because it never *rejects*, the existing test-env fetch retry (`withRetries` in `virtual-network.ts`, which only retries rejections) never fires.

The realm server's `[h2-diag]` instrumentation already rules out a server-side post-`stream` stall on the captured failures (no stalled-stream dumps, no GOAWAY, clean session open/close), which points at the Chrome↔Node h2 transport. This change makes the realm server actively detect and tear down a wedged session so the hung fetch becomes a recoverable connection error instead of a silent timeout. h2 is preserved everywhere.

## Where to start

- `packages/realm-server/server.ts` — `installHttp2Keepalive` (wires the keepalive onto every accepted h2 session) and `startSessionKeepalive` (the per-session PING loop + teardown). Both sit next to `createListener`, which now calls the keepalive on the h2 path.
- The `installHttp2Diagnostics` stalled-stream dump now also reports the session's ping liveness (`keepalive(pingRtt=… pongAge=… missedPings=…)`).
- `packages/realm-server/tests/http2-keepalive-test.ts` — unit coverage of the PING loop against a fake session.

## Key decisions

- **PING keepalive, not a client-side fetch timeout.** A blanket fetch abort/timeout in the network layer masks the transport problem; tearing the wedged session down server-side converts the silent hang into a normal connection error that the existing retry path already recovers from.
- **Close then force-destroy.** A graceful `close()` (GOAWAY) won't end a stream that's already wedged, so the session is destroyed after a short grace period to actually RST the hung stream.
- **Generous timeouts.** Worst-case detection is `maxMissedPings × (interval + pongTimeout) + grace` ≈ 55s — well under the host-test timeout, and deliberately lax so a briefly-busy server event loop never causes a false teardown (which would only cost a reconnect, never a wrong result).
- **TCP keepalive backstop** is enabled on the socket so a dead peer is also surfaced at the transport layer.

### DIFF
--- a/packages/realm-server/server.ts
+++ b/packages/realm-server/server.ts
@@ -47,6 +47,37 @@ const TLS_KEY_FILE_ENV = 'REALM_SERVER_TLS_KEY_FILE';
 // so this only ever runs in local dev / CI.
 const HTTP2_DIAGNOSTICS_ENV = 'REALM_SERVER_HTTP2_DIAGNOSTICS';
 
+// HTTP/2 PING keepalive tuning. The realm server speaks h2 to the browser on
+// the TLS path (dev/CI only; production is plain HTTP behind Traefik, see
+// createListener). An h2 session can wedge — a stream is queued but neither
+// completes nor errors — leaving the browser's `fetch` hung forever (it never
+// rejects, so client-side retry never kicks in and a host test eventually
+// times out). The PING keepalive turns that silent wedge into an observable,
+// recoverable event: every live session is pinged; a session that fails to
+// answer within the budget is closed, which RSTs its streams so the browser's
+// hung fetch rejects and its existing retry path can reissue on a fresh
+// session. Worst-case detection is maxMissedPings × (interval + pongTimeout) +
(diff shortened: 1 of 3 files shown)


habdelra:
[Claude Code 🤖] Closing in favor of a diagnostics-first approach. The keepalive was a recovery mechanism, not a root-cause fix, and the server-side h2-diag came back clean on the captured failure — which means the wedge is in a place the current stream-level sweep can't see. Replacing this with targeted session-level h2 diagnostics to actually identify the wedge before committing to a fix.

PR #5080, opened 2026-06-02
Adorn search pane improvements

lukemelia:
## Summary
Visual refinements to the shared Adorn search components (the operator-mode search sheet, and the card chooser once it opts in). No new opt-in — these polish the treatment that's already live in the search sheet.

- **Stroke no longer clips in the grid** — `.search-sheet-content`'s `overflow-y: auto` also clips horizontally, cutting off the teal outline stroke / type-label tab on the top/left edge cards. Added block + inline-start padding so they stay inside the clip region.
- **Darken the label when a selected card is hovered** — the outline stroke already shifts to the darker accent on selected+hover; the type-label tab now follows via `--adorn-label-bg` (plain hover and the resting selected state keep the light accent).
- **Room for the recents label tab** — in the compact quick-menu the hover tab flips below each recent card; raised the prompt-sheet height and biased the row's block padding so the tab isn't clipped (it overlaps the realm-name line, which is acceptable).

## Test plan
- [x] `lint:types` + eslint pass.
- [x] Manual: search sheet grid — hover/select edge cards; stroke + tab not clipped.
- [x] Manual: hover a selected card — stroke and label both darken.
- [x] Manual: compact recents quick-menu — hover tab appears below the card, fully visible.

### DIFF
--- a/packages/host/app/components/adorn/adorn-context.gts
+++ b/packages/host/app/components/adorn/adorn-context.gts
@@ -101,14 +101,18 @@ const AdornContext: TemplateOnlyComponent<AdornContextSignature> = <template>
       /* Token definitions live with the context, not in the global
          stylesheet. --boxel-teal is the light accent shipped by
          boxel-ui; the darker accent is exclusive to the Adorn
-         treatment and used when both hovered and selected. */
+         treatment and used when a selected card is hovered. */
       --adorn-accent-light: var(--boxel-teal);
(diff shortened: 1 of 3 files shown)


backspace (approved):
I noticed this in the preview deployment but it’s not related to this PR as it’s on staging too; the roundrect has faint background in the rounded areas

Image

PR #5090, opened 2026-06-03
Limit full-realm module pre-warm sweep to from-scratch indexing

habdelra:
`IndexRunner` pre-warmed the definition cache for every `.gts`/`.gjs` module in the realm on every index — from-scratch and incremental alike. That sweep is O(realm module count), not O(files changed), so a single-file edit on a module-heavy realm pays the full-realm tax: warming hundreds of modules (mostly cache hits, but serial) dominates the wall time of an incremental whose real work is a handful of cards. When the cache is cold (e.g. right after a deploy), the same sweep re-prerenders every module and the cost balloons further.

The sweep exists for one reason: to prime sibling card modules that a card references only as a string (e.g. a `Search` query's `type.module`), which never appear in any instance's runtime `deps`. Without it, a render's `lookupDefinition` for such a module fires a same-affinity sub-`prerenderModule` mid-render. That cold-cache concern is real only on from-scratch. On an incremental the module cache is already warm from the prior from-scratch, and a miss now resolves through the on-demand read-through during the visit — which is PagePool-safe (it materializes its own tab) rather than deadlocking.

This change runs the realm-wide sweep on **from-scratch only**. Incrementals pass an empty base set to `preWarmModulesTable` and keep the per-invalidation warming (bounded by the invalidation set — the invalidated module files plus the per-row deps of the invalidated cards). It also drops the incremental-only filesystem-mtimes walk that existed solely to build the sweep.

### Test

Adds a realm-server test that uses an orphan card module — defined but adopted by no instance and imported by no module — as a deterministic probe. Nothing ever invalidates it, so the only code path that can land it in the `modules` cache is the realm-wide sweep. The test asserts it is present after from-scratch indexing and absent after an incremental of an unrelated instance.

### DIFF
--- a/packages/runtime-common/index-runner.ts
+++ b/packages/runtime-common/index-runner.ts
@@ -434,20 +434,25 @@ export class IndexRunner {
       }
       current.#scheduleClearCacheForNextRender();
     }
-    // Pre-warm: combine per-row deps with a realm-wide `.gts`/`.gjs`
-    // sweep. Incremental skips `discoverInvalidations` so the
-    // filesystem-mtimes walk hasn't happened yet — call it here.
-    // Typical realm sizes make this < 200 ms; one call per job.
-    let incrementalMtimes = await current.#reader.mtimes();
-    let allRealmCardModules =
-      Object.keys(incrementalMtimes).filter(hasCardExtension);
+    // Pre-warm only the modules this batch's invalidations actually
+    // touch — the invalidated module files plus the per-row deps of the
+    // invalidated cards. The realm-wide `.gts`/`.gjs` sweep is reserved
+    // for from-scratch indexing, where the persistent module cache is
+    // cold by definition. On an incremental that cache is already
+    // populated by the prior from-scratch, so a sibling module a card
+    // references only by string resolves through the on-demand
+    // `lookupDefinition` read-through during the visit (PagePool-safe:
+    // the sub-prerender materializes its own tab) rather than paying an
+    // O(realm) sweep for a single-file edit. Passing an empty base set
+    // also skips the filesystem-mtimes walk this path would otherwise
+    // run only to build that sweep.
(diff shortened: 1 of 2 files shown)


packages/runtime-common/index-runner.ts
habdelra, line 475: why are we even calling this? is there side effect behavior that we rely on?

habdelra: [Claude Code 🤖] Not a no-op — `preWarmModulesTable` has two layers and the empty base set only drops the first one (the realm-wide `.gts`/`.gjs` sweep). The second layer still runs: for each invalidation it warms (a) the URL itself if it's an executable module, (b) the module deps recorded on that row in `boxel_index`, and (c) `adoptsFrom.module` read from disk for a novel `.json` with no prior deps. So editing a module primes that module plus the modules of the cards that depend on it, ahead of the visit that renders them — that's the "keep the per-invalidation warming, bounded by invalidation size" part of the plan.

Do we *rely* on it? Not for correctness. I wiped the `modules` table, ran an incremental, and indexing completed fine — the visit's on-demand `lookupDefinition` read-through repopulates on a miss, and that path is now PagePool-safe (it materializes its own tab instead of deadlocking behind the render). So on incrementals this is a latency front-load, not a requirement. If you'd rather drop the call entirely on incrementals and lean on the read-through, I'm happy to — it'd simplify this path. Your call.

habdelra: [Claude Code 🤖] Keeping the call. Added a comment above it (0950fee) spelling out the per-invalidation warming it still performs — the invalidated module + its dependents' per-row deps + a novel `.json`'s `adoptsFrom` — and why that front-load earns its keep on incrementals, so it no longer reads as a no-op.

backspace (approved):

PR #5099, opened 2026-06-03
Apply Burcu's design-review feedback to Adorn UI

lukemelia:
Refactors the Adorn components/modifiers and their consumers to follow the design-system conventions, applying each item from Burcu's review.

## Conventions applied
- **`data-*` over class names for DOM targeting** — the label-boundary resolver now keys off `data-adorn-context` instead of `closest('.adorn-context')`. The class stays for styling; the data attribute is the JS contract.
- **`rem` instead of `px`** — converted dimensional CSS across the Adorn primitives, the card-search selection trigger, and the card-header utility-menu trigger. SVG-internal coordinates and JS pixel math left in px.
- **Hardcoded colors → semantic tokens** — no raw hex in components; all references go through tokens.
- **`cn` helper for conditional classes** — replaced inline `{{if}}`/`{{unless}}` class-string concatenation in `AdornLabel` / `AdornSelectChip` (and the AdornContext consumer). Emits identical class strings, so behavior and selectors are unchanged.
- **`currentColor` for SVGs** — the themeable check stroke follows `color`; fixed brand parts reference a token.
- **`--boxel-highlight` instead of `--boxel-teal`** for the highlight accent, app-wide.

## Token consolidation
- `--boxel-dark-teal` is now `#00da9f` (previously `#00ebac`), so the existing `--boxel-highlight-hover` carries it rather than introducing a parallel variable. This is an app-wide shift: hover/highlight states (buttons, pills, selects, pickers, multi-select, codemirror, …) move to the slightly deeper teal.
- Added `--boxel-highlight-foreground: #0a2e1c` — the readable dark-green foreground for text/icons on a `--boxel-highlight` surface, following the established `<surface>` / `<surface>-foreground` convention. Wired through the Adorn primitives, the card-search selection trigger, and `card-header` (whose utility-menu trigger previously duplicated the same pattern with hardcoded colors).

## Visual impact
- **No change**: teal backgrounds and icon colors resolve to the same hex as before.
- **Intended changes**: text/icons on highlight surfaces shift from pure black to the dark-green `--boxel-highlight-foreground` (the editing card-header bar and the selection/utility-menu triggers), and hover/highlight teal deepens slightly via the `--boxel-dark-teal` value change.

## Docs
- New `gts-component-conventions` skill capturing these guidelines.
- The skill plus AGENTS.md now document the content-tag `<template>`-detection hazards hit during this work, including a previously-undocumented trigger (a backtick-wrapped bracket token inside a template-region comment) and its ESLint-phantom-unused-import symptom.

### DIFF
--- a/.claude/skills/gts-component-conventions/SKILL.md
+++ b/.claude/skills/gts-component-conventions/SKILL.md
@@ -1,0 +1,146 @@
+---
+name: gts-component-conventions
+description: Styling and authoring conventions for `.gts` components and their CSS in the host and boxel-ui packages, plus the content-tag `<template>`-detection hazards that silently break `.gts` parsing. Use whenever writing, reviewing, or refactoring a `.gts` component, a `<style scoped>` block, or a glimmer template — especially component-styling review passes ("apply these design-review notes", "clean up this component's CSS") and when a `.gts` file mysteriously fails to type-check or lints with phantom unused-import errors. Triggers on editing component markup/CSS, adding SVG icons, writing conditional class names, choosing colors or units, and any cascading "Cannot find name 'template'" or template-was-dropped symptom.
+---
+
+# `.gts` Component Conventions
+
+Two things live here:
+
+1. **Design-review guidelines** (from Burcu) for component markup and CSS — apply when writing or reviewing any `.gts` component or `<style scoped>` block.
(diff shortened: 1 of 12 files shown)


lukemelia:
@burieberry still more to come based on your feedback, but please review this batch

burieberry (approved):
Looks good, thanks for updating the skills!

PR #5101, opened 2026-06-03
Extract SelectionMenu into boxel-ui as a shared component

lukemelia:
Prep for bringing the bulk-selection menu to the workspace cards grid. Extract the selection control out of the card chooser's results header into a **generic boxel-ui `SelectionMenu`**, so any surface can adopt it — including the base `CardsGrid` card, which can only import boxel-ui (not host).

## The trigger is a primary `BoxelButton`
Rather than a bespoke `<button>`, the trigger is a standard `Button @kind='primary'`, so it inherits the design system's highlight colors, hover, and focus-ring behavior. The component's scoped CSS only adds what Button doesn't: the gap between the checkmark / count / caret, the readable highlight foreground (Button's primary text defaults to `--boxel-dark`), a tighter radius, the open-state deepening, and the caret flip.

## Generic / content-agnostic
The caller supplies the menu contents via `@items`, so **Select All / Deselect All** (and the inert count header) stay in the app — built in `SearchResultHeader` and handed in — and app-specific selection semantics don't leak into the design system. `@label` sets the trigger's accessible name (the card chooser passes "N cards selected").

## Shared icon + card-header dedup
Moved the `SelectionCheckmark` artwork into boxel-ui (a two-color composite, so it's a component, not part of the generated monochrome icon set) and pointed `CardHeader` at it — removing a byte-identical inline SVG and advancing **CS-11333**.

## Verification
boxel-ui + host both pass `ember-tsc` (0 errors), eslint, and prettier; existing selection-menu tests' markup/`data-test`/aria-label are unchanged. The trigger's rendering now derives from `Button` (radius/height/hover), so it's worth a fresh in-browser glance.

Stacked on #5100 → #5099; auto-retargets to `main` as those merge. (The workspace-grid integration — wiring host selection state into the base card's toolbar — is the follow-up Adorn UI issue.)


lukemelia:
[Claude Code 🤖] Folding this into #5100 for review — the extraction reworks the same selection trigger that PR introduces, so reviewing them together avoids reviewing intermediate state that this PR immediately replaces. #5100 now contains the full change (squashed): the primary-dropdown trigger + the boxel-ui SelectionMenu extraction + the SelectionCheckmark dedup.

PR #5103, opened 2026-06-04
Skip file-meta operations in listing install

FadhlanR:
## Background and Goal

Remixing a catalog listing whose example cards link to a FileDef-typed binary (avatar, thumbnail, etc.) silently fails. The install command sends a JSON:API atomic batch that includes operations with `data.type: 'file-meta'`, but the atomic handler at `packages/runtime-common/realm.ts:2107-2117` only accepts `card`/`source` resource types and returns 422 for anything else.

The breaking change was the rename of FileDef JSON:API resources from `type: 'card'` to `type: 'file-meta'`. Before the rename, atomic happily accepted the operation and wrote a `.json` sidecar at the binary's `.jpg.json` path — but **no binary bytes were ever copied**. The image still rendered because the parent card's `cardInfo.cardThumbnail.data.id` is an absolute source-catalog URL, so the remixed card linked back cross-realm. The rename didn't break a working feature — it surfaced a long-broken step (the sidecar tax) as a hard 422.

## Fix

Drop documents with `data.type === 'file-meta'` from the install command's per-instance operation builder. Binaries continue to resolve cross-realm via the parent card's relationship `data.id`, which is what was actually working pre-rename minus the no-op sidecar.

The operation-building was also extracted into `buildInstanceOperation` so the filter behavior is directly unit-testable.

## Test plan
- [x] Unit test in `listing-install-build-operation-test.ts` covers the filter behavior (file-meta → undefined, card → operation, malformed → throws)
- [x] Verify a listing with FileDef-linked thumbnails remixes successfully (staging or local with the right fixture)
- [x] Confirm remixed thumbnails still render via cross-realm reference

## Out of scope

- **Surface atomic errors in the UI.** The silent-422 swallow happens upstream of `RemixCommand` and needs its own design (the host's `ErrorDisplayService` is registry-based, not imperative). Filing as a follow-up.
- **Real binary copy + reference rewriting.** True "remix owns its binaries" is a separate feature: it requires moving bytes server-to-server AND rewriting every reference (linksTo `data.id`, hardcoded URLs in `.gts` / templates / CSS / markdown). Out of scope for this bug fix.

### DIFF
--- a/packages/host/app/commands/listing-install.ts
+++ b/packages/host/app/commands/listing-install.ts
@@ -36,6 +36,35 @@ import type NetworkService from '../services/network';
 
 const log = logger('catalog:install');
 
+export type InstanceOperation = {
+  op: 'add';
+  href: string;
+  data: LooseCardResource;
+};
+
+// file-meta resources are JSON projections of binary files in the source
+// realm. The atomic endpoint only accepts card/source types; binaries continue
+// to resolve cross-realm via the parent card's relationship data.id, so we
+// drop file-meta documents rather than try to copy them.
+export function buildInstanceOperation(
+  doc: unknown,
+  copyInstanceMeta: CopyInstanceMeta,
+  realmIdentifier: string,
+): InstanceOperation | undefined {
+  if (!doc || typeof doc !== 'object' || !('data' in doc)) {
+    throw new Error('We are only expecting single documents returned');
+  }
+  let data = (doc as { data: { type?: string } }).data;
+  if (data?.type === 'file-meta') {
+    return undefined;
+  }
+  delete (doc as any).data.id;
+  delete (doc as any).included;
+  let cardResource = (doc as any).data as LooseCardResource;
+  let href = join(realmIdentifier, copyInstanceMeta.lid) + '.json';
+  return { op: 'add', href, data: cardResource };
+}
+
 export default class ListingInstallCommand extends HostBaseCommand<
   typeof BaseCommandModule.ListingInstallInput,
   typeof BaseCommandModule.ListingInstallResult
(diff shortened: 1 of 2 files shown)


lukemelia (changes requested):
Is there a plan for this feature? Silently dropping file-meta resources seems like it might hide issues. If we can't support updating non-cards non-gts via atomic, we should raise an error if someone attempts to do this, and there should be a supported way to update files during a listing installs.

FadhlanR:
[Claude Code 🤖]

Good point worth digging into. Looking at the history, the previous atomic/remix flow also didn't upload the real binary — when the FileDef document was tagged `type: 'card'` (pre-Jan-13 rename), atomic accepted the operation and wrote an inert `.jpg.json` sidecar containing only the FileDef metadata. No bytes were ever copied to the destination realm. The thumbnails rendered because the parent card's `cardInfo.cardThumbnail.data.id` is an absolute catalog URL, so the remixed card linked cross-realm to where the bytes actually live.

So this PR is scoped to restore the pre-rename remix behavior — cross-realm link works, no destination bytes, no failing atomic op. It's removing a no-op write that was masking the same gap, not adding a new silent path.

That said, I agree we should introduce a proper feature where remix actually copies binaries into the destination realm. That's net-new scope though — bytes need to move server-to-server, and every reference (linksTo `data.id`, hardcoded URLs in `.gts`, templates, CSS, markdown) needs to be rewritten to the new URL. Happy to file a follow-up ticket so the thread doesn't get lost.

FadhlanR:
[Claude Code 🤖]

Superseded by cardstack/boxel-catalog#600. Listing commands moved out of host in #5146, so the fix moves with them. Continuing the discussion there — @lukemelia's review feedback has been carried over.

PR #5105, opened 2026-06-04
Standardize dropdown trigger arrow, sizing, and primary focus ring

lukemelia:
Three boxel-ui trigger-polish items from Burcu's review of #5083, scoped to the card-chooser/toolbar triggers (not the unrelated host chrome).

## [CS-11326](https://linear.app/cardstack/issue/CS-11326) — one standard arrow + flip on open
- The sort menu used the filled `DropdownArrowDown` and **never flipped** on open (the regression). boxel-ui's form controls (select, multi-select, picker, dropdown) already use the thin `CaretDown` chevron and flip.
- Switched the `sort-dropdown` and `selection-menu` triggers to `CaretDown`, and the sort button now flips its caret on open via its `aria-expanded` state.
- Scope decision: toolbar triggers only — the ~8 host-chrome `DropdownArrowDown` usages (file tree, submode switcher, realm dropdown, etc.) are intentionally left alone.

## [CS-11327](https://linear.app/cardstack/issue/CS-11327) — trigger doesn't shift when count changes
- The `selection-menu` count now reserves a stable 2-ch, tabular-figure width, so the trigger no longer jumps when the count crosses 1↔2 digits (e.g. a Select All going 9→10).
- Height already matches the adjacent sort button via the shared `--boxel-button-sm` min-height; the old "squished padding" was the pre-rebuild trigger.

## [CS-11328](https://linear.app/cardstack/issue/CS-11328) — primary button focus ring
- Base `BoxelButton:focus-visible` now sets the full outline shorthand (`--boxel-outline-width`/`-style`, i.e. 2px solid) instead of only `outline-color` — previously it relied on the UA default outline-style. Ring sits 2px outside.
- The primary ring darkens to match the fill on hover/active.
- The `selection-menu`'s local focus-ring override is now redundant and removed.
- Note: this `:focus-visible` change is on the shared `BoxelButton`, so every button's keyboard focus ring is now explicitly `2px solid` rather than the browser default — intended (it's the design-system `--boxel-outline`).

## Base branch
Targets `cs-11325-…` (PR #5100, the shared `SelectionMenu`), same as #5102 — keeps this diff scoped. Will rebase onto `main` and retarget once #5100 lands.

## Testing
- eslint + glint clean.
- No existing test asserts on the sort arrow or the button outline; the one `selection-menu` test targets the checkmark's `circle` (the caret has none, so it stays unambiguous).
- Behavioral checks (sort caret flip, focus-ring darkening, count not shifting) pending in-browser verification after a host rebuild.

### DIFF
--- a/packages/boxel-ui/addon/src/components/button/index.gts
+++ b/packages/boxel-ui/addon/src/components/button/index.gts
@@ -165,8 +165,12 @@ const ButtonComponent: TemplateOnlyComponent<Signature> = <template>
           transparent
         );
       }
+      /* Keyboard focus ring sits just outside the button. Set the full
+         outline shorthand (not just the color) so the ring renders
+         regardless of the UA's default outline-style. */
       .boxel-button:focus-visible {
-        outline-color: var(--ring, var(--boxel-highlight));
+        outline: var(--boxel-outline-width) var(--boxel-outline-style)
+          var(--ring, var(--boxel-highlight));
         outline-offset: 2px;
       }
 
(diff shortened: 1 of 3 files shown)


lukemelia:
[Claude Code 🤖] Split into three stacked, single-issue PRs (byte-identical to this combined diff):
- #5114 — CS-11326 (CaretDown + flip) → base `cs-11325`
- #5115 — CS-11327 (count width) → on #5114
- #5116 — CS-11328 (button focus ring) → on #5115

Closing this in favor of those.

PR #5111, opened 2026-06-04
Fix card-selection border weight, transition sync, and Adorn tag grey peek

lukemelia:
Adorn selection-overlay polish from Burcu's review of #5083. Diagnosed and verified live in the running app via DevTools computed styles + geometry.

## [CS-11329](https://linear.app/cardstack/issue/CS-11329) — selection border too thin (fully addressed)
In the card chooser, a selected item showed only the thin 1px button border. The intended teal ring was supposed to come from `AdornContext`'s `:deep(.adorn-stroke.selected)` rule, but the catalog item didn't pick it up (computed `box-shadow` on a selected item was `rgba(0,0,0,0) 0 0 0 0`). Fix: define the ring directly on `.catalog-item.adorn.selected` — a 4px `--boxel-highlight` ring, darkening to `--boxel-highlight-hover` on hover — so it stays full-weight whether or not the item is hovered. Verified: now `rgb(0,255,186) 0 0 0 4px`. (Operator-mode overlays already rendered a correct 4px ring, so this is scoped to the chooser/search `item-button`.)

## [CS-11330](https://linear.app/cardstack/issue/CS-11330) — border lags the tag (fully addressed)
The ring faded in over `box-shadow 0.2s` while the selection tag/chip render instantly, so the border appeared ~0.2s after the tag. Fix: `transition: none` on the selected ring so the two appear together. (In operator mode both were already instant — no desync there.)

## [CS-11331](https://linear.app/cardstack/issue/CS-11331) — Adorn tag polish (2 of 3 points; issue stays open)
- **Grey peek (fixed):** the label tab cast a grey `drop-shadow` filter that bled past its clipped/rounded corners; removed it (verified `filter: none`).
- **Vertical alignment when below the card (fixed):** the tab overlapped the card bottom by 2px, riding 1px too high; reduced to a 1px tuck under the bottom outline stroke (verified live, overlap 2px → 1px).
- **Border-radius (not changed, deliberately):** the tab is a separate 24px-tall flag; the card is 10px. Force-matching the radii would make the small tab look disproportionate, so the right treatment is a design call left for Burcu. CS-11331 stays open for it.

## Base branch
Targets `cs-11325-…` (PR #5100) because the Adorn overlay/tag code these touch was reworked on that stack and isn't on `main` yet — branching off `main` would patch stale code. Will rebase onto `main` once the stack lands.

## Testing
- eslint clean.
- Verified live in the running host app via DevTools computed styles + geometry (before/after values above); the dev server picked up each edit on reload.

### DIFF
--- a/packages/host/app/components/adorn/adorn-label.gts
+++ b/packages/host/app/components/adorn/adorn-label.gts
@@ -69,7 +69,6 @@ const AdornLabel: TemplateOnlyComponent<AdornLabelSignature> = <template>
       border-radius: 0.3125rem 0 0 0.3125rem;
       clip-path: polygon(0 0, calc(100% - 0.8125rem) 0, 100% 100%, 0 100%);
       z-index: 1;
-      filter: drop-shadow(0 0.3125rem 0.5rem rgba(0, 0, 0, 0.2));
     }
     /* Mirrored polygon when the label flips below the card so the
        slope still points toward the card edge. */
--- a/packages/host/app/components/card-search/item-button.gts
+++ b/packages/host/app/components/card-search/item-button.gts
@@ -450,8 +450,11 @@ export default class ItemButton extends Component<Signature> {
          4px whether or not the item is hovered, darkening on hover to match
          the rest of the Adorn treatment. `transition: none` makes the ring
          appear in lockstep with the selection tag/chip (which render
-         instantly) instead of fading in ~0.2s later. */
-      .catalog-item.adorn.selected {
+         instantly) instead of fading in ~0.2s later — and it's needed on the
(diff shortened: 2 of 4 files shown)


lukemelia:
[Claude Code 🤖] Superseded — split into 4 focused PRs (each rebased onto current `main`):

- #5117 — CS-11329 selection-ring weight (base `main`)
- #5118 — CS-11330 transition sync (stacked on #5117)
- #5119 — rename `catalog-item` → `item-button` (stacked on #5118)
- #5120 — CS-11331 grey-peek removal + 1px nudge (base `main`, independent)

Closing this bundled draft; reopen if needed.

PR #5120, opened 2026-06-04
Remove the Adorn label grey peek and sit it flush below the card (CS-11331)

lukemelia:
Resolves [CS-11331](https://linear.app/cardstack/issue/CS-11331) (2 of 3 points). Split out of #5111.

- **Grey peek (fixed):** the label tab cast a grey `drop-shadow` filter that bled past its clipped/rounded corners; removed it.
- **Below-card vertical alignment (fixed):** the tab overlapped the card bottom by 2px, riding 1px too high; reduced to a 1px tuck under the bottom outline stroke.
- **Border-radius (not changed, deliberately):** the tab is a separate 24px-tall flag; the card is 10px. Force-matching the radii would make the small tab look disproportionate, so the right treatment is a design call left for Burcu. CS-11331 stays open for it.

Base: `main` — independent of the `item-button` ring stack (#5117 → #5118 → #5119), so it can land on its own.


jurgenwerk (approved):

burieberry (approved):
border radii are fine with me as long as they're matching those of the design

PR #5144, opened 2026-06-08
Clamp adorn overlay label to card tile and truncate

jurgenwerk:
Before: 

image

After:

image


lukemelia:
The "before" is as-designed.

PR #5147, opened 2026-06-08
fix: resolve alias-form card id before sending to realm server in screenshot

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-11427/fix-screenshot-command-failing-in-alias-prefix-realms

Root cause:

ScreenshotCardCommand takes the card's in-memory id directly and POSTs it to /_screenshot-card on the realm server. In alias-prefix realms, card.id is alias form (@cardstack/catalog/SimpleCard/uuid). The realm server has no alias mapping — it only knows HTTP URLs — so when it tries to construct a URL from the alias string it throws Invalid URL and the screenshot job fails.

## Fix

- `screenshot-card.ts`: Resolve alias-form card id to HTTP URL via
  `vn.toURL()` before POSTing to `/_screenshot-card` — the realm server
  has no alias mapping and throws `Invalid URL` when given an alias-form
  id directly

## Test plan
- [x] Create a listing from an alias-prefix realm
- [x] Verify `images` field is populated after listing creation completes

### DIFF
--- a/packages/host/app/commands/screenshot-card.ts
+++ b/packages/host/app/commands/screenshot-card.ts
@@ -73,6 +73,11 @@ export default class ScreenshotCardCommand extends HostBaseCommand<
       );
     }
 
+    // Resolve alias-form RRI to HTTP URL — the realm server does not know
+    // the alias mapping and will fail to construct a URL from alias form.
+    let vn = this.loaderService.loader.getVirtualNetwork()!;
+    cardId = vn.toURL(cardId).href;
+
     // Target realm = the card's own realm. If the caller can't write there,
     // fail fast — no silent fallback to a different realm.
     let cardRealm = this.realm.realmOf(rri(cardId));
--- a/packages/host/app/commands/screenshot-card.ts
+++ b/packages/host/app/commands/screenshot-card.ts
@@ -76,17 +76,17 @@ export default class ScreenshotCardCommand extends HostBaseCommand<
     // Resolve alias-form RRI to HTTP URL — the realm server does not know
     // the alias mapping and will fail to construct a URL from alias form.
     let vn = this.loaderService.loader.getVirtualNetwork()!;
-    cardId = vn.toURL(cardId).href;
+    let cardURL = vn.toURL(cardId).href;
 
     // Target realm = the card's own realm. If the caller can't write there,
     // fail fast — no silent fallback to a different realm.
-    let cardRealm = this.realm.realmOf(rri(cardId));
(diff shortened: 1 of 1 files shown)


lucaslyl:
@codex review

packages/host/app/commands/screenshot-card.ts
```diff
@@ -73,6 +73,11 @@ export default class ScreenshotCardCommand extends HostBaseCommand<
       );
     }
 
+    // Resolve alias-form RRI to HTTP URL — the realm server does not know
+    // the alias mapping and will fail to construct a URL from alias form.
+    let vn = this.loaderService.loader.getVirtualNetwork()!;
+    cardId = vn.toURL(cardId).href;
```
backspace: This should be a new variable (`cardURL`?) because it’s now a URL and never a `@cardstack/whatever`-style RRI. Then in the `throw`s later the URL won’t show there when it might not have come in as a URL.

backspace:
apart from that comment this seems good, I’ve opened an issue to eventually have `/_screenshot-card` and other realm server endpoints accept RRIs instead of just URLs

lucaslyl:
> apart from that comment this seems good, I’ve opened an issue to eventually have `/_screenshot-card` and other realm server endpoints accept RRIs instead of just URLs

Cool, I saw it . I linked the ticket to here as a future reference. https://linear.app/cardstack/issue/CS-11458/realm-server-endpoints-should-accept-rri-directly-so-clients-dont-need . From now on I'll just rename the variable name to cardURL. https://github.com/cardstack/boxel/pull/5147/commits/242fed9b93080b2e530640c79da205bc612bd957

richardhjtan (approved):

packages/host/app/commands/screenshot-card.ts
```diff
@@ -73,15 +73,20 @@ export default class ScreenshotCardCommand extends HostBaseCommand<
       );
     }
 
+    // Resolve alias-form RRI to HTTP URL — the realm server does not know
+    // the alias mapping and will fail to construct a URL from alias form.
+    let vn = this.loaderService.loader.getVirtualNetwork()!;
```
lukemelia, line 78: @backspace @lucaslyl this seems like an abstraction/encapsulation violation and breaks the rule "virtual networks should be invisible to most of the system and just rewrite from RRI to URL at the last moment before/after network access". Do we have a plan for a better way?

backspace: It’s true that this shouldn’t need to happen, a deeper solution is tracked in CS-11458

PR #5152, opened 2026-06-08
Unified search: shared render-type resolution rule

habdelra:
Phase 0 of [Unify prerendered-card-search and card-search](https://linear.app/cardstack/project/unify-prerendered-card-search-and-card-search-concepts-b6db4a4835ad) — the single, pure render-type resolution rule (Decision #1) that the server and host both call so they pick the same type. Additive; no behavior change on existing paths.

> **Stacked on #5150** (CS-11429) — it uses that PR's `SearchRenderType`. Review the top commit here; the base will be retargeted to `main` once #5150 merges.

### The rule — `resolveRenderType({ renderType, filterOn, types })` (runtime-common `search-utils`)
- an explicit `renderType` **CodeRef** → use it;
- the `"native"` escape valve → the result's own most-derived type (`types[0]`);
- omitted → the query's `filter.on` (the common ancestor searched on); when the query has no `filter.on`, fall back to the most-derived type.

It returns the `CodeRef` used both to select the HTML column and echo `meta.renderType` (server, CS-11433) and to drive the live `CardRenderer @codeRef` (host, CS-11434). The default is `filter.on` — an intended change from today's silent `types[0]` that only takes effect on the unified path; the legacy compat shim (CS-11433) preserves today's `types[0]`-unless-`renderType` behavior until callers migrate. The rule governs only the prefer-HTML path; `dataOnly` results are the actual types, so the request carries no render spec to resolve.

### Tests
Shared contract tests run via the realm-server runner (5 cases): explicit CodeRef wins; `"native"` → `types[0]`; omitted → `filter.on`; the missing-`filter.on` fallback; and that a `dataOnly` request carries no `render` (so the rule is inapplicable).


habdelra:
If you are only testing  on the realm server, then don't put teh test  in common. that pattern is only used for tests that need to execute in both host and realm -server

habdelra:
[Claude Code 🤖] Done — both suites moved out of `runtime-common/tests` into self-contained realm-server qunit modules (importing from `@cardstack/runtime-common`), shared files + consumers removed: render-type resolution here (2f2755164c) and the unified-search contracts on #5150 (b5f1def23a).

packages/runtime-common/search-utils.ts
```diff
@@ -408,6 +408,34 @@ export function parseUnifiedSearchRequestFromPayload(
   };
 }
 
+// The single render-type resolution rule (Decision #1), shared by the server
```
backspace: Is this “Decision #1” reference something we want in the comments forever?

backspace (approved):

packages/runtime-common/search-utils.ts
habdelra: [Claude Code 🤖] Agreed — no, it should not live in the comment forever. Removed in 1528dda4cf; the rule itself stays documented, just without the decision-number reference (comments describe the enduring contract, not the rollout). Swept the same kind of reference (a `CS-11433` mention) out of the stacked #5157 too.

PR #5164, opened 2026-06-09
fix: exempt catalog contents from no-erasable-decorator lint rule

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-11461/allow-decorators-in-catalog-gts-content-files

Lint failure at Pr: 
https://github.com/cardstack/boxel-catalog/actions/runs/27181574684/job/80241602204?pr=598
image

Root cause: The root .eslintrc.js globally bans decorators via no-restricted-syntax to protect Node CLI scripts from --experimental-strip-types. But contents/ files (both .ts card/command modules and .gts Glimmer components) always go through the realm/Ember compilation pipeline — decorators like @tracked, @action, and @field are valid there.

Fix: Added two overrides in packages/catalog/.eslintrc.cjs to exempt contents/**/*.ts and contents/**/*.gts from the no-restricted-syntax decorator restriction. The .ts and .gts files need separate overrides because .gts requires ember-eslint-parser while .ts uses the default TypeScript parser.

### DIFF
--- a/packages/catalog/.eslintrc.cjs
+++ b/packages/catalog/.eslintrc.cjs
@@ -3,6 +3,14 @@
 module.exports = {
   overrides: [
     {
+      // contents/ .ts files are Boxel card/command modules that go through the
+      // realm compilation pipeline — decorators are valid here.
+      files: ['contents/**/*.ts'],
+      rules: {
+        'no-restricted-syntax': 'off',
+      },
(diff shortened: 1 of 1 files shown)


lucaslyl:
@codex review

richardhjtan (approved):

PR #5175, opened 2026-06-09
Fix `ember-cli-deploy` deprecations

backspace:
This was a side quest from #5094 that turned out to be a red herring. But it silences deployment deprecation warnings like [this](https://github.com/cardstack/boxel/actions/runs/27153811839/job/80152110190#step:8:82).

### DIFF
--- a/packages/host/config/deploy.js
+++ b/packages/host/config/deploy.js
@@ -5,7 +5,6 @@ module.exports = function (deployTarget) {
     pipeline: {
       activateOnDeploy: true,
     },
-    plugins: ['build', 'smart-compress', 'revision-data', 's3', 'cloudfront'],
     build: {},
     s3: {
       allowOverwrite: true,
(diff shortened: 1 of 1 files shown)


packages/host/config/deploy.js
```diff
@@ -1,17 +1,38 @@
...
+      // (below) so browsers and freshly-booted prerender pages pick up a
+      // new build immediately after a deploy instead of pinning a stale
+      // `index.html` for ~2 years. Plugins are auto-discovered from the
+      // installed `ember-cli-deploy-*` addons (build, smart-compress,
+      // revision-data, s3, cloudfront), so no explicit plugin list.
+      alias: {
+        s3: { as: ['assets', 'html'] },
+      },
     },
-    plugins: ['build', 'smart-compress', 'revision-data', 's3', 'cloudfront'],
```
backspace, line 8: This produces a deprecation warning and needlessly lists all plugins, so it can be safely removed.

packages/host/config/deploy.js
```diff
@@ -29,20 +50,25 @@ module.exports = function (deployTarget) {
 
   if (deployTarget === 'build-only') {
     ENV.build.environment = 'production';
-    ENV.plugins = ['build'];
+    // Run only the build; skip compression, revision data, S3 upload and
+    // CloudFront invalidation.
+    ENV.pipeline.disabled = { allExcept: ['build'] };
```
backspace, line 32: This addresses a deprecation too, but uses the updated syntax to select desired plugins.

habdelra (approved):

PR #5182, opened 2026-06-10
Cache URL canonicalization in the runtime dependency tracker

habdelra:
## What

The runtime dependency tracker re-canonicalized the same module/instance URLs on every relationship read. A `linksToMany` of N same-typed cards re-tracked that card type's entire module-dependency graph once per element, so the pure string-canonicalization work — `canonicalURL` / `normalizeModuleURL` / `normalizeInstanceURL` / `hasPathExtension` — dominated the render of any card with a large linked-card graph.

This adds two session-scoped, string-keyed caches to `RuntimeDependencyTracker` that remove the redundancy without changing what gets recorded:

- **Memoize normalization** — `normalizeByKind` results go through a `Map<string, string>`, so a repeated URL collapses to a single map lookup instead of re-slicing the string and walking the executable-extension list. `#setRoot` and the consumer normalization route through it too.
- **Dedup track operations per context** — a `WeakMap<context, Set<kind\0url>>` short-circuits a `(kind, rawURL)` already recorded under a context. Recording is idempotent, so a repeat is a pure no-op. `#currentContext()` now returns a shared frozen `EMPTY_CONTEXT` singleton so the relationship walk a template render drives — which runs outside any `withContext` scope — dedups as well (a fresh `{}` each call would defeat the WeakMap).

Both caches clear in `reset()`, so they stay bounded to one render's distinct URLs and hold only strings (no `URL` objects) — memory stays flat across renders.

## Why it's correct

Normalization is a pure function of `(kind, rawURL)`, and a node record is fully determined by `(kind, rawURL, context)`. Deduping repeats of that triple under a context can never drop or alter a dependency. Cross-context recording is preserved because dedup is keyed per context object: a dep seen in both a query and a non-query context (distinct `withContext` frames) is still recorded under both, so query-only classification is unaffected.

## Tests

Added coverage in `runtime-dependency-tracker-test.ts` for the no-active-context (`EMPTY_CONTEXT`) path and for a dep tracked first with no context and then under an explicit context. The existing query-only / cross-context / session-reset assertions continue to pin the recording semantics.

Acceptance for the underlying performance issue is verified separately with the prerender CPU profiler — the `canonicalURL` frames should fall out of the top of the affected card profiles.

Closes CS-11480.

### DIFF
--- a/packages/runtime-common/dependency-tracker.ts
+++ b/packages/runtime-common/dependency-tracker.ts
@@ -129,11 +129,16 @@ function isPromiseLike<T = unknown>(value: unknown): value is Promise<T> {
 // template render drives). A fresh object literal each call would defeat that.
 const EMPTY_CONTEXT: RuntimeDependencyTrackingContext = Object.freeze({});
 
+// NUL separates the two parts of a composite cache key. It can't appear in a
+// node kind or a URL, so distinct (kind, rawURL) pairs never collide into the
+// same key.
+const CACHE_KEY_SEPARATOR = '\0';
+
 function normalizeCacheKey(
   kind: RuntimeDependencyNodeKind,
   rawURL: string,
 ): string {
-  return `${kind}\0${rawURL}`;
+  return `${kind}${CACHE_KEY_SEPARATOR}${rawURL}`;
 }
 
 export class RuntimeDependencyTracker {
--- a/packages/runtime-common/dependency-tracker.ts
+++ b/packages/runtime-common/dependency-tracker.ts
@@ -152,11 +152,15 @@ export class RuntimeDependencyTracker {
   // Recording a dependency is idempotent, so once a (kind, rawURL) has been
   // tracked under a given context it never needs to run again. A linksToMany of
   // N same-typed cards otherwise re-tracks that type's entire module graph once
-  // per element — O(N) redundant canonicalization. Context objects are built
-  // per render / field-read and never mutated afterward, so their identity is a
-  // sound dedup key; a WeakMap lets entries fall away with the contexts. Reset
-  // alongside #nodes so a cleared node map never leaves a stale "already
-  // tracked" marker that would drop a real dependency.
+  // per element — O(N) redundant canonicalization. This dedups only stack-top
+  // contexts: those frames are built by withContext() (or the shared frozen
+  // EMPTY_CONTEXT) and never mutated, so their identity soundly stands in for
+  // their fields. Caller-supplied explicit contexts are part of the public API
+  // and structurally mutable, so identity is NOT a safe key for them — they are
+  // never deduped (matching the previous short-circuit's `!explicitContext`
+  // guard). A WeakMap lets entries fall away with the contexts; reset alongside
+  // #nodes so a cleared node map never leaves a stale "already tracked" marker
+  // that would drop a real dependency.
   #trackedByContext = new WeakMap<
     RuntimeDependencyTrackingContext,
     Set<string>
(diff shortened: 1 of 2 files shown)


packages/runtime-common/dependency-tracker.ts
```diff
@@ -283,37 +307,52 @@ export class RuntimeDependencyTracker {
...
-      rawURL === this.#lastTrackRawURL &&
-      kind === this.#lastTrackKind &&
-      context === this.#lastTrackContext
-    ) {
+    // Already recorded this (kind, rawURL) under this context — nothing more to
+    // do (see #trackedByContext). A repeat would re-derive the identical node
+    // record, so skipping it is a pure no-op.
+    let seen = this.#trackedByContext.get(context);
+    if (seen?.has(key)) {
       return;
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch — fixed in 714a2a680a. The dedup is now gated to stack-top contexts only (withContext frames and the frozen `EMPTY_CONTEXT`), which are constructed once and never mutated; explicit contexts always record, restoring the exact `!explicitContext` guard the previous short-circuit had.

In the current tree the mutation isn't actually reachable — every explicit context comes from `runtimeQueryDependencyContext` / `runtimeNonQueryDependencyContext` (fresh literals) and `runtimeDependencyContextWithSource` returns a spread copy rather than mutating — but identity-based dedup of a public, mutable API object is the wrong invariant to rely on regardless, so the guard is worth keeping. The dominant relationship fan-out (a plain `linksToMany` walked via the no-explicit-context getter) is stack-top, so the dedup win is unaffected; explicit-context paths still get the normalization memo.

Added a regression test (`does not dedup a mutated explicit context`) that tracks the same dep under one object mutated `query` → `non-query` and asserts it isn't misclassified as query-only.

packages/runtime-common/dependency-tracker.ts
```diff
@@ -124,19 +124,47 @@ function isPromiseLike<T = unknown>(value: unknown): value is Promise<T> {
...
+// Shared identity for "no active context" so a context-keyed cache can dedup
+// tracking that happens outside any withContext scope (the relationship walk a
+// template render drives). A fresh object literal each call would defeat that.
+const EMPTY_CONTEXT: RuntimeDependencyTrackingContext = Object.freeze({});
+
+function normalizeCacheKey(
+  kind: RuntimeDependencyNodeKind,
+  rawURL: string,
+): string {
+  return `${kind}\0${rawURL}`;
```
lukemelia: Does this "\0" have some meaning, or is it just a separator?

lukemelia (approved):

packages/runtime-common/dependency-tracker.ts
habdelra: [Claude Code 🤖] It's a `\0` (NUL, U+0000) used purely as a separator. The reason it's NUL rather than, say, `|` or `:` is that NUL can't appear in either component — `kind` is one of `'module' | 'instance' | 'file'`, and URLs don't contain NUL — so the two parts can never run together ambiguously and distinct `(kind, rawURL)` pairs always map to distinct keys. No semantic meaning beyond that. Happy to swap it for a more obviously-a-delimiter character if you'd prefer readability over the collision guarantee.

lukemelia: Maybe a constant name or a comment to make the use clear.

habdelra: [Claude Code 🤖] Done in 50533c18c5 — pulled it out into a named `CACHE_KEY_SEPARATOR = '\0'` constant with a comment explaining why NUL (it can't appear in a node kind or a URL, so composite keys can't collide).

PR #5195, opened 2026-06-11
Unify thumbnail picker UI to match brand-logo MarkField

FadhlanR:
## Background and Goal

[CS-11350](https://linear.app/cardstack/issue/CS-11350) — the `CardInfo` editor's thumbnail picker shows two separate fields (a heavyweight `Thumbnail Image` linksTo upload and a `Thumbnail URL` raw string input). This PR collapses them into one brand-logo-style picker: a single URL input that switches between a `Select Image` button (when empty) and an `X` clear button (when filled). The disclosure button is renamed `Change Thumbnail`, and long URLs in the default-preview pane now wrap inside their cell instead of overflowing.

## Previews

Empty state:

empty

Filled state (X clear button replaces `Select Image`):

filled

Default preview pane with a long URL — wraps inside the cell, no overflow:

long-url

## Where to start

- `packages/base/default-templates/card-info.gts` — the editor template + the new `selectThumbnailImage` / `clearThumbnail` actions. Pattern mirrors `MarkFieldEdit` in `packages/base/brand-logo.gts`.
- `packages/host/tests/integration/components/card-info-thumbnail-picker-test.gts` — focused rendering tests covering both picker states, type-to-update, X clear, Select Image (stubbed file chooser), cancel, the computed-fallback placeholder, and the URL-overflow CSS.

- The `cardInfo.cardThumbnail` (`linksTo ImageDef`) field stays in the schema; only the editor stops surfacing it. `Select Image` writes the chosen file's URL into `cardInfo.cardThumbnailURL` (the string field), matching how `MarkField` works.
- The `Theme` picker keeps its current spot inside the same disclosure (only the disclosure button copy changes) — the ticket only scopes the thumbnail UX, so the theme flow is left as-is.
- URL overflow fix: `.default-preview [data-test-edit-preview='cardThumbnailURL'] { overflow-wrap: anywhere; min-width: 0; }` — the FieldContainer already sets `overflow-wrap: break-word` which doesn't break inside URLs (no word boundaries); `anywhere` is the right escalation, and `min-width: 0` lets the grid item shrink below its content's intrinsic size.


backspace (approved):

packages/base/default-templates/card-info.gts
```diff
@@ -318,30 +344,68 @@ class CardInfoEditor extends GlimmerComponent<EditSignature> {
...
+        font-size: var(--boxel-font-size-xs);
+        color: var(--muted-foreground, var(--boxel-400));
       }
       .null-preview {
         color: var(--muted-foreground, var(--boxel-450));
       }
+      .default-preview :deep([data-test-edit-preview='cardThumbnailURL']) {
+        overflow-wrap: anywhere;
+        min-width: 0;
+      }
```
burieberry, line 412: Oof, I think base package is the one place that doesn't have this linter -- I'll fix that. In the meantime, can't use `data-test-*` attribute in css because it will get wiped off in production. Use a regular `data-*` attribute.

burieberry (approved):
I couldn't see this in the staging preview for some reason, but the screenshots look right. There's a data-test-* css usage that needs to change that I commented on. Approved once that's fixed.

PR #5201, opened 2026-06-11
Picker: toggle off by id when options rebuild across renders

FadhlanR:
## Background and Goal

In the Realm filter pill on the card search bar, clicking an already-selected realm was adding a duplicate of that realm to the trigger instead of unselecting it. Same bug latent for any other consumer of `Picker` that doesn't keep `PickerOption` object references stable across renders.

Root cause: `Picker` wraps `BoxelMultiSelectBasic` → `PowerSelect`, whose multi-select toggle uses object identity (`indexOf`) to decide add-vs-remove. `RealmPicker` builds fresh `PickerOption` objects in both `realmOptions` and `pickerSelected` on every render, so the clicked option never matches the entry already in `@selected` by identity, and power-select always treats the click as a fresh add.

The fix lives in `Picker.onChange`: normalize the incoming `selected` array by id. A duplicate id signals "toggle off" — strip it. Otherwise dedupe defensively. Keeps the contract id-based and lets consumers reconstruct option objects freely.

## Where to start

- `packages/boxel-ui/addon/src/components/picker/index.gts` — the normalization block sits ahead of the existing select-all branching so the rest of `onChange` operates on a deduped array.
- `packages/boxel-ui/test-app/tests/integration/components/picker-test.gts` — new regression test that mirrors the RealmPicker pattern (consumer stores selection by id, rebuilds options each render). Without the fix the second click stacks a duplicate; with the fix it toggles off.

### DIFF
--- a/packages/boxel-ui/test-app/tests/integration/components/picker-test.gts
+++ b/packages/boxel-ui/test-app/tests/integration/components/picker-test.gts
@@ -254,11 +254,11 @@ module('Integration | Component | picker', function (hooks) {
       .doesNotExist('Option 1 checkbox is unchecked after deselecting');
   });
 
-  // Regression: CS-11321. Consumers like RealmPicker rebuild PickerOption
-  // objects on every render rather than maintaining stable references.
-  // Power-select's identity-based toggle then fails to recognize an
-  // already-selected option, so a second click used to append a duplicate
-  // instead of unselecting. Picker.onChange normalizes by id to fix this.
+  // When a consumer reconstructs PickerOption objects on every render
+  // rather than keeping object references stable, power-select's
+  // identity-based multi-select toggle can't match the clicked option
+  // against @selected and emits a duplicate-id selection. Picker
+  // normalizes by id so a duplicate signals a toggle-off.
   test('picker toggles selection by id when consumer rebuilds option objects each render', async function (assert) {
     class SelectionController {
       @tracked selectedIds: string[] = [];
--- a/packages/boxel-ui/test-app/tests/integration/components/picker-test.gts
+++ b/packages/boxel-ui/test-app/tests/integration/components/picker-test.gts
@@ -254,6 +254,80 @@ module('Integration | Component | picker', function (hooks) {
       .doesNotExist('Option 1 checkbox is unchecked after deselecting');
   });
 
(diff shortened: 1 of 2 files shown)


packages/boxel-ui/test-app/tests/integration/components/picker-test.gts
```diff
@@ -254,6 +254,80 @@ module('Integration | Component | picker', function (hooks) {
       .doesNotExist('Option 1 checkbox is unchecked after deselecting');
   });
 
+  // Regression: CS-11321. Consumers like RealmPicker rebuild PickerOption
```
backspace: We have the `evergreen-comments` skill to discourage embedding Linear issues in comments, which are (almost?) always unhelpful

backspace (approved):
Image

this is a case where screenshots/recordings would have been helpful to see the before and after 🤓

I wonder if some more thinking is needed because the whole list jumps when you add an item, so if you want to add a few in a row, you need to track them as they move. Maybe they don’t need to be added to the top list until you leave and come back? But I don’t know what best practice is here, and this could be a future improvement.

Another future improvement: the search icon is overlapping with the placeholder text

burieberry (approved):

FadhlanR:
[Claude Code 🤖]

Thanks for the review @backspace. Both future-improvement notes are tracked separately so this PR can stay scoped to the toggle-off bug:

- [CS-11533](https://linear.app/cardstack/issue/CS-11533/realm-picker-list-jumps-when-adding-an-item-defer-reordering-until) — list jumps when adding an item; defer reordering until reopen
- [CS-11534](https://linear.app/cardstack/issue/CS-11534/realm-picker-search-icon-overlaps-placeholder-text-in-dropdown) — search icon overlaps the placeholder text

The unrelated `Realm Server Tests (4, 6)` failure (`skip-query-backed-expansion-test.ts`) and the `Host Memory Baseline` check have been re-run.

packages/boxel-ui/test-app/tests/integration/components/picker-test.gts
FadhlanR: [Claude Code 🤖]

Good catch — fixed in 35438ddac7. Rewrote the comment to describe the upstream invariant the test verifies, without the ticket reference or the "used to / to fix this" framing.

PR #5211, opened 2026-06-11
Don't steal focus from text entry when placing the Monaco cursor

habdelra:
In code submode, cursor placement runs on debounced timers: the monaco modifier restores a cursor position some time after a file opens or a definition is selected, and `MonacoService.updateCursorPosition()` unconditionally called `editor.focus()` when it finally fired. If the user had meanwhile started typing somewhere else — most commonly the AI assistant chat box — that delayed `focus()` yanked their keystrokes into the code editor, and the editor's auto-save then persisted the stray text into the open file, silently corrupting a source module.

This is also the mechanism behind a flaky cascade in the matrix Playwright suite ([example run](https://github.com/cardstack/boxel/actions/runs/27372997674/job/80890520131)). In `messages.spec.ts`, "currently viewed file is auto-attached" clicks a definition (which schedules a cursor placement at the class body of `person.gts`, line 17) and then fills the chat input. Playwright's `fill()` focuses the textarea in-page and then inserts the text via a separate CDP call into whatever element is focused *at that moment*. When the debounced cursor update fired inside that window, the chat sentence "Are there any computed fields in person.gts?" was inserted into the monaco buffer at line 17 col 36, auto-saved to the realm (`Parse Error at person.gts:17:40: 17:45` — "there" lands at columns 40–45), and every Person instance in the shared test realm went into an error state — failing 17 downstream tests for the rest of the shard.

The fix: `updateCursorPosition()` still places and reveals the cursor unconditionally, but only takes focus when doing so won't interrupt text entry elsewhere. If the active element is an input, textarea, or contenteditable outside the editor's DOM, focus is left alone and a `service:monaco` warning is logged — a diagnostic breadcrumb so any future trace shows whether this path fired.

The card URL bar is the one text-entry element that *hands off* focus to the editor: submitting a valid URL with Enter commits the navigation, so the bar blurs itself before the navigation starts. By the time the editor's cursor placement fires — at whatever point during the file load — the bar has already relinquished focus and the editor takes it as before.

Two acceptance tests cover the guard: a cursor update leaves focus on an external textarea while still moving the editor cursor, and still focuses the editor when no text-entry element has focus.

### DIFF
--- a/packages/host/app/services/monaco-service.ts
+++ b/packages/host/app/services/monaco-service.ts
@@ -24,7 +24,7 @@ import HTMLWorkerUrl from 'monaco-editor/esm/vs/language/html/html.worker.js?wor
 import JSONWorkerUrl from 'monaco-editor/esm/vs/language/json/json.worker.js?worker&url';
 import TSWorkerUrl from 'monaco-editor/esm/vs/language/typescript/ts.worker.js?worker&url';
 
-import type { SingleCardDocument } from '@cardstack/runtime-common';
+import { logger, type SingleCardDocument } from '@cardstack/runtime-common';
 
 import config from '@cardstack/host/config/environment';
 import type CardService from '@cardstack/host/services/card-service';
@@ -43,6 +43,8 @@ export type MonacoSDK = typeof _MonacoSDK;
 export type IStandaloneCodeEditor = _MonacoSDK.editor.IStandaloneCodeEditor;
 
 const { serverEchoDebounceMs } = config;
+
+const log = logger('service:monaco');
 
 // `new Worker(url)` rejects cross-origin URLs, but importScripts inside a
 // worker is allowed to fetch them. When the worker URL is on a different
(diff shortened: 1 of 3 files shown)


packages/host/app/services/monaco-service.ts
```diff
@@ -335,7 +337,30 @@ export default class MonacoService extends Service {
...
+    // fires the user may already be typing in another text-entry element,
+    // like the AI assistant chat box. Grabbing focus then would redirect
+    // their keystrokes into the editor, where auto-save would persist the
+    // stray characters into the open file. Take focus only when it won't
+    // interrupt text entry happening elsewhere.
+    let active = document.activeElement;
+    let editorDom = this.editor.getDomNode();
+    let activeElementAcceptsText =
+      active instanceof HTMLElement &&
+      !(editorDom?.contains(active) ?? false) &&
```
FadhlanR: I think !(editorDom?.contains(active)) is enough. We can remove ?? false.

FadhlanR (approved):

packages/host/app/services/monaco-service.ts
habdelra: [Claude Code 🤖] Good catch — `!undefined` already lands on `true`, so the fallback was a no-op. Removed in a88cfe98e0.

richardhjtan (approved):

PR #5231, opened 2026-06-15
fix: re-enable erasable-syntax guards in catalog contents, lifting only Decorator (CS-11461)

richardhjtan:
Follow-up to #5164.

## Background
#5164 hot-fixed a catalog CI lint failure: boxel's `no-restricted-syntax` ban gained a `Decorator` selector (commit `3af153c3ab`, the erasable-syntax guard for the ts-node → native Node migration / CS-11449), and it fired on every `@field`/`@tracked`/`@action` in catalog `contents/`. The fix set `'no-restricted-syntax': 'off'` for all of `contents/**`.

That also silently un-banned the other four non-erasable constructs — `enum`, `import =`, `export =`, and runtime `namespace`/`module` blocks — which have no more reason to be allowed in card modules than anywhere else.

## What this does
Narrows the catalog override to lift **only** the `Decorator` selector. Decorators are permanently valid in `contents/` (everything there goes through the realm/Ember compilation pipeline; `@field` is the card API), so this is the correct end state — not a grandfather. The remaining four guards now apply to contents too, matching the rest of the repo.

No file-by-file exception list is needed: catalog contents have **zero** violations of the four remaining selectors today (verified by layering the selectors back via `--rule` against freshly synced contents).

The `NO_COMPILATION_REQUIRED_TS_SELECTORS` list was duplicated verbatim in the root and `ai-bot` configs; the catalog override needs to filter it, so it's extracted to `eslint/erasable-syntax-selectors.cjs` rather than adding a third copy. The `require()` in the root and `ai-bot` `.js` configs needs the standard self-override (cf. `packages/postgres/.eslintrc.js`) to satisfy `no-var-requires`.

## Verification
- `pnpm lint:js` in `packages/catalog` (against synced `contents/`) — clean.
- Negative test: a scratch `enum` in a `contents/*.ts` file errors with the enum message; `@field` decorators stay clean.
- `eslint` on the root and `ai-bot` configs — clean (exit 0); selector lists resolve identically to before (root/ai-bot keep all 5, catalog drops only `Decorator`).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,39 +1,8 @@
 'use strict';
 
-// Selectors for TypeScript syntax that Node cannot run via
-// `--experimental-strip-types`, which only handles "erasable" TS —
-// syntax that vanishes once type annotations are stripped. New code
-// must avoid these so it can run under Node natively, without ts-node.
-const NO_COMPILATION_REQUIRED_TS_SELECTORS = [
-  {
-    selector: 'TSEnumDeclaration',
-    message:
-      'TypeScript `enum` is not erasable and requires compilation. Use a `const` object with `as const` (or a union of string literals) instead.',
-  },
-  {
-    selector: 'TSImportEqualsDeclaration',
-    message:
-      '`import =` syntax requires TypeScript compilation. Use standard ES module `import` instead.',
-  },
-  {
-    selector: 'TSExportAssignment',
-    message:
-      '`export =` syntax requires TypeScript compilation. Use a standard ES module `export default` (or named exports) instead.',
-  },
-  {
-    selector: 'Decorator',
-    message:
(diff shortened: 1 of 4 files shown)


richardhjtan:
### Verified against real catalog contents

I ran **boxel-catalog CI** with its `ci-lint.yaml` boxel checkout temporarily pinned to this branch (`cs-11461-narrow-catalog-no-restricted-syntax`), which runs `pnpm run lint` in `packages/catalog` against the actual contents.

✅ **Lint passed** — https://github.com/cardstack/boxel-catalog/actions/runs/27531451149/job/81370311776 (3m58s)

This confirms that with the narrowed override (only `Decorator` lifted, the `enum`/`import =`/`export =`/`namespace` guards re-enabled), real catalog contents lint clean: decorators are accepted and there are zero violations of the re-enabled selectors.

lukemelia (approved):
These rules are really most important for code that will run in node, which catalog contents should never.

PR #5240, opened 2026-06-15
Redesign markdown editor toolbar: docked sticky bar + view-selector dropdown

lukemelia:
## Summary

Replaces the selection-triggered floating toolbar and the three-button mode switcher in the RichMarkdown editor with a single **persistent, sticky toolbar** docked at the top of the field. A **view-selector dropdown** (Compose / Source / Preview) sits on the left and is always enabled; the formatting controls sit to its right and are **disabled until the editor has focus**.

Control set and order are unchanged: Bold, Italic, Strikethrough, Code, Link │ H1, H2, H3 │ Bullet, Numbered, Blockquote.

## Preview

Docked toolbar — formatting controls disabled until the editor has focus:

Docked toolbar, closed view selector

View selector open (compact, rounded):

View selector open

Toolbar is sticky:

Toolbar is sticky

## Changes

- **host `codemirror-context.ts`** — add `hasFocus` to `SelectionInfo`, report it on focus change, and make `wrapWith` insert an empty marker pair with the cursor centered when there is no selection (improves toolbar clicks and the `Mod-b`/`Mod-i` keymap).
- **base `view-selector.gts`** (new) — compact `BoxelSelect` wrapper for the three views; always enabled.
- **base `codemirror-editor.gts`** — remove the floating-toolbar machinery (and the `@floating-ui/dom` usage); render a sticky toolbar as the first child with a `:leadingControls` slot; focus-gate the formatting buttons; insert empty link syntax when there is no selection.
- **base `rich-markdown.gts`** — sticky toolbar layout; render the view selector into the editor's `:leadingControls` for compose/source and in a standalone bar for preview.
- **boxel-ui `select/`** — tokenize the trigger padding as `--boxel-select-trigger-padding` (the gap was already `--boxel-select-trigger-gap`) so the trigger can be compacted without a bespoke trigger component; both are documented in `select/usage.gts`. Existing `BoxelSelect` consumers are unaffected (the previous values remain the defaults).

The Add-a-card / Add-a-file menu, embed modals, fitted-format selector, and file embedding remain out of scope for this change.

## Testing

- `codemirror-editor-test`: empty-selection `wrapWith` now inserts cursor-centered markers; existing wrap-with-selection tests retained.
- `rich-markdown-field-test`: mode-button tests replaced with view-selector interactions, including focus-gated enablement and switching to/from preview.
- Type-check (`ember-tsc`) and `ember-template-lint` pass; verified the docked toolbar, compact dropdown, focus-gating, and open-state corner rounding in the running playground.


burieberry (approved):
Nice improvements over the previous iteration.

Improvement suggestions:
- I noticed the code-mirror editor doesn't focus when anywhere in the text area in clicked, making one question if it's an editable area. It only becomes active when the first line is clicked. Would be good to follow our other input/textarea conventions for hover and focus behavior and styles if possible.

Other concerns:
- There's another ViewSelector component in boxel-ui (card/list/grid view toggle) with similar type names and view_options argument. Maybe we should differentiate the naming more. The base components directory might be a better home.
- Theming. The base package templates are theme-able. I recommend checking the new components in a dark colored theme, which makes errors obvious. Most of the styles can be inherited as is by not specifying them. Hover and selected states may need adjustment. `--primary` and `--primary-foreground` for selected state, something like 5-10% of currentColor for hover background. Our dropdown theming is meh at best with tickets in the backlog, but legibility would be good.

PR #5242, opened 2026-06-15
Unified search: render-store readiness for getSearchEntriesResource

habdelra:
Part of **Unify prerendered-card-search and card-search**. Stacked on #5239 (retarget base to `main` once that lands).

A card that renders `@context.searchResultsComponent` in a prerenderable template drives `getSearchEntriesResource` through the render route. The render route's settle loop (`routes/render.ts` `#waitForRenderLoadStability`) only awaits the **render** store's `loaded()` / `loadGeneration` — but `SearchEntriesResource` tracked its in-flight v2 search solely via an `@ember/test-waiter`, which that loop never observes. The html-only result branch deposits nothing into the store, so nothing moved the load generation either, and a prerender of such a card could capture its HTML before the search resolved: an empty result list that only fills in later, in the live SPA.

The resource now resolves the render store while prerendering (gated on `__boxelRenderContext`, the same signal `prerendered-card-search` uses) and routes its v2 search fetch and a `store.trackLoad()` of every search through it. `trackLoad` bumps the load generation the moment a search starts and keeps the load pending until it resolves, so the settle loop waits for results before HTML capture — mirroring how the v1 `SearchResource` already wires render-context searches. Outside a prerender, `runtimeStore` is the injected live store and behavior is unchanged.

### Tests

- **Integration (`search-entries-test.gts`)** — the deterministic contract: under `__boxelRenderContext`, performing the search registers exactly one load on the render store, that load is the in-flight search, and the render store's `loadGeneration` advances. (An acceptance-level capture can't isolate this — the resource's test-waiter makes `await visit` wait for the search regardless of the readiness wiring; the real puppeteer prerenderer has no such settle gate.) Verified red without the `trackLoad` call, green with it.
- **Acceptance (`prerender-search-results-test.gts`)** — end-to-end: a card renders the v2 surface in its isolated template and prerenders with both results' prerendered HTML present, not an empty list.

This is a prerequisite for the codemod (CS-11439) migrating cards that render search in a `fitted` / `embedded` template, and for the Phase 2 query/collection-field prerendered-HTML work.


packages/host/tests/acceptance/prerender-search-results-test.gts
habdelra, line 1: host tests wont work--host tests bypass the render route--the very thing we want to test here. host test test thru card-prerender. you have a memory about this--cmon!?!

packages/host/tests/integration/resources/search-entries-test.gts
habdelra, line 1: same comment as above--host tests are the wrong place for this test. we are tyring to test the inside of the /render route. host tests skip that entirely

packages/host/tests/acceptance/prerender-search-results-test.gts
habdelra: [Claude Code 🤖] Agreed — moved the end-to-end coverage out of host tests. Host indexing runs through `card-prerender`, not the `/render` route, and `visit('/render/…')` + `capturePrerenderResult` is masked by `settled()` (the resource's test-waiter makes `await visit` wait for the search regardless of the readiness wiring). The faithful test now lives in the realm-server prerendering suite next to the live-search-fallback cases: a card renders `@context.searchResultsComponent` in its isolated template and prerenders with its result present. The host acceptance test is removed.

packages/host/tests/integration/resources/search-entries-test.gts
habdelra: [Claude Code 🤖] Kept this one per your follow-up: it's orthogonal to the route. It sets `__boxelRenderContext` by hand and asserts the *resource* registers its in-flight search with the render store (`trackLoad` + `loadGeneration` bump) — it never drives the `/render` route. That's the "control the inputs directly" host-test case from your test-placement rule. The actual "/render waits before capture" behavior is now the server prerendering test.

packages/host/app/resources/search-entries.ts
```diff
@@ -115,6 +116,41 @@ export class SearchEntriesResource extends Resource<Args> {
...
+  // prerender the render route settles on the *render* store's `loaded()` /
+  // `loadGeneration` (`routes/render.ts` `#waitForRenderLoadStability`), while a
+  // component's injected `store` service is the live SPA store — so the search
+  // fetch and its load-tracking must run against the render store for the
+  // prerender to wait for results before HTML capture. Outside a prerender
+  // (the live SPA) this is the injected store. Mirrors how
+  // `prerendered-card-search` and the v1 `SearchResource` route render-context
+  // work through the render store.
+  private get runtimeStore(): StoreService {
+    if ((globalThis as any).__boxelRenderContext) {
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Adopted — the gate is now `__boxelRenderContext === true`, matching the prerender header / job-priority helpers so store selection and header emission agree. (In practice the render route only ever sets it to `true`, so the truthy-non-boolean split isn't reachable today, but the strict gate is the right consistency.)

packages/host/tests/integration/resources/search-entries-test.gts
```diff
@@ -203,6 +203,65 @@ module('Integration | search-entries resource', function (hooks) {
...
+      await load;
+      assert.strictEqual(
+        search.entries.length,
+        2,
+        'the tracked load is the real search and it produced results',
+      );
+    } finally {
+      delete (renderStore as any).trackLoad;
+      delete (globalThis as any).__boxelRenderContext;
+    }
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] `trackLoad` is a `StoreService` prototype method, so the spy is an own property and `delete` restores the inherited method — that's the correct restoration here (re-assigning would leave an own-property shadow over the prototype). Added a comment noting this. The own-property-in-future scenario is a state that doesn't occur today.

backspace (approved):

richardhjtan (approved):

PR #5249, opened 2026-06-16
Apply realm rename to UI without a browser reload

lukemelia:
## CS-11586

Renaming a realm (editing its `realm.json` RealmConfig card) didn't update the realm name in the workspace chooser label or the index card title until the browser was refreshed. ([CS-11586](https://linear.app/cardstack/issue/CS-11586/realm-rename-not-applied-until-refresh))

## Why

The realm name reaches the UI through two independent paths, neither of which refreshed on a re-index:

1. **Workspace chooser label** reads `RealmService.info(url).name`, which was cached once and never refreshed. (The "In *Realm*" card-header tooltip reads the same `RealmService.info(url).name`, so it's fixed by the same change — covered transitively, not separately asserted.)
2. **Index card title** (and the browser/page title) reads the loaded CardsGrid card's `[meta].realmInfo` — injected at request time by the realm server. A `realm.json` change only invalidates the RealmConfig card (`<realm>/realm`), never the index card, so the host never re-fetched it.

## Changes

- **`services/realm.ts`** — On an incremental re-index that invalidates the RealmConfig card (the only signal a rename produces), `RealmResource` re-fetches `_info` and merges it into the existing tracked `info` object **in place** (stable reference for `@cached` getters). Deliberately narrow: `full`/`copy` re-indexes and incrementals of unrelated cards are left alone, so a refresh can't clobber client-managed publish state (`lastPublishedAt`), which is additionally preserved across the merge as defense in depth. Fetch logic extracted into a reusable `fetchInfoFromServer()`.
- **`services/store.ts`** — In `handleInvalidations`, when the RealmConfig card is invalidated, reload the realm's index card (`<realm>/index`) if loaded. The fresh fetch re-injects `realmInfo`, and `updateFromSerialized` → `notifyCardTracking` makes the title recompute.

## Tests

- Workspace chooser label updates after a rename.
- Index card title + page title update after a rename (index-card reload path).
- **Guard:** an unrelated card re-index preserves client-managed publish state (locks the invariant behind the narrow gate).

## Verification

- `pnpm lint:types`: 0 errors; `eslint` on changed files: clean.
- Acceptance tests added; full host suite + host-submode publish/unpublish tests pass in CI.

### DIFF
--- a/packages/host/app/services/realm.ts
+++ b/packages/host/app/services/realm.ts
@@ -487,9 +487,11 @@ class RealmResource {
       return;
     }
     let { info, isPublic } = await this.fetchInfoFromServer();
-    // Preserve client-managed publish state: publish()/unpublish() own
-    // `lastPublishedAt`, and `_info` may not yet reflect a just-published
-    // realm, so a rename refresh must not clobber it.
+    // `_info` always returns the full RealmInfo attribute set (unset fields
+    // come back as explicit `null`), so assigning onto the existing object
+    // can't strand a stale key. Preserve client-managed publish state:
+    // publish()/unpublish() own `lastPublishedAt`, and `_info` may not yet
+    // reflect a just-published realm, so a rename refresh must not clobber it.
     Object.assign(this.info, info, {
       isPublic,
       lastPublishedAt: this.info.lastPublishedAt,
--- a/packages/host/app/services/realm.ts
+++ b/packages/host/app/services/realm.ts
@@ -312,6 +312,20 @@ class RealmResource {
                 indexingWaiter.endAsync(this.indexingWaiterToken);
                 this.indexingWaiterToken = null;
               }
+              // A re-index may have changed realm info (e.g. the realm was
+              // renamed via its RealmConfig card). Refresh so the workspace
+              // chooser label and index card title update without a reload.
+              // For incremental indexing this only matters when the realm's
+              // config card was invalidated, so skip the refetch otherwise.
+              // Instance invalidations carry the card id (no `.json`), so the
+              // RealmConfig card at `<realm>/realm.json` appears as
+              // `<realm>/realm`.
+              if (
+                data.indexType !== 'incremental' ||
+                data.invalidations.includes(`${this.realmURL}realm`)
+              ) {
+                this.refreshInfo();
+              }
               break;
             default:
               throw assertNever(data);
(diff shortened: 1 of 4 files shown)


lukemelia:
## Code Review

### Overview
Fixes a UX gap where renaming a realm (editing its `realm.json` RealmConfig card) required a browser reload to update the workspace chooser label and index-card title. The PR addresses the two independent reactivity gaps: (a) `RealmResource.info` was cached once and never refreshed, and (b) the index card's `meta.realmInfo` is injected at request time and only the RealmConfig card—not the index card—is invalidated on a config edit. The split-fix mirrors the split data path, which is the right shape for the bug.

### Strengths
- **Narrow gating.** `refreshInfo()` only fires on `indexType === 'incremental'` *and* `invalidations.includes('${realm}realm')`. This deliberately avoids clobbering client-managed state (`lastPublishedAt`) on unrelated re-indexes—well-thought-out, and the comment captures the reasoning.
- **Stable object reference.** `Object.assign(this.info, info, …)` mutates the existing `TrackedObject` rather than replacing it, preserving identity for any `@cached` getters that closed over it. Matches the pattern used elsewhere when toggling `isIndexing`.
- **Refactor is non-invasive.** Extracting `fetchInfoFromServer()` from the dropTask makes refresh trivially reusable without changing the initial-fetch semantics (still wrapped in `fetchInfoTask`/`fetchingInfo` de-dup).
- **Test coverage hits both data paths.** Acceptance tests for the workspace chooser label and the index card title (incl. page title) — the user-visible regression-prone surfaces.

### Issues & Suggestions

**1. `Object.assign` can leave stale keys behind.** (`packages/host/app/services/realm.ts:495`)
If a future `_info` response omits a property that the tracked `info` previously had (e.g. a user clears `iconURL` and the server returns the attributes set without it), the old value will linger. The original `fetchInfoTask` builds a fresh `TrackedObject`, so this is a new asymmetry. Probably fine for the rename-only use case here, but worth a comment noting the assumption, or an explicit reset of the known optional keys before assignment.

**2. Redundant refresh after `setRealmInfoProperty`.** When `setRealmInfoProperty` PATCHes `_config`, it both updates `this.info` directly *and* triggers an indexing event that will land `<realm>/realm` in `invalidations` → `refreshInfo()` re-fetches. Harmless (idempotent), but consider gating with `clientRequestId` (the existing pattern in `store.handleInvalidations`) if you want to suppress the double fetch.

**3. No negative test for the gating.** The two new tests show that a rename does propagate. There's no test asserting that an *unrelated* card edit in the same realm does **not** trigger `refreshInfo()` (i.e. that publish state is preserved). Given that the PR's whole reason for the narrow gate is to avoid that case, a guard test would lock the invariant in.

**4. Tooltip claim not asserted.** PR description mentions the "In *Realm*" header tooltip refreshes, but no test asserts on it. It's transitively covered by the workspace-chooser path (same `RealmService.info(url).name`), so this is a doc-vs-test mismatch — either trim the PR description or add an explicit tooltip assertion.

**5. Module-scoped `let realmA: Realm;`.** (`workspace-chooser-test.gts:37`)
File-scope mutable across tests; relies on `beforeEach` to always reassign. Works today, but if a future test path skips setup it'd silently see the prior realm. Threading it through test context (`this.realmA = realm`) would be safer.

**6. Minor: comment about `.json`-stripping invalidation IDs.** The comment in `services/realm.ts` ("Instance invalidations carry the card id without `.json`") is accurate and helpful — could you mirror it as a one-liner in `services/store.ts` next to `${event.realmURL}realm` for the next reader who lands there cold?

### Risk
Low. Gating is narrow, the refresh is idempotent, and the change is additive (no behavior changes for non-rename paths). Worst-case failure mode is a missed refresh, which is the pre-PR baseline.

### Verification
Linting + types pass per PR notes. Recommend also running the publish/unpublish acceptance tests locally to confirm the `lastPublishedAt` preservation in `refreshInfoTask` (the comment claims it, and the code looks right, but the rename + publish interaction isn't directly covered).

lukemelia:
[Claude Code 🤖] Thanks for the review. Resolution (pushed in a33df48):

- **1 — `Object.assign` stale keys:** Non-issue in practice — `_info` always returns the full `RealmInfo` attribute set (unset fields come back as explicit `null`), so no key is ever omitted. Added a comment in `refreshInfoTask` documenting that assumption.
- **2 — Redundant fetch after `setRealmInfoProperty`:** Acknowledged, leaving as-is. It's one idempotent `_info` GET on icon/background quick-set; the rename flow (editing the config card in the editor) doesn't hit that path, and `clientRequestId` gating adds complexity for no correctness gain.
- **3 — Negative/guard test:** Added — a workspace-chooser test asserts an unrelated card re-index preserves client-managed publish state, locking the invariant behind the narrow gate.
- **4 — Tooltip doc-vs-test mismatch:** Trimmed the PR description to state the header tooltip is covered transitively (same `RealmService.info().name` path), not separately asserted.
- **5 — Module-scoped `let realmA`:** Leaving as-is — it matches the established convention in the same suite (`operator-mode-acceptance-test.gts` uses an identical module-scoped `let testRealm: Realm`). Changing one file in isolation would add inconsistency.
- **6 — Mirror the `.json`-stripping comment in `store.ts`:** Done.

Both previously-failing shards (Host Tests 2 and 9, incl. publish/unpublish) are green; a fresh run is validating the guard test.

backspace (approved):
I tried this in the preview deployment, works as described!

PR #5256, opened 2026-06-17
Add BFM :file references for MarkdownDef

lukemelia:
## Summary

Adds `:file[URL]` (inline) and `::file[URL]` (block) BFM syntax for embedding `FileDef` instances in markdown. Because `FileDef extends BaseDef` (not `CardDef`), `:card` cannot match files — this introduces a dedicated `file` keyword on top of the keyword-generic BFM infrastructure from CS-10570. File references resolve everywhere card references do.

Resolves [CS-10583](https://linear.app/cardstack/issue/CS-10583).

## What changed

- **marked-sync**: register `bfmExtensionsForKeyword('file')` alongside `card`, so `:file` / `::file` produce `data-boxel-bfm-*` placeholders with `data-boxel-bfm-type="file"`.
- **runtime-common**: add an `extractFileReferenceUrls` convenience wrapper mirroring `extractCardReferenceUrls`.
- **MarkdownDef**: add a computed `fileReferenceUrls` field and a query-backed `linkedFiles` field (`linksToMany(FileDef)` filtered by `in: { id }`), populate `fileReferenceUrls` in `extractAttributes`, and pass `@linkedFiles` to `MarkdownTemplate`.
- **MarkdownTemplate**: the slot-capture modifier scans both `card` and `file` placeholders. Resolved file refs render inline in atom format and block in embedded format; the file name (last path segment) is the broken-link label.
- **RichMarkdownField**: mirror the `fileReferenceUrls` + `linkedFiles` fields and pass `@linkedFiles` to its rendered views and to the editor.
- **CodeMirror editor** (`codemirror-context` + `CodeMirrorEditor`): scan card and file keywords via a shared config, tag each widget with `data-card-type`, carry `refType` through widget/render targets, resolve files via `linkedFiles` plus the shared `getCards` query (which returns file-meta for file ids), and render file widgets with the file component. `:file` refs resolve in compose-mode live preview.
- **Preview-panel `RenderedMarkdown`** (the "Rendered" toggle): load file-meta via `store.get(url, { type: 'file-meta' })`, scan/resolve file placeholders, and render resolved files with `CardRenderer` (inline atom / block embedded); unresolved files are labeled by file name.

## Demo

- **RichMarkdownPlayground** (`rich-markdown-playground-1.json`): a "File References" section embedding an image (inline + block), a CSV, and a linked markdown document — resolves in the rendered card views, the editor's compose-mode live preview, and the preview panel.
- **`bfm-showcase.md`**: a matching section for the `MarkdownDef` rendering path.

Referenced files are existing experiments-realm fixtures; images render via `ImageDef`'s atom/embedded templates.

## Testing

- `pnpm lint:types` (glint) clean across host + base; `ember-template-lint` clean on changed `.gts`; eslint clean on changed `.ts`.
- Unit coverage for the `:file` markedToHtml path and `extractFileReferenceUrls`.
- Acceptance coverage for resolved inline/block file rendering and the unresolved fallback (MarkdownDef path).
- Editor integration coverage: `:file` / `::file` produce file widget targets (with `refType`/format) and coexist with card refs.
- Preview-panel integration coverage: `:file` / `::file` placeholders and the unresolved file-name fallback.

Full browser test suites run in CI (the host harness isn't runnable locally), hence draft.


packages/base/default-templates/markdown.gts
```diff
@@ -49,21 +54,42 @@ const markdownRenderingWaiter = buildWaiter('markdown-rendering');
...
+  card?: CardDef; // present when refType === 'card' && state === 'resolved'
+  file?: FileDef; // present when refType === 'file' && state === 'resolved'
   url?: string; // present when state === 'loading' | 'unresolved'
   typeName?: string; // present when state === 'unresolved'
 }
 
+// For a `:file[URL]` ref the human-readable label is the file name (the last
+// path segment), unlike card refs whose type name is the second-to-last
+// segment (`<base>/<TypeName>/<id>`).
+function fileNameFromUrl(url: string): string {
```
lukemelia: **The Intern:** This `fileNameFromUrl` helper is byte-identical to the one I noticed at `packages/host/app/components/operator-mode/preview-panel/rendered-markdown.gts:73`. Since it's pure string slicing with no DOM/store deps, would it make sense to lift it next to `cardTypeName` (or into `runtime-common/bfm-card-references.ts` alongside the extractors) so both call sites import the same implementation? Right now if someone fixes a URL edge case in one they'll likely miss the other.

packages/base/default-templates/markdown.gts
```diff
@@ -188,17 +217,20 @@ export default class MarkDownTemplate extends GlimmerComponent<{
   captureCardSlots = modifier(
     (element: HTMLElement, _positional: unknown[]) => {
       let linkedCards = this.args.linkedCards;
+      let linkedFiles = this.args.linkedFiles;
```
lukemelia, line 209: **The Intern:** Small reactivity question — the modifier reads `this.args.linkedFiles` here, but the invocation at line 521 only threads `@linkedCards` through the positional args:

```hbs
{{this.captureCardSlots this.renderedHtml @linkedCards}}
```

The sibling `rendered-markdown.gts` passes both `this.loadedCards` and `this.loadedFiles` as positionals (see its diff around line 405). Even if Glimmer auto-tracks the `this.args.linkedFiles` read inside the modifier body, the codebase's existing convention here is to pass everything the modifier consumes as positional args so the re-run trigger is explicit. Worth at least making the two call sites consistent — would adding `@linkedFiles` as a third positional be reasonable?

packages/host/app/lib/codemirror-context.ts
```diff
@@ -118,19 +134,25 @@ class CardWidget extends WidgetType {
...
+      this.refType === other.refType
+    );
   }
 
   toDOM(): HTMLElement {
     let tag = this.kind === 'inline' ? 'span' : 'div';
     let el = document.createElement(tag);
     el.setAttribute('data-card-id', this.cardId);
     el.setAttribute('data-card-kind', this.kind);
+    el.setAttribute('data-card-type', this.refType);
```
lukemelia: **The Intern:** Naming nit — `data-card-type` reads like it's the card's type/class (display name, icon, etc.), which is what `data-card-type-display-name` and `data-card-type-icon-html` mean elsewhere in the host (e.g. `prerendered-card-search.gts`, `operator-mode-overlays.gts`, `card-list.gts`). Here it actually stores the BFM ref kind (`'card'` vs `'file'`), which is a different concept. Would `data-bfm-ref-type` or `data-ref-type` be less ambiguous for future readers? Could save someone a confused `grep` later.

packages/base/codemirror-editor.gts
```diff
@@ -637,41 +656,65 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
...
+          (getCards(this, () => {
+            let urls = this.resolvedUrlsForRefType('file');
+            if (!urls.length) return undefined;
+            return {
+              filter: {
+                in: { url: urls },
+                on: { module: `${baseRealm.url}card-api`, name: 'FileDef' },
+              },
+            };
+          }) as unknown as {
```
lukemelia: **The Intern:** The `as unknown as { instances: FileDef[]; ... }` cast is a bit of a code smell — it tells TS to take our word for it that the resource returns `FileDef[]`. Two thoughts:

1. Is `getCards` typed generically (i.e. parameterizable like `getCards<FileDef>(...)`)? If so we could drop the cast.
2. If not, even a narrow helper like
   ```ts
   function asFileResource(r: ReturnType<NonNullable<typeof getCards>>) {
     return r as unknown as { instances: FileDef[]; isLoading: boolean };
   }
   ```
   would localize the unsafety to one named spot and document why.

Mostly a readability thing — happy to leave as-is if `getCards` can't be made generic.

packages/base/codemirror-editor.gts
```diff
@@ -949,44 +993,90 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
...
+                          @displayContainer={{false}}
+                        />
+                      </div>
+                    {{/if}}
+                  {{/let}}
+                </CardContextConsumer>
+              {{else}}
+                <CardContextConsumer as |context|>
+                  {{#let
+                    (this.getCardComponent target.instance)
```
lukemelia: **The Intern:** The `{{#if (eq target.refType 'file')}} … {{else}} … {{/if}}` branches here are nearly identical — the only differences are the test attribute (`-file-slot-` vs `-card-slot-`) and `cardId=target.instance.id` vs `card=target.instance` on the modifier. About 45 lines of near-duplicated template.

A similar duplication landed in `default-templates/markdown.gts` (lines 528–610) — same shape with the same two differences.

Looking at `ElementTracker.modify` in `host/app/resources/element-tracker.ts`, the modifier accepts either `card` or `cardId`, so both branches could pass `cardId=target.instance.id` and merge into one block. Or extract a small helper component that takes `instance`, `kind`, `format`, and the test-attribute name. Not blocking, but as a first-time reader it's a lot of eye-bouncing to spot the actual difference.

Is there a reason the card branch is using `card=` (passing the whole instance) rather than `cardId=`? If not, unifying on `cardId=` would let both refTypes share one template.

packages/base/rich-markdown.gts
```diff
@@ -80,6 +82,40 @@ export class RichMarkdownField extends FieldDef {
     },
   });
 
+  /** Resolved absolute URLs of `:file[URL]` and `::file[URL]` references. */
+  @field fileReferenceUrls = containsMany(StringField, {
+    computeVia: function (this: RichMarkdownField) {
+      if (!this.content) {
+        return [];
```
lukemelia, line 98: **The Intern:** This 8-line baseUrl resolution block is verbatim the same as the one in `cardReferenceUrls` just above (lines 61–66). If a future ref type lands (per the keyword-generic BFM infra), this'll get copy-pasted a third time. Would a tiny private getter like `get refBaseUrl(): string { … }` be welcome here, or is the duplication intentional to keep each `computeVia` self-contained? Either way, fine — just flagging the smell while I have the file open.

lukemelia (commented):
_Hi! The Intern here — first time poking at the BFM file-ref path, so apologies if any of this is wrong. I read through the diff end-to-end and have a few questions/observations. Mostly small stuff and a couple of duplication smells; flagging one reactivity concern I'm less sure about. Happy to be told 'no, that's fine.'_

— The Intern

packages/base/default-templates/markdown.gts
lukemelia: [Claude Code 🤖] Good catch — lifted into `runtime-common/bfm-card-references.ts` alongside `cardTypeName` and exported. Both `default-templates/markdown.gts` and the host `rendered-markdown.gts` now import the one implementation. Fixed in a0a0226bc7.

packages/base/default-templates/markdown.gts
lukemelia: [Claude Code 🤖] Agreed on consistency — the modifier now receives `@linkedFiles` as a third positional (`{{this.captureCardSlots this.renderedHtml @linkedCards @linkedFiles}}`), matching the host `rendered-markdown.gts` convention. Fixed in a0a0226bc7.

packages/host/app/lib/codemirror-context.ts
lukemelia: [Claude Code 🤖] Renamed to `data-bfm-ref-type` (set + read are both in this file, no other references). It now reads as the BFM ref kind rather than a card type/display-name. Fixed in a0a0226bc7.

packages/base/codemirror-editor.gts
lukemelia: [Claude Code 🤖] `getCards` can't be made generic for this: its type parameter is constrained `T extends CardDef`, and `FileDef extends BaseDef` (not `CardDef`), so `getCards<FileDef>` is rejected. The `on: { ..., name: 'FileDef' }` query genuinely returns FileDef instances at runtime, hence the reinterpret. Went with your option 2 — extracted a named `asFileResource(...)` helper so the `as unknown as` lives in one documented spot instead of inline at the call site. Fixed in a0a0226bc7.

packages/base/codemirror-editor.gts
lukemelia: [Claude Code 🤖] No good reason for the card branch to pass `card=` — collapsed both refTypes into one block. The only real difference was `card=instance` vs `cardId=instance.id`, and the overlay layer (`operator-mode/overlays.gts`) already handles a string `cardDefOrId` everywhere (`getCardId`, `isSelected`, `viewCard` all branch on `typeof === 'string'`), and the card is already loaded so resolving by id is a cache hit — so unifying on `cardId=` is behavior-preserving. I unified the slot/target on a single `instance` field and kept the distinct test hooks via conditional attrs. Applied to both `codemirror-editor.gts` and `default-templates/markdown.gts`. Fixed in a0a0226bc7.

packages/base/rich-markdown.gts
lukemelia: [Claude Code 🤖] Extracted a private `refBaseUrl` getter on `RichMarkdownField`; both `cardReferenceUrls` and `fileReferenceUrls` computeVia blocks now read it. Fixed in a0a0226bc7.

packages/base/default-templates/markdown.gts
```diff
@@ -210,21 +229,39 @@ export default class MarkDownTemplate extends GlimmerComponent<{
...
           }
         }
+        let filesByUrl = new Map<string, FileDef>();
+        if (linkedFiles?.length) {
+          for (let file of linkedFiles) {
+            if (file?.id) {
+              filesByUrl.set(trimJsonExtension(file.id), file);
+            }
+          }
+        }
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] False positive — the local `resolveUrl` helper in `packages/base/default-templates/markdown.gts` (lines 80–100) already wraps its result in `trimJsonExtension(...)` on every return path:

```ts
function resolveUrl(...) {
  try {
    return trimJsonExtension(
      virtualNetwork.resolveURL(raw, baseUrl || undefined).href,
    );
  } catch {
    return trimJsonExtension(new URL(raw, baseUrl || undefined).href);
  }
  // ...
  return trimJsonExtension(raw);
}
```

So `resolvedUrl` is already canonical (no `.json` suffix) by the time it reaches `filesByUrl.get(resolvedUrl)`, matching the `filesByUrl.set(trimJsonExtension(file.id), file)` map keys. The same is true at line 288. No mismatch.

packages/base/default-templates/markdown.gts
```diff
@@ -234,37 +271,43 @@ export default class MarkDownTemplate extends GlimmerComponent<{
...
+              sizeStyle ? `${sizeStyle}; overflow: hidden` : 'overflow: hidden',
+            );
+          }
+
+          let resolvedUrl = resolveUrl(rawUrl, baseUrl, virtualNetwork);
+
+          let instance =
+            refType === 'file'
+              ? filesByUrl.get(resolvedUrl)
+              : cardsByUrl.get(resolvedUrl);
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Same false positive as above — `resolveUrl` in this file (lines 80–100) returns a `trimJsonExtension`-normalized URL on every path, so the lookup key matches the trimmed map keys. No change needed.

packages/base/codemirror-editor.gts
```diff
@@ -637,41 +669,64 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
...
+    let instancesByUrl = new Map<string, CardDef | FileDef>();
+    let addInstances = (
+      instances: (CardDef | FileDef)[] | null | undefined,
+    ) => {
+      if (!instances?.length) return;
+      for (let instance of instances) {
+        if (instance?.id) {
+          instancesByUrl.set(trimJsonExtension(instance.id), instance);
         }
       }
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] False positive — the local `resolveUrl` helper in `packages/base/codemirror-editor.gts` (lines 91–111) already returns a `trimJsonExtension`-normalized URL on every return path (try, catch, and final fallback). So `resolvedUrl` matches the `trimJsonExtension(instance.id)` map keys at the `instancesByUrl.get(resolvedUrl)` call.

packages/base/codemirror-editor.gts
```diff
@@ -637,41 +669,64 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
...
 
-    let vn = this.args.cardReferenceVirtualNetwork;
     return targets.map((target) => {
       let resolvedUrl = resolveUrl(target.cardId, baseUrl, vn);
       return {
         ...target,
-        card: cardsByUrl.get(resolvedUrl) ?? null,
+        instance: instancesByUrl.get(resolvedUrl) ?? null,
       };
     });
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Same false positive as above — `resolveUrl` in this file trims `.json`, so the key form is canonical at the lookup site.

packages/host/app/components/operator-mode/preview-panel/rendered-markdown.gts
```diff
@@ -207,35 +270,58 @@ export default class RenderedMarkdown extends Component<Signature> {
...
+                state: 'resolved',
+                format,
+                file,
+                style: resolvedStyle,
+              });
+              continue;
+            }
+          } else {
+            let card = cardsByUrl.get(resolvedUrl);
+            if (card) {
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] False positive — the local `resolveUrl` helper in this file (lines 71–82) explicitly wraps `virtualNetwork.resolveRRI(...)` in `trimJsonExtension(...)` on both branches:

```ts
function resolveUrl(...) {
  try {
    return trimJsonExtension(
      virtualNetwork.resolveRRI(raw, baseUrl ? rri(baseUrl) : undefined),
    );
  } catch {
    return trimJsonExtension(raw);
  }
}
```

So `resolveUrl(...)` returns a canonical URL, matching the trimmed `filesByUrl` / `cardsByUrl` map keys. No mismatch at line 304.

packages/host/app/components/operator-mode/preview-panel/rendered-markdown.gts
```diff
@@ -38,21 +40,30 @@ import CardRenderer from '@cardstack/host/components/card-renderer';
...
 
 type CardSlotFormat = 'atom' | 'embedded' | 'fitted' | 'isolated';
 type SlotState = 'resolved' | 'loading' | 'unresolved';
+type RefType = 'card' | 'file';
 
 interface RenderSlot {
   element: HTMLElement;
+  // 'card' refs (`:card[URL]`) resolve to CardDef instances; 'file' refs
+  // (`:file[URL]`) resolve to FileDef instances rendered without card overlays.
+  refType: RefType;
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Good catch — the doc comment was misleading. Wiring `cardComponentModifier` to file slots is intentional: the operator-mode overlay layer explicitly supports file targets (see `operator-mode-overlays.gts:397` and `getTypeForCardTarget`, which returns `'card' | 'file'`). Updated the comment to describe the actual behavior — file slots are tracked the same way as card slots, and the overlay layer distinguishes them. Fixed in 85765a10d9.

FadhlanR (approved):

PR #5257, opened 2026-06-17
Attach skill markdown files from the AI assistant skill menu

lukemelia:
Stacked on #5253.

Adds a way to attach a skill expressed as a markdown file (`MarkdownDef` with `boxel.kind: skill`) from the AI assistant skill menu. This is introduced as a **parallel** flow to the existing skill-card attach — deliberately *not* merged with it — so the skill-card path can later be removed as a clean deletion.

## Reuse the FileChooser, scoped to skills
Rather than a new chooser, this reuses the existing `chooseFile` flow, narrowed to skill markdown:
- Thread an optional `fileFieldFilter` (equality constraints on indexed file fields) through `chooseFile` → `ChooseFileModal` → `IndexedFileTree` → the index-backed `file-tree-from-index` query. A `{ kind: 'skill' }` filter shows only skill `.md` files, not all markdown. (The query is `{ 'item.on': MarkdownDef, eq: { 'item.kind': 'skill' } }`.)

## Skill menu + room wiring
- The skill menu gains a parallel "Choose a skill file to add" action (its own button + `@onChooseSkillMarkdown` callback), leaving the card button untouched.
- The chosen file's source URL flows to the same activation path the card flow uses (`UpdateRoomSkillsCommand`, which already accepts markdown skill ids and uploads their commands — see #5253).

## Display
- `RoomResource.loadSkills` loads skill markdown files through the file-meta read type (rather than downloading a card document from Matrix) and references them as file-meta, so the menu pill renders and the room's usable commands resolve.
- `SkillToggle` loads a markdown skill id as a file-meta resource and titles the pill from the frontmatter name / title (Skill cards still use `cardTitle`).

## Tests
- `file-tree-from-index`: a `{ kind: 'skill' }` `fileFieldFilter` limits the tree to skill markdown; unscoped lists every markdown file.
- file-chooser acceptance suite re-run green (shared chooser infra is additive/optional).

## Follow-ups (not in scope)
- An acceptance test driving the full menu attach → file-tree → activate path.
- The skill-toggle "Open Skill Card" menu item isn't markdown-aware yet.
- Removing the skill-card attach path once markdown skills are the standard (the motivation for keeping these flows separate).


lukemelia:
## Self-review

Walked the diff plus the supporting pieces in `lib/skill-commands.ts`, `resources/file-tree-from-index.ts`, `services/store.ts:addReference`, and the search-entry wire grammar in `runtime-common/search-entry.ts`. A few things worth calling out before this merges.

### Overall shape — good
The deliberate parallel path is the right call. `onChooseSkillMarkdown` doesn't touch the card flow, both wire into the same `attachSkillTask` → `UpdateRoomSkillsCommand`, and `isMarkdownSkillId` is a single source of truth for "load as file-meta vs. card." That gives us a clean deletion target later. The `fileFieldFilter` plumbing through `chooseFile → ChooseFileModal → IndexedFileTree → fileTreeFromIndex` is a small, well-bounded shape change.

### Issues / suggestions

**1. The `pickTask` cancel fix is broader than the PR's stated scope.** `choose-file-modal.gts:147` — previously, cancel/Escape/close called `pickTask.perform(undefined)`, the `if (deferred && selectedRealm && path)` guard failed, and the deferred was silently dropped by `resetState()`. Any `chooseFile` caller hit this. New behavior settles with `undefined`. Good fix, but:
   - The cancel regression test (`canceling the skill-file chooser re-enables the add-markdown button`) only exercises the markdown-skill flow. The underlying bug also affected the existing file-attach flows. Worth adding a generic cancel test against the non-scoped chooser, or at minimum calling this fix out as load-bearing in the PR description so it doesn't get reverted later as "extra noise."

**2. `displayTitle` (skill-toggle.gts:75–88) uses a wide structural cast instead of branching on `isMarkdownSkill`.** It works, but it relies on Skill cards never having `frontmatter` / `name` siblings that would shadow `cardTitle`. Today they don't, but the type system isn't enforcing it. Mild preference for:
   ```ts
   if (this.isMarkdownSkill) {
     let md = this.card as MarkdownDef | undefined;
     return md?.frontmatter?.name ?? md?.title ?? md?.name;
   }
   return (this.card as Skill | undefined)?.cardTitle;
   ```

**3. `doAttachSkillMarkdown` has no error handling around `chooseFile`.** `chooseFile` can reject — `loadCardDef(opts.fileType)` inside the modal's `chooseFile` is `await`ed and can throw. If it does, `restartableTask` swallows the rejection silently (no surfaced toast, no console), and the button just settles back to enabled state with no indication anything went wrong. `doAttachSkillCard` has the same shape, so this is pre-existing, but it's still worth a `try/catch` with at minimum a `console.error`.

**4. `fileFieldFilter: Record<string, unknown>` is fully permissive.** Acceptable for an internal API, but the wire grammar will reject malformed filters at runtime rather than at the call site. The only producer today is `{ kind: 'skill' }` — if we expect more callers, consider a tighter type or a small adapter that validates against a known set of indexable file fields. Lower priority; flagging in case it grows.

**5. `openSkillFile` (skill-toggle.gts:115–128) interact-mode path isn't exercised.** The new test "skill pill menu opens the skill markdown file" runs from a default state — I'd want one assertion that exercises the `submode === 'code'` branch too (`updateCodePath`). Also, the `Math.min(om.numberOfStacks(), 1)` pattern looks like it was copied from elsewhere — confirm that's actually the intent for skill *files* (vs. cards).

**6. `isMarkdownSkillId` has no unit test.** It's a small predicate with three behaviors worth pinning down: `.md` ✅, `.markdown` ✅, anything with a query string / hash ✅ (via `new URL(id).pathname`), and the falsy-URL fallback to raw string matching. None are covered. A handful of `assert.true / assert.false` cases would lock the contract for the next reader who's tempted to "simplify" the URL parsing.

**7. Comment placement in `file-tree-from-index.ts:122`.** The `// Field paths in the search-entry wire grammar are item.-prefixed` comment is great. Consider also noting *why* this lives at this layer (so callers can pass unprefixed field names like `kind` instead of `item.kind`) — that's the actual user-facing contract.

### Risk
Low. The new path is gated behind a new button + a new `onChooseSkillMarkdown` callback, the card path is untouched, and the gating in `loadSkills` is per-id. The one ambient change is the `pickTask` cancel fix — worth confirming the existing file-chooser acceptance tests cover Cancel/Escape so we don't ship a regression on the other choose-file flows.

### Verification I'd want before merging
- Run the file-chooser acceptance suite (PR notes claim green; want to see the run).
- Smoke the markdown-skill attach in interact mode and code mode (the `openSkillFile` branch).
- Verify that an already-enabled markdown skill is excluded from the chooser the same way the card flow excludes enabled skills (currently `skill picker excludes already-enabled skills` only covers card skills).

FadhlanR (approved):
Looks good other than these two points:

1.  Should we add validation in the card chooser's upload flow? When the chooser is filtered to skill files only, users could upload a non-skill file via the upload button/drag-drop. The uploaded file won't appear in the filtered file list, which could be confusing, no feedback, no error. We should either validate the uploaded file against the filter, or show a clear message explaining why it's not shown.

2. It would be good if there is a space between these buttons.

 
Image

jurgenwerk (approved):

PR #5261, opened 2026-06-17
host: Remove error details from 404 page

backspace:
The new 404 page had an error container to show details:

image

This removes it, since the error doesn’t really say much more than 404:

s 2026-06-17 at 13 38 46@2x

### DIFF
--- a/packages/host/app/components/host-mode/card.gts
+++ b/packages/host/app/components/host-mode/card.gts
@@ -10,7 +10,6 @@ import { bool, cn } from '@cardstack/boxel-ui/helpers';
 
 import CardRenderer from '@cardstack/host/components/card-renderer';
 import CardError from '@cardstack/host/components/operator-mode/card-error';
-import CardErrorDetail from '@cardstack/host/components/operator-mode/card-error-detail';
 import { getCard } from '@cardstack/host/resources/card-resource';
 
 interface Signature {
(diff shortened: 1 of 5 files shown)


FadhlanR (approved):

packages/host/app/components/host-mode/card.gts
FadhlanR: We can remove this

FadhlanR (approved):

packages/host/app/components/host-mode/card.gts
backspace: thanks, removed now

packages/host/app/components/host-mode/card.gts
```diff
@@ -125,14 +124,10 @@ export default class HostModeCard extends Component<Signature> {
     >
       {{#if this.cardError}}
         {{#if this.isNotFound}}
-          {{! A 404 (e.g. a routing rule pointing at a deleted card) gets a
-              friendly centered placeholder, while the collapsible technical
-              detail stays available below for the realm owner debugging it. }}
```
habdelra, line 130: what about an error where a dependency like a module import returns a 404, i think that 404 is propagated. in that case it's not that the card is not found its that it's in a legitimate error state because its dep is not found. i think we do want to surface that kind of error

backspace: you were right that the error was being swallowed in that case, thanks, updated now

s 2026-06-18 at 09 19 29@2x

s 2026-06-18 at 09 20 56@2x

PR #5266, opened 2026-06-17
Time out wedged test fetches so a stuck h2 stream retries instead of hanging

habdelra:
## What

A host acceptance test (`code-submode | card playground > multiple realms: edit format is enabled for writable instances of read-only card definitions`) intermittently times out at 60s. The timeout diagnostics already in the suite pinned the cause precisely:

```
[test-timeout] ... ran for 126135ms
in-flight fetches at rejection time (1):
  QUERY https://localhost:4201/base/_info (outstanding 119777ms)
```

One fetch to the base realm hung for ~120s and never settled. The same run shows other base-realm fetches on the session (`css-value`, `color`) failing and recovering — so the HTTP/2 session itself was alive, but that single `_info` stream was individually wedged.

## Why the existing safeguards don't catch it

- **`withRetries`** (the test-env fetch retry) only reacts to *rejections*. A fetch that never settles is never retried — and because the `fetcher` test-waiter stays open around it, `settled()` / `waitFor` never resolve and the test waits out its 60s budget.
- **The realm server's HTTP/2 keepalive** tears down a session only when it goes *fully silent* (misses consecutive pongs). A session that keeps answering pings while one of its streams is stuck is healthy from the keepalive's point of view, so it's never torn down. This is the "transport alive, stream wedged" mode the keepalive's own diagnostics distinguish — and it's exactly what happened here.

## The fix

Give each retryable fetch attempt in the test environment a deadline. When an attempt doesn't settle in time, it's aborted; that rejects the underlying fetch (freeing the `fetcher` waiter) and lets the existing retry loop reissue the request on a fresh stream. The deadline (15s) sits well above a healthy base-realm fetch, so a merely slow response is never aborted, and it leaves room for several retries inside the 60s test timeout.

A timed-out attempt is logged distinctly from a rejected one (`Encountered fetch timeout (>…ms) for <url>`), so if a future failure slips through, CI output shows the deadline firing and on which URL rather than blending into the transient "fetch failed" noise.

Scope is limited to the test-env retry path. Production fetch behavior — including the env-mode worker-boot `.localhost` retry, which intentionally has no deadline — is unchanged, and any caller-supplied `AbortSignal` is preserved (merged with the timeout signal).

## Test

`Unit | VirtualNetwork fetch retries` drives a stub whose first attempt mirrors a wedged stream (never resolves; only settles when aborted) and asserts the request is aborted and retried to success. A test-only `__fetchAttemptTimeoutMs` knob shrinks the deadline so the path runs in milliseconds.

### DIFF
--- a/packages/host/tests/unit/fetcher-test.ts
+++ b/packages/host/tests/unit/fetcher-test.ts
@@ -140,3 +140,61 @@ module('Unit | fetcher', function () {
     assert.strictEqual(await response.text(), 'OK', 'body text is correct');
   });
 });
+
+module('Unit | VirtualNetwork fetch retries', function (hooks) {
+  let priorEnvironment: unknown;
+  let priorTimeout: unknown;
+
+  hooks.beforeEach(function () {
+    // The per-attempt timeout is armed only in the test environment; pin it
+    // and shrink the deadline so the retry path runs in milliseconds.
+    priorEnvironment = (globalThis as any).__environment;
+    priorTimeout = (globalThis as any).__fetchAttemptTimeoutMs;
+    (globalThis as any).__environment = 'test';
+    (globalThis as any).__fetchAttemptTimeoutMs = 50;
+  });
+
+  hooks.afterEach(function () {
+    (globalThis as any).__environment = priorEnvironment;
+    if (priorTimeout === undefined) {
+      delete (globalThis as any).__fetchAttemptTimeoutMs;
+    } else {
+      (globalThis as any).__fetchAttemptTimeoutMs = priorTimeout;
+    }
+  });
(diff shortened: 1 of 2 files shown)


habdelra:
[Claude Code 🤖] Closing in favor of #5267.

The shard-9 realm-server log shows the existing PING keepalive (#5219) **did** fire on the one frame-silent session for this failure, but its force-destroy left a zombie — no `close` event, the connection lingered 185s, and the client `_info` fetch stayed hung. So the right next step is server-side: harden the keepalive teardown (also destroy the underlying socket) and record whether the teardown reaches the peer at all. That's #5267.

PR #5272, opened 2026-06-17
feat: surface frontmatter parse errors via indexing diagnostics

lukemelia:
## What & why

[CS-11548](https://linear.app/cardstack/issue/CS-11548)

When a markdown file's YAML frontmatter fails to parse, `MarkdownDef.extractAttributes` indexes the body anyway and drops everything the frontmatter declared — a skill's `commands`, its `boxel.kind`, and so on. With only a `console.warn`, an author has no signal that those declarations were lost.

This routes frontmatter parse failures through the indexing-diagnostics surface the same way `brokenLinks` findings are: the row indexes cleanly (no `has_error`) but carries the failure so it's visible to the `/_indexing-errors` endpoint and the `boxel realm indexing-errors` CLI.

## How

- **Capture** (`base/markdown-file-def.gts`): on YAML parse failure, record `{ message, line?, column? }` and route it out-of-band via `Symbol.for('boxel:file-frontmatter-parse-error')`, the same mechanism the polymorphic `frontmatter` field meta uses.
- **Lift** (`host/.../file-def-attributes-extractor.ts`): pull the symbol off the search doc into `FileDefExtractResult.frontmatterParseError`.
- **Persist** (`runtime-common/index-runner/file-indexer.ts`): merge it onto the row's `diagnostics.frontmatterParseError` on the success path.
- **Surface** (`runtime-common/realm.ts`): the `/_indexing-errors` endpoint emits one resource per finding class. A healthy row (`has_error = FALSE`) carrying a `diagnostics.frontmatterParseError` yields a `frontmatter-error` finding; one carrying `diagnostics.brokenLinks` yields a `broken-link` finding; a row with both yields both, so neither signal masks the other. A row that yields more than one finding appends the finding class to its resource id (`<type>::<url>::broken-link`) to stay JSON:API-unique. The `boxel realm indexing-errors` CLI renders each (`frontmatter parse error (line L:C): <message>`). `FrontmatterParseError` type added to `runtime-common`.

## Tests

- Host integration: invalid YAML routes the parse-error marker and still indexes the body.
- Realm-server endpoint: a `frontmatter-error` row surfaces with its payload, `has_error = FALSE`, no `errorDoc`; a healthy row with both broken links and a frontmatter error surfaces as two findings.
- boxel-cli integration: endpoint → CLI round-trip, `shortFrontmatterError` formatting, and the combined-findings case.


lukemelia:
The Intern

The patch adds the intended frontmatter diagnostic path, but it can hide pre-existing broken-link findings on rows that also have frontmatter parse errors. That regression should be addressed before considering the change correct.

Review comment:

- [P2] Preserve broken-link findings when frontmatter also fails — /Users/lmelia/p/cardstack/boxel/packages/runtime-common/realm.ts:6277-6280
  When a healthy row has both `diagnostics.brokenLinks` and `diagnostics.frontmatterParseError` (for example, a markdown skill with invalid frontmatter and a broken card reference in its body), this classifies the row only as `frontmatter-error`. The non-JSON CLI then prints only the frontmatter summary, and JSON consumers filtering for `type == "broken-link"` no longer see an existing broken-link finding for that file. Consider emitting both findings or combining the summary so the broken-link signal is not hidden.

lukemelia:
[Claude Code 🤖] Addressed in cc8c56a777.

The `_indexing-errors` endpoint emits one resource per finding class. A healthy row (`has_error = FALSE`) carrying both `diagnostics.brokenLinks` and `diagnostics.frontmatterParseError` produces two resources — a `broken-link` finding and a `frontmatter-error` finding — so a consumer filtering by `type` sees both and the CLI prints a line per finding.

A row that yields more than one finding appends the finding class to its resource id (`<type>::<url>::broken-link`) to stay JSON:API-unique; a single-finding row keeps the `<type>::<url>` id. Error rows (`has_error = TRUE`) carry `brokenLinks` as an attribute on the `indexing-error` resource, since the render failure is the headline.

Endpoint and boxel-cli integration tests cover the combined case.

lukemelia:
@habdelra I don't know too much about the indexing-errors endpoint, so give this PR a close look

habdelra (approved):
this is a perfect use the the diagnostics column and indexing errors endpoint. we handled broken links the very same way 👍

PR #5274, opened 2026-06-18
software-factory: warn when a brief reads like an adjust task but sourceCardUrl is unset

jurgenwerk:

## Why
The factory's **adjust-existing-card** flow is triggered by the brief's structured `sourceCardUrl` attribute (`realm/wiki.gts`) — *not* by its prose. When `sourceCardUrl` is set, `factory-seed.ts` emits the seed instruction `boxel realm ingest-card "<sourceCardUrl>" .` (which auto-resolves the source realm).

If a brief *describes* adjusting an existing card only in prose but never sets `sourceCardUrl`, the factory silently takes the **greenfield** path: the agent never ingests the source and rebuilds from scratch (or, as seen in the wine-cellar test runs, flails trying to guess the source realm). The failure is silent — nothing tells the operator the brief shape was wrong.

## What
At seed time, if `sourceCardUrl` is unset but the brief's title/content reads like an adjust task, emit a non-fatal `log.warn` pointing the operator at the fix:

```
Brief "<title>" reads like an "adjust existing card" task but its `sourceCardUrl`
attribute is unset — the factory will treat this as a greenfield build and the
agent will NOT ingest a source card. Set the brief's `sourceCardUrl` to the
absolute source card URL to trigger the adjust flow.
```

- The decision is an exported pure predicate, `shouldWarnMissingSourceCardUrl(brief)` = `!brief.sourceCardUrl && briefProseLooksLikeAdjust(brief)`, so it's unit-testable without capturing logs.
- The prose heuristic is a narrow keyword set ("adjust the existing", "existing card", "rather than rebuild", "source card", "ingest-card", …) — phrases that don't naturally appear in a greenfield "build a new X" brief, to avoid nagging.

This is a guardrail only; it changes no control flow.

## Testing
- New `tests/factory-seed.test.ts` (registered in `tests/index.ts`): warns on adjust-prose-without-url; silent on greenfield; silent when `sourceCardUrl` is set. All pass.
- `pnpm lint` clean (eslint / prettier / ember-tsc); `pnpm test:node` 455 pass (the one failure is the pre-existing `port-allocator` dual-stack env test).

Ticket: CS-11650. (Supersedes CS-11649, closed as a false premise — the ingest/adjust machinery already exists and is correct.)

### DIFF
--- a/packages/software-factory/src/factory-seed.ts
+++ b/packages/software-factory/src/factory-seed.ts
@@ -24,6 +24,39 @@ export function inferDarkfactoryModuleUrl(targetRealm: string): string {
 
 let log = logger('factory-seed');
 
+// Phrases that read like "modify a card that already exists" rather than
+// "build a new one". Kept deliberately narrow — these don't naturally appear
+// in a greenfield "build a new X" brief, so a match on one of them while
+// `sourceCardUrl` is unset is a strong signal the brief author meant to set it.
+const ADJUST_PROSE_SIGNALS = [
+  'adjust the existing',
+  'adjust an existing',
+  'adjusting the existing',
+  'existing card',
+  'rather than rebuild',
+  'rather than build',
+  'instead of rebuilding',
+  'ingest-card',
+  'source card',
+];
+
+function briefProseLooksLikeAdjust(brief: FactoryBrief): boolean {
+  let text = `${brief.title}\n${brief.content}`.toLowerCase();
+  return ADJUST_PROSE_SIGNALS.some((signal) => text.includes(signal));
+}
+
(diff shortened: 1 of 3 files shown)


jurgenwerk:

Closing — not needed. The factory selects the adjust flow purely on the brief's `sourceCardUrl` field (`adjust = Boolean(brief.sourceCardUrl)`); prose never causes a greenfield fallback when the field is set. This PR only warned about the inverse operator error (adjust-sounding prose with `sourceCardUrl` forgotten), which isn't worth the surface area. Ticket CS-11650 canceled.

PR #5284, opened 2026-06-18
CS-11655: add app.boxel.realm-servers event type + read/write helpers

lukemelia:
## Summary

Foundational data-model change for the [Rework Realms List Source-of-Truth](https://linear.app/cardstack/project/rework-realms-list-source-of-truth-1ba3e20e5980) project. Introduces the `app.boxel.realm-servers` account-data event type so a user's Matrix account data can carry the set of *trusted realm servers* alongside the existing flat realm list. Boot assembly (CS-11658) and lazy migration (CS-11659) land in follow-ups; this PR is plumbing-only.

- **runtime-common** — new `APP_BOXEL_REALM_SERVERS_EVENT_TYPE` constant and `AppBoxelRealmServersContent` payload type.
- **host matrix-service** — `get/set/append/remove RealmServers...AccountData` helpers mirroring the existing realms helpers. Legacy `app.boxel.realms` behavior is unchanged.
- **realm-server synapse** — parallel `appendRealmServerToUserAccountData` with the same idempotent retry-on-stomp semantics. The existing fetch/put plumbing is factored to take an `eventType` so both helpers share it.
- **realm-server upsert-permission handler** — after the existing realm append, also writes the realm-server origin to `app.boxel.realm-servers` so both keys stay in lockstep going forward.
- **mock matrix client** — handles get/set + AccountData event routing for the new key, with a new `activeRealmServers` config option.

## Acceptance criteria (from [CS-11655](https://linear.app/cardstack/issue/CS-11655))

- [x] New constant and payload type exported from `runtime-common`.
- [x] Helpers to read and write `app.boxel.realm-servers` account data exist and are typed.
- [x] Legacy `app.boxel.realms` behavior is unchanged.

## Test plan

- [x] `pnpm lint:types` clean in `runtime-common`, `host`, `realm-server`.
- [x] `pnpm lint:js` clean in `runtime-common`, `host`, `realm-server`.
- [x] `pnpm prettier --check` on all touched files.
- [ ] CI: host integration test `Integration | matrix-service | realm-servers account data` — round-trips the payload, append idempotency, remove leaves other entries.
- [ ] CI: realm-server endpoint tests — `appendRealmServerToUserAccountData` preserves prior entries; grafana upsert populates `app.boxel.realm-servers` with the realm-server origin.


packages/realm-server/synapse.ts
```diff
@@ -219,50 +222,110 @@ export async function appendRealmToUserAccountData({
...
-    let existing = await fetchRealmsAccountData(
+    let existing = await fetchAccountData(
       matrixURL,
       userId,
       userAccessToken,
+      eventType,
     );
-    let realms = Array.isArray(existing.realms) ? existing.realms : [];
-    if (realms.includes(realmURL)) {
+    let entries = Array.isArray(existing[arrayKey])
```
lukemelia, line 290: [Codex] This changes the legacy `app.boxel.realms` append behavior too: `appendRealmToUserAccountData` now routes through this helper, so any existing non-string entries in the `realms` array are filtered out and the PUT below writes back only `entries`. The previous implementation preserved the existing array contents exactly with `realms: [...realms, realmURL]`. Since this PR calls out legacy `app.boxel.realms` behavior as unchanged, please avoid sanitizing the stored array on write; if we want a string-only view for `includes`, keep that separate from the array we persist back.

lukemelia (commented):
[Codex] I reviewed PR 5284 and found one issue.

packages/realm-server/synapse.ts
lukemelia: low-risk, leaving as is

FadhlanR (approved):

packages/host/app/services/matrix-service.ts
```diff
@@ -711,6 +712,44 @@ export default class MatrixService extends Service {
     await this.realmServer.setAvailableRealmIdentifiers(newRealms.map(ri));
   }
 
+  public async getRealmServersFromAccountData(): Promise<string[]> {
+    let { realmServers = [] } =
+      ((await this.client.getAccountDataFromServer(
+        APP_BOXEL_REALM_SERVERS_EVENT_TYPE,
+      )) as { realmServers: string[] }) ?? {};
+    return realmServers;
+  }
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Leaving as-is. `app.boxel.realm-servers` is a brand-new event type with no legacy/migrated data, and the only writers (these host helpers plus the server-side append) always persist a `string[]`. The absent case is already handled by `?? {}` → `[]`. These helpers also intentionally mirror the existing legacy `realms` helpers, which use the same un-normalized cast. Same low-risk call as the parallel Codex thread above.

PR #5287, opened 2026-06-19
Add realm-server /_delegate-session endpoint for user-scoped read-only JWTs

lukemelia:
## What & why

New shared-secret-authenticated realm-server endpoint, **`POST /_delegate-session`**, that mints a **30-minute, single-realm, read-only** JWT for a named user:

```
ai-bot → POST /_delegate-session   (HMAC over body + timestamp)
         { onBehalfOf: '@user:boxel.ai', realm: 'https://realm/u/jane/' }
       ← { token, realm, permissions: ["read"] }
```

ai-bot will use this (CS-11553 / CS-11554) to read a realm on a user's behalf without a blanket `@aibot` read grant — avoiding the confused-deputy exfiltration risk where any room member could point a skills state event at someone else's realm. Implements CS-11552; follows the v1 security design settled in CS-11551 (Spec §4).

## How it works

1. **Auth** — HMAC-SHA256 over `${timestamp}.${rawBody}` with the shared secret, sent as `x-boxel-delegation-signature` + `x-boxel-delegation-timestamp` headers. Constant-time compare, ±60s timestamp window. The secret never crosses the wire, so the window genuinely bounds replays (TLS + rotation remain the defenses against leak — rotation tooling is CS-11567).
2. **Scope check** — looks up `onBehalfOf`'s permissions on the realm; requires `read` (else 403). The token grants exactly the read access the user already has.
3. **Mint** — `createJWT({ user: onBehalfOf, realm, permissions: ['read'], delegated: true }, '30m', realmSecretSeed)`.
4. **Audit** — every request is logged with a correlation id and outcome.

### Realm-side acceptance
A `["read"]`-only token would otherwise be rejected by the realm's exact-permissions-match check (`realm.ts`) on a private realm where the user has write/owner perms. So `assertRequestPermissions` now accepts a `delegated` token **only for read operations** and **only when the bound user actually has read** — a write attempt gets 403, and the exact-match invariant is untouched for all normal sessions.

## Decisions worth a reviewer's eye
- **`AI_BOT_DELEGATION_SECRET` is optional**, not required in `main.ts`. Unset → the endpoint returns **503** and mints nothing, so deployments aren't forced to provision a secret for a feature nothing consumes yet. Provisioning/rotation lands with CS-11553 / CS-11567.
- **Spec's `iss`/`aud`/`sub`/`jti` claims** are subsumed by the existing realm-token shape the verifier actually reads (`user` = subject, `realm` = audience, `realmServerURL` = issuer); the audit log is the forensic record `jti` was for.

## Testing
- `tests/server-endpoints/delegate-session-test.ts` — 12 tests: claim assertions, **end-to-end realm read accepted** for a realm-owner user, **write rejected** (read-only), auth failures (missing/invalid sig, wrong secret, stale/future timestamp), 403 no-read, 400 bad/missing body.
- HMAC sign/verify logic unit-verified locally (8/8 cases incl. the ±60s edge).
- ⚠️ The full QUnit suite (needs the dev services stack: test-pg + worker-manager + prerender) has **not** been run in this environment yet — type-check and lint are clean.

### DIFF
--- a/packages/realm-server/handlers/handle-delegate-session.ts
+++ b/packages/realm-server/handlers/handle-delegate-session.ts
@@ -1,7 +1,6 @@
 import type Koa from 'koa';
 import { randomUUID } from 'crypto';
 import {
-  ensureTrailingSlash,
   fetchRealmPermissions,
   logger,
   SupportedMimeType,
@@ -9,6 +8,7 @@ import {
 import RealmPermissionChecker from '@cardstack/runtime-common/realm-permission-checker';
 import type { CreateRoutesArgs } from '../routes.ts';
 import { createJWT } from '../jwt.ts';
+import { normalizeRealmURL } from '../utils/realm-url.ts';
 import {
   fetchRequestFromContext,
   sendResponseForBadRequest,
@@ -28,14 +28,6 @@ import {
 const DELEGATED_TOKEN_TTL = '30m';
 
 const log = logger('realm:delegate-session');
-
-function headerValue(
-  headers: Koa.Context['req']['headers'],
-  name: string,
-): string | undefined {
-  let value = headers[name];
-  return Array.isArray(value) ? value[0] : value;
-}
 
 // Mints a realm session JWT scoped to a named user's read access on a single
 // realm (CS-11552). Shared-secret authenticated (HMAC over the request body +
@@ -67,8 +59,8 @@ export default function handleDelegateSession({
 
     let auth = verifyDelegationRequest({
       secret: aiBotDelegationSecret,
-      timestamp: headerValue(ctxt.req.headers, DELEGATION_TIMESTAMP_HEADER),
-      signature: headerValue(ctxt.req.headers, DELEGATION_SIGNATURE_HEADER),
+      timestamp: ctxt.get(DELEGATION_TIMESTAMP_HEADER),
+      signature: ctxt.get(DELEGATION_SIGNATURE_HEADER),
       rawBody,
       now: Date.now(),
     });
@@ -103,23 +95,18 @@ export default function handleDelegateSession({
       return;
     }
 
-    let realmURL: URL;
-    try {
-      realmURL = new URL(realm);
-    } catch {
+    // Normalise to the canonical realm-root form so the permission lookup hits
+    // the same realm_user_permissions row a write through other endpoints
+    // produced (they key by the trailing-slash href; see normalizeRealmURL).
+    let normalizedRealmURL = normalizeRealmURL(realm);
+    if (!normalizedRealmURL) {
       await sendResponseForBadRequest(
         ctxt,
         `"realm" is not a valid URL: ${realm}`,
       );
       return;
     }
-    // Normalise to the canonical realm-root form so the permission lookup hits
-    // the same realm_user_permissions row a write through other endpoints
-    // produced (they key by the trailing-slash href; see
-    // handle-upsert-realm-user-permission).
-    realmURL.search = '';
-    realmURL.hash = '';
-    let normalizedRealmHref = ensureTrailingSlash(realmURL.href);
+    let normalizedRealmHref = normalizedRealmURL.href;
 
     // Forensic audit record (security design CS-11551): every delegation
     // request is logged with a correlation id, requester, and outcome.
@@ -132,7 +119,7 @@ export default function handleDelegateSession({
     // via a `users` grant, diverging from what the realm would accept.
     let realmPermissions = await fetchRealmPermissions(
       dbAdapter,
-      new URL(normalizedRealmHref),
+      normalizedRealmURL,
     );
     let permissionChecker = new RealmPermissionChecker(
       realmPermissions,
--- a/packages/realm-server/handlers/handle-delegate-session.ts
+++ b/packages/realm-server/handlers/handle-delegate-session.ts
@@ -2,10 +2,11 @@ import type Koa from 'koa';
 import { randomUUID } from 'crypto';
 import {
   ensureTrailingSlash,
-  fetchUserPermissions,
+  fetchRealmPermissions,
   logger,
   SupportedMimeType,
 } from '@cardstack/runtime-common';
+import RealmPermissionChecker from '@cardstack/runtime-common/realm-permission-checker';
 import type { CreateRoutesArgs } from '../routes.ts';
 import { createJWT } from '../jwt.ts';
 import {
@@ -44,6 +45,7 @@ function headerValue(
 // couldn't, and can never write.
 export default function handleDelegateSession({
   dbAdapter,
+  matrixClient,
   realmSecretSeed,
   serverURL,
   aiBotDelegationSecret,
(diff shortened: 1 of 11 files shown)


packages/runtime-common/realm.ts
```diff
@@ -3697,6 +3703,33 @@ export class Realm {
...
+
+      // Delegated read-only session (minted by the realm-server's
+      // /_delegate-session endpoint for ai-bot — CS-11552). It is bound to a
+      // single user and deliberately scoped to ['read'] even when that user
+      // has broader permissions, so neither the exact-permissions-match
+      // invariant used for normal sessions below nor the assume-user
+      // indirection applies. Enforce instead the two guarantees the delegation
+      // design promises: the session is read-only, and it grants no more than
+      // the bound user can already read.
+      if (token.delegated) {
```
lukemelia, line 3715: [Codex] This delegated-token branch does not check that the JWT's `realm` claim matches the realm currently handling the request (`requestContext.realm.url` / `this.url`). Because delegated tokens are signed with the realm-server seed and this branch skips the normal exact-permissions match, a token minted for realm A can be replayed against realm B on the same server whenever the bound user also has read on B. That breaks the advertised single-realm scope. Please reject delegated tokens whose normalized `token.realm` differs from the current realm before allowing the read.

packages/realm-server/handlers/handle-delegate-session.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    // Forensic audit record (security design CS-11551): every delegation
+    // request is logged with a correlation id, requester, and outcome.
+    let auditId = randomUUID();
+
+    let permissionsForAllRealms = await fetchUserPermissions(dbAdapter, {
+      userId: onBehalfOf,
+      onlyOwnRealms: false,
+    });
+    let userPermissions = permissionsForAllRealms[normalizedRealmHref] ?? [];
```
lukemelia: [Codex] This mint-time check uses `fetchUserPermissions`, which includes exact-user rows and public `*` grants, but the live realm authorizer also grants `users` permissions to Matrix users whose profile exists (`RealmPermissionChecker.for`). For a realm with `users: ['read']`, the user really can read the realm, but this endpoint will return 403 because there is no exact row for `onBehalfOf`. Since the endpoint is supposed to mint when the bound user actually has read, this should mirror `RealmPermissionChecker` or explicitly include validated `users` grants, with a test for that case.

lukemelia (commented):
[Codex] Code review findings from a focused pass on delegated-token scope and permission parity.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Good catch — fixed in 866653ffb1. The delegated branch in `assertRequestPermissions` now rejects a token whose normalized `realm` claim differs from the realm handling the request (`ensureTrailingSlash(token.realm) !== ensureTrailingSlash(this.url)`) before allowing the read, restoring single-realm scope. Added a test that mints a validly-signed delegated token for a different realm and asserts it's rejected (401) against this realm even though the bound user has read here.

packages/realm-server/handlers/handle-delegate-session.ts
lukemelia: [Claude Code 🤖] Agreed — fixed in 866653ffb1. The mint-time check now uses `RealmPermissionChecker` (built from `fetchRealmPermissions` + the realm-server's matrixClient) and `.can(onBehalfOf, 'read')`, so it resolves `*`, exact rows, and `users` grants identically to the realm authorizer. A realm with `users: ['read']` no longer 403s a profile-bearing user. Added two tests: minting succeeds for a `users`-grant user with a Matrix profile, and is denied when no profile exists.

packages/realm-server/handlers/handle-delegate-session.ts
```diff
@@ -0,0 +1,187 @@
...
+      );
+      return;
+    }
+    // Normalise to the canonical realm-root form so the permission lookup hits
+    // the same realm_user_permissions row a write through other endpoints
+    // produced (they key by the trailing-slash href; see
+    // handle-upsert-realm-user-permission).
+    realmURL.search = '';
+    realmURL.hash = '';
+    let normalizedRealmHref = ensureTrailingSlash(realmURL.href);
```
jurgenwerk: [Claude Code 🤖] This realm-URL normalization (`new URL` → clear `search`/`hash` → `ensureTrailingSlash(href)`) is byte-identical to `handle-upsert-realm-user-permission.ts` (lines ~104-116), and `handle-delete-realm.ts:297` has a third, slightly-divergent `normalizeRealmURL` that trailing-slashes `pathname` instead of `href`. That's three copies of the canonicalization the permission lookup keys on — if the keying convention ever changes they'll drift silently. Worth extracting a single shared `normalizeRealmURL` util.

packages/realm-server/handlers/handle-delegate-session.ts
```diff
@@ -0,0 +1,187 @@
...
+
+const log = logger('realm:delegate-session');
+
+function headerValue(
+  headers: Koa.Context['req']['headers'],
+  name: string,
+): string | undefined {
+  let value = headers[name];
+  return Array.isArray(value) ? value[0] : value;
+}
```
jurgenwerk: [Claude Code 🤖] This `headerValue()` helper reimplements Koa's `ctxt.get()`. Every other handler reads headers that way (`handle-search-v2.ts`, `handle-realm-info.ts`, `serve-index.ts`, …). Could drop the helper and use `ctxt.get(DELEGATION_TIMESTAMP_HEADER)` / `ctxt.get(DELEGATION_SIGNATURE_HEADER)` to match the house pattern.

packages/realm-server/handlers/handle-delegate-session.ts
```diff
@@ -0,0 +1,187 @@
...
+    // via a `users` grant, diverging from what the realm would accept.
+    let realmPermissions = await fetchRealmPermissions(
+      dbAdapter,
+      new URL(normalizedRealmHref),
+    );
+    let permissionChecker = new RealmPermissionChecker(
+      realmPermissions,
+      matrixClient,
+    );
+    if (!(await permissionChecker.can(onBehalfOf, 'read'))) {
```
jurgenwerk, line 128: [Claude Code 🤖] Minor: this mint-time `can(user, 'read')` check duplicates the access-time check the delegated branch in `realm.ts` now performs. It's deliberate (gives a clean 403 at mint, and the comment explains the mirroring), but it does put the "can this user read?" rule in two places that must stay in sync. Fine to keep — flagging so it's a conscious choice.

jurgenwerk (commented):
[Claude Code 🤖] Reviewed this on Matic's behalf. The security-sensitive logic holds up well: delegated tokens are unforgeable without `realmSecretSeed`, the single-realm scope check closes cross-realm replay (with a test), read-only is enforced at request time, and `delegated` can't appear on normal tokens since no other `createJWT` sets it. The throws inside the delegated branch correctly re-propagate through the surrounding catch. No correctness bugs found.

The inline notes are all cleanup/maintainability — none are blockers. One process note: the PR flags that the full QUnit suite hasn't been run locally; worth confirming CI is green before merge since the realm-side acceptance is only exercised end-to-end there.

jurgenwerk (approved):

packages/realm-server/handlers/handle-delegate-session.ts
lukemelia: OK

packages/realm-server/handlers/handle-delegate-session.ts
lukemelia: [Claude Code 🤖] Done in 22e7d22362. Extracted a single `normalizeRealmURL(string): URL | null` into `packages/realm-server/utils/realm-url.ts` and routed all three call sites through it (delegate-session, upsert-realm-user-permission, and delete-realm — whose divergent local copy is gone). Standardized on the trailing-slash-the-pathname form, which yields output identical to the `ensureTrailingSlash(href)` form once search/hash are cleared, so no keying behavior changes.

packages/realm-server/handlers/handle-delegate-session.ts
lukemelia: [Claude Code 🤖] Done in 22e7d22362. Dropped the `headerValue()` helper and switched to `ctxt.get(...)` to match the house pattern. Behavior is unchanged: `ctxt.get()` returns `""` for a missing header, and `verifyDelegationRequest` gates on `if (!timestamp || !signature)`, so a missing header is rejected exactly as the previous `undefined` was — the existing "no signature or timestamp → 401" test still covers it.

PR #5291, opened 2026-06-19
refactor: migrate boxel-cli and vscode-boxel-tools search to /_federated-search-v2 (data-only)

habdelra:
Moves the two remaining direct-HTTP search consumers — boxel-cli and vscode-boxel-tools — off the legacy `/_federated-search` and `/_search` endpoints onto the v2 `search-entry` endpoints. Both issue **data-only** queries; neither renders HTML.

## boxel-cli

`commands/search.ts` posts to `/_federated-search-v2`. Callers still author ordinary card-rooted queries; `search()` rewrites them into the `item.`-addressed search-entry grammar (`type`/`on` → `item.on`, field operators and sort keys gain the `item.` prefix), requests `fields[search-entry]=item`, and returns the `item` serializations the endpoint links in `included` — the same `card`/`file-meta` resources the old top-level `data` array carried. `SearchResult` and the CLI surface are unchanged; `parse.ts` consumes `search()` results and needs no change.

## vscode-boxel-tools

`skills.ts` skill discovery posts to `_search-v2` with the data-only fieldset, anchored on `item.on` Skill and sorted by `item.title`, reading the skill `item` serializations out of `included`.

## Why data-only

`fields[search-entry]=item` resolves to `html: false` server-side, so no prerendered `html` (and therefore no scoped `css`, which only rides under `html.styles`) is emitted — the payload carries only the serializations, never markup.

## Tests

- New `search-query.test.ts` pins the card-rooted → search-entry translation and the item-from-`included` flattening, including that the data-only fieldset is always requested.
- The `ingest-card-graph` fake realm speaks the v2 data-only contract and asserts every request carries `fields[search-entry]=item`.
- The boxel-cli search + spec-discovery integration tests pass against a live `/_federated-search-v2`.

With these moved, the legacy `/_federated-search` and `/_search` endpoints have no remaining consumers.


packages/boxel-cli/src/commands/search.ts
```diff
@@ -19,12 +19,188 @@ export interface SearchCommandOptions {
...
+// A search-entry result links its serialization through `item`; the
+// `card`/`file-meta` resource itself travels in `included`.
+interface SearchEntryDoc {
+  data?: {
+    relationships?: {
+      item?: { data?: { type: string; id: string } };
+    };
+  }[];
+  included?: { type: string; id: string }[];
+}
```
habdelra, line 165: isn't this a type that we ca nimport from runtime common (type only import)

packages/boxel-cli/src/commands/search.ts
```diff
@@ -19,12 +19,188 @@ export interface SearchCommandOptions {
...
+export function itemsFromSearchEntryDoc(
+  doc: SearchEntryDoc,
+): Record<string, unknown>[] {
+  let byIdentity = new Map<string, Record<string, unknown>>();
+  for (let resource of doc.included ?? []) {
+    if (resource.type === 'card' || resource.type === 'file-meta') {
+      // NUL-separated so a (type, id) pair can't alias another by
+      // concatenation (no type or id contains a NUL byte).
+      byIdentity.set(
+        `${resource.type}\u0000${resource.id}`,
```
habdelra: i think we have an export for teh \u0000 separqtor that we can use in runtime common--this should be a constant

packages/boxel-cli/src/commands/search.ts
```diff
@@ -19,12 +19,188 @@ export interface SearchCommandOptions {
...
+      );
+    }
+  }
+  let items: Record<string, unknown>[] = [];
+  for (let entry of doc.data ?? []) {
+    let ref = entry.relationships?.item?.data;
+    if (!ref) {
+      continue;
+    }
+    let item = byIdentity.get(`${ref.type}\u0000${ref.id}`);
```
habdelra: ditto

packages/boxel-cli/src/commands/search.ts
habdelra: [Claude Code 🤖] Verified this one — we can't import the canonical `SearchEntryCollectionDocument` here, even type-only. It lives in `document-types.ts`, which transitively reaches `query.ts → index.ts` and `code-ref.ts`, so importing it (I tested it) pulls the index's `https://cardstack.com/base/*` imports into boxel-cli's type-check — ~24 `Cannot find module 'https://cardstack.com/base/…'` errors. boxel-cli is a plain Node CLI with no card-api loader, so those modules resolve at neither type- nor run-time — the same boundary `parse.ts` documents (and why it inlines `SPEC_TYPE`).

Kept `SearchEntryDoc` as a narrow local structural type and added a comment explaining why. If you want the shared type, it'd take a separate runtime-common type-layering change (lifting the wire types + `CodeRef` off the card-api-coupled modules) — happy to do that as a follow-up.

packages/boxel-cli/src/commands/search.ts
habdelra: [Claude Code 🤖] Done. `resourceIdentity` was the right helper, but it lived in `resource-types.ts`, which cascades the index's url-style base imports into the CLI. Extracted it (+ a `RESOURCE_IDENTITY_SEPARATOR` constant) into a dependency-free `runtime-common/resource-identity.ts` leaf, re-exported from `resource-types` so every existing importer is unchanged, and this lookup now uses `resourceIdentity(...)` instead of the hand-rolled key. (2513ca3)

packages/boxel-cli/src/commands/search.ts
habdelra: [Claude Code 🤖] Same fix as the line-177 thread — this lookup now uses the shared `resourceIdentity(ref.type, ref.id)`. (2513ca3)

packages/vscode-boxel-tools/src/skills.ts
```diff
@@ -378,23 +378,29 @@ export class SkillList extends vscode.TreeItem {
       headers['Authorization'] = jwt;
     }
 
-    const searchUrl = new URL('./_search', this.realmUrl);
+    // `_search-v2` speaks the search-entry wire grammar: entry membership is
+    // addressed through `item.` (the card serialization), so the type anchor
+    // is `item.on` and sort keys carry the `item.` prefix. Request the
+    // data-only fieldset (`fields[search-entry]=item`) — skill discovery never
+    // renders HTML, so each entry carries only its full `item` serialization.
+    const searchUrl = new URL('./_search-v2', this.realmUrl);
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] This resolution is unchanged from the `./_search` call it replaces, and the realm URLs passed in are realm roots that carry a trailing slash — both the hardcoded ones (`https://app.boxel.ai/base/` etc.) and the `realmAuth.getRealmUrls()` ones — so `new URL('./_search-v2', realmUrl)` resolves correctly. Not a regression introduced here; leaving it as-is to keep the migration scoped.

packages/boxel-cli/tests/commands/search-query.test.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  it('throws on a filter member it cannot translate', () => {
+    expect(() =>
+      searchEntryRequestBody({ filter: { bogus: 1 } }, ['https://realm/a/']),
+    ).toThrow(/cannot translate filter member "bogus"/);
+  });
+});
+
+describe('itemsFromSearchEntryDoc — flatten a data-only search-entry doc to items', () => {
+  it('resolves each entry`s item from included, in entry order', () => {
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed — the test name is now a double-quoted string with the apostrophe (`entry's`). (2513ca3)

backspace (approved):

PR #5299, opened 2026-06-22
Fix host browser-bundle Node-builtin polyfills broken by ESM migration

richardhjtan:
## Problem

`dev-all` loops forever on `[wait-for-host-standby]` because the host's `/_standby` page crashes during Ember boot, so the `#standby-ready` marker never renders and the prerender's standby probe times out on every attempt.

Screenshot 2026-06-22 at 11 32 35 AM

The Vite 8 / rolldown migration dropped the automatic Node-builtin polyfilling that the previous esbuild-based optimizer provided. The host's `crypto`/`stream` CJS dependency chains now blow up in the browser one Node global at a time:

1. `stream-browserify` → `readable-stream` calls `require('util')`. With no `util` browser alias, rolldown emits a runtime `__require('util')` that throws: *"Calling `require` for 'util' in an environment that doesn't expose the `require` function."*
2. With `util` fixed, the next crash: `crypto-browserify` → `create-hash` → `ripemd160` → `readable-stream` reads `process.version.slice(...)` / `process.browser` at module-eval time. The inline `process` stub in `index.html` only defined `{ env, cwd }`, so this threw *"Cannot read properties of undefined (reading 'slice')."*

## Fix

- Add a browser polyfill alias for `util` (`util: require.resolve('util/')` in `vite.config.mjs`), declare `util` as a host dependency, and add it to the workspace catalog.
- Flesh out the inline `process` global in `index.html` with `browser`, `version`, `versions`, `platform`, and `nextTick` so the readable-stream module-eval checks pass.

## Verification

With both changes, the standby probe boots cleanly:

```
[wait-for-host-standby] browser-ready after 33s (attempt 2)
```

The `require('util')` and `reading 'slice'` page errors no longer appear in the host log, and `wait-for-host-standby` / `dev-all` no longer loops.

### DIFF
--- a/packages/host/index.html
+++ b/packages/host/index.html
@@ -56,6 +56,13 @@
     <script>
       globalThis.process = {
         env: {},
+        browser: true,
+        version: '',
+        versions: {},
+        platform: 'browser',
+        nextTick(fn, ...args) {
+          Promise.resolve().then(() => fn(...args));
+        },
         cwd() {
           return '';
         },
--- a/packages/host/package.json
+++ b/packages/host/package.json
@@ -186,6 +186,7 @@
     "tracked-built-ins": "catalog:",
     "turndown": "^7.2.4",
     "typescript": "catalog:",
+    "util": "catalog:",
     "uuid": "catalog:",
     "vite": "^8.0.8",
     "wait-for-localhost-cli": "catalog:",
--- a/packages/host/tests/index.html
+++ b/packages/host/tests/index.html
@@ -28,16 +28,13 @@
     <script>
       globalThis.process = {
         env: {},
+        browser: true,
         version: '',
-        cwd() {
-          return '/';
+        versions: {},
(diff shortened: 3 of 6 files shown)


jurgenwerk (approved):

jurgenwerk:
Maybe this is is unrelated to this branch but when I test this on my local and try to open catalog workspace, it fails with this error:

image

I tried reindexing the catalog and the issue persists.

richardhjtan:
> Maybe this is is unrelated to this branch but when I test this on my local and try to open catalog workspace, it fails with this error:
> 
> image
> I tried reindexing the catalog and the issue persists.

Yes @jurgenwerk, I face the same error in catalog index card too. Here is the fix in a separate PR: https://github.com/cardstack/boxel/pull/5301

habdelra (approved):

PR #5302, opened 2026-06-22
fix: boxel-cli __dirname under ESM breaks build-plugin / publish

jurgenwerk:
## What broke
The **boxel-cli publish** workflow on `main` is failing at the "Regenerate plugin synopsis and skills" step:

```
ReferenceError: __dirname is not defined in ES module scope
    at packages/boxel-cli/src/commands/parse.ts:69
```

## Why
The recent ESM migration switched scripts from `ts-node` (CommonJS, `__dirname` defined) to native `node` ESM (`__dirname` undefined). `build-plugin.ts` imports the command tree as ES modules, and `parse.ts` runs `findBoxelCliRoot(__dirname)` at the top level — so just importing it throws.

The migration codemod missed these boxel-cli files because they normally run as the esbuild **CJS bundle**, where `__dirname` exists. The shipped `boxel` CLI was fine; only loading the raw source as ESM (what `build-plugin` does) trips it.

## Fix
- Source now uses **`import.meta.dirname`** (native ESM) in the five spots that used `__dirname` (`parse.ts`, `test-engine.ts` ×3, `index.ts`).
- esbuild's CJS bundle doesn't fill `import.meta.dirname` (it resolves to `undefined`), so `build.ts` adds an esbuild `define` mapping `import.meta.dirname` → `__dirname` — which the single-file CJS bundle provides natively as the `dist/` dir.

So it works in both modes: native ESM source uses the real `import.meta.dirname`; the bundle rewrites it to `__dirname`.

## Verified
- `node scripts/build-plugin.ts` runs clean (the failing CI step).
- Bundled `boxel parse --help` and `boxel --help` resolve paths correctly (the top-level call that was throwing).
- `tsc` + `eslint` clean; boxel-cli unit tests pass.

CS-11688.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -94,6 +94,9 @@ module.exports = {
       // Ban the CommonJS-only `__dirname`/`__filename` globals in TS source under
       // `src/` and `scripts/` — the surface that runs or gets imported as native
       // ESM, where they are `undefined`. Bundle-only modules opt out per-line.
+      // Packages with their own `root: true` ESLint config don't inherit this and
+      // must re-declare it; among the native-ESM (`type: module`) packages only
+      // `ai-bot` is `root: true`, and it does (see its `.eslintrc.cjs`).
       files: ['**/src/**/*.{ts,mts}', '**/scripts/**/*.{ts,mts}'],
       rules: {
         'no-restricted-globals': ['error', ...CJS_GLOBALS_IN_ESM],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,6 +2,7 @@
 
 const {
   NO_COMPILATION_REQUIRED_TS_SELECTORS,
+  CJS_GLOBALS_IN_ESM,
 } = require('./eslint/erasable-syntax-selectors.cjs');
 
 const DATA_TEST_SELECTORS = [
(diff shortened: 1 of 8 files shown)


habdelra:
@lukemelia maybe we should audit the mono repo for more uses of __dirname ?

habdelra (approved):

lukemelia:
[Claude Code 🤖] Ran the audit — good news: after this PR, boxel-cli is the only exposure. There are zero `__dirname`/`__filename` references in non-`.cjs` source of the nine `type: module` packages (ai-bot, billing, bot-runner, matrix, postgres, realm-server, realm-test-harness, runtime-common, software-factory). The CS-11449 codemod already moved those off Node globals; boxel-cli got missed only because it isn't `type: module` and its source gets loaded as ESM by `build-plugin.ts`.

The catch is that a one-time audit can't *stay* clean, and this bug is invisible at runtime — on the normal path these files run as the esbuild CJS bundle (where `__dirname` is real), so nothing throws until the raw source is loaded as ESM. So I put up a durable lint guard instead: #5307 (stacked on this PR) bans `__dirname`/`__filename` in `src/`/`scripts/` TS source via `no-restricted-globals`.

It's deliberately not a blanket ban — `test-engine.ts` here is a legitimate bundle-only `__dirname` (the software-factory consumer can't use `import.meta`), and that case isn't statically distinguishable from the broken one. So the rule makes CJS globals an opt-in, documented exception: bundle-only modules opt out per line with `// eslint-disable-next-line no-restricted-globals -- <reason>`. Drafted for now to get your take on the scope.

PR #5304, opened 2026-06-22
Add `overlays` opt-out to search results so rows can render without operator-mode overlays

lucaslyl:
linear: https://linear.app/cardstack/issue/CS-11689/add-overlays-opt-out-to-search-results-so-rows-can-render-without

Adds an `overlays` flag (default `true`) to the v2 `<SearchResults>` surface,
because a consumer may not want results to carry the operator-mode overlay (the
card-type chip / options menu / selection toggle) — which lets the user trigger
it and click through to open the card. With `overlays={{false}}` rows render
plainly, while still being live cards.

`HydratableCard` blocks both overlay-registration paths when opted out: the
inert HTML's `trackElement` returns the no-op modifier, and the card context
re-provided to the live `CardRenderer` subtree has `cardComponentModifier`
blanked, so a hydrated row can't re-acquire an overlay either.

## Adds an integration test:
with the tracker present, an `overlays={{false}}` row never registers (inert or live) while still hydrating on hover.

### DIFF
--- a/packages/host/app/components/card-search/hydratable-card.gts
+++ b/packages/host/app/components/card-search/hydratable-card.gts
@@ -4,7 +4,7 @@ import Component from '@glimmer/component';
 import { cached, tracked } from '@glimmer/tracking';
 
 import { modifier } from 'ember-modifier';
-import { consume } from 'ember-provide-consume-context';
+import { consume, provide } from 'ember-provide-consume-context';
 
 import {
   CardContextName,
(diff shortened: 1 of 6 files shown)


lucaslyl:
@codex review

habdelra:
@lukemelia is the adorn system something that you can opt out of?

richardhjtan (approved):

PR #5310, opened 2026-06-22
Migrate host-mode head prefetch to the unified search API

habdelra:
Moves the host-mode published-view head prefetch off the legacy `/_federated-search-prerendered` endpoint and onto the unified search API.

`host-mode-service`'s `fetchPrerenderedHead` now issues a `search-entry`-rooted query against `/_federated-search-v2`: an html-only fieldset (`fields[search-entry]=['html']`) selecting `html.format: head`, scoped to the single card via `cardUrls`. It reads the head markup off the resolved `html` resource in `included` (reached through the entry's `html` relationship) instead of the legacy top-level `data[0].attributes.html`. The `<head>` injection — single-title handling, the boundary-marker replacement, and the undefined/null/string return contract — is unchanged.

No server change was needed: `head` is already projected on the v2 card path (`renderSet` selects `head_html`, and the engine enumerates a `head` candidate at the row's native type). This was the only consumer still calling one of the four legacy endpoints.

## Tests
- Engine: `html.format: head` returns the card head rendering, including the `cardUrls`-scoped form the host issues (head was previously exercised only against a synthetic rendering universe).
- Host: the head prefetch issues the head query in the expected shape and injects the returned `<head>` markup on the published view.

### DIFF
--- a/packages/host/app/services/host-mode-service.ts
+++ b/packages/host/app/services/host-mode-service.ts
@@ -314,6 +314,7 @@ export default class HostModeService extends Service {
       method: 'QUERY',
       headers: {
         Accept: 'application/vnd.card+json',
+        'Content-Type': 'application/json',
       },
       credentials: 'include',
       body: JSON.stringify({
--- a/packages/host/app/services/host-mode-service.ts
+++ b/packages/host/app/services/host-mode-service.ts
@@ -304,8 +304,12 @@ export default class HostModeService extends Service {
     ) {
       realmServerURL = hostModeOrigin;
     }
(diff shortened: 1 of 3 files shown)


habdelra:
[Claude Code 🤖] Addressed the low-confidence note in 933195afd0 — added `Content-Type: application/json` to the head search request, matching the other host search callers. Not a functional fix (the realm server parses the QUERY body via `request.json()` regardless of `Content-Type`), just request-shape consistency.

backspace (approved):

PR #5314, opened 2026-06-23
Fix realm config card opening as a .json file

lukemelia:
## Problem

[CS-11622](https://linear.app/cardstack/issue/CS-11622) — after editing a realm's Realm Config card and reopening it, the card rendered as a `.json` File Def instance instead of a card. A browser refresh brought it back as a card.

## Root cause

The realm config has two identities that differ only by extension: the **card** `…/realm` and the **file** `…/realm.json`. The host store (`gc-card-store.ts`) keyed its **file-meta** bucket after stripping the `.json` extension — the same normalization the **card** bucket uses. That strip is correct for cards (their ids never carry an extension) but wrong for FileDef rows, which keep theirs. So a FileDef for `…/realm.json` was stored under `…/realm`, colliding with the card id.

The stack-item classifier peeks the card id as file-meta:

```
detectStackItemTypeForTarget('…/realm')
  → store.peek('…/realm', { type: 'file-meta' })  // finds the colliding FileDef
  → returns 'file'
```

…so the card opened as a `.json` file. Latent for any `.json` file that is also a card; the realm config is the obvious victim because it's routinely touched both as a card and as a file. The intermittency matched: it only bit when a `realm.json` FileDef happened to be resident, and a refresh cleared the store.

## Fix

Key file-meta by its full URL (extension included); the card-bucket strip stays as-is. Deletion now targets the file-meta bucket by the un-stripped url, which:

- lets the card id no longer misread as a file (the reported bug), while clicking the actual `.json` file still works;
- fixes a would-be GC leak — the file-meta sweep deletes `…/realm.json` by its real key; and
- cleanly separates the two identities — deleting the card no longer evicts the same-named FileDef, and vice versa.

`makeTracked` was made consistent for the same reason (currently unused, but left correct to avoid a latent trap). As a side effect, file-meta reads/writes now clear `#gcCandidates` under the same un-stripped key the sweep uses, fixing a pre-existing key mismatch there too.

## Test

Added a regression test in `garbage-collection-test.ts`: store a card `…/realm` and a `…/realm.json` FileDef together, then assert the FileDef resolves only by its `.json` URL, the card id does **not** resolve to it, deleting the card leaves the FileDef in place, and deleting by the `.json` URL actually removes it.

Verified as a real red→green check against the running host harness (the full `Unit | identity-context garbage collection` module, 69 tests):

| State | Module | CS-11622 test |
|---|---|---|
| fix reverted (test only) | 67 passed, **2 failed** | 3 passed, **2 failed** |
| fix applied | **69 passed, 0 failed** | **5 passed, 0 failed** |

The two assertions that go red without the fix are the card-id collision read and the delete-separation — the ones tied to the bug. `ember-tsc` and `eslint` are clean.

### DIFF
--- a/packages/host/app/lib/gc-card-store.ts
+++ b/packages/host/app/lib/gc-card-store.ts
@@ -480,6 +480,10 @@ export default class CardStoreWithGarbageCollection implements CardStore {
   }
 
   delete(id: string): void {
+    // Retain the un-stripped url so the matching file-meta row (keyed with its
+    // extension) is actually deleted; the card-identity logic below operates on
+    // the stripped id, as card ids never carry an extension.
+    let rawId = id;
     id = id.replace(/\.json$/, '');
     let localId = isLocalId(id, this.#virtualNetwork) ? id : undefined;
     let remoteId = !isLocalId(id, this.#virtualNetwork) ? id : undefined;
@@ -508,7 +512,7 @@ export default class CardStoreWithGarbageCollection implements CardStore {
           this.#idResolver.removeByRemoteId(id);
         }
       }
-      this.deleteFromAll(remoteId);
+      this.deleteFromAll(remoteId, rawId);
       this.#idResolver.removeByRemoteId(remoteId);
     }
   }
@@ -669,6 +673,8 @@ export default class CardStoreWithGarbageCollection implements CardStore {
   }
 
   makeTracked(remoteId: string) {
+    // File-meta is keyed by the full URL; card buckets by the stripped id.
+    let fileMetaId = remoteId;
     remoteId = remoteId.replace(/\.json$/, '');
     let instance = this.#nonTrackedCardInstances.get(remoteId);
     if (instance) {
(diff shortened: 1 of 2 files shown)


lukemelia (commented):
[Codex] I found one deletion-path issue that looks worth fixing before merge.

packages/host/app/lib/gc-card-store.ts
```diff
@@ -508,7 +512,7 @@ export default class CardStoreWithGarbageCollection implements CardStore {
           this.#idResolver.removeByRemoteId(id);
         }
       }
-      this.deleteFromAll(remoteId);
+      this.deleteFromAll(remoteId, rawId);
```
lukemelia: [Codex] This still lets deleting or GC'ing the `.json` FileDef remove the same-named card. `delete(fileUrl)` strips `id` to `.../realm` before computing `remoteId`, then this call deletes card buckets by that stripped `remoteId` while deleting file-meta by `rawId`. So when the file-meta sweep calls `this.delete(instance.id)` for `.../realm.json`, `deleteFromAll()` removes `#cardInstances['.../realm']` as well. The new regression test deletes the card before deleting the file, so it doesn't catch the inverse case. Please separate the file-meta-only delete path for `.json` ids, or add an assertion that `store.getCard(cardId)` still returns the card after `store.delete(fileUrl)`.

lukemelia: [Claude Code 🤖] Good catch — confirmed and fixed in d8daecedb7.

You're right: `delete('…/realm.json')` stripped `.json` before computing `remoteId`, then `deleteFromAll` re-stripped and deleted `#cardInstances['…/realm']`, so the GC sweep deleting a resident `.json` FileDef (`this.delete(fileDef.id)`) would evict the live card. And since `.json` is the card-source extension, this wasn't unique to the realm config — it was any card whose source FileDef happened to be resident.

Fix takes the route you suggested: a `.json` url now goes through a file-meta-only path (`deleteFileMeta`) and never runs the card-identity deletion. `deleteFromAll` reverts to a single arg and still clears file-meta for non-`.json` file urls (`.png`, `.gts`, …), which is the reason I couldn't just strip the extension everywhere.

The regression test now asserts both directions, including the inverse you flagged: after `store.delete(fileUrl)`, `store.getCard(cardId)` still returns the card. That assertion fails on the prior code and passes now.

backspace (approved):

PR #5317, opened 2026-06-23
ai-bot: read a user's realm with a short-lived, read-only token

jurgenwerk:

## Why we need this

ai-bot needs to **read realm content on a user's behalf** for the pull-model skill loader (CS-11554): the bot reads skills out of the realms a user can access (their own realm, and any other realm they have read access to).

Today the only ways to do that are bad:

- give `@aibot` a **standing read grant on every realm** — far more access than it needs, all the time, with a big blast radius if the bot's credentials leak; or
- don't let the bot read user realms at all.

This adds a third, least-privilege option: the bot asks for **just-in-time access** that is read-only, scoped to a single realm, expires in 30 minutes, is tied to the specific user, and can never read more than that user could already read themselves. So the bot reads exactly what it needs, when it needs it, without holding broad standing access.

## What this does

The bot asks the realm server for a short-lived token — **30 minutes, one realm, read-only, tied to that user** — and uses it to read that realm on the user's behalf. It works for any realm the user can already read, not just their own.

The realm server already shipped the endpoint that hands out these tokens (#5287). This PR is the ai-bot side, plus the small shared module both sides use to sign and check the request so their formats can't drift apart.

## What's in the PR

- **Shared sign/verify module** that ai-bot and the realm server both import — one definition, no copy-paste.
- **ai-bot token manager**: gets a token per (user, realm), caches it, and refreshes it before it expires.
- **Wiring in `main.ts`** behind `AI_BOT_DELEGATION_SECRET`. If that secret isn't set, the feature is simply off.

Nothing calls this at runtime yet — the consumer is the pull-model skill loader (CS-11554). The manager is wired up but dormant until then.

## How I tested it

There's no chat flow that triggers this yet, so I drove it directly against a local stack with the realm server started with `AI_BOT_DELEGATION_SECRET` set:

1. Asked for a token for `@user:localhost` on the `catalog` realm.
2. Token came back **read-only, 30-min TTL, scoped to just that realm**.
3. Used it to **read** the realm → **200** ✅
4. Used it to **write** the realm → **403** ✅ (read-only is enforced — and that user can't write catalog anyway)
5. Write with **no token** → **401** (writes really are gated)

Plus the automated tests below; lint + type-check clean.

## Tests

`packages/ai-bot/tests/user-delegated-realm-server-session-test.ts` (9 tests): the client signs in a way the server accepts; correct endpoint/origin; error-code mapping; and the manager's disabled path, cache hit, proactive refresh, per-key isolation, origin derivation, and invalidate. The realm-server `delegate-session` tests cover the server side.


lukemelia (approved):
Copilot issues look important to address, otherwise LGTM

PR #5319, opened 2026-06-23
Lazy migration on host boot — seed app.boxel.realm-servers from legacy realms

lukemelia:
## What

Self-healing migration: an account with only the legacy `app.boxel.realms` list gets an `app.boxel.realm-servers` entry without a privileged batch job. On host boot, when `app.boxel.realm-servers` is absent (boot falls back to the legacy `app.boxel.realms` list), derive the realm-server(s) backing the account's existing realms and persist them to the new key. Subsequent fresh boots take the authoritative trusted-servers assembly path.

## How

- `MatrixService.start()` — in the legacy-fallback branch, when realms exist, derive and write `app.boxel.realm-servers` via `RealmServerService.getRealmServersForRealms(realms)` + the existing `setRealmServersInAccountData` helper. Wrapped best-effort so a derivation failure never breaks boot.
- `RealmServerService.getRealmServersForRealms(realms)` — derives each realm's server from its JWT `realmServerURL` claim (read from the per-realm session token), falling back to **this host's own realm server** for realms with no token. It never uses the bare realm-URL origin: a realm's origin can differ from its realm server (e.g. the shared base realm at `cardstack.com`), and persisting such a foreign origin would make the next boot's `assertOwnRealmServer` reject the list and log the user out. Results are normalized via `normalizeRealmServerURL`, so realms backed by one server collapse to a single entry.
- The legacy `app.boxel.realms` key is **retained** (not deleted) for rollback safety.
- Sticky legacy path: a session that booted from the legacy realm list stays on that path for the lifetime of the `MatrixService` instance (`bootedFromLegacyRealmsList`). The migration's write to `app.boxel.realm-servers` only takes effect on the next fresh session — switching the same instance to the trusted path would re-derive the list from `_realm-auth` for no benefit and could drop realms the trusted servers don't advertise.
- The `app.boxel.realm-servers` AccountData listener short-circuits while `bootedFromLegacyRealmsList` is set, so neither the migration's own write nor the echo from `startClient()`'s initial sync re-runs trusted-servers assembly mid-boot.
- Idempotent: gated on `app.boxel.realm-servers` being genuinely empty, so a re-boot of the same legacy session (where the key is now present) doesn't re-write it; the migration fires at most once per account, and is a no-op for accounts with no realms.
- Session-token localStorage parsing is centralized in a shared `readSessionTokens()` helper.

## Acceptance criteria

- [x] An account with only `app.boxel.realms` gets `app.boxel.realm-servers` written on next boot, derived from each realm's JWT `realmServerURL` claim with an own-server fallback (never the bare realm-URL origin).
- [x] An account that already has `app.boxel.realm-servers` is untouched.
- [x] `app.boxel.realms` is retained (not deleted).
- [x] A legacy-booted session stays on the legacy path for its lifetime; the migration write doesn't flip it mid-boot.

## Tests

- Boot migration fires for a legacy account and writes the derived realm-server (own server, not the base-realm origin); legacy `app.boxel.realms` retained; available realms still assemble.
- The migration self-write echo — including the `app.boxel.realm-servers` event re-emitted by `startClient()`'s initial sync — does not flip the boot to the trusted path.
- A re-boot of the same migrated session stays on the legacy path (realm-servers not re-derived or duplicated).
- Already-migrated account: `app.boxel.realm-servers` left unchanged (migration does not run).

### DIFF
--- a/packages/host/app/services/matrix-service.ts
+++ b/packages/host/app/services/matrix-service.ts
@@ -182,6 +182,12 @@ export default class MatrixService extends Service {
   // gains content at runtime. Login-related side effects (`loginToRealms`,
   // `loadMoreAuthRooms`) still run regardless.
   private trustedRealmServersAuthoritative = false;
+  // Set only while the boot-time lazy migration writes `app.boxel.realm-servers`
+  // itself, so the AccountData listener ignores that self-write rather than
+  // re-running trusted-servers assembly mid-boot. The migration has already
+  // assembled this session from the equivalent legacy realm list; the new key
+  // only needs to take effect on the next boot.
+  private migratingRealmServersAccountData = false;
   @tracked private _currentRoomId: string | undefined;
   @tracked private timelineLoadingState: Map<string, boolean> =
     new TrackedMap();
@@ -421,6 +427,13 @@ export default class MatrixService extends Service {
               break;
             }
             case APP_BOXEL_REALM_SERVERS_EVENT_TYPE: {
+              // The boot-time lazy migration's own write echoes back here; it
+              // has already assembled this session from the equivalent legacy
+              // realm list, so ignore the self-write rather than re-running
+              // assembly mid-boot.
+              if (this.migratingRealmServersAccountData) {
+                break;
+              }
               let realmServers = e.event.content.realmServers as string[];
               this.trustedRealmServersAuthoritative = realmServers.length > 0;
               if (this.trustedRealmServersAuthoritative) {
@@ -942,7 +955,16 @@ export default class MatrixService extends Service {
               if (derivedRealmServers.length > 0) {
                 if (isTesting())
                   console.warn('[start-phase] migrateRealmServersAccountData');
-                await this.setRealmServersInAccountData(derivedRealmServers);
+                // Guard so this self-write doesn't re-trigger trusted-servers
+                // assembly via the AccountData listener while boot is still
+                // running — this session is already assembled from the legacy
+                // list; the new key takes effect on the next boot.
+                this.migratingRealmServersAccountData = true;
+                try {
+                  await this.setRealmServersInAccountData(derivedRealmServers);
+                } finally {
+                  this.migratingRealmServersAccountData = false;
+                }
               }
             } catch (err) {
               console.error(
--- a/packages/host/app/services/matrix-service.ts
+++ b/packages/host/app/services/matrix-service.ts
@@ -954,23 +954,23 @@ export default class MatrixService extends Service {
           )) as { realms: string[] } | null;
           userRealmURLs = legacyRealmsData?.realms ?? [];
 
-          // Lazy migration: this account predates `app.boxel.realm-servers`
-          // (the key was absent or empty, so boot fell back to the legacy
-          // realm list above). Seed the new key with the realm-server backing
-          // the user's existing realms so future boots take the authoritative
-          // trusted-servers assembly path. We use `getRealmServersForRealms`,
-          // which derives the server from each realm's JWT `realmServerURL`
-          // claim and falls back to this host's own realm server — never the
-          // bare realm-URL origin. That matters because a realm URL's origin
-          // can differ from its realm server (e.g. the shared base realm at
-          // cardstack.com); persisting such a foreign origin would make the
-          // next boot's `assertOwnRealmServer` reject the list and log the
-          // user out. The legacy `app.boxel.realms` key is intentionally
-          // retained for rollback safety during the transition window. Gated
-          // on `trustedServers` being genuinely empty so a re-boot of this
-          // already-legacy session (where the key we just wrote is now
-          // present) doesn't re-write it. A no-op for an account with no
-          // realms. Best-effort: a failure must not break boot.
+          // Lazy migration: this account has no `app.boxel.realm-servers`
+          // entry (the key was absent or empty, so boot fell back to the
+          // legacy realm list above). Seed the new key with the realm-server
+          // backing the user's existing realms so subsequent boots take the
+          // authoritative trusted-servers assembly path. We use
+          // `getRealmServersForRealms`, which derives the server from each
+          // realm's JWT `realmServerURL` claim and falls back to this host's
+          // own realm server — never the bare realm-URL origin. That matters
+          // because a realm URL's origin can differ from its realm server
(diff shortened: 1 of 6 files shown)


packages/host/app/services/realm-server.ts
```diff
@@ -372,6 +364,39 @@ export default class RealmServerService extends Service {
...
+  // origin in tests) collapse to that one server.
+  deriveRealmServerURLsForRealms(realms: string[]): string[] {
+    let sessionTokens = this.readSessionTokens();
+    let realmServerURLs = new Set<string>();
+    for (let realmURL of realms) {
+      let normalizedRealmURL = ensureTrailingSlash(realmURL);
+      let token = sessionTokens[normalizedRealmURL] ?? sessionTokens[realmURL];
+      let claims = token ? realmClaimsFromRawToken(token) : undefined;
+      let realmServerURL =
+        claims?.realmServerURL ?? new URL(normalizedRealmURL).origin;
```
lukemelia: [Codex] This fallback can persist a server URL that the next boot will reject. When a legacy realm has no local session token, we write `new URL(realm).origin`; if that realm is `baseRealm.url` (already used in the legacy fallback fixture) or a custom/content-origin realm, the migrated `app.boxel.realm-servers` list can include `https://cardstack.com/` or another non-own origin. On the next fresh boot `fetchUserRealmsFromTrustedServers()` calls `assertOwnRealmServer()` before any fallback, rejects the list, and `MatrixService.start()` logs the user out. I think the migration needs to either filter/skip non-own derived origins or reuse the existing own-server fallback behavior from `getRealmServersForRealms()` when there is no JWT claim.

lukemelia (commented):
[Codex] Code review: I found one boot-regression risk in the lazy migration.

packages/host/app/services/realm-server.ts
lukemelia: [Claude Code 🤖] Good catch — confirmed and fixed in 586e7384c8.

You're right: deriving a realm's server from `new URL(realm).origin` could persist a foreign origin (`https://cardstack.com/` from the base realm, or any custom-domain realm) into `app.boxel.realm-servers`. On the next fresh boot the trusted-servers path runs `assertOwnRealmServer()` first, rejects the multi/non-own list, and `start()`'s catch calls `logout()`. It was masked in tests because in the test env the base-realm origin and the own-server origin both collapse to `localhost:4201`.

Fix: dropped the origin-based `deriveRealmServerURLsForRealms` and reused `getRealmServersForRealms()` for the migration, exactly as you suggested — it derives from each realm's JWT `realmServerURL` claim and falls back to this host's own realm server, skipping token-less realms rather than guessing their origin. So the migration now only ever persists the user's own realm server.

packages/host/tests/integration/matrix-service-boot-assembly-test.ts
```diff
@@ -182,3 +188,124 @@ module(
     });
   },
 );
+
+// CS-11659: lazy migration on host boot. A user who predates
```
backspace: Should we make `evergreen-comments` a default? (Can we?)

backspace (approved):

packages/host/tests/integration/matrix-service-boot-assembly-test.ts
lukemelia: Probably by editing AGENTS.md

packages/host/app/services/matrix-service.ts
```diff
@@ -421,6 +433,13 @@ export default class MatrixService extends Service {
               break;
             }
             case APP_BOXEL_REALM_SERVERS_EVENT_TYPE: {
+              // The boot-time lazy migration's own write echoes back here; it
+              // has already assembled this session from the equivalent legacy
+              // realm list, so ignore the self-write rather than re-running
+              // assembly mid-boot.
+              if (this.migratingRealmServersAccountData) {
+                break;
+              }
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Good catch — fixed in 3588c5b673.

Confirmed: the `migratingRealmServersAccountData` flag only spanned the synchronous `setRealmServersInAccountData()` write. `startClient()`'s initial sync then re-emits every account-data key (the adjacent comment about the `app.boxel.realms` re-emission already notes this), and that second `app.boxel.realm-servers` echo arrived with the flag back to `false` — so the listener ran trusted-servers assembly mid-boot and overwrote the legacy-assembled realm list, defeating the sticky-legacy guarantee.

Rather than the narrower `bootedFromLegacyRealmsList && !postLoginCompleted`, I guarded the whole case on `bootedFromLegacyRealmsList`. It's set before the migration write and persists for the instance lifetime, so it covers both echoes (the synchronous self-write *and* the startClient sync echo) and matches `start()`'s instance-lifetime stickiness — a legacy-booted session never switches to the trusted path until the next fresh session, in the listener as well as in `start()`. The narrow `migratingRealmServersAccountData` flag is now redundant and removed.

This was invisible in tests because the mock client only re-emitted the legacy `app.boxel.realms` event on `startClient`. I made the mock re-emit `app.boxel.realm-servers` too (mirroring the real initial sync) and added a regression test asserting a legacy-migration boot stays non-authoritative through the echo.

packages/host/app/services/realm-server.ts
```diff
@@ -332,15 +332,7 @@ export default class RealmServerService extends Service {
...
-    let sessionTokens: Record<string, string> = {};
-    let sessionStr =
-      window.localStorage.getItem(SessionLocalStorageKey) ?? '{}';
-
-    try {
-      sessionTokens = JSON.parse(sessionStr) as Record<string, string>;
-    } catch {
-      sessionTokens = {};
-    }
+    let sessionTokens = this.readSessionTokens();
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Right — the description was stale after 586e7384c8 dropped the origin-based approach. Reconciled the PR description: there is no `deriveRealmServerURLsForRealms`; the migration uses `getRealmServersForRealms()`, which derives each realm's server from its JWT `realmServerURL` claim (with this host's own realm server as the fallback for token-less realms) and never consults the bare realm-URL origin. The "How", acceptance-criteria, and "Tests" sections now match the shipped code (and cover the sticky-legacy guard added in 3588c5b673).

PR #5320, opened 2026-06-23
Remove crypto-browserify polyfill from host

lukemelia:
## What

Eliminates the `crypto-browserify` dependency from the host package.

The host's Vite config aliased bare `crypto` → `crypto-browserify` solely to satisfy one static `import { createHmac } from 'crypto'` in `runtime-common/url-signature.ts`. That import is on the module's **Node-only** synchronous signing path — which the host never executes. The host only calls `createURLSignature`, the async **WebCrypto** (`crypto.subtle`) path. The polyfill's distinctive exports were already being tree-shaken out of the built bundle; the alias only mattered at build-resolution time.

## How

- Split `runtime-common/url-signature.ts` so it holds **only** the browser/WebCrypto `createURLSignature` — no `crypto` import, safe to bundle.
- New `runtime-common/url-signature-node.ts` holds the synchronous Node path (`createURLSignatureSync`, `verifyURLSignature`), importing from `node:crypto`. (The `"./*": "./*.ts"` exports wildcard covers the new subpath.)
- Point realm-server's two consumers (`handlers/handle-download-realm.ts`, `tests/server-endpoints/download-realm-test.ts`) at the new module.
- Drop the `{ find: 'crypto', ... }` Vite alias, the `crypto-browserify` host dependency, and its catalog entry. `pnpm install` removed crypto-browserify and 34 transitive packages.

The browser and Node consumers were fully disjoint, so the split is low-risk — no caller needed both paths from one module.

## Verification

- `runtime-common`, `realm-server`, and `host` all type-check clean (`ember-tsc --noEmit`).
- `host` Vite build succeeds — confirming nothing in the bundle graph still needs a bare-`crypto` resolution.

### DIFF
--- a/packages/host/package.json
+++ b/packages/host/package.json
@@ -102,7 +102,6 @@
     "camelcase": "catalog:",
     "concurrently": "catalog:",
     "content-tag": "catalog:",
-    "crypto-browserify": "catalog:",
     "date-fns": "catalog:",
     "decorator-transforms": "catalog:",
     "ember-animated": "catalog:",
--- a/packages/host/vite.config.mjs
+++ b/packages/host/vite.config.mjs
@@ -270,7 +270,6 @@ export default defineConfig(({ mode }) => ({
   resolve: {
     alias: [
       { find: 'path', replacement: require.resolve('path-browserify') },
-      { find: 'crypto', replacement: require.resolve('crypto-browserify') },
       { find: 'stream', replacement: require.resolve('stream-browserify') },
       { find: /^util$/, replacement: require.resolve('util/') },
       // recast's main.js eagerly requires 'fs'; we stub it for the browser.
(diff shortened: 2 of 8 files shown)


lukemelia (commented):
[Codex] Reviewed the PR diff and did not find any blocking code-review issues.

I checked that the actual GitHub diff is limited to the expected package/dependency changes plus the URL-signature split; `packages/host/memory-baseline.json` is not part of the PR diff. I also ran a lightweight local parity check confirming `createURLSignature` and `createURLSignatureSync` still produce the same URL-safe HMAC for the same URL after stripping `sig`, and that `verifyURLSignature` accepts it.

Residual risk: I did not rerun the full host or realm-server test/build suites locally.

backspace (approved):

PR #5327, opened 2026-06-24
Reconcile eligible live searches against in-memory Store state

lukemelia:
## What

Wires the runtime-common instance matcher/comparator into `SearchResource` so an **eligible live search** derives its displayed result set from the server result reconciled against in-memory Store state — the list self-corrects the instant a card is created, edited, or deleted, before the realm reindexes. One-shot `StoreService.search()` and non-live searches are unaffected.

This builds the full-recompute version; an incremental / dirty-set optimization is out of scope and left for separate, measurement-gated work.

## How

- **`SearchResource.displayedInstances`** (new, backs the `instances` getter): for an eligible live search it
  - keeps every server-returned instance unless it now demonstrably fails the filter locally (an `unresolvable` predicate never removes a server result),
  - adds Store candidates that match the filter but are absent from the server result, scoped to the query's target realm(s) (`unresolvable` candidates are not added),
  - dedupes by id and re-sorts with the query's sort comparator.
- **Eligibility** (`isClientFilterEligible`): live search + matcher loaded + a complete result set (`instances.length === meta.page.total`) + a client-evaluable filter. Anything else (non-live, paginated/incomplete, `matches`/full-text) is a server-only passthrough.
- **Reactivity**: recomputes on Store mutation, not only when the server search re-runs — it reads the tracked identity map (covers create/delete) and a new `StoreService.instanceMutationVersion` signal (covers in-place edit/save).
- **`StoreService`** exposes the candidate pool (`allCardInstances` / `allFileMetaInstances`) and a `getMatchAPI()` slice for the matcher.

## Notable correctness detail

The Store keys each instance under **both** its local id and its remote id, so an un-deduped candidate pool returns the same card twice — which the merge would render as two rows. `allCardInstances` / `allFileMetaInstances` dedupe by instance identity, and the merge dedupes by id as a second guard. Regression test included.

## Demo

`experiments-realm/client-filter-playground.gts` — two live searches (active / archived) over one Store, each sorted by priority. Create, archive/activate, re-prioritize, or delete a widget and both lists reconcile in place with no `_federated-search` round-trip (the server reindex catches up a beat later). The list renders whenever it has results, so a background live-refresh doesn't blank it.

## Tests

`tests/integration/resources/search-test.ts`, `client-side Store filtering step` module:
- locally added matching card appears with no server round-trip; a server card that no longer matches locally is removed; editing a candidate toggles membership reactively; merged set is ordered per the sort.
- a surfaced candidate is not duplicated despite dual local/remote-id keying (no duplicate ids in the displayed set).
- passthrough cases: one-shot `store.search`, non-live, incompletely-loaded (paginated), and non-client-evaluable (`matches`) all stay server-only.

## Notes for reviewers

- `isLoading` is `search.isRunning`, so it flips `true` on every background live-refresh. Consumers that render a loading branch ahead of their results will blank on each round-trip — the demo orders its template to render results when present to avoid this. A resource-level `isRefreshing` / initial-load-only distinction would give consumers flash-free behavior by default; left out here to avoid changing shared `isLoading` semantics.

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -1,4 +1,8 @@
-import { registerDestructor } from '@ember/destroyable';
+import {
+  isDestroyed,
+  isDestroying,
+  registerDestructor,
+} from '@ember/destroyable';
 import type Owner from '@ember/owner';
 import { setOwner } from '@ember/owner';
 import { service } from '@ember/service';
@@ -21,6 +25,10 @@ import type {
 import {
   subscribeToRealm,
   isFileDefInstance,
+  isFileDefCodeRef,
+  isClientEvaluable,
+  matchInstanceAgainstFilter,
+  makeInstanceComparator,
   logger as runtimeLogger,
   normalizeQueryForSignature,
   buildQueryParamValue,
@@ -28,8 +36,12 @@ import {
   ri,
   RealmPaths,
   runtimeDependencyContextWithSource,
+  type CardAPIForMatching,
+  type CodeRef,
+  type MatchResult,
+  type RealmResourceIdentifier,
 } from '@cardstack/runtime-common';
-import type { Query } from '@cardstack/runtime-common/query';
+import type { Filter, Query } from '@cardstack/runtime-common/query';
 
 import type { CardDef } from 'https://cardstack.com/base/card-api';
 import type { FileDef } from 'https://cardstack.com/base/file-api';
@@ -41,6 +53,70 @@ import type RealmServerService from '../services/realm-server';
 import type StoreService from '../services/store';
 
 const waiter = buildWaiter('search-resource:search-waiter');
+
+// Retains Store references for a changing set of instance ids and reconciles
+// them as that set changes, releasing all on teardown. `addReference` /
+// `dropReference` can mutate tracked Store state (e.g. `autoSaveStates`), so
+// reconciliation is deferred to a microtask — it must never run inside the
+// tracked computation that declares the set. Used for the live-search
+// candidates surfaced by the client-side merge but absent from the server
+// result: `updateInstances` references only the server result, so without this
+// a displayed candidate at reference count zero could be swept by the Store GC
+// while still on screen.
+class CandidateReferenceRetainer {
+  #held = new Set<string>();
+  #desired = new Set<string>();
+  #flushScheduled = false;
+  #torn = false;
+
+  constructor(
+    private getStore: () => StoreService,
+    parent: object,
+  ) {
+    registerDestructor(parent, () => {
+      this.#torn = true;
+      this.#desired = new Set();
+      this.#reconcile();
+    });
+  }
+
+  // Declare the ids currently in use. Safe to call from a tracked getter — the
+  // reference mutation happens later, in a microtask, outside this frame.
+  retain(ids: string[]): void {
+    if (this.#torn) {
+      return;
+    }
+    this.#desired = new Set(ids);
+    if (this.#flushScheduled) {
+      return;
+    }
+    this.#flushScheduled = true;
+    queueMicrotask(() => {
+      this.#flushScheduled = false;
+      if (!this.#torn) {
+        this.#reconcile();
+      }
+    });
+  }
+
+  #reconcile(): void {
+    if (this.#held.size === 0 && this.#desired.size === 0) {
+      return;
+    }
+    let store = this.getStore();
+    for (let id of this.#held) {
+      if (!this.#desired.has(id)) {
+        store.dropReference(id);
+      }
+    }
+    for (let id of this.#desired) {
+      if (!this.#held.has(id)) {
+        store.addReference(id);
+      }
+    }
+    this.#held = new Set(this.#desired);
+  }
+}
 
 export interface Args<T extends CardDef | FileDef = CardDef> {
   named: {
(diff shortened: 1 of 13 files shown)


packages/host/app/resources/search.ts
```diff
@@ -337,8 +396,131 @@ export class SearchResource<
...
+    // type-gated and, when they have results, name their kind here anyway.
+    let isFileSearch =
+      serverResults.length > 0 && isFileDefInstance(serverResults[0]);
+    let candidatePool: (CardDef | FileDef)[] = isFileSearch
+      ? this.runtimeStore.allFileMetaInstances()
+      : this.runtimeStore.allCardInstances();
+
+    // Add candidates the server didn't return but that match locally, scoped to
+    // the query's target realm(s). `unresolvable` candidates are not added.
+    let added = candidatePool.filter(
```
lukemelia, line 539: [Codex] Locally added candidates are returned from `instances`, but they never go through the reference bookkeeping in `updateInstances` (`addReference`/`dropReference` only sees `_instances`, which remains the server result set). That means a local card shown only by this merge can still have a zero reference count and be swept by the Store GC while the search resource is rendering it. Please retain/release references for the merged displayed IDs, or otherwise make these candidate rows reachable for as long as they are exposed.

packages/host/app/resources/search.ts
```diff
@@ -337,8 +396,131 @@ export class SearchResource<
...
+
+    let serverIds = new Set(
+      serverResults.map((instance) => instance.id).filter(Boolean) as string[],
+    );
+
+    // The candidate pool is drawn from the same kind (card vs file-meta) as the
+    // result set. A non-empty set names its kind directly; an empty set is
+    // ambiguous, so we default to cards — file-meta searches are always
+    // type-gated and, when they have results, name their kind here anyway.
+    let isFileSearch =
```
lukemelia: [Codex] This makes empty file-meta live searches choose the card candidate pool, so they never surface a locally hydrated FileDef until the server returns at least one file row. A FileDef query can be complete with zero rows (`instances.length === meta.page.total === 0`), which passes the eligibility gate above; in that state `serverResults[0]` is unavailable and the default to `allCardInstances()` causes matching FileDef candidates to be skipped. Please derive the pool from the query/file-meta dispatch (for example the FileDef type filter or `meta.isFileMeta`) instead of from the first returned row.

packages/host/app/resources/search.ts
lukemelia: [Claude Code 🤖] Fixed. Confirmed: `updateInstances` retains references only for `_instances` (the server result), and the GC `sweep` roots reachability at `hasReferences(localId)` + transitive links — so a candidate surfaced only by the merge could sit at reference count zero and be swept while still displayed.

The merge now retains Store references for the displayed-but-not-server candidate ids via a new `CandidateReferenceRetainer`: it diffs the id set, adds/drops references accordingly, and releases everything on teardown. Reconciliation is deferred to a microtask rather than run inline in the `displayedInstances` getter, because `dropReference` can mutate tracked Store state (`autoSaveStates`) and that must not happen inside the tracked computation. Server-result references stay owned by `updateInstances`; the retainer only covers the `added` set, and the two never hold the same id simultaneously (added excludes server ids), so counts stay correct across transitions.

packages/host/app/resources/search.ts
lukemelia: [Claude Code 🤖] Fixed. The candidate pool is now dispatched from the query's target type rather than from the first returned row: a new `isFileDefCodeRef(ref)` (runtime-common) recognizes the base `FileDef` ref and the known extension subtypes, and the merge checks the filter's top-level `on`/`type` against it. So a complete file-meta search with zero server rows still selects `allFileMetaInstances()` and surfaces a locally hydrated FileDef. The first-row check remains only as a fallback for a query with no top-level type ref. Caveat worth noting: a realm-defined custom FileDef subtype that isn't in the extension map won't be recognized by ref alone — the `// TODO: realm metadata configuration` in file-def-code-ref.ts is the same limitation.

packages/host/tests/integration/resources/search-test.ts
```diff
@@ -1363,4 +1365,455 @@ module(`Integration | search resource`, function (hooks) {
...
+
+    innerHooks.afterEach(function () {
+      restoreFetch?.();
+      restoreFetch = undefined;
+    });
+
+    const bookRef = { module: testRRI('book'), name: 'Book' };
+    let abdelRahmanQuery: Query = {
+      filter: { on: bookRef, eq: { 'author.lastName': 'Abdel-Rahman' } },
+    };
```
habdelra, line 1399: this doesn't look like a v2 query. where is the `item` leg of the query?

habdelra:
can you also double check that you are not using any deprecated search surfaces?

packages/host/tests/integration/resources/search-test.ts
lukemelia: [Claude Code 🤖] That's a v1 `Query` because it's the *input* shape `SearchResource` / `@context.getCards` / `StoreService.search` consume — the unified-search guidance lists the programmatic instance path as `StoreService.search(query) → CardDef[]`. The v2 `search-entry` query, including the `item` leg, is built internally: `store.search` → `fetchAndHydrateSearchResults` → `fetchSearchDoc` → `fetchSearchDocUncoalesced`, which calls `searchEntryWireQueryFromQuery(query, { fields: ['item'] })` (store.ts:1262) and POSTs it to `_federated-search-v2` (store.ts:1276). The `{ fields: ['item'] }` selection *is* the item leg — the instance path needs the full serialization to hydrate cards into the Store. So this rides v2 end-to-end; nothing here calls `/_federated-search` or the prerendered endpoints.

The `_federated-search` substring the test's mock counts on also matches `_federated-search-v2` (the real endpoint it forwards to) — the mock just tallies round-trips so the test can assert the client-side recompute makes none.

lukemelia:
[Claude Code 🤖] Checked against the unified-search deprecation guidance — no deprecated search surfaces are used. Everything rides v2:

- `SearchResource` / `@context.getCards` → `StoreService.search` → `_federated-search-v2`, with a `{ fields: ['item'] }` search-entry query built via `searchEntryWireQueryFromQuery` (store.ts:1262, 1276).
- The playground renders its queried instances with its own markup via `@context.getCards` (the supported reactive-instance path), not `@context.prerenderedCardSearchComponent`.
- Grepping the full diff for deprecated identifiers — `/_search`, `/_federated-search`, `*-prerendered`, `getPrerenderedSearch`, `PrerenderedCardSearch`, `prerenderedCardSearchComponent`, `PrerenderedCardLike`/`PrerenderedCardData`, `SearchResultSection`/`ItemButton` — returns nothing.

See the inline thread on the test query for where the v2 `item` leg is constructed.

habdelra (approved):

PR #5334, opened 2026-06-25
Make client-side filter type gate tolerant of module-ref spellings

lukemelia:
## Problem

`main`'s host test suite is red — `markdown BFM card references` (acceptance) and `RichMarkdownField: linkedCards …` (integration) fail because query-backed fields render nothing.

This is a latent interaction between two PRs that were each green in isolation:

- **#5327** (cs-11416) added the client-side search filtering step (`SearchResource.displayedInstances`), which reconciles an eligible live search's server result against in-memory Store state. Its CI ran at 03:57 UTC.
- **#5148** ("Change cross-resource references to prefix form") canonicalized refs and landed at 13:58 UTC — *after* #5327's CI, so the two were never tested together until #5327 merged.

## Root cause

The filtering step gates each candidate by the query's `on`/`type` ref via `instanceIsType`, which compared the instance class's identified ref against the filter ref with **exact string equality** (`a.module === b.module`).

After #5148, a class identified under one module spelling (e.g. the realm's real URL `https://realm-server.ci.localhost/base/card-api`) no longer string-matched a filter ref carrying an equivalent spelling (RRI / virtual alias). The gate returned `no-match`, so reconciliation dropped a server-returned instance the **server had correctly matched** (the server tolerates equivalent spellings via `internalKeyFor` / `internalKeysFor`).

Concretely, the `linkedCards` query is `{ in: { id: [...] }, on: { module: <card-api>, name: 'CardDef' } }`. The server returns the linked card; the client's exact-string type gate drops it; nothing renders.

## Fix

Resolve both module refs through the `VirtualNetwork` before comparing in `codeRefEquals`, mirroring the tolerance the server already applies in `internalKeyFor`. `CardAPIForMatching` gains a `virtualNetwork` member (supplied by `store.getMatchAPI()`).

Added a unit-test case asserting the type gate matches across equivalent module spellings.

## Verification

- `runtime-common` and `host` type-check clean; eslint clean on changed files.
- The failing markdown acceptance + `RichMarkdownField` integration tests are the regression coverage (they exercise the exact `on`-gated query path).

### DIFF
--- a/packages/runtime-common/instance-filter-matcher.ts
+++ b/packages/runtime-common/instance-filter-matcher.ts
@@ -1,6 +1,6 @@
 import { isEqual } from 'lodash-es';
 
-import { getField, identifyCard } from './code-ref.ts';
+import { canonicalModuleKey, getField, identifyCard } from './code-ref.ts';
 import {
   isAnyFilter,
   isCardTypeFilter,
@@ -473,41 +473,6 @@ function codeRefEquals(
     );
   }
   return isEqual(a, b);
-}
-
-// Reduce a module reference to a single canonical key, collapsing every
-// equivalent spelling the VirtualNetwork knows about so two refs that point at
-// the same module compare equal regardless of how they were written.
-function canonicalModuleKey(
-  module: string,
-  virtualNetwork: VirtualNetwork,
-): string {
-  let href: string;
-  try {
-    // Resolves a prefix-form RRI (e.g. `@cardstack/base/card-api`) to its
-    // mapped URL; an absolute URL is returned unchanged.
-    href = virtualNetwork.resolveURL(module, undefined).href;
-  } catch {
-    // Unresolvable reference (e.g. a scoped prefix with no mapping): fall back
-    // to the raw form. The `a.module === b.module` fast path already covered
-    // the only way two such refs can be equal.
-    return module;
-  }
-  // `resolveURL` leaves an absolute URL untouched, so a virtual spelling and
-  // its real target (registered via `addURLMapping`, e.g.
-  // `https://cardstack.com/base/…` ↔ the deployed realm URL) stay distinct.
-  // Collapse virtual onto real here; real and unmapped URLs pass through.
-  try {
-    let real = virtualNetwork.mapURL(href, 'virtual-to-real');
-    if (real) {
-      href = real.href;
-    }
-  } catch {
-    // `href` wasn't a parseable absolute URL — leave it as resolved.
-  }
-  // Collapse the real URL onto its portable prefix form when a registered realm
-  // prefix matches, so `@scope/…` and the real URL also land on one key.
-  return virtualNetwork.unresolveURL(href);
 }
 
 // -- sort comparator ---------------------------------------------------------
--- a/packages/runtime-common/instance-filter-matcher.ts
+++ b/packages/runtime-common/instance-filter-matcher.ts
@@ -463,30 +463,51 @@ function codeRefEquals(
       return true;
     }
     // The class is identified under the module spelling it was loaded from,
-    // while the filter ref can carry an equivalent spelling (RRI / real-URL /
-    // virtual-alias). Resolve both through the VirtualNetwork before comparing
-    // so the type gate doesn't drop an instance over a cosmetic URL difference
-    // — the same tolerance the server applies in `internalKeyFor`.
+    // while the filter ref can carry an equivalent spelling (prefix RRI /
+    // real-URL / virtual-alias). Reduce both to a single canonical key before
+    // comparing so the type gate doesn't drop an instance over a cosmetic URL
+    // difference — the tolerance the server applies in `internalKeyFor`.
     return (
-      resolveModuleURL(a.module, virtualNetwork) ===
-      resolveModuleURL(b.module, virtualNetwork)
+      canonicalModuleKey(a.module, virtualNetwork) ===
+      canonicalModuleKey(b.module, virtualNetwork)
     );
   }
   return isEqual(a, b);
 }
 
-function resolveModuleURL(
+// Reduce a module reference to a single canonical key, collapsing every
+// equivalent spelling the VirtualNetwork knows about so two refs that point at
+// the same module compare equal regardless of how they were written.
+function canonicalModuleKey(
   module: string,
   virtualNetwork: VirtualNetwork,
 ): string {
+  let href: string;
   try {
-    return virtualNetwork.resolveURL(module, undefined).href;
+    // Resolves a prefix-form RRI (e.g. `@cardstack/base/card-api`) to its
+    // mapped URL; an absolute URL is returned unchanged.
+    href = virtualNetwork.resolveURL(module, undefined).href;
   } catch {
-    // A reference the VirtualNetwork can't resolve (e.g. a scoped prefix it has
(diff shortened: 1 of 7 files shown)


packages/runtime-common/instance-filter-matcher.ts
```diff
@@ -425,26 +433,62 @@ function negate(result: MatchResult): MatchResult {
...
   }
   return isEqual(a, b);
 }
 
+function resolveModuleURL(
+  module: string,
+  virtualNetwork: VirtualNetwork,
+): string {
+  try {
+    return virtualNetwork.resolveURL(module, undefined).href;
```
lukemelia: [Codex] `resolveURL()` does not apply `addURLMapping()` aliases; it only resolves absolute URLs as themselves and prefix-form RRIs through `realmMappings`. That means this still compares `https://cardstack.com/base/card-api` (or the hardcoded test realm URL registered via `addURLMapping`) against the resolved real URL as different hrefs, and the new test above registers exactly that kind of URL mapping so it should still fail. Please canonicalize URL mappings too, for example by mapping the resolved href through `virtualNetwork.mapURL(..., "virtual-to-real")` before comparing, or use a helper that mirrors the server/index canonicalization for both URL aliases and prefix mappings.

lukemelia: [Claude Code 🤖] Good catch — you're right. `resolveURL()` collapses a prefix-form RRI onto its mapped URL (via `realmMappings`/`toURL`) but leaves an absolute URL untouched, so it never applies an `addURLMapping` virtual↔real alias. That alias is exactly the client-side mismatch here (the class is identified under the virtual base URL while the query ref carries the real one), so the original comparison still missed and the new test failed.

Fixed in 615a66aa3b: `canonicalModuleKey` now does `resolveURL` → `mapURL(…, 'virtual-to-real')` → `unresolveURL`, so prefix RRIs, real URLs, and virtual aliases all reduce to one key. Verified against a real `VirtualNetwork` that `https://cardstack.com/base/card-api`, `https://realm-server.ci.localhost/base/card-api`, and `@cardstack/base/card-api` all collapse to `@cardstack/base/card-api`.

habdelra:
[Claude Code 🤖]

Heads-up: there's a **second** host failure from the same root cause that this PR's diff doesn't reach, so shard 15 will stay red after this merges.

**Failing test:** `Unit | isFileDefCodeRef: recognizes a known extension subtype ref` (Host Tests shard 15).

**Same root cause, different call site.** Just like `codeRefEquals`, `isFileDefCodeRef` compares module spellings with exact string equality — but it compares against `FILEDEF_CODE_REF_BY_EXTENSION`, whose entries are still built in **full-URL** form while #5148 moved `baseFileRef`/`baseRRI` to **prefix** form:

```ts
// packages/runtime-common/file-def-code-ref.ts
function baseModule(name: string): RealmResourceIdentifier {
  return `${baseRealm.url}${name}` as RealmResourceIdentifier;  // -> https://cardstack.com/base/markdown-file-def
}
```

The test (and the real caller, `resources/search.ts` → `isFileDefCodeRef(filterRef?.on ?? filterRef?.type)`) passes prefix-form refs: `baseRRI('markdown-file-def')` → `@cardstack/base/markdown-file-def`. The base-`FileDef` assertion passes because it compares against `baseFileRef` (which #5148 converted to prefix form); only the **extension-subtype** assertion fails, because the map didn't get converted. Net runtime effect: a file-meta live search whose `on`/`type` targets a FileDef *subtype* (`MarkdownDef`, `PngDef`, …) no longer dispatches to the file-meta candidate pool, so it skips local reconciliation.

**Suggested fix** — the one entry #5148 missed; point `baseModule` at the prefix-form root so the map matches `baseFileRef` and the prefix-form refs callers pass:

```ts
import { baseRealmRRI, baseFileRef } from './constants.ts';

function baseModule(name: string): RealmResourceIdentifier {
  return `${baseRealmRRI}${name}` as RealmResourceIdentifier;
}
```

The existing `file-def-code-ref-test.ts` is the regression coverage.

One thing to sanity-check before taking it: `resolveFileDefCodeRef` shares this map and routes non-`://` modules through `virtualNetwork.resolveURL(...)` (the branch the `.mismatch` entry already uses). Prefix entries would flow through that branch too, so its below-VN output for the indexer/realm shifts from the hardcoded `https://cardstack.com/base/…` to the VN-resolved URL. If you'd rather not touch `resolveFileDefCodeRef`'s output, the alternative is to make `isFileDefCodeRef` spelling-tolerant the same way you did `codeRefEquals` (canonicalize via the VirtualNetwork before comparing), threading the network in from `resources/search.ts`.

Flagging here rather than opening a competing PR since it's squarely in this PR's "tolerant of module-ref spellings" scope. Happy to take it as a separate PR if you'd prefer.

packages/runtime-common/instance-filter-matcher.ts
```diff
@@ -425,26 +433,83 @@ function negate(result: MatchResult): MatchResult {
...
+// the same module compare equal regardless of how they were written.
+function canonicalModuleKey(
+  module: string,
+  virtualNetwork: VirtualNetwork,
+): string {
+  let href: string;
+  try {
+    // Resolves a prefix-form RRI (e.g. `@cardstack/base/card-api`) to its
+    // mapped URL; an absolute URL is returned unchanged.
+    href = virtualNetwork.resolveURL(module, undefined).href;
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Good call — switched to `toURLHref()`. The canonicalization helper moved to `canonicalModuleKey` in `code-ref.ts` (now shared with `isFileDefCodeRef`) and its first step uses the memoized `virtualNetwork.toURLHref(module)` instead of `resolveURL(...).href`, so the type-gate hot path no longer allocates a `URL` per call. Equivalent resolution here since these refs are always absolute URLs or prefix RRIs (no relative/`/`-rooted inputs that would need `resolveURL`'s base handling). Fixed in 93c014ca73.

lukemelia:
[Claude Code 🤖] Thanks — fixed in 93c014ca73, and you nailed shard 15.

I took the second option (make `isFileDefCodeRef` spelling-tolerant) rather than re-pointing `baseModule` at `baseRealmRRI`, precisely to avoid the `resolveFileDefCodeRef` output shift you flagged: that helper feeds the indexer (`realm.ts`, `index-runner/visit-file.ts`, `file-indexer.ts`), so changing its below-VN output is a stored-ref behavior change I didn't want to take on in a "get main green" fix. `isFileDefCodeRef` has a single caller (`resources/search.ts`), so threading the network through is contained.

Concretely: extracted the canonicalize-via-VirtualNetwork logic into a shared `canonicalModuleKey` in `code-ref.ts` and use it from both `codeRefEquals` and `isFileDefCodeRef`, so prefix-form / real-URL / virtual-alias spellings all collapse to one key before comparison. `resolveFileDefCodeRef` is untouched. Verified against a real `VirtualNetwork` that the base FileDef ref and the subtype refs (`MarkdownDef`, `PngDef`) match while `CardDef`, a non-base `Book` ref, `undefined`, and a structured ref correctly don't. The existing `file-def-code-ref-test.ts` (updated to supply a VN configured like host boot) is the regression coverage.

habdelra:
i think that remaining flaky test is actually fixed in this PR https://github.com/cardstack/boxel/pull/5333

PR #5339, opened 2026-06-25
Add archived_at to realm_metadata + archive query helpers

lukemelia:
## What

Adds the storage and query layer for the owner-controlled realm archive flag.

## Changes

- **Migration** (`packages/postgres/migrations/1782417591926_add-archived-at-to-realm-metadata.js`): adds a nullable `archived_at timestamptz` column to `realm_metadata`. `NULL` = active; non-null = the time the realm was archived. Existing and new rows default to `NULL`.
- **Query helpers** (`packages/runtime-common/db-queries/realm-metadata-queries.ts`, exported via `runtime-common/index.ts`):
  - `archiveRealm(dbAdapter, realmURL)` — upserts `archived_at = now()`, so a realm with no metadata row yet can still be archived.
  - `unarchiveRealm(dbAdapter, realmURL)` — upserts `archived_at = NULL`; idempotent when no row exists.
  - `isRealmArchived(dbAdapter, realmURL)` — boolean.
  - `fetchArchivedRealmsForOwner(dbAdapter, username)` — joins archived `realm_metadata` rows to `realm_user_permissions` where the user holds `realm-owner`, returning the owner's archived realm URLs (newest first).

## Why this shape

Archiving is a global, per-realm state rather than per-user, so it belongs on `realm_metadata` — the mutable per-realm flags table — not on a permissions or per-user table. A nullable timestamp captures both the active/archived boolean and when the transition happened in one column.

## Testing

- Migration applies up, rolls back down, and re-applies cleanly; the column is confirmed nullable with no default.
- New `realm archive helpers` module in `packages/realm-server/tests/queries-test.ts` covers set → archived, clear → active, idempotent clear, and the owner-scoped archived list (excludes other owners' realms and realms where the user only has read). The file passes 8/8.

### DIFF
--- a/packages/realm-server/tests/queries-test.ts
+++ b/packages/realm-server/tests/queries-test.ts
@@ -5,9 +5,13 @@ import { v4 as uuidv4 } from 'uuid';
 
 import type { PgAdapter } from '@cardstack/postgres';
 import {
+  archiveRealm,
   fetchAllRealmsWithOwners,
+  fetchArchivedRealmsForOwner,
   fetchUserPermissions,
   insertPermissions,
+  isRealmArchived,
+  unarchiveRealm,
 } from '@cardstack/runtime-common';
 
 import { upsertPublishedRealmInRegistry } from '../lib/realm-registry-writes.ts';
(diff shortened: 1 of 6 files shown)


packages/runtime-common/db-queries/realm-metadata-queries.ts
```diff
@@ -0,0 +1,58 @@
...
+// realm_metadata rows to realm_user_permissions where the user holds the
+// realm-owner permission.
+export async function fetchArchivedRealmsForOwner(
+  dbAdapter: DBAdapter,
+  username: string,
+): Promise<string[]> {
+  let results = (await query(dbAdapter, [
+    `SELECT rm.url
+     FROM realm_metadata rm
+     INNER JOIN realm_user_permissions rup ON rup.realm_url = rm.url
```
lukemelia, line 64: [Codex] This join will also return published snapshots if they get an archived metadata row: `handle-publish-realm` grants the publishing user `realm-owner` on the published URL, and the existing owner-facing permission query filters published `realm_registry` rows out for that reason. Please add the same published-realm exclusion here, plus a test that archives a published URL owned by the user, so the archived workspace list only contains source realms.

packages/runtime-common/db-queries/realm-metadata-queries.ts
```diff
@@ -0,0 +1,58 @@
...
+
+// Mark a realm archived. Upserts the realm_metadata row so a realm that
+// never had a metadata row (no other flags ever set) can still be
+// archived. Re-archiving an already-archived realm refreshes the
+// timestamp.
+export async function archiveRealm(dbAdapter: DBAdapter, realmURL: URL) {
+  await query(dbAdapter, [
+    `INSERT INTO realm_metadata (url, archived_at) VALUES (`,
+    param(realmURL.href),
+    `, now()) ON CONFLICT (url) DO UPDATE SET archived_at = now(), updated_at = now()`,
```
lukemelia: [Codex] This helper lives in runtime-common and accepts a generic `DBAdapter`, but `now()` is Postgres-only. The SQLite adapter will execute this raw SQL string as-is, so `archiveRealm`/`unarchiveRealm` will fail if they are called from a host/SQLite path. Existing shared query helpers use `dbExpression({ pg: "now()", sqlite: "CURRENT_TIMESTAMP" })` for this; can we make these timestamps adapter-specific before exporting this as a runtime-common API?

packages/runtime-common/db-queries/realm-metadata-queries.ts
lukemelia: [Claude Code 🤖] Good catch — confirmed `handle-publish-realm` grants the owner `realm-owner` on the published URL. `fetchArchivedRealmsForOwner` now excludes `realm_registry` rows with `kind = 'published'` (the same filter `fetchUserPermissions(onlyOwnRealms)` uses), and I added a test that archives a published URL owned by the user and asserts it's omitted. Fixed in 6a12bbe7a1.

packages/runtime-common/db-queries/realm-metadata-queries.ts
lukemelia: [Claude Code 🤖] Agreed — switched the timestamps to `dbExpression({ pg: 'now()', sqlite: 'CURRENT_TIMESTAMP' })` in both `archiveRealm` and `unarchiveRealm`, so the shared runtime-common API renders correctly on the SQLite path instead of emitting Postgres-only `now()`. Fixed in 6a12bbe7a1.

backspace (approved):

habdelra (approved):

packages/runtime-common/db-queries/realm-metadata-queries.ts
```diff
@@ -0,0 +1,73 @@
...
+  let results = (await query(dbAdapter, [
+    `SELECT rm.url
+     FROM realm_metadata rm
+     INNER JOIN realm_user_permissions rup ON rup.realm_url = rm.url
+     WHERE rm.archived_at IS NOT NULL
+       AND rup.realm_owner = true
+       AND rm.url NOT IN (SELECT url FROM realm_registry WHERE kind = 'published')
+       AND rup.username =`,
+    param(username),
+    `ORDER BY rm.archived_at DESC`,
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Good point — added a secondary `rm.url ASC` sort key so the order is deterministic when several realms share an `archived_at` second (SQLite's `CURRENT_TIMESTAMP` is 1s-resolution). Fixed in 7c0717bfe7.

packages/realm-server/tests/queries-test.ts
```diff
@@ -221,4 +225,127 @@ module(basename(import.meta.filename), function () {
...
+      const realmURL = new URL('http://example.com/never-archived/');
+
+      await unarchiveRealm(dbAdapter, realmURL);
+      assert.false(
+        await isRealmArchived(dbAdapter, realmURL),
+        'realm remains active',
+      );
+    });
+
+    test('fetchArchivedRealmsForOwner returns only the owner’s archived realms', async function (assert) {
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Fixed — switched the test title to a plain ASCII apostrophe. 7c0717bfe7.

PR #5344, opened 2026-06-26
ai-bot: the assistant can read a skill directly from the realm server

jurgenwerk:
Here is a video demonstrating:

1. Informing the AI about a set of available skills and where to get them if needed. **In next work, we will create curated skill index files and attach it at the beginning of the system prompt**. This is so that AI will be aware of the available skills but won't load them until it actually needs them
2. The bot pulling the skills on demand **from the realm server directly** (but still visible as command results in the chat UI). The reasons we want to fetch from the server directly: no Matrix uploads at enable time, no dependency on the user's host tab being open, content is always fresh
3. AI then receives the skills on its own demand (in a "pull" mechanism, as opposed to our current "push" mechanism where we inject all the attached skills upfront in the prompt)

https://github.com/user-attachments/assets/c232e8e4-1a21-450a-8eb6-a8fb25369b44

More details:

## What this does

The assistant can now grab a file from your realm **by itself, while it's answering** — usually a skill's instructions, but really any file. Before, every enabled skill's full text was stuffed into the prompt up front. Now the bot reads only what it needs, when it needs it.

## How it works (the simple version)

1. Mid-answer, the bot thinks "I need that file" and calls a tool: `readRealmFile(realm, url)`.
2. The bot fetches the file **itself**, using a short-lived, read-only pass made out in *your* name — so it can only read what you already could. No browser, no round-trip; it happens in the same turn.
3. It reads the file's contents into its own working memory, keeps going, and answers you.
4. You see a pill in the chat — **"Read file: demo/SKILL.md"** — that spins while loading and turns to ✓ when done (or red with a reason if it failed). That pill is just a record; the file's *text* goes to the model, not into the chat.

## Why `readRealmFile` and not `loadSkill`

There's nothing skill-specific about it — it just reads a file. So it gets an honest, general name. (It's distinct from the existing host `readFile`, which runs in your browser; this one runs server-side with a delegated token.)

## Mixed tool calls

Bot-run tools (like this) finish *this* turn; host commands finish *next* turn. Those two timings don't mix. So if the model asks for a read **and** a host command in the same round, the read is **explicitly rejected** with a clear message ("readRealmFile cannot be mixed with host-dispatched commands in the same round") instead of being silently dropped — and the host command still runs.

## Safety

- Only in **single-human rooms**, so "you" is unambiguous.
- **Read-only and scoped to you** — can't read what you can't, can't write.
- **Always live** — the current file, never a stale snapshot.
- A missing file / no access / bad URL comes back as an error the model can read (and a red pill), never a crash.
- Inert unless the delegation secret is configured (already set in our deployed envs).

## Also in here (unrelated)

A one-line CSS fix removing a stray border line that appeared above the AI action bar during diff-apply.

## Tests

Unit tests for the fetch/executor (delegated-token retry, 404 / disabled / forbidden), the tool-call classifier (bot vs host, mixed), and the read loop; verified live with single and multiple reads, rendered in order.


jurgenwerk (commented):
[Claude Code 🤖] Self-review — bugs/risks I can substantiate, worst first.

### 1. (High, security) No single-human-room guard before minting delegation
`main.ts` runs the loadSkill loop whenever `delegatedRealmSessions.enabled`, with `onBehalfOf: senderMatrixUserId` (main.ts:597). In a room with more than one human, the bot mints a read token for whoever sent the message and reads *their* realms — the confused-deputy case the security design calls out. It's gated by the secret and the guard is a known follow-up, but **this must land before the secret is enabled anywhere**. Right now nothing stops it.

### 2. (Medium, correctness) A rejected delegated token is never invalidated
`executeLoadSkill` (lib/load-skill.ts) returns an error on a non-2xx file response but never calls `delegatedRealmSessions.invalidate(...)`. Tokens are cached for ~30 min; if the user's read is revoked inside that window, the cached token gets a 401/403 and **every loadSkill for that (user, realm) keeps failing for up to 30 minutes**. Should invalidate on 401/403 and retry once. Note the deps are `Pick<…, 'getToken'>`, so the fix also has to widen them to include `invalidate`.

### 3. (Medium) loadSkill calls can be silently dropped — no result, no error
`response-state` drops loadSkill from the emitted command requests, and the bot only executes it when `buildLoadSkillFollowup` returns a followup. But it returns `[]` when the model mixes loadSkill with a host command, and the loop stops at `LOAD_SKILL_MAX_ROUNDS`. In those cases the loadSkill call is **neither executed nor handed to the host** — the model "called a tool" and gets nothing back, so the turn can end with empty/confused content. At minimum it should feed back an error tool result (or stop advertising the tool once at the cap).

### 4. (Medium, UX — the untested path) Multi-round rendering into one Matrix message
The responder is reused across rounds and `ResponseState.updateContent` *replaces* `latestContent`. Round 1 (the loadSkill round) and round 2 (the answer) stream into the same message, so any visible round-1 content is overwritten and the message can shrink/reset mid-stream. This is the multi-round behavior I haven't exercised live (the single-round happy path was verified). Needs a real >1-round chat before trusting it.

### 5. (Low) `executeLoadSkill` follows redirects
Unlike the manual script (`redirect: 'manual'`), the executor uses default fetch, so a stray 30x is followed silently. A file GET shouldn't redirect, but worth pinning to surface surprises.

### 6. (Low) No size cap on fetched content
`executeLoadSkill` returns `await response.text()` unbounded; a large `references/` file is injected wholesale into the next prompt. Fine for SKILL.md, risky for arbitrary reference files.

### 7. (Low) `generationId` reflects only the last round
Cost is summed across rounds (good), but `generationId` used for credit tracking is the last round's id only — minor attribution skew across a multi-round turn.

Nothing here is a crash on the normal (delegation-off) path — that stays byte-identical. #1 gates enabling the feature; #2/#3 are the substantive correctness fixes; #4 is the thing to verify live.

jurgenwerk:
[Claude Code 🤖] Fixes pushed for the review (numbering matches the review items above; written `(n)` so GitHub doesn't auto-link them as PRs — sorry about the earlier mangled links):

- **(1) single-human-room guard** — counts non-bot members; loadSkill is neither offered nor executed unless exactly one human is in the room, so the bot never reads a realm on an ambiguous user's behalf. (`getResponse` now takes an explicit `offerLoadSkill` flag.)
- **(2) token invalidation** — a cached token rejected with 401/403 is invalidated and re-minted once before failing, instead of being wedged for the rest of its TTL.
- **(5) redirects** — the skill fetch uses `redirect: 'manual'`; **(6) size cap** — fetched content is bounded so a large references file can't blow up the next round's prompt.

New tests cover the retry and the cap; types/lint/responder suite green.

Still open (intentionally not patched here):
- **(3)** — when the model mixes loadSkill with a host command, or exceeds the round cap, the loadSkill call is dropped rather than answered. Closing it cleanly needs a host-vs-bot tool-precedence decision, so I left it as a known limitation.
- **(4)** — multi-round rendering into one Matrix message still needs a real >1-round chat to verify; the single-round path is confirmed.

habdelra (approved):

jurgenwerk:
idea by @lukemelia: when AI requests a tool call, the ai bot can make a command request with a special property named something like `handledByServer` and then the host can know it doesn't have to execute it (because the ai bot already loaded the skill directly from the server). This is so that we can show a visual indicator in the chat history telling the user the necessary skills got loaded (will also be helpful to debug in case of problems)

lukemelia:
> idea by @lukemelia: when AI requests a tool call, the ai bot can make a command request with a special property named something like `handledByServer` and then the host can know it doesn't have to execute it (because the ai bot already loaded the skill directly from the server).

Or, even better might be `executedBy: "ai-bot"`. A little more information dense, and could possibly help us when we scale to multi-user or multi-bot rooms.

lukemelia (changes requested):
Let's re-implement with the approach we discussed before merging.

lukemelia:
One other question to consider: there is nothing specific to skills in the `loadSkill` tool AFAIK. Should it just be a `readFile` tool?

jurgenwerk:
(codex) Review note: I think the current `readRealmFile` path is carrying too much special-case control flow.

The main issue I see is that mixed tool rounds are still lossy. `readRealmFileOnlyToolCalls()` only handles the round when *every* tool call is `readRealmFile`, while `ResponseState` also strips `readRealmFile` out of the host-visible command requests. So if the model emits `readRealmFile` plus a normal host command in the same round, the read is neither executed nor surfaced to the host.

I think the simpler shape is a single post-completion tool-call pass:
- if ai-bot has a local executor for a tool, run it and append a `tool` result
- otherwise emit a normal host command request
- keep `executedBy: "ai-bot"` only as timeline/debug metadata, not as a separate control path

That would remove most of the branching between `read-realm-file-loop.ts` and `main.ts`, and it seems closer to the re-implementation direction already being discussed here.

packages/host/app/lib/matrix-classes/message-builder.ts
```diff
@@ -331,6 +331,29 @@ export default class MessageBuilder {
         );
       }) as CommandResultEvent | undefined);
 
+    // ai-bot ran this one itself (executedBy, e.g. readRealmFile), so the host
+    // never resolves a command class or runs it. Skip the skill lookup below —
+    // it's pure async churn here (an `await store.get` per enabled skill) that
+    // would leave the marker pill blank for a beat while it runs. Build the
+    // command synchronously: 'applying' (loading) until the result event lands,
+    // then applied (success) or invalid + reason (failure).
+    if (commandRequest.executedBy) {
```
jurgenwerk: should this be `commandRequest.executedBy == 'ai-bot'`?

packages/host/app/lib/command-auto-execute.ts
```diff
@@ -18,13 +18,18 @@ export const CHECK_CORRECTNESS_COMMAND_NAME = 'checkCorrectness';
...
+  command: Pick<MessageCommand, 'name' | 'requiresApproval' | 'executedBy'>,
   activeLLMMode: LLMMode | undefined,
   isOwnedByCurrentAgent: boolean,
 ): boolean {
   if (!isOwnedByCurrentAgent) {
     return false;
   }
+  // Already executed by a server-side actor (e.g. ai-bot ran readRealmFile);
+  // the host only records it in the timeline and never executes it.
+  if (command.executedBy) {
```
jurgenwerk: should this be command.executedBy == 'ai-bot'?

packages/host/app/services/command-service.ts
```diff
@@ -456,6 +463,11 @@ export default class CommandService extends Service {
       message.eventId,
     );
     for (let messageCommand of message.commands) {
+      // Server-executed (e.g. ai-bot's readRealmFile): not the host's to run, so
+      // not the host's to invalidate when processing wedges.
+      if (messageCommand.executedBy) {
```
jurgenwerk: Same comment as elsewhere

packages/host/app/services/command-service.ts
```diff
@@ -660,6 +672,11 @@ export default class CommandService extends Service {
 
   //TODO: Convert to non-EC async method after fixing CS-6987
   run = task(async (command: MessageCommand) => {
+    // Server-executed (e.g. ai-bot's readRealmFile): nothing for the host to run.
+    // Guards the manual "Try Anyway" path as well as any auto-execution.
+    if (command.executedBy) {
```
jurgenwerk: Same comment as elsewhere

packages/host/app/services/command-service.ts
```diff
@@ -773,6 +790,11 @@ export default class CommandService extends Service {
 
   async validate(command: MessageCommand): Promise<boolean> {
     let error: string | undefined;
+    // Server-executed (e.g. ai-bot's readRealmFile): the host has no command class
+    // to resolve, and never runs it, so there is nothing to validate.
+    if (command.executedBy) {
```
jurgenwerk: Same comment as elsewhere

packages/host/app/lib/matrix-classes/message-builder.ts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Good call. `executedBy` could later be `host` (or another actor), and the truthiness check would then make the host skip running its *own* command. Switched to `=== AI_BOT_EXECUTOR` here and at all six sites, updated the field doc, and added a test that a non-ai-bot executor isn't short-circuited. (9814e66408)

packages/host/app/lib/command-auto-execute.ts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Yep — switched to `=== AI_BOT_EXECUTOR` (truthiness would wrongly skip a `host`-executed command). Done at all six sites + a test. (9814e66408)

packages/host/app/services/command-service.ts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Done — same fix: now `=== AI_BOT_EXECUTOR`. (9814e66408)

packages/host/app/services/command-service.ts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Done — same fix: now `=== AI_BOT_EXECUTOR`. (9814e66408)

packages/host/app/services/command-service.ts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Done — same fix: now `=== AI_BOT_EXECUTOR`. (9814e66408)

jurgenwerk:
@lukemelia the implementation should now be more in line with what we discussed - I also attached a video in the PR description

jurgenwerk:
@lukemelia regarding the 2 different flows, it's because of an ordering problem

The ordering problem was where bot does the read and the answer in one turn, and that turn has a single "Thinking…" bubble that the answer streams into. When the bot also tried to show a "Read file…" command request/result for the read, that landed below the answer, which looked weird and out of place

Host commands get correct ordering for free: the bot ends its turn, the host runs the command and posts the command result, and that starts a new turn = a new message. So the request, result, and continuation naturally land in order.

On the other hand, readRealmFile stays in one turn: the bot runs it itself and keeps writing the answer in the same turn. So we replace the "Thinking…" message with the command request(s), post the command result(s), and push the answer into a new message after manually arranging the order

lukemelia:
> The ordering problem was where bot does the read and the answer in one turn

I'm going to try creating a patch to this PR we can review that uses a multi-turn approach, keeping host-executed commands and aibot-executed commands more similar to each other and how they are treated my ai-bot's main processing engine. I think it could simpilify things a lot.

jurgenwerk:
I noticed another a real problem in this implementation - the skill content is not persisted, and it's only available for the current turn.

jurgenwerk:
Working now:

image

lukemelia (approved):
Let's re-implement with the approach we discussed before merging.

PR #5348, opened 2026-06-26
Enforce data-test-* selector lint rules across root-config packages

burieberry:
## Summary

`data-test-*` attributes are stripped from production builds by `ember-test-selectors`, so using them as CSS/DOM selectors (e.g. `querySelector('[data-test-foo]')` or `[data-test-foo] { ... }` in a `<style>` block) works in tests but silently breaks everywhere else. This PR closes the gaps where that guard wasn't being applied.

## Changes

- **Shared selector definition.** Extracted the `no-restricted-syntax` selectors into `eslint/data-test-selectors.cjs` and re-export them from the root `.eslintrc.js`. Packages with their own `root: true` config don't inherit the monorepo root, so they import and re-declare the guard from this shared module instead of duplicating the selectors.
- **Applied the ESLint guard** to source/app code in `packages/host`, `packages/boxel-ui/addon`, and `packages/boxel-ui/test-app`, and to `contents/` card code in `packages/catalog` — each scoped so tests, which legitimately select on `data-test-*`, are exempt.
- **Template lint coverage.** The `no-data-test-selector` template rule now also scans `<style>` blocks (CSS lives in a raw `TextNode`, so attribute selectors there were previously missed), and the plugin's recommended config now exempts test templates (`tests/`, `*-test.*`, `*.test.*`).
- **Wired up `software-factory` template linting.** Added a `.template-lintrc.js`, the `ember-template-lint` dependency, and `lint:hbs` / `lint:hbs:fix` scripts so its `realm/` `.gts` files are linted. Fixed the two violations this surfaced: a malformed `template-lint-disable` comment in `document.gts` and `this.args.model` → `@model` usage in `test-results.gts`.

## Verification

- `ember-template-lint realm` lints all 20 files (18 `.gts` + 2 `.test.gts`) with zero errors.
- No remaining non-test `[data-test-` selector usages in the newly-covered source/content directories.


packages/base/default-templates/card-info.gts
```diff
@@ -200,7 +201,8 @@ class CardInfoEditor extends GlimmerComponent<EditSignature> {
               {{on 'click' this.toggleThumbnailEditor}}
               data-test-toggle-thumbnail-editor
             >
-              Change Thumbnail
+              Change
+              {{#unless @hideThemeChooser}}Theme & {{/unless}}Thumbnail
```
backspace: This seems unrelated to the lint rule, is it an inclusion from elsewhere?

Image

burieberry: Yes sorry it’s unrelated. Noticed this regression while fixing the lint error on that file

backspace (approved):

PR #5351, opened 2026-06-27
Build searchable-driven search-doc generation + parity harness + baseline

habdelra:
Builds the searchable-driven search-doc generator and the parity tooling, **without making it authoritative** — store-driven `searchDoc` stays in charge of production indexing until the cutover (CS-11724). Stacked on #5350 (CS-11721).

### Generator — `searchDocFromFields(instance)` (`packages/base/card-api.gts`)
Parallel to `searchDoc`. Derives link depth from the explicit `searchable` annotation instead of from what the render loaded, and loads the named link targets itself (targeted loading) rather than relying on store residency.

- **Routes** are dotted paths rooted at the indexed card's link fields. Depth is governed **entirely by the annotations on the card being indexed** — a card pulled in as a link target does **not** re-consult its own `searchable`. `true` makes the immediate ("self") link searchable; a dotted path makes a deeper (n+1) link searchable; arrays combine.
- The owner's store is threaded through the recursion (a contained `FieldDef` value may not be store-associated, but a link reached through it must still load against the owner's store). Targeted loading degrades a missing/broken/unloadable target to `{ id }`.
- Cycle clipping, `{ id }` for unfollowed/broken/not-loaded links, `linksToMany` id normalization, and the query-backed-field skip are preserved from the store-driven path. Link targets enumerate their **declared** type, dropping unqueryable polymorphic-subtype bloat.

### Parity
- **Differential test** (`packages/host/tests/integration/searchable-search-doc-test.gts`): the searchable-driven path follows a searchable link to the same target, with the same contained data, the store-driven render loaded. Whole-doc byte-equality is intentionally **not** asserted here — the new spec keeps `{ id }` for every relationship while the store-driven path omits unused links via `usedLinksToFieldsOnly`; reconciling that to an identical doc is the cutover's gate, after the migration (CS-11723) reproduces today's depth.
- **`searchable-parity-diff.ts`** (`packages/realm-server/scripts`): the realm-scale post-migration validator — diffs a realm's live store-driven search docs (`boxel_index`) against the searchable-driven output per card, ignoring `_cardType` and (optionally) the intended shallow-link difference. Data-gathering documented inline.
- **Baseline**: `ctse/military-pigeon` render-vs-searchDoc split captured from staging (≈ 933:1) and recorded on the ticket for the follow-on (prerender off the hot path) to measure against.

### Tests (host integration, run in CI)
`searchable-search-doc-test.gts` — 9 cases: the **routes-come-only-from-the-indexed-card** trio (target's own `searchable` dormant when pulled in; honored only when itself indexed; a dotted route on the indexer drives the deeper expansion), self link, n+1 route, contains-routing through a contained value, cycle clip → `{ id }`, missing/broken link → `{ id }`, unannotated link → `{ id }`, declared-type enumeration drops subtype bloat, and the differential expansion-parity check.

Verified: glint + `tsc` clean (base/host/runtime-common/realm-server), eslint clean, and all 9 integration tests green on a live `test-services:host` stack.


packages/base/card-api.gts
```diff
@@ -4546,6 +4546,343 @@ export function searchDoc<CardT extends BaseDefConstructor>(
...
+): Promise<CardDef | undefined> {
+  let resident = store.getCard(reference);
+  if (resident && (resident as any)[isSavedInstance] === true) {
+    return resident;
+  }
+  // A missing / broken / unloadable target degrades to `{ id }` upstream. The
+  // store may surface that either as a returned `CardError` or a thrown
+  // rejection (e.g. a 404 / invalid-URL on the load path), so guard both.
+  try {
+    let cardDoc = await store.loadCardDocument(reference);
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Real bug — fixed in 55e7e85c30. A not-loaded link's reference is now resolved against the owner's `relativeTo` (via the same `makeAbsoluteURL` the doc already uses) **before** the targeted lookup/load, so a relative `links.self` like `./hassan` resolves and the searchable link expands instead of degrading to `{ id }`. Applied to both the `linksTo` and `linksToMany` paths.

packages/realm-server/scripts/searchable-parity-diff.ts
```diff
@@ -0,0 +1,151 @@
...
+
+// A relationship slot captured as `{ id }`-only (or null) carries no contained
+// data — it's a bare reference. The store-driven path omits unused links
+// entirely while the new path keeps the `{ id }`; under --ignore-shallow-links
+// both forms are treated as "no data here".
+function isShallowLink(value: unknown): boolean {
+  if (value == null) return true;
+  if (typeof value !== 'object') return false;
+  let keys = Object.keys(value as object);
+  return keys.length === 1 && keys[0] === 'id';
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch — fixed in 55e7e85c30. `isShallowLink` now recurses through arrays: a plural slot is shallow when every element is shallow (and an empty plural is shallow), so `--ignore-shallow-links` covers an unrendered `linksToMany` (`[{ id }]` vs absent) the same way it covers a singular link.

packages/base/card-api.gts
```diff
@@ -4546,6 +4546,350 @@ export function searchDoc<CardT extends BaseDefConstructor>(
...
+          await searchableQueryableValue(
+            field!.card,
+            rawValue,
+            tails,
+            nextStack,
+            store,
+          ),
+        ]);
+        break;
+      }
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] This is a deliberate, spec-sanctioned choice rather than a bug. The project designates contained polymorphism as an implementation-ease choice (declared-vs-runtime contained fields are both acceptable for a directly-contained polymorphic FieldDef), so the generator enumerates the declared type consistently. I tried switching contained enumeration to the runtime value — it regresses the link case, because `getFields(instance)` resolves a `linksTo` field's `.card` to the *runtime subtype*, which then defeats the intended declared-type drop for link targets (penName leaked into an expanded `linksTo(SimpleAuthor)` of a FancyAuthor). Consistent declared-type enumeration avoids that and keeps link-bloat dropping correct; any byte-parity on polymorphic contains is reconciled at the cutover (CS-11724).

packages/base/card-api.gts
```diff
@@ -4546,6 +4546,350 @@ export function searchDoc<CardT extends BaseDefConstructor>(
...
+          if (item == null) {
+            continue;
+          }
+          let v = await searchableQueryableValue(
+            field!.card,
+            item,
+            tails,
+            nextStack,
+            store,
+          );
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Same as the `contains` thread: declared-type enumeration is deliberate (the project treats contained polymorphism as an implementation-ease choice), and enumerating the runtime item resolves the link `field.card` to the runtime subtype, breaking the intended declared-type drop for `linksToMany` targets. Kept consistent declared-type.

packages/realm-server/scripts/searchable-parity-diff.ts
```diff
@@ -0,0 +1,155 @@
...
+  let keys = new Set([...Object.keys(l), ...Object.keys(g)]);
+  for (let key of [...keys].sort()) {
+    let lv = (l as SearchDoc)[key];
+    let gv = (g as SearchDoc)[key];
+    if (ignoreShallowLinks && isShallowLink(lv) && isShallowLink(gv)) {
+      continue;
+    }
+    let ls = JSON.stringify(lv);
+    let gs = JSON.stringify(gv);
+    if (ls !== gs) {
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in 29a3b69bea — `diffDoc` now serializes with keys sorted at every level (array order preserved) before comparing, so jsonb-vs-JS key-order differences no longer show as false divergences.

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+  test('a link with no searchable annotation stays { id }', async function (assert) {
+    let doc = await loadAndGenerate(`${testRealmURL}ArticleShallow/sh1`);
+    assert.deepEqual(
+      doc.author,
+      { id: authorUrl },
+      'an unannotated link is captured as { id } only',
+    );
+  });
+
+  test('a self-referential link clips the cycle to { id }', async function (assert) {
```
habdelra: lets see a cycle tests that also has 3 cards that participate in a cycle

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+import { setupRenderingTest } from '../helpers/setup';
+
+let loader: Loader;
+let realm: Realm;
+
+// Exercises the searchable-driven generator `searchDocFromFields` (CS-11722).
+// The central property under test (Hassan-confirmed): search-doc depth is
+// sourced ONLY from the `searchable` annotations on the card being indexed —
+// a card pulled in as a link target does NOT re-consult its own `searchable`.
+module('Integration | searchable search doc', function (hooks) {
```
habdelra, line 43: I don't see any linksToMany tests, each tests should be paired with ah linksToMany form of the test

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+  test('a self-referential link clips the cycle to { id }', async function (assert) {
+    let doc = await loadAndGenerate(`${testRealmURL}Person/p1`);
+    assert.deepEqual(
+      doc.friend,
+      { id: `${testRealmURL}Person/p1` },
+      'a self link clips to { id } via the cycle guard',
+    );
+  });
+
+  test('a searchable link to a missing target degrades to { id }', async function (assert) {
```
habdelra, line 880: what does a linksToMany form of this look like when you are missing a specific link in the collection? what about missing all the links in a collection?

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: what does this look like when you have a linksToMany that participates in a cycle?

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+    class Person extends CardDef {
+      static displayName = 'Person';
+      @field name = contains(StringField);
+      @field friend = linksTo(() => Person, { searchable: true });
+    }
+    // A FieldDef that itself holds a link, so a route can pass THROUGH a
+    // contained value to reach a deeper link (contains-routing).
+    class ArticleMeta extends FieldDef {
+      static displayName = 'ArticleMeta';
+      @field editor = linksTo(Author);
```
habdelra: we need to have a linksToMany on this and tests that explore that too

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+  });
+
+  setupLocalIndexing(hooks);
+  let mockMatrixUtils = setupMockMatrix(hooks);
+  setupCardLogs(
+    hooks,
+    async () => await loader.import(`${baseRealm.url}card-api`),
+  );
+
+  hooks.beforeEach(async function () {
```
habdelra, line 58: I'm not seeing any tests that test thru the "." in in the dotted fields, where is a test like "foo.bar", or "foo.bar.baz". we should assert that we see contained fields in the traversed links

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+    class Pair extends CardDef {
+      static displayName = 'Pair';
+      @field name = contains(StringField);
+      @field left = linksTo(Agent);
+      @field right = linksTo(Agent);
+    }
+    class ArticleArray extends CardDef {
+      static displayName = 'ArticleArray';
+      @field title = contains(StringField);
+      @field pair = linksTo(Pair, { searchable: ['left', 'right'] });
```
habdelra: how about an array where there is a shared ancestor like ['foo.bar', 'foo.baz'], does these collapse correctly?

packages/realm-server/scripts/searchable-parity-diff.ts
habdelra, line 1: where are the tests for this!?!?

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+    let doc = await loadAndGenerate(`${testRealmURL}ArticleDeep/d1`);
+    assert.strictEqual(
+      doc.author?.agent?.name,
+      'Agent Smith',
+      'the route `author.agent` declared on the indexed card drives the depth',
+    );
+  });
+
+  test('a link with no searchable annotation stays { id }', async function (assert) {
+    let doc = await loadAndGenerate(`${testRealmURL}ArticleShallow/sh1`);
```
habdelra, line 738: we should also assert that contained fields are always included in search doc too

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+    assert.strictEqual(
+      doc.pair?.right?.name,
+      'Agent Smith',
+      'the second route in the array (`right`) expands',
+    );
+  });
+
+  test('a route through a contained field reaches a deeper link', async function (assert) {
+    let doc = await loadAndGenerate(`${testRealmURL}ArticleContains/c1`);
+    assert.strictEqual(
```
habdelra, line 989: i feel like we are really light on the assertiuons here--there should be contained fields on the fieldDef you are linking thru that you assert AND contained fields in the actual link that you should be asserting (as well as the link itself)

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -0,0 +1,538 @@
...
+    // A FieldDef that itself holds a link, so a route can pass THROUGH a
+    // contained value to reach a deeper link (contains-routing).
+    class ArticleMeta extends FieldDef {
+      static displayName = 'ArticleMeta';
+      @field editor = linksTo(Author);
+    }
+    class ArticleContains extends CardDef {
+      static displayName = 'ArticleContains';
+      @field title = contains(StringField);
+      @field meta = contains(ArticleMeta, { searchable: 'editor' });
```
habdelra: what about containsMany for Field def. consider there are 4 possible combinations here: contains including a linksTo field, contains including linksToMany field, containsMany including linksTo field containsMany including linksToMany field

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra, line 43: what about tests for bad paths? empty string, null, undefined, empty array, array with null or undefined. do we have a test for a path that is impossible to exist a field that is not actually declared? or perhaps an impossible dotted  path too like 'missing.foo', both for linksTo and linksToMany

habdelra: what happens when searchable is `false`?

packages/realm-server/scripts/searchable-parity-diff.ts
```diff
@@ -0,0 +1,177 @@
...
+
+// A relationship slot captured as `{ id }`-only (or null) carries no contained
+// data — it's a bare reference. The store-driven path omits unused links
+// entirely while the new path keeps the `{ id }`; under --ignore-shallow-links
+// both forms are treated as "no data here".
+// Order-insensitive serialization for comparison: object keys are sorted at
+// every level (array order is preserved). Postgres `jsonb` and JS object
+// construction can emit the same data with different key order, so a plain
+// `JSON.stringify` comparison would report noisy false divergences.
+function stableStringify(value: unknown): string {
```
habdelra: there should be a npm package that does stable json stringifies that we are using, no need to reinvent this

packages/realm-server/scripts/searchable-parity-diff.ts
```diff
@@ -0,0 +1,177 @@
...
+): string[] {
+  let diffs: string[] = [];
+  let strip = (d: SearchDoc) => {
+    let { _cardType, ...rest } = d;
+    return rest;
+  };
+  let l = strip(live ?? {});
+  let g = strip(generated ?? {});
+  let keys = new Set([...Object.keys(l), ...Object.keys(g)]);
+  for (let key of [...keys].sort()) {
```
habdelra: I would also check that same package for a way to iterate thru the doc in a stable manner too

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added a 3-card cycle: `Ring` (r1→r2→r3→r1) driven by `searchable: 'next.next'`. The test walks all three hops then asserts the 4th re-entry clips to `{ id }` via the cycle guard.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added linksToMany coverage throughout: `Team` (expands every element; the pulled-in member's own `searchable` stays dormant), `TeamDeep` (deep route into each element), `TeamShallow` (unannotated → array of `{ id }`), plus broken/missing/empty cases and a plural cycle (`RingM`).

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added both: `Team/missone` (one real member + one missing → expanded member then `{ id }` for the missing slot) and `Team/missall` (all missing → array of `{ id }`). `Team/empty` covers the empty-collection → `null` case.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added `RingM` — a 3-card linksToMany cycle (`searchable: 'nexts.nexts'`); the test asserts it walks the ring then clips on re-entry to `[{ id }]`.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Done — the contained FieldDef (`Crew`) now carries both a `linksTo` (`lead`) and a `linksToMany` (`roster`), with tests routing through it via both `contains` and `containsMany`.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added `ArticleHop3` with the 3-segment route `author.company.hq` (the a.b.c case) — asserts the third hop expands while a sibling link stays `{ id }`. Contained-scalar assertions added throughout the contains/containsMany tests.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added `ArticleShared` with `['company.ceo', 'company.hq']`: the shared `company` ancestor is expanded once and both deeper links resolve under it.

packages/realm-server/scripts/searchable-parity-diff.ts
habdelra: [Claude Code 🤖] Added `packages/realm-server/tests/searchable-parity-diff-test.ts` — unit coverage for the differ (`diffDoc`/`isShallowLink`/`shallowIds`): `_cardType` ignored, key-order insensitivity, changed-id reported under `--ignore-shallow-links`, and absent-vs-shallow handling.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added — every contains/containsMany test now asserts the contained scalar is present, plus a dedicated "contained scalar fields are always present regardless of searchable" test.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Strengthened — the `Crew` tests now assert the FieldDef's own contained field (`label`) AND the linked target's contained field (`name`), on top of the link expansion itself.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added all four combinations via the `Crew` matrix: contains→linksTo, contains→linksToMany, containsMany→linksTo, containsMany→linksToMany.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: [Claude Code 🤖] Added bad-path coverage: empty string, empty array, `null`, an array with a non-string entry, and an impossible deep path (`agent.deeper`). Both the generator's route seeding and the definition-build validator were hardened to degrade gracefully on these instead of throwing.

habdelra: [Claude Code 🤖] `searchable` now accepts `false` (type widened to `boolean | string | string[]`) as an explicit "not searchable" — same result as omitting it (`{ id }`). Test added, and the validator handles it too.

packages/realm-server/scripts/searchable-parity-diff.ts
habdelra: [Claude Code 🤖] Switched to `safe-stable-stringify` (already a dependency in runtime-common and host) and dropped the hand-rolled version.

packages/realm-server/scripts/searchable-parity-diff.ts
habdelra: [Claude Code 🤖] No stable iteration needed — the comparison unions both sides' keys into a `Set` and compares each independently, so iteration order can't change the result. I removed the key sort entirely; `safe-stable-stringify` handles the value serialization.

packages/host/tests/integration/searchable-search-doc-test.gts
habdelra: NO!! ther eis no such thing as not searchable!!! WTF?! this is not in the requirements. i just wanted false as a test to see how we handle band input

habdelra: [Claude Code 🤖] You're right — I over-reached. Reverted: `Searchable` is back to `true | string | string[]` (no `false`, no "not searchable" concept). `false` is now treated purely as bad input — both the route seeding and the definition-build validator degrade it to `{ id }` without throwing, and the test is reframed as a bad-input case (`searchable: false as any`).

Separately, removing `false` surfaced that the generator had been wrongly pulled into every card's dependency closure via a card-api re-export; that re-export is removed too, so `searchable` is no longer a universal card dep.

lukemelia (approved):

PR #5367, opened 2026-06-30
Persist applied command status across reload (CS-11736)

FadhlanR:
## Problem

After a command is applied in the AI assistant (most visibly an auto-executed command in `act` mode), it shows `applied` during the session but reverts to `ready` after a page reload — the applied state isn't persisted.

**Root cause.** `command-service.run` stamps the `commandResult` event's `m.relates_to.event_id` with the *latest* `m.replace` edit id `Y` of the streamed bot message (intentional — CS-11045 — so the ai-bot's normalized `/messages` view correlates correctly). On reload the timeline filter strips all `m.replace` edits, so only the original event `X` is loaded (with aggregated content) and `Y` no longer exists as a distinct event. The host's two reload-side matchers still keyed off `event_id`, so the dangling `Y` link broke and the status fell back to the literal default `'ready'`.

## Fix

Correlate both reload-side matchers by the stable, globally-unique `commandRequestId` (the LLM tool-call id) instead of the drifting `event_id` — mirroring the correlation the ai-bot already uses:

- `MessageBuilder.buildMessageCommand` — drop the `event_id` condition; match on type + rel_type + `commandRequestId`.
- `RoomResource.updateMessageCommandResult` — locate the owning bot message by "result's `commandRequestId` ∈ message's command-request ids" instead of `event_id` equality.

The wire format (`event_id = Y`) is **unchanged**, so ai-bot prompt construction / checkCorrectness behavior is untouched. The downstream `_messageCache` key still resolves to `X` on both the live and reload paths.

## Screen Recording

https://github.com/user-attachments/assets/bd35c493-cb13-4e32-9c02-ec22494fcc88

## Test

Adds a host regression test that reproduces the post-reload state directly (original message `X` carrying the command request + a `commandResult` linked to a stripped edit id `Y`) and asserts the command still renders `[data-test-apply-state="applied"]`. Fails on `main`, passes with this change.

> Note: the host integration suite couldn't be run locally (no Docker for Postgres/Synapse in this environment); relying on CI. Type-check (`ember-tsc`) and eslint pass on the changed files.

Linear: CS-11736

### DIFF
--- a/packages/host/app/lib/matrix-classes/message-builder.ts
+++ b/packages/host/app/lib/matrix-classes/message-builder.ts
@@ -322,11 +322,13 @@ export default class MessageBuilder {
       this.builderContext.commandResultEvent ??
       (this.builderContext.events.find((e: any) => {
         let r = e.content['m.relates_to'];
-        // CS-11736: correlate purely by commandRequestId (the globally-unique
-        // LLM tool-call id), not the result's m.relates_to.event_id. After a
-        // reload the m.replace edits are stripped, so the result's link id (the
-        // final edit Y) no longer matches the loaded original event X — but
-        // commandRequestId is stable across edits and survives reload.
+        // Correlate the result to its command by commandRequestId (the
+        // globally unique LLM tool-call id), not by the result's
+        // m.relates_to.event_id. A reload strips the m.replace edits and loads
+        // only the original event, so the result's link id — pointing at the
+        // final edit — matches no loaded event. commandRequestId is stable
+        // across edits and present on every one, so it resolves the command on
+        // both the live and reload paths.
         return (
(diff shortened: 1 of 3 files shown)


lukemelia (approved):
Need the /evergreen-comments skill applied

FadhlanR:
[Claude Code 🤖] Applied `/evergreen-comments`. Stripped the ticket-ID prefixes and change-narrative ("used to key off event_id", "the fix keys", "Without/with Phase B") from the correlation and cache-key comments and the two new test names, restating each as the current contract. Also cleaned the neighboring streaming-snapshot test (comment, name, and `cs-…` fixture id) since it had the same rot. Behavior is unchanged — comments, test names, and fixture strings only. Pushed as f116797.

PR #5374, opened 2026-06-30
Error at query time when a filter path crosses a non-searchable relationship

habdelra:
A filter whose path resolves in the schema but crosses a `linksTo`/`linksToMany` hop that was never made `searchable` matches nothing silently — that hop's target is only a bare `{ id }` in the search doc, so the data the filter wants simply isn't there. The query compiler now throws a distinct, actionable error instead of returning empty.

### How it works

As `walkFilterFieldPath` crosses each `linksTo`/`linksToMany` hop in a filter path, it checks that the hop's target is in the search doc — reconstructing searchability from the **same route model the search-doc generator uses** (`base/searchable.ts`): routes are seeded from the queried card's own `searchable` annotations and narrowed to the target's tails on each hop, and a link is searchable exactly when a route reaches it. The pure route helpers (`routesForField`, `matchSearchableRoutes`) are extracted into `runtime-common/searchable-routes.ts` and shared by both the generator and the check, so "this link is expanded into the doc" and "this link is queryable" can't drift apart.

Filter handlers seed the routes from the queried card and pass them into the walk; sort paths pass none, so only filter paths are gated. The thrown `FilterRefersToNonsearchableFieldError` is distinct from the existing "nonexistent field" error and names the offending relationship plus the `searchable` annotation to add.

### What stays allowed / unchanged

- **`<link>.id`** reference filters (e.g. `customer.id`) — the `{ id }` sentinel is always present, so these are allowed even when the link isn't searchable.
- **Genuinely nonexistent fields** still raise the original "nonexistent field" error: `getField` resolves each segment before that segment's hop is checked, so a typo surfaces as the resolution error.
- **Sort paths** are not gated — only filters.
- **Query-backed relationships** report a separate reason (`query-backed`); they're never in the search doc at all and can't be filtered through.

This lands after the migration + cutover (CS-11723, CS-11724), so every filter path that resolves today is already annotated — the only paths it rejects are genuinely unsearchable ones.

### Tests

`index-query-engine-test.ts` gains a fixture with searchable / non-searchable / query-backed links and covers: unsearchable path → new error; deeper-hop error; query-backed error; searchable path passes; `.id` allowed through a non-searchable link; nonexistent field beyond a searchable link → old error. The existing test card's queried-through relationships are annotated `searchable` so the pre-existing through-link query tests keep passing.

Verified green: host `Unit | query` (76), host `Integration | search resource` (40), realm-server `eq-containment`/`matches-filter` (23, Postgres), `skip-query-backed`/`search-entries`/`superseded` (32), and `indexing-test` (61).

Closes CS-11726.

### DIFF
--- a/packages/runtime-common/index-query-engine.ts
+++ b/packages/runtime-common/index-query-engine.ts
@@ -59,9 +59,11 @@ import {
 } from './index-structure.ts';
 import {
   getFieldDef,
+  getImmediateFieldDef,
   type Definition,
   type FieldDefinition,
 } from './definitions.ts';
+import { matchSearchableRoutes, routesForField } from './searchable-routes.ts';
 import {
   isFilterRefersToNonexistentTypeError,
   type DefinitionLookup,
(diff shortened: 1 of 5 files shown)


packages/runtime-common/index-query-engine.ts
```diff
@@ -1370,6 +1444,11 @@ export class IndexQueryEngine {
   private async handleFieldArity(fieldArity: FieldArity): Promise<Expression> {
     let { path, value, type, pluralValue, usePluralContainer } = fieldArity;
     let definition = await this.getDefinition(type);
+    // Every field-path filter (eq / in / contains / range, including their null
+    // forms) routes through exactly one `fieldArity` node, so this is the single
+    // chokepoint that sees each filter path once. Sort paths use `fieldQuery`
+    // directly and are intentionally not gated here.
+    await this.assertFilterPathSearchable(definition, path);
```
habdelra: this is a really awkward place to do this. this function is specificlaly around determining the cardinality of the field. it lowers cohesion to do path validation at this point. its truly a coincidence that this happens here, and a changing iomplmentation might mean that this is no longer an ideal place for this. in teh requirements we indicate that walkFilterFieldPath fi where we should be validating the path--it makes more sense there because you are walking the path spcifically. please refactor this asertion so that it is not buried in this particular handler

habdelra: [Claude Code 🤖] Agreed — moved the assertion into `walkFilterFieldPath` (884c2c5). It now takes the card's `searchable` routes and validates each `linksTo`/`linksToMany` hop as the walk crosses it, narrowing the routes to the target's tails on the way down. The filter handlers seed the routes from the queried card; sort paths pass none, so only filter paths are gated. `handleFieldArity` no longer carries any validation logic — it just provides the routes. Because `getField` resolves each segment before its hop is checked, a nonexistent field still surfaces as the "nonexistent field" error.

packages/runtime-common/searchable-routes.ts
```diff
@@ -0,0 +1,73 @@
+import type { Searchable } from 'https://cardstack.com/base/card-api';
+
+// Shared route logic for `searchable`-driven search-doc depth. A route is a
+// dotted field path rooted at a card's own link fields: `searchable: true`
+// names the immediate ("self") link (a head-only route), a dotted path names
+// the n+1 link reached through this field's target, and an array combines
+// routes. Depth is governed ENTIRELY by the routes seeded from the card being
+// indexed — a card pulled in as a link target never re-consults its own
+// `searchable`; only the inherited route tails continue into it.
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch — tightened the header in 1c53945. It now states `searchable` governs links only, that a route can root at any field, and that on a `contains`/`containsMany` field `true` is inert (value always present) while a path is meaningful only to reach a link *through* the contained value — matching the canonical `Searchable` docs in `base/card-api.gts`.

lukemelia (approved):

PR #5383, opened 2026-07-02
Fix markdown embed toolbar not swapping to Edit pencil for later refs

FadhlanR:
## Summary

The markdown editor toolbar swaps its `+` ("Add embed") button for a pencil ("Edit embed") when the caret sits inside a `:card[…]` / `::card[…]` (or `:file` / `::file`) reference. In a document with several embeds, the first embed worked but the caret inside later embeds kept showing `+`.

## Root cause

`extractBfmRefRanges()` skips any directive that falls inside a code region, and code regions came from `INLINE_CODE_RE`. That regex allowed an inline code span's content to cross a blank line, so a stray lone backtick — e.g. in a bullet like `Keyboard shortcuts (Mod-\`)` — paired lazily, across the blank line and intervening prose, with the first backtick of a later ```` ``` ```` fenced block. The result was a spurious "inline code" region spanning everything in between, so every `:card` / `::card` directive it covered was treated as code and never got a range. No range → `_currentBfmRef` stayed null → the toolbar never swapped to the pencil. The first embed (before the stray backtick) was unaffected.

## Fix

Constrain inline code spans so their content cannot contain a blank line — a paragraph break ends the span, matching CommonMark and the render-side tokenizer. This is a strict tightening: single-line and soft-wrapped inline code still match; only cross-blank-line pairing is removed. The same regex is shared by `extractBfmReferences()`, so both call sites are corrected consistently.

### Before
Screenshot 2026-07-01 at 21 55 02

### After
Screenshot 2026-07-02 at 16 42 02

## Testing

- Unit: new `extractBfmRefRanges` regression asserting an inline and a block ref after a lone backtick both surface while a ref genuinely inside a fence stays skipped. Existing inline/fenced-code tests stay green.
- Integration (`codemirror embed toolbar`): new case placing the caret in the 2nd (inline) and 3rd (block) embeds after a lone backtick, asserting the Edit pencil appears and Add does not, plus a prose position that still shows Add.
- Verified locally: unit module 132/132 pass; the integration module passes against the running dev stack.

https://linear.app/cardstack/issue/CS-11753

### DIFF
--- a/packages/host/tests/integration/components/codemirror-embed-toolbar-test.gts
+++ b/packages/host/tests/integration/components/codemirror-embed-toolbar-test.gts
@@ -297,6 +297,76 @@ module('Integration | codemirror embed toolbar', function (hooks) {
     );
   });
 
+  test('caret inside a directive after a lone backtick still swaps to Edit', async function (assert) {
+    // A stray lone backtick (here in `Mod-\``) must not pair — across a blank
+    // line — with the backtick fence below: that would form a spurious
+    // inline-code region hiding every directive in between, leaving their
+    // toolbar stuck on "+". The caret in each embed must show the Edit pencil.
+    let harness = new ContentHarness();
+    let content = [
+      '- Keyboard shortcuts (Mod-`)',
+      '',
+      `Inline card reference: :card[${mango}]`,
+      '',
+      `::card[${mango}]`,
+      '',
+      '```typescript',
+      'let greeting = "hi";',
+      '```',
+    ].join('\n');
+    harness.content = content;
+    await renderEditorAndModal({ CodeMirrorEditor, harness });
+
+    await waitFor('[data-test-codemirror-editor] .cm-content', {
+      timeout: 5000,
+    });
+
+    let editor = document.querySelector(
+      '[data-test-codemirror-editor] .cm-editor',
+    ) as HTMLElement | null;
+    let view = editor ? cmContext.EditorView.findFromDOM(editor) : null;
+    view?.focus();
+
+    let inlineStart = content.indexOf(`:card[${mango}]`);
+    let blockStart = content.indexOf(`::card[${mango}]`);
+
+    let placeCaret = async (head: number) => {
+      view?.dispatch({ selection: { anchor: head, head } });
+      await settled();
+    };
+
+    // Start of the inline directive.
+    await placeCaret(inlineStart);
+    await waitFor('[data-test-toolbar="edit-embed"]', { timeout: 5000 });
+    assert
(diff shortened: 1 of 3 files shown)


packages/host/tests/integration/components/codemirror-embed-toolbar-test.gts
```diff
@@ -297,6 +297,76 @@ module('Integration | codemirror embed toolbar', function (hooks) {
     );
   });
 
+  test('caret inside a directive after a lone backtick still swaps to Edit', async function (assert) {
```
habdelra, line 304: why do we have commented out tests?

lukemelia (approved):

packages/host/tests/integration/components/codemirror-embed-toolbar-test.gts
FadhlanR: [Claude Code 🤖] These aren't commented-out tests — the four `//` lines under the `test(...)` line were an explanatory comment describing the scenario the (fully active) test exercises. I can see how sitting directly beneath the declaration made it read like a disabled body, so I've moved the comment above the `test(...)` line in bc3168a so the body now starts with real code.

packages/runtime-common/bfm-card-references.ts
```diff
@@ -8,8 +8,15 @@ import type { TokenizerAndRendererExtension } from './marked.mts';
...
 // Match code spans with any number of consecutive backticks as delimiter
-// (e.g. `code`, ``code``, ```code```).
-const INLINE_CODE_RE = new RegExp('(`+)([\\s\\S]*?)\\1', 'g');
+// (e.g. `code`, ``code``, ```code```). Content may span soft line breaks but
+// not a blank line: a paragraph break ends the span (per CommonMark). Without
+// the blank-line guard a stray lone backtick pairs lazily across blank lines
+// with a later fence, producing a spurious code region that swallows real
+// directives.
+const INLINE_CODE_RE = new RegExp(
+  '(`+)((?:(?!\\n[ \\t]*\\n)[\\s\\S])*?)\\1',
```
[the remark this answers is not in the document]
FadhlanR: [Claude Code 🤖] Good catch — fixed in bc3168a. The paragraph-break guard now allows an optional `\r` on each newline (`\r?\n[ \t]*\r?\n`), so a `\r\n\r\n` blank line ends the code span just like `\n\n`. Verified against `extractBfmRefRanges` that on CRLF input the lone backtick no longer pairs with the later fence: both intervening `:card[...]`/`::card[...]` refs surface while the genuinely-fenced ref stays skipped. Added a CRLF regression test alongside the existing LF one.

PR #5384, opened 2026-07-02
Add Re-index action to workspace chooser tile menu (CS-11741)

FadhlanR:
[CS-11741](https://linear.app/cardstack/issue/CS-11741)

## What

Adds a **Re-index** item to the "..." (kebab) menu on each owned workspace tile in the workspace chooser. Previously the only way to re-index a workspace was the admin `_full-reindex` endpoint.

The action calls the existing `RealmService.fullReindex()`, which POSTs to the realm's `_full-reindex` endpoint and wraps the call in the indexing animation (self-restoring on error). No backend changes.

## Details

- **Owner-gated**: the item only appears on tiles the user owns (mirrors the Archive item's `canReindexWorkspace` shape).
- **Disabled while indexing**: gated on the live `realmInfo.isIndexing` flag rather than `task.isRunning`. The `dropTask` resolves the instant the 204 lands (sub-second), while the reindex itself runs much longer — `isIndexing` stays true for the whole pass. Repeat clicks are server-safe regardless (the reindex queue coalesces them), so this guard is purely a UX nicety.
- **Visible feedback**: the tile's `RealmIcon` now passes `@canAnimate={{true}}` (as the card URL bar already does), so the indexing pulse shows for the whole queued→running→done window.
- **Failure handling**: a failed reindex surfaces an auto-dismissing inline error on the tile.

## Tests

New dedicated acceptance module `workspace-chooser-reindex-test.gts` (mirrors the sibling archive test + the `full-reindex-realm` integration test): owner-gating, POST-to-endpoint + tile animation, disabled-while-indexing, and failure surfacing an inline error.

## Verification

- `pnpm lint:js` — clean
- `pnpm lint:types` — no new errors (only the pre-existing baseline `boxel-ui` module-resolution noise)
- Acceptance suite run pending (local dev stack still coming up); will confirm on this PR.

## Screen Recording

https://github.com/user-attachments/assets/78cd821e-7fda-4d6d-b3db-3e2d2631d0f9


packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
```diff
@@ -1306,6 +1352,28 @@ export default class Workspace extends Component<Signature> {
...
+    this.clearReindexError();
+    try {
+      await this.realm.fullReindex(this.args.realmIdentifier);
+    } catch (error: any) {
+      this.reindexError = String(error?.message ?? error);
+      // Auto-dismiss after a few seconds. A plain setTimeout (not an
+      // ember-concurrency timeout) so it does not keep test `settled()` waiting.
+      this.reindexErrorTimer = setTimeout(() => {
+        this.reindexError = undefined;
+      }, 5000);
```
habdelra: do we really want errors to auto dismiss?

FadhlanR: [Claude Code 🤖] Good call — agreed it shouldn't auto-dismiss. Removed the timer entirely. The failure now renders as a persistent banner across the top of the tile and stays until the user closes it with the banner's ✕ button (or it clears automatically when they retry the reindex). This also drops the stale-timer bookkeeping Copilot flagged.

reindex error banner

habdelra (approved):

PR #5385, opened 2026-07-02
Close race window in StoreService.patch() between snapshot and write-back

richardhjtan:
## Summary

There's a catalog listing-create command that creates a listing card, then kicks off about nine background tasks concurrently against that same card to auto-populate it — name, summary, tags, categories, license, examples, and specs get set directly, while a screenshot and thumbnail get attached via `PatchCardInstanceCommand` → `store.patch()`. Fields were silently disappearing after listing creation — a tag or category set by one task would vanish once another task's patch landed later.

The cause is in `StoreService.patch()`: it snapshots the instance, resolves the patch's relationship target (a network fetch), then reapplies the snapshot wholesale — so whatever another task wrote to the same instance during that fetch gets reverted when the stale snapshot writes back.

This reorders the method to resolve the relationship before taking the snapshot instead of after, so the snapshot can't go stale. No behavior change for the non-concurrent case.

## Why `patch()` resolves relationships at all

This isn't required to persist the patch — the server only needs the pointer, not the target's content. It's there because `LinksTo.deserialize()`'s lazy fallback (cache → `included[]` → `not-loaded` sentinel) assumes a live UI template that re-renders once the real value arrives, but most `patch()` callers are Commands — plain code with no equivalent "wait for it." Without the eager load, a patched relationship would come back as an unresolved placeholder for any caller whose target isn't already cached, with no automatic recovery.

This PR is intentionally scoped to the concurrency bug rather than also removing that eager load — `patch()`'s relationships path has other callers (the AI chat edit tools, code-mode field editing, submission-retry logic) that haven't been audited for whether they assume synchronous hydration.

## Test plan

- Added an integration test that gates the relationship-resolution step on a controlled `Deferred`, mutates an unrelated field on the same live instance while `patch()` is in flight, and asserts that field survives.
- Verified both directions locally: reverting just the reorder in `store.ts` fails the test (`actual: "Hassan"`, `expected: "Race Winner"`); with the reorder in place, it passes.

### DIFF
--- a/packages/host/app/services/store.ts
+++ b/packages/host/app/services/store.ts
@@ -967,6 +967,28 @@ export default class StoreService extends Service implements StoreInterface {
     if (opts?.doNotPersist) {
       await this.stopAutoSaving(instance);
     }
+    // Resolve any linked-card relationships first. This can require a
+    // network fetch, and a sibling task elsewhere may mutate other fields on
+    // this same live instance while we wait. Snapshotting the instance below
+    // (for the merge + write-back further down) only after this resolves
+    // keeps that snapshot from going stale and reverting the sibling's write.
+    let linkedCards = await this.loadPatchedInstances(patch, instance.id);
+    for (let [field, value] of Object.entries(linkedCards)) {
+      if (field.includes('.')) {
+        let parts = field.split('.');
+        let leaf = parts.pop();
+        if (!leaf) {
+          throw new Error(`bug: error in field name "${field}"`);
+        }
+        let inner = instance;
+        for (let part of parts) {
+          inner = (inner as any)[part];
+        }
+        (inner as any)[leaf.match(/^\d+$/) ? Number(leaf) : leaf] = value;
+      } else {
+        (instance as any)[field] = value;
+      }
+    }
     let doc = await this.cardService.serializeCard(instance, {
       omitQueryFields: true,
     });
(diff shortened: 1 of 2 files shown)


habdelra:
can you provide more context for the reason of this PR? why do we need to load links when we are doing a document patch? linked fields are not part of the patch payload, these are side loaded data. the document being patched should be able to be patched without loading links at all. (a patch should not touch the `included[]` part of the JSON:API documetn only the field is truly a brand new instance you already have it locally and are supplying a local id.

richardhjtan:
> can you provide more context for the reason of this PR? why do we need to load links when we are doing a document patch? linked fields are not part of the patch payload, these are side loaded data. the document being patched should be able to be patched without loading links at all. (a patch should not touch the `included[]` part of the JSON:API documetn only the field is truly a brand new instance you already have it locally and are supplying a local id.

Sure @habdelra 

There is a catalog listing create command that creates a listing card, then kicks off about nine background tasks concurrently against that same card to auto-populate it — name, summary, tags, categories, license, examples and specs get set directly, while a screenshot and thumbnail get attached via PatchCardInstanceCommand → store.patch(). I was seeing fields silently disappear after listing creation — a tag or category set by one task would vanish once another task's patch landed later. I traced it to store.patch(): it snapshots the instance, resolves the patch's relationship target (a network fetch), then reapplies the snapshot wholesale — so whatever another task wrote to the same instance during that fetch gets reverted when the stale snapshot writes back. That's what this PR fixes: resolving the relationship before taking the snapshot instead of after, so the snapshot can't go stale.

On why patch() resolves the relationship at all — you're right it's not needed to persist the patch, the server only needs the pointer. It's there because LinksTo.deserialize()'s lazy fallback (cache → included[] → not-loaded sentinel) assumes a live UI template that re-renders once the real value arrives, but most patch() callers are Commands — plain code with no equivalent "wait for it." Without the eager load, a patched relationship would come back as an unresolved placeholder for any caller that doesn't already have the target cached, with no automatic recovery.

I intentionally kept this PR scoped to the concurrency bug rather than also removing that eager load — patch()'s relationships path has other callers (the AI chat edit tools, code-mode field editing, submission-retry logic) I haven't audited for whether they assume synchronous hydration

habdelra (approved):

PR #5389, opened 2026-07-02
Isolate per-file indexing failures instead of aborting the whole realm

richardhjtan:
## Origin
Found while debugging the Virtual Piano App in the public catalog: its in-card song search showed "No Music Sheets found" when the app ran from the catalog realm, yet worked after remixing into a personal realm. First root cause: the local catalog realm's `boxel_index` was completely empty — its from-scratch reindex had gotten through 652 of 1166 files, hit one stuck prerender-visit request (a 120s abort on a listing-thumbnail PNG), and the whole job died, discarding all the already-indexed rows. The realm still mounted "successfully," so every search against it silently returned nothing. This PR fixes that all-or-nothing failure mode. (The same investigation surfaced a second, independent bug in how search results are hydrated for alias-prefixed realms, fixed in #5388.)

## Summary
- `IndexRunner.tryToVisit` rethrew any non-404 error from a file's visit — e.g. a transport-level failure of the index visit's own prerender-server request (timeout/abort), or a reader error — which unwound the entire `fromScratch`/`incremental` visit loop and skipped `batch.done()`. (HTML prerendering runs in a separate job; the request at issue is the index pass's own visit.)
- Since `batch.done()` is what promotes the working table into the live `boxel_index`, a single file's transport-level failure discarded every other already-successfully-indexed file in the same job — leaving the realm mounted but with an empty or stale search index, and no error surfaced anywhere.
- The indexer already isolates per-file failures for in-band render errors (an `error_doc` row for just that card), but a transport-level failure — where no response ever arrives — bypassed that model entirely and became all-or-nothing for the whole realm.
- Fix: catch non-404 failures per file and record a `file-error` entry for that URL, mirroring the existing `error_doc` pattern, so the rest of the realm still indexes and commits.
- Follow-up from review (Codex + @habdelra): `Batch.invalidate()` pre-tombstones every existing row type at a URL, so the file-error write alone would let an existing card's untouched `instance` tombstone be promoted — silently deleting a previously-good card from search over a transient error. And deciding "was this a card?" by re-reading the file was itself fragile: if the visit failed on the read path, the re-read fails the same way. The batch now records which live row types `tombstoneEntries` overwrote per URL (`Batch.tombstonedLiveTypes`), and `tryToVisit` writes the companion `instance-error` based on that index-derived answer — the source re-parse remains only as the fallback for brand-new files, which have no prior row to lose.

## Test plan
- [x] Regression test (`indexing-test.ts` › per-file failure isolation): a wrapping `Prerenderer` throws an abort-style transport error for one armed URL; asserts the incremental job still commits (sibling card stays searchable), both the `file` and `instance` rows carry the error without being tombstoned, the instance keeps its last-known-good `pristine_doc`, and a later successful rewrite clears the error state
- [x] Second regression test for the brand-new-card fallback: a new card whose first visit fails still gets both error rows via the source-parse fallback (no prior index row for the oracle to consult)
- [x] Red/green verified: with the fix reverted to main's version, the isolation test fails — the whole incremental-index job rejects with the injected transport error
- [x] Full `indexing-test.ts` run: 63/63 pass
- [x] `tsc --noEmit` clean on changed files
- [x] Branch rebased onto latest main

### DIFF
--- a/packages/runtime-common/index-runner.ts
+++ b/packages/runtime-common/index-runner.ts
@@ -8,6 +8,7 @@ import {
   logger,
   hasCardExtension,
   hasExecutableExtension,
+  isCardResource,
   SupportedMimeType,
   jobIdentity,
   Deferred,
@@ -691,13 +692,43 @@ export class IndexRunner {
             }),
           );
       error.message = message;
-      let entry: FileErrorIndexEntry = {
+      let fileEntry: FileErrorIndexEntry = {
         type: 'file-error',
         error,
       };
-      await this.batch.updateEntry(url, entry);
+      await this.batch.updateEntry(url, fileEntry);
       this.#dependencyResolver.invalidateRelationshipDependencyRowCache(url);
       this.stats.fileErrors++;
+      // `Batch.invalidate()` already tombstoned every type this URL
+      // previously had in the index — for an existing card that's both
+      // `instance` and `file`. Overwriting only the `file` tombstone
+      // above would let batch.done() promote the untouched `instance`
+      // tombstone, silently removing a previously-good card from search
+      // over a transient error. Re-parse the file ourselves (the throw
+      // above lost visitFileForIndexingFused's internal determination)
+      // and write a matching instance-error row when it's a card, so
+      // the last-known-good instance survives the same as the in-band
+      // render-error path.
+      if (url.href.endsWith('.json')) {
+        try {
+          let fileRef = await this.#reader.readFile(url);
+          let resource = fileRef?.content
+            ? (JSON.parse(fileRef.content)?.data as unknown)
+            : undefined;
+          if (resource && isCardResource(resource)) {
+            let instanceEntry: InstanceErrorIndexEntry = {
+              type: 'instance-error',
+              error,
+            };
+            await this.batch.updateEntry(url, instanceEntry);
+            this.stats.instanceErrors++;
+          }
+        } catch (parseErr) {
+          this.#log.warn(
+            `${jobIdentity(this.#jobInfo)} could not determine whether ${url.href} is a card instance after its visit failed: ${(parseErr as Error)?.message}`,
+          );
+        }
+      }
     }
   }
 
--- a/packages/runtime-common/index-runner.ts
+++ b/packages/runtime-common/index-runner.ts
@@ -18,6 +18,7 @@ import {
   type LooseCardResource,
   type InstanceEntry,
   type InstanceErrorIndexEntry,
+  type FileErrorIndexEntry,
   type RealmInfo,
   type FromScratchResult,
   type IncrementalResult,
@@ -35,7 +36,12 @@ import { moduleFrom } from './code-ref.ts';
 import type { RealmResourceIdentifier } from './realm-identifiers.ts';
 import type { CacheScope, DefinitionLookup } from './definition-lookup.ts';
 import type { VirtualNetwork } from './virtual-network.ts';
-import { isCardError } from './error.ts';
+import {
+  CardError,
+  coerceErrorMessage,
+  isCardError,
+  serializableError,
+} from './error.ts';
 import type { IndexingProgressEvent } from './worker.ts';
 import { canonicalURL } from './index-runner/dependency-url.ts';
 import { IndexRunnerDependencyManager } from './index-runner/dependency-resolver.ts';
@@ -658,9 +664,40 @@ export class IndexRunner {
         this.#log.info(
           `${jobIdentity(this.#jobInfo)} tried to visit file ${url.href}, but it no longer exists`,
         );
-      } else {
-        throw err;
+        return;
       }
+      // A transport-level failure (prerender timeout/abort, network error)
+      // never reaches performCardIndexing/performFileIndexing's own
+      // error-entry construction — visitFileForIndexingFused rethrows
+      // before calling indexCardWithResult/indexFileWithResults. Left
+      // uncaught here, one file's failure propagates out of the
+      // fromScratch/incremental visit loop, skips batch.done(), and
+      // discards every other successfully-visited file's rows for the
+      // whole job. Persist a file-error row instead so the failure is
+      // isolated to this URL, matching the error_doc pattern used for
+      // in-band render errors.
+      let message = coerceErrorMessage(
+        err,
+        `Indexing failed for ${url.href} with no error message (${jobIdentity(this.#jobInfo)})`,
+      );
+      this.#log.warn(
+        `${jobIdentity(this.#jobInfo)} failed to index ${url.href}, recording file-error: ${message}`,
+      );
+      let error = isCardError(err)
+        ? serializableError(err)
+        : serializableError(
+            Object.assign(new CardError(message, { status: 500 }), {
+              stack: (err as Error)?.stack,
+            }),
+          );
+      error.message = message;
+      let entry: FileErrorIndexEntry = {
+        type: 'file-error',
+        error,
+      };
+      await this.batch.updateEntry(url, entry);
+      this.#dependencyResolver.invalidateRelationshipDependencyRowCache(url);
+      this.stats.fileErrors++;
     }
   }
 
--- a/packages/runtime-common/index-runner.ts
+++ b/packages/runtime-common/index-runner.ts
@@ -667,16 +667,18 @@ export class IndexRunner {
         );
         return;
       }
-      // A transport-level failure (prerender timeout/abort, network error)
-      // never reaches performCardIndexing/performFileIndexing's own
-      // error-entry construction — visitFileForIndexing rethrows
-      // before calling indexCardWithResult/indexFileWithResults. Left
-      // uncaught here, one file's failure propagates out of the
-      // fromScratch/incremental visit loop, skips batch.done(), and
-      // discards every other successfully-visited file's rows for the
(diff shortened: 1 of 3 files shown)


packages/runtime-common/index-runner.ts
```diff
@@ -544,9 +550,40 @@ export class IndexRunner {
...
+            Object.assign(new CardError(message, { status: 500 }), {
+              stack: (err as Error)?.stack,
+            }),
+          );
+      error.message = message;
+      let entry: FileErrorIndexEntry = {
+        type: 'file-error',
+        error,
+      };
+      await this.batch.updateEntry(url, entry);
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Confirmed — Batch.invalidate() tombstones every existing type at a URL up front (instance + file for a card), and the file-error write alone left the instance tombstone to be promoted, silently deleting the card from search on a transient error. Fixed in 5e88b86aba: now also writes an instance-error entry when the failed URL is a card, so the last-known-good instance survives the same way the in-band render-error path preserves it.

packages/runtime-common/index-runner.ts
```diff
@@ -544,9 +550,40 @@ export class IndexRunner {
...
         );
-      } else {
-        throw err;
+        return;
       }
+      // A transport-level failure (prerender timeout/abort, network error)
+      // never reaches performCardIndexing/performFileIndexing's own
+      // error-entry construction — visitFileForIndexingFused rethrows
+      // before calling indexCardWithResult/indexFileWithResults. Left
+      // uncaught here, one file's failure propagates out of the
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Fair point on the missing regression test for the mid-batch isolation behavior — flagged as open in the PR description since simulating a transport-level prerender abort isn't trivial in the current IndexRunner test harness. Open to suggestions on the lowest-effort way to cover it if you have one.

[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Regression test added in caea2d7a16. The seam turned out to already exist: test realm setup accepts a `prerenderer` override, so the test wraps the real test prerenderer with one whose `prerenderVisit` throws an abort-style transport error for a single armed URL (the failure mode in-band error paths can't reach, since those require a response document). It asserts the incremental job still commits — sibling card stays searchable — while the failed card's `file` and `instance` rows carry the error un-tombstoned with last-known-good `pristine_doc` preserved, and that a later successful rewrite clears the error. Red/green verified: against main's unfixed index-runner this test fails with the whole job rejecting.

packages/runtime-common/index-runner.ts
```diff
@@ -602,8 +609,69 @@ export class IndexRunner {
         this.#log.info(
           `${jobIdentity(this.#jobInfo)} tried to visit file ${url.href}, but it no longer exists`,
         );
-      } else {
-        throw err;
+        return;
+      }
+      // A transport-level failure (prerender timeout/abort, network error)
```
habdelra: just FYI prerender doesn't actually ever happen in this job, it's been split out into a different job

habdelra: [Claude Code 🤖] Building on the FYI above: worth rewording "prerender timeout/abort" so it doesn't read as HTML prerendering — this job's visit is the index pass (meta/types/icon/file-extract); HTML rendering runs in its own job. The failure class is still real here because the index pass rides the prerender server (`prerenderVisit`), so something like "a transport-level failure of the index visit's prerender-server request (timeout/abort, network error)" keeps the example accurate without implying HTML rendering happens in this job.

packages/runtime-common/index-runner.ts
```diff
@@ -602,8 +609,69 @@ export class IndexRunner {
...
+      // previously had in the index — for an existing card that's both
+      // `instance` and `file`. Overwriting only the `file` tombstone
+      // above would let batch.done() promote the untouched `instance`
+      // tombstone, silently removing a previously-good card from search
+      // over a transient error. Re-parse the file ourselves (the throw
+      // above lost visitFileForIndexing's internal determination)
+      // and write a matching instance-error row when it's a card, so
+      // the last-known-good instance survives the same as the in-band
+      // render-error path.
+      if (url.href.endsWith('.json')) {
```
habdelra: [Claude Code 🤖] This companion write is what protects an existing card from the promoted `instance` tombstone, but it hinges on a *second* successful `readFile` + parse. When the original failure was a reader/storage error rather than a prerender-transport one, this re-read most likely fails the same way, we land in the catch below, and the tombstone gets promoted — the exact silent-delete this block exists to prevent. (The earlier automated comment gestures at this; I verified the path is reachable.)

Since the hazard is the tombstone, the index is a more reliable oracle than the file: decide "was this a card?" from whether the URL had a non-deleted `instance` row before invalidation — either query the production table for it or have `invalidate()`/`tombstoneEntries()` report the row types it tombstoned per URL — and write the `instance-error` on that basis. The source re-parse can stay as the fallback for the brand-new-file case, where there's no prior row and nothing to lose.

habdelra (commented):
[Claude Code 🤖] Reviewed with a focus on how this interacts with the prerender-split work (the index/prerender job separation, the `prerendered_html` tables, and the two-channel error model), plus general correctness. Verdict: **no conflict with the split — this is complementary — and the fix is sound.** Two inline comments; the non-line-specific findings are below.

**Split integration (all verified against `main`):**
- The catch sits only in the index-visit loop, and its error rows land on `boxel_index.error_doc` — exactly where index-channel errors belong now that render errors live on `prerendered_html.error_doc`. No cross-channel bleed.
- The prerender job's own visit (`index-runner/prerender-html-visit.ts`) already isolates per-URL failures and writes render-error rows with dep enrichment — so this PR closes the same all-or-nothing hole on the index side that the render side already handles. The two channels now share the isolation model.
- A failed URL still flows to the spawned prerender job as an `update` change, which handles its own rendering/failure independently; error rows written here get the batch's generation like any row, so freshness comparisons stay coherent; and the render-side reconcile sweep keys on render errors, so an index-error row causes no repair churn.

**Correctness claims verified:**
- `invalidate()` does tombstone every existing row type up front, so the `instance-error` companion write is load-bearing — without it an existing card's untouched `instance` tombstone is promoted at `done()`. Good catch folding that in.
- `updateEntry`'s error path preserves last-known-good state (types favored, `last_known_good_deps` merged into the error deps), so recovery-by-rewrite works — and the test proves it.
- The writer's refuse-empty-error-message guard is satisfied via `coerceErrorMessage` + the explicit `error.message` assignment, avoiding the historical black-hole-row failure mode.

**Non-contiguous observations (no change requested):**
- *Failure visibility shifts from loud to quiet*: a from-scratch that used to die now "succeeds" with N error rows and warn logs — which is the point — but the signal now lives in `stats.fileErrors`/`instanceErrors` and queryable `has_error` rows rather than a failed job. Worth keeping in mind when eyeballing job dashboards.
- *No automatic retry for transport-failed rows*: recovery requires something to re-index the URL (edit, dep invalidation, from-scratch). That matches existing error-row semantics; if transient infra failures turn out to be common in practice, an index-channel reconcile analog could be a follow-on — out of scope here.

The regression test is well-constructed: it injects the one failure mode the in-band error paths can't produce (an abort before any response document exists) at the real `Prerenderer` seam, and asserts the four things that matter — error rows un-tombstoned, `pristine_doc` preserved, sibling card still searchable (proving `done()` ran), and the error washing out on a successful rewrite.

packages/runtime-common/index-runner.ts
richardhjtan: [Claude Code 🤖] Implemented in fb08d83cc7 as suggested: `tombstoneEntries` now records the live (non-deleted) row types it overwrote per URL, exposed as `Batch.tombstonedLiveTypes(url)`, and `tryToVisit` decides "was this a card?" from that — no file read involved for existing cards, so a reader/storage failure can no longer promote the instance tombstone. The source re-parse stays as the fallback for the brand-new-file case only (no prior row to lose). Added a second regression test covering that fallback branch (new card + armed visit failure → both error rows land un-tombstoned); the existing test now exercises the oracle path since mango had a live instance row. 63/63 pass on indexing-test.ts. One honest caveat: the reader-failing-twice scenario itself isn't directly simulated (no public seam to break the realm adapter's read), but the decision path no longer depends on a read at all for existing cards, which is what closes the hazard.

packages/runtime-common/index-runner.ts
```diff
@@ -602,8 +609,69 @@ export class IndexRunner {
...
+        try {
+          let fileRef = await this.#reader.readFile(url);
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Fixed in fb08d83cc7 — the card decision now comes from the index (which live row types the batch tombstoned at that URL) rather than a second file read, so a read-path failure can't cause the tombstone promotion. Re-parse remains only for brand-new files with no prior row to protect.

packages/runtime-common/index-runner.ts
richardhjtan: [Claude Code 🤖] Thanks — reworded in fb08d83cc7: the comment now says the failure is a transport-level failure of the index visit's own prerender-server request (or a reader error), and notes explicitly that HTML prerendering runs in a separate job.

habdelra (approved):

PR #5395, opened 2026-07-03
Stamp _isCardInstance and cardTitle into file search docs during indexing

FadhlanR:
## What

Every realm file gets a `type='file'` index row, and a card-instance `.json` is dual-indexed (an `instance` row plus a `file` row). The fused visit computes "is this a card instance" at index time (`parsedCardResource`) but the file indexer never learned it. This PR threads that fact through to `performFileIndexing` and stamps two synthetic keys into the file row's `search_doc`:

- **`_isCardInstance: true`** — marks the `file` row of a dual-indexed card-instance `.json` so a mixed cards+files search can exclude it (the card already appears via its `instance` row). Stamped on the dependency-error entry too, so a card `.json` whose card is in an error state stays out of file search. Only stamped when true; plain file docs don't carry the key. Added to the searchable-parity ignore set alongside `_cardType`.
- **`cardTitle`** — the file's name aliased under the key card docs use (CardDef's title field is named `cardTitle`). The search sheet's term filter substring-matches via `contains: {cardTitle}` (full-text `matches` only matches whole words) and A-Z sort compiles to `search_doc->>'cardTitle'` with NULLS LAST, so without the alias files never prefix-match and always sink below every card in a mixed result set. It is not a claim that the file is a card.

Both keys are stamped generically in `file-indexer.ts`; no `packages/base` or query-engine changes.

## Rollout caveat

Pre-existing index rows lack both keys until their realm reindexes — until then, card-instance `.json` files will linger in file search and files will sort/match degraded, per realm. Forced reindex vs. natural cadence is a ship-time decision for the downstream mixed-search work.

## Testing

Extended `packages/realm-server/tests/indexing-test.ts`:
- the file entry for a card-instance `.json` fixture carries `_isCardInstance: true` and `cardTitle` = file name
- a plain text file's search doc carries `cardTitle` = file name and no `_isCardInstance` key

### DIFF
--- a/packages/runtime-common/index-runner/file-indexer.ts
+++ b/packages/runtime-common/index-runner/file-indexer.ts
@@ -176,11 +176,11 @@ export async function performFileIndexing({
   //   appears via its `instance` row). It rides on the dependency-error entry
   //   too, so a card .json whose card is in an error state stays out of file
   //   search. Stamped only when true; plain file docs don't carry the key.
-  // - `cardTitle` on a file doc is NOT a claim that the file is a card: it is
-  //   the file's display title (its name) aliased under the key card docs use
-  //   (CardDef's title field is named `cardTitle`), so one mixed query can
-  //   substring-match (`contains: {cardTitle}`) and A-Z sort
-  //   (`search_doc->>'cardTitle'`, NULLS LAST) cards and files uniformly.
+  // - `_title` is the row's display title (a file's is its name) under a
+  //   neutral key that card docs also carry (stamped alongside `_cardType`
+  //   during card render), so one mixed query can substring-match
+  //   (`contains: {_title}`) and A-Z sort (`search_doc->>'_title'`, NULLS
+  //   LAST) cards and files uniformly.
   let searchData = {
     url: fileURL,
     sourceUrl: fileURL,
@@ -188,7 +188,7 @@ export async function performFileIndexing({
     contentType,
     ...(extractResult.searchDoc ?? {}),
     ...(isCardInstance ? { _isCardInstance: true } : {}),
-    cardTitle: name,
+    _title: name,
   };
 
   // Runtime deps are the source of truth. Use index-backed lookup only to
--- a/packages/runtime-common/index-runner/file-indexer.ts
+++ b/packages/runtime-common/index-runner/file-indexer.ts
@@ -32,6 +32,9 @@ export interface FileIndexerOptions {
   lastModified: number;
   resourceCreatedAt: number;
   hasModulePrerender?: boolean;
+  // True when this file is a card-instance .json — the fused visit dual-
+  // indexes those (an `instance` row plus this `file` row).
+  isCardInstance?: boolean;
   realmURL: URL;
   auth: string;
   jobInfo: JobInfo;
@@ -59,6 +62,7 @@ export async function performFileIndexing({
   lastModified,
   resourceCreatedAt,
   hasModulePrerender,
+  isCardInstance,
   realmURL: _realmURL,
   auth: _auth,
   jobInfo,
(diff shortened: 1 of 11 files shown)


packages/runtime-common/index-runner/file-indexer.ts
```diff
@@ -164,6 +168,29 @@ export async function performFileIndexing({
...
+  //   substring-match (`contains: {cardTitle}`) and A-Z sort
+  //   (`search_doc->>'cardTitle'`, NULLS LAST) cards and files uniformly.
+  let searchData = {
+    url: fileURL,
+    sourceUrl: fileURL,
+    name,
+    contentType,
+    ...(extractResult.searchDoc ?? {}),
+    ...(isCardInstance ? { _isCardInstance: true } : {}),
+    cardTitle: name,
```
habdelra: shouldn't we prefix this with "_" in order to not clobber user namespace like we do with _cardType and _isCardInstance

habdelra: Also, we could remove the "card" ambiguity completely if we just use "_fileTitle". it still feels weird that we use the term "card" when this is definitely not a card

FadhlanR: [Claude Code 🤖]

Good callout on the naming — digging in, both points (the `_` prefix and `_fileTitle`) run into the same constraint, so let me lay it out.

The target UX for files-in-search is a single **interleaved A–Z list** of cards and files together (not a cards block then a files block). That forces the sort/filter onto **one shared `search_doc` key** across both row types:

- A multi-key sort like `[{by: cardTitle}, {by: fileTitle}]` doesn't interleave — each key sorts `NULLS LAST`, so the primary key partitions all cards ahead of all files (files are null on `cardTitle`). You get cards-then-files, not A–Z.
- A `COALESCE(cardTitle, fileTitle)` sort *would* interleave, but the query's `sort.by` is a single field path — it can't express a coalesce.

So the file row's title has to live under the **same key the card row uses**. Cards key on `cardTitle` — it's a real declared `CardDef` field (`card-api.gts`), so it can't be renamed and it resolves in `getField()` for free. That's why the file row aliases its name under `cardTitle` here rather than `_fileTitle`: a `_fileTitle` (or `_cardTitle` on the file only) would be a *different* key from the card side, so files would sort/filter as null and fall out of the unified list.

That leaves two real options:

- **A (this PR):** keep `cardTitle` on the file row. Minimal, works with the shared-key sort today. Downside is exactly what you flagged — `cardTitle` on a non-card reads oddly.
- **B (follow-up):** introduce a neutral synthetic `_title` stamped on **both** card and file docs, sort/filter on `_title`, and add a `_title` synthetic-field case in `getField()` like `_cardType`. That gets the proper `_` prefix *and* drops the "card" wording *and* keeps the single shared key.

B is the clean end state and addresses both your points, but it touches the card index path + query engine, so I'd rather not fold it into this PR. Proposing we keep A here and track B as a follow-up — sound good?

habdelra: i see--yeah _title is probably a better choice if its something that needs to be consistent across cards and files

FadhlanR: [Claude Code 🤖]

Done — went with `_title`. It's now stamped on both card docs (alongside `_cardType` in the prerender meta route) and file docs, treated as a queryable synthetic `StringField` in the query engine, and excluded from the searchable-parity diff — same treatment as `_cardType`. The misleading `cardTitle` alias on file rows is gone.

packages/runtime-common/index-runner/file-indexer.ts
```diff
@@ -179,13 +206,7 @@ export async function performFileIndexing({
...
       type: 'file-error',
       error: normalizedDependencyError,
-      searchData: {
-        url: fileURL,
-        sourceUrl: fileURL,
-        name,
-        contentType,
-        ...(extractResult.searchDoc ?? {}),
-      },
+      searchData,
```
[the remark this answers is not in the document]
FadhlanR: [Claude Code 🤖]

Good catch — fixed. The `instance-error` / `file-error` branch now captures the production version once and assigns `search_doc` *after* the spread, overlaying the fresh `searchData` onto the last-known-good doc (`{ ...production.search_doc, ...entry.searchData }`). So the freshly-stamped synthetic keys (`_title`, `_isCardInstance`) survive on error rows even when the prior production row predates them, while last-known-good fields still carry forward. Added a unit test in `index-writer-test.ts` covering exactly that rollout case.

habdelra (approved):
looks like there are still failing tests, but otherwise looks ok to me

burieberry (approved):

PR #5396, opened 2026-07-03
ai-bot: render markdown skills entrypoints in the prompt 

jurgenwerk:
Today, when a skill is enabled in a room, its entire instruction text is included in every prompt sent to the LLM. This PR changes that for one skill — Source Code Editing: instead of its full ~7KB text, the prompt carries a ~0.5KB entry-point skill that says the skill exists and where it lives. The bot fetches the actual instructions with `readRealmFile` only when the user asks for a code edit — once per conversation, visible in the chat as a "Read file" step.

Source Code Editing was chosen because it's the easiest skill to prove the mechanism with: no command bindings and a single self-contained document.

**Before (main) — pushed card, snapshot at enable time**

```
Enter code mode
  │
  ▼
loadDefaultSkills('code')
  = [boxel-development, boxel-environment, Skill/source-code-editing]  ← card
  │
  ▼
UpdateRoomSkillsCommand
  loads the card + companion .md, UPLOADS full content to Matrix
  │
  ▼
room state: enabledSkillCards[ source-code-editing (snapshot) ]
  │
  ▼
EVERY bot turn:
  download snapshot from Matrix → JSON.parse → full ~7KB
  pasted into the system prompt
                                   (edits to the skill? invisible
                                    until someone re-enables it)
```

**After (this PR) — pulled markdown, live at use time**

```
Enter code mode (or create room in code mode)
  │
  ▼
loadDefaultSkills('code')
  = [boxel-development, boxel-environment]           ← unchanged cards
  + skills/code-mode/SKILL.md                         ← entry point (URL)
  │
  ▼
UpdateRoomSkillsCommand
  loads the SKILL.md, uploads only the ~0.5KB entry point
  │
  ▼
room state: enabledSkillCards[ ..., code-mode/SKILL.md ]
  │
  ▼
EVERY bot turn:
  entry-point body in prompt: "code skills exist; load before editing"
  │
  ▼  only when the user actually asks for a code edit
readRealmFile(realm, url)
  → realm server, delegated user-scoped token
  → LIVE skills/source-code-editing/SKILL.md
  → instructions enter the conversation once, reused after
                                   (edit the skill file → next
                                    fetch sees it immediately)
```

Changes:
- `getEnabledSkills` can read markdown skills (it JSON-parsed everything as cards before; frontmatter parser now shared via runtime-common)
- code mode enables the entry point (on switch and on room creation); source-code-editing dropped from the pushed defaults
- mise tasks set the dev delegation secret so `readRealmFile` works locally

The other two code-mode skills, Boxel Development and Boxel Environment, stay pushed as cards: the bot doesn't yet read command bindings from markdown-skill frontmatter (Boxel Environment carries 21 of them), and their content needs an editorial pass before converting. Flipping them is a tracked follow-up.

Merge cardstack/boxel-skills#84 first (it adds the two SKILL.md files this references). Verified end to end locally; unit tests + clean type checks.


jurgenwerk (commented):
[Claude Code 🤖] Marked the dual-path (push + pull) seams inline — which parts are the transition and what goes away when.

packages/host/app/services/ai-assistant-panel-service.ts
```diff
@@ -453,6 +453,11 @@ export default class AiAssistantPanelService extends Service {
           }
 
           ({ roomId } = await createRoomCommand.execute(input));
+          // Rooms created while already in code mode miss the mode-switch
+          // hook that enables the entry-point skill; add it here.
+          if (this.operatorModeStateService.state.submode === 'code') {
+            await this.matrixService.activateCodeModeEntryPoint(roomId);
```
lukemelia: Fine for this validation phase, but I think with the entrypoint approach, we can eliminate activation on submode switch

packages/runtime-common/ai/prompt.ts
```diff
@@ -436,14 +437,60 @@ async function getEnabledSkills(
...
+        return (JSON.parse(content) as LooseSingleCardDocument)?.data;
       }),
     );
   }
   return [];
 }
 
+// A skill enabled as a markdown file (SKILL.md), rather than a Skill card.
+export function isMarkdownSkillFile(fileDef: SerializedFileDef): boolean {
+  return /\.(md|markdown)$/i.test(fileDef.sourceUrl ?? fileDef.name ?? '');
```
lukemelia, line 488: a markdown skill file is only a markdown skill file if it has frontmatter where boxel.kind == 'skill'

lukemelia (approved):

packages/runtime-common/ai/prompt.ts
```diff
@@ -436,14 +437,60 @@ async function getEnabledSkills(
...
+        // pipeline expects, so skillCardsToMessages and getTools need no
+        // special-casing. The JSON.parse fallthrough below is the legacy
+        // pushed-card path; it is removed when the push model is retired.
+        if (isMarkdownSkillFile(cardFileDef)) {
+          let { title, body } = parseMarkdownSkill(content, cardFileDef);
+          return {
+            id: cardFileDef.sourceUrl,
+            type: 'card',
+            attributes: { title, instructions: body },
+          } as unknown as LooseCardResource;
```
jurgenwerk: Probably should not be LooseCardResource

PR #5416, opened 2026-07-07
Bridge worker-originated realm events through the worker manager

habdelra:
Background jobs run in worker child processes that hold no matrix client, so they can't broadcast realm events. Every client-facing realm event today is emitted by the web-tier `Realm` in response to a request that awaited its indexing job. A job whose result nobody in the web tier awaits has no path to the client.

This adds a generic, typed worker→manager→realm-server bridge so a worker-originated request reaches the realm server exactly once (no per-replica fan-out). Its first request type, `broadcast-realm-event`, broadcasts a realm event to subscribed hosts without the worker holding a matrix client.

## Flow

```
task → TaskArgs.reportRealmEvent(event)              (transport-agnostic; mirrors reportProgress)
     → Worker fan-out
     → worker.ts   sendWorkerRequest(type, payload)  → IPC  worker-request|{type,payload}
     → worker-manager.ts  worker.on('message')  → dispatchWorkerRequest (switch on type)
     → forwardWorkerRealmEvent  resolve origin + canonical realmURL from --fromUrl/--toUrl, HMAC-sign
     → POST /_worker-request   (realm server)
     → handle-worker-request.ts  verify signature → dispatch on type
         broadcast-realm-event: normalizeRealmURL → reconciler.lookupOrMount → realm.broadcastEvent()
     → matrix session rooms
```

## Design

- **Generic transport.** A worker request is a `{ type, payload }` envelope; both the manager and the realm server dispatch on `type`. `broadcast-realm-event` owns the url-mapping resolution + realm broadcast. Another worker-originated request is a distinct `type` plus a handler on each side — the transport doesn't change. The task-facing API stays `reportRealmEvent(event)`.
- **Canonical realm-URL routing.** `Realm.url` — the realm-registry key that `reconciler.lookupOrMount` matches and that hosts subscribe to — is `main.ts`'s `hrefs[i][0]`: the `--fromUrl` for URL-form mappings and the `--toUrl` for realm-prefix mappings. The manager forwards the event carrying that canonical URL while POSTing to the `--toUrl` origin (e.g. base maps `cardstack.com/base/` → `app.boxel.ai/base/`). Realm URLs are canonicalized with `normalizeRealmURL` on both the manager match and the endpoint lookup, since registry keys are canonicalized.
- **Exactly-once, no per-replica fan-out.** The single worker manager posts once → the realm server broadcasts once, avoiding the duplicate broadcasts a Postgres `NOTIFY` re-broadcast would produce, and HTTP can carry a full from-scratch invalidation set. Delivery is at-least-once — a lost HTTP response after a successful broadcast triggers a re-send on retry, which consumers tolerate because realm events drive idempotent re-invalidation.
- **Auth.** HMAC-over-body-and-timestamp shared secret, the same pattern `/_delegate-session` uses, keyed on `REALM_SECRET_SEED` (which both the manager and the realm server already hold — no new secret or env var). Sign / verify / client live in one node-only module (`runtime-common/worker-request.ts`).
- **Delivery** is best-effort with a short retry (retry transport / 5xx, treat 4xx as terminal). A dropped event degrades to the same staleness the periodic re-index / catch-up sweep already tolerates.

The bridge is additive and inert until a task calls `reportRealmEvent`.

## Testing

- `worker-request-signature-test` — HMAC sign/verify: tamper, wrong-secret, missing/malformed/stale timestamp, window edges.
- `worker-request-forwarder-test` — resolution across all three mapping shapes (URL-form from≠to, realm-prefix, from==to), canonical-URL normalization, best-effort retry (5xx and thrown-fetch retried, 4xx terminal), unmanaged-realm drop, and the manager dispatch-on-type hop.
- `server-endpoints/worker-request-test` — end to end against the live matrix stack: a signed request broadcasts the event to a subscribed host; realmURL canonicalization; and 401 (missing / invalid / wrong-secret / stale signature), 404 (realm not hosted here), 400 (unknown type / malformed body).

`runtime-common` and `realm-server` type-check and lint clean.


packages/realm-server/lib/worker-realm-event-forwarder.ts
```diff
@@ -0,0 +1,105 @@
...
+
+// Forward a worker-originated realm event to the realm server over the
+// authenticated /_worker-event endpoint. Best-effort with a short retry on
+// transport / 5xx errors: a dropped event degrades to the same staleness the
+// periodic re-index / catch-up sweep already tolerates, so a final failure is
+// logged, not fatal. A 4xx (bad request / auth) won't improve on retry, so we
+// stop early. Returns true on delivery.
+export async function forwardWorkerRealmEvent({
+  event,
+  urlMappings,
```
habdelra: it would be nice if this was a more generic mechanism so it coulkd have an arbitrary payload, and we just happen to use a specialized event where the payload is urlMappings that get resolved

packages/realm-server/worker-manager.ts
```diff
@@ -978,6 +980,33 @@ async function startWorker(
...
+          let event: RealmEventContent;
+          try {
+            event = JSON.parse(payload) as RealmEventContent;
+          } catch (e) {
+            log.error(`Failed to parse realm event from worker ${name}: ${e}`);
+            return;
+          }
+          forwardWorkerRealmEvent({
+            event,
+            urlMappings,
```
habdelra, line 1001: it would be nice if the payload could be more generic instead of tied to url mappings

habdelra: maybe we could have an "eventType" that on the realm server can map to a specific type of event handler where in our case we resolve the url mappings for our event type

packages/realm-server/lib/worker-realm-event-forwarder.ts
```diff
@@ -0,0 +1,105 @@
...
+      `dropping worker realm event for unmanaged realm ${event.realmURL} (eventName=${event.eventName})`,
+    );
+    return false;
+  }
+  // Normalize the event's realmURL to the canonical (to) realm url so the
+  // endpoint's registry lookup resolves it regardless of which mapping side
+  // matched.
+  let normalizedEvent = {
+    ...event,
+    realmURL: target.href,
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in efe4076. You're right — `Realm.url` is the `--fromUrl` for URL-form mappings (`main.ts` backfills `hrefs[i][0]`), so rewriting the event to the `--toUrl` would make `reconciler.lookupOrMount` 404 those events (e.g. base maps `cardstack.com/base/` → `app.boxel.ai/base/`). The manager now keeps the canonical registry `realmURL` and only POSTs to the `--toUrl` origin. Added a from≠to (base) forwarder test that would have caught this.

packages/realm-server/lib/worker-realm-event-forwarder.ts
habdelra: [Claude Code 🤖] Done in efe4076. The bridge is now a generic transport: the worker sends `worker-request|{type,payload}` over IPC and the manager dispatches on `type`. `broadcast-realm-event` is the first specialization that owns the url-mapping resolution + realm broadcast; a new worker→manager request is just a new `type` + handler, no transport change. Task-facing API stays `reportRealmEvent(event)`.

packages/realm-server/worker-manager.ts
habdelra: [Claude Code 🤖] Done in efe4076 — the IPC payload is now the generic `{ type, payload }` envelope; url-mapping resolution lives only inside the `broadcast-realm-event` handler (`worker-request-forwarder.ts`), not in the transport.

habdelra: [Claude Code 🤖] Done in efe4076 — the realm-server endpoint (renamed `/_worker-request`) now dispatches on `type` too, so a request type maps to its own handler server-side. `broadcast-realm-event` resolves the realm and broadcasts; future types add a `case` without touching the endpoint plumbing.

packages/realm-server/lib/worker-realm-event-forwarder.ts
```diff
@@ -0,0 +1,105 @@
...
+  realmURL: string,
+): URL | undefined {
+  for (let [from, to] of urlMappings) {
+    let fromHref = from instanceof URL ? from.href : from;
+    if (realmURL === to.href || realmURL === fromHref) {
+      return to;
+    }
+  }
+  return undefined;
+}
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in efe4076 — `resolveWorkerRealmEventTarget` now canonicalizes the incoming realmURL via `normalizeRealmURL` before matching, so a trailing-slash/spelling difference no longer silently drops the event.

packages/realm-server/handlers/handle-worker-event.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    let realmURL = event.realmURL;
+    // Route through the reconciler so a realm that hasn't been touched on this
+    // process since the last restart still mounts on demand — the request may
+    // land on any replica behind the load balancer, not necessarily one that
+    // already has the realm mounted.
+    let realm;
+    try {
+      realm = await reconciler.lookupOrMount(realmURL);
+    } catch (e: any) {
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in efe4076 — the endpoint (now `handle-worker-request.ts`) runs `normalizeRealmURL` before `reconciler.lookupOrMount`, so a non-canonical realmURL resolves against the canonicalized registry key. Added an endpoint test that posts a trailing-slash-less realmURL and asserts it resolves.

packages/realm-server/routes.ts
```diff
@@ -309,6 +310,10 @@ export function createRoutes(args: CreateRoutesArgs) {
   // Shared-secret authenticated (HMAC over body + timestamp); auth is handled
   // inside the handler because the signature covers the request body.
   router.post('/_delegate-session', handleDelegateSession(args));
+  // Handles a worker-originated request bridged in through the worker manager,
+  // dispatched on its `type` (CS-11808). Shared-secret authenticated (HMAC over
+  // body + timestamp), same as /_delegate-session — auth is inside the handler.
+  router.post('/_worker-request', handleWorkerRequest(args));
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch — updated the PR description to the current shape: the endpoint is `/_worker-request`, the IPC message is `worker-request|{type,payload}`, and the flow dispatches on `type` (`broadcast-realm-event` being the first). The stale `/_worker-event` / `realm-event|` names are gone.

lukemelia (approved):

PR #5428, opened 2026-07-08
feat: mixed cards + files entry search with scope and client parity

FadhlanR:
Closes CS-11775, CS-11776.

## What

Makes the search **entry** API return card instances and files from a single query, instead of routing all-or-nothing by filter type, and updates every consumer to the new mechanism so the API change and its consumers land together. Kind is discriminated entirely by the caller's filter — a card-type-anchored filter matches only `instance` rows via `types`, a FileDef-anchored filter only `file` rows — so no per-kind routing heuristic is needed and no wire member is added.

## API changes (CS-11775)

- **`_search` gains an `'all'` mode** (`WHERE i.type IN ('instance','file')`) beside the existing `'instance'`/`'file'` modes. Nothing is deduped by default: a dual-indexed card `.json` (an `instance` row and a `file` row sharing one `url`) surfaces as two rows via `GROUP BY (url, type)`; COUNT/LIMIT/OFFSET run over the same grouping so `total` stays consistent. Instance rows honor `includeErrors`; file rows only surface healthy.
- **`_isCardInstance` is a queryable synthetic field**: `eq: { item._isCardInstance: null }` keeps cards + plain files and drops the card `.json` file row (the key is stamped only on that row). The positive NULL match is load-bearing — `not: { eq: { _isCardInstance: true } }` would compare NULL on every other row and drop everything.
- **Unified projection + assembly.** `search()` carries the file-only columns in either projection when file rows are in scope; `searchEntries` folds the former instance and file-meta loops into one loop that branches per row on `row.type`. The single-URL instance GET pins `'instance'` so a bare file URL never resolves as a card entry.
- The federated fan-out/merge is untouched.

## Consumer changes (CS-11776)

Every consumer falls into one of two filter idioms:

- **Cards-only** → anchor with `{ type: baseCardRef }` unless a positive type ref is already present (the engine shares one `types` cross-join alias, so two positive type conditions cannot compose). Shared helper `anchorQueryToCardInstances` lives in `runtime-common/query-field-utils.ts`.
- **Mixed-but-deduped** → compose `eq: { _isCardInstance: null }` (legacy spelling) / `'item._isCardInstance': null` (wire spelling).

| Consumer | Fix |
|---|---|
| Search sheet (`query-builder.ts`) | term filter matches the synthetic `_title` key so files match by name; sheet queries compose the `_isCardInstance` dedup |
| Card choosers (modal + mini) | new `cardsOnly` option on `buildSearchQuery`/`buildRecentsQuery`, passed as `@cardsOnly` through `PanelContent`; the preselected-card `store.search` query is anchored defensively |
| Recents queries (incl. compact) | scoped the same way — their `cardUrls` also name the dual-indexed `.json` file rows |
| Host-mode head prefetch | `'item._isCardInstance': null` in the wire query (its `cardUrls` matched both rows and the file row's head could win the order tiebreak — the playwright `head-tags` failure) |
| create-listing example query | same wire-level dedup |
| cards-grid "All Cards" | anchored with `{ type: baseCardRef }`; the "All Files" group owns the file side. (Collateral: `cards-grid` wasn't in the original ticket scope — the mixed default forces it.) |
| `search-cards` AI command | anchors an otherwise-unscoped query via the shared helper |
| Playground, detail-panel, create-specs, sync-openrouter, file-tree | no change — already type-anchored (self-discriminating) or file-only by design |

When a positive type condition is present (a picked type or a typed base filter), neither anchor nor dedup is added: one kind is already selected, and a card `.json` matching an explicit file-type filter correctly surfaces once, as a file.

## Tests

- Realm-server engine tests: mixed default returns both kinds; `eq item._isCardInstance null` dedups; positive card-type anchor stays cards-only; the head-scoped single-URL query mirrors the prefetch dedup.
- Host: query-builder unit expectations updated for `_title` + scoping; implicitly cards-only realm/indexing test queries anchored; new "All Cards stays cards-only" integration test.

## Deployment & backward compatibility

No transitional wire mechanism ships with this change — compatibility is handled operationally. The asymmetry that drives the plan:

- **Old client → new server** degrades softly at worst. NULL semantics keep most existing queries card-only automatically (a positive predicate on a card field, a `not: { eq: … }`, or a positive type anchor all drop file rows, since file rows lack those `search_doc` keys / `types`). Only filter-less/sort-only queries, `cardUrls` queries (a card `.json` URL matches both of its rows), and queries on keys files also carry (`_title`) see mixed rows — extra `file-meta` entries that current clients already deserialize; no errors.
- **New client → old server** is a hard break: the `_isCardInstance` synthetic field is only queryable as of this PR, so an old realm-server rejects it as an unknown field.

### Prerequisite: index state (per environment, staging then prod)

Dedup relies on `_isCardInstance: true` being stamped on every card-instance `.json` file row (stamping landed separately with the CS-11774 indexing change; this PR only consumes it, and includes no reindex trigger). Before deploying, verify a full reindex has completed since that stamping deploy — e.g. spot-check `boxel_index` that `file`-type rows for card `.json`s have `search_doc->>'_isCardInstance' = 'true'`. If not, force a full reindex and wait. Un-restamped cards appear twice in the search sheet until their realm reindexes.

### Deploy order

1. **realm-server first** — the mixed default goes live; a not-yet-updated host doesn't send the dedup filter, so the only skew passed through is soft (transient dupes in search results, no errors).
2. **host second**, once realm-server is healthy — it starts sending `eq: { item._isCardInstance: null }` and the skew window closes.

Rolling back is the reverse: **host first, then realm-server**, so the deployment never passes through new-host + old-realm (the erroring direction). A realm-server rollback needs no index migration — the stamped key is inert to old code.

### Post-deploy smoke tests

- Search sheet: cards **and** plain files appear; no card appears twice.
- Cards-grid "All Cards": cards only; errored instances still show their error tile.
- Card chooser / MiniCardChooser: cards only.
- `search-cards` AI command: card instances only.
- Recents: each entry appears once.
- Published view: head prefetch resolves a single card.

### Client follow-up

boxel-cli's filter-less and `cardUrls` searches will show the extra file rows until it adopts the dedup filter — tracked in CS-12052, to be released together with or after the realm-server deploy (an old server rejects the new filter).


docs/cs-11775-11776-mixed-search-plan.md
habdelra, line 1: i don't think we want to commit this file

habdelra:
I believe for for "all" type of query it will only work with full text search using the "matches" filter, as there are no fields in common between file-def and card-def. we should return a 400 error if the client uses combines an "all" with something that is not a "matches" filter.

habdelra:
Can you include the new wire format for the "all" mode? I'm having a hard time finding it the tests

habdelra:
what happens when you mix `_isCardInstance:null` with `type: xxx`? what does that even mean?

habdelra:
what does `_isCardInstance: false` mean? why is the guidance to use `_isCardInstance: null` instead of `_isCardCardInstance: false`

docs/cs-11775-11776-mixed-search-plan.md
FadhlanR: [Claude Code 🤖] Removed in 00f8008c8d — the doc was a planning scratchpad and should not have been part of the PR.

FadhlanR:
[Claude Code 🤖] @habdelra answering the discussion points together, since they are related:

**Is there an `'all' wire mode? / 400 for non-`matches` filters**

There is no new wire member — the request format is unchanged. Searching both kinds is simply the endpoint's default now: with no type filter, a query matches `instance` and `file` rows alike. Kind discrimination happens entirely through the filter's type ref — `{ type: <any card type> }` (e.g. the CardDef base ref) matches only card instances via `types`, a FileDef-anchored ref matches only files. On the response side the only change is that a mixed result can interleave `file-meta` resources with card resources, in the same envelope as before.

On restricting mixed queries to `matches`: the two kinds do share queryable fields — the synthetic `_title` key (card title / file name) is stamped on both row kinds, and the search sheet relies on `contains: { _title: term }` matching files by name. Full-text `matches` spans both too. A filter on a card-only field never matches file rows (the key is absent from their search doc), which acts as the intended narrowing rather than an error — and since mixed is the default with no opt-in member, a 400 on non-`matches` filters would reject every typed query. Filters on genuinely nonexistent fields already error today.

That said — would you prefer an explicit wire member (e.g. `scope: 'instance' | 'file' | 'all'`) over filter-driven discrimination? If you think the implicitness will bite API consumers, adding one is straightforward.

**`_isCardInstance` semantics (`null` vs `false`, mixing with `type:`)**

Agreed the current shape is confusing, and we're changing it. The key is stamped on a card `.json`'s *file* row and really means "this file is a card instance's source", so the plan is to rename it to `_isCardSourceFile` and make `eq: { _isCardSourceFile: false }` the dedup spelling — an absent key will be treated as `false` at query translation, so the load-bearing `null` match goes away and `false` means what it reads as. Mixing it with `type: xxx` is redundant but harmless: a card-type filter already selects instance rows, which never carry the key; our query builders skip the dedup filter whenever a positive type ref is present.

habdelra:
so one thing we floated a bunch of times was using a type of baseDef to mean both card defs and file defs since they both inherit from this type, what did you think of that idea?

habdelra:
(also i have claude working on a more thorough code review as we speak)

packages/runtime-common/index-query-engine.ts
```diff
@@ -813,39 +842,71 @@ export class IndexQueryEngine {
...
+    // resource + native renderings. Each is gated on `i.type = 'file'` so a
+    // mixed query never drags a card row's (potentially multi-megabyte)
+    // `search_doc` / `isolated_html` / `markdown`: within a `(url, type)`
+    // group the type is constant, so the CASE yields the value for a file
+    // group and NULL for an instance group (which the assembly loop ignores
+    // — it only calls `fileEntryFromResult` on `file` rows). Appended to
+    // either projection only when file rows are in scope.
+    let fileOnly = (col: string) =>
+      `ANY_VALUE(CASE WHEN i.type = 'file' THEN ${col} END)`;
+    let fileColumns = `, ${fileOnly('i.search_doc')} as search_doc, ${fileOnly(
```
habdelra, line 877: [Claude Code 🤖] Two of these columns are dead weight, and they're the two potentially large ones. The mixed assembly loop in `realm-index-query-engine.ts` reads a file row through exactly two functions:

- `enumerateFileRenderings(file)` — reads `fittedHtml` / `embeddedHtml` / `atomHtml` / `headHtml`, all of which the renderSet projection already carries dual-read.
- `fileResourceFromIndex(url, file)` — reads `searchDoc`, `resource` (pristine_doc), `types`, `lastModified`, `resourceCreatedAt`.

Neither reads `markdown` or `isolated_html`. `markdown` for a file row is the file's extracted content and can be very large (this is the same column whose oversized base64 runs forced the `markdown_search_text()` capping machinery for the tsvector index), and both get hauled per file row on every mixed query in *both* projections — including `dataOnly`, whose own comment says "no rendered HTML entries."

Suggest trimming `fileColumns` to what the assembly actually consumes: `search_doc`, `last_modified`, `resource_created_at`, `realm_url`, `indexed_at`. If a future consumer needs a file's markdown or isolated rendering through the mixed path, it can be added alongside the code that reads it.

packages/runtime-common/index-query-engine.ts
```diff
@@ -1914,14 +1939,19 @@ async function getField(
...
-      // (e.g. the cards grid sorts by the card's display name; a mixed
-      // cards+files search sorts/filters both row types on '_title'). the
-      // index-runner / prerender meta route inject these into the searchDoc
-      // during index time.
+      // this is a little awkward--we have the need to treat '_cardType',
+      // '_title' and '_isCardInstance' as fields that we can query against
+      // from the index (e.g. the cards grid sorts by the card's display name;
+      // a mixed cards+files search sorts/filters both row types on '_title',
+      // and dedups a card's dual-indexed `.json` file row with
+      // `eq: { _isCardInstance: null }`). the index-runner / prerender meta
```
habdelra: [Claude Code 🤖] The comment says `_isCardInstance` "is only queried for existence (`eq: null`)", but nothing enforces that, and the engine compiles other spellings in ways that quietly diverge between adapters. Worth understanding the mechanics here because they also shape the rename that was agreed in the PR discussion:

- `eq: { _isCardInstance: null }` → `fieldEqFilter`'s null branch → `search_doc ->> '_isCardInstance' IS NULL`. Type-agnostic, works on both adapters. This is the only spelling the PR uses — good.
- `eq: { _isCardInstance: true }` → the positive-polarity path routes to `handleJsonContainsQuery`, which sees a non-string leaf and degrades to `->>` extraction equality with a **boolean** bind. On Postgres `->>` yields the text `'true'` and node-pg happens to serialize the bind as `'true'`, so it matches by coincidence; on SQLite `->>` of JSON `true` yields integer `1` and the comparison depends on how the adapter binds a JS boolean. Same query, adapter-dependent answer — exactly the class of divergence the expression layer exists to prevent.

For the planned rename to `_isCardSourceFile` with `eq: { _isCardSourceFile: false }` treating an absent key as `false`: that rewrite can't live in this `getField` shim (it only fabricates a field definition; it never sees the value). It belongs in `fieldEqFilter` as a synthetic-key special case that emits, at positive polarity, something like:

```
( search_doc ->> '_isCardSourceFile' IS NULL OR search_doc ->> '_isCardSourceFile' = 'false' )
```

and stays polarity-aware so `not: { eq: { _isCardSourceFile: false } }` negates the whole disjunction (the same absent-path NULL trap that motivated the `FilterPolarity` machinery for `@>` vs `->>`). Until then, a comment here noting that only the `null` spelling is supported — or a validation error for non-null values — would keep someone from discovering the adapter divergence in production.

packages/runtime-common/query-field-utils.ts
```diff
@@ -403,6 +404,33 @@ export function getTypeRefsFromFilter(
...
+// so a second positive type condition would compile to an unsatisfiable
+// AND over that alias.
+//
+// Residual gap: a mixed `any: [{ type: … }, { <untyped> }]` counts as
+// carrying a positive ref (so the anchor is skipped), yet its untyped branch
+// can still match file rows. This is intentional — anchoring such a filter is
+// worse: the added `{ type: baseCardRef }` ANDs against the same `types`
+// alias as the typed branch and matches nothing. We prefer the narrower skip
+// (a small file-row leak) over an anchor that drops the caller's own results.
+export function anchorQueryToCardInstances(query: Query): Query {
```
habdelra: [Claude Code 🤖] This helper is doing its best with the tools available, but the "residual gap" comment frames a workaround as a permanent design. The root cause is worth naming precisely, because fixing it at the engine level would delete this heuristic entirely.

**Why the heuristic exists.** Every `{ type: … }` condition in a query compiles to a predicate against a *single shared* `jsonb_array_elements_text(types)` cross-join alias (`expressionToSql` dedupes table-valued functions by column, so all type conditions share `types0_array_element`). The fan-out gives each index row one output row per element of its `types` array, `WHERE` filters those, and `GROUP BY` keeps any url with ≥1 surviving row. That gives type conditions *exists-one-element* semantics, which composes wrongly under both `AND` and `NOT`. Verified directly against Postgres:

```sql
WITH idx(url, types) AS (
  VALUES ('spec-1',   '["Spec","CardDef","BaseDef"]'::jsonb),
         ('author-1', '["Author","CardDef","BaseDef"]'::jsonb))
-- every([{type: Author}, {type: CardDef}]): needs ONE element equal to both
SELECT url FROM idx CROSS JOIN LATERAL jsonb_array_elements_text(types) e
WHERE e = 'Author' AND e = 'CardDef' GROUP BY url;
--> 0 rows (should be author-1: its types contain both)

-- every([{type: baseCardRef}, {not: {type: Spec}}]) — what this helper emits
SELECT url FROM idx CROSS JOIN LATERAL jsonb_array_elements_text(types) e
WHERE e = 'CardDef' AND NOT (e = 'Spec') GROUP BY url;
--> BOTH rows: spec-1's CardDef element satisfies both conjuncts
```

The second result means the anchored filter this helper produces for the search-sheet shape (`every: [{type: baseCardRef}, {not: {type: specRef}}]`) does not actually exclude Specs — and neither did the bare `not: {type: specRef}` before this PR. That's a pre-existing engine bug, not this PR's, but this PR is now building consumer-facing heuristics on top of it.

**The engine-level fix.** Make `typeCondition` a self-contained membership predicate instead of a reference to the shared alias — a new resolved expression node rendered per adapter, exactly the pattern `JsonContains` established: Postgres `i.types @> '["<key>"]'::jsonb`, SQLite `EXISTS (SELECT 1 FROM json_each(i.types) WHERE value = ?)`. Then:

- `every([{type: A}, {type: B}])` means intersection (satisfiable — a card of type A also carries B in its chain),
- `not: {type: Spec}` genuinely excludes Specs,
- this helper can add its anchor **unconditionally** — no `getTypeRefsFromFilter` scan, no skip heuristic, and the documented `any:` leak disappears,
- rows whose `types` is empty stop vanishing whenever any branch of the filter references types (today the cross join against an empty array eliminates the row before `WHERE` even runs).

That's a follow-up ticket, not this PR — but I'd reword this comment to present the skip as an interim workaround for the shared-alias composition limits rather than as intended behavior, and link the follow-up once it exists.

One more subtlety worth a sentence in the doc comment: anchoring with `{ type: baseCardRef }` also drops instance rows whose `types` array is empty — instances whose render failed outright have no `types` (dependency-error rows keep theirs). For `store.search` callers that's moot (error rows are excluded there anyway), but for any `includeErrors` surface the anchor is *stricter* than "cards only": it's "successfully-typed cards only."

packages/host/app/utils/card-search/query-builder.ts
```diff
@@ -84,20 +86,67 @@ function buildTypeFilter(
...
+}
+
+function hasPositiveTypeRef(filter: Filter | undefined): boolean {
+  if (!filter) {
+    return false;
+  }
+  return getTypeRefsFromFilter(filter)?.some((r) => !r.negated) ?? false;
+}
+
+function scopeFilters(
```
habdelra, line 180: [Claude Code 🤖] Two consolidation points, both aimed at giving the "which kind(s) does this query select?" decision exactly one home:

1. `hasPositiveTypeRef` + the skip-if-typed branch here re-implement the decision `anchorQueryToCardInstances` makes in runtime-common, just against a filter array instead of a `Query`. If the shared helper grew a filter-level form (e.g. `filterSelectsSingleKind(filter)` used by both), the two call sites can't drift — and the Codex thread above about `cardsOnly` being bypassed by a positively-typed *file* filter gets fixed in one place rather than two. (That gap is real: `hasPositiveTypeRef` treats any positive ref as "one kind already selected," so `cardsOnly: true` + a FileDef-typed baseFilter happily returns files.)

2. The `{ eq: { _isCardInstance: null } }` literal built here (and its `DEDUP_FILTER` twin in the unit test, and the `'item._isCardInstance': null` wire spellings in host-mode-service / create-listing-modal / panel-content) should be a single named export from runtime-common next to `anchorQueryToCardInstances` — e.g. `excludeCardSourceFileRows()` returning the filter, plus its wire-key constant. Beyond readability (the NULL-match reads backwards — "keep rows where `_isCardInstance` is null" *keeps* the card instances), the discussion already agreed to rename the key to `_isCardSourceFile` with inverted semantics; with a named helper that rename is a one-file change instead of a grep across five consumers and their tests.

packages/host/app/services/host-mode-service.ts
```diff
@@ -320,7 +320,14 @@ export default class HostModeService extends Service {
       body: JSON.stringify({
         realms: [realmRoot],
         cardUrls: [cardJsonURL],
-        filter: { eq: { htmlQuery: { eq: { format: 'head' } } } },
+        // `item._isCardInstance: null` drops the card `.json`'s dual-indexed
+        // file row, which shares the `cardUrls` URL with the instance entry.
+        filter: {
+          eq: {
+            htmlQuery: { eq: { format: 'head' } },
+            'item._isCardInstance': null,
```
habdelra: [Claude Code 🤖] This site is the clearest illustration of why the dedup deserves a better spelling than a raw filter literal. The intent here is not "filter by a field" at all — it's "this `cardUrls` lookup should resolve to the card entry, not the card's `.json` file entry." That is *scope* selection, and the engine already models it as scope: `RealmIndexQueryEngine.searchEntries` grew an `entryTypeScope` parameter in this PR, and the single-URL `cardDocument` path pins `'instance'` through it for exactly this reason. Internal callers get a clean scope parameter; wire callers like this one have to know a synthetic-key NULL-match incantation to say the same thing.

Concretely I'd suggest one of:
- expose the scope on the wire (see my comment on `entryTypeScope` in realm-index-query-engine.ts) so this request says `scope: 'cards'` — self-describing, independent of index-stamp state, and immune to the "un-restamped rows appear twice until reindex" window the PR description documents; or
- at minimum, replace the literal with the shared named helper (see the query-builder thread) so the meaning and the planned `_isCardSourceFile` rename live in one place. Same applies to the twin literal in create-listing-modal.gts and the ternary in panel-content.gts.

packages/runtime-common/realm-index-query-engine.ts
```diff
@@ -271,20 +272,27 @@ export class RealmIndexQueryEngine {
   async searchEntries(
     searchEntryQuery: SearchEntryQuery,
     opts?: Options,
+    // Internal scope (not a wire member): membership/list queries search both
+    // kinds in one query ('all', the default) — a card row and a file row
+    // carry non-overlapping `types`, so the caller's filter alone
+    // discriminates kind. The single-URL instance GET pins 'instance' so a
+    // bare file URL never resolves as a card entry.
+    entryTypeScope: 'instance' | 'all' = 'all',
```
habdelra: [Claude Code 🤖] This parameter is the right abstraction — my suggestion is to let wire callers use it too, rather than keeping it internal-only.

The comment says "not a wire member" as if that's a feature, but look at the asymmetry it creates: `cardDocument` pins `'instance'` by passing an argument, while the host-mode head prefetch — which wants *exactly the same pin* for its single-URL query — has to express it as `eq: { 'item._isCardInstance': null }`, which only works because an indexing pass stamped a synthetic key on the row it wants to drop. Same intent, two mechanisms, and the wire one is coupled to index-stamp state (the PR's own deployment section documents the "duplicates until reindexed" window that coupling buys).

A wire `scope: 'cards' | 'files' | 'all'` (default `'all'`, preserving this PR's semantics and its old-client compat story) would give consumers a way to state kind selection that:

- pins `i.type` directly, so it works for **any** filter shape — no interaction with the shared `types` cross-join alias, none of the anchor-composability heuristics in `anchorQueryToCardInstances`/`scopeFilters`, no `any:`-branch leak;
- doesn't silently drop errored instances the way a `{ type: baseCardRef }` anchor does (error rows can lack `types`; `i.type` is always present);
- makes the `_isCardInstance` dedup needed only where a query is *genuinely* mixed (the search sheet), instead of being the idiom every cards-only consumer must remember.

Compat cost is the same as this PR already accepts: an old server would reject the unknown member just as it rejects the unknown `_isCardInstance` field, and the same host-after-server deploy order covers both. This also directly answers the "what's the wire format for 'all'?" question from the PR discussion — the mode becomes visible in the request instead of being an ambient default.

packages/runtime-common/realm-index-query-engine.ts
```diff
@@ -317,6 +326,82 @@ export class RealmIndexQueryEngine {
     let fullItemRoots: (CardResource<Saved> | FileMetaResource)[] = [];
 
     for (let row of results) {
+      // A `file` row (mixed 'all' scope) renders natively and carries no
+      // ancestor coercion — its renderings hang off its own type's entry with
+      // no renderTypeKey, and its resource is the synthesized `file-meta`. This
+      // is the file counterpart of the instance branch below; kind is decided
+      // per row on `row.type`.
+      if (row.type === 'file') {
```
habdelra, line 355: [Claude Code 🤖] This branch is a near-verbatim copy of the per-file loop body in `searchEntriesFileMeta` (icon collection → `enumerateFileRenderings` + htmlQuery match → scoped-CSS attach → `buildHtmlResource` loop → `itemAsFallback` gating → `fileResourceFromIndex` → sparse/full item split → `buildEntryResource`). Two copies of that assembly will drift the first time one of them learns something new — the css-attach rule and the icon-dedup rule are exactly the kind of details that get fixed in one place.

Two ways to converge, either fine as a fast-follow:

1. Extract the loop body into a shared `assembleFileRow(file, fieldset, resolvedHtmlQuery, accumulators)` helper both paths call.
2. Go one step further in the direction this PR already points: `_search` supports `'file'` as an entry type, so `searchEntriesFileMeta` could become `searchEntries(searchEntryQuery, opts, 'file')` once the unified loop handles file rows — which, after this PR, it does. That would leave exactly one assembly path for all three scopes, with `searchFiles`/`fileEntryFromResult` remaining for the non-entry callers.

packages/host/tests/unit/anchor-query-to-card-instances-test.ts
```diff
@@ -0,0 +1,68 @@
...
+    let query: Query = { filter: { type: authorRef } };
+    assert.deepEqual(anchorQueryToCardInstances(query), query);
+  });
+
+  test('a filter with a positive `on` scoping is left untouched', function (assert) {
+    let query: Query = { filter: { on: authorRef, eq: { status: 'ready' } } };
+    assert.deepEqual(anchorQueryToCardInstances(query), query);
+  });
+
+  test('a negated-only type ref carries no positive anchor, so it is anchored', function (assert) {
```
habdelra: [Claude Code 🤖] Heads-up on what this composed filter actually does at the SQL level, because the query *shape* asserted here doesn't deliver the semantics its name implies. `every: [{type: baseCardRef}, {not: {type: specRef}}]` compiles both type conditions against the one shared `jsonb_array_elements_text(types)` alias, and `GROUP BY` keeps any url with one passing element — so a Spec instance (types `[Spec, CardDef, BaseDef]`) survives via its `CardDef` element, which satisfies `e = 'CardDef' AND NOT (e = 'Spec')`. I verified this fan-out behavior directly against Postgres (details in my comment on `anchorQueryToCardInstances`).

Not a regression — the unanchored `not: {type: specRef}` had the same leak before this PR — and this unit test is still right to pin the helper's output shape. But it's worth (a) an engine-level test that asserts the *result set* for this composition, marked/skipped as documenting the known shared-alias limitation, and (b) not letting this test's existence read as confirmation that the anchored filter excludes Specs. The real fix is making type conditions self-contained membership predicates (per the other thread); that follow-up can flip the engine test to its intended expectation.

habdelra (commented):
[Claude Code 🤖] Full review, focused on how this lands against the query engine's architecture. TL;DR: **the engine-level mechanics are right and I'd have built the `_search` layer the same way; the feedback is concentrated on the application layer — where kind-selection intent lives, and one engine limitation this PR works around that we should fix instead.**

## What lands right

The core design respects the engine's underpinnings precisely where it matters:

- **Mixed scope belongs in `_search`, expressed as row-kind gating** (`i.type IN` via the OR'd branches), not as filter rewriting. The `instance`/`file` branches keep their distinct error semantics, and the shared `everyCondition` means the data and count queries can't disagree.
- **`GROUP BY (url, type)`** is the correct grouping once a dual-indexed card `.json` puts two rows behind one url — and the count rewrite to `COUNT(*)` over a `DISTINCT (url, type)` subquery correctly accounts for SQLite's single-expression `COUNT(DISTINCT …)` limit. The `type` tiebreaker added to every ORDER BY path (including the wrapped-projection variant) keeps pagination deterministic. This is exactly the discipline the two-adapter constraint demands.
- **Kind discrimination through the filter is sound** because of how the search doc works: a card-only field filter requires an `on:` type ref to resolve (which brings a `types` condition that file rows can't match), and file rows simply lack card keys. The self-narrowing is a property of the schema walk, not a heuristic.
- **Deleting the `queryTargetsFileMeta` all-or-nothing routing** is a genuine simplification — a FileDef-typed filter now works through the same path as everything else.
- The deployment/compat analysis in the description is thorough and honest about the hard direction.

## On restricting mixed queries to `matches` (discussion follow-up)

I'd argue against a 400 for non-`matches` filters in mixed scope. The two kinds *do* share a queryable surface — it's just synthetic: `_title` (contains + sort), `matches` (GIN full-text), `_cardType`, and now the dedup key. The search sheet's own mixed query is `every: [{not: {type: spec}}, {any: [{matches}, {contains: {_title}}]}, dedup]` — a strict matches-only rule would reject it. The synthetic-key set effectively **is the schema of a mixed search**; rather than policing filters, I'd document that contract (a short section wherever the entry API wire format is documented: "these are the keys both row kinds carry; anything else narrows to one kind by construction").

## On the `baseDef` type-anchor idea (discussion follow-up)

Conceptually attractive — one mechanism for everything — but three mechanical facts work against it today:

1. **BaseDef's key is in neither kind's `types` array.** `getTypes` (host `render/meta.ts`) explicitly `break`s at `baseRef`, so instance chains end at `CardDef`; file rows carry `[<file subtype>, FileDef]` (`file-indexer.ts`). A `{type: baseDef}` filter would match zero rows. Making it work needs an indexer change plus a full reindex of every realm.
2. **Type conditions don't compose.** All type refs in a query share one `jsonb_array_elements_text(types)` cross-join alias, so `every: [{type: baseDef}, {not: {type: spec}}]` doesn't mean what it reads as (verified against Postgres — see the `anchorQueryToCardInstances` thread). A baseDef anchor would immediately collide with this.
3. **Rows with empty `types` vanish** whenever the filter references the types alias — errored instances would silently drop out of "everything" queries.

So: if we want `baseDef`-as-universe eventually, the typeCondition composability fix below is a prerequisite anyway. For now, an explicit wire scope (or the PR's ambient default) is mechanically cleaner.

## Recommendations (details in the inline threads)

1. **Expose the scope on the wire** (`scope: 'cards' | 'files' | 'all'`, default `'all'`). The PR already models scope internally (`entryTypeScope`) and pins it for `cardDocument`; wire callers deserve the same verb instead of filter incantations. It pins `i.type` directly — no shared-alias heuristics, no dependence on index-stamp state, doesn't drop errored instances the way a `types` anchor does, and it answers "what's the wire format for `'all'`?" by making the mode visible in the request. Compat cost is identical to what the PR already accepts for the new synthetic field.
2. **Centralize the dedup idiom** behind one named runtime-common export (and its wire spelling) instead of five scattered `_isCardInstance: null` literals — which also makes the agreed `_isCardSourceFile` rename a one-file change. The rename's absent-as-`false` semantics need to live in `fieldEqFilter` (polarity-aware), not the `getField` shim — sketch in the inline comment.
3. **File a follow-up: make `typeCondition` a self-contained membership predicate** (pg `types @> '["key"]'::jsonb`, sqlite `EXISTS (SELECT 1 FROM json_each(types) WHERE value = ?)` — a resolved per-adapter node, same pattern as `JsonContains`). This fixes a latent pre-existing bug (`not: {type: spec}` does not actually exclude Specs — verified), makes positive type conditions AND-composable, and lets `anchorQueryToCardInstances` drop its skip-heuristic and documented `any:` leak.
4. **Consider hoisting the cards-only anchor into `store.search`** — its doc comment already promises "instances only," and the helper's skip-on-positive-type behavior means typed (card *or* file) queries pass through untouched. One choke point instead of per-call-site patches, and future `store.search` callers can't forget.
5. Smaller items inline: trim the unused `markdown`/`isolated_html` from `fileColumns` (potentially large per-row payloads the mixed assembly never reads), and converge the duplicated file-row assembly between the new mixed branch and `searchEntriesFileMeta`.

## Tests

The engine tests are well-aimed (mixed default, NULL dedup, positive-anchor narrowing, and the `_title` sort-direction guard is a nice touch). Two gaps: an `'all'` + `includeErrors` test (the asymmetry — errored instances in, errored files out — is load-bearing and currently only documented in a comment), and an engine-level result test for the anchored not-spec composition, which belongs with the follow-up in (3).

None of this blocks the PR conceptually — (1) and (2) are worth doing before merge since they shape the wire surface consumers will start depending on; (3)–(5) work fine as fast-follows.

habdelra:
(lol, these fable code reviews are _fantastic_ :star_struck: )

packages/host/app/components/card-search/constants.ts
```diff
@@ -55,7 +55,12 @@ export const SORT_OPTIONS: SortOption[] = [
           module: baseRRI('card-api'),
           name: 'CardDef',
         },
-        by: 'cardTitle',
+        // The sheet is mixed cards + files, so sort on the synthetic `_title`
+        // key that both row types carry (a card's title — mirrored from
+        // `cardTitle` at index time — and a file's name). Sorting on
+        // `cardTitle` alone would leave file rows NULL and sink them all below
+        // the cards under NULLS LAST.
+        by: '_title',
```
habdelra, line 63: [Claude Code 🤖] The server side of this sort is right, but the client-side mirror doesn't know `_title` exists. `makeInstanceComparator` → `sortValue` (runtime-common/instance-filter-matcher.ts) resolves the sort path through `resolvePath`, which looks `_title` up as a real card field, finds nothing, and yields `null` for every instance — so under NULLS-LAST everything ties and falls through to the URL tiebreak. Wherever the live reconciliation path re-sorts a merged result set locally (`SearchResource.displayedInstances` applies the local comparator to the *entire* displayed set), an A-Z `_title` sort silently becomes URL order, overriding the server's correct ordering. The pre-PR spelling `cardTitle` resolved as a real field, so this is a new divergence.

See my review-level comment on local search parity — the matcher and comparator need the same synthetic-key shim the engine's `getField` has, and this sort is one of the three idioms in this PR that currently fall through it.

habdelra (commented):
[Claude Code 🤖] Follow-up review on **local search parity** — the client-side counterpart of the query engine. The store runs live queries against in-memory instances for immediate response and reconciles the server's results as they arrive (`SearchResource.displayedInstances` in `host/app/resources/search.ts`, backed by `runtime-common/instance-filter-matcher.ts`, whose header promises it is "kept faithful to the server"). **This PR teaches the server three new query idioms without teaching the client matcher any of them, and the failure mode is not "degraded instant results" — it's the reconciler actively removing correct server results.**

## The mechanics

`displayedInstances` keeps a server-returned instance only while it doesn't *demonstrably* fail the filter locally: `'no-match'` removes it; `'unresolvable'` keeps it. The matcher resolves a filter path via `resolvePath`, which looks fields up on the hydrated instance. A synthetic search-doc key (`_title`, `_cardType`, `_isCardInstance`) is not a field on any instance, so the path contributes **no values** and — critically — `sawUnresolvable` stays false, so the predicate returns `'no-match'`, not `'unresolvable'`. The engine's `getField` has a shim that fabricates definitions for these keys; the matcher has no counterpart.

## The three divergences this PR introduces

1. **`eq: { _isCardInstance: null }` blanks a reconciled result set.** Server: key absent from every instance row → `IS NULL` → matches all cards. Client: no such field → `'no-match'` for every instance → every server result is filtered out of the display and no local candidate can be added. A live `SearchResource` query using this PR's own documented dedup recipe renders **zero rows** while holding a correct server response.
2. **`contains: { _title: term }` does the same** for recents-style term filters. The pre-PR spelling `cardTitle` was a real field and evaluated correctly locally; `_title` resolves to nothing → `'no-match'` for every instance.
3. **`sort: [{ by: '_title', on: CardDef }]` degrades to URL order locally.** `sortValue` yields `null` for every instance → NULLS-LAST ties → id tiebreak — and `displayedInstances` re-sorts the *merged* set with the local comparator, overriding the server's correct ordering. (Inline comment on `constants.ts`.)

Why tests haven't caught it: today's consumers of these idioms (search sheet, recents) go over the **entry API** resource, which doesn't instance-match — so the breakage is latent there. But `getSearchResource` is precisely the surface handed to **cards** via `@context.store`, and the dedup filter is now the documented recipe for mixed-safe queries; the first card author who follows it gets a blank live query. There's also an amusing proof of how idiom-dependent this is: cards-grid's `not: { eq: { _cardType: null } }` accidentally agrees with the server locally (inner `'no-match'` negates to `'match'`), while the positive `eq: { _isCardInstance: null }` — same synthetic-key mechanics, opposite polarity — blanks the set.

## Proposed approach

Mirror the engine's own pattern: the engine centralizes synthetic-key knowledge in its `getField` shim; give the matcher the exact counterpart, in the same file the engine documents them, so the next synthetic key updates both sides or neither.

1. **A synthetic-key resolution step ahead of `resolvePath`** (used by `matchEq`/`matchIn`/`matchContains` and `sortValue`), returning the value both sides agree the key denotes:
   - `_title` → the instance's title (`cardTitle` for cards — the same value the index stamps; a FileDef's name).
   - `_cardType` → the card type's display name (what the meta route stamps).
   - `_isCardInstance` → for a hydrated `CardDef`, always `null` (the stamp exists only on a card `.json`'s **file** row, which never hydrates as a CardDef) — so `eq: null` matches, restoring parity. For a hydrated `FileDef` the stamp isn't derivable from the resource (it lives in the search doc, not the file's attributes), so return `'unresolvable'` — which the reconciler already treats as "trust the server."
2. **A conservative guard in `isClientEvaluable`**: a filter path whose head is an *unshimmed* underscore-prefixed key should make the query not client-evaluable (server-only passthrough). That turns "someone added a synthetic key on the server and forgot the matcher" from silently-blanked results into merely no-instant-results — the failure mode you want by default.
3. **`makeInstanceComparator.sortValue`** takes the same shim for `_title` (and `_cardType`) sorts.
4. When the agreed `_isCardSourceFile` rename lands, its matcher shim should ship **in the same commit** as the engine-side `fieldEqFilter` rewrite — the two files are one contract, and a cross-reference comment between the engine's `getField` shim and the matcher would make that hard to forget.

One adjacent structural note for a follow-up rather than this PR: `displayedInstances` dispatches a single candidate pool (cards *or* file-meta) by sniffing the filter's top-level type ref or the first server row. Once mixed result sets are the default, a mixed live query can only reconcile one kind, and first-row sniffing becomes order-dependent. Fine to defer — but worth capturing in the same ticket that tracks mixed-search parity, since the sheet will presumably want live file results eventually.

If it's useful I can sketch the shim as a concrete diff — it's small and self-contained (`instance-filter-matcher.ts` + one guard in `isClientEvaluable`).

habdelra (changes requested):
This PR actually breaks the local search parity in a destructive way. see the followup review comment for the details.

habdelra:
FYI, if it helps I used this skill to help guide this code review https://github.com/cardstack/boxel/pull/5467

FadhlanR:
[Claude Code 🤖] Rebased onto latest main and pushed a revised implementation addressing the review feedback. Summary of what changed since the last review:

**Engine — type conditions now compose (root-cause fix, not a workaround)**
Replaced the shared \`jsonb_array_elements_text(types)\` cross-join alias with a self-contained \`TypesContains\` membership predicate (Postgres \`@>\`, SQLite \`json_each\` EXISTS, \`COALESCE\` so NULL/absent \`types\` are a definite FALSE). Verified on both adapters: \`not:{type:X}\` genuinely excludes X, \`every:[{type:A},{type:B}]\` is a satisfiable intersection, and rows with empty \`types\` survive negation. This removes the need for the anchor heuristic and its documented \`any:\` leak — \`anchorQueryToCardInstances\` is deleted.

**Wire \`scope\` member**
Kind selection is now a first-class wire concept: \`scope: 'cards' | 'files' | 'all'\` (default \`'all'\`) pins \`boxel_index.type\` directly, independent of filter shape. It answers "what's the wire format for 'all'?" by making the mode explicit, doesn't drop errored instances the way a \`{type}\` anchor does, and is immune to the un-restamped-rows window. Consumers (store.search, host-mode head prefetch, create-listing, recents, choosers) pin scope instead of the \`_isCardInstance\` incantation; the federated-search cache key includes it. \`scope: 'all'\` returns both rows of a dual-indexed card; deduping stays an explicit filter.

**Client search parity (the CHANGES_REQUESTED blocker)**
The matcher now understands the synthetic search-doc keys via a shim at the top of \`resolvePath\` (covers all match predicates and the sort comparator), plus an \`isClientEvaluable\` guard so an *unshimmed* underscore key becomes server-only rather than blanking reconciled results. \`friendlyCardType\` moved to runtime-common so the index stamp and the matcher share one implementation.

**Key rename + BaseDef in the type chain**
\`_isCardInstance\` → \`_isCardInstanceFile\`, canonical spelling \`eq: false\` (compiles to an adapter-agnostic existence test, fixing the old \`eq:true\` pg/sqlite divergence). One home in \`search-doc-keys.ts\` + an \`excludeCardInstanceFileRows()\` helper. BaseDef now terminates both instance and file type chains, so \`{ type: baseRef }\` spans both kinds and composes.

**Smaller items**
cardsOnly choosers hide file types in the type picker; trimmed the unread \`markdown\`/\`isolated_html\` from the mixed file projection; folded \`searchEntriesFileMeta\` into \`searchEntries(..., 'file')\`; documented \`scope\` + the mixed-search shared-key contract in \`docs/search.md\`.

**Tests (green locally on both adapters)**
Postgres \`search-entries-engine\` 29/29 and \`indexing\` 61/61; SQLite engine 79 tests / 168 assertions; matcher parity 32/78; query-builder 17/21; index-writer 52/142. New coverage for the composition fix (not-spec exclusion, intersection), BaseDef, the three scopes, \`_isCardInstanceFile\` true/false, and the matcher synthetic-key shim.

**Deployment note:** requires a full reindex (stamp rename + BaseDef in \`types\`). There is no index-version knob and the deploy auto-reindex only fires on a boxel-ui dist checksum change, so trigger it explicitly post-deploy (\`_grafana-full-reindex\` or \`REALM_SERVER_FULL_INDEX_ON_STARTUP=true\`). Host-after-server deploy order (an old server 400s on the new \`scope\` member).

docs/search.md
```diff
@@ -229,3 +229,32 @@ let response = await request
...
+searchEntryWireQueryFromQuery(query, { fields: ['item'], scope: 'cards' });
+```
+
+### The mixed-search key contract
+
+Cards and files carry non-overlapping fields, so a filter that names a card field narrows to cards by construction, and a FileDef-typed filter narrows to files. A mixed (`scope: 'all'`) query can filter and sort across both kinds only through the keys both row kinds carry:
+
+- `matches` — full-text search over both kinds' content.
+- `_title` — the row's display title (a card's `cardTitle`, a file's name); usable in `contains` and `sort`.
+- `_cardType` — the card type's display name.
```
habdelra: [Claude Code 🤖] One correction and one addition to this key-contract list — worth getting exactly right since this section is now the canonical answer to "how do I query across both kinds":

1. **`_cardType` doesn't belong under "keys both row kinds carry"** — it's stamped only during card render (`render/meta.ts`); file rows never get it (that asymmetry is exactly what makes cards-grid's `not: { eq: { _cardType: null } }` a cards-only filter). Listing it here invites someone to sort a mixed result set on `_cardType` and wonder why every file sinks to the NULLS-LAST tail. I'd move it out of this list and describe it as a card-only synthetic — useful in mixed queries precisely *because* files lack it.

2. **A sentence on when `{ type: <BaseDef> }` beats `scope: 'all'`** would head off the "we now have two spellings for everything" confusion: `scope` is query-global (it pins `boxel_index.type` for the whole request), while a BaseDef anchor is a *filter node* — composable inside the tree, e.g. one branch of an `any:` that must span both kinds while a sibling branch is card-typed, or an intersection with another type condition. If a caller just wants "everything", `scope: 'all'` is the spelling; the BaseDef anchor is for kind-spanning *within* a filter composition.

packages/host/app/services/store.ts
```diff
@@ -1263,8 +1265,26 @@ export default class StoreService extends Service implements StoreInterface {
...
+    // discriminates. An otherwise-unscoped query is pinned to 'cards' so the
+    // common "search for cards" case doesn't surface a card's dual-indexed
+    // `.json` file row (or plain files) — the choke point that replaces the
+    // former per-call-site card anchor, while leaving file/typed searches
+    // (e.g. SearchResource's file-meta queries) untouched.
+    let typeRefs = query.filter
+      ? getTypeRefsFromFilter(query.filter)
+      : undefined;
+    let hasPositiveType = typeRefs?.some((r) => !r.negated) ?? false;
+    let scope: SearchEntryScope | undefined = hasPositiveType
```
habdelra, line 1294: [Claude Code 🤖] Making `store.search` the choke point is the right structure — every caller (search-cards tool, preselected-card query, and anything future) now inherits the pin without per-call-site edits. Two exceptions have crept into the premise "a positive type ref already selects a kind", though, both introduced elsewhere in this same PR:

1. **`{ type: baseRef }` is now a positive ref that selects *both* kinds** (BaseDef terminates both chains as of this PR). A `store.search({ filter: { type: baseRef } })` therefore takes the `hasPositiveType` branch, stays `'all'`-scoped, and returns FileDef hydrations *plus both rows of every card `.json`* — probably not what a `store.search` caller ever wants, given the method's "instances only"-flavored contract.
2. **A mixed `any:`** — `any: [{ type: SomeCard }, { contains: { _title: 'x' } }]` counts as positively typed, but its untyped branch still matches file rows in `'all'` scope.

Neither has a live caller today, so this is a comment-tightening ask, not a bug report: the doc comment should say the pass-through is for *concrete* card/file type refs, and name these two shapes as the known gaps. If you want to close gap 1 cheaply, treat a baseRef-equivalent ref as *not* kind-selecting (pin `'cards'` for it) — `store.search`'s contract arguably wants that regardless.

packages/runtime-common/index-query-engine.ts
```diff
@@ -1308,6 +1334,31 @@ export class IndexQueryEngine {
     onRef: CodeRef,
     polarity: FilterPolarity = 'positive',
   ): CardExpression {
+    if (
+      currentField(key) === CARD_INSTANCE_FILE_KEY &&
```
habdelra, line 1320: [Claude Code 🤖] The existence-test compilation is the cleanest possible resolution of the `null`-vs-`false` ambiguity — `eq: false` now reads as what it does, `IS [NOT] NULL` sidesteps the pg-`'true'`-vs-sqlite-`1` rendering divergence entirely, and polarity inverts for free. Nice.

One precedence subtlety worth a sentence in the comment: this special case keys on the *leaf name* (`currentField(key)`) and fires before any definition lookup, so a card that defines a real boolean field literally named `_isCardInstanceFile` gets existence semantics instead of value comparison (`eq: false` would *fail* to match a card whose field is explicitly `false` — the key is present). That's the opposite precedence from the `getField` shim below, which only fabricates a synthetic definition when the real-field lookup fails. Checking the definition first isn't practical here (this method is sync; lookup is async), and the matcher's `isClientEvaluable` guard already treats every unshimmed `_`-prefixed key as server-only — so the system has de facto reserved `_`-prefixed top-level keys for synthetics. I'd just make that reservation explicit: one sentence in `search-doc-keys.ts` (and the docs' key-contract section) saying user-defined fields must not use `_`-prefixed names at the search-doc top level, rather than trying to make this branch definition-aware.

packages/runtime-common/expression.ts
```diff
@@ -94,6 +95,19 @@ export interface JsonContains {
...
 }
 
+// Self-contained membership test: does the JSON array in `column` contain
+// `key`? Rendered per adapter (Postgres `@>`, SQLite `json_each` EXISTS).
+// Unlike a `jsonb_array_elements_text` cross join — which fans a row out into
+// one row per array element and so gives a type condition exists-one-element
+// semantics that miscompose under AND/NOT — this is a single per-row scalar
+// predicate, so type conditions compose correctly (a real `NOT` exclusion, an
+// AND intersection) and no GROUP BY is needed to recollapse the fan-out.
+export interface TypesContains {
```
habdelra, line 99: [Claude Code 🤖] This node is exactly the right shape — per-adapter rendering like `JsonContains`, and the `COALESCE` guard means an empty/never-indexed `types` is a definite FALSE instead of a vanished row, which fixes the third artifact of the old fan-out (rows disappearing whenever any filter branch referenced the alias). The engine tests pinning `not:`-exclusion and `every:`-intersection close out what were previously silent misfires.

Two small follow-ons, neither blocking:

1. **`TableValuedEach` is now consumer-less.** `typeCondition`, `hasInstanceType`, and `hasFileType` were its only emitters; what remains is the node kind, its factory, its `expressionToSql` branch, and — easy to forget — the SQLite `adjustSQL` rewrites that existed solely for it (`jsonb_array_elements_text(` → `json_each(`, the `_array_element` alias fixup, and the `case jsonb_typeof … end` strip). Dead per-adapter rewrite machinery is where bit-rot hides; worth deleting, here or as an immediate follow-up.
2. **Postgres can now index-serve type filters.** The old cross-join could never use an index; `types @> '["key"]'::jsonb` can, given `CREATE INDEX … USING gin (types jsonb_path_ops)` on `boxel_index`(+`_working`). Type-anchored queries are the hottest filter shape in the system, so that migration is probably worth a ticket even if perf is acceptable today.

packages/realm-server/handlers/handle-search.ts
```diff
@@ -127,6 +127,14 @@ export default function handleSearch(opts: {
     if (parsed.cardUrls?.length) {
       cacheKeyOpts.cardUrls = parsed.cardUrls;
     }
+    // `scope` changes which row kinds the response contains, so it must key the
+    // cache — otherwise a `scope: 'cards'` and a `scope: 'all'` request for the
+    // same query would collide on one ETag/body. Fold only when set (absent ==
+    // the default 'all', so keying `undefined` would not fragment but omitting
+    // it keeps the key identical to pre-scope requests).
+    if (parsed.scope) {
```
habdelra: [Claude Code 🤖] Good catch folding `scope` into the cache key — I'd missed that interaction entirely. One residual asymmetry: an *explicit* `scope: 'all'` is truthy, so it keys differently from an absent scope even though the two are semantically identical — same query, two cache entries/ETags. Normalizing the default away keeps the comment's own goal ("absent == the default 'all'") true for both spellings:

```ts
if (parsed.scope && parsed.scope !== 'all') {
  cacheKeyOpts.scope = parsed.scope;
}
```

habdelra (commented):
[Claude Code 🤖] Follow-up review of the reworked revision. **TL;DR: this is an exemplary response to review — every architectural thread from the last pass was adopted and, in several places, improved on. The remaining asks are small: one doc correction, a few comment-tightenings (inline), and one real must-do — the PR description still describes the superseded design and needs a rewrite, including its deployment section.**

## What this revision gets right

I verified each of these against the branch rather than taking the diff's word for it:

- **Kind selection is now a first-class verb.** The wire `scope: 'cards' | 'files' | 'all'` member (validated with a 400 on bad values, folded into the response cache key) replaces both filter-anchor gymnastics and the `queryTargetsFileMeta` routing heuristic. Internal callers and wire callers now speak the same language, and the single-URL GET pins scope by accept-header kind through the same parameter.
- **The dedup finally reads as what it does.** `_isCardInstance: null` → `excludeCardInstanceFileRows()` / `eq: { _isCardInstanceFile: false }`, compiled as an existence test (`IS [NOT] NULL`) — adapter-agnostic, polarity-clean, and `eq: true` gains a sensible meaning (select exactly the card-source `.json` rows, now tested). The rename plus `search-doc-keys.ts` as the single source of the synthetic-key spellings addresses the scattered-literals concern completely.
- **Type conditions now mean what they say.** The `TypesContains` membership node (per-adapter rendering on the `JsonContains` pattern, with the `COALESCE` empty-array guard) fixes all three artifacts of the shared cross-join alias: `not: { type: X }` genuinely excludes, `every` of two type refs is an intersection, and empty-`types` rows no longer vanish. The new engine tests pin each of these — including the not-spec case that was silently broken before.
- **BaseDef as the composable universe anchor** — with the chains extended consistently in all three `getTypes` implementations, and the prerender loop smartly skipping the abstract BaseDef ancestor rendering so the index doesn't bloat.
- **One assembly loop.** `searchEntriesFileMeta` is gone; the `'file'` scope routes through the unified `searchEntries` path, so file-row assembly can't drift.
- **Client parity is closed.** The matcher shims all three synthetic keys (with `friendlyCardType` extracted to runtime-common so the `_cardType` stamp and the shim share one definition), files resolve as `unresolvable` (trust-the-server), and the `isClientEvaluable` underscore-guard turns any *future* unshimmed synthetic key into "no instant results" instead of "reconciler blanks correct results". Tests cover each shim and the guard.
- The projection trim, the boxel-cli scope passthrough (folding in what was a follow-up ticket), the cards-only type-picker filtering, and the `docs/search.md` contract section all landed too.

## On the clarity of "search both kinds"

That was the open design question, and I think this revision answers it well. There's now a layered, unambiguous vocabulary: **`scope`** says which row kinds a query spans (explicit `'all'` for both); **type refs** narrow within that scope (with `{ type: BaseDef }` as the kind-spanning anchor *inside* filter compositions, now that type conditions compose); **`excludeCardInstanceFileRows()`** is the one remaining special case — "each card once" in a mixed result — and its `eq: false` spelling plus the named helper make the intent legible at every call site. Two documentation nits keep that story crisp (inline on `docs/search.md`): `_cardType` is listed as a both-kinds key but is stamped on card rows only, and a sentence on *when* to reach for the BaseDef anchor vs `scope: 'all'` would pre-empt the "two spellings for the same thing" question.

## The must-do: rewrite the PR description

The body still describes the previous design — "no wire member is added", the `_isCardInstance: null` idiom, `anchorQueryToCardInstances`, and a consumer table that no longer matches the code. Since this PR is the reference for how mixed search works, that's worth fixing before merge. In particular the **deployment section needs re-deriving**, because two things in this revision change the index-state story:

1. **The stamp rename**: rows indexed under the old `_isCardInstance` key don't carry `_isCardInstanceFile`, so until a realm reindexes, its card `.json` file rows pass the new dedup filter and appear as duplicates in mixed results.
2. **BaseDef in `types`**: `{ type: baseRef }` matches nothing on rows indexed before this deploy.

Both self-heal through the deploy-triggered full reindex, but the visible window (transient dupes; BaseDef anchors under-matching) is exactly what the operational notes exist to describe — and the current text's "verify a reindex has completed since the stamping deploy" prerequisite is now obsolete (the relevant reindex is the one *this* deploy triggers).

## Small things (inline threads)

`store.search`'s "positive type ref selects a kind" premise now has two exceptions (BaseDef refs, mixed `any:`) worth naming in its comment; the `fieldEqFilter` special case implies a `_`-prefix reservation that should be stated once in `search-doc-keys.ts`; `TableValuedEach` and its SQLite `adjustSQL` rewrites are now dead code; a GIN index on `types` is the natural perf follow-up now that type filters are `@>`-served; and the cache key can normalize explicit `scope: 'all'` to match the absent default.

None of the inline items block merge on their own — the description rewrite is the one thing I'd insist on before this lands.

habdelra (approved):

docs/search.md
FadhlanR: [Claude Code 🤖] Both taken. `_cardType` is out of the shared-key list — it now appears after the list as a card-only synthetic, with the two consequences spelled out: it works as a kind discriminator in mixed queries precisely because files lack it, and it makes a poor mixed sort key (files all sink to the NULLS-LAST tail). Also added the paragraph contrasting query-global `scope` with the filter-composable BaseDef anchor, and a note reserving `_`-prefixed top-level keys for synthetics (from the query-engine thread).

packages/host/app/services/store.ts
FadhlanR: [Claude Code 🤖] Took the cheap close for gap 1: a BaseDef ref no longer counts as kind-selecting — `hasPositiveType` skips refs equal to `baseRef`, so `{ type: baseRef }` pins to `'cards'` like an untyped query, which is what `store.search`'s instances-only contract wants. A new integration test in `store-search-test.gts` pins this (a BaseDef-typed search returns only card instances, no file rows). The doc comment now says the pass-through is for *concrete* type refs and names the mixed-`any:` shape as the known gap.

packages/runtime-common/index-query-engine.ts
FadhlanR: [Claude Code 🤖] Made the reservation explicit rather than making the branch definition-aware: `search-doc-keys.ts` now states that `_`-prefixed top-level search-doc keys are reserved for synthetics (the engine resolves them by leaf name before any field-definition lookup, so a colliding user-defined field would silently get the synthetic's semantics), and the docs' key-contract section carries the same warning.

packages/runtime-common/expression.ts
FadhlanR: [Claude Code 🤖] Deleted the dead node — the `TableValuedEach` interface, `tableValuedEach()` factory, both union entries, the `expressionToSql` render branch, and the `makeExpression` pass-through arm. One correction on the adjustSQL list, though: those rewrites were not exclusive to this node. `index-writer.ts`'s invalidation query still emits raw `jsonb_array_elements_text(i.deps) … deps_array_element` SQL on the SQLite path and depends on the `json_each(` substitution, the `CROSS JOIN LATERAL` downgrade, and the `_array_element` alias fixup — so those stay. Only the `case jsonb_typeof … end` strip was emitted solely by this node, and that one is removed. The GIN index migration is filed as CS-12082.

packages/realm-server/handlers/handle-search.ts
FadhlanR: [Claude Code 🤖] Applied — an explicit `'all'` now folds the same as an absent scope (`parsed.scope && parsed.scope !== 'all'`), so the two spellings share one cache entry/ETag; the comment states the normalization.

PR #5444, opened 2026-07-09
Render the broken-ref visual in the markdown-embed chooser

FadhlanR:
Linear: [CS-11678: Broken card / broken file rendering](https://linear.app/cardstack/issue/CS-11678/broken-card-broken-file-rendering)

## Problem

A `:card[...]` / `:file[...]` reference that no longer resolves (deleted, moved, no permission) rendered the same empty *"Search for a … & preview its format here"* placeholder as a not-yet-picked slot. The chooser's `loadTarget` collapsed "still loading / nothing picked" and "resolution failed" into one `selectedTarget = undefined`, so a broken ref was indistinguishable from an empty pane — in the chooser preview and in the Edit modal.

## Change

The chooser preview pane and the Edit modal now render the same rich broken-ref visual that broken `linksTo` / `linksToMany` field values already use: the placeholder box with a type/file label plus the reveal overlay carrying the URL, a copy button, and error diagnostics.

To reach that component from host with a plain **static import** (rather than a runtime module load), `BrokenLinkTemplate` is relocated from `base` to **`@cardstack/boxel-ui`** — the shared component package both base and host already consume, and whose scoped CSS is already shipped in the host bundle. Its three outside dependencies are decoupled so boxel-ui takes on **no dependency on `runtime-common`** (which would invert the existing `runtime-common → boxel-ui` edge into a cycle):

- `cardTypeName` → a caller-supplied `@typeName` arg (base card sites pass the card type name; the file chooser passes the filename; falls back to `'Card'`).
- `SerializedError` → a local `BrokenLinkErrorDoc` interface listing only the fields the overlay reads (`SerializedError` is a structural superset, so base callers pass one unchanged).
- base `ViewCardFn` → a local `BrokenLinkViewFn`.

### base (visual parity, no behavior change)
The five call sites (`card-api`, `links-to-editor`, `links-to-many-component`) switch to the boxel-ui import and pass `@typeName={{cardTypeName reference}}` — reproducing the label the component previously computed internally, so the shipped `linksTo` / `linksToMany` broken UI is unchanged.

### host chooser (the ticket)
`tab-panel` tracks a `selectedError`, mounts the pane on target-**or**-error, and threads a broken state (url / state / type name / error doc) through `pane` into `MarkdownEmbedPreview`, which renders `BrokenLinkTemplate` in place of the embed. The Edit modal's existing **Remove / Replace** tile is the fix/remove affordance; there is no "Open anyway" in the chooser. The inline document render (`rendered-markdown.gts`) is left as-is.

## Tests

- New boxel-ui **test-app** suite for the relocated component (box + `@typeName`, reveal overlay + copy, not-found vs error) — **passing locally (4/4)**.
- `broken-link-template-test` re-pointed at the boxel-ui import + `@typeName`.
- Broken-state cases added to the chooser suites: `markdown-embed-preview` (404 + error, card vs file label, no resolved embed), `markdown-embed-preview-pane` (broken preview surfaces; CTA still serializes the URL), `markdown-embed-chooser-modal` (Edit-mode broken preload shows the broken preview + Remove/Replace, no empty placeholder).
- `realm-indexing-test` dep snapshots: dropped the two now-stale `@cardstack/base/default-templates/broken-link-template` lines (the `link-off` icon stays — still reachable via the boxel-ui bundle).
- `linksto-broken-link-placeholder-test` / `linksto-many-broken-link-placeholder-test` are the regression guard for the shipped `linksTo` UI (unchanged type-name value).

Verified locally: glint/type-check clean (boxel-ui addon, host, test-app), eslint + template-lint clean, boxel-ui test-app suite green. Host integration tests rely on CI (they need the realm-server dev stack).


packages/boxel-ui/addon/src/components/broken-link/index.gts
```diff
@@ -54,68 +86,70 @@ function parseHttpUrl(url: string): URL | null {
 // single-colour boxel icon can't express. Shared by the reveal trigger and the
 // overlay title; `@size` sets both svg dimensions so each caller can match its
 // context.
-class WarningIcon extends GlimmerComponent<{
-  Element: SVGElement;
+const WarningIcon: TOC<{
```
burieberry: can you get a version of this where the fill and stroke properties are already in the svg? Then put it through image optimizer or compress it, and place it in icons in boxel-ui raw icons

packages/boxel-ui/addon/src/components/broken-link/usage.gts
```diff
@@ -0,0 +1,155 @@
...
+import BrokenLinkTemplate, {
+  type BrokenLinkErrorDoc,
+  type BrokenLinkFormat,
+  type BrokenLinkState,
+} from './index.gts';
+
+export default class BrokenLinkUsage extends Component {
+  @tracked brokenUrl = 'https://example.com/realm/Author/exploded-card-id';
+  @tracked typeName = 'Author';
+  @tracked noun = 'card';
```
burieberry: `noun` is a very vague property name. how about calling `noun` -> `itemType` and calling `typeName` -> `displayName`? If there's no displayName, you can fall back to the capitalized itemType. If there's any way of knowing the itemType from the payload, we could use that too

burieberry (commented):
I made a couple comments. Other than those, can you make sure all css properties properly match the designs: https://app.zeplin.io/project/6a1f141256365a9e08e74147/screen/6a1f14c4aadcc7b409bc2366
I also recommend using our context button for the x on the overlay.

packages/boxel-ui/addon/src/components/broken-link/index.gts
FadhlanR: [Claude Code 🤖] Done in 013c0c5. Added `raw-icons/warning-triangle-filled.svg` with fill/stroke baked in per-path (SVGO drops a root `fill`, so the colour lives on each path), ran it through the icon pipeline (`pnpm rebuild:icons` → SVGO + committed `src/icons/warning-triangle-filled.gts`), and the component imports that instead of the inline SVG. The triangle uses the design amber `#ffa515`.

packages/boxel-ui/addon/src/components/broken-link/usage.gts
FadhlanR: [Claude Code 🤖] Renamed in 013c0c5: `noun`→`itemType` (now typed `'card' | 'file'`) and `typeName`→`displayName`. When `displayName` is absent the label falls back to the capitalized `itemType` ("Card" / "File"). `itemType` is caller-supplied from the payload — `linksTo` sites pass `'card'`; the BFM chooser derives card-vs-file from the ref type.

packages/host/app/components/markdown-embed-chooser/preview/index.gts
```diff
@@ -198,12 +245,24 @@ export default class MarkdownEmbedPreview extends Component<Signature> {
...
-          />
+          {{#if @target}}
+            <Embed
+              @target={{@target}}
+              @format={{this.renderFormat}}
+              @kind={{this.kind}}
+              @sizeStyle={{this.sizeStyle}}
+            />
+          {{else if @brokenUrl}}
+            <BrokenLinkTemplate
```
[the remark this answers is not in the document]
FadhlanR: [Claude Code 🤖] Fixed in 013c0c5. Broken refs now render through a `BrokenEmbed` placement wrapper that mirrors the resolved-embed inline slot, so an inline non-atom broken ref previews `inline-block` (matching how it serializes) instead of a block. The wrapper adds no card-frame and never sets `overflow: hidden`, so the reveal overlay stays unclipped. Added a preview test covering the inline placement.

packages/host/app/components/markdown-embed-chooser/preview/index.gts
```diff
@@ -144,6 +160,37 @@ export default class MarkdownEmbedPreview extends Component<Signature> {
...
+  // The broken template requires a state + error doc; default them so the arg
+  // types stay optional at this boundary (the parent only sets them alongside
+  // `brokenUrl`, but the correlation isn't expressible in the type).
+  private get brokenState(): BrokenLinkState {
+    return this.args.brokenState ?? 'error';
+  }
+
+  private get brokenErrorDoc(): BrokenLinkErrorDoc {
+    return this.args.errorDoc ?? {};
+  }
```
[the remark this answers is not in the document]
FadhlanR: [Claude Code 🤖] Done in 013c0c5 — the fallback returns a module-level frozen `EMPTY_ERROR_DOC` constant instead of a fresh `{}` each render.

FadhlanR:
[Claude Code 🤖] Thanks for the review — all addressed in 013c0c5:

**CSS vs the designs** — went through the broken-ref styling against the Zeplin screen. Most of it already used `--boxel-*` tokens; the real corrections were the atom pill + error-box radii → `--boxel-border-radius-2xs` (5px) and the hairlines → `--boxel-border-color` (≈ the design's `#d1d1d1`, vs the lighter `--boxel-200` before). Panel width/height keep the design's fixed px (350 wide, 155–600 tall). One deliberate departure from tokens: the warning triangle bakes the design amber `#ffa515`, since a two-tone icon can't ride a single colour var.

**Context button** — the overlay × is now `<ContextButton @icon='close'>`.

**Scope note** — the full V2 error panel (scroll / copy-link / placement fallbacks) is a much larger feature than this PR; happy to file a follow-up ticket for it rather than pull it in here.

While verifying the pencil-edit flow live I also found and fixed two bugs (same root cause): editing a card ref opened the chooser with a blank preview and threw a Glimmer backtracking assertion. The document-relative ref (`../Type/id`) wasn't resolved to an absolute URL before `store.get`, so it fast-failed mid-render. It's now resolved first, with a regression test that reproduces the assertion when the fix is reverted.

https://github.com/user-attachments/assets/fb0a4655-c888-4c73-9e58-56bea9f3c1d6

packages/boxel-ui/addon/src/components/broken-link/index.gts
```diff
@@ -695,20 +732,12 @@ export default class BrokenLinkTemplate extends GlimmerComponent<{
...
-        display: inline-flex;
-        align-items: center;
-        justify-content: center;
-        width: 1.25rem;
-        height: 1.25rem;
         margin: -2px -2px 0 0;
-        border-radius: 50%;
+        --boxel-icon-button-width: 1.25rem;
+        --boxel-icon-button-height: 1.25rem;
         color: var(--boxel-400);
```
burieberry: likely don't need these if using size argument. color is black, no need to change

packages/boxel-ui/addon/src/components/broken-link/index.gts
```diff
@@ -373,15 +397,24 @@ export default class BrokenLinkTemplate extends GlimmerComponent<{
...
+                width='16'
+                height='16'
+                aria-hidden='true'
+              />
               <span data-test-broken-link-headline>{{this.headline}}</span>
             </span>
-            <label
-              for={{this.toggleId}}
+            <ContextButton
               class='overlay-close'
```
burieberry, line 409: try `@size='extra-small'` or small

burieberry (approved):
Approved after below changes

packages/boxel-ui/addon/src/components/broken-link/index.gts
FadhlanR: [Claude Code 🤖] Done in 87910378d4. Swapped the hard-coded `@width='13' @height='13'` for `@size='extra-small'` on the close ContextButton — icon now renders at 16px (matching the title warning icon) with the button box at 1.25rem/20px (same as before).

packages/boxel-ui/addon/src/components/broken-link/index.gts
FadhlanR: [Claude Code 🤖] Done in 87910378d4. Dropped the `--boxel-icon-button-width/height` overrides (now covered by `@size`) and the `color: var(--boxel-400)` + `:hover` colour rule from `.overlay-close` — the button keeps the default black with the ghost variant's own hover background. Only `flex: none` + the corner margin remain.

PR #5450, opened 2026-07-09
Fix start-host script reference in README

cmgardella:
## Summary
- The README referenced `scripts/start-host`, but the actual script is `scripts/start-host.sh`.

## Test plan
- [x] Confirmed `scripts/start-host.sh` is the actual filename in the repo.
- [x] Updated both references in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ There exists a "dev" mode in which we can use ember-cli to host the card runtime
 
 ### With staging or production
 
-You can develop the host application locally backed by the staging or production infrastructure by running `scripts/start-host [environment]`.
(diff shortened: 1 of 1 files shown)


richardhjtan (commented):
[Claude Code 🤖] Checked the claim against the tree: every `scripts/` path the README names, and whether the two edited references are the only stale ones.

**No blocking issues — the change is correct and complete.** `scripts/start-host.sh` is the real filename, and after this diff every `scripts/` path in the README resolves to a file that exists.

The ask here is a merge rather than a change: this has sat unreviewed while `main` moved ~2000 commits, and it still reports `MERGEABLE` / `CLEAN` with green checks. A docs one-liner that stays open is a reader hitting `scripts/start-host: No such file or directory` for no reason.

PR #5477, opened 2026-07-13
feat: generate skill tool definitions at index time and serve them via file-meta

jurgenwerk:
This is a development step where we want to have functioning tools coming from .md skill files that are pulled on demand by the model. Once the pulled skill is read in the ai bot, the bot will fetch the indexed tool definition and send it to the model so that it can call the tools defined in pulled skills. 

`SkillFrontmatterField.fromFrontmatter` now generates each declared tool's LLM definition during file extraction and stamps it onto the skill's file-meta, so consumers get ready-to-use tool definitions straight from the index — no module loader needed on the consumer side.

```
skills/my-skill/SKILL.md
  boxel.tools: [{ codeRef, requiresApproval }]
        │
        ▼  file-extract visit (prerender page, full host runtime)
  SkillFrontmatterField.fromFrontmatter:
  resolve codeRef absolute → load tool class → getInputJsonSchema
        │
        ├─ ok ───────► file-meta resource
        │                frontmatter.tools[n] = { codeRef (absolute),
        │                                         functionName,
        │                                         requiresApproval,
        │                                         definition (LLM tool) }
        │
        └─ failure ──► diagnostics.toolSchemaErrors
                         row still indexes; remaining tools still enrich;
                         surfaced by /_indexing-errors and
                         boxel-cli realm indexing-errors

  search_doc ──► keeps tools as authored; definitions never land there
  deps ────────► tool module URLs, so editing a realm-hosted tool
                 module reindexes the referencing skill
```

- Only the indexing path enriches: the file-extract render route hands the extractor an owner-carrying `toolContext`, and without one `fromFrontmatter` skips generation — interactive extract paths are unchanged.
- The enriched value and the failure diagnostics travel out-of-band on symbol channels (`FRONTMATTER_FILE_META_VALUE_SYMBOL`, `FRONTMATTER_DIAGNOSTICS_SYMBOL`), keeping both off the flat `search_doc`.
- `ToolField` gains an optional `definition` JsonField so enriched resources rehydrate host-side.
- Stamping the absolute codeRef also fixes the known relative-module functionName mismatch, since rehydrated tools now recompute from it.

- Deploy side effects to be aware of: legacy Skill cards' `commands` entries now serialize with `definition: null`, so each enabled legacy skill card re-uploads its tool definitions once (the content-hash cache repopulates), and committed skill fixtures gain `definition: null` diffs on next save. Skill rows indexed before this deploys stay un-enriched until their file changes or the realm fully reindexes — plan a reindex when the pull-model consumer starts reading stamped definitions.

Verified locally against the dev stack: after touching a skill markdown file that declares `boxel.tools` (reindex is per-file incremental), its file-meta carries the stamped definitions —

```sh
curl -sk -H "Accept: application/vnd.card.file-meta+json" \
  https://localhost:4201/skills/skills/catalog-listing/SKILL.md \
  | jq '.data.attributes.frontmatter.tools[0]'
```

returns the authored tool entry (just a `codeRef` in the YAML) plus the index-generated additions — resolved `codeRef`, `functionName`, explicit `requiresApproval`, and the ready-to-use `definition`:

```json
{
  "codeRef": {
    "name": "default",
    "module": "@cardstack/catalog/commands/listing-create"
  },
  "definition": {
    "type": "function",
    "function": {
      "name": "listing-create_5cac",
      "parameters": {
        "type": "object",
        "required": [
          "attributes",
          "description"
        ],
        "properties": {
          "attributes": {
            "type": "object",
            "required": [
              "codeRef",
              "targetRealm"
            ],
            "properties": {
              "codeRef": {
                "type": "object",
                "properties": {
                  "name": {
                    "type": "string"
                  },
                  "module": {
                    "type": "string"
                  }
                }
              },
              "openCardIds": {
                "type": "array",
                "items": {
                  "type": "string"
                }
              },
              "targetRealm": {
                "type": "string"
              }
            }
          },
          "description": {
            "type": "string"
          }
        }
      },
      "description": "Create a catalog listing for a card, field, skill, theme, or component"
    }
  },
  "functionName": "listing-create_5cac",
  "requiresApproval": true
}
```


habdelra (approved):

packages/base/tool-field.gts
```diff
@@ -37,6 +38,15 @@ export class ToolField extends FieldDef {
...
   });
 
+  // The tool's ready-to-use LLM tool definition (`{ type: 'function',
+  // function: { name, description, parameters } }`), generated from the tool
+  // class's input schema at indexing time and stamped onto the skill's
+  // file-meta resource. Present only on tools rehydrated from an enriched
+  // index row — a tool authored in frontmatter has no value here until the
+  // file indexes. Consumers that need a schema and find none must generate
+  // it themselves (see the host's `uploadToolDefinitions`).
+  @field tool = contains(JsonField);
```
lukemelia: Since this is `ToolField` it is a little odd to have a field named `tool`. Maybe it should be `schema` or `definitionJson` or something like that?

packages/host/app/routes/render/file-extract.ts
```diff
@@ -98,6 +98,9 @@ export default class RenderFileExtractRoute extends Route<Model> {
       contentHash,
       contentSize,
       buildError: this.#buildError.bind(this),
+      // This route is the indexing path, so skill tool schemas are generated
+      // here and persisted with the row.
+      generateToolSchemas: true,
```
lukemelia: Why is this needed?

lukemelia (changes requested):
Would it be possible to do the tool schema work in `fromFrontmatter` within `SkillFrontmatterField`?

packages/base/tool-field.gts
lukemelia: [Claude Code 🤖] Renamed to `definition` in #5482 — it holds the whole LLM tool definition (`{ type: "function", function: { … } }`), not just the parameters schema, and the `Json` suffix felt redundant given the field is a `JsonField`. Happy to switch to one of your suggested spellings if you prefer.

packages/host/app/routes/render/file-extract.ts
lukemelia: [Claude Code 🤖] Schema generation loads each declared tool's module and instantiates the tool class, so it should only run on the indexing path — the interactive extract paths (room file attach, the store's direct file-meta fallback) shouldn't pay that module-loading cost, and their consumers already generate definitions on demand (`uploadToolDefinitions`). This render route is the indexing path, so it opts in. In #5482 the boolean is gone: the route instead passes the owner-carrying `ToolContext` that tool construction needs, and the presence of that context is what enables the work.

lukemelia:
[Claude Code 🤖] Yes — it works nicely there. #5482 (based on this branch) moves the schema work into `SkillFrontmatterField.fromFrontmatter`: the hook is now async and receives `{ fileURL, toolContext? }`, the skill subclass generates the definitions itself, and the generic extractor no longer inspects skill shapes — it just lifts the enriched frontmatter and any `ToolSchemaError`s off the same out-of-band symbol channels the frontmatter parse error already uses (keeping the multi-KB definitions out of `search_doc`). That PR also renames the `ToolField` field to `definition` per your other comment.

jurgenwerk:
@copilot resolve the merge conflicts in this pull request

lukemelia (approved):

PR #5478, opened 2026-07-13
Fix catalog remix leaving base-realm module refs unresolvable in the install destination

richardhjtan:
## Summary
- `isInBaseRealm` (used by `planInstanceInstall`/`resolveTargetCodeRef` to decide whether a module needs to be copied into the install destination) only recognized a base-realm module via its *resolved real backing URL* (e.g. `https://localhost:4201/base/skill`, via `VirtualNetwork.unresolveURL()`). A module still in its *literal* symbolic `https://cardstack.com/base/` form — which is how base refs actually appear in serialized cards/specs, and in a `VirtualNetwork` with no realm mappings registered — fell through and got wrongly treated as needing to be copied, producing a broken `adoptsFrom.module` relative to the new install folder instead of the actual base realm module. Now checks both forms.
- `live-test.js` already patched `resolvedCatalogRealmURL` for its own bootstrapping loader, but only on a throwaway spread copy of the config module — not the shared module object every test's own app instance imports via `setupApplicationTest`. One-line fix: mutate the shared object directly. No production runtime code (`network.ts`) needs to change at all.

Companion fix in [boxel-catalog#652](https://github.com/cardstack/boxel-catalog/pull/652) addresses the actual `adoptsFrom` rewrite bug this uncovered; both are needed for a catalog remix/install to correctly resolve module references end to end.

## Test plan
- [x] `pnpm lint:types` clean in `packages/runtime-common`
- [x] `Unit | Catalog | Install Plan Builder` suite (13/13 pass, including the base-realm-literal-URL case flagged by Codex review)
- [x] Full live test suite against a real catalog realm: 137 tests, 86 pass, 51 skip (unrelated pre-existing skips), 0 fail
- [x] Verified end-to-end against boxel-catalog#652's own CI (temporarily pinned its checkout to this branch, confirmed both Lint and Live Tests pass, then reverted the pin)

### DIFF
--- a/packages/runtime-common/catalog.ts
+++ b/packages/runtime-common/catalog.ts
@@ -7,7 +7,7 @@ import type { CardDef } from '@cardstack/base/card-api';
 import { RealmPaths, join } from './paths.ts';
 import type { ResolvedCodeRef } from './code-ref.ts';
 import { resolveAdoptedCodeRef } from './code-ref.ts';
-import { baseRealm, baseRealmRRI, realmURL } from './constants.ts';
+import { baseRealmRRI, realmURL } from './constants.ts';
 import { logger } from './log.ts';
 import type { LocalPath } from './paths.ts';
 import { rri } from './realm-identifiers.ts';
@@ -23,23 +23,20 @@ export interface Listing extends CardDef {
   skills: any[];
 }
 
-// A codeRef.module for a base-realm class may show up in either form: the
-// literal symbolic `https://cardstack.com/base/` URL, or the base realm's
-// real backing URL (e.g. `https://localhost:4201/base/skill`).
-// VirtualNetwork.unresolveURL() can canonicalize either form to the
-// `@cardstack/base/` RRI prefix, but only when the base realm's URL and
-// realm mappings are actually registered on that VirtualNetwork instance —
-// a bare VirtualNetwork (e.g. the plan-install unit test) leaves both
-// forms unchanged, so the literal-URL check below is still needed.
-// Without it, a base-realm module can get wrongly treated as something
-// that needs to be copied into the install destination.
+// A codeRef.module for a base-realm class may reach the planner as the
+// literal symbolic `https://cardstack.com/base/` URL or as the base realm's
+// real backing URL (e.g. `https://localhost:4201/base/skill`), so
+// canonicalize once to RRI form and compare against the `@cardstack/base/`
+// prefix. This requires the base realm's URL and realm mappings to be
+// registered on the VirtualNetwork the caller hands us — every production
+// VirtualNetwork registers them at construction (see the host's network
+// service), and tests must do the same. The durable end-state is for refs
+// to already be canonical before they reach this layer, at which point the
+// unresolveURL call collapses to a no-op prefix check.
 function isInBaseRealm(
   module: RealmResourceIdentifier,
   virtualNetwork: VirtualNetwork,
 ): boolean {
-  if (module.startsWith(baseRealm.url)) {
-    return true;
-  }
   return virtualNetwork.unresolveURL(module).startsWith(baseRealmRRI);
 }
 
--- a/packages/runtime-common/catalog.ts
+++ b/packages/runtime-common/catalog.ts
@@ -7,7 +7,7 @@ import type { CardDef } from '@cardstack/base/card-api';
 import { RealmPaths, join } from './paths.ts';
 import type { ResolvedCodeRef } from './code-ref.ts';
 import { resolveAdoptedCodeRef } from './code-ref.ts';
-import { realmURL } from './constants.ts';
+import { baseRealmRRI, realmURL } from './constants.ts';
 import { logger } from './log.ts';
 import type { LocalPath } from './paths.ts';
 import { rri } from './realm-identifiers.ts';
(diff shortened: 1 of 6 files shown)


packages/runtime-common/catalog.ts
```diff
@@ -23,7 +23,20 @@ export interface Listing extends CardDef {
...
+// a real URL back to its canonical RRI-prefix form (`@cardstack/base/skill`)
+// via the registered realm mapping, which is what actually identifies base
+// realm membership post-RRI-refactor. Without this, a base-realm module
+// gets wrongly treated as something that needs to be copied into the
+// install destination.
+function isInBaseRealm(
+  module: RealmResourceIdentifier,
+  virtualNetwork: VirtualNetwork,
+): boolean {
+  return virtualNetwork.unresolveURL(module).startsWith(baseRealmRRI);
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Good catch — fixed in 2ccc65a. `isInBaseRealm` now checks the literal `https://cardstack.com/base/` prefix first, falling back to the RRI-normalized `unresolveURL` check only when that doesn't match. The plan-install unit test with the bare `VirtualNetwork()` (and the base-realm skill install live test) both pass now.

packages/runtime-common/catalog.ts
```diff
@@ -23,7 +23,25 @@ export interface Listing extends CardDef {
...
+// A codeRef.module for a base-realm class may show up in either form:
+// the literal symbolic `https://cardstack.com/base/` URL (e.g. when built
+// directly, without going through a VirtualNetwork with real mappings
+// registered), or the base realm's real backing URL (e.g.
+// `https://localhost:4201/base/skill`) once resolved through one that has
+// them. VirtualNetwork.unresolveURL() only normalizes the latter — it maps
+// a real URL back to its canonical RRI-prefix form (`@cardstack/base/skill`)
+// via the registered realm mapping — so check the literal symbolic form
+// first. Without both, a base-realm module can get wrongly treated as
+// something that needs to be copied into the install destination.
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] You're right, that was overstated — fixed in b6059fb. The comment now says the deciding factor is whether the base realm's URL/realm mappings are registered on that VirtualNetwork instance, not which URL form is passed in.

packages/runtime-common/catalog.ts
```diff
@@ -23,7 +23,25 @@ export interface Listing extends CardDef {
...
+// A codeRef.module for a base-realm class may show up in either form: the
+// literal symbolic `https://cardstack.com/base/` URL, or the base realm's
+// real backing URL (e.g. `https://localhost:4201/base/skill`).
+// VirtualNetwork.unresolveURL() can canonicalize either form to the
+// `@cardstack/base/` RRI prefix, but only when the base realm's URL and
+// realm mappings are actually registered on that VirtualNetwork instance —
+// a bare VirtualNetwork (e.g. the plan-install unit test) leaves both
+// forms unchanged, so the literal-URL check below is still needed.
+// Without it, a base-realm module can get wrongly treated as something
+// that needs to be copied into the install destination.
```
habdelra: hmmmm, this makes me feel like we are papering over a bug. i would expect that the actual materialized base realm URL is only visible in the virtual network layer. they fact that it leaks out is probably not good. i think that might be the bug you need to solve. also this probably touches the base realm RRI work that @backspace is doing

backspace: We have CS-12078 to follow up on this; Claude agrees it’s papering over, but will need a deeper fix

richardhjtan: [Claude Code 🤖] Agreed the multi-form matching was the wrong interim shape — reworked in 0ac8c85. `isInBaseRealm` now canonicalizes the incoming ref once via `unresolveURL` and does a single `@cardstack/base/` prefix check; the literal-URL branch is gone. It only existed for the plan-install unit test's bare `VirtualNetwork()`, which now registers the base realm mapping the same way the host's network service does at construction. This matches the interim guidance in CS-12078 (single canonicalization-at-read until refs arrive canonical); the durable RRI-native planner — resolving in RRI space and confining `toURL` to the fetch boundary — stays tracked there since it depends on the serialization/data-migration work. Verified: plan-install unit tests 13/13, catalog live install+remix suites 9/9 against this branch.

richardhjtan:
[Claude Code 🤖] @habdelra @backspace — following up on this thread to get a decision on how to proceed.

Since your comments, the multi-form matching is gone: 0ac8c85 reworked `isInBaseRealm` to a single `unresolveURL()` canonicalization plus one `@cardstack/base/` prefix check, which is the interim shape CS-12078 describes (single canonicalization-at-read until the migration runs). The literal-URL branch only existed because the plan-install unit test used a bare `VirtualNetwork()`; the test now registers the base-realm mapping the same way the host's network service does.

Locally this is green: plan-install unit tests 13/13, catalog live install + remix suites 9/9. Without the fix on `main`, the card-listing install live test times out and the skill-listing install fails the run.

The question: merge this as the stopgap with CS-12078 as the durable fix, or close it and wait for the RRI-native planner? The dependent piece is boxel-catalog #652 (nine install/remix live tests), plus the remix/install behavior for base-realm-adopting content in the meantime.

Happy to go either way — just need the call so #652 isn't in limbo.

habdelra:
I would prefer to hold on this and fix the root issue. bandaids are a maintenance burden

richardhjtan:
Closing this PR as we lean towards fixing the root cause from the discussion above. Thanks

cc: @habdelra @backspace

PR #5493, opened 2026-07-14
Bridge the markdown-embed chooser through CardContext instead of a global

FadhlanR:
## Background and Goal

Base-realm card code can't statically import host services, so the markdown editor's embed chooser was reached through a `_CARDSTACK_MARKDOWN_EMBED_CHOOSER` global: the chooser modal registered itself on `globalThis` at mount and `runtime-common` exposed `chooseMarkdownEmbed` / `editMarkdownEmbed` accessors that base code called.

This replaces that `globalThis` hop with the existing Glimmer `CardContext` — the same channel already used to hand host capabilities (`searchResultsComponent`, `store`, `getCard`) down to base-realm cards. `operator-mode/container.gts` now provides the `markdown-embed-chooser` service on `CardContext.markdownEmbedChooser`, and `codemirror-editor.gts` consumes it via `@consume(CardContextName)`. The modal no longer touches `globalThis`, and the two `runtime-common` accessor functions are removed. No behavior change to the chooser flow.

**Scope note:** this converts only the markdown-embed chooser. The card and file chooser globals (`_CARDSTACK_CARD_CHOOSER`, `_CARDSTACK_FILE_CHOOSER`) are left unchanged — their logic still lives inside the modal components, so converting them is a larger, separate effort.

- **Reused `CardContext` rather than a new context name.** `codemirror-editor` already consumes `CardContext`, and the chooser is the same kind of host capability as `searchResultsComponent`. The field is optional (`markdownEmbedChooser?`) — contexts with no chooser modal (prerender, freestyle, some tests) simply don't set it, and the toolbar handlers guard on it, preserving the prior "unavailable → warn + no-op" behavior.
- **Propagation through re-providers.** The components that re-provide `CardContext` down the operator-mode tree (`stack-item`, `playground-panel`, `preview-panel`, `spec-preview`) all spread `{ ...this.cardContext, ... }`, so the field reaches `codemirror-editor` on every edit path. A future from-scratch re-provider inside operator mode would need to keep spreading it.


lukemelia (approved):
👍  nice improvement

PR #5497, opened 2026-07-14
Fix flaky matrix head-tags test: gate published-realm navigation on prerender readiness

habdelra:
## Background

Publishing a realm returns before the published realm finishes re-indexing and prerendering, so navigating to a freshly published realm "cold" races that work — the served HTML can still be missing its prerendered head tags.

## The flake

The "the HTML response from a published realm has relevant meta tags" test navigated to the published realm immediately after publishing and asserted on `og:title`. Intermittently that meta element was not yet in the document, surfacing as `element(s) not found`.

## The fix

Gate the navigation on `waitForPublishedMarker` — the same readiness probe the sibling host-mode head-tags test in this file already uses and documents — so the published document is prerendered before the browser asks for it.

## Testing

`packages/matrix` type-checks clean; CI exercises the published-realm flow.

### DIFF
--- a/packages/matrix/tests/head-tags.spec.ts
+++ b/packages/matrix/tests/head-tags.spec.ts
@@ -85,6 +85,13 @@ test.describe('Head tags', () => {
 
     let publishedRealmURLString = `https://${user.username}.localhost:4205/new-workspace/index`;
 
+    // Publishing returns before the published realm finishes re-indexing and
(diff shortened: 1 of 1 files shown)


habdelra:
[Claude Code 🤖] Folding this into #5496, which now carries both the head-tags readiness gate (cherry-picked here) and the workspace-creation boot-index gate. The two flakes share the same root — tests racing realm indexing — so they are landing as one PR.

PR #5499, opened 2026-07-14
Compact rendering for bot-executed readRealmFile tool calls

lukemelia:
Linear: [CS-12080](https://linear.app/cardstack/issue/CS-12080/use-correctness-check-style-compact-rendering-for-skill-read-tool)

Tool calls that ai-bot fulfills itself (`readRealmFile`, marked `executedBy: 'ai-bot'`) are status indicators rather than actionable requests — the host never runs them, so there is no Apply button to present. They now share the correctness checks' compact rendering: a status circle plus a one-line description.

The rows stay distinguishable from correctness checks: the label reads "Read file(s): …", and bot-executed rows appear inline within a regular assistant message (avatar, timestamp, prose), unlike correctness checks, which render as standalone meta-less compact messages.

A compact readRealmFile row with a check circle and 'Read file: pirate-speak/SKILL.md' label, above a full-size host tool call with an Apply button

A burst of readRealmFile calls arrives as separate messages. When such a message carries nothing but bot-executed tool indicators (no prose, reasoning, or attachments) it gets a `bot-tools-only` class, and a headerless (`meta-hidden`) message of that kind stacks flush against a preceding one — headered or not, since the run's first message may carry the timestamp header. The flush rule keys on `meta-hidden`, so it takes full effect together with #5500, which drops timestamp headers within a same-author run:

image

Changes:
- `MessageTool` gains an `isBotExecuted` getter.
- `RoomMessage` computes `@isCompact` per tool: compact when the message is a correctness check (as before) or when the tool is bot-executed.
- `RoomMessage` marks tool-indicator-only messages with `bot-tools-only`; a headerless one gets `margin-top: 0` after another of its kind at the conversation level.
- Integration tests covering the compact bot-executed row, the full-size host tool call, and the `bot-tools-only` marking.


jurgenwerk (approved):
NIce

PR #5510, opened 2026-07-15
Use a Set for O(1) included-dedup in loadLinks (+ fix flaky host-test socket timeout)

habdelra:
This PR carries two independent changes.

## O(1) included-dedup in `loadLinks`

`RealmIndexQueryEngine.loadLinks` assembles the JSON:API `included[]` array for a search response by walking the transitive closure of linked resources. Two dedup checks run inside that walk — whether a relationship's target is already included, and whether a resource is already included — and both were linear scans (`included.find(...)`) over an array that grows to the full closure size, making the walk O(n²) in the number of included resources.

This mirrors each included resource's id into a `Set` and turns both checks into O(1) lookups, keeping the array and the Set in sync as resources are pushed.

## Fix for the flaky host-test socket timeout

Host test shards intermittently fail with a synthetic `Browser timeout exceeded: 60s`, attributed to whichever test ran last, even though every real test passes. testem emits that when the browser's socket.io control channel drops before the suite reports done. On a two-vCPU CI runner the whole service stack and Chrome contend for CPU, so the browser can be too starved to answer a socket.io ping within the 60s testem derives from `browser_disconnect_timeout` — the control socket drops mid-run while the browser is healthy and still finishing the suite.

`socket_heartbeat_timeout` (socket.io's `pingTimeout`) is raised to 300s so a slow-but-progressing browser stays connected. `browser_no_activity_timeout` (120s) still bounds a genuinely stuck browser that stops producing results. testem's control-channel log is also teed to a per-shard CI artifact (`host-testem-debug-*`) so any recurrence is diagnosable from the artifact.

### DIFF
--- a/.github/workflows/ci-host.yaml
+++ b/.github/workflows/ci-host.yaml
@@ -734,6 +734,19 @@ jobs:
           name: testem-log-${{ matrix.shardIndex }}
           path: junit/host-testem.log
           retention-days: 30
+      # testem's internal server log — records the browser↔testem socket
+      # handshake (attach / disconnect / disconnectCount / reconnection-limit
+      # refusal / process exit) that the TAP output omits. It's the only
+      # server-side view of the intermittent post-suite `Browser timeout
+      # exceeded` synthetic failure.
+      - name: Upload testem debug log
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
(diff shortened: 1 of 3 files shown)


habdelra:
ugh--theres a bunch of flaky tests that i'm trying to solve that this is stumbling over. let me land those first

backspace (approved):

PR #5518, opened 2026-07-16
AI Assistant: fix visual bugs when handling tool calls

jurgenwerk:
## Problem

Three visual regressions in the AI assistant's dark panel:

image

1. **Compact tool calls** (bot-executed tools such as `readRealmFile`, shown as "Read files" rows) displayed an unreadable gray blob. The `preparing` state is designed as a full-size "Working…" pill, but compact mode clamps the indicator to a 16px circle with `overflow: hidden`, clipping the pill into the blob.
2. **The full-size "Working…" pill** rendered as an opaque light-gray capsule instead of its dark capsule with an animated gradient ring. The theme's semantic token layer defines light-theme disabled-button tokens (`--boxel-button-disabled-background` etc.); the AI panel is always dark but sits outside the dark theme scope, so the pill's disabled button picked up light-theme colors.
3. **The action bar's "Apply Diff" spinner** rendered near-black on the dark bar. `LoadingIndicator` now colors its icon from its variant system, whose default resolves to the light theme's foreground; the inline icon color also overrides the `--icon-color` the bar used to inherit down (previously teal).

| | Before | After |
|---|---|---|
| Compact in-progress circle | compact before | compact after |
| Working pill | pill before | pill after |
| Apply Diff spinner | spinner before | spinner after |

## Fix

- In compact mode, the `preparing` state renders the same `CircleSpinner` the `applying` state uses, so a streaming tool call reads as an in-progress spinner consistent with the rest of the compact indicators.
- The "Working…" pill styling (width, gradient ring pseudo-elements, dark background) is scoped to `:not(.compact)`, so full-size tool calls keep the pill.
- The pill's disabled button pins `--boxel-button-disabled-background/-foreground/-border` to its own transparent/dark design so the global light-theme disabled tokens can't repaint it, and its label drops a step to 10px.
- The action bar passes `var(--boxel-highlight)` to the Apply Diff spinner explicitly, restoring its teal color.
- The ApplyButton usage page gains an `isCompact` knob, and the ActionBar gets a usage page, so all these states render in `/_freestyle`.

## Testing

- Added an integration test asserting that a streaming compact tool call renders the spinner rather than clipped "Working…" pill text.
- Verified renderings via the `/_freestyle` usage pages (screenshots above).

https://claude.ai/code/session_013o85ejRNCBobEQr1oYFz1T


habdelra (approved):

burieberry:
This seems related to the recent theming related button updates that was not caught in percy. I'll take a look

burieberry (approved):

PR #5521, opened 2026-07-16
Memoize canonicalURL across the index pass

habdelra:
## What

`canonicalURL()` runs once per `(row, dep)` pair while ordering invalidations by dependency, and again per dep during the per-row dependency bookkeeping in the card and file indexers. The same module and relationship deps recur on nearly every card in a realm, so the large majority of these calls resolve the same `(relativeTo, url)` inputs — each one calling `virtualNetwork.resolveURL()` and allocating a transient `URL`.

This adds a pass-scoped `string → string` memo owned by `IndexRunnerDependencyManager`, keyed on `(relativeTo, url)`:

- Shared across invalidation ordering, the relationship-dependency extractor, and index-backed error lookup.
- Cleared at each pass boundary by `reset()`, which already runs at the start of both from-scratch and incremental indexing.
- A hit returns the cached canonical string and skips `resolveURL` entirely, dropping the CPU and transient-allocation cost on the duplicate calls.

The set of registered realm prefixes is stable within a pass, so `(relativeTo, url)` fully determines the result — the memo is a pure cache of `canonicalURL`, not a behavior change. It is bounded by the distinct pairs seen in a pass and dropped when the pass ends.

## How

- `canonicalURL()` gains an optional `memo` parameter; callers that don't pass one behave exactly as before.
- The dependency-normalization helpers forward the optional memo.
- `IndexRunnerDependencyManager` owns the map, passes it to the extractor and the index-backed error lookup, uses it in invalidation ordering, and exposes a memoized `canonicalURL` method that `card-indexer` uses for its direct calls.

## Test plan

- New unit test (`canonical-url-memo-test`) covering: identical result with and without the memo, `resolveURL` skipped on a repeated `(relativeTo, url)` pair, the key distinguishing `relativeTo`, and recomputation after the memo is cleared.
- Existing indexing suites exercise the threaded dependency-ordering and per-row bookkeeping paths.

---
_Generated by [Claude Code](https://claude.ai/code/session_01F9vhaczjPPtkbW3pcQyAFt)_


packages/realm-server/tests/canonical-url-memo-test.ts
habdelra, line 1: there is no need to follow this pattern for tests in runtime common when you only need server test. runSharedTest is supposed to be used for isomorphic tests that run in both the server and the browser. that is not the case here. get rid of the runtime-common test and move to here

habdelra: [Claude Code 🤖] Done — removed the runtime-common shared module and moved the test directly into this file as a server-only QUnit test (stub network and assertions inlined, no `runSharedTest`). Pushed in 018ee3ab.

---

packages/runtime-common/index-runner/dependency-resolver.ts
```diff
@@ -102,7 +121,12 @@ export class IndexRunnerDependencyManager {
       }
       let base = new URL(row.url);
       for (let dep of row.deps ?? []) {
-        let normalized = canonicalURL(dep, base.href, this.#virtualNetwork);
+        let normalized = canonicalURL(
+          dep,
+          base.href,
+          this.#virtualNetwork,
+          this.#canonicalURLMemo,
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch — routed that call site through the manager's own `canonicalURL()` wrapper so all canonicalization inside `IndexRunnerDependencyManager` goes through one entry point. Fixed in 2cfb14f3.

---

lukemelia (approved):

PR #5528, opened 2026-07-16
Fix boxel-ui docs scroll behavior and other improvements

burieberry:
Fixes scroll issues in the boxel-ui Freestyle docs, improves the icons grid, reworks the mobile nav into a proper drawer, and adds theme importing.

## Scroll fixes
- Patch `FreestyleMenu#scrollActiveItemIntoView` so the nav's scroll-spy scrolls only the nav's own container instead of using `scrollIntoView`, which was hijacking the in-progress page scroll whenever the active menu item was off-screen.
- Set `overflow: visible` on the guide's `CardContainer` — its `overflow: hidden` made it the containing scroll box, so the sticky nav sidebar never stuck to the viewport.
- Set `overflow: visible` on `.FreestyleGuide-content` too: its vendor `overflow: auto` made it the containing scroll box for sticky descendants, but the page body is what actually scrolls (the column is never height-constrained), so nothing inside could stick. Also cap it with `min-width: 0; max-width: 100%` so wide content (tables, code) can never push the column past its row.

## Mobile nav drawer
- The nav is now a modal drawer on small screens: closed by default (users land on scrollable content), opened via the hamburger.
- Translucent backdrop that dismisses on tap; selecting a menu entry also closes the drawer. Body scroll-lock applies only while the drawer is open.
- Slides in with a transform transition; full-height `inset` positioning replaces the old `top: 70px` magic numbers; the hamburger reflects state via `aria-expanded`.

## Icons grid improvements
- Added a "Show All" toggle that removes the grid's max-height so the full icon set can be browsed without the inner scroll area (including on mobile, where the capped height is now a `max-height`).
- The grid header (filter, Show All, count) is sticky while scrolling through the icons, with a shadow that appears only while pinned (IntersectionObserver sentinel — no scroll listeners).
- Toggling Show All off scrolls back to the top of the icons section instead of stranding the page deep in collapsed content.
- Replaced the raw filter `<input>` with `BoxelInput` + `FieldContainer`, and swapped hardcoded hex colors for theme variables (`var(--border)`, `color-mix` hover).

## Themes
- Consolidated theme definitions into `tcn-themes.ts` / `cs-themes.ts` barrels and added new themes (Amethyst Haze, Sage Garden, Boxel Brand Guide, …); display names are derived from the export names (`SoftPop` → "Soft Pop").
- New "Import Theme" panel: paste a theme export (e.g. from tweakcn), and only its `:root` / `.dark` variable blocks are kept — `@theme` blocks and resets are stripped. Apply is disabled until a valid block is present; names are de-duplicated.
- Fonts: the static Google Fonts link now covers every family the built-in themes reference, and imported themes get their `--font-*` families loaded dynamically (one link per family, so unknown fonts fail silently to their fallback stacks).

## Other polish
- Theme settings header: flattened the layout into three inline `FieldContainer`s (Theme / Cycle Themes / Dark Mode) and dropped the floating-position media-query CSS.
- Small typography tweaks (font-size tokens for subtitle, descriptions, mobile sizes) and a themed even-row color for the API/CSS-vars tables.


backspace (approved):
I don’t think this is blocking and maybe just a side-effect of the preview deployment but switching between icons and components causes a full refresh for me:

Image

PR #5534, opened 2026-07-17
Persist search sheet state across close/reopen

FadhlanR:
Resolves [CS-12109](https://linear.app/cardstack/issue/CS-12109).

## Problem

Closing the operator-mode search sheet discards the entire search — reopening starts from a blank slate. The `@mode` gate at `search-sheet/index.gts` destroys the whole `SearchPanel → PanelContent → SearchResults → SearchEntriesResource` subtree on close, dropping the query, filters, sort, view, pagination, and results.

## Approach

A new in-memory, session-scoped **`SearchSheetState` service** owns the sheet's inputs (query text, selected types/realms, sort, view toggle, pagination, scroll offset), the **derived query** (`mainQuery`), and the one **`SearchEntriesResource`** (`mainSearch`) itself. Components become pure consumers — nothing is snapshotted or re-seeded.

- The resource is **owned by the service**, so it outlives the `@mode`-gated subtree. On reopen the subtree remounts and rehydrates from the service; because the resource never died, the results are already present (no blank flash, no re-run for an unchanged query).
- `mainQuery` is derived from the **service's own inputs**, not the panel's asynchronously-restored filter state, so a type-filtered reopen never transiently runs unfiltered.
- Once a search is active, the resource **and its realm subscriptions stay live for the whole session** — a plain close does *not* tear them down. Because `mainQuery` is always page-capped, incremental-index / `prerender_html` events in searched realms trigger a background re-run while the sheet is closed. This is a deliberate tradeoff (fresh results on reopen) versus the ongoing subscription cost; the state fully clears on Cancel/Escape, logout/realm reset, or a page reload.
- Persistence is **opt-in** via a `@persist` flag passed only by the search sheet, so the card choosers (which render their own `SearchPanel` instances) keep their own clean, per-invocation state — their `SearchResults` stays component-owned.
- The service registers with the `reset` service, so **logout / realm reset clears it**; it is in-memory only (no `localStorage`), so a **page reload also clears it**.
- `resetState()` runs only on explicit **Cancel / Escape** (not on plain close / blur / result-select), and drives the resource to its dormant branch (idle query → unsubscribe every realm → clear rows/meta). Reopening a persisted search opens straight to the **results view**, gated on the service's `hasActiveSearch` (term **or** type **or** realm) so filter-only searches restore too.

## Files

- `services/search-sheet-state.ts` (new) — the session-scoped store: persisted inputs, derived `mainQuery`, the service-owned `mainSearch` resource, `resetState()`, reset-registered.
- `resources/search-entries.ts` — deferred task start (avoids a mutate-after-consume backtracking assertion when a query is already set at mount) + a two-flag `isLoading` that reports loading across the deferred start.
- `search-sheet/index.gts` — state backed by the service; filters written into the service; reset only on explicit Cancel/Escape; `SearchPanel` remounted on an external trigger (e.g. code mode's "Find instances") so it re-reads the freshly-reset filter state.
- `search/panel.gts` — additive `initialActiveSort` / `onSortChange` / `persist` args (choosers unaffected).
- `search/panel-content.gts` — view + pagination behind `@persist`; main-results `SearchResults` consumes the service's resource.
- `search/search-results.gts` — optional `@resource` arg (host-only); choosers keep the internally-created resource.
- `operator-mode/submode-layout.gts` — reopen to the results view when a search is active.
- `modifiers/persist-scroll-position.ts` (new) — restores the results-list scroll offset on reopen (reveal-gated to avoid a mid-transition jump).

## Verification

`lint:types`, `lint:js`, and `lint:hbs` pass in `packages/host`. Host acceptance + unit coverage in `interact-submode-test.gts` and `search-sheet-state-service-test.ts`: persist across close/reopen (including a filter-only search), no re-run for an unchanged query, scroll restore, Cancel/Escape resets, empty-query reopen to the compact prompt, in-memory (reset/reload) clears, choosers stay clean.


packages/host/app/components/operator-mode/submode-layout.gts
```diff
@@ -346,7 +349,11 @@ export default class SubmodeLayout extends Component<Signature> {
 
   @action private openSearchSheetToPrompt() {
     if (this.searchSheetMode === SearchSheetModes.Closed) {
-      this.searchSheetMode = SearchSheetModes.SearchPrompt;
+      // Reopen straight to the results view when a search is persisted, so the
+      // restored results are shown immediately rather than the compact prompt.
+      this.searchSheetMode = this.searchSheetState.searchKey.trim()
+        ? SearchSheetModes.SearchResults
+        : SearchSheetModes.SearchPrompt;
```
habdelra, line 360: [Claude Code 🤖] **Reopen gate misses filter-only searches — confirms the Codex bot's flag on this line, verified against the code path.**

**The claim.** Reopening a *filter-only* persisted search (a type or realm selected, empty term) sends the sheet to the compact prompt instead of the results view, so the restored results stay hidden until the user changes the search.

**The mechanism.** This gate keys on `this.searchSheetState.searchKey.trim()`, but the service treats a search as active when there's a term **or** a type **or** a realm — see `hasActiveSearch` in `services/search-sheet-state.ts`, which is exactly what drives whether `mainQuery` (and so the resource) produces results. Concrete path: code mode's "Find instances" calls `openSearch('', ref)` (`detail-panel.gts` → `searchForInstances`) → `doExternallyTriggeredSearch('', ref)` in `search-sheet/index.gts` sets `searchKey = ''` and `selectedTypes = [ref]`. A plain click-outside close keeps that state; reopening via the search button lands here → `''.trim()` is falsy → `SearchSheetModes.SearchPrompt`. Prompt ⇒ `isCompact` is true, and in `sheet-results.gts` the main-query sections render only in the non-compact `{{else}}` branch — the `{{#if @isCompact}}` branch shows recents only. The type-filtered rows are live in the resource but never shown.

**The fix.** Gate on the service's own notion of an active search — it already covers term / URL / type / realm in one place, so the check can't drift the way a term-only re-derivation here already has. Make `hasActiveSearch` public (drop `private` on the getter) and:

```suggestion
      // Reopen straight to the results view when a search is persisted, so the
      // restored results are shown immediately rather than the compact prompt.
      this.searchSheetMode = this.searchSheetState.hasActiveSearch
        ? SearchSheetModes.SearchResults
        : SearchSheetModes.SearchPrompt;
```

`hasActiveSearch` is also true for a URL key (it's non-empty), so the URL-paste reopen — which likewise only renders its section in the results view — keeps working; the existing acceptance test for that stays green.

**Scope.** Regression in this PR's new reopen behavior; should-fix here (small, and the single-source gate stops the term-vs-active-search drift from recurring). Not a crasher — blocking is your call.

---

packages/host/app/resources/search-entries.ts
```diff
@@ -318,11 +335,46 @@ export class SearchEntriesResource extends Resource<Args> {
...
+    void Promise.resolve().then(() => {
+      if (isDestroyed(this) || isDestroying(this)) {
+        loaded.fulfill();
+        return;
+      }
+      this.search.perform().then(
+        () => loaded.fulfill(),
+        () => loaded.fulfill(),
+      );
+    });
```
habdelra, line 360: [Claude Code 🤖] **Confirmation — nothing to change; a guard rail for the next editor.**

This deferred task start is load-bearing and correct, and it's the kind of thing a later "just call `perform()` directly" simplification would silently break, so it's worth calling out that it's intentional:

- Starting the task a microtask after `modify()` (instead of a synchronous `perform()`) keeps a `<SearchResults>` that mounts with a query already set — a chooser, the playground, or the sheet reopening restored — from flipping `search.isRunning` inside the render that just consumed it (Glimmer's mutate-after-consume backtracking assertion).
- The paired `isLoading` change (reporting loading across the gap via the untracked `#previousQuery` / `#hasSettled`) is what stops a just-set query from reading momentarily as settled-with-no-results — the failure that made the code-mode playground autogenerate a blank "Untitled" instance.

Verified the settle path is right: `#hasSettled` flips in the task's `finally`, so a *failed* run also settles (no permanent "Searching…"), while `hasCompletedFullRun` stays gated to a successful full run for the partial-refresh precondition — the two-flag split is deliberate. The `Deferred` keeps the prerender load-tracking synchronous so render readiness is unaffected. Good as is.

---

habdelra (commented):
[Claude Code 🤖] **Review — reactivity & lifetime of the service-owned search resource**

This review focused on the two things most likely to bite in this design: the reactivity/lifetime contract of the new session-scoped `SearchEntriesResource`, and whether "reopen restores the search" is complete across *every* way a search becomes active (term, URL, type, realm).

**Bottom line: the architecture is sound and the reactivity handling is careful and correct. One confirmed gap (filter-only searches don't reopen to their results) is worth closing in this PR, and the PR description's "Approach" section now contradicts the shipped behavior and should be reconciled. No crashers; host + realm CI is green.**

**What lands right**
- Moving the query and the `SearchEntriesResource` into the session-scoped service is the right call over the earlier snapshot/re-seed approach — the results survive close/reopen because the resource literally outlives the `@mode`-gated subtree, with no snapshot-vs-live handoff to keep correct. Deriving `mainQuery` from the *service's own* inputs (not the panel's async-restored filter state) is what makes a type-filtered reopen never transiently run unfiltered.
- The deferred task start + two-flag `isLoading` (`#hasSettled` vs `hasCompletedFullRun`) is subtle but correct — details in the inline confirmation on `search-entries.ts`. It's a genuine guard rail worth keeping intact.
- The controlled/uncontrolled split on `PanelContent` (choosers pass none of the service args and keep their ephemeral local state; `SearchResults` branches once at construction on whether `@resource` is present) keeps the blast radius off every other consumer of the search components.

**Recommendations**
1. **[should-fix, this PR]** Reopen gate misses filter-only searches — see the inline thread on `openSearchSheetToPrompt`. Confirms the Codex flag with a verified path ("Find instances" → close → reopen); gate on the service's `hasActiveSearch` rather than `searchKey.trim()`.
2. **[non-blocking, please reconcile]** The description's Approach still reads: *"there is no background work while the sheet is closed (the search resource and its realm subscriptions tear down)."* The shipped service does the opposite, and documents it so in `search-sheet-state.ts`: once a search is active, the resource **and its realm subscriptions stay live for the whole session**, and a plain close does **not** stop them. Because `mainQuery` is always page-capped (`page: { size: SECTION_DISPLAY_LIMIT_FOCUSED }`), the resource never qualifies for the selective or partial-refresh paths (`query.page === undefined` is required for both), so every incremental-index / `prerender_html` event in any searched realm triggers a **full re-run** — ongoing while the sheet is closed, until Cancel/Escape or a page reload. That's a defensible, deliberate tradeoff, but it's a real ongoing-cost change from the old "close tears everything down" model, and a plain click-outside close leaving a live re-running search is easy for a user not to realize. Please align the description with the code and confirm the tradeoff is intended.
3. **[non-blocking, coverage]** Nothing pins the filter-only reopen from (1): the `mainQuery` unit test exercises only term-based derivation (not type-only), and no acceptance test reopens after a type-only "Find instances". A regression test there would lock the fix and the intended results-view reopen.

**Adjacent, out of scope**: I traced the `ScrollToFocusedSection` (scroll-to-0 on a restored `focusedSection`) vs `persistScrollPosition` (restore saved offset) interaction on reopen — the restore's per-frame re-apply burst outlasts the one-shot scroll-to-0, and the scroll-to-0 fires during `restoring === true` so it isn't recorded — so no visible conflict. Noting it only so it's on record; nothing to change.

Submitted as a comment — approve / request-changes is the maintainer's call.

---

FadhlanR:
[Claude Code 🤖] Addressed the review feedback in bf847b705e:

1. **[should-fix] Filter-only reopen** — `openSearchSheetToPrompt` now gates on the service's own `hasActiveSearch` (term OR type OR realm) instead of `searchKey.trim()`, so code mode's "Find instances" (a type filter with an empty term) reopens to its live results rather than the recents-only compact prompt. `hasActiveSearch` dropped `private`; URL-paste reopen is unaffected (a URL key is still a non-empty, active key).
2. **[coverage]** Added a type-only unit case on `hasActiveSearch`/`mainQuery` and an acceptance test that reopens a filter-only search to the results view. The full `search sheet persistence` module (7 tests) and the service unit module (6 tests) pass locally.
3. **[docs]** Reconciled the PR description with the shipped service-owned-resource design — corrected the stale "no background work while closed / subscriptions tear down" claim to the actual behavior (resource + subscriptions stay live for the session; background re-runs on index events are a deliberate tradeoff), and refreshed the Files list.

The confirmation on the deferred task start in `search-entries.ts` needs no change — left intact.

habdelra (approved):

PR #5537, opened 2026-07-17
Factory agent pulls skills on demand instead of front-loading them

jurgenwerk:
The factory's per-issue agent used to receive a resolver-picked set of skills pasted in full into every system prompt, with a token budget that silently truncated whatever didn't fit. It now gets progressive disclosure:

```
resolver ──► load:      workflow skill only ──► full text in system prompt
         └─► suggested: marked in index    ──► agent decides

skill index (one line per skill: name — description)
         └─► read_skill({ skill })            ──► SKILL.md + reference filenames
             read_skill({ skill, reference }) ──► one reference document
```

- Skill sources are exactly two: `.agents/skills-orchestrator/` (factory workflow skills, still front-loaded) and `packages/boxel-cli/plugin/skills/` (the domain library, on demand). The root `.agents/skills/` copies (`boxel-development`, `ember-best-practices`) are no longer read.
- The keyword resolver is demoted to a hint: its picks render as **(suggested)** markers in the index instead of gating what the agent can see.
- Realm-lifecycle skills (`realm-sync`, `file-ops`, `realm-history`, `profile`, `boxel-environment`) are excluded from the index — their write commands would fight the orchestrator's workspace→realm sync — and the index preamble tells the agent not to run realm-mutating `boxel` commands skills mention.
- `--skills-dir <path>` replaces the on-demand library with an operator-curated folder: nothing bundled is indexed or readable, no exclusion filtering, fail-fast at startup if the folder yields no skills. Front-loaded workflow skills are unaffected.

Verified on a staging run: the implementation-issue prompt shrank to one front-loaded skill plus an 18-line index, and the agent made 12 targeted `read_skill` calls (the core skill plus 11 of its 32 references, each pulled right before the work it covers) with all validators green in a single iteration.

### DIFF
--- a/packages/software-factory/.claude/CLAUDE.md
+++ b/packages/software-factory/.claude/CLAUDE.md
@@ -13,6 +13,10 @@ pnpm factory:go --brief-url <url> --target-realm <url>
 - `--debug` — verbose logs.
 - `--agent openrouter` — use the opencode-OpenRouter passthrough agent
   (routes via the realm-server's `/_openrouter/chat/completions` proxy).
+- `--skills-dir <path>` — replace the on-demand skill library with this
+  directory (bundled plugin skills are not read; workflow skills still
+  front-load from the factory package). Fails fast if it yields no
+  skills. Repeatable.
 
 ## Running tests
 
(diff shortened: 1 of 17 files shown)


jurgenwerk:
Close in favor of https://github.com/cardstack/boxel/pull/5549

PR #5539, opened 2026-07-17
Preserve compound field data backed by unexported field defs in server-side writes

habdelra:
Server-side instance writes (POST, PATCH, `/_atomic`) re-serialize the card document through the definition-driven file serializer before it lands on disk. The serializer walks the card's stored `Definition` to normalize each field, and for compound (FieldDef-backed) fields it descends into a child definition looked up from the field's `fieldOrCard` code ref.

The definition store only holds entries for exported classes. A compound field whose declared type is an unexported FieldDef carries a `fieldOf`/`ancestorOf` code ref, which can't be resolved to a definition. The serializer treated that unresolvable lookup as "this field has no fields": every `containsMany` entry serialized to `{}` (array length preserved, attribute data gone), a `contains` value was dropped outright, and any relationship whose dotted path crosses such a field was dropped. Editing such a card in the UI or pushing it through `/_atomic` silently wiped its compound field data on disk.

The serializer now passes values through verbatim whenever a compound field's child definition can't be resolved, and preserves relationships whose dotted path crosses such a field. Schema-aware normalization (relativizing inner code refs, stripping computed fields) is skipped for exactly that subtree — an unnormalized code ref is recoverable, dropped field data is not. Fields the definition doesn't declare at all are still dropped, as before.

The PATCH handler had a second, independent path to the same symptom: it built its merge base from the index's `pristine_doc` rather than the stored file. The index is downstream of the file and can lag it — an in-flight or backlogged index job hands back stale state, and merging a patch over a stale base silently reverts every field the index hasn't caught up on. The merge base is now the stored source file, read inside the same per-realm write lock as the write itself.

Behavioral consequences of the merge-base change: a PATCH against an instance whose file exists but isn't indexed yet now succeeds instead of returning 404; a PATCH against a file that isn't parseable as a card document fails with a clear error instead of overwriting the file with just the patch; and the stored file keeps only the fields the file and the patch actually specify — PATCH no longer materializes the index's schema-defaulted nulls (e.g. an untouched `cardInfo`) into the file, matching how POST and `/_atomic` writes already store documents. GET responses are unaffected since they serialize from the index, which materializes the full field shape.

New realm-server tests cover the three surfaces: PATCH preservation of `containsMany`/`contains` values and nested relationships backed by an unexported FieldDef, PATCH merging against a file the index hasn't caught up with, and `/_atomic?waitForIndex=true` preserving compound entries pushed as raw source.

https://claude.ai/code/session_01BzRJdC6yxAU8oXY42qn4pW

### DIFF
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -2457,6 +2457,8 @@ module(basename(import.meta.filename), function () {
           let cardFile = join(dir.name, 'realm_server_1', 'test', entry);
           assert.ok(existsSync(cardFile), 'card json exists');
           let card = readJSONSync(cardFile);
+          // The stored file carries only the fields the file and the patch
+          // actually specify — unset fields are not materialized into it.
           assert.deepEqual(
             card,
             {
@@ -2464,7 +2466,6 @@ module(basename(import.meta.filename), function () {
                 type: 'card',
                 attributes: {
                   firstName: 'Van Gogh',
-                  cardInfo,
                 },
                 meta: {
                   adoptsFrom: {
@@ -2556,13 +2557,14 @@ module(basename(import.meta.filename), function () {
           let cardFile = join(dir.name, 'realm_server_1', 'test', entry);
           assert.ok(existsSync(cardFile), 'card json exists on disk');
           let card = readJSONSync(cardFile);
+          // Only the values the patch specifies land on disk; unset nested
+          // fields (cardThumbnailURL) are not materialized into the file.
           assert.deepEqual(
             card.data.attributes?.cardInfo,
             {
               name: 'Mango Card',
               notes: 'a friendly dog',
               summary: 'good boy',
-              cardThumbnailURL: null,
             },
             'nested cardInfo values persisted to disk by file-serializer',
           );
@@ -2575,6 +2577,25 @@ module(basename(import.meta.filename), function () {
             'test',
             'person-1.json',
           );
+          // A PATCH stores the file in canonical serialized form (e.g. the
+          // adoptsFrom module ref is written without its executable
+          // extension), so a first PATCH of a hand-authored fixture may
+          // rewrite it once. Prime with one PATCH so the no-op assertions
+          // below measure the steady state.
+          await request
+            .patch('/person-1')
+            .send({
+              data: {
+                type: 'card',
+                meta: {
+                  adoptsFrom: {
+                    module: rri('./person'),
+                    name: 'Person',
+                  },
+                },
+              },
+            })
+            .set('Accept', 'application/vnd.card+json');
           let initialStat = statSync(cardFile);
 
           let initialResponse = await request
@@ -2693,6 +2714,23 @@ module(basename(import.meta.filename), function () {
         });
 
         test('no-op PATCH response carries an ETag matching the existing one', async function (assert) {
+          // Prime once so the stored file is in canonical serialized form;
+          // the no-op assertions below measure the steady state (see the
+          // no-op lastModified test).
+          await request
+            .patch('/person-1')
+            .send({
+              data: {
+                type: 'card',
+                meta: {
+                  adoptsFrom: {
+                    module: rri('./person'),
+                    name: 'Person',
+                  },
+                },
+              },
+            })
+            .set('Accept', 'application/vnd.card+json');
           let initialResponse = await request
             .get('/person-1')
             .set('Accept', 'application/vnd.card+json');
@@ -2722,7 +2760,7 @@ module(basename(import.meta.filename), function () {
           );
         });
 
-        test('patches card when index entry is an error using pristine doc', async function (assert) {
+        test('patches card when index entry is an error', async function (assert) {
           let cardURL = `${testRealmHref}person-1`;
           let errorDoc = {
             message: 'render failed',
@@ -2975,7 +3013,6 @@ module(basename(import.meta.filename), function () {
                   type: 'card',
                   attributes: {
                     firstName: 'Paper',
-                    cardInfo,
                   },
                   relationships: {
                     friend: {
@@ -3511,7 +3548,6 @@ module(basename(import.meta.filename), function () {
                   type: 'card',
                   attributes: {
                     firstName: 'Paper',
-                    cardInfo,
                   },
                   relationships: {
                     friend: {
@@ -3762,7 +3798,6 @@ module(basename(import.meta.filename), function () {
                   type: 'card',
                   attributes: {
                     firstName: 'Paper',
-                    cardInfo,
                   },
                   relationships: {
                     friend: {
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -4000,6 +4000,7 @@ module(basename(import.meta.filename), function () {
                 containsMany,
                 field,
                 linksTo,
+                linksToMany,
                 CardDef,
                 FieldDef,
               } from "@cardstack/base/card-api";
@@ -4011,6 +4012,7 @@ module(basename(import.meta.filename), function () {
                 @field at = contains(DatetimeField);
                 @field headline = contains(StringField);
                 @field author = linksTo(() => Author);
+                @field coauthors = linksToMany(() => Author);
               }
 
               export class Author extends CardDef {
@@ -4105,6 +4107,14 @@ module(basename(import.meta.filename), function () {
                       headline: 'latest entry',
                     },
                   },
+                  relationships: {
+                    // JSON:API to-many `data: [...]` form nested under
+                    // the unexported compound field; must survive as
+                    // indexed `.N` link keys on disk.
+                    'entries.0.coauthors': {
+                      data: [{ id: './author-1', type: 'card' }],
+                    },
+                  },
                   meta: {
                     adoptsFrom: {
                       module: rri('./log'),
@@ -4153,6 +4163,15 @@ module(basename(import.meta.filename), function () {
                 },
               },
               'relationship nested in compound entry persisted to disk',
+            );
+            assert.deepEqual(
+              card.data.relationships?.['entries.0.coauthors.0'],
+              {
+                links: {
+                  self: './author-1',
+                },
+              },
+              'to-many data form nested in compound entry persisted as indexed link keys',
             );
             assert.deepEqual(
               card.data.relationships?.owner,
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -6,7 +6,8 @@ import { join, basename } from 'path';
 import type { RealmHttpServer as Server } from '../server.ts';
 import type { DirResult } from 'tmp';
 import fsExtra from 'fs-extra';
-const { existsSync, readJSONSync, statSync, writeFileSync } = fsExtra;
+const { existsSync, readFileSync, readJSONSync, statSync, writeFileSync } =
+  fsExtra;
 import type {
   Realm,
   Relationship,
@@ -4246,6 +4247,42 @@ module(basename(import.meta.filename), function () {
                 },
               ],
               'compound entries survive a PATCH of an unrelated field',
+            );
+          });
+
+          test('PATCH against a corrupt stored file fails without overwriting it', async function (assert) {
+            let cardFile = join(
+              dir.name,
+              'realm_server_1',
+              'test',
+              'log-1.json',
+            );
+            let corruptContent = 'not a card document {';
+            writeFileSync(cardFile, corruptContent);
+
+            let response = await request
+              .patch('/log-1')
+              .send({
+                data: {
+                  type: 'card',
+                  attributes: {
+                    logTitle: 'Should not land',
+                  },
+                  meta: {
+                    adoptsFrom: {
+                      module: rri('./log'),
+                      name: 'Log',
+                    },
+                  },
+                },
+              })
+              .set('Accept', 'application/vnd.card+json');
+
+            assert.strictEqual(response.status, 500, 'HTTP 500 status');
+            assert.strictEqual(
+              readFileSync(cardFile, 'utf8'),
+              corruptContent,
+              'the stored file is left untouched',
             );
           });
         },
--- a/packages/realm-server/tests/card-endpoints-test.ts
+++ b/packages/realm-server/tests/card-endpoints-test.ts
@@ -3948,6 +3948,254 @@ module(basename(import.meta.filename), function () {
           assert.strictEqual(response.status, 200, 'HTTP 200 status');
         });
       });
+
+      module(
+        'compound fields backed by an unexported FieldDef',
+        function (hooks) {
+          setupPermissionedRealmCached(hooks, {
+            realmURL,
+            permissions: {
+              '*': ['read', 'write'],
+              '@node-test_realm:localhost': ['read', 'realm-owner'],
+            },
+            fileSystem: {
+              'log.gts': `
+              import {
+                contains,
+                containsMany,
+                field,
+                linksTo,
+                CardDef,
+                FieldDef,
+              } from "@cardstack/base/card-api";
+              import StringField from "@cardstack/base/string";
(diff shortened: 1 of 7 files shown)


packages/runtime-common/file-serializer.ts
```diff
@@ -382,6 +391,18 @@ async function processRelationships({
...
 
+    if (fieldDefinition === 'unresolvable') {
+      // The path crosses a compound field whose child definition can't
+      // be resolved (an unexported class). We can't tell what kind of
+      // relationship this is, but we know the card declares the holder
+      // field — preserve the relationship in normalized form rather
+      // than dropping it.
+      result[relationshipKey] = Array.isArray(value)
+        ? value.map((entry) => normalizeRelationship(entry))
+        : normalizeRelationship(value);
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed — `normalizeRelationship` only derives links from a single-resource `data`, so the array form lost its targets here. The unresolvable branch now expands `data: [...]` into the indexed `.N` keys to-many relationships are stored under, one entry per resource identifier, each normalized individually. Covered by a disk assertion in the compound-fields PATCH test.

packages/runtime-common/file-serializer.ts
```diff
@@ -454,7 +481,7 @@ async function resolveDottedFieldDef(
       virtualNetwork,
     );
     if (!next) {
-      return undefined;
+      return 'unresolvable';
     }
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] This is deliberate, and the scope is narrower than it looks. `resolveChildDef` returns undefined only when the segment's declared type has a non-resolved code ref — i.e. an unexported class (`fieldOf`/`ancestorOf` shape). A resolved ref whose definition is genuinely missing makes `lookupDefinition` throw `FilterRefersToNonexistentTypeError`, which fails the whole write loudly rather than reaching the `unresolvable` path. So `owner.someRel` only preserves when `owner`'s target type is itself an unexported class — the same declared-field-with-unschematizable-type situation as the compound case, where dropping the key is silent data loss on a write path that overwrites the stored source. Restricting preservation to contains/containsMany would reinstate exactly that loss for `linksTo(() => UnexportedCard)` holders. Truly unknown paths (a segment the definition doesn't declare) still return undefined and are dropped, so schema filtering of junk keys is unchanged.

packages/runtime-common/file-serializer.ts
```diff
@@ -420,14 +455,20 @@ async function processRelationships({
 // `meta.fields` sub-tree at the root; we walk it in parallel with the
 // path segments so polymorphic per-segment overrides drive the
 // definition lookup.
+//
+// Returns `undefined` when a segment does not exist on its definition
+// (or descends past a primitive) — the path names nothing the card
+// declares. Returns `'unresolvable'` when a segment exists but its
+// child definition can't be resolved (an unexported class), so the
+// caller can preserve the value rather than treat it as unknown.
```
habdelra, line 474: [Claude Code 🤖] **Confirmation — the tri-state boundary is exactly right, plus one contract it silently depends on.**

I traced every way `resolveChildDef` can hand this walk an undefined child (which this function then reports as `'unresolvable'`):

- **A genuinely unknown key never reaches it.** A missing segment fails `getImmediateFieldDef` and returns `undefined` — the caller still drops it, same as before.
- **An unexported class is the only producer.** Its `fieldOrCard` is a `fieldOf`/`ancestorOf` ref, and a polymorphic `meta.fields` override whose `adoptsFrom` is itself such a ref behaves identically (`codeRefWithAbsoluteIdentifier` passes type-refs through, so `isResolvedCodeRef` fails) — both take the preserve path.
- **A resolved ref naming a missing type does NOT take this path.** `CachingDefinitionLookup.lookupDefinition` throws `FilterRefersToNonexistentTypeError` rather than returning undefined (`packages/runtime-common/definition-lookup.ts`, `lookupDefinitionWithContext`), so a deleted/renamed module still fails the whole write loudly instead of quietly preserving stale data.

So `'unresolvable'` means precisely "the card declares this holder field, but its type is not an export the definition store can hold" — the one case where verbatim preservation is the only non-lossy option.

Two properties worth knowing, neither blocking:

1. Segments *past* the unresolvable hop are unverifiable, so a bogus deeper key (`entries.0.notARealField`) is preserved rather than dropped. Right bias — a false preserve costs a stray key; a false drop costs data.
2. The precision leans on `DefinitionLookup.lookupDefinition`'s throwing contract (`Promise<Definition>`, never undefined-for-missing). If an implementation ever returned undefined for a resolved-but-missing type, "type was deleted" would silently reroute into "preserve verbatim". If you want that pinned, a sentence on the `DefinitionLookup` interface is the cheap spot.

Class: confirmation; non-blocking.

packages/runtime-common/file-serializer.ts
```diff
@@ -382,6 +391,32 @@ async function processRelationships({
...
+        // entry per resource identifier, before normalizing each.
+        value.data.forEach((item, index) => {
+          result[`${relationshipKey}.${index}`] = normalizeRelationship({
+            ...(value.meta ? { meta: value.meta } : {}),
+            data: item,
+          });
+        });
+      } else {
+        result[relationshipKey] = normalizeRelationship(value);
+      }
```
habdelra, line 423: [Claude Code 🤖] **The declared-`linksToMany` fan-out later in this function still flattens the exact input shape this branch now expands — pre-existing, not introduced here, but the two branches now disagree.**

This expansion itself is correct: each `data` entry goes through `normalizeRelationship({ data: item })`, whose single-resource branch converts `id` → `links.self`, and the results land under the indexed `.N` keys the file format stores to-many links under. The `entries.0.coauthors.0` assertion pins it.

**The resolvable twin.** Walk the same `{ data: [...] }` value through the declared path below (`fieldDefinition.type === 'linksToMany' && Array.isArray(value.data)`): `normalizeRelationship(value)` runs on the *whole* value first — an array-valued `data` falls through every branch, `delete processedValue.data` erases the identifiers, and the fan-out then stamps `{ links: processedValue.links, meta: … }` (i.e. `links: undefined`) onto every `.N` key. All targets are lost; and when `links` *is* present alongside the array, every entry gets the same single `self`. Net effect today: a to-many `data: [...]` under an unexported compound field survives, while the same relationship on a field the definition **can** resolve is silently emptied.

**Reachability.** `isRelationship` accepts array-valued `data` (its `'type' in data` probes don't fire on arrays), and `promoteLocalIdsToRemoteIds` only rewrites `lid` entries — so this shape reaches the serializer via `/_atomic` raw source and hand-built POST/PATCH bodies.

**The way out** is the block right here: per-item `normalizeRelationship({ ...(value.meta ? { meta: value.meta } : {}), data: item })` in the declared fan-out, replacing the shared `processedValue.links`.

Class: pre-existing bug, made visible by this branch; non-blocking — fine as a follow-up, though it's a ~6-line mirror if you'd rather not ship the two branches disagreeing.

packages/runtime-common/realm.ts
```diff
@@ -4757,32 +4757,42 @@ export class Realm {
...
+        }
+        original = cloneDeep(existingDoc.data) as CardResource;
+      } catch (err: any) {
+        return systemError({
+          requestContext,
+          message: `cannot apply patch, the existing file for ${instanceURL} is not a valid card document: ${err.message}`,
+          additionalError: err,
+          id: instanceURL,
+        });
+      }
```
habdelra, line 4795: [Claude Code 🤖] **Confirmation — the stored-file merge base preserves the concurrent-PATCH guarantee, and the failure triage is complete.** What I verified, for the record:

**Lost-update safety.** `withWriteLock` (a Postgres advisory lock keyed on the realm URL) serializes read→merge→write across replicas. Inside it, `_batchWriteUnlocked` awaits both the adapter write and — because this call site omits `waitForIndex`, taking the default-true branch — the post-write `performIndex()` before the lock releases. The next waiter's `readFileAsText` therefore opens the file only after the previous writer finished it; with close-to-open consistency on the shared filesystem, that read sees the committed bytes. File reads were already load-bearing cross-replica (source GETs, `_batchWriteUnlocked`'s own no-op content check), so no new consistency assumption class is introduced.

**Path correctness.** The PATCH route pattern is `'/.+(?<!.json)'`, so `localPath` never carries the extension and `${localPath}.json` names the right file.

**Error triage.** `isCardResource` requires `meta` with a valid `adoptsFrom` CodeRef, and its `'fields' in meta` probe throws inside this `try` for `meta: null` — so unparseable JSON, non-card JSON, and mangled meta all reach the `systemError` branch. That branch also changes the corrupt-file-with-healthy-index case from "silently clobber the corrupt file with patch-over-index-state" to "500 and leave the bytes for recovery" — the right trade, since whatever the corrupt file held is exactly what a clobber would destroy. A file that exists but was never indexed now merges fine, and the response still comes from the index because the in-lock `performIndex()` has run by then.

**Type guard on disk refs.** `internalKeyFor` resolves the file's relative `adoptsFrom` against the extensionless instance URL; URL-join drops the last path segment either way, so refs written relative to `<instance>.json` resolve identically, and `trimExecutableExtension` makes a fixture's `./person.gts` compare equal to a patch's `./person`. One tightening worth noting: the old index-error fallback could make the type-change guard vacuous (original's `adoptsFrom` defaulted to the patch's own); the guard now always compares against the file's real type.

The conditions under which this stops being safe: the adapter write becoming deferred (returning before durability) or the index flush moving outside the lock — either would let the next waiter read stale state.

Class: confirmation; non-blocking.

packages/runtime-common/realm.ts
```diff
@@ -4757,32 +4757,42 @@ export class Realm {
...
       original.meta ??= { adoptsFrom: patch.meta.adoptsFrom };
       original.meta.adoptsFrom =
         original.meta.adoptsFrom ?? patch.meta.adoptsFrom;
```
habdelra: [Claude Code 🤖] These two fallbacks are dead now that the merge base is guarded by `isCardResource`: it returns false unless `meta` is an object whose `adoptsFrom` is a valid CodeRef, so by this point `original.meta.adoptsFrom` always exists — the `??=` never fires and the `?? patch.meta.adoptsFrom` arm is unreachable. Likewise the `originalClone.meta?.adoptsFrom &&` condition on the type-change guard below is always truthy. Harmless, but they read as though meta-less stored files are supported on this path, which the guard forecloses (those land in the `systemError` branch). Worth deleting for the next reader.

Class: dead code left behind by the merge-base change; nit, non-blocking.

packages/realm-server/tests/card-endpoints-test.ts
```diff
@@ -3948,6 +3983,273 @@ module(basename(import.meta.filename), function () {
...
+              'log-1.json',
+            );
+            // Edit the stored file directly — no realm write, no file
+            // watcher in this fixture — so the index still reflects the
+            // original content. A PATCH of an unrelated field must keep
+            // this edit: the merge base is the stored file, and using the
+            // (lagging) index instead would silently revert it.
+            let onDisk = readJSONSync(cardFile);
+            onDisk.data.attributes.cardInfo = { notes: 'edited on disk' };
+            writeFileSync(cardFile, JSON.stringify(onDisk, null, 2));
```
habdelra, line 4207: [Claude Code 🤖] **The one behavioral consequence without a test pin: PATCH against a corrupt stored file must fail rather than overwrite.** The `systemError` branch in `patchCardInstance` is what prevents an unreadable merge base from degrading into "write just the patch" — but nothing exercises it: the two index-error tests corrupt `boxel_index`, not the file, and this test writes valid JSON. A regression back to overwrite-on-parse-failure would pass the whole suite.

This module already has the exact pattern needed — one more test that `writeFileSync`s garbage (or valid JSON that isn't a card document) into `log-1.json`, PATCHes, and asserts 500 plus byte-identical file content afterward would pin the contract. The sibling behavior — 404 when the `.json` file is missing even if the index still has a row — is worth pinning the same way.

Class: test gap for behavior this PR introduces; non-blocking, but recommended while the machinery is fresh.

habdelra (commented):
[Claude Code 🤖] This review focused on the two places a wrong branch silently destroys user data: the serializer's new preserve-vs-drop boundary, and the PATCH merge-base move — both sit directly on the write path.

**Bottom line: no blocking issues.** Every mechanism claim in the description checks out against the code, and the affected suites pass locally against this branch: `card-endpoints-test.ts` + `atomic-endpoints-test.ts` 79/79 green, plus a fresh 60/60 `card-endpoints-test.ts` run at head `78f80435` covering the to-many expansion and its new assertion.

**What lands right:**
- The tri-state lives at the correct altitude. Distinguishing "path names nothing the card declares" (drop) from "path crosses a type the definition store cannot hold" (preserve) in `resolveDottedFieldDef` is the minimal precise fix; the tempting wrong version — blanket preservation of unknown relationship keys — would have turned typo'd keys into permanent file residue. I verified the boundary is exact (inline confirmation): unknown keys still drop, and resolved-but-missing types still fail the write loudly because `lookupDefinition` throws rather than returning undefined.
- The merge-base move keeps the critical-section shape and simplifies its guarantee: instead of "wait for indexing inside the lock so the next waiter's index read is fresh", the next waiter now reads the same bytes the lock ordered. Write and index flush both complete inside the lock, so the concurrent-PATCH lost-update guarantee holds (inline confirmation has the full trace).
- Sparse files converge PATCH with POST and `/_atomic` on stored shape, and GET responses are untouched since they serialize from the index. The tests' priming pattern honestly documents the one-time canonicalization rewrite a hand-authored file gets on first PATCH.

1. Mirror the to-many `data: [...]` expansion into the declared `linksToMany` fan-out — the resolvable branch still empties the shape the unresolvable branch now preserves (pre-existing, non-blocking; file-serializer thread).
2. Pin the corrupt-file `systemError` path with a test — it's the only new behavior a regression could silently undo (card-endpoints-test thread).
3. Delete the now-dead `original.meta` fallbacks in `patchCardInstance` (nit; realm.ts thread).

**Adjacent, out of scope:** containsMany per-item polymorphic overrides cannot influence relationship-path resolution — `parseRelationshipKey` strips the index (its own comment calls the transform lossy) and `isMeta` rejects the array-shaped `meta.fields` entry — so a relationship declared only on an item's polymorphic subtype under a *resolvable* declared type still drops silently. This PR narrows that family of loss (unexported declared types now preserve), but the lossy key-cleaning root remains; noted for whoever picks up the schema-driven key parsing the existing comment already asks for.

packages/runtime-common/file-serializer.ts
habdelra: [Claude Code 🤖] The dependence on the throwing contract is now stated where it lives: `resolveChildDef`'s doc comment spells out that undefined means "ref can't name an export" and that a resolved-but-missing definition throws in `lookupDefinition` before reaching the preserve path.

packages/runtime-common/file-serializer.ts
habdelra: [Claude Code 🤖] Fixed — the declared-`linksToMany` fan-out now derives each indexed entry from its own resource identifier via `normalizeRelationship`, the same expansion the unresolvable branch uses, so a to-many `data: [...]` with no relationship-level links keeps its targets. Behavior with relationship-level links present is unchanged (each entry still carries them).

packages/runtime-common/realm.ts
habdelra: [Claude Code 🤖] Thanks for the trace — nothing further to change here.

packages/runtime-common/realm.ts
habdelra: [Claude Code 🤖] Removed — `isCardResource` requires `meta` with a valid `adoptsFrom` code ref, so the guard makes both fallbacks unreachable. Only the `delete original.meta.lastModified` normalization remains.

packages/realm-server/tests/card-endpoints-test.ts
habdelra: [Claude Code 🤖] Added — a garbage-file PATCH test now pins the behavior: 500 response and the stored bytes left untouched.

lukemelia (approved):
Should we consider making it a lint error to define a FieldDef/CardDef/FileDef subclass that is not exported?

burieberry (approved):

PR #5540, opened 2026-07-17
Change loader module cache to use canonical RRI

backspace:
If a prefix mapping exists, the prefix RRI is used as the key.

This adds `VirtualNetwork#onMappingChange` so the cache can be thrown away if prefix mappings are added or removed.

### DIFF
--- a/packages/host/tests/unit/loader-test.ts
+++ b/packages/host/tests/unit/loader-test.ts
@@ -266,6 +266,32 @@ module('Unit | loader', function (hooks) {
     }
   });
 
+  test('a realm-mapping change discards the module cache', async function (assert) {
+    // The loader keys its module cache by canonical RRI form, whose
+    // relationship to a URL is only stable between mapping changes. A mapping
+    // add/remove must therefore discard the cache so an entry can't survive
+    // under a spelling that no longer resolves the same way.
+    let virtualNetwork = getService('network').virtualNetwork;
+    await loader.import(`${testRealmURL}f`);
+    assert.true(
+      loader.isModuleLoaded(`${testRealmURL}f`),
+      'module is cached after import',
+    );
+    virtualNetwork.addRealmMapping('@test-loader-discard/', testRealmURL);
+    try {
+      assert.false(
+        loader.isModuleLoaded(`${testRealmURL}f`),
+        'adding a realm mapping discards the module cache',
+      );
+    } finally {
+      virtualNetwork.removeRealmMapping('@test-loader-discard/');
+    }
+    assert.false(
+      loader.isModuleLoaded(`${testRealmURL}f`),
+      'removing a realm mapping also discards the module cache',
+    );
+  });
+
   test('isModuleLoaded returns false for a module that has not been imported', function (assert) {
     assert.false(
       loader.isModuleLoaded(`${testRealmURL}a`),
--- a/packages/host/tests/unit/loader-test.ts
+++ b/packages/host/tests/unit/loader-test.ts
@@ -292,6 +292,35 @@ module('Unit | loader', function (hooks) {
     );
   });
 
+  test('dispose() unsubscribes a discarded loader from mapping-change notifications', async function (assert) {
+    // LoaderService replaces its loader on every module edit / session
+    // boundary. Each loader subscribes to realm-mapping changes; without
+    // dispose() the VirtualNetwork keeps that subscription — pinning the
+    // loader and its whole module cache indefinitely. After dispose(), a
+    // mapping change must no longer reach this loader.
+    //
+    // Probe with a net-zero add+remove rather than isModuleLoaded straight
+    // after an add: a mapping add shifts the cache-key form itself, so a
+    // still-cached module would fail to look up under its original URL even
+    // when nothing was discarded. Restoring the mapping restores the key
+    // form, so a surviving cache entry is observable again — which it is only
+    // if the loader ignored both notifications.
+    let virtualNetwork = getService('network').virtualNetwork;
+    let discarded = Loader.cloneLoader(loader);
+    await discarded.import(`${testRealmURL}f`);
+    assert.true(
+      discarded.isModuleLoaded(`${testRealmURL}f`),
+      'module is cached after import',
+    );
+    discarded.dispose();
+    virtualNetwork.addRealmMapping('@test-loader-dispose/', testRealmURL);
+    virtualNetwork.removeRealmMapping('@test-loader-dispose/');
+    assert.true(
+      discarded.isModuleLoaded(`${testRealmURL}f`),
+      "a disposed loader's cache survives mapping changes (subscription released)",
+    );
+  });
+
   test('isModuleLoaded returns false for a module that has not been imported', function (assert) {
     assert.false(
       loader.isModuleLoaded(`${testRealmURL}a`),
--- a/packages/host/tests/unit/loader-test.ts
+++ b/packages/host/tests/unit/loader-test.ts
@@ -281,7 +281,7 @@ module('Unit | loader', function (hooks) {
     try {
       assert.false(
         loader.isModuleLoaded(`${testRealmURL}f`),
-        'adding a realm mapping discards the module cache',
+        'module is not visible under its pre-mapping spelling after a mapping add',
       );
     } finally {
       virtualNetwork.removeRealmMapping('@test-loader-discard/');
--- a/packages/runtime-common/loader.ts
+++ b/packages/runtime-common/loader.ts
@@ -210,6 +210,13 @@ export class Loader {
   private fetchImplementation: Fetch;
   private resolveImport: (moduleIdentifier: string) => string;
   private virtualNetwork: VirtualNetwork | undefined;
+  // Unsubscribe for the realm-mapping-change listener registered below. The
+  // VirtualNetwork outlives any single loader (LoaderService replaces the
+  // loader on every module edit / session boundary), so a loader that isn't
+  // unsubscribed when it's discarded stays pinned — along with its whole
+  // module cache — by the listener the network still holds. `dispose()`
+  // releases it; the owner calls that before dropping the loader.
+  private unsubscribeMappingChange: (() => void) | undefined;
   // When the host runs inside a prerender, `setTimeout` is suppressed by
   // the render-timer-stub so the default sleep used by
   // `fetchWithTransientRetry` would never resolve and a transient 5xx on
@@ -235,11 +242,19 @@ export class Loader {
     // relationship to a real URL is only stable between realm-mapping changes.
     // Discard the RRI-keyed caches whenever a mapping is added or removed so an
     // entry can't outlive the spelling it was keyed under.
-    this.virtualNetwork?.onMappingChange(() => {
+    this.unsubscribeMappingChange = this.virtualNetwork?.onMappingChange(() => {
       this.modules.clear();
       this.moduleCanonicalURLs.clear();
       this.knownDepsCache.clear();
     });
+  }
+
+  // Release the realm-mapping-change subscription so this loader can be
+  // garbage-collected once discarded. Only detaches the listener — it does not
+  // clear the caches, since a clone made from this loader carries them forward.
+  dispose() {
+    this.unsubscribeMappingChange?.();
+    this.unsubscribeMappingChange = undefined;
   }
 
   getVirtualNetwork(): VirtualNetwork | undefined {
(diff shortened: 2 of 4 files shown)


packages/runtime-common/loader.ts
```diff
@@ -231,6 +231,15 @@ export class Loader {
...
     this.virtualNetwork = options?.virtualNetwork;
+    // Module caches are keyed by canonical RRI form (see moduleCacheKey), whose
+    // relationship to a real URL is only stable between realm-mapping changes.
+    // Discard the RRI-keyed caches whenever a mapping is added or removed so an
+    // entry can't outlive the spelling it was keyed under.
+    this.virtualNetwork?.onMappingChange(() => {
+      this.modules.clear();
+      this.moduleCanonicalURLs.clear();
+      this.knownDepsCache.clear();
+    });
```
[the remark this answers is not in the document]
backspace: fixed in 2023cfa

packages/runtime-common/loader.ts
```diff
@@ -231,6 +238,23 @@ export class Loader {
...
     this.virtualNetwork = options?.virtualNetwork;
+    // Module caches are keyed by canonical RRI form (see moduleCacheKey), whose
+    // relationship to a real URL is only stable between realm-mapping changes.
+    // Discard the RRI-keyed caches whenever a mapping is added or removed so an
+    // entry can't outlive the spelling it was keyed under.
+    this.unsubscribeMappingChange = this.virtualNetwork?.onMappingChange(() => {
+      this.modules.clear();
+      this.moduleCanonicalURLs.clear();
+      this.knownDepsCache.clear();
+    });
```
habdelra, line 249: [Claude Code 🤖] **Confirmation — the clear-on-change design is correct as landed, and here are the conditions under which it stays that way.** No change requested; this documents what I verified so the trade-off is on record.

**What this replaces.** The previous key was deliberately mapping-*insensitive* (virtual-alias URL via `mapURL(real-to-virtual)`), precisely so a mapping change couldn't orphan cached entries. This change inverts that: the key is now mapping-*sensitive* (`unresolveURL`), and the instability is handled by discarding the caches on every mapping change. The discard is strictly better than orphaning — an orphaned URL-form entry would become reachable again after a mapping remove restores the key form, re-serving a stale module copy; the discard makes that impossible.

**Why the blast radius is acceptable in production.** I enumerated every production caller of `addRealmMapping`/`removeRealmMapping`: host — `makeVirtualNetwork()` in `packages/host/app/services/network.ts` (base/catalog/skills/openrouter prefixes at service construction) and `shimExternals` in `packages/host/app/lib/externals.ts` (boxel-ui prefix); server — boot-time registration in `packages/realm-server/main.ts` and `worker.ts` (and the server constructs no VN-attached `Loader`s in production code at all — I grepped for every `new Loader(` site). `RealmServerService.setAvailableRealmIdentifiers` — the one path that changes the realm list at runtime — only mutates its in-memory `availableRealms` array, never the VN. So in production the listener fires only during boot, before any meaningful module cache exists; all mid-session add/remove traffic is test-scoped.

**The condition under which this stops being safe.** If realm-prefix mappings ever become runtime-dynamic in the host (e.g. registering user-realm prefixes at login or on workspace creation — a plausible direction for RRI-native work), each add wipes every live loader's compiled modules. Already-rendered card instances keep holding classes evaluated from the discarded cache while fresh imports re-evaluate new class objects, so `instanceof` / polymorphic-field identity checks across that boundary diverge. Whoever makes mappings dynamic needs to pair it with a loader-generation story, not just rely on this listener.

**Mid-flight safety, verified.** A clear during an in-flight import degrades to a refetch, not a wedge: `advanceToState`'s loop re-reads `getModule` every iteration and its `undefined` arm calls `fetchModule` again, and Loader-level shims re-serve through `_fetch`'s `moduleShims` lookup (keyed by URL, untouched by the clear), with `fetchModule`'s shimmed branch re-establishing the module entry and canonical URL.

**One caveat worth knowing.** `unresolveURL`'s second step chases `resolveURLMapping(url, 'virtual-to-real')`, so the key derivation also depends on `urlMappings` — and `addURLMapping` does not notify. That's fine under the current contract (`urlMappings` is append-only with no removal API and is populated only at boot, so it's outside the instability window), but if a removal API is ever added it needs to notify too.

packages/runtime-common/loader.ts
```diff
@@ -231,6 +238,23 @@ export class Loader {
...
+  }
+
+  // Release the realm-mapping-change subscription so this loader can be
+  // garbage-collected once discarded. Only detaches the listener — it does not
+  // clear the caches, since a clone made from this loader carries them forward.
+  dispose() {
+    this.unsubscribeMappingChange?.();
+    this.unsubscribeMappingChange = undefined;
   }
```
habdelra, line 259: [Claude Code 🤖] **The comment's rationale for not clearing the caches is factually wrong: clones do not carry the caches forward.** Non-blocking, but worth fixing before merge — it misdescribes `cloneLoader`'s contract in exactly the place a future reader will look when deciding whether `dispose()` may clear.

**What `cloneLoader` actually does.** It constructs a fresh `Loader` (empty `modules` / `moduleCanonicalURLs` / `knownDepsCache`) and copies only `moduleShims`, re-registering each via `clone.shimModule(...)`. The compiled-module state is deliberately *not* inherited — the clone rebuilds it from the HTTP fetch cache, which is the whole point of the swap (`resetLoader`'s comment in `packages/host/app/services/loader-service.ts` says exactly this: "we keep the fetch cache so we can take advantage of HTTP caching when rebuilding the loader state"). So "a clone made from this loader carries them forward" can't be the reason not to clear; clearing the parent's maps would be invisible to any clone.

**The real reasons not to clear, which the code gets right.** (1) The discarded loader may still be draining in-flight work — `LoaderService` swaps `this.loader` while async imports against the old instance are mid-flight, and clearing under them would force re-evaluation and split class identity within those operations. (2) Clearing buys nothing for memory: once the listener is detached and the owner drops its reference, the maps are collected wholesale.

Suggested rewording:

```suggestion
  // Release the realm-mapping-change subscription so this loader can be
  // garbage-collected once discarded. Only detaches the listener — the caches
  // are left intact because a discarded loader may still be draining in-flight
  // imports, and once nothing references it the maps are collected wholesale.
  dispose() {
    this.unsubscribeMappingChange?.();
    this.unsubscribeMappingChange = undefined;
  }
```

packages/host/tests/unit/loader-test.ts
```diff
@@ -266,6 +266,61 @@ module('Unit | loader', function (hooks) {
...
+    assert.true(
+      loader.isModuleLoaded(`${testRealmURL}f`),
+      'module is cached after import',
+    );
+    virtualNetwork.addRealmMapping('@test-loader-discard/', testRealmURL);
+    try {
+      assert.false(
+        loader.isModuleLoaded(`${testRealmURL}f`),
+        'adding a realm mapping discards the module cache',
+      );
```
habdelra, line 285: [Claude Code 🤖] **Nit, non-blocking: this assertion's message overclaims what it observes — it would pass even if the add discarded nothing.** The dispose test's own comment explains the mechanism: a mapping add shifts the cache-key form itself. After `addRealmMapping('@test-loader-discard/', testRealmURL)`, `isModuleLoaded` unresolves the lookup to `@test-loader-discard/f`, while an undiscarded entry would still sit under its import-time key `${testRealmURL}f` — a miss either way. I traced both paths through `moduleCacheKey` → `unresolveURL` to confirm.

The assertion that actually pins the discard is the final one: after the remove restores the key form, a surviving entry would resurface and `isModuleLoaded` would return true. (Strictly, the pair pins "at least one of add/remove discards" — a remove-only implementation would also pass — but that disjunction is the observable contract; the stale-resurrection scenario is exactly what the last assertion catches, and there's no public-API probe that can distinguish which event did the discarding.)

Suggested message that matches what's observed:

```suggestion
      assert.false(
        loader.isModuleLoaded(`${testRealmURL}f`),
        'module is not visible under its pre-mapping spelling after a mapping add',
      );
```

The test comment at the top is accurate as written; only this message oversells. Also: the net-zero add+remove probe in the dispose test below, with its explanatory comment, is genuinely good test design — that comment will save the next person real time.

habdelra (commented):
[Claude Code 🤖] Review focus: the semantics of the new mapping-sensitive cache key (`unresolveURL` in `moduleCacheKey`), the invalidation/subscription lifecycle, and whether the leak fix covers every loader-replacement site. All claims below were verified against the checked-out branch, not the diff view.

**Bottom line: no blocking issues.** The RRI-keyed cache, the clear-on-mapping-change subscription, and the `dispose()` plumbing are correct as landed. One code comment misdescribes `cloneLoader`'s behavior and should be fixed before merge (one-line change, thread on `dispose()` in `loader.ts`).

**What lands right:**

- The cache key now matches the canonical identifier form the loader already emits everywhere else (`canonicalIdentifier` feeds `identify()` output and dependency lists) — one canonicalization scheme instead of two coexisting ones (virtual-alias for keys, RRI for outputs). I traced all three spellings through the new key: resolved real URL and RRI prefix collapse via the `realmMappings` scan, and the virtual-alias spelling collapses via the `resolveURLMapping` chase — single class identity holds across every spelling.
- Clear-on-change converts the documented instability (RRI→URL is only stable between mapping changes) into an invalidation event rather than a stale-key hazard, and it's strictly safer than the orphaning the previous stable-key design accepted — detail and the boot-time-only production evidence in the thread on the constructor subscription.
- The unsubscribe fix is complete: `dispose()` is called at all three loader-replacement sites in `LoaderService` (`resetSessionBoundary`, both `resetLoader` branches) plus the destructor. I grepped the tree for other production `new Loader(`/`cloneLoader` sites — there are none outside the host service (the realm server registers mappings at boot but constructs no VN-attached loaders); test-created loaders subscribe to per-test VNs that die with the test owner.
- Both new tests ran in CI (confirmed in the merged host test report) and the net-zero add+remove probe in the dispose test is exactly the right observable for "subscription released".

**Recommendations:**

1. Fix the `dispose()` comment — clones do not carry the caches forward; corrected wording suggested in the thread on `packages/runtime-common/loader.ts`. (Non-blocking but cheap and worth doing here.)
2. Optional: reword the mapping-add assertion message in `loader-test.ts` so it doesn't overclaim — thread there has the detail.

**Adjacent, out of scope:**

- `moduleCacheKey` now runs `unresolveURL` on every `getModule`/`setModule`. Cost is at parity with the removed path (both pay `resolveURLMapping` — URL construction plus `RealmPaths` — on the common user-realm miss; the base-realm real-URL case actually got cheaper since the `realmMappings` scan hits before any URL construction). If loader ops ever show up in render profiles again, a VN-level memo for `unresolveURL` mirroring `toURLHrefCache` (cleared on mapping change, which `onMappingChange` now makes trivial) is the natural follow-up.
- `knownDepsCache` keys remain resolved-URL-form (`resolveImport` output) rather than `moduleCacheKey` form, so two spellings of one root can hold duplicate entries. Correctness is covered by the clear-on-change; unifying the keying is a nicety for whoever continues the RRI-native migration.
- Percy reports 5 visual diffs awaiting review — worth a glance before merge; nothing in this diff should change rendering, so they're likely environmental, but that's for Percy review to confirm.

packages/runtime-common/loader.ts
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Verified the fix resolves this thread completely. `dispose()` detaches the listener via the retained unsubscribe, and I checked every loader-discard path in `LoaderService`, not just the two the finding named: `resetSessionBoundary` and both branches of `resetLoader` dispose the outgoing loader before replacing it, and the destructor disposes the final one at service teardown. The remaining `Loader` constructions in the tree are test-scoped and subscribe to per-test VirtualNetworks that die with the test owner, so no unsubscribe is needed there. The unit test pins the released subscription with a net-zero mapping add+remove against a disposed clone, and it ran green in CI.

habdelra (approved):

packages/host/tests/unit/loader-test.ts
backspace: addressed in 46e9a65a0c

packages/runtime-common/loader.ts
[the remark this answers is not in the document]
backspace: updated in 91784aa02c

backspace: also in 91784aa02c

packages/runtime-common/virtual-network.ts
```diff
@@ -35,11 +35,29 @@ export class VirtualNetwork {
   // mapping is added or removed (both clear the cache).
   private toURLHrefCache = new Map<string, string>();
 
+  // Notified whenever a realm-prefix mapping is added or removed. Consumers
+  // that key caches by the RRI form a mapping produces (e.g. the Loader's
+  // module cache) subscribe here to discard those entries when the mapping
+  // set changes — the RRI→URL relationship is only stable between changes.
+  private mappingChangeListeners = new Set<() => void>();
```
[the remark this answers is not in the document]
backspace: fixed in 2421aa759e

PR #5542, opened 2026-07-17
test: Add CI for npm-installed CLI

backspace:
`boxel parse` is silently broken in the npm release of Boxel CLI. The test suite didn’t catch this because it runs using monorepo conventions, not the npm package.

This makes several changes:
1. a `runBoxel` test helper to test CLI behaviour, not in-code behaviour
2. which can run with `dist/index.js` as before or path to an npm-installed package
3. the existing CI job now runs against a tarred version of the package
4. new CI jobs run after unstable and stable npm releases to exercise the just-released package

# Deliberate test failure

The failed CI job is known, to demonstrate the problem of undetected brokenness, and that this new CI structure would fix it. The fix that causes the test to pass is in #5551, separated for demonstration and ease of review.


packages/boxel-cli/scripts/run-cli-suite.ts
```diff
@@ -0,0 +1,197 @@
...
+  let attempt = 0;
+  for (;;) {
+    let result = spawnSync(
+      'npm',
+      ['view', `${PKG_NAME}@${version}`, 'version'],
+      { encoding: 'utf8' },
+    );
+    if (result.status === 0 && result.stdout.trim()) {
+      console.log(`Resolved ${PKG_NAME}@${version} → ${result.stdout.trim()}`);
+      return;
```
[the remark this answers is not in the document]
backspace: fixed in d874d0670b

packages/boxel-cli/tests/integration/parse.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  {
+    name: 'plain-glimmer',
+    covers: '@glimmer/component + @tracked + contains(NumberField)',
+  },
+  {
+    name: 'tracked-format-class',
+    covers: '@tracked inside a static isolated format class',
+  },
+  { name: 'runtime-common', covers: 'bare @cardstack/runtime-common import' },
+  { name: 'boxel-host-tools', covers: '@cardstack/boxel-host/tools/* import' },
```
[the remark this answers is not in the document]
backspace: the failure is deliberate, and fixed in #5551

packages/boxel-cli/tests/integration/parse.test.ts
```diff
@@ -0,0 +1,84 @@
...
+const CLEAN_FIXTURES: { name: string; covers: string }[] = [
+  {
+    name: 'plain-glimmer',
+    covers: '@glimmer/component + @tracked + contains(NumberField)',
+  },
+  {
+    name: 'tracked-format-class',
+    covers: '@tracked inside a static isolated format class',
+  },
+  { name: 'runtime-common', covers: 'bare @cardstack/runtime-common import' },
```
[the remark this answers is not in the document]
backspace: this is recorded in CS-12254

packages/boxel-cli/tests/helpers/run-boxel.ts
```diff
@@ -0,0 +1,144 @@
...
+      cwd: options.cwd,
+      env,
+      stdio: ['pipe', 'pipe', 'pipe'],
+      timeout: options.timeout ?? 60_000,
+    });
+
+    let stdout = '';
+    let stderr = '';
+    child.stdout.on('data', (chunk) => (stdout += chunk));
+    child.stderr.on('data', (chunk) => (stderr += chunk));
```
habdelra, line 124: [Claude Code 🤖]

**Non-blocking (introduced here): `stdout`/`stderr` accumulate via per-chunk `Buffer`→`string` coercion, which corrupts multi-byte UTF-8 that straddles a chunk boundary.**

No encoding is set on the streams, so each `data` event delivers a `Buffer`, and `stdout += chunk` runs `chunk.toString()` on *each chunk independently*. When a multi-byte character (an emoji, an accented letter) has its bytes split across two `data` events — which an OS pipe can do at any buffer boundary — each half decodes to U+FFFD (`�`) and the character is lost.

**Where it can bite.** Two paths route non-ASCII realm content through here and then `JSON.parse` it: `file read` returning card content, and the `.json()` helper on any `--json` command whose payload contains user text. A card with an emoji in a field, on a large-enough response, would make `json()` throw or an equality assertion fail — intermittently, keyed to where the pipe happens to split. It's a latent flake, not a guaranteed failure, which is exactly why it would be painful to debug later.

**Fix** — let Node's `StringDecoder` buffer partial sequences across events by setting the encoding once:

```suggestion
    child.stdout.setEncoding('utf8');
    child.stderr.setEncoding('utf8');
    child.stdout.on('data', (chunk) => (stdout += chunk));
    child.stderr.on('data', (chunk) => (stderr += chunk));
```

(Collecting `Buffer`s and doing a single `Buffer.concat(...).toString('utf8')` at `close` is equivalent.)

---

packages/boxel-cli/tests/integration/file-read.test.ts
```diff
@@ -107,53 +135,73 @@ describe('file read (integration)', () => {
...
-
-    let result = await read(realmUrl, 'test-card.json', {
-      profileManager: emptyManager,
-    });
-    expect(result.ok).toBe(false);
-    expect(result.error).toContain('No active profile');
-
-    fs.rmSync(emptyDir, { recursive: true, force: true });
+    let emptyHome = fs.mkdtempSync(path.join(os.tmpdir(), 'boxel-empty-'));
+    new ProfileManager(path.join(emptyHome, '.boxel-cli'));
```
habdelra: [Claude Code 🤖]

**Non-blocking cleanup (introduced by the conversion): this `new ProfileManager(...)` is a no-op and can be deleted.**

The constructor only reads config (`loadConfig`) — it does not create the config dir (`ensureConfigDir` isn't called from it) — and the instance is never assigned or used. The "no active profile" behavior this test asserts comes entirely from spawning the CLI with `HOME=emptyHome` (an empty `mkdtemp` dir with no `profiles.json`), so the line does nothing.

The mechanical conversion left the identical dead statement in ~15 files — a grep for `new ProfileManager(path.join(emptyHome, '.boxel-cli'))` hits file-write, file-delete, file-list, file-touch, file-lint, read-transpiled, run-command, realm-archive (×2), realm-list, realm-remove, realm-cancel-indexing, realm-wait-for-ready, and realm-indexing-errors. Dropping the line in each (and the now-unused `ProfileManager` import wherever this is its only use) keeps these "empty home ⇒ no active profile" tests reading as exactly what they check.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** This reviews the new test-execution machinery — the `runBoxel` subprocess helper and the `run-cli-suite.ts` install-and-run harness — the CI wiring that drives it, and whether the ~30 in-process→subprocess test conversions preserve behavior. I checked out the head branch and traced the module-resolution and packaging claims against the actual `parse` tsconfig and `bundled-types/`.

**Bottom line: the design is sound and I found no blocking correctness bug in the new machinery. The one thing to get right before merge is sequencing — the suite is deliberately red and can only go green once *two* separate follow-ups land, not one (see thread 2 below).** Two non-blocking cleanups are in inline threads.

**What lands right.**
- The split between *which* binary runs (`BOXEL_CLI_BIN`, read once in the parent) and *how* the suite drives it (identical `runBoxel` calls) is the right seam: one suite, three execution contexts (local `dist`, packed tarball, published registry), no per-context test forks.
- `sanitizedParentEnv()` stripping `BOXEL_*` is a real correctness win, not just hygiene — it makes a developer's shell match CI, and the smoke test asserting a leaked `BOXEL_ENVIRONMENT` can't shift behavior pins it.
- The conversions strengthen coverage rather than merely relocating it: the error-path tests now assert `exitCode` alongside the JSON payload, which the in-process calls couldn't observe. `parse.test.ts`'s `plain-glimmer` fixture is a faithful superset of the removed `parse-glimmer-types.test.ts`, and the `deliberate-type-error` fixture that proves glint actually ran (guarding the "produced no TS diagnostics" false-pass) is the crux of the whole PR and is well-constructed.
- `pnpm pack` (not `npm pack`) to rewrite `catalog:` / `workspace:*`, then `npm install` for hoisting, is the correct combination to reproduce a real user's `node_modules` layout.

**On the open review threads (verified against the branch):**

1. **"Wait for the exact published version" (P1) — fixed correctly.** `boxel-cli-publish.yml` now exposes a `version` output from each publish job (read from `package.json`), and both verify jobs install `needs.<job>.outputs.version` — a concrete semver, not the moving `unstable`/`latest` dist-tag. `waitForPublishedVersion` only meaningfully polls for a concrete version and loudly warns on a dist-tag. The chain is airtight. (As a bonus confirmation: `verify-stable` correctly omits the `published` guard that `verify-unstable` needs — the stable job always publishes when it succeeds, whereas the unstable push path can no-op.)

2. **"`runtime-common` fixture can't resolve" (P1) — confirmed, and this is the sequencing point.** `@cardstack/runtime-common` is a `devDependency` (`workspace:*`), so it isn't installed in the npm layout. The generated parse tsconfig maps `@cardstack/base/*`, `@cardstack/host/*`, `@cardstack/boxel-host/commands/*`, `@cardstack/boxel-ui/*`, and a `'*'` catch-all → `bundled-types/host-types/`. The bare `@cardstack/runtime-common` in `fixtures/parse/runtime-common/home.gts` falls through to that catch-all, and `bundled-types/host-types/@cardstack/` ships only `host/`, not `runtime-common/` — so that fixture's `expect(errors).toEqual([])` fails today. Per your own thread notes, the `boxel-host-tools` gap is fixed in the companion PR but `runtime-common` is tracked separately, so **the parse suite needs both fixes to go green; the companion PR alone leaves the `runtime-common` case red.**

3. **"`boxel-host-tools` fixture" (P1) — deliberate, no action here.** Same mechanism (no path mapping, and `tools` is excluded from the bundled host subdirs); addressed by the companion PR.

**Recommendations.**
1. Sequence the merge so `main` is never left with a red CLI job. This flips the existing green "Boxel CLI Tests" job to run the packed install, which fails until *both* the `boxel-host-tools` fix and the `runtime-common` follow-up land — turning that job red on every unrelated PR and failing `verify-unstable` on every subsequent unstable publish. Land them together, or, if they can't ship together, mark the two known-failing fixtures pending (a documented `it.fails`/skip keyed to the missing mappings) so the demonstration is preserved without reddening `main`. (Non-blocking on code; blocking on merge order.)
2. `run-boxel.ts`: set the stream encoding to avoid multi-byte UTF-8 corruption on chunk boundaries — see thread. (Non-blocking.)
3. Drop the dead `new ProfileManager(...)` statement duplicated across ~15 converted tests — see thread. (Non-blocking.)

**Adjacent, out of scope.** `waitForPublishedVersion` shells out to `sleep` via `execFileSync('sleep', ...)`, which is POSIX-only. Harmless because it runs only on the Linux publish runners, but it's the one spot where the otherwise cross-platform helpers (the rest honors `USERPROFILE` for Windows) wouldn't run on Windows. Not asking for a change.

---

packages/boxel-cli/tests/helpers/run-boxel.ts
backspace: updated in 5b3d302

packages/boxel-cli/tests/integration/file-read.test.ts
backspace: fixed in 70bb97f

backspace:
> Sequence the merge so `main` is never left with a red CLI job.

The fix PR is already approved. My plan is to merge this once approved, then rebase the fix in #5551, then merge it. I know this will result in `main` briefly having a failing test, but I think that’s fine, as the fix will come just after.

habdelra (approved):

PR #5546, opened 2026-07-18
CardChooserModal: mixed cards+files mode with {id, kind} return

lukemelia:
Closes CS-12205. Phase 2 of the mixed cards+files chooser (parent CS-11809). Stacked on #5531 (CS-12204, file `id` unification) — review/merge that first.

## What

Extend `CardChooserModal` to run in a mixed cards + files mode via `chooseCard(query, { includeFiles: true })`, scoped by a base filter, returning each pick tagged with its `kind`. Most of the data layer already exists from #5428 (mixed search is the default; file rows already render as result tiles).

- **Parameterize `cardsOnly`:** the two hardcoded `@cardsOnly={{true}}` (`SearchPanel`, `Content`) become `{{not state.includeFiles}}`, so mixed mode surfaces file rows and the type picker offers file types.
- **Widen the query type:** `CardChooserQuery.filter` now allows `AnyFilter` so a mixed `any` base filter type-checks.
- **`{id, kind}[]` return contract:** expose `kind` (`'card' | 'file'`) on `RenderableSearchEntryLike` / the `SearchEntry` view-model, derived by reusing the existing file-detection heuristic in `search-entries.ts` (no new classification path). Thread a kind-tagged selection payload (`SearchSelection` / `SelectedSearchItem`) through the result tile → modal so `pickCards` returns `ChosenItem[]`. Cards-only callers are unaffected — the `chooseCard` wrapper projects mixed results back to bare id string(s) unless `includeFiles` is set.
- **`chooseCardTitle`** (ancillary): reviewed — an `any` filter falls through to the neutral default title, no crash. Left as-is per the ticket's non-blocking note; Phase 3 supplies the real title.

## Tests

- **`search-entries-test`**: a card row is tagged `kind: 'card'`; both the item-only and html-only file rows are tagged `kind: 'file'`.
- **`card-search-adorn-test`** (result tile): clicking a card row emits `{ id, kind: 'card' }`; a file row emits `{ id, kind: 'file' }` — the tile is what stamps the kind onto the payload.
- **`card-chooser-mixed-mode-test`** (new): a real-index modal driven by the global `chooseCard({ includeFiles: true, multiSelect: true })` with the canonical `{ type: baseRef }` mixed filter surfaces both a card instance and a markdown file, and resolves kind-correct `{ id, kind }` picks.
- Updated `card-search-selection-menu-test` for the `SelectedSearchItem` selection type.

`lint:types` (host + runtime-common), eslint, and ember-template-lint are green.

## Note

Depends on #5531. The `kind` on file rows relies on the file `id` unification only for downstream exclusion (Phase 3); this PR's kind derivation is independent. No reindex required for this PR on its own.

### DIFF
--- a/packages/host/app/components/card-chooser/modal.gts
+++ b/packages/host/app/components/card-chooser/modal.gts
@@ -37,6 +37,7 @@ import type { Query } from '@cardstack/runtime-common/query';
 import { getFilterTypeRefs } from '@cardstack/host/utils/card-search/type-filter';
 import {
   isNewCardArgs,
+  removeCardJsonExtension,
   type SearchSelection,
   type SelectedSearchItem,
 } from '@cardstack/host/utils/card-search/types';
@@ -92,13 +93,17 @@ type State = {
   lockConsumingRealm?: boolean;
 };
 
-function normalizeCardUrl(url: string): string {
-  return url.replace(/\.json$/, '');
-}
-
 function selectionEquals(a: SearchSelection, b: SearchSelection): boolean {
   if (!isNewCardArgs(a) && !isNewCardArgs(b)) {
-    return normalizeCardUrl(a.id) === normalizeCardUrl(b.id);
+    if (a.kind !== b.kind) {
+      // A card instance and its backing `.json` file share a normalized id but
+      // are distinct picks; kind keeps them from toggling each other.
+      return false;
+    }
+    // `.json` stripping is a card-id convention; a file id is already canonical.
+    return a.kind === 'file'
+      ? a.id === b.id
+      : removeCardJsonExtension(a.id) === removeCardJsonExtension(b.id);
   }
   if (isNewCardArgs(a) && isNewCardArgs(b)) {
     return a.realmURL === b.realmURL;
@@ -522,11 +527,20 @@ export default class CardChooserModal extends Component<Signature> {
           new URL(selectedItem.realmURL),
         );
         if (newCardId) {
-          chosen.push({ id: newCardId.replace(/\.json$/, ''), kind: 'card' });
+          chosen.push({
+            id: removeCardJsonExtension(newCardId)!,
+            kind: 'card',
+          });
         }
       } else {
         chosen.push({
-          id: selectedItem.id.replace(/\.json$/, ''),
+          // `.json` stripping is a card-id convention; a file id (e.g. a card's
+          // backing `Foo/1.json`) is already its canonical id — stripping it
+          // would yield the card-instance id under a `file` kind.
+          id:
+            selectedItem.kind === 'file'
+              ? selectedItem.id
+              : removeCardJsonExtension(selectedItem.id)!,
           kind: selectedItem.kind,
         });
       }
--- a/packages/host/app/components/card-chooser/modal.gts
+++ b/packages/host/app/components/card-chooser/modal.gts
@@ -87,6 +87,12 @@ type State = {
   errorMessage?: string;
   baseFilter?: Filter;
   availableRealmUrls: string[];
+  // The realm scope from the `chooseCard` query. When present, the search is
+  // pinned to these realms (see `initialSelectedRealmsForPanel` /
+  // `lockSelectedRealmsForPanel`) instead of fanning out across every
+  // available realm — which, in mixed mode, would render a tile for every
+  // file in every realm.
+  requestedRealms?: string[];
   hasPreselectedCard?: boolean;
   consumingRealm?: URL;
   preselectConsumingRealm?: boolean;
@@ -124,7 +130,7 @@ export default class CardChooserModal extends Component<Signature> {
             @baseFilter={{state.baseFilter}}
             @initialSelectedRealms={{this.initialSelectedRealmsForPanel}}
             @initialSelectedTypes={{this.initialSelectedTypesForPanel}}
-            @lockSelectedRealms={{state.lockConsumingRealm}}
+            @lockSelectedRealms={{this.lockSelectedRealmsForPanel}}
             @cardsOnly={{not state.includeFiles}}
             as |Bar Content|
           >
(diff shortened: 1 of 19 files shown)


packages/host/app/components/card-chooser/modal.gts
```diff
@@ -509,14 +522,19 @@ export default class CardChooserModal extends Component<Signature> {
           new URL(selectedItem.realmURL),
         );
         if (newCardId) {
-          cardIds.push(newCardId.replace(/\.json$/, ''));
+          chosen.push({ id: newCardId.replace(/\.json$/, ''), kind: 'card' });
         }
+      } else {
+        chosen.push({
+          id: selectedItem.id.replace(/\.json$/, ''),
```
lukemelia: [Claude Code 🤖] **Correctness — `.json` stripped from file-kind selections.**

This `else` branch handles both `card` and `file` kinds, but unconditionally strips a trailing `.json` from the id. A file-meta row backed by a `.json` file (a card's backing file, e.g. `.../Foo/1.json`, or a standalone json file) gets rewritten to `.../Foo/1` while still tagged `kind: 'file'`. The caller then routes on `kind: 'file'` and tries to open `.../Foo/1` as a file — but that's the card-instance id, not a file id, so the attach/open resolves the wrong resource or fails.

`.json`-stripping is a card-only convention (see `removeCardJsonExtension`'s own doc). The old code only ever held card selections, so the invariant held; it isn't re-established now that file kinds flow through here.

```suggestion
      } else {
        chosen.push({
          id:
            selectedItem.kind === 'file'
              ? selectedItem.id
              : selectedItem.id.replace(/\.json$/, ''),
          kind: selectedItem.kind,
        });
      }
```

packages/host/app/components/card-chooser/modal.gts
```diff
@@ -83,20 +92,13 @@ type State = {
...
 
-function selectionEquals(
-  a: string | NewCardArgs,
-  b: string | NewCardArgs,
-): boolean {
-  if (typeof a === 'string' && typeof b === 'string') {
-    return normalizeCardUrl(a) === normalizeCardUrl(b);
+function selectionEquals(a: SearchSelection, b: SearchSelection): boolean {
+  if (!isNewCardArgs(a) && !isNewCardArgs(b)) {
+    return normalizeCardUrl(a.id) === normalizeCardUrl(b.id);
```
lukemelia: [Claude Code 🤖] **Correctness (plausible) — `selectionEquals` conflates a card and its backing `.json` file across kinds.**

`normalizeCardUrl` strips `.json` from both sides before comparing, but `id` is now the only thing compared — `kind` is ignored. In a multiSelect mixed chooser, a card instance (`Foo/1`, `kind: card`) and its file-meta row (`Foo/1.json`, `kind: file`) normalize to the same string and compare equal. `selectFromSearch`'s toggle then treats them as the same selection: selecting the file row when the card is already selected silently toggles the card *off* (and vice-versa), so the user can never hold both, and the surviving pick may carry the wrong `kind`.

If holding both a card and its json file is intended to be possible, the equality check needs to include `kind` when both sides are non-`NewCardArgs`.

packages/host/app/components/card-chooser/modal.gts
```diff
@@ -83,20 +92,13 @@ type State = {
   lockConsumingRealm?: boolean;
 };
 
-function isNewCardArgs(item: string | NewCardArgs): item is NewCardArgs {
-  return typeof item !== 'string' && 'realmURL' in item;
-}
-
 function normalizeCardUrl(url: string): string {
```
lukemelia: [Claude Code 🤖] **Cleanup (plausible) — four copies of the `.json`-stripping rule.**

The shared `removeCardJsonExtension` already does exactly this. Here it's re-implemented as the local `normalizeCardUrl` (used by `selectionEquals`) plus inline `.replace(/\.json$/, '')` calls in `pickCard` (this file) and again in `card-search/sheet-results.gts` and `card-search/result-section.gts`. Four independent copies of the same card-id normalization must stay in sync; a future change to the convention will update some and not others, producing selections whose ids compare unequal to themselves across components. Consider consolidating on `removeCardJsonExtension`.

packages/host/app/components/card-chooser/modal.gts
lukemelia: [Claude Code 🤖] Fixed in a94688c8. `pickCard` now strips `.json` only for `kind: card`; file ids pass through unchanged. Also fixed the same strip in the upstream "Select All" path (`sheet-results` `allCards`), which feeds `state.selectedCards` before the modal — the modal-only fix alone would have left Select All corrupting file ids. New regression test in `card-chooser-mixed-mode-test` drives a standalone `.json` file through Select All and asserts the `.json` id survives with kind=file.

packages/host/app/components/card-chooser/modal.gts
lukemelia: [Claude Code 🤖] Fixed in a94688c8. `selectionEquals` now returns false when the two kinds differ, so a card instance and its backing `.json` file no longer toggle each other in a multi-select chooser. It also applies the `.json`-strip only to card kinds (file ids compared exactly), consistent with the fix in the pick path.

packages/host/app/components/card-chooser/modal.gts
lukemelia: [Claude Code 🤖] Fixed in a94688c8. Consolidated the selection-path copies onto the shared `removeCardJsonExtension`: dropped the local `normalizeCardUrl` and the inline `.replace(/\.json$/, "")` calls in `pickCard`, `sheet-results`, and `result-section`. (Left the ~40 other unrelated `.replace(/\.json$/, "")` uses across host alone — those are a different concern than the selection-equality path flagged here.)

backspace (approved):

PR #5547, opened 2026-07-18
AI Assistant skill menu: one mixed chooser for cards + skill files

lukemelia:
Closes CS-12206. Phase 3 (the user-facing payoff) of the mixed cards+files chooser (parent CS-11809). Stacked on #5546 (CS-12205, mixed chooser) → #5531 (CS-12204, file `id`). Review/merge those first.

## What

One button in the AI Assistant skill menu now attaches **either** a Skill card **or** a skill-bearing markdown file, from a single search chooser.

- Collapse `attachSkillCard` + `attachSkillMarkdown` into one action that opens the mixed chooser via `chooseCard(query, { includeFiles: true })`. The base filter spans both kinds — `any` of `{ type: skillCardRef }` and `{ on: markdownDefRef, eq: { kind: 'skill' } }`.
- **Route by kind:** a card pick → `onChooseCard(id)`, a file pick → `onChooseSkillMarkdown(id)`. Both already flow to the same `attachSkillTask` in `matrix/room.gts`, so no consumer change.
- **Exclude already-attached** (locked decision #3): `not: { eq: { id } }` clauses built client-side from the menu's own skill list. Card skills are excluded immediately; file skills once their rows carry `id` in the search doc (Phase 1 reindex) — so the feature ships without waiting on a deployed reindex.
- **Retire the file-tree skill path** (locked decision #4): the second button and its `chooseFile`-based attach are removed. `chooseFile` / `FileChooserModal` stay for their other consumers (message file attach, Playground, markdown-embed) — untouched.

## Tests (`skills-test.gts`)

- The single button attaches a **skill card** (existing test, now through the mixed chooser) and a **skill markdown file** (rewritten off the retired file-tree flow to click the file's result tile).
- Already-attached **card** and **markdown** skills are both excluded from the picker.
- The retired skill-file button is gone.
- Canceling the chooser re-enables the add button.

`lint:types` (host), eslint, and ember-template-lint are green.

## Note

Depends on #5546 and #5531. In CI the index is fresh (Phase 1 `id` stamping active), so the markdown-skill exclusion is exercised end-to-end. In a deployed environment, file-skill exclusion becomes effective after the affected realms are reindexed; card-skill exclusion needs no reindex.

### DIFF
--- a/packages/host/app/components/ai-assistant/skill-menu/index.gts
+++ b/packages/host/app/components/ai-assistant/skill-menu/index.gts
@@ -17,6 +17,8 @@ import { baseRRI, chooseCard, skillCardRef } from '@cardstack/runtime-common';
 import SkillToggle from '@cardstack/host/components/ai-assistant/skill-menu/skill-toggle';
 import PillMenu from '@cardstack/host/components/pill-menu';
 
+import { skillsRealmURL } from '@cardstack/host/lib/utils';
+
 import type { RoomSkill } from '@cardstack/host/resources/room';
 import type RealmServerService from '@cardstack/host/services/realm-server';
 
@@ -206,12 +208,13 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
           ...exclusions,
         ],
       },
-      // Scope to the user's own workspaces. The mixed chooser (`includeFiles`)
-      // renders a tile per matching row; without a realm scope it fans out
-      // across every server realm — including large shared realms — and the
-      // intended skills may never finish rendering. Skills a user attaches
-      // come from their workspaces.
-      realms: this.realmServer.userRealmIdentifiers,
+      // Scope to the user's own workspaces plus the shared Boxel Skills realm,
+      // where most attachable skills live. A bounded scope keeps the mixed
+      // chooser (`includeFiles`) from fanning out across every server realm
+      // (base, catalog, …) and stalling on their contents.
+      realms: [
+        ...new Set([...this.realmServer.userRealmIdentifiers, skillsRealmURL]),
+      ],
     };
     let chosen = await chooseCard(query, { includeFiles: true });
     if (!chosen) {
--- a/packages/host/app/components/ai-assistant/skill-menu/index.gts
+++ b/packages/host/app/components/ai-assistant/skill-menu/index.gts
@@ -208,14 +208,29 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
           ...exclusions,
         ],
       },
-      // Scope to the user's own workspaces plus the shared Boxel Skills realm,
-      // where most attachable skills live. A bounded scope keeps the mixed
-      // chooser (`includeFiles`) from fanning out across every server realm
-      // (base, catalog, …) and stalling on their contents.
+      // Scope to the realms where attachable skills live: the user's own
+      // workspaces, any catalog realms, and the shared Boxel Skills realm.
       realms: [
-        ...new Set([...this.realmServer.userRealmIdentifiers, skillsRealmURL]),
+        ...new Set([
+          ...this.realmServer.userRealmIdentifiers,
+          ...this.realmServer.catalogRealmIdentifiers,
+          skillsRealmURL,
+        ]),
       ],
     };
+    // TEMP DIAGNOSTIC — remove before merge.
+    console.warn(
+      '[SKILL-SCOPE] user=',
+      JSON.stringify(this.realmServer.userRealmIdentifiers),
+      'catalog=',
+      JSON.stringify(this.realmServer.catalogRealmIdentifiers),
+      'skills=',
+      skillsRealmURL,
+      'available=',
+      JSON.stringify(this.realmServer.availableRealmIdentifiers),
+      'FINAL=',
+      JSON.stringify(query.realms),
+    );
     let chosen = await chooseCard(query, { includeFiles: true });
     if (!chosen) {
       return;
--- a/packages/host/app/components/ai-assistant/skill-menu/index.gts
+++ b/packages/host/app/components/ai-assistant/skill-menu/index.gts
@@ -179,14 +179,20 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
   // (`includeFiles`) surfaces both in a single list and tags each pick with its
   // kind, which routes the result to the matching attach callback.
   private doAttachSkill = restartableTask(async () => {
-    // Exclude already-attached skills. The `not: { eq: { id } }` clauses are
-    // built client-side from the menu's own skill list, so card skills are
-    // excluded immediately; file skills are excluded once their rows carry
-    // `id` in the search doc (Phase 1 reindex), and the feature ships without
-    // waiting on that reindex.
+    // Exclude already-attached skills, matched client-side against the menu's
+    // own skill list by `id`. Card skills are excluded immediately; file skills
+    // once their rows carry `id` in the search doc (Phase 1 reindex), and the
+    // feature ships without waiting on that reindex.
+    //
+    // The exclusion must be NULL-safe: a bare `not: { eq: { id } }` drops any
+    // row whose `id` is absent, because negated `eq` compiles to `NOT (id = X)`
+    // and `NOT (NULL = X)` is NULL (excluded) under three-valued logic. On a
+    // realm not yet carrying the stamped `id`, every skill-file row lacks `id`,
+    // so that would hide *all* skill files whenever any skill is attached. The
+    // `{ eq: { id: null } }` branch (an `IS NULL` test) keeps those rows.
     let exclusions =
       this.args.skills?.map((skill: RoomSkill) => ({
-        not: { eq: { id: skill.cardId } },
+        any: [{ eq: { id: null } }, { not: { eq: { id: skill.cardId } } }],
       })) ?? [];
     let query = {
       filter: {
--- a/packages/host/app/components/ai-assistant/skill-menu/index.gts
+++ b/packages/host/app/components/ai-assistant/skill-menu/index.gts
@@ -1,6 +1,7 @@
 import { fn } from '@ember/helper';
 import { on } from '@ember/modifier';
 import { action } from '@ember/object';
+import { service } from '@ember/service';
 import Component from '@glimmer/component';
 
 import { tracked } from '@glimmer/tracking';
@@ -17,6 +18,7 @@ import SkillToggle from '@cardstack/host/components/ai-assistant/skill-menu/skil
 import PillMenu from '@cardstack/host/components/pill-menu';
 
 import type { RoomSkill } from '@cardstack/host/resources/room';
+import type RealmServerService from '@cardstack/host/services/realm-server';
 
 // A skill expressed as a markdown file is a `MarkdownDef` whose frontmatter
 // declares `boxel.kind: skill`. One branch of the mixed chooser's base filter
@@ -134,6 +136,8 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
     </style>
   </template>
 
+  @service declare private realmServer: RealmServerService;
+
   @tracked private isExpanded = false;
   @tracked private isAttachingSkill = false;
 
@@ -196,6 +200,12 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
           ...exclusions,
         ],
       },
+      // Scope to the user's own workspaces. The mixed chooser (`includeFiles`)
+      // renders a tile per matching row; without a realm scope it fans out
+      // across every server realm — including large shared realms — and the
+      // intended skills may never finish rendering. Skills a user attaches
+      // come from their workspaces.
+      realms: this.realmServer.userRealmIdentifiers,
     };
     let chosen = await chooseCard(query, { includeFiles: true });
     if (!chosen) {
--- a/packages/host/app/components/ai-assistant/skill-menu/index.gts
+++ b/packages/host/app/components/ai-assistant/skill-menu/index.gts
@@ -11,18 +11,16 @@ import pluralize from 'pluralize';
 
 import { Button } from '@cardstack/boxel-ui/components';
 
-import { baseRRI, skillCardRef } from '@cardstack/runtime-common';
-import { chooseCard, chooseFile } from '@cardstack/runtime-common';
+import { baseRRI, chooseCard, skillCardRef } from '@cardstack/runtime-common';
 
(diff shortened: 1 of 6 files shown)


packages/host/app/components/ai-assistant/skill-menu/index.gts
```diff
@@ -184,53 +170,56 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
...
+  // kind, which routes the result to the matching attach callback.
+  private doAttachSkill = restartableTask(async () => {
+    // Exclude already-attached skills. The `not: { eq: { id } }` clauses are
+    // built client-side from the menu's own skill list, so card skills are
+    // excluded immediately; file skills are excluded once their rows carry
+    // `id` in the search doc (Phase 1 reindex), and the feature ships without
+    // waiting on that reindex.
+    let exclusions =
       this.args.skills?.map((skill: RoomSkill) => ({
         not: { eq: { id: skill.cardId } },
```
lukemelia: [Claude Code 🤖] **Correctness — `not:{eq:{id}}` exclusions hide ALL skill files on un-reindexed realms.**

These client-built exclusions are ANDed into the mixed `every`. A skill-markdown row on a realm not yet carrying the Phase-1 stamped `id` has `search_doc->>'id'` = NULL, so `NOT (id = <cardId>)` is NULL under three-valued logic and the row fails the WHERE clause — for *every* skill-file row, not just the attached one. Since the assistant typically boots with skills already attached, a user on an un-reindexed realm opens the chooser and sees only skill cards; no skill files appear.

CI is green because its index is fresh (all files carry `id`), so this only bites deployed realms pre-reindex. The header comment says un-stamped files "are excluded once their rows carry id" (i.e. would still show) — NULL-under-NOT does the opposite and hides them all. Consider making a missing `id` not drop the row (exclude the file kind by `url`, or apply id-exclusions to card rows only).

packages/host/app/components/ai-assistant/skill-menu/index.gts
```diff
@@ -184,53 +170,56 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
...
+          },
+          ...exclusions,
+        ],
       },
+      // Scope to the user's own workspaces. The mixed chooser (`includeFiles`)
+      // renders a tile per matching row; without a realm scope it fans out
+      // across every server realm — including large shared realms — and the
+      // intended skills may never finish rendering. Skills a user attaches
+      // come from their workspaces.
+      realms: this.realmServer.userRealmIdentifiers,
```
lukemelia: [Claude Code 🤖] **Behavior change (intentional — flagging for the record) — scoping to `userRealmIdentifiers` drops shared/catalog skills.**

`userRealmIdentifiers` is `availableRealms` filtered to `type === 'user'`, so skills that live in a shared/catalog/base realm (e.g. a dedicated skills realm) fall out of scope and no longer appear — a capability the old card/file choosers had. This was a deliberate call to stop the mixed chooser fanning out across every realm, but it narrows discoverability. If shared-realm skills should stay attachable, broaden the scope (e.g. `[...userRealmIdentifiers, ...catalogRealmIdentifiers]`).

lukemelia: [Claude Code 🤖] **Correctness (edge case) — an empty `userRealmIdentifiers` silently re-opens the fan-out.**

`realms` is snapshotted here at click time. If no realm is typed `'user'` (or the list is not yet populated), `userRealmIdentifiers` is `[]`, and the modal guard `if (this.state?.requestedRealms?.length)` (card-chooser/modal.gts:266) is falsy for an empty array → it falls back to searching all `availableRealmUrls`. That is exactly the fan-out this scoping was meant to prevent. Worth handling `[]` deliberately rather than treating it as "unscoped".

packages/host/app/utils/card-search/query-builder.ts
```diff
@@ -210,12 +210,18 @@ export function buildSearchQuery(
...
+    // When typeFilter is present, strip a *simple* baseFilter's type constraint
+    // as redundant — the picked subtype already implies it. Only safe for a
+    // single-type base filter. A compound base filter (e.g. the skill menu's
+    // `any: [{ type: skillCardRef }, { on: markdownDefRef, eq: { kind: 'skill' }
+    // }]`) must stay intact: stripping the type branch collapses the `any` and
+    // promotes the sibling field constraint into a required top-level `every`
+    // clause — which would wrongly exclude the other branch's kind (skill cards
+    // vanish, only markdown skills survive).
+    const effectiveBaseFilter =
+      typeFilter && isCardTypeFilter(baseFilter)
```
lukemelia: [Claude Code 🤖] **Cleanup — the `any`-collapse fix is band-aided at the call site; the defect stays latent in `stripTypeFromFilter`.**

Both call sites now gate on `isCardTypeFilter(baseFilter)`, so `stripTypeFromFilter` is only ever passed a bare `{type}` node — its `every`/`any`/`not` recursion is now unreachable from here. The underlying bug (recursing into an `any` and stripping a branch collapses the OR) still lives in the function, so a future caller passing a compound base filter without this guard re-hits the skill-card-vanishes bug. Consider fixing inside `stripTypeFromFilter` (don't strip within an `any`) so the invariant holds regardless of caller.

packages/host/app/components/ai-assistant/skill-menu/index.gts
lukemelia: [Claude Code 🤖] Fixed in 691ecd3. The exclusions are now NULL-safe — each is `any: [{ eq: { id: null } }, { not: { eq: { id } } }]`, so a row with an absent `id` (an un-reindexed skill-file row) is kept, while a present-and-matching id is still excluded. Host tests green.

packages/host/app/utils/card-search/query-builder.ts
lukemelia: [Claude Code 🤖] Fixed in 691ecd3. Moved the guard into `stripTypeFromFilter` itself — it never strips within an `any` now (dropping/collapsing a branch changes which rows match), so the invariant holds for any caller. Removed the call-site `isCardTypeFilter` band-aids, and added a regression unit test asserting a compound `any` base filter stays intact.

packages/host/app/components/ai-assistant/skill-menu/index.gts
lukemelia: [Claude Code 🤖] Addressed in 19f4965. The chooser now scopes to `userRealmIdentifiers` **plus the shared Boxel Skills realm** (`skillsRealmURL`), so shared skills are visible by default — the boxel-skills realm is a *dedicated* realm (`skillsRealmURL`), not a `showAsCatalog` catalog realm, so that is the right target. Host integration tests pass with the 2-realm scope; still resolving fallout in the matrix e2e skill tests from the widened scope before this is fully green.

lukemelia: [Claude Code 🤖] Not changed this round (deferred by design). If `userRealmIdentifiers` is empty the modal still falls back to all realms; with the explicit `skillsRealmURL` now always in the scope array the array is never empty in practice, which narrows this edge case, but a hard guard against the empty-scope fallback would make it airtight. Happy to add that if you want it.

packages/host/app/components/ai-assistant/skill-menu/index.gts
```diff
@@ -184,53 +172,65 @@ export default class AiAssistantSkillMenu extends Component<Signature> {
...
+    // row whose `id` is absent, because negated `eq` compiles to `NOT (id = X)`
+    // and `NOT (NULL = X)` is NULL (excluded) under three-valued logic. On a
+    // realm not yet carrying the stamped `id`, every skill-file row lacks `id`,
+    // so that would hide *all* skill files whenever any skill is attached. The
+    // `{ eq: { id: null } }` branch (an `IS NULL` test) keeps those rows.
+    let exclusions =
       this.args.skills?.map((skill: RoomSkill) => ({
-        not: { eq: { id: skill.cardId } },
+        any: [{ eq: { id: null } }, { not: { eq: { id: skill.cardId } } }],
       })) ?? [];
```
habdelra, line 198: [Claude Code 🤖] **Confirmation (verified, non-blocking) — the NULL-safe exclusion is correct, and the markdown case hinges on `RoomSkill.cardId`.** Putting the linchpin on the record: for a markdown-file skill, `RoomSkill.cardId` is the file's `sourceUrl` (`resources/room.ts` sets `cardId: skillCard.sourceUrl`), and after the file-`id` unification a skill-file row's search-doc `id` is that same source URL — so `not: { eq: { id: skill.cardId } }` lines up against the row's `id` and hides an already-attached markdown skill exactly the way it does a card. Verified end-to-end by the new "excludes an already-attached markdown skill" test.

The documented degradation is genuinely harmless: on a realm not yet reindexed, an attached markdown skill's row has `id = NULL`, the `{ eq: { id: null } }` branch keeps it, so it stays visible and re-selectable — but re-picking it just re-runs `attachSkillTask` → `UpdateRoomSkillsTool`, which keys enabled skills in a `Map` by `sourceUrl` (idempotent), so a duplicate pick is a no-op. Nothing to change here.

---

packages/host/app/components/card-search/panel.gts
```diff
@@ -67,7 +67,11 @@ export default class SearchPanel extends Component<Signature> {
...
     let realmURLs = this.args.initialSelectedRealms;
-    if (!realmURLs || realmURLs.length === 0) {
+    // Only auto-focus a section when the search is pinned to a single realm.
+    // Focusing collapses every other realm's section to its header, so for a
+    // multi-realm scope (e.g. the skill chooser's user + catalog + skills)
+    // focusing the first realm would hide the results of all the others.
+    if (!realmURLs || realmURLs.length !== 1) {
       return null;
     }
     return `realm:${realmURLs[0].href}`;
```
habdelra, line 77: [Claude Code 🤖] **Confirmation (verified, non-blocking) — this shared-component change is safe today; noting the blast radius for the next editor.** `SearchPanel` feeds `initialFocusedSection` to both the card-chooser modal and the search sheet, so I checked every consumer that the `>= 1` → `=== 1` change could touch:

- the search sheet and the mini chooser pass no `initialSelectedRealms` → `null` either way (unchanged).
- the card-chooser modal's `initialSelectedRealms` is `query.realms`; of all `chooseCard(...)` call sites, only the skill menu passes ≥2 realms. The others (`create-listing-modal`, `playground-panel`, the spec/attachment choosers) pass 0–1.

So the only realized behavior change is the intended one. The condition under which this matters again: a future `chooseCard` caller that pins ≥2 realms *and* wants the first one focused would now get no focus. Since `isCollapsed()` hides every non-focused section, auto-focusing the first of several realms was arguably never desirable — this reads as a correct generalization rather than a scoped workaround.

---

packages/host/app/utils/card-search/query-builder.ts
```diff
@@ -211,8 +211,10 @@ export function buildSearchQuery(
   let filters: Filter[];
   if (baseFilter) {
     // When typeFilter is present, strip the baseFilter's (parent) type
-    // constraint as redundant — the picked subtype already implies it. Safe
-    // because the type picker only offers subtypes of the baseFilter type.
+    // constraint as redundant — the picked subtype already implies it.
+    // `stripTypeFromFilter` only unwraps a top-level `{ type }` node and leaves
+    // any `any`-alternation intact, so a compound base filter passes through
+    // whole.
```
habdelra, line 217: [Claude Code 🤖] **Doc accuracy (minor, non-blocking).** This says `stripTypeFromFilter` "only unwraps a top-level `{ type }` node," but the function also recurses into `every` (stripping bare `{ type }` children) and into `not`. The claim that actually matters here — a compound `any` base filter passes through whole — is correct and is exactly what the fix guarantees; it's just the "only a top-level node" phrasing that's imprecise (an `every: [{ type: X }, { eq: … }]` base filter would still have its `{ type: X }` stripped). Consider "unwraps top-level and `every`-nested `{ type }` nodes but leaves any `any`-alternation intact" so the next reader doesn't over-trust the "only top-level" wording.

---

habdelra (commented):
[Claude Code 🤖]

Reviewed with a focus on the mixed-chooser correctness path — filter construction (the exclusions plus the base `any`), the kind-routing of a pick, and the two shared-component edits (`stripTypeFromFilter`, `SearchPanel` focus) this feature drags in — since those are where a single skill-menu change can ripple into other choosers.

**Bottom line: no blocking code issues. The prior review threads are all correctly resolved (I re-verified each against the code), and the design is sound. The one thing gating merge is CI, which is red on the head commit.**

**What lands right**
- Collapsing the two attach buttons into one `chooseCard(query, { includeFiles: true })` and routing by `chosen.kind` is clean; both branches already funnel to the same `attachSkillTask`, so there's no consumer churn. `chosen.id` is the right value for each kind — the modal deliberately skips `.json`-stripping for `file` picks, so a markdown file's id stays its source URL, which is what `attachSkillTask` wants.
- The NULL-safe exclusion (`any: [{ eq: { id: null } }, { not: { eq: { id } } }]`) is the correct three-valued-logic fix, and shipping ahead of a deployed reindex is a reasonable call given the degradation is harmless (see the inline note on the exclusion).
- Moving the `any`-preservation into `stripTypeFromFilter` itself, rather than guarding at the two call sites, is the right home for the invariant, and the new unit test pins it.

**Re-verification of the resolved threads (all correctly closed)**
1. NULL-safe exclusions — correct; the markdown case is verified via `RoomSkill.cardId === sourceUrl` (inline).
2. Scope now user + catalog + skills — and because `skillsRealmURL` is always in the set, the earlier "empty `userRealmIdentifiers` re-opens the fan-out" edge case is closed in practice (the `realms` array is never empty), which settles that deferred thread too.
3. `any`-collapse moved into the helper — correct, with regression coverage.

**Recommendations**
1. **Get CI green before merge (blocking for merge, not a code defect).** The head commit shows a host-test error (shard 13 — the aggregate's `1 🔥` / `1 ❌`) plus a Host Memory Baseline failure. I could not attribute the shard-13 error to a specific test: testem's per-test output lands in the junit artifact, not job stdout, and the artifact download was blocked in my environment. Please re-run shard 13 to establish flaky vs. real; the memory-baseline job is typically informational (a test was added here, which can shift the baseline).
2. Optional: tighten the `buildSearchQuery` comment about `stripTypeFromFilter` (inline note).

**Adjacent, out of scope**: the pre-existing "skill picker excludes already-enabled skills" test fills the search field with `'Exanple'` (a typo for "Example") — not touched by this PR and evidently still passing, but worth a fix by whoever next edits that file.

---

habdelra (approved):

PR #5548, opened 2026-07-19
Honor embed format and custom size in the compose editor

FadhlanR:
## Summary

In the field's edit (compose) view, embedded cards/files didn't honor a BFM directive's format or custom size — most visibly `fitted` (and `isolated`) width/height was ignored, so the embed rendered at the wrong height. The saved views and the edit "preview" sub-mode rendered sizes correctly, so the bug was specific to the CodeMirror compose editor.

Fixes CS-12112.

## Root cause

The format/size specifier (`::card[url | fitted w:400 h:300]`, `| isolated`, …) was dropped at three points in the compose pipeline:

- `CardWidget` never recorded the directive's format/size — `toDOM` emitted only id/kind/ref-type.
- The target notifier hardcoded `format: kind === 'inline' ? 'atom' : 'embedded'`.
- The live-preview template rendered `<RefComponent @format=… />` with no sizing, and `CardWidgetTarget.format` couldn't even represent `fitted`/`isolated`.

## Fix

Thread format/width/height through the widget → DOM data-attrs → target → template chain, reusing the existing `parseBfmSizeSpec` / `bfmRefFormatAndSize` helpers so the compose editor matches the saved and preview markdown renderers exactly — including `overflow: hidden` for fitted. Applies to both `:card` and `:file` embeds.

Two equality guards also had to learn about the size fields so a size-only edit (e.g. `w:400` → `w:600`) actually re-renders: `CardWidget.eq` (else CodeMirror reuses the stale DOM element) and the `_applyTargets` structural-identity skip check (else the Glimmer re-render is skipped).

## Testing

- `pnpm lint:types` clean; `pnpm lint:js` clean.
- New `codemirror-context` integration tests cover block fitted with explicit size, block isolated, atom/embedded defaults, and file-embed parity. Full module 51/51.
- `codemirror embed toolbar` 8/8 and `RichMarkdownField` 25/25 — no regressions.

### Before

Screenshot 2026-07-21 at 19 59 18

### After

Screenshot 2026-07-21 at 19 48 57

### DIFF
--- a/packages/host/app/components/markdown-embed-chooser/format-selection.ts
+++ b/packages/host/app/components/markdown-embed-chooser/format-selection.ts
@@ -278,6 +278,18 @@ export default class EmbedFormatSelection {
     }
   }
 
+  // A Custom-size fitted embed must carry at least one explicit dimension —
+  // otherwise it serializes to the size-less bare `fitted`, which defeats the
+  // point of the Custom option. Every named format (atom/embedded/isolated and
+  // the fixed fitted variants) already carries its size, so it is always
+  // insertable; the consumer disables the insert CTA when this is false.
+  get hasValidSize(): boolean {
+    if (this.category !== 'custom') {
+      return true;
+    }
+    return this.width !== undefined || this.height !== undefined;
+  }
+
   // True once any format/placement/size field has diverged from the seeded
   // snapshot. Target identity (the user swapped in a different card/file) is
   // tracked separately by the tab panel, which owns the resolved target.
--- a/packages/host/app/lib/codemirror-context.ts
+++ b/packages/host/app/lib/codemirror-context.ts
@@ -147,9 +147,10 @@ class CardWidget extends WidgetType {
     readonly cardId: string,
     readonly kind: 'inline' | 'block',
     readonly refType: 'card' | 'file' = 'card',
-    // Size specifier fields (as strings, for a clean DOM data-attr round-trip),
-    // parsed from the directive's `| spec` segment. Undefined when absent.
-    readonly format?: string,
+    // Size specifier fields parsed from the directive's `| spec` segment.
+    // `width`/`height` stay strings for a clean DOM data-attr round-trip;
+    // `format` is the parsed render format. Undefined when absent.
+    readonly format?: BfmRefFormat,
     readonly width?: string,
     readonly height?: string,
   ) {
@@ -668,11 +669,12 @@ function buildLinkDecorations(
 // ── Card decorations ───────────────────────────────────────────────────────
 
 // Parse a directive's inner content (between `[` and `]`) into the URL/id and
-// the CardWidget size args. Kept as strings so they round-trip cleanly through
-// DOM data-attrs, where `notifyTargets` re-derives format + style.
+// the CardWidget size args. `width`/`height` are kept as strings so they
+// round-trip cleanly through DOM data-attrs, where `notifyTargets` re-derives
+// format + style.
 function parseCardWidgetContent(content: string): {
   cardId: string;
-  format?: string;
+  format?: BfmRefFormat;
   width?: string;
   height?: string;
 } {
--- a/packages/host/app/lib/codemirror-context.ts
+++ b/packages/host/app/lib/codemirror-context.ts
@@ -45,8 +45,11 @@ import {
 } from '@codemirror/view';
 
 import {
+  type BfmRefFormat,
   type BfmRefRange,
+  bfmRefFormatAndSize,
   extractBfmRefRanges,
+  parseBfmSizeSpec,
 } from '@cardstack/runtime-common/bfm-card-references';
 
 // ── Card widget target interface ────────────────────────────────────────────
@@ -54,11 +57,15 @@ import {
 export interface CardWidgetTarget {
   element: HTMLElement;
   cardId: string;
-  format: 'atom' | 'embedded';
+  format: BfmRefFormat;
   kind: 'inline' | 'block';
   // 'card' refs (`:card[URL]`) resolve to CardDef instances; 'file' refs
   // (`:file[URL]`) resolve to FileDef instances.
   refType: 'card' | 'file';
+  // Inline sizing (`width`/`height`, plus `overflow: hidden` for fitted) derived
+  // from the directive's size specifier. Undefined for non-fitted formats.
+  // Mirrors the style the saved/preview markdown renderers apply.
+  style?: string;
 }
 
 // ── State effect for opening card search ────────────────────────────────────
@@ -140,6 +147,11 @@ class CardWidget extends WidgetType {
     readonly cardId: string,
     readonly kind: 'inline' | 'block',
     readonly refType: 'card' | 'file' = 'card',
+    // Size specifier fields (as strings, for a clean DOM data-attr round-trip),
+    // parsed from the directive's `| spec` segment. Undefined when absent.
+    readonly format?: string,
+    readonly width?: string,
+    readonly height?: string,
   ) {
     super();
   }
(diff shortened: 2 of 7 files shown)


packages/host/app/components/markdown-embed-chooser/format-selection.ts
```diff
@@ -278,6 +278,18 @@ export default class EmbedFormatSelection {
...
+  // otherwise it serializes to the size-less bare `fitted`, which defeats the
+  // point of the Custom option. Every named format (atom/embedded/isolated and
+  // the fixed fitted variants) already carries its size, so it is always
+  // insertable; the consumer disables the insert CTA when this is false.
+  get hasValidSize(): boolean {
+    if (this.category !== 'custom') {
+      return true;
+    }
+    return this.width !== undefined || this.height !== undefined;
+  }
```
habdelra, line 291: [Claude Code 🤖] **Editing an existing bare `::card[url | fitted]` disables the "Done" CTA, so the unchanged embed can't be accepted.** The automated reviewer flagged this; I traced it end to end and it reproduces.

**The path.** Bare `fitted` is a first-class supported form — `parseBfmSizeSpec('fitted')` returns `{ format: 'fitted' }`, and `serializeBfmSizeSpec` emits bare `fitted` for a dimensionless fitted spec. When the pencil opens the chooser on such a directive, `codemirror-editor.gts` calls `editEmbed({ sizeSpec: 'fitted', kind: 'block' })`, and `deriveFormatSeeds('fitted', 'block')` takes the fitted default branch → `{ format: 'custom', width: undefined, height: undefined, kind: 'block' }`. The constructed `EmbedFormatSelection` therefore has `category === 'custom'` with empty `widthInput`/`heightInput`, so `hasValidSize` returns `false` → `pane.gts`'s `ctaDisabled` is `true` → the "Done" button is disabled. The user can't accept the embed they opened without either typing a dimension (which silently rewrites the directive) or removing it.

**Scope: regression introduced here** — before this guard the CTA was always enabled, so the edit round-trip worked. It's narrow (only bare `fitted`; named variants and explicit W×H seed dimensions and are unaffected), so non-blocking — but it does break a valid edit round-trip.

**A way out.** The guard wants to block a *new* Custom insert left empty, not a *seeded* embed opened unchanged; dirtiness distinguishes the two — e.g. in `pane.gts`:
```js
private get ctaDisabled(): boolean {
  return !this.args.selection.hasValidSize && this.args.selection.isDirty;
}
```
An unchanged bare-fitted edit has `isDirty === false`, so "Done" stays enabled and re-serializes the same `| fitted`; a fresh Custom-with-no-dims pick is dirty (selection moved off `atom`) and stays blocked — so the acceptance test added here still passes. Alternatively, scope the block to new inserts / preserve seeded bare-fitted as valid, as the automated reviewer suggested.

**Coverage.** The new acceptance test covers the *insert* path (pick Custom → disabled → enter dims → enabled). The gap is the *edit* path: open the chooser on an existing `::card[url | fitted]` and assert "Done" is enabled and accepting re-inserts the bare `fitted`.

---

packages/host/app/lib/codemirror-context.ts
```diff
@@ -851,12 +932,30 @@ function createCardTargetNotifier(
...
+                kind === 'inline' ? 'atom' : 'embedded',
+              );
+              // Fitted slots carry the width/height plus `overflow: hidden` so
+              // the resolved instance occupies the requested footprint.
+              let style =
+                format === 'fitted'
+                  ? sizeStyle
+                    ? `${sizeStyle}; overflow: hidden`
+                    : 'overflow: hidden'
+                  : undefined;
```
habdelra, line 951: [Claude Code 🤖] **Confirming this is correct, and documenting the invariant it now depends on.** This block is the compose-editor twin of the saved/preview renderer in `operator-mode/preview-panel/rendered-markdown.gts`, and I verified the two match hop for hop: the same `bfmRefFormatAndSize(..., kind === 'inline' ? 'atom' : 'embedded')` call, and the identical fitted-style formula (`${sizeStyle}; overflow: hidden`, else `overflow: hidden`). The atom-vs-embed class split added in `codemirror-editor.gts` mirrors `markdown-bfm-card-slot--inline` / `--inline-embed` there as well. That parity is the whole point of the fix and the thing a future edit must preserve — the two derivations have no shared home, so a change to one silently drifts from the other. Extracting a single shared helper ("derive slot format + resolved style from BFM attrs") would remove the drift risk, but that's a follow-up, not this PR.

Also confirming `htmlSafe(target.style)` in the template is safe here: `style` is assembled only from `bfmRefFormatAndSize`, which admits width/height only after `/^\d+%$/` / `/^\d+$/` gates, so no free-form text reaches the `style` attribute.

---

habdelra (commented):
[Claude Code 🤖] **Reviewed for parity with the saved/preview markdown renderers.** The fix's stated contract is that the compose editor should size and format embeds exactly as the saved view and edit-preview already do, so I traced the widget → `data-boxel-bfm-*` attr → `bfmRefFormatAndSize` → target → template chain against `rendered-markdown.gts` and verified each hop.

**Bottom line: the threading is correct and the twin parity genuinely holds. One non-blocking regression in the edit-mode CTA — already flagged by the automated reviewer, now verified with a concrete repro.**

**What lands right:**
- The widget → DOM attrs → target → template chain reuses the same helpers the render side uses, so `fitted` / `isolated` / explicit-size embeds now derive identically in compose and in saved views (verified hop by hop — see the confirmation on `codemirror-context.ts`).
- Both re-render guards are needed and correct: `CardWidget.eq` (format/width/height) forces CM6 to rebuild the widget DOM on a size-only edit, and the `_applyTargets` structural skip-check (format/style, guarded by the `length` check) forces the Glimmer re-render. Without both, a `w:400 → w:600` edit would silently no-op.
- The new integration tests exercise the real resolved-card path, not just target shape — the fitted slot's computed `width` / `height` / `overflow` are asserted on the rendered element.

**Recommendation (non-blocking):**
1. Keep a bare `::card[url | fitted]` editable — opening the chooser on one currently disables "Done" because the seed lands in the `custom` category with no dimensions. Gate the disable on dirtiness (or scope it to new inserts) and add the edit-path test. Detail in the thread on `format-selection.ts`.

**Adjacent, out of scope:**
- `CardWidgetTarget` is declared twice — in `codemirror-editor.gts` (base) and `codemirror-context.ts` (host) — and this change extends both in lockstep. They're structurally identical; worth collapsing to one exported type before they drift, but not this PR's job.
- The compose editor's loading/unresolved fallback (`codemirror-card-fallback`) carries no size style, whereas `rendered-markdown.gts` sizes its loading shimmer — so a block fitted embed doesn't reserve its footprint until the card resolves. Minor, transient, pre-existing structure; flagging only because the PR's goal is renderer parity.

---

habdelra (approved):

packages/host/app/components/markdown-embed-chooser/format-selection.ts
```diff
@@ -278,6 +278,18 @@ export default class EmbedFormatSelection {
...
+  // A Custom-size fitted embed must carry at least one explicit dimension —
+  // otherwise it serializes to the size-less bare `fitted`, which defeats the
+  // point of the Custom option. Every named format (atom/embedded/isolated and
+  // the fixed fitted variants) already carries its size, so it is always
+  // insertable; the consumer disables the insert CTA when this is false.
+  get hasValidSize(): boolean {
+    if (this.category !== 'custom') {
+      return true;
+    }
+    return this.width !== undefined || this.height !== undefined;
```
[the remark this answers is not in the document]
FadhlanR: [Claude Code 🤖] Good catch — fixed in 302d600157. The insert CTA's disable is now conjoined with dirtiness: `ctaDisabled = !hasValidSize && isDirty`. A bare `::card[url | fitted]` opened for editing seeds the Custom category with no dimensions but is not dirty, so Done stays enabled and accepting re-serializes the same `| fitted`. A fresh Custom pick left empty is dirty (it moved off the atom default), so it stays blocked. Added an edit-path acceptance test: it opens the chooser on an existing bare `::card[url | fitted]`, asserts Done is enabled, and confirms accepting round-trips the directive unchanged.

PR #5549, opened 2026-07-20
Design-first software factory: lifecycle phases, screenshot-reviewed cards, control/product realm split

jurgenwerk:
Reworks the software factory into a design-first pipeline — the design language is decided before any card code is written, every card is screenshotted and reviewed against its acceptance criteria before it counts as done, and the run works explicit lifecycle phases that stop at an operator-chosen point. The package README documents the operator surface; `docs/architecture.md` and `docs/run-trace.md` cover internals.

```
brief ──▶ seeds (Project + board + issues, mechanical)
            └─▶ bootstrap (enriches Project, plans issues)     [sonnet]
                  └─▶ design foundation: brand guide, tokens,  [flagship]
                      family sheet — screenshot-critiqued
                        └─▶ per issue: DESIGN turn (HTML mockup [flagship]
                            → screenshot → critique) → BUILD    [sonnet]
                            turn forked from the design session
                              └─▶ validation → render gate →
                                  review turn (approve / bounce once)
                                    └─▶ hardening: QUnit pass per
                                        shipped card; tests gate
                                        everything after
                                          └─▶ polishing (opt-in)
```

Example run (from `packages/software-factory`, dev stack up):

```sh
pnpm factory:go \
  --brief-url https://localhost:4201/software-factory/Wiki/sticky-note \
  --target-realm https://localhost:4201/user/sticky-note/ \
  --control-realm https://localhost:4201/user/sticky-note-control/ \
  --to-phase hardening --debug
```

Both realms are created if missing; re-running the same command resumes from the board. Watch the run live via the `Runs/sticky-note` card in the control realm.

The pipeline:

- Design-first is the only mode. The old build-immediately loop is removed; a design-foundation issue (brand guide, `design/tokens.css`, family coherence sheet) blocks all implementation, and each issue runs a design turn then a build turn forked from it.
- Screenshot-reviewed cards: finished issues are rendered via the prerenderer (a card that errors can't proceed) and the screenshots judged against acceptance criteria by a reviewer that approves, or bounces once with comments; review-filed defect issues take a diagnose-and-fix turn with no design round.
- Lifecycle phases via `--to-phase design | implementation | hardening | polishing` (default `implementation`). Later-phase issues stay on the board as operator decision points. Hardening synthesizes one QUnit test-pass issue per shipped card, registered with the scheduler directly (no waiting on the index); once test files exist, every subsequent issue's validation runs them.
- Seeded tracker: the Project and IssueTracker board are written mechanically before any agent turn — the board is live from second zero and every seed carries its project link inline.

Operations:

- `--control-realm` splits machinery from deliverable: issues, tracker, validation artifacts, and the run log live in a control realm; the product realm keeps only the built cards, so run-log churn never invalidates the product index.
- Per-turn model budgets: bootstrap/build/hardening default to claude-sonnet-5 (claude backend only); design turns deliberately stay on the session flagship; `--review-model`/`--review-effort`, `--fix-model`/`--fix-effort`, `--build-model`/`--build-effort` override per turn type.
- Live run log (`Runs/<slug>`): a query-backed activity feed that streams design screenshots, validation results, and scheduler narration as the run happens; an orchestrator monitor (`--monitor-level`) adds stall and telemetry notes.
- Run-trace telemetry: every run appends NDJSON spans (per-turn model/tokens/cache-reads, validation, render gate, sync) to `<workspace>/.factory-trace/`; schema and jq recipes in `docs/run-trace.md`.
- `--repo-url` port flow: a PORT-ANALYSIS issue clones and studies a real GitHub app (code, tests, fixtures) before bootstrap plans the card family.
- On-demand skills: a lean always-on core plus `list_skills`/`read_skill` discovery, with `.claude/skills` materialized into the workspace.
- Resume safety: re-running the same command resumes from the board — done seeds stay done, stranded in-review issues re-arm, and the authoritative local workspace reconciles stale index reads.
- Fresh-checkout provisioning: `pnpm factory:setup` builds the missing prerequisites; `factory:go` preflights them and fails once with an actionable list instead of crashing mid-run.

Note for running tests locally: `control-plane-sync.test.ts` uses the `node:test` runner (`node --test`), not the qunit harness. The port-allocator dual-stack test expects Linux socket semantics and fails locally on macOS — on main too, unrelated to this branch.


packages/software-factory/src/factory-agent/opencode.ts
```diff
@@ -481,6 +487,13 @@ export class OpencodeFactoryAgent implements LoopAgent {
     if (issueType === 'bootstrap' && context.briefUrl) {
       return assembleBootstrapPrompt({ context, loader: this.promptLoader });
     }
+    if (issueType === 'hardening' && !context.validationContext) {
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 12904bf — the OpenRouter dispatcher now mirrors the Claude backend's order: prime, then acceptance (before any issue-type branch), then design/analysis/bootstrap/hardening. A review turn gets the acceptance-walkthrough prompt on both backends.

packages/software-factory/src/control-plane-sync.ts
```diff
@@ -0,0 +1,329 @@
...
+        return { ok: false, error: listing.error, pulled: 0 };
+      }
+      let paths = (listing.filenames ?? []).filter(isControlPath);
+      let pulled = 0;
+      for (let relPath of paths) {
+        let localPath = join(this.workspaceDir, relPath);
+        let localExists = await fileExists(localPath);
+        let read = await this.client.read(this.controlRealm, relPath);
+        if (!read.ok || typeof read.content !== 'string') {
+          continue;
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 12904bf — an unreadable listed file now fails the pull with the paths and errors named, and the entrypoint refuses to start the run on a failed pull instead of proceeding against incomplete resume state. Covered by a new test in control-plane-sync.test.ts.

packages/software-factory/src/factory-issue-loop-wiring.ts
```diff
@@ -263,22 +460,71 @@ export async function runFactoryIssueLoop(
...
     });
 
   // 6. Run issue loop
   log.info(`Starting issue loop: targetRealm=${targetRealm}`);
 
+  let runLog: RunLogWriter | undefined;
+  {
+    let runSlug = (config.briefUrl.split('/').pop() ?? 'factory-run')
+      .replace(/\.json$/i, '')
+      .toLowerCase();
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 12904bf — the slug now takes the last nonempty path segment and falls back to `factory-run` if nothing usable remains, so trailing-slash brief/repo URLs get distinct, nonempty run-log paths.

packages/software-factory/src/factory-agent/opencode.ts
```diff
@@ -478,9 +484,54 @@ export class OpencodeFactoryAgent implements LoopAgent {
 
   private buildPrompt(context: AgentContext): string {
     let issueType = (context.issue as Record<string, unknown>).issueType;
+    if (context.primeTurn === true) {
+      return this.promptLoader.load('prime', {
+        project: context.project,
+        knowledge: context.knowledge,
```
habdelra, line 490: [Claude Code 🤖]

**`buildPrompt` here dispatches none of the four new turn types the Claude backend handles — regression, blocking for `--agent openrouter`.** This confirms and widens the Codex acceptance-turn flag.

**The divergence.** `claude-code.ts` `buildUserPrompt` (the twin) dispatches, in order: `primeTurn` → `prime`, `acceptanceTurn` → `acceptance-walkthrough`, `issueType === 'design'` → `issue-design-foundation`, `issueType === 'analysis'` → `issue-analysis`, then bootstrap/hardening/iterate/implement. This `buildPrompt` handles only bootstrap, hardening, `validationContext`, and the default `assembleImplementPrompt` — I grepped the whole file and it contains zero references to `primeTurn`, `acceptanceTurn`, `design`, or `analysis`. So under `--agent openrouter`, every one of those four turns silently runs the implement prompt.

**Why it ships unvalidated edits (the acceptance case).** The render gate is on by default regardless of backend (`factory-issue-loop-wiring.ts` `config.renderGate !== false`), and all turns share one `agent` instance. So an acceptance turn reaches this `buildPrompt`, falls through to `assembleImplementPrompt`, and runs as an *implementation* turn that edits product files. `issue-loop.ts` then `await syncWorkspace()` right after the review turn (pushing those edits), reads `.factory-scratch/review-verdict.json`, and — since the implement prompt never writes that file (only `acceptance-walkthrough.md` does, and only for a `rework` verdict) — the missing-file `catch` defaults to `{ verdict: 'approve' }`. Net: edits made by a turn that was supposed to be read-only review are synced and auto-approved without validation.

**The design case is arguably worse for the stated goal:** a `design` foundation issue under `--agent openrouter` runs as a plain implement turn, so the "design-first is the only mode" invariant doesn't hold on that backend at all.

**The fix.** Mirror the Claude backend: dispatch `context.acceptanceTurn` (and `primeTurn`/`design`/`analysis`) *before* the issue-type branches. If OpenRouter genuinely isn't meant to support the design-first pipeline, then fail loud — `throw` for these turn types — rather than silently falling through to implement, which is the dangerous outcome.

---

packages/software-factory/src/control-plane-sync.ts
```diff
@@ -0,0 +1,340 @@
...
+      }
+      let paths = (listing.filenames ?? []).filter(isControlPath);
+      let pulled = 0;
+      let unreadable: string[] = [];
+      for (let relPath of paths) {
+        let localPath = join(this.workspaceDir, relPath);
+        let localExists = await fileExists(localPath);
+        let read = await this.client.read(this.controlRealm, relPath);
+        if (!read.ok || typeof read.content !== 'string') {
+          // A listed file we cannot read means the pull is INCOMPLETE — if
```
habdelra, line 225: [Claude Code 🤖]

**A failed read of a *listed* control file is silently treated as a successful pull — data loss on split-realm resume.** Confirms the Codex P1; here is the traced blast radius.

**Mechanism.** `pull()` lists control paths, then for each one reads it and `continue`s on `!read.ok`. The file is neither mirrored into the workspace nor recorded in `lastPushed`, yet the method still returns `{ ok: true }`. So a transient read failure is indistinguishable, downstream, from "the file doesn't exist remotely."

**Why that loses data.** On resume, the entrypoint calls `createSeedIssue`, whose existence check reads the *local* workspace — `readCard(workspaceDir, SEED_ISSUE_FILE)` in `factory-seed.ts` — with a comment stating it relies on this pull having mirrored the seed locally. If the read of `Issues/bootstrap-seed.json` transiently failed here, that file isn't local, so `seedMatchesAndDone(...)` is false and `createSeedIssue` rewrites the seed to `status=backlog`. `factory-seed.ts` even documents the invariant this breaks: *"this path only fires for a new/changed brief or a never-completed seed."* Then `controlSync.sync()` runs — `lastPushed` has no entry for that path (the read failed), so the hash check can't match and the fresh backlog stub is pushed over the enriched/done remote seed. Same hazard for the seeded Project. Result on resume: completed work re-run and enriched metadata lost.

**The fix.** A read failure on a path that `listFiles` just returned is an error, not an absence: record it and return `{ ok: false }` (or retry the read), so the caller doesn't proceed as though the remote state were complete. At minimum, don't let a listed-but-unreadable file look absent to `createSeedIssue`.

Class: regression (new code). Blocking-ish — resume is a first-class feature of this pipeline.

---

packages/software-factory/src/host-import-manifest.ts
```diff
@@ -0,0 +1,161 @@
...
+    });
+    let names: string[] = [];
+    for (let entry of entries) {
+      if (!entry.isFile() || !entry.name.endsWith('.ts')) continue;
+      let abs = join(entry.parentPath, entry.name);
+      let rel = relative(hostToolsDir, abs)
+        .split(sep)
+        .join('/')
+        .replace(/\.ts$/, '');
+      names.push(rel);
```
habdelra: [Claude Code 🤖]

**The manifest is derived from the host *file tree*, but the authoritative valid-import set is the runtime *shim registry* — and they disagree, so this gate fails valid code and its "did you mean" fix points at a specifier that 404s.** New finding (not in the automated review). Two concrete divergences:

**1. Subdir tools (`bot-requests/*`).** `readdir(..., { recursive: true })` here yields `bot-requests/create-listing-pr-request`. But `shimHostTools` in `packages/host/app/tools/index.ts` registers that module under a **flat** name — `shimHostToolModule(virtualNetwork, 'create-listing-pr-request', …)` (line 229-233) — so the only specifier that resolves at runtime is `@cardstack/boxel-host/tools/create-listing-pr-request`. Consequence in `findHostImportViolations`: a card writing that valid flat import gets `name = 'create-listing-pr-request'`, `manifest.has(name)` is **false** (manifest only holds `bot-requests/create-listing-pr-request`) → violation → `imports-step` returns `passed:false` and fails the iteration on correct code. Worse, `closestManifestEntry` substring-matches and suggests `…/tools/bot-requests/create-listing-pr-request`; that *is* in the manifest so it passes the gate, but the shim never registered it → 404 in the browser, the exact failure this gate exists to prevent. `buildHostToolsSkill` injects the same broken subdir specifier into every agent context. Affects both current subdir tools (`create-listing-pr-request`, `send-bot-trigger-event`).

**2. The `commands/` alias (lines 118-126 + skill text 79).** This file and the injected skill assert `@cardstack/boxel-host/commands/...` "NO LONGER EXISTS" and "will crash at runtime." The host says otherwise: `shimHostToolModule` (index.ts:106-113) shims **both** `tools/${name}` and `commands/${name}` and the comment states the alias is deliberately retained. The shipped precedent card `packages/experiments-realm/bot-request-demo.gts:6` imports `@cardstack/boxel-host/commands/create-listing-pr-request` — valid, still-resolving code that this gate flags with a doubly-false message ("no tool named 'create-listing-pr-request' exists either"). An agent copying that precedent gets its correct code failed. Banning `commands/` to steer toward the canonical `tools/` spelling may be intentional — but then it should be a warning, not `passed:false`, and the "no longer exists / crashes at runtime" justification is factually wrong.

**Root cause + fix (covers both).** Derive the valid-import set from `shimHostTools` (or an exported module list it drives) rather than from the file tree — that registry is the real contract: it flattens subdir names and registers the `commands/` alias. That single change fixes the false-positives and stops the skill from teaching a 404 specifier.

Class: regression (new gate). Non-blocking — the gate is active by default whenever the host checkout is readable (`factory-issue-loop-wiring.ts:274` → `validation-pipeline.ts:278`), but the `commands/`→`tools/` suggestion lets an agent self-correct on that path; the subdir path has no safe escape.

---

packages/software-factory/src/issue-scheduler.ts
```diff
@@ -76,9 +84,42 @@ export class IssueScheduler {
...
   /** Load (or reload) the full issue list from the store. */
   async loadIssues(): Promise<void> {
     this.issues = await this.issueStore.listIssues();
+    for (let [id, local] of this.localIssues) {
+      if (this.issues.some((i) => i.id === id)) {
+        this.localIssues.delete(id);
+      } else {
+        this.issues.push(local);
+      }
+    }
```
habdelra, line 122: [Claude Code 🤖]

**The local overlay re-pushes its *original* snapshot, frozen at `backlog`, so a locally-registered issue that ends `blocked` is re-picked and re-run every outer cycle until the index catches up.** New finding.

**Mechanism.** `addLocalIssues` stores the snapshot in `localIssues` at registration status (`backlog` for a synthesized hardening issue). `refreshIssueState` updates `this.issues[idx]` but never `this.localIssues`, and this loop re-pushes the untouched `local` snapshot whenever the issue isn't yet indexed. Control-plane raw writes "index asynchronously, sometimes minutes behind under load" (this file's own comment), so the window is real.

**Path to the re-run.** The issue is picked, the agent blocks → `issue-loop.ts` writes `status:'blocked'` to disk, sets `exitReason='blocked'`, and `break`s **without** adding to `exhaustedIssues` (only the max-iterations path does that). Next `loadIssues` re-pushes the frozen `backlog` copy; `getUnblockedIssues` filters to `backlog | in_progress` (line 210), so it's eligible again. The reconciliation guard in `issue-loop.ts` reads `readLocalStatus` = `blocked` but only skips when it's `done`, so it proceeds and re-selects. The issue re-runs each cycle until the index reflects the `blocked` status (making it ineligible) or `maxOuterCycles` is hit — wasted turns and a run that can't cleanly drain.

**The fix.** Have `refreshIssueState` (and the blocked/`updateIssue` path) also update the `localIssues` entry, or extend the `issue-loop.ts` reconciliation guard to skip+exclude on a local `blocked` status the way it does for `done`.

Class: regression (the overlay is new here). Non-blocking but wasteful.

---

packages/software-factory/src/factory-issue-loop-wiring.ts
```diff
@@ -263,22 +460,80 @@ export async function runFactoryIssueLoop(
...
     });
 
   // 6. Run issue loop
   log.info(`Starting issue loop: targetRealm=${targetRealm}`);
 
+  let runLog: RunLogWriter | undefined;
+  {
+    // Last NONEMPTY path segment: a trailing slash (common for --repo-url)
+    // must not yield an empty slug, or every such run writes Runs/.json
+    // and successive runs overwrite each other's history.
```
habdelra, line 476: [Claude Code 🤖]

**Empty run-log slug on a trailing-slash brief URL — path collisions across runs.** Confirms the Codex P2.

`'…/sticky-note/'.split('/').pop()` returns `''`, and `'' ?? 'factory-run'` keeps `''` (empty string isn't nullish, so `??` doesn't fire). `runSlug` then flows into `RunLogWriter`, which builds `Runs/${runSlug}.json` → `Runs/.json` and `RunLogEntries/${runSlug}-${seq}.json` → `RunLogEntries/-000001.json` (run-log.ts:262, 485). Any two trailing-slash brief URLs in the same workspace reuse those exact paths and mix or overwrite their run histories.

**The fix.** Take the last *nonempty* path segment (e.g. filter out empties before `pop()`), and fall back to `'factory-run'` when the result is empty rather than only when it's nullish.

Class: regression (new code). Non-blocking, easy fix.

---

packages/software-factory/src/run-log.ts
```diff
@@ -0,0 +1,2394 @@
...
+ * Extract "real card" paths from an agent turn's tool-call log: instance
+ * JSON files written outside the tracker/validation/design folders. These
+ * become card-ready entries that embed the live card.
+ */
+export function cardPathsFromToolCalls(
+  toolCalls: { tool: string; args: Record<string, unknown>; result: unknown }[],
+  limit = 3,
+): string[] {
+  const EXCLUDED =
+    /^(Issues|Projects|Boards|Knowledge(?: |%20)Articles|Spec|Validations|Runs|RunLogEntries|design|design-history|\.factory-scratch)\//;
```
habdelra: [Claude Code 🤖]

**Drift note (non-blocking, in sync today):** the "which written paths are product card instances vs. scaffolding" contract is now encoded in three places that must move in lockstep:

- this `EXCLUDED` regex (line 1134), in `cardPathsFromToolCalls`,
- the byte-identical `EXCLUDED_INSTANCE_DIRS` regex earlier in this file (~line 934), in the instance-path tracker,
- `EXCLUDED_DIRS` as a hand-kept `Set` in `instance-discovery.ts:28`.

They agree right now. The first time the factory starts writing to a new scaffolding directory, whoever adds it has to update all three or the run log / screenshot discovery will treat that scaffolding as a shippable card. Worth extracting one shared predicate (note the regex forms also carry `.factory-scratch` and the `%20`-encoded space variant that the `Set` omits — reconcile those when you unify).

Class: follow-up / maintainability.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I empirically verified the three automated (Codex) findings against the checked-out head, then ran twin-divergence and drift checks across the seams where this rework is most likely to have left one half of a contract behind: the two agent backends (Claude vs. OpenCode/OpenRouter), the new split-realm control-plane sync, and the new host-import gate. CI is green — all three Software Factory shards plus Lint — so I did not re-audit the passing tests; I focused on paths CI doesn't exercise (the OpenRouter backend, resume, the import gate against real host modules).

**Bottom line.** No correctness problem on the default `--agent claude` happy path that CI covers. But there are two data-integrity / ship-safety bugs off that path — one on `--agent openrouter`, one on split-realm resume — plus a new host-import gate that fails *valid* code and steers the agent toward a runtime-404 specifier. For an experimental package none of these are hard blockers, but 1–3 below meaningfully undercut features this PR advertises. Verdict is advisory — approve / request-changes is your call.

- Review "bounce once" is correctly bounded — a second `rework` verdict logs and ships rather than looping (`reviewBounces` in `issue-loop.ts`).
- The phase-split design/build budget handling avoids the "taste turn silently ran on the cheap model" trap and documents why the override is unconditional (`issue-loop.ts` ~1409-1416).
- `render-gate.ts` (URL join safe because the realm is trailing-slash-normalized before construction), `screenshot-execution.ts` (traversal constrained, browser closed in `finally`), and `transient-agent-error.ts` (2 retries, exponential backoff) all check out.

**Findings — class · blocking?** (detail in the inline threads):
1. **OpenCode backend misdispatches every new turn type** — regression · blocking for `--agent openrouter`. `buildPrompt` handles none of `primeTurn` / `acceptanceTurn` / `design` / `analysis`; all fall through to the implement prompt. Sharpest consequence: an acceptance turn edits product files, they're synced, and a missing verdict auto-approves → unvalidated ship. Confirms + widens the Codex acceptance flag. → `factory-agent/opencode.ts`
2. **Control-plane `pull()` swallows failed reads** → seed recreated as a `backlog` stub and re-pushed over enriched remote state on resume — regression · blocking-ish. Confirms Codex P1 with the traced `createSeedIssue` consequence. → `control-plane-sync.ts`
3. **Host-import gate fails valid `bot-requests/*` and `commands/*` imports** and its "did you mean" suggestion / injected skill point at a specifier that 404s — regression · non-blocking. Root cause: the manifest is built from the host file tree instead of the shim registry. → `host-import-manifest.ts`
4. **Scheduler local overlay frozen at `backlog`** re-runs blocked hardening issues each outer cycle until the index catches up — regression · non-blocking. → `issue-scheduler.ts`
5. **Empty run-log slug on a trailing-slash brief URL** → colliding `Runs/.json` / `RunLogEntries/-000001.json` — regression · non-blocking. Confirms Codex P2. → `factory-issue-loop-wiring.ts`

Findings 3 and 4 were not in the automated review.

**Adjacent, out of scope (flagged for whoever touches it next):**
- The "non-product instance dirs" contract is triplicated (`run-log.ts` ×2 + `instance-discovery.ts`) — in sync today, drift risk. Inline note on `run-log.ts`.
- `issue-scheduler.ts` `pickNextIssue`'s doc comment says "status = ready | in_progress", but there's no `ready` in the `IssueStatus` union; the code correctly filters `backlog | in_progress`. Stale comment only — code is right.

---

packages/software-factory/src/factory-agent/opencode.ts
jurgenwerk: [Claude Code 🤖] All four turn types were already dispatched in 12904bf (prime → acceptance → design → analysis, before the issue-type branches, mirroring the Claude backend) — this review ran against the prior head. Agreed on the wider framing: the fall-through-to-implement default was the dangerous part, and acceptance dispatching first is what closes it.

packages/software-factory/src/control-plane-sync.ts
jurgenwerk: [Claude Code 🤖] Fixed in 12904bf (prior head): an unreadable listed file now fails the pull with the paths named, and the entrypoint refuses to start on a failed pull — exactly to protect the createSeedIssue existence check and the enriched Project on resume. Test added in control-plane-sync.test.ts.

packages/software-factory/src/factory-issue-loop-wiring.ts
jurgenwerk: [Claude Code 🤖] Fixed in 12904bf (prior head): last nonempty segment, with a `factory-run` fallback when nothing usable remains.

packages/software-factory/src/host-import-manifest.ts
jurgenwerk: [Claude Code 🤖] Confirmed both divergences and fixed in c285432 by adopting your root-cause suggestion: the manifest is now parsed from the shim registry source (`shimHostToolModule(virtualNetwork, '<name>'` registrations in `packages/host/app/tools/index.ts`) instead of the file tree — flat names for subdir tools come out correct by construction. A known name under `commands/` is no longer a violation (the alias resolves; the injected skill now says `tools/` is canonical and `commands/` is a legacy alias, dropping the false "no longer exists" claim), and an unknown name under either spelling gets a near-match suggestion. Tests updated to model the registry, including the flat-name-for-subdir-file case.

packages/software-factory/src/issue-scheduler.ts
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in c285432 — `refreshIssueState` now updates the overlay entry alongside the working list, so the status that comes back through the workspace-file fallback (e.g. `blocked`) is what `loadIssues` re-pushes while the index lags, and the issue stops being eligible. Test added: a blocked local issue is not re-picked.

packages/software-factory/src/run-log.ts
jurgenwerk: [Claude Code 🤖] Done in c285432 rather than deferred — the three copies collapsed into one predicate (`isNonProductPath` in `src/product-paths.ts`) used by both run-log call sites and instance discovery. Unified to the superset as you suggested: the `%20`-encoded space form and `.factory-scratch` now apply everywhere, including the discovery path that previously missed them.

habdelra (approved):

PR #5557, opened 2026-07-20
Interrupt in-flight prerender renders when the client disconnects

habdelra:
## Background

When a card is indexed (or its HTML is prerendered on demand), a worker sends the render request to the prerender manager, which proxies it to one of the prerender servers. The prerender server drives a headless Chrome tab through the host app's render routes — a "visit": one page acquisition, then a sequence of steps (route transitions, an HTML capture per format, a metadata pass), each executed as CDP calls against the tab.

Client disconnects already propagate through this chain as an abort signal. If the worker gives up on a slow render, or the manager's own caller disconnects, the manager aborts its upstream fetch; the prerender server's HTTP layer notices its socket close and fires an `AbortController` whose signal is threaded down through the prerenderer into the render pipeline. (The manager half of this — client socket close aborts the upstream fetch — is pinned by an existing test in `prerender-manager-test.ts`.)

The gap was in how the prerender server *observed* that signal. It was consulted at a few fixed checkpoints: while a render was queued (waiting on the render semaphore or the per-affinity tab queue), and at the boundaries between a visit's passes. Between checkpoints — that is, while a render step was actually executing — nothing looked at it.

## The failure this produces

Every render step runs under `withTimeout`, a watchdog that races the step against a timer (60 s per step by default). A card whose render wedges the page's JS run loop means the in-flight CDP call cannot return, so the step burns its full 60 s; the timeout path then spends several more seconds on hang diagnostics (a main-thread responsiveness probe, CPU sampling, a paused-stack capture) — all to explain the hang to a caller who, in this scenario, disconnected long ago. Only when the visit reaches its next pass boundary does the cancellation actually surface, and if no checkpoint remains between the burn and the end of the visit, it never surfaces at all: the visit runs every remaining step to completion and serializes a response onto a dead socket.

On a saturated deployment this is expensive in exactly the wrong moment: renders were observed continuing for minutes (in one case over seven) after their requests had been aborted, with the tab occupied for that whole window and every queued render for the same affinity waiting behind it. The burned capacity is also invisible — the caller's job has already failed, so nothing is persisted about the work the tab kept doing.

## What this changes

**`withTimeout` races the abort signal.** The watchdog now takes the visit's abort signal and adds it to the race alongside the step promise and the timer. An abort mid-step rejects immediately with a `rendering`-state `PrerenderCancelledError` — the same error the pass-boundary checks produce — so the existing cancel handling upstream takes over. Two deliberate details:

- The abort *rejects* rather than resolving the timeout marker, which means the timeout path's hang diagnostics are skipped entirely. Those diagnostics exist to explain a hang to a waiting caller; there is no caller.
- The signal spans the whole request while the race is per-step, so each race detaches its abort listener on the way out — a multi-step visit doesn't stack listeners on the shared signal.

**Raw page operations get the same race.** A handful of page operations run outside `withTimeout` (the session/capability setup evaluates at the start of a visit, the file-data stash, the captured-deps read, the command-runner's setup and transition). Against a wedged page these hang until Puppeteer's protocol timeout. They are now wrapped in a small `abortable(signal, fn)` helper that races the operation against the signal with the same rejection semantics. In both `withTimeout` and `abortable`, the abandoned operation's promise stays pending until the tab is torn down; its eventual rejection ("Target closed") is absorbed by the already-settled race, so it can't escape as an unhandled rejection.

**Cancel-handler disposal keeps the warm context and is awaited.** A `rendering`-state cancel has always disposed the affinity's tab — the tab's route state is uncertain mid-transition, and closing the page is precisely what interrupts whatever CDP call the cancelled render abandoned (page close is a browser-process operation, so it lands even when the page's JS thread is pegged). Two aspects of that disposal are tuned here:

- The shared `BrowserContext` is retained. The context holds the realm's warm HTTP cache and session state, which are unaffected by the cancel; the next visit for this affinity spawns a fresh page in the warm context instead of paying the cold module-source waterfall.
- The teardown stays awaited (the default) rather than backgrounded. The caller is gone, so awaiting delays nobody — and it means that by the time the cancellation propagates out, the pool's slot accounting has already released the page. The next visit for the affinity then materializes a fresh page normally. A backgrounded close would leave the closing page counting against the pool ceiling, and a visit arriving in that window can find no standby and get needlessly routed through the browser-restart recovery lane. This is a deliberate divergence from the render-error eviction path, which backgrounds its teardown because it runs mid-visit with a live caller still waiting for the error response.

**Cancel log lines say why.** The prerender server's close listeners abort with a reason string, which rides through `PrerenderCancelledError` into the `render cancelled` log line.

## What doesn't change

Queued-state cancellation (bailing out of the semaphore / tab-queue waits), the pass-boundary checks, and the timeout path for a render whose caller is still connected all behave exactly as before — the timeout race, its diagnostics, and the canonical `Render timeout` error doc are untouched when the signal never fires. The manager needed no changes.

One boundary worth naming: the abort interrupts the *server's wait*, not Chrome's work. An in-page hot loop keeps spinning until the disposal's page close lands; the win is that the visit stops consuming its slot, responds to nothing, and hands the affinity a fresh tab immediately, instead of after multiple watchdog cycles.

## Where the changes live

- `packages/realm-server/prerender/utils.ts` — `withTimeout` gains the signal race
- `packages/realm-server/prerender/prerender-cancel.ts` — the `abortable` helper
- `packages/realm-server/prerender/render-runner.ts` — threads the signal into every step's `withTimeout` call and wraps the raw page operations
- `packages/realm-server/prerender/prerenderer.ts` — cancel-handler disposal retains the shared context
- `packages/realm-server/prerender/prerender-app.ts` — abort reasons on the close listeners

## Testing

Unit tests (no Chrome) pin the race mechanics in `prerender-cancellation-test.ts`: an abort mid-step rejects with a `rendering`-state cancel well before the step timeout; an already-aborted signal prevents the step from starting; the timeout path is byte-for-byte unchanged when the signal never fires; completed steps detach their abort listeners; and a late rejection from an abandoned operation is absorbed rather than surfacing as an unhandled rejection. These were verified to fail against the previous `withTimeout`.

An integration test (real Chrome) in `prerendering-test.ts` holds a render step in-flight, aborts the signal mid-step, and asserts the visit rejects promptly with the cancellation error — then runs a follow-up visit on the same affinity and asserts it renders normally, pinning that the cancelled tab's disposal neither poisons the affinity nor detours the next visit through the restart lane.

https://claude.ai/code/session_015N4eRPdSqvC4uAR1Th1tea

### DIFF
--- a/packages/realm-server/prerender/prerender-app.ts
+++ b/packages/realm-server/prerender/prerender-app.ts
@@ -395,7 +395,9 @@ export function buildPrerenderApp(options: {
       let ac = new AbortController();
       const onClientClose = () => {
         if (ctxt.res.writableEnded) return;
-        ac.abort();
+        // The reason string rides on `PrerenderCancelledError.message`
+        // so cancel log lines say why the render was interrupted.
+        ac.abort('client disconnected');
       };
       ctxt.res.on('close', onClientClose);
       try {
@@ -656,7 +658,9 @@ export function buildPrerenderApp(options: {
     let ac = new AbortController();
     const onClientClose = () => {
       if (ctxt.res.writableEnded) return;
-      ac.abort();
+      // The reason string rides on `PrerenderCancelledError.message`
+      // so cancel log lines say why the render was interrupted.
+      ac.abort('client disconnected');
     };
     ctxt.res.on('close', onClientClose);
     try {
--- a/packages/realm-server/prerender/prerender-cancel.ts
+++ b/packages/realm-server/prerender/prerender-cancel.ts
@@ -6,8 +6,11 @@
 // that signal from its Koa handler and threads it down through
 // `Prerenderer` → `RenderRunner` → `PagePool.getPage` so queued
 // waits (render semaphore, per-affinity tab queue) can bail out
-// without entering render, and an in-flight render can be torn
-// down cleanly via `#maybeEvict`.
+// without entering render, and in-flight render steps race the
+// signal (`withTimeout` / `abortable`) so a mid-render abort
+// interrupts the step immediately — even against a wedged renderer
+// whose CDP calls would otherwise pin the visit until a protocol
+// timeout — and the Prerenderer tears the tab down.
 //
 // `PrerenderCancelledError` is the distinct-named rejection
 // callers look for to decide "this is a user cancel, route the tab
@@ -57,5 +60,48 @@ export function throwIfAborted(
       state,
       reason: typeof signal.reason === 'string' ? signal.reason : undefined,
     });
+  }
+}
+
+// Races `run()` against the signal so a page operation that can't
+// observe cancellation itself (a CDP evaluate against a wedged
+// renderer never returns) is abandoned the moment the caller goes
+// away, throwing `PrerenderCancelledError` instead of waiting on the
+// operation's own (protocol-level) timeout. The abandoned promise
+// keeps running until the cancel handler disposes the page — its
+// eventual rejection is absorbed by the race, never surfacing as an
+// unhandled rejection.
+export async function abortable<T>(
+  signal: AbortSignal | undefined,
+  run: () => Promise<T>,
+  state: PrerenderCancelState = 'rendering',
+): Promise<T> {
+  if (!signal) {
+    return run();
+  }
+  throwIfAborted(signal, state);
+  let onAbort: (() => void) | undefined;
+  try {
+    return await Promise.race([
+      run(),
+      new Promise<never>((_resolve, reject) => {
+        onAbort = () =>
+          reject(
+            new PrerenderCancelledError({
+              state,
+              reason:
+                typeof signal.reason === 'string' ? signal.reason : undefined,
+            }),
+          );
+        signal.addEventListener('abort', onAbort, { once: true });
+      }),
+    ]);
+  } finally {
+    // The signal outlives any one operation (it spans the whole
+    // request), so every race must detach its listener or a visit's
+    // worth of steps accumulates listeners on the shared signal.
+    if (onAbort) {
+      signal.removeEventListener('abort', onAbort);
+    }
   }
 }
--- a/packages/realm-server/prerender/prerender-cancel.ts
+++ b/packages/realm-server/prerender/prerender-cancel.ts
@@ -31,6 +31,10 @@ export type PrerenderCancelState = 'queued' | 'rendering' | 'releasing';
 export class PrerenderCancelledError extends Error {
   name = 'PrerenderCancelledError';
   state: PrerenderCancelState;
+  // Why the caller aborted (e.g. the close listener's disconnect
+  // reason). Carried as its own field — not only embedded in the
+  // message — so log lines can print it without re-parsing.
+  reason?: string;
   constructor(
     opts?: { state?: PrerenderCancelState; reason?: string } | string,
   ) {
@@ -44,6 +48,7 @@ export class PrerenderCancelledError extends Error {
     }
     super(reason ? `prerender cancelled: ${reason}` : 'prerender cancelled');
     this.state = state;
+    this.reason = reason;
   }
 }
 
--- a/packages/realm-server/prerender/prerender-cancel.ts
+++ b/packages/realm-server/prerender/prerender-cancel.ts
@@ -93,7 +93,17 @@ export async function abortable<T>(
                 typeof signal.reason === 'string' ? signal.reason : undefined,
             }),
           );
-        signal.addEventListener('abort', onAbort, { once: true });
+        // AbortSignal never replays `abort` for listeners attached
+        // after the fact, so re-check before attaching. No await
+        // separates the entry check above from this attach, so no
+        // abort can land between them — the re-check keeps the
+        // no-missed-abort guarantee local instead of resting on that
+        // ordering.
+        if (signal.aborted) {
+          onAbort();
+        } else {
+          signal.addEventListener('abort', onAbort, { once: true });
+        }
       }),
     ]);
   } finally {
--- a/packages/realm-server/prerender/prerenderer.ts
+++ b/packages/realm-server/prerender/prerenderer.ts
@@ -61,6 +61,11 @@ export class Prerenderer {
   #affinityIdleEvictMs: number;
   #semaphore: AsyncSemaphore;
   #restartInFlight: Promise<void> | null = null;
+  // Count of actual browser restarts performed (coalesced callers share
+  // one). Read by tests to assert a flow did NOT pay a restart — e.g. a
+  // visit following a cancelled render must acquire a fresh page rather
+  // than detour through the restart recovery lane.
+  #browserRestartCount = 0;
   // `clearCache` batch ownership (CS-10758 step 3). Maps affinityKey to
   // `{ batchId, since }` for the batch that currently owns the affinity's
   // warm loader. See `#gateClearCache` for the full policy. Populated on
@@ -243,7 +248,8 @@ export class Prerenderer {
     let elapsed = Date.now() - startedAt;
     log.info(
       `render cancelled after ${elapsed}ms in state=${err.state} ` +
-        `affinity=${affinityKey} target=${target}`,
+        `affinity=${affinityKey} target=${target} ` +
+        `reason=${err.reason ?? '<none>'}`,
     );
     if (err.state === 'rendering') {
       try {
@@ -256,9 +262,13 @@ export class Prerenderer {
         // instead of racing a background close at the pool ceiling
         // (which reads as "no standby available" and needlessly
         // routes that visit through the browser-restart recovery
-        // lane). The shared BrowserContext is retained so that next
-        // visit reuses the warm HTTP cache rather than paying the
-        // cold module-source waterfall.
+        // lane). That guarantee is scoped to visits arriving after
+        // this disposal: a waiter already holding the tab-queue
+        // lease when the cancel lands still receives the doomed
+        // page, since the lease handoff carries no revalidation.
+        // The shared BrowserContext is retained so the next visit
+        // reuses the warm HTTP cache rather than paying the cold
+        // module-source waterfall.
         await this.#pagePool.disposeAffinity(affinityKey, {
           retainSharedContext: true,
         });
@@ -293,6 +303,13 @@ export class Prerenderer {
       this.#renderRunner.clearIconMemo(affinityKey);
       log.debug(`batch ${batchId} released ownership of ${affinityKey}`);
     }
+  }
+
+  // Read-only observability accessor used by tests. Callers outside of
+  // tests should not rely on this; it's a debugging surface, not a
+  // stable API.
+  getBrowserRestartCount(): number {
+    return this.#browserRestartCount;
   }
 
   // Read-only observability accessor used by tests. Callers outside of
@@ -909,6 +926,7 @@ export class Prerenderer {
 
   async #runRestart(): Promise<void> {
     log.warn('Restarting prerender browser');
+    this.#browserRestartCount++;
     await this.#pagePool.closeAll();
     await this.#browserManager.restartBrowser();
     await this.#pagePool.warmStandbys().catch((e) => {
--- a/packages/realm-server/prerender/prerenderer.ts
+++ b/packages/realm-server/prerender/prerenderer.ts
@@ -247,18 +247,19 @@ export class Prerenderer {
     );
     if (err.state === 'rendering') {
       try {
-        // Same disposal envelope as the render-error eviction path:
-        // mark the entries closing synchronously (so a queued waiter
-        // for this affinity is routed to a fresh page instead of
-        // adopting the abandoned — possibly wedged — one), close them
-        // in the background (a wedged tab's teardown must not
-        // serialize this cancel handler), and keep the shared
-        // BrowserContext so the next visit reuses its warm HTTP
-        // cache instead of paying the cold module-source waterfall.
-        // Closing the page is also what interrupts any CDP call the
-        // cancelled render abandoned mid-flight.
+        // Closing the page is what interrupts any CDP call the
+        // cancelled render abandoned mid-flight. The teardown is
+        // awaited: the caller is gone, so nothing is delayed by
+        // waiting, and by the time the cancellation propagates the
+        // pool's slot accounting has already released this page —
+        // the affinity's next visit materializes a fresh page
+        // instead of racing a background close at the pool ceiling
+        // (which reads as "no standby available" and needlessly
+        // routes that visit through the browser-restart recovery
+        // lane). The shared BrowserContext is retained so that next
+        // visit reuses the warm HTTP cache rather than paying the
+        // cold module-source waterfall.
         await this.#pagePool.disposeAffinity(affinityKey, {
-          awaitIdle: false,
           retainSharedContext: true,
         });
         this.#renderRunner.clearAuthCache(affinityKey);
--- a/packages/realm-server/prerender/prerenderer.ts
+++ b/packages/realm-server/prerender/prerenderer.ts
@@ -247,7 +247,20 @@ export class Prerenderer {
     );
     if (err.state === 'rendering') {
       try {
-        await this.#pagePool.disposeAffinity(affinityKey);
+        // Same disposal envelope as the render-error eviction path:
+        // mark the entries closing synchronously (so a queued waiter
+        // for this affinity is routed to a fresh page instead of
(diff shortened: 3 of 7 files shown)


packages/realm-server/prerender/render-runner.ts
```diff
@@ -433,6 +437,7 @@ export class RenderRunner {
         },
         opts?.timeoutMs,
         this.#profileContext(affinityKey, command, 'command-runner'),
+        signal,
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed. The catch converted a cancellation into a `status: 'error'` response, so the Prerenderer's cancel handler never disposed the abandoned tab and the pool got it back still busy. The catch rethrows `PrerenderCancelledError`, matching the module/screenshot/visit attempts, whose errors propagate uncaught to the per-path cancel branches in the Prerenderer.

packages/realm-server/prerender/render-runner.ts
```diff
@@ -1759,6 +1781,7 @@ export class RenderRunner {
               },
               opts?.timeoutMs,
               this.#profileContext(affinityKey, url, 'file isolated/0', jobId),
+              signal,
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed. The stash cleanup runs under the same signal race as the render steps, with the cancellation swallowed so the `finally` never masks the original throw. Skipping the cleanup on cancel is safe: the stash exists to keep stale file data from bleeding into a reused tab, and a cancelled visit's page is disposed immediately after — there is no next render to protect.

packages/realm-server/prerender/prerender-cancel.ts
```diff
@@ -59,3 +62,46 @@ export function throwIfAborted(
...
+      }),
+    ]);
+  } finally {
+    // The signal outlives any one operation (it spans the whole
+    // request), so every race must detach its listener or a visit's
+    // worth of steps accumulates listeners on the shared signal.
+    if (onAbort) {
+      signal.removeEventListener('abort', onAbort);
+    }
+  }
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] The gap isn't reachable in this control flow: the entry `throwIfAborted`, the invocation of `run()`, and the listener attach all execute within one synchronous section — no `await` separates them, and an abort (which arrives via the response socket's `close` event) can only be dispatched between turns of the event loop, never inside a synchronous section. `run()` being invoked before the attach therefore can't observe an abort the entry check missed.

That said, the guarantee rested on a non-local property (no intervening await), so the executor re-checks `signal.aborted` before attaching and rejects immediately when set — a refactor that introduces an await between the check and the attach can no longer open a missed-abort window.

packages/realm-server/prerender/utils.ts
```diff
@@ -1634,11 +1640,43 @@ export async function withTimeout<T>(
...
+            reject(
+              new PrerenderCancelledError({
+                state: 'rendering',
+                reason:
+                  typeof signal.reason === 'string' ? signal.reason : undefined,
+              }),
+            );
+          signal.addEventListener('abort', abortListener, { once: true });
+        }),
+      );
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Same analysis as the `abortable` thread: no `await` separates the entry check from the listener attach, and abort dispatch requires an event-loop turn, so the window doesn't exist in this control flow. The racer's executor re-checks `signal.aborted` before attaching all the same, keeping the no-missed-abort guarantee local rather than resting on that ordering.

packages/realm-server/prerender/prerender-app.ts
```diff
@@ -395,7 +395,9 @@ export function buildPrerenderApp(options: {
       let ac = new AbortController();
       const onClientClose = () => {
         if (ctxt.res.writableEnded) return;
-        ac.abort();
+        // The reason string rides on `PrerenderCancelledError.message`
+        // so cancel log lines say why the render was interrupted.
+        ac.abort('client disconnected');
```
habdelra, line 400: [Claude Code 🤖] **The abort reason never reaches a log line — the stated motivation for passing a reason isn't wired through.** This comment (and its twin on the composite-visit route's close listener below) says the reason string "rides on `PrerenderCancelledError.message` so cancel log lines say why the render was interrupted". The first half is true and unit-pinned: `throwIfAborted`, `withTimeout`, and `abortable` all thread `signal.reason` into the error, and the cancellation unit test asserts the message contains the reason. But no cancel-path log line prints that message. I enumerated the sites: `Prerenderer.#handlePrerenderCancel` logs `render cancelled after <n>ms in state=<state> affinity=<key> target=<url>` — no reason — and the HTTP-layer swallow in the route catch logs only `prerender cancelled before completion requestId=<id>` at debug. So the reason is captured but observable nowhere.

**The way out.** One line in `Prerenderer.#handlePrerenderCancel`: append the error's message (or a `reason=` field) to the `render cancelled` line. Alternatively trim both comments to stop claiming log coverage — but wiring the reason through is the better half, since distinguishing "client disconnected" from any future second abort source is the point of passing a reason at all.

**Scope.** Regression against this PR's own stated behavior (the abort mechanics themselves work). Non-blocking, but it belongs in this PR because two code comments and the description assert the log behavior.

packages/realm-server/prerender/prerenderer.ts
```diff
@@ -247,7 +247,21 @@ export class Prerenderer {
...
+        // the affinity's next visit materializes a fresh page
+        // instead of racing a background close at the pool ceiling
+        // (which reads as "no standby available" and needlessly
+        // routes that visit through the browser-restart recovery
+        // lane). The shared BrowserContext is retained so that next
+        // visit reuses the warm HTTP cache rather than paying the
+        // cold module-source waterfall.
+        await this.#pagePool.disposeAffinity(affinityKey, {
+          retainSharedContext: true,
+        });
```
habdelra, line 274: [Claude Code 🤖] **Mostly confirmation, with one hole this comment shouldn't imply is closed: a waiter already queued on the doomed tab.**

**What's verified right.** The ordering claim holds: `RenderRunner`'s visit `finally { release() }` runs before the rejection reaches this handler, so the pool's slot accounting has indeed released the page by the time `disposeAffinity` runs. `retainSharedContext: true` matches the retention `#evictAffinity` already uses on the render-error path, and leaving `retainConsoleErrors` unset is correct here — no caller remains to read an error doc. The awaitIdle divergence from the eviction path is also sound: eviction runs mid-visit with a live caller owed a response; this path has nobody to delay. And the eviction invariant is untouched — a mid-render cancel rejects with `PrerenderCancelledError`, bypassing `#maybeEvict` entirely, so eviction on render errors stays gated on `#evictionReason` (timeout title / explicit evict flag).

**The hole.** `release()` → `releaseTab()` hands the tab-queue lease to the next waiter *before* the cancellation propagates here. `PagePool.#selectEntryForAffinity` filters `entry.closing` only at selection time; a same-affinity waiter already parked on that entry's `TabQueue` receives the lease with no revalidation, and this handler's `disposeAffinity` then closes the page under it. That waiter's first CDP call rejects with a protocol error — not a `PrerenderCancelledError` — so `prerenderVisit`'s catch routes it through `#restartBrowser()`: the recovery lane this comment describes avoiding, now paid by a bystander.

**Scope.** Pre-existing, not this PR's bug — pass-boundary cancels have always had the same release-then-dispose ordering — but this PR makes `rendering`-state cancels fire more often (that's its purpose), so the window gets more traffic. Follow-up, not this PR: after a `TabQueue` acquire resolves, re-check `entry.closing` (or membership in `#affinityPages`) and re-select instead of returning the doomed entry. Nothing to change here beyond knowing the "next visit materializes a fresh page" guarantee is scoped to visits that arrive after disposal completes.

packages/realm-server/prerender/utils.ts
```diff
@@ -1588,7 +1589,11 @@ export async function withTimeout<T>(
   fn: () => Promise<T>,
   timeoutMs = cardRenderTimeout,
   profileContext?: RenderProfileContext,
+  signal?: AbortSignal,
```
habdelra, line 1592: [Claude Code 🤖] Confirmation — no change requested. Three properties verified here, plus the condition under which each would rot:

**Coverage.** The signal only helps if every step passes it, and a missed site regresses silently (the parameter is optional). Enumerated: `withTimeout` is imported by exactly one production file, `render-runner.ts`, and all ten call sites there pass `signal` — the command-runner wait, screenshot, module, file-extract, the card `runTimedStep` funnel, card isolated, file isolated, file icon, file head, and the file `steps` loop. (`trace-capture.ts` declares an unrelated private `withTimeout`.) This rots the day someone adds an eleventh call site without the argument; if this signature keeps growing, folding `signal` into a required options parameter would turn that omission into a type error.

**Rejection containment.** Losing racers can't escape as unhandled rejections: `Promise.race` subscribes to every element, so the abandoned `renderFn()` promise's eventual "Target closed" rejection after disposal is absorbed whether the abort or the timer won. The entry `throwIfAborted` sits before `activeRenderCount++`, so a pre-aborted call can't unbalance the counter (the decrement lives in the `finally` of a `try` the throw never enters). The per-step listener detach in `finally` keeps the request-scoped signal from accumulating one listener per step.

**Timeout path unchanged.** The abort racer is pushed only when `signal` is present, and the timeout marker, its diagnostics, and the canonical timeout error doc are byte-identical when the signal never fires — pinned by the unit test for that case. Ran the cancellation unit suite at this head: 25/25 pass with a clean exit, which also pins the no-unhandled-rejection property (the suite fails on a late one).

packages/realm-server/tests/prerendering-test.ts
```diff
@@ -7534,6 +7534,69 @@ module(basename(import.meta.filename), function () {
...
+        renderOptions: { cardRender: true },
+      });
+      assert.ok(
+        followUp.response.card?.isolatedHTML?.includes('Maple'),
+        'follow-up visit on the same affinity renders normally',
+      );
+      assert.notOk(
+        followUp.response.pageUnusableError,
+        'follow-up visit is not poisoned by the cancelled render',
+      );
```
habdelra, line 7598: [Claude Code 🤖] **What this follow-up pins is narrower than "the next visit doesn't detour through the restart lane".** If the follow-up *did* take the restart lane, these assertions would still pass: `prerenderVisit`'s catch restarts the browser and retries, and the retry would render the card normally with no `pageUnusableError`. So the test pins "the affinity stays usable after a mid-render cancel" — valuable, and the right primary claim — but it cannot distinguish a clean fresh-page acquisition from a restart-recovery detour.

The abort-timing half of the test is deterministic, verified: `simulateTimeoutMs` pins the isolated-html step via a delay inside the `withTimeout` callback, and `onTabAcquired` fires after the queued checkpoint, so the delayed abort can only surface as a `rendering`-state cancel well inside the step.

**If pinning the no-restart property matters** — it is the reason the cancel handler awaits disposal — the cheap strengthening is asserting the follow-up's acquisition didn't pay a browser restart: bound `followUp.timings.launchMs` well below restart cost, or assert on a restart counter/log hook if one is exposed to tests.

**Scope.** Follow-up-grade test hardening, non-blocking; as written this is still a real regression net for the affinity-poisoning failure.

habdelra (commented):
[Claude Code 🤖] Review lens: the cancellation lifecycle end to end — listener hygiene on the request-scoped signal, rejection containment for abandoned page operations, coverage of the signal across every render step, and the cancel path's interaction with the page pool's slot accounting and the eviction contract. Every claim below was verified against the checked-out head; the cancellation unit suite and the package typecheck were run locally.

**Bottom line: no blocking issues. One fix recommended in this PR — the abort reason is captured but never logged, contradicting two new code comments and the description. Everything else is confirmation or pre-existing follow-up material.**

What lands right (verified, not courtesy):

- All ten `withTimeout` call sites in `render-runner.ts` receive the signal, and `withTimeout` has no other production importer — no step is left observing cancellation only at pass boundaries. Detail in the `utils.ts` thread.
- The eviction invariant is preserved: a mid-render cancel rejects with `PrerenderCancelledError` and bypasses `#maybeEvict` entirely, so eviction on render *errors* remains gated on `#evictionReason` (timeout title / explicit evict flag), untouched by this diff. The cancel-path disposal is the pre-existing `rendering`-state contract; this PR tunes its retention and documents the awaited teardown. Detail in the `prerenderer.ts` thread.
- Rejection containment is structural, not incidental: both races subscribe to the abandoned operation's promise, so its post-disposal "Target closed" rejection is absorbed whichever racer won. Unit-pinned for `abortable`; same construct in `withTimeout`. The 25-test cancellation suite passes at this head with a clean exit.
- The abort reason string is safe end to end: every consumer of this signal (`throwIfAborted`, `withTimeout`, `abortable`, `TabQueue.acquire`, `AsyncSemaphore.acquire`) guards `typeof signal.reason === 'string'` and constructs its own `PrerenderCancelledError`; nothing calls native `signal.throwIfAborted()` (which would throw the raw string) or hands this signal to an API expecting a `DOMException` reason. Verified by grep across the prerender directory.
- The `runCommandAttempt` rethrow closes the one catch in `render-runner.ts` that could swallow a cancellation — the only other catch is `#evictAffinity`'s dispose guard, verified by enumerating catch blocks; the module/screenshot/visit attempts propagate rejections naturally to the per-path cancel branches.
- The listener re-check at attach time: independently traced the control flow and concur with the existing threads — the entry check, `run()` invocation, and attach form one synchronous section, and abort dispatch requires an event-loop turn, so the re-check hardens an already-unreachable window rather than fixing a live bug.

Recommendations:

1. Wire the abort reason into the `render cancelled` log line, or drop the claim from the two route comments — see the `prerender-app.ts` thread. In this PR.
2. File the queued-waiter hole in the release-then-dispose window as a page-pool follow-up: a waiter already parked on the doomed entry's `TabQueue` can be handed the tab before disposal closes it, landing that bystander in the browser-restart lane — see the `prerenderer.ts` thread. Pre-existing.
3. Optionally strengthen the follow-up-visit assertion so it can distinguish fresh-page acquisition from restart recovery — see the `prerendering-test.ts` thread. Follow-up.

Verification notes: cancellation unit suite 25/25 with clean exit; `ember-tsc --noEmit` clean in `packages/realm-server`; the real-Chrome integration module was not re-run here (another realm-server test run was active and the suite is single-process), so that module rests on the author's run and CI.

Adjacent, out of scope: the timeout path's hang diagnostics are not signal-gated — a caller that disconnects after the timer has already won still pays several seconds of CDP probing for nobody. Same class of waste this change removes elsewhere, much smaller magnitude; only worth touching if the diagnostics burst ever shows up in saturation triage.

packages/realm-server/prerender/prerender-app.ts
habdelra: [Claude Code 🤖] Confirmed — the reason stopped at the error's message and no cancel-path log line printed it. The cancellation error carries `reason` as its own field (unit-pinned alongside the message form), and `#handlePrerenderCancel`'s `render cancelled` line prints `reason=<value>` (`<none>` when the signal aborted without one), so a client disconnect is distinguishable in the logs from any other abort source. The close-listener comments' claim is accurate against this wiring.

packages/realm-server/prerender/prerenderer.ts
habdelra: [Claude Code 🤖] Agreed on all points. The disposal comment names the boundary explicitly: the fresh-page guarantee covers visits arriving after disposal, while a waiter already holding the tab-queue lease receives the doomed page, since the lease handoff carries no revalidation. Post-acquire revalidation (or draining parked waiters with a cancellation at dispose time, so they re-select instead of discovering the closed page via a protocol error) is filed as a follow-up on the internal tracker.

packages/realm-server/prerender/utils.ts
habdelra: [Claude Code 🤖] Thanks for the enumeration — especially pinning the rot condition on the optional parameter. Folding `signal` into a required options argument so a missed call site becomes a type error is the right shape if this signature grows again; leaving it as-is while the call-site count is stable and single-file.

packages/realm-server/tests/prerendering-test.ts
habdelra: [Claude Code 🤖] Confirmed — the success assertions alone can't tell a clean acquisition from restart-lane recovery. The Prerenderer exposes a browser-restart counter as a test observability accessor (incremented per actual restart; coalesced callers share one), and the test asserts the counter is unchanged across the cancelled visit and its follow-up — pinning the no-restart property deterministically rather than through a timing bound.

FadhlanR (approved):

PR #5558, opened 2026-07-20
Wire the ai-bot to-device streaming channel end-to-end (ai-bot side)

lukemelia:
## Summary

- **CS-12261** — adds `APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE = 'app.boxel.response-stream'` and the `AppBoxelResponseStreamContent` interface (fields: `roomId`, `parentEventId`, `sequence`, `body`, `reasoning`, `toolRequests`) to `packages/runtime-common/matrix-constants.ts`.
- **CS-12262** — adds `APP_BOXEL_ORIGINATING_DEVICE_ID_KEY` and stamps the sending device id onto every boxel user-prompt event (`sendMessage` in `packages/host/app/services/matrix-service.ts`). Also adds the optional key to the `CardMessageContent` interface in `packages/base/matrix-event.gts`.
- **CS-12263** — implements `AI_BOT_STREAMING_MODE=to-device` in the ai-bot's `Responder`: throttled state changes route to `matrixClient.sendToDevice()` targeted at the (userId, deviceId) that composed the prompt, carrying an `app.boxel.response-stream` payload. The **final** consolidated event still lands as a room edit (routed via `flush` → `sendMessageEvent`, not the throttled internal) so durable state is always in the room — to-device previews are ephemeral by design.
- Also merged in **CS-12260** (`off` mode) as a base commit, since CS-12263 sits on top of the `streamingMode` gate that CS-12260 introduced. #5556 remains open with the standalone CS-12260 patch; whichever of the two lands first, the other's diff auto-shrinks.

Together this is the full ai-bot side of the [Improve Synapse Performance by Killing Streaming](https://linear.app/cardstack/project/improve-synapse-performance-by-killing-streaming-1d7d8ac245b1) project's to-device path. The host receiver ([CS-12264](https://linear.app/cardstack/issue/CS-12264)) is the last remaining wire — until it lands, running with `AI_BOT_STREAMING_MODE=to-device` produces to-device previews the browser doesn't yet consume, and the UX falls back to "thinking → complete response arrives at once", identical to `off` mode.

## Design notes

- **Final always lands as a room event.** The throttled internal checks `responseState.isStreamingFinished` and routes to `sendMessageEvent` (room edit) whenever true, regardless of mode. `flush` in `finalize` uses `sendMessageEvent` directly. So the leading-edge case where the throttled internal fires during `finalize` still produces a room edit, not a to-device preview.
- **Older-client fallback.** When mode is `to-device` but the prompt lacks the originating device id (older client without CS-12262), the responder falls back to `off` behavior — no mid-stream sends, one final room edit. Deployment can enable `to-device` mode without waiting for every user's session to refresh.
- **Preview payload shape.** Full accumulated body / reasoning / toolRequests, with a monotonic per-turn `sequence` for last-writer-wins reconciliation on the client. Not deltas — gaps and reordering are tolerated by design.
- **Encryption.** to-device previews use the plaintext `client.sendToDevice()` path today. Acceptable for Cardstack's own Synapse (the operator can see the traffic either way) and for previews that will also land in the room as the final. For federated deployments or a stronger threat model, swap to `client.encryptAndSendToDevice()` as follow-up.
- **`toolRequests` shape.** Previews run `toolCalls` through the same `toCommandRequest()` the room event uses, so both channels carry `{ id, name, arguments: <object> }`; `arguments` is `{}` until the streamed JSON completes. Typed `unknown[]` on the shared interface (the receiver, CS-12264, validates at the boundary), but the runtime shape now matches the room event.
- **No `isFinal` field.** The final state always lands as a room edit, never a to-device preview, so an `isFinal` preview flag would never be set — the room edit replacing the thinking placeholder is already the terminal signal. Dropped from the schema; reintroduce if the receiver ever needs an explicit "discard preview state now" signal.
- **Multi-step turns keep streaming.** The originating device id is taken from the triggering event, or — when a continuation is triggered by a tool / code-patch result that carries none — from the most recent stamped user-prompt event in the turn's history. So a turn that continues after a tool call still previews to the device that started it, rather than dropping to `off` behavior.
- **`sendToDevice` takes a `Map`, not a plain object.** matrix-js-sdk's `sendToDevice(eventType, contentMap)` requires `Map<userId, Map<deviceId, content>>` and iterates it internally (`recursiveMapToObject`); a plain nested object throws `TypeError: contentMap is not iterable` before anything reaches the wire — swallowed by the preview send's `try/catch`, so previews would have failed silently. The Responder builds the real `Map`. `FakeMatrixClient.sendToDevice` now rejects a non-`Map` argument so a regression fails the unit suite the way it fails in production.

## Test plan

- [x] Ai-bot: 22 `Responding` tests pass, including two new ones — `\`to-device\` streaming mode: streams previews to the target device and lands one final consolidated room event`, and `\`to-device\` streaming mode without a target device falls back to \`off\` behavior`. The CS-12260 `off` mode test also still passes.
- [x] All 5 `Responder Cancellation` tests pass — `finalize({isCanceled:true})` still routes the final send to the room-edit path.
- [x] Typecheck: no errors touching any of the modified files (the `../base/...` / `@cardstack/boxel-ui/*` errors are the pre-existing local-dist issue).
- [x] Verified the `FakeMatrixClient` `Map` guard bites: temporarily reverting the Responder to the plain-object payload flips the `to-device` streaming unit test to failing (`not ok`), because the swallowed `TypeError` leaves no preview sent. Restored, back to green.
- [ ] Matrix e2e (`packages/matrix/tests/response-stream-to-device.spec.ts`): drives a real matrix-js-sdk client against the dockerized Synapse — delivers a `response-stream` preview to the target device and asserts the real SDK rejects a plain-object `contentMap`. Typechecks + lints locally; first real run to be watched in CI (needs the full Synapse + realm-server stack, not run locally).
- [ ] Staging: set `AI_BOT_STREAMING_MODE=to-device` on the ai-bot task after CS-12264 lands. Verify (a) the browser renders the streaming preview, (b) only 2 room events land per turn, (c) other devices in the same account see the final response but no live preview.

## Commits

- `f85e59c65e` — CS-12261: `app.boxel.response-stream` schema
- `7dd96110a8` — CS-12262: stamp originating device id on prompts
- `91eb30c3de` — merge CS-12260 (`off` mode) as base
- `0ff3c38c15` — CS-12263: `to-device` streaming mode
- `c5ad13b90b` — fix: send previews as a `Map`; harden the fake + add a real-SDK matrix e2e boundary test
- `16619683b9` — review findings: normalize preview `toolRequests` via `toCommandRequest`; drop `isFinal`; thread the originating device id through multi-step turns

### DIFF
--- a/packages/ai-bot/lib/responder.ts
+++ b/packages/ai-bot/lib/responder.ts
@@ -1,5 +1,9 @@
 import { logger } from '@cardstack/runtime-common';
-import { APP_BOXEL_CODE_PATCH_CORRECTNESS_MSGTYPE } from '@cardstack/runtime-common/matrix-constants';
+import {
+  APP_BOXEL_CODE_PATCH_CORRECTNESS_MSGTYPE,
+  APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE,
+  type AppBoxelResponseStreamContent,
+} from '@cardstack/runtime-common/matrix-constants';
 import { isToolOrCodePatchResult } from '@cardstack/runtime-common/ai';
 
 import { errorReporter } from './sentry.ts';
@@ -49,7 +53,18 @@ export class Responder {
     );
   }
 
-  constructor(client: MatrixClient, roomId: string, agentId: string) {
+  private client: MatrixClient;
+  private streamPreviewTarget: { userId: string; deviceId: string } | undefined;
+  private _streamPreviewSequence = 0;
+
+  constructor(
+    client: MatrixClient,
+    roomId: string,
+    agentId: string,
+    streamPreviewTarget?: { userId: string; deviceId: string },
+  ) {
+    this.client = client;
+    this.streamPreviewTarget = streamPreviewTarget;
     this.matrixResponsePublisher = new MatrixResponsePublisher(
       client,
       roomId,
@@ -81,6 +96,19 @@ export class Responder {
     return 'room-edits';
   }
 
+  // Whether mid-turn state changes should trigger a throttled preview send.
+  // In `off` mode we never send mid-turn events. In `to-device` mode without a
+  // preview target (older client that didn't stamp its device id on the prompt)
+  // we also skip mid-turn — the sensible fallback since we can't target a
+  // preview at anyone in particular. The final consolidated room edit still
+  // lands from `finalize`/`flush`.
+  private get shouldStreamMidTurn(): boolean {
+    const mode = this.streamingMode;
+    if (mode === 'off') return false;
+    if (mode === 'to-device' && !this.streamPreviewTarget) return false;
+    return true;
+  }
+
   async ensureThinkingMessageSent() {
     await this.matrixResponsePublisher.ensureThinkingMessageSent();
   }
@@ -96,10 +124,52 @@ export class Responder {
   sendMessageEventWithThrottlingInternal: () => unknown = throttle(
     () => {
       this.needsMessageSend = false;
-      this.sendMessageEvent();
+      // The final consolidated event always lands as a room edit — never a
+      // to-device preview — because to-device is ephemeral and durable state
+      // must be in the room. Intermediate previews in to-device mode go over
+      // sendToDevice targeted at the originating device only.
+      if (
+        this.streamingMode === 'to-device' &&
+        this.streamPreviewTarget &&
+        !this.responseState.isStreamingFinished
+      ) {
+        this.sendToDevicePreview();
+      } else {
+        this.sendMessageEvent();
+      }
     },
     Number(process.env.AI_BOT_STREAM_THROTTLE_MS ?? 250),
   );
+
+  private sendToDevicePreview = async (): Promise<void> => {
+    if (!this.streamPreviewTarget) return;
+    const parentEventId = this.matrixResponsePublisher.originalResponseEventId;
+    if (!parentEventId) {
+      // Haven't sent the thinking placeholder yet — the client would have
+      // nothing to attach the preview to.
+      return;
+    }
+    const payload: AppBoxelResponseStreamContent = {
+      roomId: this.matrixResponsePublisher.roomId,
+      parentEventId,
+      sequence: this._streamPreviewSequence++,
+      body: this.responseState.latestContent ?? '',
+      reasoning: this.responseState.latestReasoning ?? '',
+      toolRequests: this.responseState.toolCalls ?? [],
+      isFinal: false,
+    };
+    try {
+      await this.client.sendToDevice(APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE, {
+        [this.streamPreviewTarget.userId]: {
+          [this.streamPreviewTarget.deviceId]: payload,
+        },
+      } as any);
+    } catch (e) {
+      // Preview loss is non-fatal; the final room edit still lands. Log at
+      // debug so a wedged homeserver doesn't spam sentry.
+      log.debug('to-device response-stream preview send failed', e);
+    }
+  };
 
   sendMessageEvent = async () => {
     // Only send if the delta is meaningful, unless we are finalizing.
@@ -169,7 +239,7 @@ export class Responder {
       isStreamingFinished: this.responseState.isStreamingFinished,
       responseStateChanged,
     });
-    if (responseStateChanged && this.streamingMode !== 'off') {
+    if (responseStateChanged && this.shouldStreamMidTurn) {
       await this.sendMessageEventWithThrottling();
     }
 
--- a/packages/ai-bot/lib/responder.ts
+++ b/packages/ai-bot/lib/responder.ts
@@ -158,12 +158,20 @@ export class Responder {
       toolRequests: this.responseState.toolCalls ?? [],
       isFinal: false,
     };
+    // matrix-js-sdk's sendToDevice takes a Map<userId, Map<deviceId, content>>
+    // and iterates it internally — a plain nested object throws
+    // `TypeError: contentMap is not iterable` at runtime.
+    const contentMap = new Map([
+      [
+        this.streamPreviewTarget.userId,
+        new Map([[this.streamPreviewTarget.deviceId, payload]]),
+      ],
+    ]);
     try {
-      await this.client.sendToDevice(APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE, {
-        [this.streamPreviewTarget.userId]: {
-          [this.streamPreviewTarget.deviceId]: payload,
-        },
-      } as any);
+      await this.client.sendToDevice(
+        APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE,
+        contentMap,
+      );
     } catch (e) {
       // Preview loss is non-fatal; the final room edit still lands. Log at
       // debug so a wedged homeserver doesn't spam sentry.
--- a/packages/ai-bot/lib/responder.ts
+++ b/packages/ai-bot/lib/responder.ts
@@ -15,7 +15,9 @@ import type { FunctionToolCall } from '@cardstack/runtime-common/helpers/ai';
 import type OpenAI from 'openai';
 import type { ChatCompletionSnapshot } from 'openai/lib/ChatCompletionStream';
 import type { MatrixEvent as DiscreteMatrixEvent } from 'matrix-js-sdk';
-import MatrixResponsePublisher from './matrix/response-publisher.ts';
+import MatrixResponsePublisher, {
+  toCommandRequest,
+} from './matrix/response-publisher.ts';
 import ResponseState from './response-state.ts';
 import type { MatrixClient } from 'matrix-js-sdk';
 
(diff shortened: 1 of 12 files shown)


packages/ai-bot/lib/responder.ts
```diff
@@ -88,11 +124,53 @@ export class Responder {
...
+      roomId: this.matrixResponsePublisher.roomId,
+      parentEventId,
+      sequence: this._streamPreviewSequence++,
+      body: this.responseState.latestContent ?? '',
+      reasoning: this.responseState.latestReasoning ?? '',
+      toolRequests: this.responseState.toolCalls ?? [],
+      isFinal: false,
+    };
+    try {
+      await this.client.sendToDevice(APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE, {
```
lukemelia: [Claude Code 🤖] **This to-device send throws at runtime against the real SDK — the whole preview feature is silently dead in prod.** `matrix-js-sdk@38.3.0` types `sendToDevice(eventType, contentMap)` with `SendToDeviceContentMap = Map<string, Map<string, Record<string, any>>>`, and its implementation both calls `utils.recursiveMapToObject(contentMap)` and does `for (const [userId, deviceMessages] of contentMap)` — both throw `TypeError: contentMap is not iterable` when handed the plain nested object built here. The `as any` cast hides the type error, and the surrounding `try/catch` swallows the `TypeError` at `log.debug`, so every preview send fails silently and no preview ever reaches the client. Tests pass only because `FakeMatrixClient.sendToDevice` stores the object without iterating it.

```suggestion
      await this.client.sendToDevice(APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE, new Map([
        [this.streamPreviewTarget.userId, new Map([[this.streamPreviewTarget.deviceId, payload]])],
      ]));
```

packages/runtime-common/matrix-constants.ts
```diff
@@ -135,6 +135,38 @@ export const APP_BOXEL_HAS_CONTINUATION_CONTENT_KEY =
...
+  // keys the client's preview state so concurrent turns don't collide.
+  parentEventId: string;
+  // Monotonic per turn. Payload carries the full accumulated state (not
+  // deltas), so gaps or reordering are non-fatal — last-writer-wins by sequence.
+  sequence: number;
+  body: string;
+  reasoning: string;
+  // Same wire shape as APP_BOXEL_TOOL_REQUESTS_KEY on the room event, though
+  // each request's `arguments` JSON may still be partial during streaming.
+  toolRequests: unknown[];
```
lukemelia, line 160: [Claude Code 🤖] **`toolRequests` does not actually match the room-event wire shape the comment claims.** The room event serializes tool calls through `toCommandRequest()` in `response-publisher.ts`, producing `Partial<ToolRequest>` = `{ id, name, arguments: <parsed object> }`. But `sendToDevicePreview` (responder.ts:158) sends `this.responseState.toolCalls` verbatim — raw OpenAI `ChatCompletionSnapshot.Choice.Message.ToolCall` objects shaped `{ index, id, type, function: { name, arguments: <JSON string> } }`. So a client that parses `toolRequests` the same way for both channels (name at `.name` vs `.function.name`, args as object vs JSON string) will mis-read every preview. Either run the previews through `toCommandRequest` too, or fix this comment to document the raw OpenAI shape.

packages/runtime-common/matrix-constants.ts
```diff
@@ -135,6 +135,38 @@ export const APP_BOXEL_HAS_CONTINUATION_CONTENT_KEY =
...
+  // Set on the final preview immediately before ai-bot flushes the room edit;
+  // lets the client discard preview state promptly.
+  isFinal: boolean;
```
lukemelia: [Claude Code 🤖] **`isFinal` is documented as a signal that is never actually sent.** The comment says it is “Set on the final preview immediately before ai-bot flushes the room edit,” but the final consolidated state always goes out as a room edit via `sendMessageEvent` — the throttle branch in responder.ts (`... && !this.responseState.isStreamingFinished`) routes the final send to a room edit, never a to-device preview, and `sendToDevicePreview` hardcodes `isFinal: false`. So no preview ever carries `isFinal: true`, and a client relying on it to “discard preview state promptly” will never get the signal. Either emit a final `isFinal: true` to-device preview, or drop the field and this comment.

packages/ai-bot/tests/helpers/fake-matrix-client.ts
```diff
@@ -97,6 +101,19 @@ export class FakeMatrixClient extends MatrixClient {
     return this.sentEvents;
   }
 
+  async sendToDevice(
```
lukemelia, line 105: [Claude Code 🤖] **This fake masks the real `sendToDevice` type contract, letting a prod-breaking payload pass tests.** The override types `contentMap: any` and just pushes it onto an array, so the tests never exercise the `Map<string, Map<...>>` shape the real `matrix-js-sdk@38.3.0` client requires (see the responder.ts:162 finding). Consider having the fake assert `contentMap instanceof Map` (and normalize via `recursiveMapToObject`) before storing, so a plain-object payload fails the test the way it fails in production.

packages/ai-bot/main.ts
```diff
@@ -467,7 +468,26 @@ Common issues are:
               ? JSON.parse(event.getContent().data)
               : event.getContent().data;
           const agentId = contentData.context?.agentId;
-          const responder = new Responder(client, room.roomId, agentId);
+          // Route to-device streaming previews (see APP_BOXEL_STREAMING_MODE
+          // handling in Responder) at the device that composed this prompt.
+          // Absent on older clients that didn't stamp the id — Responder falls
+          // back to the `off`-mode behavior in that case.
+          const originatingDeviceId = (
```
lukemelia: [Claude Code 🤖] The originating device id is only read from the single triggering `event`. For turns triggered by a tool/code-patch **result** event (`isToolOrCodePatchResult`), that event carries no `app.boxel.originating-device-id`, so `streamPreviewTarget` is `undefined` and `shouldStreamMidTurn` falls back to off — meaning multi-step turns that continue after a tool call stream no previews at all, even on the same device that started the turn. If that is intended, worth a comment; if not, the device id likely needs to be threaded from the originating user-prompt event of the turn rather than the immediate trigger.

packages/ai-bot/lib/responder.ts
lukemelia: [Claude Code 🤖] Fixed in c5ad13b90b. The Responder now builds a real `Map<userId, Map<deviceId, content>>` and the `as any` is gone, so the compiler polices this call site. To stop it regressing silently: `FakeMatrixClient.sendToDevice` now rejects a non-`Map` contentMap (see the fake finding), and a new `packages/matrix/tests/response-stream-to-device.spec.ts` drives the real matrix-js-sdk client against the dockerized Synapse — it delivers a preview to the target device and asserts the real SDK rejects a plain-object contentMap. Verified the guard bites: temporarily reverting to the plain object flips the to-device unit test to `not ok`.

packages/runtime-common/matrix-constants.ts
lukemelia: [Claude Code 🤖] Fixed in 16619683b9. Previews now run `toolCalls` through the same `toCommandRequest()` the room event uses (exported from `response-publisher.ts` for reuse), so both channels carry `{ id, name, arguments: <object> }`; `arguments` comes through as `{}` until the streamed JSON completes. Added a unit test asserting the normalized shape, and updated this schema comment to match.

packages/runtime-common/matrix-constants.ts
lukemelia: [Claude Code 🤖] Fixed in 16619683b9 by dropping the `isFinal` field and its comment. As you note, the final state always lands as a room edit (never a to-device preview), so the field was never set true — and the room edit replacing the thinking placeholder is already the terminal signal for the client. If the CS-12264 receiver turns out to want an explicit "discard preview state now" signal, we can reintroduce it then.

packages/ai-bot/tests/helpers/fake-matrix-client.ts
lukemelia: [Claude Code 🤖] Fixed in c5ad13b90b. `sendToDevice` now throws a `TypeError` on a non-`Map` contentMap (mirroring the real SDK) and normalizes via `recursiveMapToObject` for storage, so existing assertions keep reading a plain object while a plain-object *payload* now fails the suite the way it fails in prod. Confirmed: reverting the Responder to the plain-object send flips the to-device unit test to `not ok` (the swallowed `TypeError` leaves no preview sent).

packages/ai-bot/main.ts
lukemelia: [Claude Code 🤖] Good catch — not intended. Fixed in 16619683b9. When the immediate trigger (a tool / code-patch result) carries no device id, we now fall back to the most recent stamped user-prompt event in the turn's history, so multi-step turns keep streaming to the device that started the turn instead of dropping to off after the first tool call. It still falls back to off only when no event in the turn was ever stamped (older client without CS-12262).

backspace (approved):

PR #5560, opened 2026-07-20
PosterBoard step 2: render cards at grid positions

burieberry:
Render prerendered html versions of selected cards on the poster board.

Sample card is in experiments-realm. Just renders the attached cards.

The live card tests are in the same directory and are not part of CI. They can be run locally via `https://localhost:4200/tests/index.html?liveTest=true&realmURL=https%3A%2F%2Flocalhost%3A4201%2Fexperiments%2F&filter=poster-board`

**Known limitation:** tiles linking a card in a prefix-mapped realm (`@cardstack/base/...`, `@cardstack/catalog/...`) render the not-found placeholder. Card links are canonicalized to the realm alias, but the search `cardUrls` filter only matches the index's URL form, so those cards get no rows back. Fixing that is outside this card: [CS-12744](https://linear.app/cardstack/issue/CS-12744) (search should match every equivalent URL form) and [CS-12745](https://linear.app/cardstack/issue/CS-12745) (realm write path rewrites alias links to resolved URLs).

Linear: [CS-12214](https://linear.app/cardstack/issue/CS-12214)

poster-board-render-cards

### DIFF
--- a/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
+++ b/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
@@ -34,32 +34,27 @@
       },
       "cards.3": {
         "links": {
-          "self": "@cardstack/base/Theme/boxel-brand-guide"
+          "self": "../../llm-model-environment/Model/bd5a43c0-0681-476d-af80-be4ad46d908c"
         }
       },
       "cards.4": {
         "links": {
-          "self": "@cardstack/base/Theme/cardstack-brand-guide"
+          "self": "../../Country/argentina"
         }
       },
       "cards.5": {
         "links": {
-          "self": "../../llm-model-environment/Model/bd5a43c0-0681-476d-af80-be4ad46d908c"
+          "self": "../../llm-model-environment/Environment/3f6cde92-99b2-4bb5-a471-20d907cb682b"
         }
       },
       "cards.6": {
         "links": {
-          "self": "../../Country/argentina"
+          "self": "../../carving-turn-entry"
         }
       },
       "cards.7": {
         "links": {
-          "self": "../../llm-model-environment/Environment/3f6cde92-99b2-4bb5-a471-20d907cb682b"
-        }
-      },
-      "cards.8": {
-        "links": {
-          "self": "../../carving-turn-entry"
+          "self": "../../DetailedStyleReference/8af54c49-13b5-4bd4-9b2d-95dd569c1f44"
         }
       }
     }
--- a/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
+++ b/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
@@ -31,6 +31,36 @@
         "links": {
           "self": "../../Contact/a01fc5c9-d70d-4b9c-aae4-384cf2b79b25"
         }
+      },
+      "cards.3": {
+        "links": {
+          "self": "@cardstack/base/Theme/boxel-brand-guide"
+        }
+      },
+      "cards.4": {
+        "links": {
+          "self": "@cardstack/base/Theme/cardstack-brand-guide"
+        }
+      },
+      "cards.5": {
+        "links": {
+          "self": "../../llm-model-environment/Model/bd5a43c0-0681-476d-af80-be4ad46d908c"
+        }
+      },
+      "cards.6": {
+        "links": {
+          "self": "../../Country/argentina"
+        }
+      },
+      "cards.7": {
+        "links": {
+          "self": "../../llm-model-environment/Environment/3f6cde92-99b2-4bb5-a471-20d907cb682b"
+        }
+      },
+      "cards.8": {
+        "links": {
+          "self": "../../carving-turn-entry"
+        }
       }
     }
   }
--- a/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
+++ b/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
@@ -3,13 +3,34 @@
     "type": "card",
     "attributes": {
       "cardInfo": {
-        "name": "Demo Poster Board"
-      }
+        "name": "Demo Poster Board",
+        "summary": null,
+        "cardThumbnailURL": null,
+        "notes": null
+      },
+      "frameSettings": []
     },
     "meta": {
       "adoptsFrom": {
         "module": "../poster-board",
         "name": "PosterBoard"
+      }
+    },
+    "relationships": {
+      "cards.0": {
+        "links": {
+          "self": "../../Author/jane-doe"
+        }
+      },
+      "cards.1": {
+        "links": {
+          "self": "../../Author/0b9c06fd-3833-4947-a0b8-ac24b8e71ee7"
+        }
+      },
+      "cards.2": {
+        "links": {
+          "self": "../../Contact/a01fc5c9-d70d-4b9c-aae4-384cf2b79b25"
+        }
       }
     }
   }
--- a/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
+++ b/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
@@ -48,6 +48,30 @@
         {
           "x": null,
           "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
         }
       ]
     },
@@ -100,12 +124,42 @@
       },
       "tiles.8.card": {
         "links": {
-          "self": "../../Chess/c09f9bb3-df3f-4337-a6bd-cc4c96a5caa3"
+          "self": "@cardstack/base/Theme/boxel-brand-guide"
         }
       },
       "tiles.9.card": {
         "links": {
-          "self": "../../index"
+          "self": "@cardstack/catalog/4b6602-wine-cellar-card-definition/WineBottle/011f0239-9a66-405f-a874-32bb4146ee7b"
+        }
+      },
+      "tiles.10.card": {
+        "links": {
+          "self": "@cardstack/base/cards/brand-guide"
+        }
+      },
+      "tiles.11.card": {
+        "links": {
+          "self": "@cardstack/catalog/Theme/cardstack-brand-guide"
+        }
+      },
+      "tiles.12.card": {
+        "links": {
+          "self": "@cardstack/skills/index"
+        }
+      },
+      "tiles.13.card": {
+        "links": {
+          "self": "@cardstack/catalog/c7eda3-travel-itinerary-planner/TravelItinerary/7e50c44f-7124-4347-b48d-b2433896fb6e"
+        }
+      },
+      "tiles.14.card": {
+        "links": {
+          "self": "../../GardenItem/c73d38e9-bd83-4089-9667-0b29dbc37008"
+        }
+      },
+      "tiles.15.card": {
+        "links": {
+          "self": "@cardstack/catalog/4b6602-wine-cellar-card-definition/WineBottle/a227be75-3301-4f02-b99a-66e05fe87432"
         }
       }
     }
--- a/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
+++ b/packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
@@ -8,7 +8,48 @@
         "cardThumbnailURL": null,
         "notes": null
       },
-      "frameSettings": []
+      "tiles": [
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        },
+        {
+          "x": null,
+          "y": null
+        }
+      ]
     },
     "meta": {
       "adoptsFrom": {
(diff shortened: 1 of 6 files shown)


backspace (approved):
I’m curious why the base cards don’t render, is there an issue for this? Might be something for me to look into

packages/experiments-realm/poster-board/poster-board.test.gts
```diff
@@ -1,26 +1,124 @@
...
+// downstream `saveCard` / `loader.import` calls check for real.
+function loaderService(): { loader: Loader } {
+  return getService('loader-service') as unknown as { loader: Loader };
+}
+
+// The board renders its tiles through `@context.searchResultsComponent`, which
+// only the host app provides (operator mode / index / prerender routes). The
+// stub below stands in for it: it captures each query the board issues and
+// yields a test-controlled entry set, so tile mapping and fallbacks are
+// exercised deterministically without a live prerender index.
```
habdelra: [Claude Code 🤖] **Non-blocking (test coverage / follow-up).** This comment says tile mapping *and fallbacks* are exercised deterministically, but neither test in this file actually reaches a fallback branch:

- `brokenSlotAt` → `BrokenLinkTemplate` (the error / not-found slot) never runs — both tests link only healthy `present` cards.
- The synthesized `missingEntryErrorDoc` "not-found" placeholder (`{{#unless results.isLoading}}` + `{{#if ref}}` when no entry matches the ref) never runs — every linked card has a matching `stubEntry`.
- `results.isLoading` is hard-coded `false` in the stub, so the loading branch is never taken either.

These fallbacks are the subtlest logic in the component, so leaving them uncovered is where a regression would hide. The missing-entry placeholder is cheap to pin without a broken link: link a saved card, leave `stubEntries` empty (isLoading stays false), and assert `[data-test-poster-board-broken-tile]` renders with `@state='not-found'`. If you'd rather lean on the live tests for the error paths, drop "and fallbacks" from this comment so it doesn't read as covering something it doesn't. Not a blocker for an experiments-realm demo.

---

packages/experiments-realm/poster-board/poster-board.gts
```diff
@@ -43,6 +99,103 @@ class Isolated extends Component<typeof PosterBoard> {
...
+    let owner = this.args.model as unknown as PosterBoard | undefined;
+    if (!owner) {
+      return undefined;
+    }
+    let { membership } = getRelationshipMembershipState(owner, 'cards');
+    let rel = (membership ?? [])[index];
+    return rel && (rel.kind === 'error' || rel.kind === 'not-found')
+      ? rel
+      : undefined;
+  };
```
habdelra, line 277: [Claude Code 🤖] **Non-blocking (cleanup).** `getRelationshipMembershipState(owner, 'cards')` is recomputed here on every `brokenSlotAt(index)` call, on top of the call in `linkedRefs` (which `tilePlacements`, `tilesQuery`, `entryFor`, and `refAt` all read through). Each call re-runs `getField` + `peekAtField` and rebuilds a fresh `membership` array.

To be fair to the design: Glimmer's per-reference caching keeps this off the pan/zoom hot path — these getters don't read rig state, so a pan frame doesn't recompute them — so it isn't a per-frame cost. It's the same array rebuilt several times per membership change, and it splits "read the cards' membership" across two call sites that must stay in sync.

Consider one source of truth, e.g. `@cached get slots()` returning `getRelationshipMembershipState(owner, 'cards').membership ?? []` once; then `linkedRefs` is `this.slots.map((s) => s.reference)`, `brokenSlotAt(i)` is `this.slots[i]` (test its kind), and `entryFor` / `refAt` index the same array. Purely a readability/efficiency cleanup.

---

packages/experiments-realm/poster-board/poster-board.gts
```diff
@@ -43,6 +99,103 @@ class Isolated extends Component<typeof PosterBoard> {
...
+      let setting = settings.find((s) => Number(s.cardIndex) === index);
+      // Number() guards against non-numeric values in hand-edited JSON
+      let x = Number(setting?.x);
+      let y = Number(setting?.y);
+      if (setting && Number.isFinite(x) && Number.isFinite(y)) {
+        return { index, x, y };
+      }
+      return defaultPlacement(index);
+    });
+  }
```
habdelra, line 202: [Claude Code 🤖] **Non-blocking (design note, relevant to step 3).** Frame settings are matched to cards by array index (`Number(s.cardIndex) === index`, where `index` is the position in the `cards` `linksToMany`). That means removing or reordering a linked card silently reassigns every persisted position after it to a different card — the settings don't move with the card they were authored for.

Step 3 (drag-to-move) will start writing real positions into these records, so the index-keyed model becomes load-bearing exactly when a mismatch is most visible (a user drags card A, later removes an earlier card, and card A's saved position now belongs to card B). Worth considering keying each `FrameSettingsField` by the linked card's reference/id instead of its slot index, so a position tracks its card across insert/remove/reorder. Not this PR's blocker — flagging now because step 3 builds directly on this shape.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** I focused on the tile-rendering pipeline — how the board turns linked-card references into a search query, maps results back to slots, and handles the broken/missing paths — plus the test's fidelity to that pipeline, since the stub replaces the real search engine.

**Bottom line: no blocking issues.** The query construction and reference-based mapping are correct and match established call sites; the design is sound for an experiments-realm demo. All findings are non-blocking (one test-coverage gap, one cleanup, one design note for step 3).

**What lands right (and is safe to rely on):**
- **The query** — `searchEntryWireQueryFromQuery({}, { scope: 'cards' })` + `cardUrls: refs.map((r) => r + '.json')` + `filter: { eq: { htmlQuery: { eq: { format: 'fitted' } } } }` — is exactly the shape `host-mode-service.ts` and `search-cards.ts` use. `scope: 'cards'` maps to `entryType: 'instance'` in `realm-index-query-engine.ts`, dropping each card's dual-indexed file row, and `cardUrls` compiles to `i.url IN (...)` on the `.json` file URLs. Verified against those call sites and the wire grammar in `search-entry.ts`.
- **Mapping by `entry.id === ref`** rather than by result position is correct: `entry.id` is the extensionless card URL and `linkedRefs` reads the same reference from membership; the reversed-order test pins this.
- **Reactivity is safe.** `SearchEntriesResource.modify` compares queries with a deep `isEqual`, so `tilesQuery` returning a fresh object each render does **not** re-issue the search; and `tilesQuery` reads only membership (never rig state), so pan/zoom don't re-run it. This is the subtle failure mode a display-on-a-canvas card could easily hit, and it's avoided.
- **Reading references via `getRelationshipMembershipState`** (a pure read that never triggers `lazilyLoadLink`) is the right call for a display board — tiles render from references without fetching the linked instances.

**On the earlier automated P1 (window-scoped keydown):** resolved in this head — the keydown listener is element-scoped (`{{on 'keydown'}}` on `.poster-board-root`), no `window` listener, so the double-fire concern doesn't apply to this commit. Noting so it isn't re-raised.

**@backspace, on why `@cardstack/base` linked cards don't render:** I ruled out the obvious cause — the base realm *is* searched. `realm-server`'s `availableRealms` seeds `{ type: 'base', url: baseRealm.url }`, and with no `realms` on the query the resource searches every available realm identifier, base included. So the issue is downstream of realm selection. The distinguishing factor is the reference *form*: the relative links (`../../Author/jane-doe`) resolve to canonical realm URLs, while `@cardstack/base/Theme/...` is the odd one out — most likely it isn't resolving to the canonical `https://cardstack.com/base/...` that both the `cardUrls` filter and the `entry.id` match expect (or those Theme cards simply have no prerendered fitted rendering). Quickest confirmation: log `this.tilesQuery.cardUrls` for the base slots against the `entry.id`s that come back — a mismatch confirms the reference-form theory. Host-side follow-up, not this PR.

**Numbered recommendations (all non-blocking):**
1. Test the fallback branches, or soften the stub comment that claims they're covered — see the thread on the stub comment in `poster-board.test.gts`.
2. Consolidate the repeated `getRelationshipMembershipState('cards')` into one source of truth — see the thread on `brokenSlotAt`.
3. Consider keying frame settings by card reference rather than slot index before step 3 persists positions — see the thread on `tilePlacements`.

**Adjacent, out of scope:**
- The `tsconfig.json` `include` reaches into `../host/node_modules/qunit-dom/dist/es/install.d.ts` to pull qunit-dom's assertion augmentations into the experiments-realm project. It works because host depends on qunit-dom, but it couples this project's typecheck to host's `node_modules` layout and qunit-dom's internal dist path; a shared test-types reference would be sturdier. (I couldn't confirm the exact path resolves in this environment — `node_modules` isn't installed here.)
- The board relies on `@overlays` defaulting to `true` for its tile chrome; passing it explicitly would be robust to a future default change. This matches the unguarded `(component @context.searchResultsComponent)` convention already used across experiments-realm (`components/grid.gts`, `components/card-list.gts`), so it's consistent as-is.

---

habdelra (approved):

packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
```diff
@@ -3,14 +3,60 @@
...
+      },
+      "cards.2": {
+        "links": {
+          "self": "../../Contact/a01fc5c9-d70d-4b9c-aae4-384cf2b79b25"
+        }
+      },
+      "cards.3": {
+        "links": {
+          "self": "../../llm-model-environment/Model/bd5a43c0-0681-476d-af80-be4ad46d908c"
+        }
```
richardhjtan: [Claude Code 🤖] Swapping the two base-realm links out removes the only thing in the tree that demonstrated the problem, while the description still asserts it: "`@cardstack/base` linked cards did not render. This seems to require a fix in host."

What's left after this diff is a reviewer's question with no way to reproduce the answer, and a diagnosis in that thread — that the `@cardstack/base/...` reference form doesn't resolve to the canonical `https://cardstack.com/base/...` that both the `cardUrls` filter and the `entry.id` match expect — that nothing in the repo can now confirm or refute.

The ask is an answer, not a code change: if the cause is known, put it in the description and drop the "seems to require a fix in host" hedge; if it isn't, file it so the finding doesn't leave with the fixture. To be explicit, I am *not* asking for a base link to be kept — a permanently broken tile in a demo card is worse than a missing repro.

Non-blocking.

richardhjtan (commented):
[Claude Code 🤖] Re-review after the force-push that dismissed the earlier approval, plus an independent pass over the canvas's accessibility contract, which the earlier reviews didn't cover.

**No blocking issues, and re-approval is cheap: the code is unchanged.** `poster-board.gts`, `poster-board.test.gts` and `tsconfig.json` are byte-identical to the commit that was approved — a `git diff` between the two heads over those paths is empty. The whole content change is `demo-poster-board.json`, where the two `@cardstack/base/Theme/...` links became realm-relative ones and a `DetailedStyleReference` replaced the dropped eighth slot. All eight targets resolve to files that exist in `packages/experiments-realm`, and `frameSettings` is still `[]`, so the index-keyed placement concern raised earlier isn't exercised by the reslotting.

I also chased the placeholder-flash risk in `{{#unless results.isLoading}}` and it isn't there: `isLoading` is `search.isRunning`, which goes true inside the same `modify()` that re-performs on a query change, so there is no window with a fresh query and a stale `entries`.

Two asks in the threads:

1. The base-realm question got closed by deletion rather than by an answer — see the thread on `cards.3`.
2. `role='application'` is step 1's, but this PR is what makes it expensive — see the thread on the root element.

**Still open from the earlier review** (the code didn't change, so none of these moved): the untested fallback branches, the duplicated `getRelationshipMembershipState('cards')` read, and keying frame settings by card reference rather than slot index. The third is the one worth settling before step 3 writes real positions through it.

**Adjacent, out of scope.** The HUD's `Fit` button runs `resetView`, which sends the camera to the world origin at 100% rather than fitting anything. Those coincide while every tile is at its default grid position and the grid is smaller than the viewport — eight `cardsgrid-tile`s at 170×250 with a 32px gap over 4 columns is 818×574 from the origin, so it holds today. Step 3's drag-to-move is what breaks it: a card dragged off to one side stays outside the viewport after pressing `Fit`. Pre-existing from step 1, so not this PR's to fix — but the content bounds are `min`/`max` over `tilePlacements`, which this PR is what introduces, so whoever writes step 3 has the pieces.

packages/experiments-realm/poster-board/poster-board.gts
```diff
@@ -168,25 +324,78 @@ class Isolated extends Component<typeof PosterBoard> {
...
       {{OnInsert this.handleInserted}}
       {{on 'wheel' this.handleWheel}}
       {{on 'pointerdown' this.handlePointerDown}}
       {{on 'pointermove' this.handlePointerMove}}
       {{on 'pointerup' this.handlePointerUp}}
       {{on 'pointercancel' this.handlePointerUp}}
       {{on 'keydown' this.handleKeyDown}}
+      role='application'
+      aria-label='Poster board canvas'
+      tabindex='0'
```
richardhjtan: [Claude Code 🤖] Pre-existing — `role='application'` and `tabindex='0'` are both in the base version of this file — but this PR is what makes the role expensive, so it's worth a decision here rather than after.

`role='application'` tells a screen reader to stop interpreting the subtree and forward keystrokes to the widget, which switches browse mode off over everything inside it. Until this PR the subtree held a background grid and a hint paragraph, so that cost close to nothing. This PR fills it with card HTML: every tile is now ordinary readable content that a screen-reader user can no longer navigate with the reader's own commands.

What the role buys in exchange is the three Shift-modified zoom shortcuts on this element. There is no keyboard path between tiles, no keyboard pan, and no arrow-key model — so it removes the navigation a screen-reader user has and offers nothing in its place. The APG reserves `application` for widgets implementing a full keyboard model.

`role='region'` (or no role) keeps the `aria-label` and the `tabindex='0'` the keydown listener needs, and leaves the tiles readable. If a self-contained canvas really is the intent, the role should land together with tile-to-tile focus movement rather than ahead of it.

Non-blocking; pre-existing, now load-bearing.

burieberry: [Claude Code 🤖] Agreed — switched the root to `role='region'` in aabecb6d90. The keydown listener only needs the `tabindex`, and the toolbar buttons already cover zoom, so nothing depended on `application`. Will revisit it if arrow-key tile navigation lands.

packages/experiments-realm/poster-board/poster-board.test.gts
burieberry: [Claude Code 🤖] Both fallbacks are now pinned in this file: `falls back to a not-found placeholder when a linked card has no index entry` (empty entry set with `isLoading` false, asserts `@state='not-found'`) and `reports a failed search as an error, not as missing cards`. The membership-reported error/not-found branch in `brokenSlotAt` still relies on the live tests, and the comment now only claims the missing-entry fallback.

packages/experiments-realm/poster-board/poster-board.gts
burieberry: [Claude Code 🤖] This went away with the model change: each `BoardTile` holds its own `linksTo` card, so `tileLinkState(tile)` reads one relationship with a single member. `linkedRefs` and `brokenSlotAt` each still call it, but there is no shared membership array to rebuild anymore.

packages/experiments-realm/poster-board/poster-board.gts
burieberry: [Claude Code 🤖] Done — tiles are now `containsMany(BoardTile)`, where each tile carries its `card` link plus `x`/`y`, so a position can't drift from its card on insert, remove, or reorder. `keeps a persisted position with its tile when an earlier tile is removed` pins it.

packages/experiments-realm/poster-board/PosterBoard/demo-poster-board.json
burieberry: [Claude Code 🤖] The cause is now known and ticketed, and the PR description says so instead of hedging. Card links are canonicalized to the realm alias (`@cardstack/base/...`, `@cardstack/catalog/...`), but the search `cardUrls` filter is an exact match on the index's URL form, so those tiles get zero rows and fall back to the not-found placeholder. That's CS-12744 (search should match every equivalent URL form); CS-12745 covers the realm write path rewriting alias links to resolved URLs, found while reproducing. The board resolves references through the store on its side, which is as far as this card can take it.

PR #5564, opened 2026-07-21
Add trailing slash normalisation in host mode

backspace:
When a host mode route map exists, a request for `/pricing/` should be treated as one for `/pricing`. With this running locally:

s 2026-07-21 at 15 25 02@2x

If a route map has a `/pricing/` route, it should act as `/pricing`; the editor now warns about this. I also updated the leading slash to match the font of the field, because the mismatched `/`es were irksome.

<table>
<tr>
<td>Before</td>
<td>
s 2026-07-21 at 15 06 31@2x
</td>
</tr>
<tr>
<td>After</td>
<td>
s 2026-07-21 at 15 01 02@2x
</td>
</tr>
</table>

I chose HTTP 308 according to [this flow chart](https://mlk.nfshost.com/choosing-an-http-status-code/HTTP-2XX-3XX-Status-Codes.svg).

### DIFF
--- a/packages/realm-server/handlers/serve-index.ts
+++ b/packages/realm-server/handlers/serve-index.ts
@@ -25,6 +25,7 @@ import {
   getPublishedRealmInfo,
   hasPublicPermissions,
   isIndexedCardInstance,
+  matchHostRoutingRule,
   type RealmRoutingDeps,
 } from '../lib/realm-routing.ts';
 import type { RealmRegistryReconciler } from '../lib/realm-registry-reconciler.ts';
@@ -248,7 +249,16 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
         let cardURL = requestURL;
         let isCardInstance = await isIndexedCardInstance(cardURL, routingDeps);
         if (!isCardInstance) {
-          return next();
+          // A bare routed sub-path (e.g. /pricing) is not itself an indexed
+          // card instance — its host routing rule maps it to a card that
+          // lives elsewhere (e.g. /pages/pricing). Only fall through to the
+          // module resolver (→ 404) when the path also matches no routing
+          // rule. Without this, the bare form 404s while the trailing-slash
+          // form — which skips this gate via isIndexRequest — renders.
+          let matchedRule = await matchHostRoutingRule(requestURL, routingDeps);
+          if (!matchedRule) {
+            return next();
+          }
         }
       }
     }
@@ -292,6 +302,45 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
 
     ctxt.type = 'html';
 
+    // Resolve the realm once and reuse it for canonicalization, the
+    // permissions check, and the routing-map lookup below. `findOrMountRealm`
+    // can fall back to a DB probe when the in-memory registry is cold, so we
+    // don't want to pay that cost more than necessary on the hot HTML path.
+    let routedRealm = await findOrMountRealm(requestURL, routingDeps);
+
+    // Canonicalize routed paths BEFORE conditional-GET handling.
+    // A host routing rule declared as '/pricing' matches both '/pricing' and
+    // '/pricing/' (RealmPaths.local strips trailing slashes). The canonical
+    // form is the realm mount pathname joined with the rule's declared path —
+    // so the realm-root rule '/' keeps its trailing slash while a sub-path
+    // rule has none. Redirect any other form with a 308 so bookmarks, shared
+    // links and crawlers converge on one URL, and so relative links in the
+    // served HTML resolve against a stable document base (a trailing slash
+    // shifts that base and breaks './'-relative hrefs in the prerendered
+    // page). This runs ahead of the ETag/304 block below so a repeat request
+    // under a non-canonical form still redirects rather than 304-ing.
+    let routingMap: { path: string; id: string }[] = [];
+    if (routedRealm) {
+      routingMap = await routedRealm.getHostRoutingMap();
+      if (routingMap.length > 0) {
+        let realmURL = new URL(routedRealm.url);
+        realmURL.protocol = requestURL.protocol;
+        let pathInRealm = '/' + new RealmPaths(realmURL).local(requestURL);
+        let rule = routingMap.find((r) => r.path === pathInRealm);
+        if (rule) {
+          let canonicalPathname =
+            realmURL.pathname + rule.path.replace(/^\//, '');
+          if (requestURL.pathname !== canonicalPathname) {
+            ctxt.redirect(
+              new URL(canonicalPathname + requestURL.search, requestURL).href,
+            );
+            ctxt.status = 308;
+            return;
+          }
+        }
+      }
+    }
+
     let cardURL = requestURL;
     let isIndexRequest = requestURL.pathname.endsWith('/');
     if (isIndexRequest) {
@@ -329,11 +378,6 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
         return;
       }
     }
-    // Resolve the realm once and reuse for both the permissions check and
-    // the routing-map lookup below. `findOrMountRealm` can fall back to a
-    // DB probe when the in-memory registry is cold, so we don't want to
-    // pay that cost twice on the hot HTML path.
-    let routedRealm = await findOrMountRealm(requestURL, routingDeps);
     let publicPermissions = await hasPublicPermissions(
       routedRealm,
       routingDeps,
@@ -350,21 +394,21 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
     // CS-10055: host routing rules in the realm config can map a bare path
     // (e.g. /whitepaper) to a target card. When the requested path matches
     // a rule, rewrite cardURL so the head/isolated/scoped CSS fetched
-    // below render the routed target. The same map is also written into
-    // the @cardstack/host/config/environment meta tag further down so the
-    // SPA can resolve the path post-hydration.
-    let routingMap: { path: string; id: string }[] = [];
-    if (routedRealm) {
-      routingMap = await routedRealm.getHostRoutingMap();
-      if (routingMap.length > 0) {
-        let realmURL = new URL(routedRealm.url);
-        realmURL.protocol = requestURL.protocol;
-        let realmPaths = new RealmPaths(realmURL);
-        let pathInRealm = '/' + realmPaths.local(requestURL);
-        let rule = routingMap.find((r) => r.path === pathInRealm);
-        if (rule) {
-          cardURL = new URL(rule.id);
-        }
+    // below render the routed target. The same map (resolved above for
+    // canonicalization) is also written into the
+    // @cardstack/host/config/environment meta tag further down so the SPA
+    // can resolve the path post-hydration.
+    if (routedRealm && routingMap.length > 0) {
+      let realmURL = new URL(routedRealm.url);
+      realmURL.protocol = requestURL.protocol;
+      let realmPaths = new RealmPaths(realmURL);
+      let pathInRealm = '/' + realmPaths.local(requestURL);
+      let rule = routingMap.find((r) => r.path === pathInRealm);
+      if (rule) {
+        // The request has already been canonicalized above, so `rule`
+        // matches the canonical form here; rewrite cardURL so the
+        // head/isolated/scoped CSS fetched below render the routed target.
+        cardURL = new URL(rule.id);
       }
     }
 
--- a/packages/realm-server/handlers/serve-index.ts
+++ b/packages/realm-server/handlers/serve-index.ts
@@ -255,8 +255,18 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
           // module resolver (→ 404) when the path also matches no routing
           // rule. Without this, the bare form 404s while the trailing-slash
           // form — which skips this gate via isIndexRequest — renders.
+          //
+          // The rule match is additionally gated on public read: consulting
+          // the routing map for a non-public realm would leak which bare
+          // paths are configured routes via the 200-vs-404 difference (a real
+          // route falls through to the generic 200 shell, a non-route 404s).
+          // Requiring public read makes routed and non-routed bare paths
+          // behave identically (both next() → 404) on a non-public realm.
           let matchedRule = await matchHostRoutingRule(requestURL, routingDeps);
-          if (!matchedRule) {
+          if (
+            !matchedRule ||
+            !(await hasPublicPermissions(matchedRule.realm, routingDeps))
+          ) {
             return next();
           }
         }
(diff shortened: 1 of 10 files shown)


packages/runtime-common/realm.ts
```diff
@@ -6794,6 +6794,14 @@ export class Realm {
...
         let instance = (rule as Record<string, unknown>).instance;
         if (typeof path !== 'string') return [];
+        // Normalize a rule authored with a trailing slash ('/pricing/') to
+        // its canonical slash-free form ('/pricing'). Request paths are
+        // matched slash-insensitively (RealmPaths.local strips trailing
+        // slashes), so an un-normalized '/pricing/' rule would never match;
+        // normalizing here also feeds the correct canonical form to the
+        // serve-index redirect and the client-side routing map. The
+        // realm-root rule '/' is preserved.
+        let normalizedPath = path.replace(/\/+$/, '') || '/';
```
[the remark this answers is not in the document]
backspace: fixed in aceba4a891

packages/realm-server/handlers/serve-index.ts
```diff
@@ -292,6 +302,45 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
...
+    // so the realm-root rule '/' keeps its trailing slash while a sub-path
+    // rule has none. Redirect any other form with a 308 so bookmarks, shared
+    // links and crawlers converge on one URL, and so relative links in the
+    // served HTML resolve against a stable document base (a trailing slash
+    // shifts that base and breaks './'-relative hrefs in the prerendered
+    // page). This runs ahead of the ETag/304 block below so a repeat request
+    // under a non-canonical form still redirects rather than 304-ing.
+    let routingMap: { path: string; id: string }[] = [];
+    if (routedRealm) {
+      routingMap = await routedRealm.getHostRoutingMap();
```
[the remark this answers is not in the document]
backspace: ac48cd74f4

packages/realm-server/handlers/serve-index.ts
```diff
@@ -248,7 +249,16 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
...
+          // A bare routed sub-path (e.g. /pricing) is not itself an indexed
+          // card instance — its host routing rule maps it to a card that
+          // lives elsewhere (e.g. /pages/pricing). Only fall through to the
+          // module resolver (→ 404) when the path also matches no routing
+          // rule. Without this, the bare form 404s while the trailing-slash
+          // form — which skips this gate via isIndexRequest — renders.
+          let matchedRule = await matchHostRoutingRule(requestURL, routingDeps);
+          if (!matchedRule) {
+            return next();
+          }
```
habdelra, line 270: [Claude Code 🤖] **This gate reopens the private-route disclosure that moving the routing lookup behind the permission gate just closed — on a different signal (200 vs 404 instead of the 308).** Blocking-worthy, but it's your/security's call.

**Mechanism.** This gate only runs for non-`text/html`, non-`vnd` requests (the `else` branch at line 225 — i.e. `*/*` from crawlers, curl, unfurlers). For those clients, `matchHostRoutingRule` calls `getHostRoutingMap()` and decides route-vs-not *before* any permission check. On a **non-public** published realm:

- `GET <realm>/pricing` (a real rule, target elsewhere) → `isCardInstance` false → `matchHostRoutingRule` matches → we do **not** `return next()` → fall through to the main block → `hasPublicPermissions` is false → generic Boxel shell, **200** (line 348-359).
- `GET <realm>/nonexistent` (no rule) → `matchHostRoutingRule` returns undefined → `return next()` → module resolver → **404**.

So an unauthenticated `*/*` caller can enumerate which bare sub-paths are configured routes on a private realm by diffing 200 vs 404. That's the same class of leak the permission-gate reordering addressed for the `text/html` 308 path; the comment at lines 361-367 even asserts "a non-public realm never has its routing map consulted" — but this gate consults `getHostRoutingMap()` for exactly that realm.

**Verified by tracing** the head branch: `text/html` requests skip this block entirely (they take the `if (includesHtmlMimeType)` branch, which only handles `vnd`), so the disclosure test for `/pricing/` + `text/html` doesn't exercise this path. Pre-PR the bare form uniformly 404'd, so this is newly introduced here.

**The fix.** Gate the bare-path serve on public permissions too, so route and non-route behave identically on a non-public realm. `matchHostRoutingRule` already resolved the realm, so a `hasPublicPermissions` check here is cheap:

```ts
if (!isCardInstance) {
  let matchedRule = await matchHostRoutingRule(requestURL, routingDeps);
  if (!matchedRule || !(await hasPublicPermissions(matchedRule.realm, routingDeps))) {
    return next();
  }
}
```

Non-public realm then 404s for both route and non-route bare paths — no oracle — which is also what the lines 361-367 comment claims already happens.

**Scope:** regression introduced by this PR (the bare-path serving is new), same disclosure class as the redirect fix.

---

packages/realm-server/handlers/serve-index.ts
```diff
@@ -363,6 +375,26 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
...
+          if (requestURL.pathname !== canonicalPathname) {
+            ctxt.redirect(
+              new URL(canonicalPathname + requestURL.search, requestURL).href,
+            );
+            ctxt.status = 308;
+            return;
+          }
+          // Rewrite cardURL so the head/isolated/scoped CSS fetched below
+          // render the routed target.
           cardURL = new URL(rule.id);
```
habdelra: [Claude Code 🤖] **The canonicalization decision now lives in two places, and only one of them is guarded.** Non-blocking; a simplification worth doing while the code is fresh.

`matchHostRoutingRule` was added to return a full `MatchedHostRoutingRule` — `{ realm, rule, canonicalPathname }` — and its doc calls `canonicalPathname` "the single URL the rule should be served under." But its only caller (the gate above) uses it purely as a boolean, and this block re-derives the same three things independently: it re-reads `getHostRoutingMap()`, re-runs `routingMap.find(...)`, and recomputes `canonicalPathname` (`realmURL.pathname + rule.path.replace(/^\//, '')` here vs. the identical expression at `realm-routing.ts:146`). So the canonicalization formula has two homes that will drift the first time one learns something the other doesn't.

There's a second, smaller consequence: `matchHostRoutingRule` wraps `realmPaths.local(requestURL)` in a try/catch (it can throw with `status:404` when the URL isn't in-realm — see `RealmPaths.local`), but this inline copy at line 375 doesn't. I couldn't construct a case where it throws given `routedRealm` resolved (both sides do the same protocol-adjusted, no-VirtualNetwork `inRealm` check), so this isn't a live crash — but consolidating onto the helper would pick up the guard for free.

**Suggestion:** have the `text/html` path call `matchHostRoutingRule` too and use its result for the redirect + `cardURL` rewrite, so the match/canonicalize logic exists once. That also lets the gate above thread its already-computed match through instead of this block re-resolving the realm and map a second time on the hot HTML path (the "resolve the realm once" comment at lines 305-308 is already undercut on the `*/*` path, where `findOrMountRealm` runs at 234, inside `matchHostRoutingRule` at 258, and again at 309).

**Scope:** follow-up / simplification, not a correctness bug.

---

packages/runtime-common/host-routing-validation.ts
```diff
@@ -5,6 +5,19 @@
 
 const VALID_PATH_PATTERN = /^\/(?:[A-Za-z0-9._~/-]|%[0-9A-Fa-f]{2})*$/;
 
+/**
+ * Canonical form of a routing rule path: a trailing slash is stripped so
+ * `/pricing/` and `/pricing` compare and match identically, with the realm
+ * root `/` preserved. This is the single source of truth for that
+ * normalization — `Realm.getHostRoutingMap` uses it to build the map, and
+ * `findDuplicateRoutingPaths` / `validateRoutingPath` use it so the editor's
+ * collision detection and advisories agree with how routes actually resolve.
```
habdelra: [Claude Code 🤖] **"The single source of truth for that normalization" isn't quite true — there's a pre-existing twin on the client.** Non-blocking, and it predates this PR — but the docstring makes a claim the tree doesn't back up, so worth reconciling now that this function exists.

`packages/host/app/services/host-mode-service.ts` has `canonicalizeRoutingPath()` doing the same job (strip trailing slashes, preserve `/`), and `resolveRoutedPath` runs it on both the injected map key and Ember's `params.path`. Since `@cardstack/runtime-common` is already imported there (line 7), that copy could import `normalizeRoutingPath` instead, making this genuinely the one home.

They aren't byte-identical, either: `normalizeRoutingPath('//')` → `'/'` (the `|| '/'` fallback), while `canonicalizeRoutingPath('//')` → `''`. Real map keys arrive server-normalized so this edge isn't reachable in practice today, but it's exactly the kind of latent divergence a shared helper is supposed to prevent.

**Suggestion:** either drop the "single source of truth" clause, or (better) fold the host's `canonicalizeRoutingPath` onto this export as a follow-up. **Scope:** follow-up / twin-implementation drift, not this PR's bug.

---

packages/realm-server/tests/server-endpoints/index-responses-test.ts
```diff
@@ -1605,4 +1605,300 @@ module(`server-endpoints/${basename(import.meta.filename)}`, function () {
...
+        assert.notStrictEqual(
+          response.status,
+          308,
+          'no route-specific redirect is emitted for a non-public realm',
+        );
+        assert.strictEqual(
+          response.status,
+          200,
+          'serves the generic shell for a non-public realm',
+        );
```
habdelra, line 1900: [Claude Code 🤖] **This test guards the disclosure on the one path that can't leak it, and skips the path that can.** Tied to the gate finding in `serve-index.ts`.

It asserts `/pricing/` + `Accept: text/html` returns 200 not 308 on a non-public realm — good, but `text/html` never reaches the early bare-path gate (it takes the `includesHtmlMimeType` branch, which only handles `vnd`). The leak I flagged is on the **bare** form with a **generic** Accept: `GET <realm>/pricing` + `Accept: */*` returns 200 while `GET <realm>/nonexistent` + `*/*` returns 404, which distinguishes a real route from a non-route on a private realm.

So this test would stay green even with the oracle present. A negative-space case that actually pins the property:

```ts
test('a non-public realm does not disclose bare routes to a generic Accept header', async function (assert) {
  await dbAdapter.execute(
    `DELETE FROM realm_user_permissions WHERE realm_url = '${publishedRealmURLString}'`,
  );
  let routed = await request
    .get(`${publishedRealmPath}pricing`)      // real rule
    .set('Host', publishedRealmHost)
    .set('Accept', '*/*');
  let bogus = await request
    .get(`${publishedRealmPath}not-a-route`)  // no rule
    .set('Host', publishedRealmHost)
    .set('Accept', '*/*');
  assert.strictEqual(
    routed.status,
    bogus.status,
    'a routed and a non-routed bare path are indistinguishable on a non-public realm',
  );
});
```

On the current branch `routed.status` is 200 and `bogus.status` is 404, so this fails until the gate checks permissions. **Scope:** test-coverage gap on the same finding.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with a focus on the request-routing state machine in `serve-index.ts` — specifically how the new bare-path branch interacts with the permission gate, the ETag fast-path, and the trailing-slash canonicalization — plus the shared normalization contract across the server map builder, the editor validators, and the client resolver.

**Bottom line: the trailing-slash normalization and canonical-redirect work is solid and well-tested, but the new `*/*` bare-path gate reintroduces the exact private-route disclosure this PR otherwise took care to close — on a non-public realm it now answers 200 for a real route and 404 for a non-route. That's the one issue I'd resolve before merge; everything else is follow-up or cleanup.**

**What lands right**
- Moving the routing-map lookup + 308 behind `hasPublicPermissions` (while keeping the ETag/304 fast-path ahead of it, since a revalidation discloses nothing) is the correct ordering, and the reasoning is captured in the comment at lines 361-367. My one caveat is that the *bare-path* branch doesn't yet honor that same ordering — see the inline thread.
- Centralizing trailing-slash handling in `normalizeRoutingPath` and reusing it across `getHostRoutingMap`, `findDuplicateRoutingPaths`, and `validateRoutingPath` is the right shape: the editor's collision banner, the path advisory, and the server match now agree by construction. The `.find()`-makes-the-second-target-unreachable case the earlier bot review raised is genuinely handled by the duplicate detection now.
- `matchHostRoutingRule` guarding `RealmPaths.local` against its `status:404` throw is the right instinct — I only wish the `text/html` path reused it rather than re-implementing the match inline (inline thread).
- Choosing 308 (not 302) so the method/body are preserved and the canonical form is treated as permanent is correct for this use.

**Recommendations** (detail in the inline threads)
1. **Gate the bare-path serve on public permissions** so a non-public realm can't be probed via 200-vs-404. — `serve-index.ts`, the `!isCardInstance` block. *Should-fix.*
2. **Add the negative-space test** for bare + `*/*` on a non-public realm; the existing disclosure test only covers `text/html` + trailing slash, which never hits the leaky branch. — `index-responses-test.ts`.
3. **Consolidate the canonicalization** — `matchHostRoutingRule` returns a full match (rule + `canonicalPathname`) that the caller uses only as a boolean while the HTML path recomputes it; unify so the formula and the `local()` guard live once. — `serve-index.ts`. *Follow-up.*
4. **Reconcile the "single source of truth" claim** with the client's pre-existing `canonicalizeRoutingPath` twin in `host-mode-service.ts`. — `host-routing-validation.ts`. *Follow-up.*

**Adjacent, out of scope**
- The `Host Memory Baseline` check is red while Host and Realm test suites are green; it reads as the standalone baseline job rather than anything in this diff, but worth a glance before merge.
- `getHostRoutingMap` normalizes an empty (`''`) authored path to `/`, so a hand-edited `realm.json` rule with a blank path + a linked instance would become a realm-root rule. This actually *aligns* the server with the editor (which maps an unset path to `/`) and the client, so it's arguably a latent fix rather than a bug — noting it only so the empty→root behavior is a conscious choice.

None of these block from my side except at your discretion on #1; submitting as a comment so approve/request-changes stays with a human reviewer.

---

packages/realm-server/handlers/serve-index.ts
backspace: as of 3d5836e7ae `hasPublicPermissions` is required

packages/runtime-common/host-routing-validation.ts
backspace: fixed in 1e8eb6ec74

packages/realm-server/handlers/serve-index.ts
backspace: in e1ad34283b the HTML route now calls the existing normalisation instead of duplicating it

packages/realm-server/tests/server-endpoints/index-responses-test.ts
backspace: a better test for this is in 3d5836e7ae

habdelra (approved):

PR #5568, opened 2026-07-21
Docs: ai-bot streaming modes, event schema, and throttle defaults

lukemelia:
Updates the ai-bot README to match the code and document the streaming knobs.

## Changes

- **Corrected throttle defaults**: `AI_BOT_STREAM_THROTTLE_MS` is `250` ms (was documented as `600`) and `AI_BOT_STREAM_MIN_DELTA` is `0` (was documented as `300`).
- **Streaming modes**: new section documenting `AI_BOT_STREAMING_MODE` — `room-edits` (default), `off`, and `to-device` — what each delivers and the to-device device-targeting fallback.
- **Event schema**: documents the `app.boxel.response-stream` to-device event payload (`AppBoxelResponseStreamContent`) field by field, including its cumulative-state / last-writer-wins-by-`sequence` semantics.
- **Rationale**: explains the homeserver-load trade-off between the three modes.

## Base

Stacked on `cs-12261-response-stream-event-schema`, which implements the streaming modes and event type this documents.

## Test plan

Docs-only; verified every value and field against `packages/ai-bot/lib/responder.ts`, `packages/ai-bot/main.ts`, and `packages/runtime-common/matrix-constants.ts`.

### DIFF
--- a/packages/ai-bot/README.md
+++ b/packages/ai-bot/README.md
@@ -54,14 +54,14 @@ While the model generates a response, ai-bot can surface the in-flight text so t
 
 Regardless of mode, the final consolidated response always lands as a room event — to-device previews are ephemeral and not persisted, so durable state must live in the room.
 
-**`to-device` fallback.** to-device previews need to know which device to target. The client stamps its device id (`matrixClient.getDeviceId()`) on the prompt event under the `app.boxel.originating-device-id` key. If ai-bot can't find that id for a turn — an older client that never stamped it, or a tool/code-patch continuation that carries no prompt — it skips mid-turn previews for that turn (as in `off`), and the final room event still lands.
+**`to-device` fallback.** to-device previews need to know which device to target. The client stamps its device id (`matrixClient.getDeviceId()`) on the prompt event under the `app.boxel.originating-device-id` key. A tool/code-patch continuation carries no device id on its own triggering event, so ai-bot falls back to the most recent stamped prompt in the turn's history — streaming continues across multi-step turns on the originating device. Only when no event anywhere in the turn's history carries the id — e.g. an older client that never stamped it — does ai-bot skip mid-turn previews for that turn (as in `off`); the final room event still lands either way.
 
 ### Streaming throttle tuning
 
-Applies to `room-edits` and `to-device` mid-turn sends; `off` sends nothing mid-turn.
+`off` sends nothing mid-turn.
 
-- `AI_BOT_STREAM_THROTTLE_MS` (default `250`): min ms between mid-turn sends.
-- `AI_BOT_STREAM_MIN_DELTA` (default `0`): min new body characters before sending an update. New reasoning, a changed tool call, the first content, and the final send always go through regardless of this threshold.
+- `AI_BOT_STREAM_THROTTLE_MS` (default `250`): min ms between mid-turn sends. Gates both `room-edits` and `to-device`.
+- `AI_BOT_STREAM_MIN_DELTA` (default `0`): min new body characters before a `room-edits` update is sent. New reasoning, a changed tool call, the first content, and the final send always go through regardless of this threshold. Gates `room-edits` only — `to-device` previews ignore it and send on every throttle window.
 
 ### `app.boxel.response-stream` to-device event
 
@@ -69,14 +69,14 @@ In `to-device` streaming mode, ai-bot emits `app.boxel.response-stream` to-devic
 
 Content payload (`AppBoxelResponseStreamContent` in `packages/runtime-common/matrix-constants.ts`):
 
-| Field           | Type        | Description                                                                                                                                                                                                    |
-| --------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `roomId`        | `string`    | Room the response belongs to.                                                                                                                                                                                  |
-| `parentEventId` | `string`    | Event id of the thinking placeholder this stream will eventually replace. Keys the client's preview state so concurrent turns don't collide.                                                                   |
-| `sequence`      | `number`    | Monotonic per turn. Since the payload is cumulative, gaps or reordering are safe — apply last-writer-wins by sequence.                                                                                         |
-| `body`          | `string`    | Accumulated response text so far.                                                                                                                                                                              |
-| `reasoning`     | `string`    | Accumulated reasoning text so far.                                                                                                                                                                             |
-| `toolRequests`  | `unknown[]` | Tool calls in the same wire shape as the room event's `app.boxel.tool-requests` key: each entry is `{ id, name, arguments: <object> }`, where `arguments` is `{}` until the streamed tool-call JSON completes. |
+| Field           | Type        | Description                                                                                                                                                                                                   |
+| --------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `roomId`        | `string`    | Room the response belongs to.                                                                                                                                                                                 |
+| `parentEventId` | `string`    | Event id of the thinking placeholder this stream will eventually replace. Keys the client's preview state so concurrent turns don't collide.                                                                  |
+| `sequence`      | `number`    | Monotonic per turn. Since the payload is cumulative, gaps or reordering are safe — apply last-writer-wins by sequence.                                                                                        |
+| `body`          | `string`    | Accumulated response text so far.                                                                                                                                                                             |
+| `reasoning`     | `string`    | Accumulated reasoning text so far.                                                                                                                                                                            |
+| `toolRequests`  | `unknown[]` | Tool calls in the same wire shape as the room event's `app.boxel.toolRequests` key: each entry is `{ id, name, arguments: <object> }`, where `arguments` is `{}` until the streamed tool-call JSON completes. |
 
 ### Rationale
 
--- a/packages/ai-bot/README.md
+++ b/packages/ai-bot/README.md
@@ -44,11 +44,43 @@ Start the server with `pnpm start` or `pnpm start-dev` for live reload.
     - `pnpm dlx clinic open .clinic/ai-bot-bp`
     - If needed, open the trace directly: `pnpm dlx clinic open '.clinic/ai-bot-bp/*clinic-bubbleprof/*-traceevent'`
 
-### Streaming behavior tuning
+### Streaming modes
 
-- You can adjust mid-stream edit frequency to balance responsiveness vs. Matrix churn:
-  - `AI_BOT_STREAM_THROTTLE_MS` (default `600`): min ms between edit sends.
-  - `AI_BOT_STREAM_MIN_DELTA` (default `300`): min new characters before sending an edit (final send always occurs).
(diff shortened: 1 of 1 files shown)


packages/ai-bot/README.md
```diff
@@ -44,11 +44,43 @@ Start the server with `pnpm start` or `pnpm start-dev` for live reload.
...
+| `sequence`      | `number`    | Monotonic per turn. Since the payload is cumulative, gaps or reordering are safe — apply last-writer-wins by sequence.                                                                                         |
+| `body`          | `string`    | Accumulated response text so far.                                                                                                                                                                              |
+| `reasoning`     | `string`    | Accumulated reasoning text so far.                                                                                                                                                                             |
+| `toolRequests`  | `unknown[]` | Tool calls in the same wire shape as the room event's `app.boxel.tool-requests` key: each entry is `{ id, name, arguments: <object> }`, where `arguments` is `{}` until the streamed tool-call JSON completes. |
```
habdelra: [Claude Code 🤖] **The wire key is `app.boxel.toolRequests`, not `app.boxel.tool-requests`.** A reader who greps a room event's JSON for `app.boxel.tool-requests` will find nothing.

The room event stores its tool calls under the constant `APP_BOXEL_TOOL_REQUESTS_KEY`, which is defined as camelCase:

```
// packages/runtime-common/matrix-constants.ts:12
export const APP_BOXEL_TOOL_REQUESTS_KEY = 'app.boxel.toolRequests';
```

and `sendMessageEvent` writes it under exactly that key (`packages/runtime-common/ai/matrix-utils.ts:68`: `[APP_BOXEL_TOOL_REQUESTS_KEY]: encodeToolRequests(toolRequests)`). The kebab-case spelling is easy to reach for because the *event type* on this same page (`app.boxel.response-stream`) really is kebab-case — but the tool-requests *content key* is camelCase. The source comment this table paraphrases gets it right (`matrix-constants.ts:156`: "Same wire shape as `APP_BOXEL_TOOL_REQUESTS_KEY`").

Suggested fix: change `app.boxel.tool-requests` → `app.boxel.toolRequests`.

Class: regression (introduced by this PR). Non-blocking but should-fix — the whole point of this PR is doc/code fidelity, and this is the one value a reader would copy verbatim.

---

packages/ai-bot/README.md
```diff
@@ -44,11 +44,43 @@ Start the server with `pnpm start` or `pnpm start-dev` for live reload.
...
+- `off`: no mid-turn updates at all. The room shows the thinking placeholder, then a single consolidated bot message when the turn completes. Cheapest for the homeserver; the client shows no streaming.
+- `to-device`: mid-turn previews are sent as ephemeral [`app.boxel.response-stream`](#appboxelresponse-stream-to-device-event) to-device messages targeted at the one device that composed the prompt, and only the final consolidated state lands as a room event. Streaming UX with a fraction of the room-event churn. See the event schema below.
+
+Regardless of mode, the final consolidated response always lands as a room event — to-device previews are ephemeral and not persisted, so durable state must live in the room.
+
+**`to-device` fallback.** to-device previews need to know which device to target. The client stamps its device id (`matrixClient.getDeviceId()`) on the prompt event under the `app.boxel.originating-device-id` key. If ai-bot can't find that id for a turn — an older client that never stamped it, or a tool/code-patch continuation that carries no prompt — it skips mid-turn previews for that turn (as in `off`), and the final room event still lands.
```
habdelra: [Claude Code 🤖] **This says a tool/code-patch continuation loses streaming, but the code deliberately keeps it streaming.** The paragraph lists "a tool/code-patch continuation that carries no prompt" as a case where ai-bot can't find the device id and falls back to `off`-mode behavior. That's the opposite of what `main.ts` does — and the multi-step-turn behavior is arguably the most important nuance of this fallback.

**What the code actually does.** When a turn is triggered by a tool/code-patch result, the triggering event itself carries no `app.boxel.originating-device-id`. `main.ts` handles that by scanning backward through the turn's event history for the most recent stamped prompt event:

```
// packages/ai-bot/main.ts:492-507
let streamPreviewTarget = readDeviceTarget(event.getContent(), event.getSender());
if (!streamPreviewTarget) {
  for (let i = eventList.length - 1; i >= 0; i--) {
    const candidate = eventList[i] as ...;
    const target = readDeviceTarget(candidate.content, candidate.sender);
    if (target) { streamPreviewTarget = target; break; }
  }
}
```

The comment right above it (`main.ts:471-478`) states the intent explicitly: "A continuation triggered by a tool / code-patch result carries no device id, so fall back to the most recent stamped user-prompt event in the turn's history — **otherwise multi-step turns would stop streaming after the first tool call** even on the originating device."

So a tool/code-patch continuation does **not** drop to `off`-mode previews; the fallback resolves it to the originating device and streaming continues across the multi-step turn.

**When the id genuinely isn't found** (→ `off`-mode behavior, `shouldStreamMidTurn` false at `responder.ts:110`): no event anywhere in the turn's history carries the key — i.e. an older client that never stamped it (or the stamped prompt has aged out of the history window).

**Suggested rewrite** of the "can't find that id" list:

> If ai-bot can't find that id anywhere in the turn's history — e.g. an older client that never stamped it — it skips mid-turn previews for that turn (as in `off`), and the final room event still lands. A tool/code-patch continuation carries no device id on its own triggering event, but ai-bot falls back to the most recent stamped prompt in the turn's history, so streaming continues across multi-step turns.

Class: regression (introduced by this PR). Should-fix — as written it documents the exact failure mode the base branch's fallback was written to prevent.

---

packages/ai-bot/README.md
```diff
@@ -44,11 +44,43 @@ Start the server with `pnpm start` or `pnpm start-dev` for live reload.
...
+
+### Streaming throttle tuning
+
+Applies to `room-edits` and `to-device` mid-turn sends; `off` sends nothing mid-turn.
+
+- `AI_BOT_STREAM_THROTTLE_MS` (default `250`): min ms between mid-turn sends.
+- `AI_BOT_STREAM_MIN_DELTA` (default `0`): min new body characters before sending an update. New reasoning, a changed tool call, the first content, and the final send always go through regardless of this threshold.
```
habdelra: [Claude Code 🤖] **`AI_BOT_STREAM_MIN_DELTA` has no effect in `to-device` mode — only `AI_BOT_STREAM_THROTTLE_MS` does.** The section intro says the tuning "Applies to `room-edits` and `to-device` mid-turn sends," which is true for the throttle but overstates `MIN_DELTA`.

**The two knobs live on different code paths.** The throttle wraps the dispatcher (`responder.ts:126-144`), which routes a mid-turn tick to `sendToDevicePreview()` in to-device mode and to `sendMessageEvent()` otherwise — so `THROTTLE_MS` gates both modes. But the min-delta gate is implemented *only* inside `sendMessageEvent`:

```
// packages/ai-bot/lib/responder.ts:207-212
const shouldSend =
  isFinal || isFirstContent || reasoningDelta > 0 ||
  toolCallsChanged || contentDelta >= minDelta;
```

`sendToDevicePreview` (`responder.ts:146-189`) reads no `MIN_DELTA` and applies no character-delta gate — it emits the accumulated state on every throttle tick where the response state changed at all. So with, say, `AI_BOT_STREAM_MIN_DELTA=300`, room-edits would coalesce until 300 new body chars accumulate, while to-device would still fire every throttle window. (At the default `0` this is invisible, which is why it's easy to miss.)

Suggested tweak: scope the intro, e.g. "`AI_BOT_STREAM_THROTTLE_MS` applies to both `room-edits` and `to-device` mid-turn sends; `AI_BOT_STREAM_MIN_DELTA` gates only `room-edits` (and the final consolidated send) — to-device previews send on every throttle window."

Class: pre-existing behavior, now being documented (not a code bug in this PR). Non-blocking — worth correcting so an operator tuning to-device doesn't reach for a knob that does nothing there.

---

habdelra (commented):
[Claude Code 🤖] Reviewed as a fidelity check: since this is docs-only, I traced every documented value, field, and behavior back to the code it claims to describe (`responder.ts`, `response-publisher.ts`, `main.ts`, `matrix-utils.ts`, `matrix-constants.ts`) on the head branch.

**Bottom line: the doc is accurate on nearly everything, with two claims that contradict the code and one that overstates a knob's reach. No blocking issues; all three are quick doc edits.**

**What lands right (verified, not skimmed):**
- Throttle defaults match: `AI_BOT_STREAM_THROTTLE_MS` = `250` (`responder.ts:143`) and `AI_BOT_STREAM_MIN_DELTA` = `0` (`responder.ts:193`) — the old `600`/`300` values were indeed stale.
- The `MIN_DELTA` exception list ("new reasoning, a changed tool call, the first content, and the final send always go through") maps one-to-one onto the `shouldSend` disjunction at `responder.ts:207-212`.
- Streaming modes and their delivery: `room-edits` uses `m.replace` (`matrix-utils.ts:42-47`), `off` sends nothing mid-turn, `to-device` previews go over `sendToDevice` and only the final consolidated state lands as a room event (`responder.ts:133-141`) — all correct.
- Event schema: every `AppBoxelResponseStreamContent` field matches (`matrix-constants.ts:146-161`), `sequence` is per-turn monotonic from 0 (`responder.ts:157`), and `arguments: {}` until the streamed JSON parses is exactly `toCommandRequest`'s catch branch (`response-publisher.ts:33-41`).

**Recommendations:**
1. Fix the tool-requests wire key: `app.boxel.tool-requests` → `app.boxel.toolRequests` — see the thread on the schema table (should-fix).
2. Rework the `to-device` fallback paragraph — a tool/code-patch continuation keeps streaming via the turn-history fallback in `main.ts`, rather than dropping to `off`; see the thread on that paragraph (should-fix).
3. Scope the throttle-tuning intro so `MIN_DELTA` isn't credited with gating to-device previews — see the thread on that section (non-blocking).

**Adjacent, out of scope:** the `body` character-delta gate (`contentDelta >= minDelta`) keys off body length only; reasoning-only growth is covered by the separate `reasoningDelta > 0` term. That's consistent and not worth documenting — noting it only so the "body characters" wording in the table isn't later "fixed" to include reasoning.

---

packages/ai-bot/README.md
lukemelia: [Claude Code 🤖] Fixed in 72c85ebf14 — the table now reads `app.boxel.toolRequests` (camelCase), matching `APP_BOXEL_TOOL_REQUESTS_KEY`.

packages/ai-bot/README.md
lukemelia: [Claude Code 🤖] Fixed in 72c85ebf14. The paragraph now states that a tool/code-patch continuation resolves its target from the most recent stamped prompt in the turn history, so streaming continues across multi-step turns; only a turn whose history carries no stamped id (e.g. an older client) drops to `off`-mode behavior.

packages/ai-bot/README.md
lukemelia: [Claude Code 🤖] Fixed in 72c85ebf14. The tuning section now scopes each knob: `AI_BOT_STREAM_THROTTLE_MS` gates both `room-edits` and `to-device`, while `AI_BOT_STREAM_MIN_DELTA` gates `room-edits` only — to-device previews send on every throttle window.

habdelra (approved):

PR #5570, opened 2026-07-22
Client: hydrate ai-bot to-device streaming previews into room messages

lukemelia:
Consume ai-bot's `app.boxel.response-stream` to-device streaming previews on the client so the browser renders streaming UX when ai-bot runs in `AI_BOT_STREAMING_MODE=to-device` — without a Matrix room event per throttle window.

## What this does

- **Central binding (matrix-service).** All raw-client event bindings live in `matrix-service`, so it binds `ClientEvent.ToDeviceEvent` centrally, filters for `app.boxel.response-stream`, and fans payloads out to `RoomResource`s that register via a new `onResponseStreamPreview(handler)` (returns a disposer). Each handler self-filters by roomId. This deviates from the ticket's literal "`matrixClient.on` in room.ts" because the raw client is private and centralizing bindings is the established pattern.
- **Reconciliation (room.ts).** `RoomResource` registers in `modify()` / disposes in `teardown()`. `hydrateResponseStreamPreview()` shapes each preview as the `CardMessage` edit it mirrors and runs it through the existing `MessageBuilder.updateMessage` path, reusing its tool-request chunk merging.
- **Ordering.** The synthetic event's `origin_server_ts` is pinned to the message's creation time so a preview never advances past — and thus never suppresses (via `applyToolRequestChunk`) — the real, later final room edit. A per-message `sequence` gate drops out-of-order / duplicate previews; payloads carry full accumulated state, so last-writer-wins is safe.
- **Stall timeout.** No `room-message.gts` change was needed: `updateMessage` stamps a fresh `updated` on every applied preview, which is exactly the value the streaming stall-timeout getter compares against — so long responses reset the timeout for free.

## Test coverage

mock-matrix had no to-device support, so this adds `simulateToDeviceEvent` (mock client + `ClientEvent.ToDeviceEvent` on the mock SDK + `MockUtils`), plus an integration test covering preview hydration (body + reasoning), the sequence gate dropping stale previews, and unknown-parent / cross-room no-ops.

## Base

Stacked on `cs-12261-response-stream-event-schema`, which defines the `app.boxel.response-stream` event schema this consumes.

## Test plan

- [x] Typecheck (`lint:types`) clean for all changed files.
- [x] ESLint / template-lint clean.
- [ ] Host integration suite — not run locally (needs the full realm + Matrix server harness); relying on CI.

### DIFF
--- a/packages/host/app/resources/room.ts
+++ b/packages/host/app/resources/room.ts
@@ -25,9 +25,13 @@ import {
   getToolRequests,
   isToolResultRelType,
   APP_BOXEL_DEBUG_MESSAGE_EVENT_TYPE,
+  APP_BOXEL_MESSAGE_MSGTYPE,
   APP_BOXEL_REALM_SERVER_EVENT_MSGTYPE,
+  APP_BOXEL_REASONING_CONTENT_KEY,
+  APP_BOXEL_TOOL_REQUESTS_KEY,
   APP_BOXEL_LLM_MODE,
   DEFAULT_FALLBACK_MODEL_ID,
+  type AppBoxelResponseStreamContent,
   type LLMMode,
 } from '@cardstack/runtime-common/matrix-constants';
 
@@ -94,6 +98,12 @@ interface Args {
 export class RoomResource extends Resource<Args> {
   #skillIds = new Set<string>();
   #hasRegisteredDestructor = false;
+  #responseStreamPreviewDisposer: (() => void) | undefined;
+  // Highest to-device preview `sequence` applied per streaming message
+  // (keyed by parentEventId). Previews carry full accumulated state, so an
+  // out-of-order or duplicate delivery is simply dropped rather than regressing
+  // the message to older content.
+  #lastPreviewSequence = new Map<string, number>();
   private _messageCache: TrackedMap<string, Message> = new TrackedMap();
   private _nameEventsCache: TrackedMap<string, RoomNameEvent> =
     new TrackedMap();
@@ -123,6 +133,10 @@ export class RoomResource extends Resource<Args> {
     this.processing = this.processRoomTask.perform(named.roomId);
     if (!this.#hasRegisteredDestructor) {
       this.#hasRegisteredDestructor = true;
+      this.#responseStreamPreviewDisposer =
+        this.matrixService.onResponseStreamPreview((payload) =>
+          this.hydrateResponseStreamPreview(payload),
+        );
       registerDestructor(this, () => this.teardown());
     }
   }
@@ -131,6 +145,9 @@ export class RoomResource extends Resource<Args> {
     this.processRoomTask.cancelAll();
     this.activateLLMTask.cancelAll();
     this.activateLLMModeTask.cancelAll();
+    this.#responseStreamPreviewDisposer?.();
+    this.#responseStreamPreviewDisposer = undefined;
+    this.#lastPreviewSequence.clear();
     for (let id of this.#skillIds ?? []) {
       this.store.dropReference(id);
     }
@@ -713,6 +730,66 @@ export class RoomResource extends Resource<Args> {
         continuedFromMessage.continuedInMessage = message;
       }
     }
+  }
+
+  // Hydrate an in-flight `app.boxel.response-stream` to-device preview (ai-bot
+  // in AI_BOT_STREAMING_MODE=to-device) into the same Message the final room
+  // edit will eventually reconcile. Shaping the preview as the CardMessage edit
+  // it mirrors lets us reuse MessageBuilder.updateMessage — including its tool
+  // request chunk merging — and its `setUpdated(new Date())` resets the
+  // streaming stall timeout so long responses don't trip the fallback.
+  private async hydrateResponseStreamPreview(
+    payload: AppBoxelResponseStreamContent,
+  ) {
+    if (!payload || payload.roomId !== this.roomId) {
+      return;
+    }
+    let message = this._messageCache.get(payload.parentEventId);
+    // Either the thinking placeholder this preview belongs to hasn't loaded
+    // yet, or the turn already landed its final consolidated room edit. In both
+    // cases the room-event path is authoritative and reconciles the true state,
+    // so there is nothing for an ephemeral preview to add.
+    if (!message || message.isStreamingOfEventFinished) {
+      return;
+    }
+    let lastSequence =
+      this.#lastPreviewSequence.get(payload.parentEventId) ?? -1;
+    if (payload.sequence <= lastSequence) {
+      return;
+    }
+    this.#lastPreviewSequence.set(payload.parentEventId, payload.sequence);
+
+    let author = this.upsertRoomMember({
+      roomId: payload.roomId,
+      userId: this.matrixService.aiBotUserId,
+    });
+    // origin_server_ts is pinned to the message's creation time so a preview
+    // never advances past — and thus never suppresses (see
+    // MessageBuilder.applyToolRequestChunk) — the real, later final room edit.
+    let syntheticEvent = {
+      type: 'm.room.message',
+      event_id: payload.parentEventId,
+      room_id: payload.roomId,
+      sender: this.matrixService.aiBotUserId,
+      origin_server_ts: message.created.getTime(),
+      content: {
+        msgtype: APP_BOXEL_MESSAGE_MSGTYPE,
+        body: payload.body,
+        [APP_BOXEL_REASONING_CONTENT_KEY]: payload.reasoning,
+        [APP_BOXEL_TOOL_REQUESTS_KEY]: payload.toolRequests,
+        isStreamingFinished: false,
+      },
+    } as unknown as CardMessageEvent;
+
+    let messageBuilder = new MessageBuilder(syntheticEvent, getOwner(this)!, {
+      roomId: payload.roomId,
+      effectiveEventId: payload.parentEventId,
+      author,
+      index: message.index ?? 0,
+      events: this.events,
+      skills: this.skills,
+    });
+    await messageBuilder.updateMessage(message);
   }
 
   private async updateMessageCommandResult({
--- a/packages/host/app/resources/room.ts
+++ b/packages/host/app/resources/room.ts
@@ -793,6 +793,18 @@ export class RoomResource extends Resource<Args> {
     message: Message,
     payload: AppBoxelResponseStreamContent,
   ) {
+    // The synchronous gate checked isStreamingOfEventFinished at *enqueue* time,
+    // but this apply may have waited behind a predecessor on the chain while the
+    // final consolidated room edit landed and finalized the message. Re-check
+    // now: updateMessage rewrites body / reasoning / isStreamingFinished
+    // synchronously (before its first await), so an apply that *starts* after
+    // finalization would un-finish the completed message with stale content —
+    // leaving it stuck "streaming" until the stall timeout trips. An apply that
+    // started before finalization is safe: the final edit's synchronous writes
+    // still land last.
+    if (message.isStreamingOfEventFinished) {
+      return;
+    }
     try {
       let author = this.upsertRoomMember({
         roomId: payload.roomId,
--- a/packages/host/app/resources/room.ts
+++ b/packages/host/app/resources/room.ts
@@ -12,6 +12,7 @@ import { TrackedMap } from 'tracked-built-ins';
 
 import {
   isCardInstance,
+  logger,
   rri,
   type LooseSingleCardDocument,
 } from '@cardstack/runtime-common';
@@ -85,6 +86,8 @@ export type RoomSkill = {
   isActive: boolean;
 };
 
+const responseStreamLog = logger('matrix:response-stream');
+
 interface Args {
   named: {
     roomId: string | undefined;
@@ -102,8 +105,12 @@ export class RoomResource extends Resource<Args> {
   // Highest to-device preview `sequence` applied per streaming message
   // (keyed by parentEventId). Previews carry full accumulated state, so an
   // out-of-order or duplicate delivery is simply dropped rather than regressing
-  // the message to older content.
+  // the message to older content. Both maps are pruned when a message finalizes
+  // (see hydrateResponseStreamPreview) and cleared in teardown().
   #lastPreviewSequence = new Map<string, number>();
+  // Serializes preview applies per streaming message so their async tool-request
+  // builds land in sequence order rather than promise-completion order.
+  #previewApplyChain = new Map<string, Promise<void>>();
   private _messageCache: TrackedMap<string, Message> = new TrackedMap();
   private _nameEventsCache: TrackedMap<string, RoomNameEvent> =
     new TrackedMap();
@@ -148,6 +155,7 @@ export class RoomResource extends Resource<Args> {
     this.#responseStreamPreviewDisposer?.();
     this.#responseStreamPreviewDisposer = undefined;
     this.#lastPreviewSequence.clear();
+    this.#previewApplyChain.clear();
     for (let id of this.#skillIds ?? []) {
       this.store.dropReference(id);
     }
@@ -738,9 +746,16 @@ export class RoomResource extends Resource<Args> {
   // it mirrors lets us reuse MessageBuilder.updateMessage — including its tool
   // request chunk merging — and its `setUpdated(new Date())` resets the
   // streaming stall timeout so long responses don't trip the fallback.
-  private async hydrateResponseStreamPreview(
-    payload: AppBoxelResponseStreamContent,
-  ) {
+  //
+  // This method gates an incoming preview (roomId / staleness / duplicate)
+  // synchronously and then enqueues its apply on a per-message chain. The gate
+  // must be synchronous: it decides ordering by arrival, and
+  // applyResponseStreamPreview awaits async tool-request builds, so without
+  // serialization two overlapping previews would resolve their tool args in
+  // promise-completion order rather than sequence order (every synthetic event
+  // pins the same origin_server_ts, so MessageBuilder.applyToolRequestChunk's
+  // timestamp guard can't reorder them).
+  private hydrateResponseStreamPreview(payload: AppBoxelResponseStreamContent) {
     if (!payload || payload.roomId !== this.roomId) {
       return;
     }
@@ -750,6 +765,13 @@ export class RoomResource extends Resource<Args> {
     // cases the room-event path is authoritative and reconciles the true state,
     // so there is nothing for an ephemeral preview to add.
     if (!message || message.isStreamingOfEventFinished) {
+      // Once the message is finalized its tracking entries are dead — prune
+      // them opportunistically so a long-lived room doesn't accumulate one per
+      // AI turn (teardown() is the backstop).
+      if (message?.isStreamingOfEventFinished) {
+        this.#lastPreviewSequence.delete(payload.parentEventId);
+        this.#previewApplyChain.delete(payload.parentEventId);
+      }
       return;
     }
     let lastSequence =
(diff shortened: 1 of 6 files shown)


packages/host/app/resources/room.ts
```diff
@@ -715,6 +732,66 @@ export class RoomResource extends Resource<Args> {
...
+
+    let messageBuilder = new MessageBuilder(syntheticEvent, getOwner(this)!, {
+      roomId: payload.roomId,
+      effectiveEventId: payload.parentEventId,
+      author,
+      index: message.index ?? 0,
+      events: this.events,
+      skills: this.skills,
+    });
+    await messageBuilder.updateMessage(message);
```
lukemelia: [Claude Code 🤖] **Unhandled promise rejection — previews have no error path.** `hydrateResponseStreamPreview` is invoked fire-and-forget (room.ts:138, and matrix-service `onToDeviceEvent` ignores the returned promise), so a rejection here has nowhere to go. `updateMessage` is not pure-sync: when a preview carries a *new* tool-request id it calls `buildMessageCommand`, which awaits skill/loader resolution and can throw. The room-event path swallows this inside `processRoomTask`'s try/catch; this path has none, so a failed skill resolve becomes an unhandled rejection (Sentry noise / a failed test) even though a dropped preview is harmless (the final room edit reconciles). Suggest wrapping the body in try/catch and logging at debug, mirroring ai-bot's own best-effort `sendToDevicePreview`.

packages/host/app/resources/room.ts
```diff
@@ -715,6 +732,66 @@ export class RoomResource extends Resource<Args> {
...
+    // so there is nothing for an ephemeral preview to add.
+    if (!message || message.isStreamingOfEventFinished) {
+      return;
+    }
+    let lastSequence =
+      this.#lastPreviewSequence.get(payload.parentEventId) ?? -1;
+    if (payload.sequence <= lastSequence) {
+      return;
+    }
+    this.#lastPreviewSequence.set(payload.parentEventId, payload.sequence);
```
lukemelia, line 782: [Claude Code 🤖] **Previews are applied concurrently, so the `sequence` gate does not guarantee final ordering.** The gate blocks *entry* of a stale preview, but each to-device event fires its own un-awaited `hydrateResponseStreamPreview` (unlike the room-event loop in `processRoomTask`, which `await`s each `updateMessage` in turn). When a preview introduces a new tool id, `updateMessage` awaits `buildMessageCommand`, so two in-flight previews interleave. Because every synthetic event pins `origin_server_ts` to the same `message.created` time, `applyToolRequestChunk`'s `origin_server_ts < toolRequestEventTs` guard never fires between previews — the tool args written last are decided by promise-completion order, not `sequence`, so an older preview can win. Body/reasoning are safe (set synchronously in entry order); only tool args race. Serializing per message (a promise chain keyed by parentEventId) closes this and composes with the try/catch above.

packages/host/app/resources/room.ts
```diff
@@ -715,6 +732,66 @@ export class RoomResource extends Resource<Args> {
...
+    // origin_server_ts is pinned to the message's creation time so a preview
+    // never advances past — and thus never suppresses (see
+    // MessageBuilder.applyToolRequestChunk) — the real, later final room edit.
+    let syntheticEvent = {
+      type: 'm.room.message',
+      event_id: payload.parentEventId,
+      room_id: payload.roomId,
+      sender: this.matrixService.aiBotUserId,
+      origin_server_ts: message.created.getTime(),
+      content: {
```
lukemelia: [Claude Code 🤖] **The synthetic event clobbers fields it does not model.** `updateMessage` unconditionally writes several fields from `this.event`: for `APP_BOXEL_MESSAGE_MSGTYPE` it sets `attachedCardIds`/`attachedCardsAsFiles`/`attachedFiles` (→ empty, since the preview has no `data`), and always sets `continuationOf`/`hasContinuation` (→ reset), `status` (→ undefined), and `reloadBillingData` (→ false). For a plain streaming response these are empty/false anyway and the final room edit restores the truth, so it is mostly benign today — but it is fragile: any bot message that legitimately carried an attachment, a continuation, or a billing-reload flag on its placeholder would have it wiped on every preview until finalization. Worth a comment noting the preview intentionally only owns body/reasoning/toolRequests, or narrowing which fields `updateMessage` touches.

packages/host/app/resources/room.ts
```diff
@@ -715,6 +732,66 @@ export class RoomResource extends Resource<Args> {
...
+        msgtype: APP_BOXEL_MESSAGE_MSGTYPE,
+        body: payload.body,
+        [APP_BOXEL_REASONING_CONTENT_KEY]: payload.reasoning,
+        [APP_BOXEL_TOOL_REQUESTS_KEY]: payload.toolRequests,
+        isStreamingFinished: false,
```
lukemelia: [Claude Code 🤖] **`isStreamingFinished: false` here is redundant and a latent un-finish hazard.** Previews are inherently mid-stream, and the top-of-handler `message.isStreamingOfEventFinished` guard already prevents applying a preview after the final edit. But if a preview ever slips past that guard (e.g. the ordering race above), `updateMessage` calls `setIsStreamingFinished(false)` and un-finishes a completed message. Since the value is never anything but `false`, omitting the key (so `updateMessage` passes `undefined`, a no-op) removes the footgun without changing behavior.

packages/host/app/resources/room.ts
```diff
@@ -94,6 +98,12 @@ interface Args {
 export class RoomResource extends Resource<Args> {
   #skillIds = new Set<string>();
   #hasRegisteredDestructor = false;
+  #responseStreamPreviewDisposer: (() => void) | undefined;
+  // Highest to-device preview `sequence` applied per streaming message
+  // (keyed by parentEventId). Previews carry full accumulated state, so an
+  // out-of-order or duplicate delivery is simply dropped rather than regressing
+  // the message to older content.
+  #lastPreviewSequence = new Map<string, number>();
```
lukemelia, line 110: [Claude Code 🤖] **`#lastPreviewSequence` grows for the lifetime of the room.** One entry is added per streamed bot message and only ever cleared in `teardown()`. A long-lived room with many AI turns accumulates a dead entry per turn. Minor, but since each streaming message is finalized by its final room edit, its sequence entry is dead the moment `isStreamingOfEventFinished` flips — you could delete the key at that point, or prune opportunistically.

packages/host/app/resources/room.ts
lukemelia: [Claude Code 🤖] Fixed in 4377f26. The apply now runs inside a try/catch that logs at debug (channel `matrix:response-stream`) and swallows the error, mirroring ai-bot's best-effort `sendToDevicePreview` — a dropped preview is reconciled by the final room edit, so it should never surface as an unhandled rejection.

packages/host/app/resources/room.ts
lukemelia: [Claude Code 🤖] Fixed in 4377f26. Split into a synchronous gate (`hydrateResponseStreamPreview`) and an async apply (`applyResponseStreamPreview`); applies are now serialized per `parentEventId` via a promise chain (`#previewApplyChain`), so a new tool id's async `buildMessageCommand` lands in sequence order rather than promise-completion order. Body/reasoning were already safe — they're written before the first await, in arrival (sequence) order.

packages/host/app/resources/room.ts
lukemelia: [Claude Code 🤖] Addressed in 4377f26 with a comment on the synthetic event, noting the preview intentionally owns only body/reasoning/toolRequests and that the other fields `updateMessage` writes (attached*, continuationOf/hasContinuation, status, reloadBillingData) reset to empty/false on every apply and are restored by the final room edit. Behavior left as-is (benign for a streaming response); the comment flags the fragility for anyone who later gives a placeholder a legitimate attachment/continuation/billing flag.

packages/host/app/resources/room.ts
lukemelia: [Claude Code 🤖] Not taking this one — the suggested fix doesn't hold. `setIsStreamingFinished(undefined)` is not a no-op and doesn't remove the un-finish hazard: `isStreamingOfEventFinished` returns `_isStreamingFinished === true`, so if a finished message ever slipped past the guard, passing `undefined` still flips `true → undefined` and the getter goes `false` — un-finished, exactly as `false` would. The actual protection against that race is the per-message serialization in 4377f26 plus the existing `isStreamingOfEventFinished` guard, not omitting the key. Keeping the explicit `false` as an honest statement that a preview is always mid-stream.

packages/host/app/resources/room.ts
lukemelia: [Claude Code 🤖] Fixed in 4377f26. Both `#lastPreviewSequence` and `#previewApplyChain` are now pruned when a preview arrives for an already-finalized message (`isStreamingOfEventFinished`), with `teardown()` as the backstop. It's opportunistic (the fallback you suggested): a turn whose last preview arrives before finalization leaves a single entry until teardown, since this resource doesn't own the room-event path where the flag actually flips.

packages/host/app/resources/room.ts
```diff
@@ -715,6 +740,112 @@ export class RoomResource extends Resource<Args> {
...
+
+      let messageBuilder = new MessageBuilder(syntheticEvent, getOwner(this)!, {
+        roomId: payload.roomId,
+        effectiveEventId: payload.parentEventId,
+        author,
+        index: message.index ?? 0,
+        events: this.events,
+        skills: this.skills,
+      });
+      await messageBuilder.updateMessage(message);
```
habdelra, line 847: [Claude Code 🤖] **A preview applied *after* the turn finalizes clobbers the completed message — body/reasoning regress and the message is un-finished, so it renders as stuck "streaming" and eventually trips the stall-timeout error, for a response that actually succeeded.** Blocking; regression introduced here. This is the residual of the un-finish hazard from the `isStreamingFinished: false` thread — see the end.

**Mechanism.** `isStreamingOfEventFinished` is checked only in the *synchronous* gate `hydrateResponseStreamPreview` (line 767), i.e. at *enqueue* time. The write happens later in this method, which never re-checks it before `updateMessage`. And there is no ordering guarantee between `#previewApplyChain` and `processRoomTask`'s `loadRoomMessage` — they are two independent async writers of the same `Message`. So this interleaving is reachable:

1. Preview seq=N passes the gate (not yet finalized) and is enqueued; its predecessor on the chain is still awaiting `buildMessageCommand` (a new tool id → async skill/loader resolve).
2. The final consolidated room edit lands via `processRoomTask` → `updateMessage`, setting `body`=final and `isStreamingFinished`=true. `message.isStreamingOfEventFinished` is now `true`.
3. The chain drains: this method runs `updateMessage` with the synthetic event, which unconditionally `setBody(payload.body)` (`message-builder.ts:210`), `setReasoningContent(...)` (`:211`), and — because the synthetic content carries `isStreamingFinished: false` — `setIsStreamingFinished(false)` (`:216`).

Nothing runs afterward to re-reconcile (the final edit was the last room event), so the message is left with stale body/reasoning and `_isStreamingFinished === false`. `room-message.gts`'s `streamingTimeout` getter (`:128`, `:132`) then keeps the spinner up and, after `STREAMING_TIMEOUT_MS`, shows the timeout message.

Tool *args* are safe — `applyToolRequestChunk`'s `origin_server_ts < toolRequestEventTs` guard (`message-builder.ts:328`) blocks the stale write, since the final edit bumped `toolRequestEventTs` to its later ts while the preview's ts is pinned to `message.created`. But `body`/`reasoning`/`isStreamingFinished` have no such guard.

**Why current defenses don't cover it.** The decline on the `isStreamingFinished: false` thread is right that omitting the key wouldn't help (`undefined` un-finishes too), but the conclusion — "serialization + the `isStreamingOfEventFinished` guard protect it" — misses that the guard runs only at *enqueue* and serialization only orders previews *among themselves*, not against the room-event path.

**The fix.** Re-check at apply time, after acquiring the turn and before `updateMessage`:

```js
private async applyResponseStreamPreview(message, payload) {
  if (message.isStreamingOfEventFinished) {
    return; // final room edit already reconciled — nothing for an ephemeral preview to add
  }
  try {
    ...
```

This is sufficient: `updateMessage` writes body/reasoning/finished synchronously *before* its first `await` (the tool loop), so if the apply *starts* before finalization the final edit's synchronous writes still land last. The only unsafe case is an apply that *starts* after finalization — which this guard rejects.

---

packages/host/tests/integration/components/ai-assistant-panel/to-device-streaming-test.gts
```diff
@@ -0,0 +1,241 @@
...
+      body: string,
+      reasoning = '',
+    ) {
+      simulateToDeviceEvent(APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE, {
+        roomId,
+        parentEventId,
+        sequence,
+        body,
+        reasoning,
+        toolRequests: [],
```
habdelra: [Claude Code 🤖] **The two hardest behaviors in this PR have no executable coverage.** Non-blocking, but worth closing before merge since both were the subject of self-review threads.

1. **Tool-request serialization is untested.** Every test drives `preview(...)` with `toolRequests: []`, so `#previewApplyChain` — the headline change of the second commit — is never exercised. Serialization exists precisely because when a preview introduces a *new* tool id, `updateMessage` awaits `buildMessageCommand`, and since all previews pin the same `origin_server_ts`, `applyToolRequestChunk`'s ts guard can't reorder them, so promise-completion order would otherwise decide the winner. A test that fires two overlapping previews carrying tool requests (higher sequence introducing/replacing a tool id, with the lower-sequence build made slower to resolve) and asserts the higher-sequence args win would fail if `#previewApplyChain` were removed. As written, deleting the serialization keeps this suite green.

2. **The `updated`-timestamp assertion doesn't pin what it claims.** Line 181 asserts `messages[0].updated.getTime() >= createdAt`, but `updated` is initialized to `new Date()` at message construction (`message-builder.ts:106`) while `created` is the placeholder's `origin_server_ts` — so `updated >= created` already holds *before any preview applies*. The assertion passes even if `setUpdated` is never called. Capture `messages[0].updated` *before* the previews and assert it strictly advances after, so the test actually covers the stall-timeout reset its name describes.

---

packages/host/app/services/matrix-service.ts
```diff
@@ -2708,6 +2711,34 @@ export default class MatrixService extends Service {
...
+    (payload: AppBoxelResponseStreamContent) => void
+  >();
+
+  private onToDeviceEvent = (e: MatrixEvent) => {
+    if (e.getType() !== APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE) {
+      return;
+    }
+    let payload = e.getContent() as AppBoxelResponseStreamContent;
+    for (let handler of this.#responseStreamPreviewHandlers) {
+      handler(payload);
```
habdelra, line 2737: [Claude Code 🤖] **`onToDeviceEvent` trusts the payload without verifying the sender.** Follow-up / defense-in-depth, non-blocking — practical risk is low, but flagging because this path has a weaker trust model than the room-event path it feeds.

Unlike room events (authored in-room by ai-bot, which the room-event path implicitly trusts), a to-device message can be sent by *any* Matrix user directly to a device — room membership is not required to deliver one. This handler forwards any `app.boxel.response-stream` payload straight to hydration, which renders `body`/`reasoning` into the assistant's bubble. Checking the sender restores parity with the room-event trust model:

```js
private onToDeviceEvent = (e: MatrixEvent) => {
  if (e.getType() !== APP_BOXEL_RESPONSE_STREAM_EVENT_TYPE) {
    return;
  }
  if (e.getSender() !== this.aiBotUserId) {
    return;
  }
  ...
```

Exploitability is genuinely low: an attacker would need a device to target plus a *currently-streaming* `parentEventId`, which for a 1:1 ai room requires room visibility they don't have; and `body` is HTML-escaped in `Message.bodyHTML` (`escapeHtmlOutsideCodeBlocks` before `sanitize:false`), so this is content-spoofing, not XSS. Still cheap to close, and it also short-circuits the fan-out to every `RoomResource` for unrelated to-device traffic.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with a focus on the concurrency contract between the new to-device preview path and the existing room-event path, since that's where this change is load-bearing and where the self-review threads concentrated. Checked out at 4377f26; traced `hydrateResponseStreamPreview` / `applyResponseStreamPreview` against `MessageBuilder.updateMessage`, `applyToolRequestChunk`, the `Message` streaming-finished / `updated` getters, and `room-message.gts`'s stall-timeout getter. CI (host / realm / matrix suites) is green on the head commit.

**Bottom line: one blocking correctness issue** — a preview whose apply is delayed past the final room edit un-finishes and regresses the completed message (details inline on `applyResponseStreamPreview`). The rest is a test-coverage gap on the two trickiest behaviors and one non-blocking hardening suggestion. The overall design is sound and the self-review already resolved the highest-value issues.

**What lands right.**
- Centralizing the `ClientEvent.ToDeviceEvent` binding in matrix-service and fanning out via `onResponseStreamPreview` disposers matches the existing binding pattern and unbinds cleanly in `unbindEventListeners` / `teardown` — no listener leak.
- Reusing `updateMessage` by shaping the preview as the `CardMessage` edit it mirrors is the right call: pinning `origin_server_ts` to `message.created` genuinely makes `applyToolRequestChunk`'s ts guard protect the final edit's tool args from a late preview (verified — the guard blocks `created < finalTs`).
- The "stall timeout resets for free" claim checks out: `streamingTimeout` reads `Number(this.message.updated)` and `updateMessage` stamps a fresh `updated`, so no `room-message.gts` change was needed.
- Per-message serialization is the correct shape for the tool-arg ordering problem given the shared `origin_server_ts`, and isolating apply failures behind try/catch is right for a fire-and-forget handler.

**On the resolved threads.** The serialization, failure-isolation, pruning, and field-clobber-documentation fixes all landed correctly and I agree with them. The one declined thread (`isStreamingFinished: false`) is where the blocking issue lives — the decline is half right (omitting the key doesn't fix it) but the residual race is still open and needs the apply-time re-check, not a change to the literal. Detail in the inline comment.

**Recommendations**
1. (Blocking) Re-check `message.isStreamingOfEventFinished` at the top of `applyResponseStreamPreview` — inline on `room.ts`.
2. Add a tool-request preview test that would fail without `#previewApplyChain`, and tighten the weak `updated` assertion — inline on the test file.
3. (Follow-up) Verify the to-device sender is the ai-bot in `onToDeviceEvent` — inline on `matrix-service.ts`.

**Adjacent, out of scope.** `applyResponseStreamPreview` calls `upsertRoomMember` for `aiBotUserId` on every apply, but `updateMessage` never reads `builderContext.author` (only `buildMessage` does) — so that's dead work per preview. Harmless; not asking you to change it here.

---

packages/host/app/resources/room.ts
lukemelia: [Claude Code 🤖] Fixed in 4a5ff8b257. `applyResponseStreamPreview` now re-checks `message.isStreamingOfEventFinished` after it acquires its turn on the chain and returns before `updateMessage` if the final room edit already finalized the message.

As you noted, this is sufficient: `updateMessage` writes body/reasoning/isStreamingFinished synchronously before its first `await`, so an apply that *starts* before finalization still lets the final edit's synchronous writes land last — only an apply that *starts* after finalization is unsafe, and the guard rejects exactly that.

packages/host/app/services/matrix-service.ts
lukemelia: [Claude Code 🤖] Fixed in 4a5ff8b257. `onToDeviceEvent` now returns unless `e.getSender() === this.aiBotUserId`, restoring parity with the room-event trust model and short-circuiting the fan-out for unrelated to-device traffic. Added a test that a spoofed-sender preview is dropped while the genuine ai-bot sender still applies (the to-device mock now threads an optional sender).

packages/host/tests/integration/components/ai-assistant-panel/to-device-streaming-test.gts
lukemelia: [Claude Code 🤖] Both addressed in 4a5ff8b257.

1. **`updated` assertion** — now captures `updated` after the placeholder loads but *before* any preview, and asserts it *strictly* advances. The old `>= createdAt` held from construction (`updated` seeds to `new Date()`, `created` is the earlier placeholder ts), so it passed even if `setUpdated` never ran; the new form fails in that case.

2. **Tool-request coverage** — added a test that fires two overlapping previews carrying the *same* tool id (no settle between, so both enqueue on `#previewApplyChain` before either finishes) and asserts the message ends with a single tool whose args come from the higher-sequence preview. This exercises the chain and the dedup re-check for the tool path, which every prior test skipped with `toolRequests: []`.

One honest caveat on the mutation-kill you described: with no enabled skill, `buildMessageCommand` resolves without awaiting, so there is no build interleaving to invert — the added test stays green whether or not the chain is present. A test that genuinely *fails* when `#previewApplyChain` is removed needs a timing seam to force the lower-sequence build to resolve last (an enabled skill plus a controllable skill-load delay), which we do not have today. I can add that seam if you want the stronger guarantee — say the word.

habdelra (approved):

PR #5574, opened 2026-07-22
Remove GOOGLE_AUTH_ENABLED flag; gate Google sign-in on homeserver capability

FadhlanR:
## What & why

The "Continue with Google" button on the login screen was gated by **two** conditions:
1. a build-time `GOOGLE_AUTH_ENABLED` feature flag, and
2. a runtime check for whether the homeserver advertises the `oidc-google` identity provider (`googleSsoAvailable`).

Now that Synapse serves the Google OIDC config, the build-time flag is redundant. This PR removes it so the button is shown **purely based on homeserver capability** — it appears wherever the homeserver's `/_matrix/client/v3/login` flows advertise the `oidc-google` IdP, and stays hidden everywhere else.

## Changes

- Remove `GOOGLE_AUTH_ENABLED` from host config (`config/environment.js`, `app/config/environment.ts`) and from `environment-service.ts` (the `googleAuthEnabled` field is gone).
- `login.gts`: always probe the homeserver login flows on mount and derive `showGoogleButton` from `googleSsoAvailable` alone; the `environmentService` injection is no longer needed.
- `login-test.gts`: drop the now-impossible "flag off" test, remove the flag setters from the remaining tests, and re-title the positive case to reflect capability-driven behavior.

## Behavior change

Because the button is now capability-driven, it will render in **any** environment whose homeserver advertises the `oidc-google` IdP (currently staging and production). There is no longer a per-environment build flag to hide it.

## Testing

- `eslint` passes on all changed files.
- Login component integration tests updated; full `lint:types` + test suite run in CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,7 +483,7 @@ mise run infra:stop-admin
 
 #### Google Sign-In (local dev)
 
-The dev Synapse template ships with a Google OIDC block and a custom `user_mapping_provider` (`packages/matrix/support/synapse/modules/boxel_oidc_mapping_provider.py`) that auto-links Google sign-ins to existing Matrix accounts by verified email. The whole block is **gated on env vars** — without them, `packages/matrix/support/synapse/index.ts` strips the OIDC block from the generated `homeserver.yaml` so Synapse boots cleanly for unrelated work. The host's "Sign in with Google" button is also gated on the `GOOGLE_AUTH_ENABLED` feature flag (defaults `true` in `development`).
+The dev Synapse template ships with a Google OIDC block and a custom `user_mapping_provider` (`packages/matrix/support/synapse/modules/boxel_oidc_mapping_provider.py`) that auto-links Google sign-ins to existing Matrix accounts by verified email. The whole block is **gated on env vars** — without them, `packages/matrix/support/synapse/index.ts` strips the OIDC block from the generated `homeserver.yaml` so Synapse boots cleanly for unrelated work. The host's "Sign in with Google" button appears whenever the homeserver's login flows advertise the `oidc-google` IdP, so it shows up automatically once the OIDC block is in place.
 
 To enable Google sign-in locally:
 
--- a/packages/host/app/components/matrix/login.gts
+++ b/packages/host/app/components/matrix/login.gts
@@ -25,7 +25,6 @@ import {
   isMatrixError,
   type MatrixError,
 } from '@cardstack/host/lib/matrix-utils';
-import type EnvironmentService from '@cardstack/host/services/environment-service';
 import type MatrixService from '@cardstack/host/services/matrix-service';
 
 import AuthButton from './auth-button';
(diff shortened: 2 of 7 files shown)


FadhlanR (commented):
[Claude Code 🤖]

This review focused on completeness of the flag removal: whether every home of `GOOGLE_AUTH_ENABLED` — code, config, tests, and docs — moved together, since a half-removed flag leaves dead config and misleading docs behind.

**Bottom line: the code change is correct and the tests are right-sized, but the removal is incomplete — two references to the flag survive outside the touched files. Neither blocks, but both belong in this PR because they're the same removal.**

**What lands right.** Collapsing `showGoogleButton` to `googleSsoAvailable` alone and always running `detectGoogleSso` on mount is the correct capability-driven shape. The test reduction is sound: the dropped "flag off" test is genuinely impossible once the flag is gone, and the remaining three (`PASSWORD_ONLY_FLOW` hides, `OTHER_SSO_FLOW` hides, `GOOGLE_SSO_FLOW` shows) still pin the real contract — the button follows what `/_matrix/client/v3/login` advertises. Removing the `environmentService` injection and the `@tracked` import in `environment-service.ts` is clean.

**Two surviving references (both cleanup, non-blocking, this PR):**

1. `packages/host/config/staging.env` still sets `GOOGLE_AUTH_ENABLED=true`. After this PR, `packages/host/config/environment.js` no longer reads `process.env.GOOGLE_AUTH_ENABLED` (that read was deleted along with the `featureFlags.GOOGLE_AUTH_ENABLED` entry), so this deploy env var is now dead — nothing consumes it. Drop the line to keep staging's env in sync with what the config actually reads.

2. `README.md` (Google Sign-In section) still states the host button "is also gated on the `GOOGLE_AUTH_ENABLED` feature flag (defaults `true` in `development`)." That sentence is now false — the button is gated purely on the homeserver advertising `oidc-google`. Suggested replacement, e.g.: "The host's 'Sign in with Google' button appears whenever the homeserver's login flows advertise the `oidc-google` IdP." The rest of that section (Synapse OIDC env gating, the verify step) stays accurate — step 4's "the button will now appear" reads even better under capability-driven behavior.

**One behavior note worth confirming at merge.** `production.env` never set `GOOGLE_AUTH_ENABLED`, so under the old code the button was hidden in production regardless of what prod Synapse advertised. This PR makes it capability-driven, so the button will now appear in production the moment prod Synapse advertises `oidc-google`. That matches the intent in the description; just worth confirming prod Synapse is actually serving the `oidc-google` IdP (otherwise the button silently stays hidden — which is the desired fail-safe, but means the feature isn't live until the homeserver side is in place).

Recommendations:
1. Remove `GOOGLE_AUTH_ENABLED=true` from `packages/host/config/staging.env` — dead env var post-removal.
2. Update the `GOOGLE_AUTH_ENABLED` sentence in the README Google Sign-In section to capability-driven wording.

PR #5576, opened 2026-07-22
Hide AI assistant by default

lukemelia:
## Summary

The AI Assistant panel opened by default on a first-ever visit. This defaults it to **closed** so the initial experience — particularly for new users during onboarding — isn't dominated by the assistant.

Resolution order is unchanged aside from the first-visit default:

- **Fresh visit, never toggled** → closed
- **URL `operatorModeState` carries `aiAssistantOpen`** → wins (open or closed)
- **Opened last session (same login)** → reopens, via the remembered `boxel-ai-assistant-open` localStorage preference
- **On logout** → the remembered preference is cleared, so the next login starts closed

## Changes

- `operator-mode-state-service`: `readPersistedAiAssistantOpen()` now defaults to `false` (was `true`) for first-ever visits.
- `local-storage-keys`: `clearLocalStorage()` (called on logout) now removes the `boxel-ai-assistant-open` key.

## Test coverage

- **Unit** (`operator-mode-state-service-test`): closed-by-default with no URL/localStorage; remembered-open path; existing URL-precedence tests unchanged.
- **Acceptance**: end-to-end closed-by-default when the URL omits `aiAssistantOpen` (`ai-assistant-test`); logout clears the preference (`operator-mode-acceptance-test`).

Resolves [CS-12273](https://linear.app/cardstack/issue/CS-12273/hide-ai-assistant-by-default)

### DIFF
--- a/packages/matrix/helpers/index.ts
+++ b/packages/matrix/helpers/index.ts
@@ -31,6 +31,7 @@ interface ProfileAssertions {
 interface LoginOptions {
   url?: string;
   showAllCards?: boolean; //default true
+  openAiAssistant?: boolean; //default false; opens the AI assistant panel after login
 }
 
 // Shared SQL executor is created on demand for tests
@@ -157,7 +158,21 @@ export async function reloadAndOpenAiAssistant(page: Page) {
 }
 
 export async function openAiAssistant(page: Page) {
-  await page.locator('[data-test-open-ai-assistant]').click();
+  // The AI assistant panel is closed by default; the toggle button both opens
+  // and closes it. Wait for the toggle to render (operator mode's open/closed
+  // state is restored synchronously before first paint, so once the button is
+  // present the panel's presence already reflects that state) and only click
+  // when the panel is actually closed — otherwise a call made when it is
+  // already open (restored from URL/localStorage state) would toggle it back
+  // closed.
+  let openButton = page.locator('[data-test-open-ai-assistant]');
+  await openButton.waitFor();
+  let panel = page.locator('[data-test-ai-assistant-panel]');
+  if (await panel.isVisible()) {
+    return;
+  }
+  await openButton.click();
+  await panel.waitFor();
 }
 
 export async function createRealm(
@@ -440,6 +455,10 @@ export async function login(
   }, localStorageAuth);
 
   await openRoot(page, opts?.url);
+
+  if (opts?.openAiAssistant) {
+    await openAiAssistant(page);
+  }
 }
 
 export async function enterWorkspace(
--- a/packages/host/app/services/operator-mode-state-service.ts
+++ b/packages/host/app/services/operator-mode-state-service.ts
@@ -143,12 +143,13 @@ export type ModuleInspectorView = 'schema' | 'spec' | 'preview';
 export const DEFAULT_MODULE_INSPECTOR_VIEW: ModuleInspectorView = 'schema';
 
 // Read the user's persisted AI Assistant open/closed preference. Defaults to
-// `true` for first-ever visits; URL state still takes precedence over this.
+// `false` for first-ever visits, so the panel is closed unless the URL state
+// (which takes precedence) or a remembered preference opens it.
 function readPersistedAiAssistantOpen(): boolean {
   let raw = window.localStorage.getItem(AiAssistantOpen);
   if (raw === 'true') return true;
   if (raw === 'false') return false;
-  return true;
+  return false;
 }
 
 export default class OperatorModeStateService extends Service {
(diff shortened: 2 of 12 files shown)


packages/matrix/helpers/index.ts
```diff
@@ -157,7 +158,21 @@ export async function reloadAndOpenAiAssistant(page: Page) {
...
+  // closed.
+  let openButton = page.locator('[data-test-open-ai-assistant]');
+  await openButton.waitFor();
+  let panel = page.locator('[data-test-ai-assistant-panel]');
+  if (await panel.isVisible()) {
+    return;
+  }
+  await openButton.click();
+  await panel.waitFor();
 }
```
habdelra, line 176: [Claude Code 🤖]

**Confirmation, plus the one invariant this early-return leans on.** The idempotency is correct, and it's what makes the whole test migration safe: `data-test-open-ai-assistant` is rendered by `AiAssistantButton` in `submode-layout.gts` whenever `!isAiAssistantHidden` (any submode except Host), independent of open/closed — so `openButton.waitFor()` resolves in both states rather than hanging. The panel (`data-test-ai-assistant-panel`, on the outer `<div>` in `ai-assistant/panel.gts`) is a sibling `ResizablePanel` gated only by `{{#if aiAssistantPanelService.isOpen}}` in the same template, so it renders in the same Glimmer pass as the button. Once the button is visible the panel's presence already reflects `isOpen`, which is exactly what the comment claims — verified against the templates, and it holds today.

**Guard-rail for whoever edits the panel next.** `isVisible()` is a point-in-time read with no auto-wait, so its correctness depends entirely on "panel presence is synchronous with the button." If someone later gates `data-test-ai-assistant-panel` behind async room loading — moving the attribute off the outer div, or wrapping the panel body so the attribute only appears after `roomResource` resolves — then in the already-open case this reads false before the panel paints, the helper clicks, and it toggles the panel **closed**. Nothing asserts it stayed open, so that failure would surface as a confusing downstream "panel not found." Keeping the data-test attribute on the always-rendered outer element preserves the invariant.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I focused on two things: whether the closed-by-default resolution order is preserved end-to-end (URL → remembered preference → default), and whether the matrix test migration is *complete* — flipping the default silently breaks any Playwright spec that assumed the room panel was present after navigating, so a missed spec is the main risk here.

**Bottom line: correct and well-covered — no blocking issues.** The behavior change is a one-line default flip plus a logout-clear, both guarded by focused unit + acceptance tests, and the matrix migration covers every spec and every test that touches the panel.

**What lands right:**
- The resolution order is genuinely unchanged except for the first-visit default. `deserialize()` reads `rawState.aiAssistantOpen ?? readPersistedAiAssistantOpen()`, so an explicit URL value still wins and a remembered `true`/`false` in localStorage still wins over the new default. The unit suite pins all six cells of that matrix (URL true/false over localStorage; localStorage true/false with no URL key; and the bare default), which is the right altitude to lock this behavior down.
- Logout clearing is wired correctly. `clearLocalStorage()` — called from matrix-service `logout()` — now removes `AiAssistantOpen`, and the acceptance test asserts the key is `null` after signout, so "next login starts closed" is executable rather than assumed.
- The test migration is complete. Every matrix spec referencing a panel-dependent selector (auth-rooms, correctness-checks, messages, room-creation, skills, tools) opens the assistant; the SDK-only spec (response-stream-to-device) correctly stays untouched. Within the large files, every test reaches the open call — directly, via the `openAiAssistant: true` login option, via `reloadAndOpenAiAssistant`, or via a shared setup helper (e.g. the `auto-attachment` block's `setupAutoAttachmentUser`, which the diff updated). Making `openAiAssistant` idempotent is what lets the reload/relogin paths call it unconditionally without risking a toggle-closed — see the inline note for the invariant that rests on.

**Nuance on the "URL wins (open or closed)" framing.** `deserialize` does honor an explicit `aiAssistantOpen: false` (the unit test "URL aiAssistantOpen=false wins over localStorage=true" proves it), but the app never *produces* such a URL: `rawStateWithSavedCardsOnly()` only writes the key when it's truthy. So the app-generated URL only ever carries `true`; the explicit-`false`-in-URL branch is reachable only via a hand-crafted URL — which is exactly what the new acceptance test builds to exercise the default fallback rather than the URL-wins path. Not a bug and not changed here, just noting the description's "(open or closed)" slightly overstates what the running app emits.

**Recommendation:**
1. Confirm the matrix Playwright job lands green on the head commit. The sticky status reflects the host suite on the first commit (the source change), which is 0-fail; the matrix specs are the suite that actually exercises the second commit's helper rewrite. I verified their completeness statically but can't run the dockerized Playwright suite from here.

**Adjacent, out of scope.** `clearLocalStorage()` still leaves `SessionLocalStorageKey` and `SpecSelection` in place — presumably intentional, and unrelated to this change; flagging only for whoever next audits what logout forgets.

---

burieberry (approved):

PR #5579, opened 2026-07-22
ai-bot: per-turn telemetry line for streaming-mode comparison

lukemelia:
Instrumentation to quantify the ai-bot streaming-mode work with measured before/after numbers rather than estimates.

## What

Every finished ai-bot turn emits one structured, greppable line from the `Responder`:

```
[turn-telemetry] mode=… durationMs=… roomEvents=… toDeviceEvents=… promptTokens=… completionTokens=… canceled=… roomId=… agentId=… responseEventId=…
```

These are the per-turn numbers needed to compare the `room-edits`, `off`, and `to-device` streaming modes under load: room-event volume is what drives Synapse CPU, and to-device count / duration / tokens show whether cutting room events costs latency.

## How

- **Room-event count** is tallied at the single choke point every room event passes through — `MatrixResponsePublisher` (thinking placeholder, streamed edits, continuation splits, and errors) — so the count reflects true Matrix load regardless of mode.
- The **Responder** tallies to-device previews and token usage (from the end-of-stream `chunk.usage`) and emits the line from `finalize()`.
- Counts are also exposed via a `turnTelemetry` getter so tests assert them directly instead of parsing the log.

Format is single-line `key=value` to match the repo's other request-timing channels (e.g. `realm:requests dur=`), so it parses cleanly with Loki's `pattern` + `unwrap`.

## Testing

- New unit test asserts the per-turn counts (`roomEvents`, `toDeviceEvents`, tokens, mode) for a `to-device` turn.
- Existing responder / streaming-mode tests pass unchanged (the line renders correctly across `room-edits` / `off` / `to-device`).

### DIFF
--- a/packages/ai-bot/lib/matrix/response-publisher.ts
+++ b/packages/ai-bot/lib/matrix/response-publisher.ts
@@ -63,6 +63,11 @@ export default class MatrixResponsePublisher {
   readonly responseState: ResponseState;
   eventSizeMax = DEFAULT_EVENT_SIZE_MAX;
   responseEvents: ResponseEventData[] | undefined;
+  // Count of Matrix room events this publisher has sent this turn (thinking
+  // placeholder, streamed message edits, continuation splits, and errors).
+  // Read by the Responder's per-turn telemetry line to compare room-event
+  // volume across streaming modes.
+  roomEventsEmitted = 0;
   private sendingMessage = Promise.resolve(); // track pending send operation
 
   get currentResponseEvent() {
@@ -146,6 +151,7 @@ export default class MatrixResponsePublisher {
           ),
           reasoningAndContent.reasoning,
         );
+        this.roomEventsEmitted++;
         if (!this.currentResponseEvent.eventId) {
           this.currentResponseEvent.eventId = messageEvent.event_id;
         }
@@ -189,6 +195,7 @@ export default class MatrixResponsePublisher {
           ),
         contentAndReasoning.reasoning,
       );
+      this.roomEventsEmitted++;
       if (!this.currentResponseEvent.eventId) {
         this.currentResponseEvent.eventId = messageEvent.event_id;
       }
@@ -209,6 +216,7 @@ export default class MatrixResponsePublisher {
       this.originalResponseEventId,
       opts,
     );
+    this.roomEventsEmitted++;
   }
 
   async ensureThinkingMessageSent() {
@@ -231,5 +239,6 @@ export default class MatrixResponsePublisher {
     this.responseEvents = [
       new ResponseEventData(initialMessage.event_id, this.eventSizeMax),
     ];
+    this.roomEventsEmitted++;
   }
 }
--- a/packages/ai-bot/lib/responder.ts
+++ b/packages/ai-bot/lib/responder.ts
@@ -59,14 +59,18 @@ export class Responder {
   private streamPreviewTarget: { userId: string; deviceId: string } | undefined;
   private _streamPreviewSequence = 0;
 
-  // Per-turn telemetry, logged once from finalize() to compare streaming modes.
-  // startedAt is set at construction — the turn boundary as the Responder sees
-  // it. Room-event count lives on the publisher; to-device previews and token
-  // usage are tallied here.
+  // Per-turn telemetry, logged once per turn to compare streaming modes.
+  // startedAt is the turn boundary as the Responder sees it (construction, i.e.
+  // before prompt construction and the per-user cost-lock wait); streamStartedAt
+  // marks the first chunk, so streamMs isolates the generation+streaming+finalize
+  // window from that mode-independent pre-generation time. Room-event count lives
+  // on the publisher; to-device previews and token usage are tallied here.
   private startedAt = Date.now();
+  private streamStartedAt: number | undefined;
   private toDeviceEventsEmitted = 0;
   private promptTokens: number | undefined;
   private completionTokens: number | undefined;
+  private telemetryLogged = false;
 
   constructor(
     client: MatrixClient,
@@ -244,6 +248,10 @@ export class Responder {
     chunk: OpenAI.Chat.Completions.ChatCompletionChunk,
     snapshot: ChatCompletionSnapshot,
   ) {
+    // Mark the start of the streaming window on the first chunk so streamMs
+    // measures generation+streaming+finalize, not the pre-generation wait.
+    this.streamStartedAt ??= Date.now();
+
     // reasoning does not support snapshots, so we need to handle the delta
     const newReasoningContent = (
       chunk.choices?.[0]?.delta as { reasoning?: string }
@@ -315,7 +323,14 @@ export class Responder {
         agentId: this.matrixResponsePublisher.agentId,
       },
     });
-    return await this.matrixResponsePublisher.sendError(error, opts);
+    let result = await this.matrixResponsePublisher.sendError(error, opts);
+    // An errored turn still emits room events (the placeholder + this error
+    // event, plus any mid-turn edits already sent), and several error paths in
+    // main.ts end the turn here without ever calling finalize(). Log telemetry
+    // so that room-event volume is not missing from the comparison; the guard
+    // in logTurnTelemetry keeps it to a single line when finalize() also runs.
+    this.logTurnTelemetry();
+    return result;
   }
 
   async flush() {
@@ -357,10 +372,17 @@ export class Responder {
 
   // Per-turn measurements used to compare streaming modes. Exposed as a getter
   // so tests can assert the counts directly rather than parse the log line.
+  // durationMs is whole-turn wall-clock (includes pre-generation queue/lock
+  // wait); streamMs isolates the streaming window and is undefined for turns
+  // that errored before the first chunk.
   get turnTelemetry() {
     return {
       mode: this.streamingMode,
       durationMs: Date.now() - this.startedAt,
+      streamMs:
+        this.streamStartedAt === undefined
+          ? undefined
+          : Date.now() - this.streamStartedAt,
       roomEvents: this.matrixResponsePublisher.roomEventsEmitted,
       toDeviceEvents: this.toDeviceEventsEmitted,
       promptTokens: this.promptTokens,
@@ -375,11 +397,19 @@ export class Responder {
   // One structured, greppable line per turn so a scripted load run can compare
   // Matrix event volume and latency across streaming modes in Loki. Keep it
   // single-line key=value to stay consistent with the repo's other
-  // request-timing channels (e.g. realm:requests `dur=`).
+  // request-timing channels (e.g. realm:requests `dur=`). Fires at most once
+  // per turn — from finalize() on the normal/canceled paths, or from onError()
+  // on the error paths that never finalize — so every turn that emitted room
+  // events also emits exactly one line.
   private logTurnTelemetry() {
+    if (this.telemetryLogged) {
+      return;
+    }
+    this.telemetryLogged = true;
     let t = this.turnTelemetry;
     log.info(
       `[turn-telemetry] mode=${t.mode} durationMs=${t.durationMs} ` +
+        `streamMs=${t.streamMs ?? ''} ` +
         `roomEvents=${t.roomEvents} toDeviceEvents=${t.toDeviceEvents} ` +
         `promptTokens=${t.promptTokens ?? ''} ` +
         `completionTokens=${t.completionTokens ?? ''} ` +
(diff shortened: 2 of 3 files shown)


packages/ai-bot/lib/matrix/response-publisher.ts
```diff
@@ -63,6 +63,11 @@ export default class MatrixResponsePublisher {
   readonly responseState: ResponseState;
   eventSizeMax = DEFAULT_EVENT_SIZE_MAX;
   responseEvents: ResponseEventData[] | undefined;
+  // Count of Matrix room events this publisher has sent this turn (thinking
+  // placeholder, streamed message edits, continuation splits, and errors).
+  // Read by the Responder's per-turn telemetry line to compare room-event
+  // volume across streaming modes.
+  roomEventsEmitted = 0;
```
habdelra, line 70: [Claude Code 🤖] **Confirmation (nothing to change) — the choke-point claim holds.** The counter is placed correctly, and I verified the "single choke point" premise rather than taking it on faith.

**Every response-path room send goes through this class and increments the counter.** Grepping the ai-bot lib for room-event send sites (`sendMessageEvent(` / `sendErrorEvent(` / `client.sendEvent(`) outside tests, the response path resolves to exactly four call sites, all here and all followed by `roomEventsEmitted++`: the thinking placeholder (`ensureThinkingMessageSent`), the final consolidated send (`sendMessage`), each continuation split (the `while` loop in `sendMessage`), and the error event (`sendError`) — the four sources your comment names.

**The other two room-send sites in the package are correctly excluded.** `code-patch-correctness.ts` (`client.sendEvent`) runs on the `publishCodePatchCorrectnessMessage` early-return path, and `debug.ts` only fires in debug mode — neither reaches `finalize()`, so no telemetry line is emitted for them and there's no under/over-count.

**The thinking placeholder is counted once even though `main.ts` calls `ensureThinkingMessageSent()` twice** (once for the definitely-triggers case, once after prompt construction) — the `if (this.initialMessageSent) return;` guard short-circuits the second call before the increment. So `roomEvents` for a normal turn is `thinking(1) + edits/splits + final`, exactly as intended.

Non-blocking, and I'd keep it as-is.

---

packages/ai-bot/lib/responder.ts
```diff
@@ -340,5 +352,39 @@ export class Responder {
       await this.sendMessageEventWithThrottling();
     }
     await this.flush();
+    this.logTurnTelemetry();
```
habdelra, line 370: [Claude Code 🤖] **Finding (measurement gap, non-blocking, likely a follow-up): errored turns emit room events but no telemetry line, so the Loki totals won't reconcile with real Matrix load.** `logTurnTelemetry()` is reachable only through `finalize()`, and `finalize()` is not called on every turn-end path.

**The paths, traced in `main.ts`:**
- Normal completion → `responder.finalize()` — logs. ✅
- User abort (`APIUserAbortError`) → `responder.finalize({ isCanceled: true })` — logs, `canceled=true`. ✅
- Chat-history processing error → `onError(...)` then `responder.finalize()` — logs. ✅
- **Generic generation/finalization error** → `onError(error)` with **no** `finalize()`. ❌
- **Credit-tracking error** → `onError(...); return;` — no `finalize()`. ❌
- **Insufficient credits** → `onError(...); return;` — no `finalize()`. ❌

Each of those three ❌ turns still sends room events that this PR counts — at minimum the thinking placeholder plus the error event (both increment `roomEventsEmitted` on the publisher) — but the turn produces no `[turn-telemetry]` line, so that room-event volume is invisible to the comparison. If error rates differ across the three streaming modes (e.g. the `413 event too large` path is more likely in a mode that sends larger/more room edits), the mode comparison is biased by exactly the turns most worth counting.

**It's also internally inconsistent:** the history-processing path finalizes after `onError`, but the generic error path doesn't — so "does an errored turn log telemetry?" depends on *which* error.

**Two ways out**, your call which fits scope:
1. Accept it and document it — tighten the getter/method comment to "logged once per *completed or canceled* turn; errored turns are not measured," so the analyst knows the totals exclude error turns by design. Lowest-risk for an instrumentation PR.
2. Emit on the error paths too — e.g. call `logTurnTelemetry()` from `onError` (guarded so it fires at most once per turn, mirroring `isFinalized`), or add a `finalize()` to the generic-error/credit branches. This makes the room-event totals actually reconcile.

I'd lean (1) for this PR since the scripted load run targets successful turns, but the current comment reads as "one line per turn," which overstates the coverage. Not blocking.

---

packages/ai-bot/lib/responder.ts
```diff
@@ -59,6 +59,15 @@ export class Responder {
   private streamPreviewTarget: { userId: string; deviceId: string } | undefined;
   private _streamPreviewSequence = 0;
 
+  // Per-turn telemetry, logged once from finalize() to compare streaming modes.
+  // startedAt is set at construction — the turn boundary as the Responder sees
+  // it. Room-event count lives on the publisher; to-device previews and token
+  // usage are tallied here.
+  private startedAt = Date.now();
```
habdelra: [Claude Code 🤖] **Finding (interpretation caveat for `durationMs`, non-blocking): this span is whole-turn wall-clock, not streaming latency — and under a load run the non-streaming part can dominate.** The PR description frames `durationMs` as showing "whether cutting room events costs latency," so it's worth being precise about what it actually measures.

`startedAt` is stamped when `new Responder(...)` runs, which in `main.ts` is *before* `getPromptParts(...)` (prompt construction) and *before* the `withUserCostLock` block wrapping generation. So `durationMs` = prompt construction + `waitForPendingCreditTracking` + `validateAICredits` + **per-user cost-lock acquisition** + generation + streaming + finalize.

The streaming mode only affects the send/finalize portion; everything before it is common-mode. The cost lock in particular serializes a user's requests across all rooms/replicas, so if the load harness fires concurrent requests as one (or few) users, lock-wait time is both large and mode-independent — it can swamp the very difference you're trying to isolate.

Not a code defect, and no change required to the counter itself. Options: (a) leave it and note in the analysis that `durationMs` is turn-wall-clock including queueing, comparing modes only in aggregate; or (b) if you want a latency signal that isolates the streaming cost, stamp a second start at the first send (or reuse the existing `requestStart`/`firstChunkAt` timing already computed in `main.ts`) and report that span instead. The `startedAt` comment ("the turn boundary as the Responder sees it") is accurate but understates how much pre-generation time is folded in.

---

packages/ai-bot/tests/responding-test.ts
```diff
@@ -352,6 +352,58 @@ module('Responding', (hooks) => {
...
+      let toDeviceCount = fakeMatrixClient.getSentToDeviceEvents().length;
+      await responder.finalize();
+
+      let telemetry = responder.turnTelemetry;
+      assert.equal(telemetry.mode, 'to-device', 'mode is recorded');
+      assert.equal(
+        telemetry.roomEvents,
+        2,
+        'room events = thinking placeholder + one final consolidated event',
+      );
```
habdelra, line 386: [Claude Code 🤖] **Finding (test coverage, non-blocking): the assertion that matters most for this PR's purpose — that `roomEvents` is *high* in `room-edits` and *low* elsewhere — isn't pinned.** This test verifies the `to-device` "after" number (`roomEvents === 2`, one preview channel, tokens captured), which is good. But the whole reason the line exists is the *comparison*: room-edits should produce many room events (one per mid-turn throttled edit), `off`/`to-device` should produce ~2.

Nothing here exercises the `room-edits` branch, so if a future change stopped incrementing `roomEventsEmitted` on the mid-turn `sendMessage()` edits (line 198 in `response-publisher.ts`), this suite would stay green while the central metric silently under-reported — the "before" side of every comparison would collapse to ~2 and no test would notice.

I traced the `room-edits` path to confirm it's easy to assert: with `AI_BOT_STREAMING_MODE` unset (or `'room-edits'`) and no preview target, each throttled `onChunk`→`sendMessageEventWithThrottling` resolves to `sendMessageEvent()`→`publisher.sendMessage()`, so N distinct content deltas across the 300ms ticks yield thinking(1) + N edits + final. A second test that drives a few content chunks in `room-edits` and asserts `telemetry.roomEvents > 2` and `telemetry.toDeviceEvents === 0` would lock down the contract the instrumentation is built to measure. Worth adding here; not blocking the merge.

---

habdelra (commented):
[Claude Code 🤖] Reviewed as an instrumentation change: whether the per-turn line measures what it claims (room-event volume, to-device volume, latency, tokens), and whether the counts are placed where they can't drift from real Matrix load. Verified against the checked-out branch; AI-bot Tests and Lint are green.

**Bottom line: no blocking issues.** The room-event counter is in the right place — `MatrixResponsePublisher` really is the single choke point for the response path, so `roomEvents` reflects true room-event volume across all three modes, which is the load-driving metric the work needs. The findings below are all non-blocking and concern how faithfully the *emitted* numbers map to the questions the load run wants to answer.

- **Choke-point placement.** All four response-path room sends route through the publisher and increment the counter; the two other room-send sites in the package (code-patch-correctness, debug) are on paths that never finalize, so they're correctly excluded. The thinking placeholder is counted exactly once despite `ensureThinkingMessageSent()` being invoked twice in `main.ts`. Detail in the thread on `roomEventsEmitted`.
- **Mode discrimination is real.** In `room-edits` each throttled edit is a distinct Matrix event, so counting every `sendMessage()` is the correct Synapse-load proxy; in `to-device` the mid-turn previews go over `sendToDevice` and never touch the room counter, so `roomEvents` collapses to ~2 — exactly the reduction the effort aims to demonstrate.
- **`canceled` is captured correctly:** `finalize({ isCanceled })` → `updateIsStreamingFinished(true, isCanceled)` sets the flag before `logTurnTelemetry()` reads it. Token capture reuses the existing single-shot `chunk.usage` handling.
- Exposing the counts via the `turnTelemetry` getter and asserting those rather than parsing the log line is the right call.

1. Errored turns emit room events but no telemetry line — `logTurnTelemetry()` fires only from `finalize()`, which the generic-error, credit-tracking, and insufficient-credits paths skip. Either document "completed/canceled turns only" or emit on those paths so the Loki totals reconcile. See the thread on `this.logTurnTelemetry()`.
2. `durationMs` is whole-turn wall-clock (includes prompt construction + credit checks + per-user cost-lock wait), not streaming latency; under concurrent load the lock-wait can swamp the mode difference. See the thread on the `startedAt` comment.
3. Add a `room-edits`/`off` assertion so the "before" side of the comparison (`roomEvents > 2`, `toDeviceEvents === 0`) is pinned, not just the `to-device` "after" number. See the thread on the test.

**Minor / adjacent, out of scope:** `promptTokens`/`completionTokens` render as empty (`promptTokens=`) when usage never arrived, which won't `unwrap` cleanly in Loki — but per (1) those turns mostly don't log anyway, so it's cosmetic. And `sendError` increments the counter without awaiting `sendErrorEvent(...)` (pre-existing non-await), so a failed error-send still counts by one; negligible.

---

packages/ai-bot/lib/responder.ts
lukemelia: [Claude Code 🤖] Fixed in 12efa2680b (went with option 2). `logTurnTelemetry` now also fires from `onError()`, guarded by a `telemetryLogged` flag so it emits at most once per turn — the normal/canceled paths still log from `finalize()`, and the generic-error / credit / insufficient-credits paths now log from `onError()`. So error turns (which carry the placeholder + error event, plus any mid-turn edits already sent) reconcile in the Loki totals, and the 413-heavy `room-edits` case is no longer silently undercounted. New test asserts one emission across an `onError`-without-`finalize` turn plus a late `finalize`.

packages/ai-bot/lib/responder.ts
lukemelia: [Claude Code 🤖] Fixed in 12efa2680b (option b). Added a `streamMs` field stamped from the first chunk to finalize, alongside the existing whole-turn `durationMs`. `streamMs` isolates the generation+streaming+finalize window from the mode-independent pre-generation time (prompt construction + cost-lock wait), so the latency comparison is not swamped by lock contention under concurrent load. Both are in the line and the `turnTelemetry` getter; `streamMs` is empty for turns that errored before the first chunk. Updated the `startedAt` comment to say so explicitly.

packages/ai-bot/tests/responding-test.ts
lukemelia: [Claude Code 🤖] Added in 12efa2680b. New test drives 5 content chunks in the default `room-edits` mode (no preview target) and asserts `telemetry.roomEvents > 2` and `telemetry.toDeviceEvents === 0`, pinning the "before" side so a regression that stopped counting mid-turn edits can no longer keep the suite green while the metric collapses to ~2.

habdelra (approved):

PR #5580, opened 2026-07-22
Add ProcessCard base card for Workspace setup-progress

lukemelia:
## What

Adds `ProcessCard` (`packages/base/process-card.gts`), a base CardDef representing a long-running setup-progress job (indexing, import, a guided setup flow).

## Why

A realm's index card surfaces setup-progress jobs on its Home tab by querying the realm for `ProcessCard` instances via a `codeRef` (`codeRef(here, './process-card', 'ProcessCard')`) rather than a static import. The field names here are therefore a contract that consumer reads, and the card must exist in the base realm for those jobs to resolve and render.

(Unrelated to the `Wiki/process-card.json` fixture in the software-factory realm, which is a documentation page that adopts `Wiki` and merely shares the filename.)

## Fields (the contract)

`listingName`, `stage`, `processStatus` (StringField), `progressDone`/`progressTotal` (NumberField), `startedAt` (DateTimeField), and an optional linked `setupSurvey`. Plus computed `cardTitle` (falls back to "Setup process") and getters `percentComplete` / `progressLabel` / `statusLabel`, driving a shared isolated/embedded/fitted template. The bar is the shared themed `ProgressBar` from boxel-ui; its progressbar ARIA semantics (`role="progressbar"`, `aria-value*`) are added to the shared component in #5586.

## Tests

- `packages/host/tests/integration/components/process-card-test.gts` — 10 tests covering the getters (fraction, clamping, unknown-total, status default) and rendering (aria attributes, copy, fill width, fallbacks).
- `packages/host/tests/helpers/base-realm.ts` — exports `ProcessCard` from the base-realm test helper (mirrors the `CardsGrid` wiring) so tests exercise the real base card via the loader.

Test plan: runs green locally against a full dev stack — 10 passed, 0 failed. Clean under glint (`ember-tsc --noEmit`), `ember-template-lint`, and prettier.

### DIFF
--- a/packages/base/process-card.gts
+++ b/packages/base/process-card.gts
@@ -100,7 +100,7 @@ class ProcessTemplate extends Component<typeof ProcessCard> {
 }
 
 export class ProcessCard extends CardDef {
-  static displayName = 'Process Card';
+  static displayName = 'Process';
   static icon = ProgressIcon;
 
   // The name shown on the setup bar; falls back to the card title.
--- a/packages/base/process-card.gts
+++ b/packages/base/process-card.gts
@@ -88,7 +88,8 @@ export class ProcessCard extends CardDef {
   static displayName = 'Process';
   static icon = ProgressIcon;
 
-  // The name shown on the setup bar; falls back to the card title.
+  // The name shown on the setup bar. `cardTitle` reads this, defaulting to
+  // "Setup process" when it is unset.
   @field listingName = contains(StringField);
   // Human-readable label for the current step, e.g. "Importing files".
   @field stage = contains(StringField);
--- a/packages/base/process-card.gts
+++ b/packages/base/process-card.gts
@@ -1,5 +1,3 @@
-import { htmlSafe } from '@ember/template';
-
 import {
   CardDef,
   Component,
@@ -11,6 +9,7 @@ import {
 import NumberField from './number';
 import DateTimeField from './datetime';
 
+import { ProgressBar } from '@cardstack/boxel-ui/components';
 import ProgressIcon from '@cardstack/boxel-icons/progress';
 
 // A setup-progress job: a long-running task (indexing, import, a guided
@@ -28,15 +27,7 @@ class ProcessTemplate extends Component<typeof ProcessCard> {
         </span>
         <h3 class='process-card__name'><@fields.cardTitle /></h3>
       </header>
-      <div
-        class='process-card__bar'
-        role='progressbar'
-        aria-valuenow={{@model.percentComplete}}
-        aria-valuemin='0'
-        aria-valuemax='100'
-      >
-        <div class='process-card__fill' style={{@model.barStyle}}></div>
-      </div>
+      <ProgressBar @value={{@model.percentComplete}} @max={{100}} />
       <footer class='process-card__footer'>
         <span class='process-card__count'>{{@model.progressLabel}}</span>
         <span
@@ -56,40 +47,34 @@ class ProcessTemplate extends Component<typeof ProcessCard> {
       .process-card__header {
         display: flex;
         flex-direction: column;
-        gap: var(--boxel-sp-xxxs);
+        gap: var(--boxel-sp-3xs);
         min-width: 0;
       }
       .process-card__stage {
-        color: var(--boxel-450);
-        font: 500 var(--boxel-font-xs);
+        color: var(--muted-foreground);
+        font-weight: 500;
+        font-size: var(--boxel-font-size-xs);
+        line-height: var(--boxel-line-height-xs);
         letter-spacing: var(--boxel-lsp-sm);
         text-transform: uppercase;
       }
       .process-card__name {
         margin: 0;
-        font: 600 var(--boxel-font);
+        font-weight: 600;
+        font-size: var(--boxel-font-size);
+        line-height: var(--boxel-line-height);
         overflow: hidden;
         text-overflow: ellipsis;
         white-space: nowrap;
-      }
-      .process-card__bar {
-        height: 6px;
-        border-radius: var(--boxel-border-radius-sm);
-        background-color: var(--boxel-200);
-        overflow: hidden;
-      }
-      .process-card__fill {
-        height: 100%;
-        border-radius: inherit;
-        background-color: var(--boxel-highlight);
-        transition: width 0.3s ease;
       }
       .process-card__footer {
         display: flex;
         justify-content: space-between;
         gap: var(--boxel-sp-xs);
-        color: var(--boxel-450);
-        font: 500 var(--boxel-font-xs);
+        color: var(--muted-foreground);
+        font-weight: 500;
+        font-size: var(--boxel-font-size-xs);
+        line-height: var(--boxel-line-height-xs);
         letter-spacing: var(--boxel-lsp-sm);
       }
       .process-card__status {
@@ -147,10 +132,6 @@ export class ProcessCard extends CardDef {
 
   get statusLabel(): string {
     return this.processStatus ?? 'running';
-  }
-
-  get barStyle() {
-    return htmlSafe(`width: ${this.percentComplete}%`);
   }
 
   static isolated = ProcessTemplate;
--- a/packages/base/process-card.gts
+++ b/packages/base/process-card.gts
@@ -1,0 +1,159 @@
+import { htmlSafe } from '@ember/template';
+
+import {
+  CardDef,
+  Component,
+  StringField,
+  contains,
+  field,
+  linksTo,
+} from './card-api';
+import NumberField from './number';
+import DateTimeField from './datetime';
+
+import ProgressIcon from '@cardstack/boxel-icons/progress';
+
+// A setup-progress job: a long-running task (indexing, import, a guided
+// setup flow) whose progress the workspace surfaces on the Home tab's setup
+// bar. The `Workspace` card discovers these by querying the realm for
+// `ProcessCard` instances, so the field names here are a contract the
+// Workspace reads (`listingName`, `progressDone`, `progressTotal`, `stage`,
+// `startedAt`, `setupSurvey`, `processStatus`).
+class ProcessTemplate extends Component<typeof ProcessCard> {
+  <template>
+    <article class='process-card'>
+      <header class='process-card__header'>
+        <span class='process-card__stage'>
+          {{if @model.stage @model.stage 'In progress'}}
+        </span>
+        <h3 class='process-card__name'><@fields.cardTitle /></h3>
+      </header>
+      <div
+        class='process-card__bar'
+        role='progressbar'
+        aria-valuenow={{@model.percentComplete}}
+        aria-valuemin='0'
+        aria-valuemax='100'
+      >
+        <div class='process-card__fill' style={{@model.barStyle}}></div>
+      </div>
+      <footer class='process-card__footer'>
+        <span class='process-card__count'>{{@model.progressLabel}}</span>
+        <span
+          class='process-card__status'
+          data-status={{@model.statusLabel}}
+        >{{@model.statusLabel}}</span>
+      </footer>
+    </article>
+    <style scoped>
+      .process-card {
+        display: flex;
+        flex-direction: column;
+        gap: var(--boxel-sp-xs);
+        padding: var(--boxel-sp);
+        container-type: inline-size;
+      }
+      .process-card__header {
+        display: flex;
+        flex-direction: column;
+        gap: var(--boxel-sp-xxxs);
+        min-width: 0;
+      }
+      .process-card__stage {
+        color: var(--boxel-450);
+        font: 500 var(--boxel-font-xs);
+        letter-spacing: var(--boxel-lsp-sm);
+        text-transform: uppercase;
+      }
+      .process-card__name {
+        margin: 0;
+        font: 600 var(--boxel-font);
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+      }
+      .process-card__bar {
+        height: 6px;
+        border-radius: var(--boxel-border-radius-sm);
+        background-color: var(--boxel-200);
+        overflow: hidden;
+      }
+      .process-card__fill {
+        height: 100%;
+        border-radius: inherit;
+        background-color: var(--boxel-highlight);
+        transition: width 0.3s ease;
+      }
+      .process-card__footer {
+        display: flex;
+        justify-content: space-between;
+        gap: var(--boxel-sp-xs);
+        color: var(--boxel-450);
+        font: 500 var(--boxel-font-xs);
+        letter-spacing: var(--boxel-lsp-sm);
+      }
+      .process-card__status {
+        text-transform: capitalize;
+      }
+    </style>
+  </template>
+}
+
+export class ProcessCard extends CardDef {
+  static displayName = 'Process Card';
+  static icon = ProgressIcon;
+
+  // The name shown on the setup bar; falls back to the card title.
+  @field listingName = contains(StringField);
+  // Human-readable label for the current step, e.g. "Importing files".
+  @field stage = contains(StringField);
+  // Progress as a fraction: `progressDone` of `progressTotal` items.
+  @field progressDone = contains(NumberField);
+  @field progressTotal = contains(NumberField);
+  // When the process began; drives the Home tab's ETA estimate.
+  @field startedAt = contains(DateTimeField);
+  // Lifecycle status; the Workspace treats a missing value as 'running'.
+  @field processStatus = contains(StringField);
+  // Optional themed survey collected as part of setup (e.g. a questionnaire
+  // whose answers gate later steps). Linked rather than contained so it can
+  // be its own themed card.
+  @field setupSurvey = linksTo(() => CardDef);
+
+  @field cardTitle = contains(StringField, {
+    computeVia: function (this: ProcessCard) {
+      return this.listingName ?? 'Setup process';
+    },
+  });
+
+  // Percentage complete, clamped to 0–100. Returns 0 when the total is
+  // unknown so the bar renders empty rather than NaN-wide.
+  get percentComplete(): number {
+    let total = this.progressTotal ?? 0;
+    if (total <= 0) {
+      return 0;
+    }
+    let pct = Math.round(((this.progressDone ?? 0) / total) * 100);
+    return Math.max(0, Math.min(100, pct));
+  }
+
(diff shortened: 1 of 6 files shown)


packages/base/process-card.gts
```diff
@@ -0,0 +1,159 @@
...
+        display: flex;
+        flex-direction: column;
+        gap: var(--boxel-sp-xs);
+        padding: var(--boxel-sp);
+        container-type: inline-size;
+      }
+      .process-card__header {
+        display: flex;
+        flex-direction: column;
+        gap: var(--boxel-sp-xxxs);
```
burieberry: 3xs

packages/base/process-card.gts
```diff
@@ -0,0 +1,159 @@
...
+        min-width: 0;
+      }
+      .process-card__stage {
+        color: var(--boxel-450);
```
burieberry: --muted-foreground

packages/base/process-card.gts
```diff
@@ -0,0 +1,159 @@
...
+        font: 500 var(--boxel-font-xs);
```
burieberry: instead of font shorthand, use font-weight, font-size, line-height (font-family is inherited). only use font shorthand if you don't want font-family to be themed

packages/base/process-card.gts
```diff
@@ -0,0 +1,159 @@
...
+      .process-card__bar {
+        height: 6px;
+        border-radius: var(--boxel-border-radius-sm);
+        background-color: var(--boxel-200);
+        overflow: hidden;
+      }
+      .process-card__fill {
+        height: 100%;
+        border-radius: inherit;
+        background-color: var(--boxel-highlight);
```
burieberry: --primary

packages/base/process-card.gts
```diff
@@ -0,0 +1,159 @@
...
+        transition: width 0.3s ease;
+      }
+      .process-card__footer {
+        display: flex;
+        justify-content: space-between;
+        gap: var(--boxel-sp-xs);
+        color: var(--boxel-450);
```
burieberry: --muted-foreground

packages/base/process-card.gts
```diff
@@ -0,0 +1,159 @@
...
+        margin: 0;
+        font: 600 var(--boxel-font);
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+      }
+      .process-card__bar {
+        height: 6px;
+        border-radius: var(--boxel-border-radius-sm);
+        background-color: var(--boxel-200);
```
burieberry: maybe --muted

burieberry (commented):
There's a ProgressBar component that already has the theming in place that might be sufficient. I used it in `software-factory/realm/overview.gts` for software-factory setup progress.

lukemelia:
[Claude Code 🤖] Thanks for the review — all applied in the latest two commits:

- **ProgressBar** — swapped the hand-rolled track+fill `<div>` for `<ProgressBar @value={{@model.percentComplete}} @max={{100}} />`. That also resolves the bar-color notes for free, since ProgressBar themes its track with `--muted` and fill with `--primary` internally. Dropped the now-unused `barStyle` getter.
- **`font` shorthand** — replaced every `font:` usage (`.process-card__stage`, `__name`, `__footer`) with separate `font-weight` / `font-size` (`--boxel-font-size-*`) / `line-height` (`--boxel-line-height-*`), leaving `font-family` to inherit.
- **`--muted-foreground`** — applied to the stage and footer text (was `--boxel-450`).
- **`3xs`** — header gap is now `--boxel-sp-3xs` (was the deprecated `--boxel-sp-xxxs`).

The rendering tests now assert the shared progress bar renders (`[data-test-boxel-progress-bar]`); `percentComplete`'s math is already covered by its own unit tests. Ran locally — 10/10 pass.

I also captured the two conventions that weren't yet written down — no `font` shorthand, and reach for an existing themed boxel-ui component before hand-rolling — in the `gts-component-conventions` skill, so the next card gets it right up front.

packages/base/process-card.gts
```diff
@@ -0,0 +1,140 @@
...
+class ProcessTemplate extends Component<typeof ProcessCard> {
+  <template>
+    <article class='process-card'>
+      <header class='process-card__header'>
+        <span class='process-card__stage'>
+          {{if @model.stage @model.stage 'In progress'}}
+        </span>
+        <h3 class='process-card__name'><@fields.cardTitle /></h3>
+      </header>
+      <ProgressBar @value={{@model.percentComplete}} @max={{100}} />
```
habdelra, line 30: [Claude Code 🤖] **The progress bar lost its progressbar ARIA semantics in the switch to the shared component — non-blocking, but the PR body and a test name now over-claim accessibility.**

**What the shared component actually renders.** `ProgressBar` (`packages/boxel-ui/addon/src/components/progress-bar/index.gts`) is a plain nest of `<div>`s: the outer element only takes an optional `aria-label`, and there is no `role`, `aria-valuenow`, `aria-valuemin`, or `aria-valuemax` anywhere in its template. This call site doesn't pass `@label` either, so to a screen reader the bar is an unlabeled, roleless graphic.

**Why this is a regression across the PR's own commits.** The first commit's hand-rolled bar carried `role='progressbar'` + `aria-valuenow={{@model.percentComplete}}` + `aria-valuemin='0'` / `aria-valuemax='100'`, and the test asserted all four. Moving to the shared `ProgressBar` (the right call for theming — it's the reviewer's suggestion and it fixes the token/color notes for free) silently dropped those semantics. Two stale claims are left behind:
- the PR description still advertises "an accessible `role="progressbar"` bar";
- the rendering test is titled *"renders the progress bar with accessible attributes and copy"* but now only asserts `[data-test-boxel-progress-bar]` exists — no accessible attribute is checked.

**The way out.** The clean fix is to add `role="progressbar"` + `aria-valuenow/min/max` (and an `aria-label`) to the shared `ProgressBar` itself — that restores the semantics here *and* for every other consumer, and is a small, self-contained follow-up rather than this PR's job. At minimum in this PR, correct the description and the test title so they don't claim accessibility the card no longer has. (Passing `@label={{@model.progressLabel}}` would give it an aria-label, but `ProgressBar` also renders `@label` *visibly* inside the track, which would duplicate the footer count — so that's not the right lever.)

Class: regression (a11y) introduced within this PR; non-blocking. Fixing the stale claims is in-scope; the shared-component ARIA is a follow-up.

---

packages/base/process-card.gts
```diff
@@ -0,0 +1,140 @@
...
+  // Optional themed survey collected as part of setup (e.g. a questionnaire
+  // whose answers gate later steps). Linked rather than contained so it can
+  // be its own themed card.
+  @field setupSurvey = linksTo(() => CardDef);
+
+  @field cardTitle = contains(StringField, {
+    computeVia: function (this: ProcessCard) {
+      return this.listingName ?? 'Setup process';
+    },
+  });
```
habdelra, line 112: [Claude Code 🤖] **Confirmation with one thing to make a conscious choice: this override drops `cardInfo.name` as a title source — which is consistent with some base cards but not others.**

Overriding `cardTitle` with a `computeVia` is a well-established idiom in the base realm, so the mechanism is fine. But there are two camps in how the existing overrides treat the user-editable `cardInfo.name`:

- **Replace it outright with a domain field** — `Tag` returns `this.name`, `Country` returns `this.name`, `CardsGrid` returns `this.realmName`. This override matches that camp (`this.listingName ?? 'Setup process'`).
- **Prefer `cardInfo.name`, then a domain field, then a default** — `StyleReference`: `this.cardInfo?.name ?? this.styleName ?? 'Untitled Style'`.

The consequence of the first camp here: if a process instance is given a name through the standard `cardInfo.name` mechanism but no `listingName`, its title everywhere the platform shows `cardTitle` (chooser, stack header, …) is the literal `'Setup process'`, never the name the user set. That's fine *if* these jobs are always machine-created with `listingName` populated — which is likely, given they're discovered by query — so I'm flagging it as a choice to make deliberately rather than a defect. If you want a user rename to win, the `StyleReference` shape (`this.listingName ?? this.cardInfo?.name?.trim() || 'Setup process'`) gets you there.

Related doc nit: the `listingName` comment says "falls back to the card title," but `listingName` is a plain `StringField` with no fallback of its own, and `cardTitle` in turn falls back to `listingName`. The fallback lives in the (forthcoming) consumer, not on this field — worth rewording so the two comments don't read as circular.

Class: confirmation + optional refinement; non-blocking, low priority.

---

packages/base/process-card.gts
```diff
@@ -0,0 +1,140 @@
...
+    return `${this.progressDone ?? 0} of ${total} items`;
+  }
+
+  get statusLabel(): string {
+    return this.processStatus ?? 'running';
+  }
+
+  static isolated = ProcessTemplate;
+  static embedded = ProcessTemplate;
+  static fitted = ProcessTemplate;
```
habdelra, line 140: [Claude Code 🤖] **One template for all three formats — worth a visual check in the smallest `fitted` sizes.**

`isolated`, `embedded`, and `fitted` all point at `ProcessTemplate`. That's a fine simplification for isolated/embedded, but the `fitted` format is rendered across a wide range of small aspect ratios (the grid tiles and strip/badge variants that `DefaultFittedTemplate` normally adapts to). This template is a fixed `article` → `header` (stage + title) + `ProgressBar` (1.5em tall) + `footer`, wrapped in `padding: var(--boxel-sp)`. In the shortest fitted strips that stack will overflow its container.

Because the card is folded away as "System" machinery it'll rarely be shown fitted, so this is low priority — but it can still surface in the card chooser's fitted tiles, so please eyeball it there, or give `fitted` a compact variant if it clips. Not blocking.

Class: design / visual-verification; non-blocking, low priority.

---

packages/base/process-card.gts
```diff
@@ -0,0 +1,140 @@
...
+    return Math.max(0, Math.min(100, pct));
+  }
+
+  // "3 of 12 items", or '' when the total is unknown.
+  get progressLabel(): string {
+    let total = this.progressTotal;
+    if (total == null) {
+      return '';
+    }
+    return `${this.progressDone ?? 0} of ${total} items`;
```
habdelra, line 131: [Claude Code 🤖] **"items" is hardcoded, but the card documents itself as covering more than item-counted work.** The class comment describes this as a job that may be "indexing, import, a guided setup flow" — for a guided flow the natural unit is steps, for an import it may be files. `progressLabel` always renders `"N of M items"`, with no way for a consumer to override the noun. Since the field set is explicitly a contract other cards read, consider either making the unit a field (e.g. `progressUnit` defaulting to "items") or dropping the noun ("N of M") so the label stays honest across job types. Non-blocking, but cheaper to settle now than after consumers depend on the "items" wording.

Class: follow-up / contract wording; non-blocking.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with two lenses: (1) does a base card that exists purely as a *contract* for a not-yet-landed consumer hold up on its own, and (2) did the design-review swap to the shared `ProgressBar` land cleanly. I checked out the head branch and traced the fields, the template, the shared `ProgressBar`, the base `CardDef.cardTitle` it overrides, and the failing CI check.

**Bottom line: no blocking issues. The card is small, well-tested, and idiomatic; the design-review feedback was correctly applied. The items worth acting on before merge are the stale accessibility claims and confirming the memory-baseline CI failure is the expected "new module, no baseline" case.**

**What lands right**
- The `ProgressBar` swap is the correct response to the review thread — it's the shared themed component, and it resolves the `--muted`/`--primary` token notes for free because `ProgressBar` themes its own track/fill internally. The unused `barStyle` getter was correctly dropped.
- The getters are the right shape and are unit-tested directly rather than through the DOM: `percentComplete` clamps to 0–100 and returns 0 (not `NaN`) when the total is unknown or ≤ 0; `progressLabel` returns `''` on unknown total; `statusLabel` defaults to `'running'`. `@max={{100}}` against a value already normalized to a percentage is correct (double-clamping in `ProgressBar` is harmless).
- Module wiring is sound: base modules resolve by path (`@cardstack/base/process-card`), so no manifest/registration entry is needed beyond the file itself; the `base-realm.ts` test-helper addition correctly mirrors the `CardsGrid` pattern.
- The `gts-component-conventions` skill additions (no `font` shorthand; reach for an existing themed boxel-ui component before hand-rolling; steer off the numbered palette toward semantic role tokens) are accurate and written evergreen.

**On the earlier review thread**
burieberry's `ProgressBar` suggestion is fully applied and the six styling threads are all addressed in the latest commits — nothing left open there.

**Cross-cutting: the failing "Host Memory Baseline" check is this PR's new test module.** The job log shows it failing with *"2 module(s) exceeded hard memory threshold"*, and the two flagged modules are the two with no baseline row — one of which is `Integration | Card | process-card` at 5.9 MB (the other, an `ai-assistant-panel` module, isn't from this branch). A new host test module has no baseline entry, so it's measured against an absolute hard cap. 5.9 MB is not an obvious leak for a `setupBaseRealm` test (peers on the same run: `read-source` 5.7 MB, `rendered-markdown` 11.3 MB), so this reads as "new module, no baseline" rather than retention gone wrong — and the PR's `mergeable_state` is `unstable`, not `blocked`, so it's advisory rather than a required gate. Please confirm the intended flow: the `Update Host Memory Baseline` job was skipped on this PR (it runs on main), so the baseline should regenerate post-merge. If you want to be sure it's not genuine retention, a quick pass with the host-test-memory-leak lens on the two `renderCard` calls would settle it.

**One thing this PR can't verify, by design.** There is no consumer in the tree yet — no "System (n)" folding logic and nothing querying for `ProcessCard` — so the field-name spellings are the entire load-bearing surface and nothing pins them to what the forthcoming Workspace index card will read. That's acceptable for landing the card ahead of its consumer; just worth an integration test *when the consumer lands* that actually resolves `ProcessCard` via the `codeRef` and reads these fields, so a later rename of e.g. `progressTotal` fails loudly instead of silently.

**Recommendations** (all non-blocking; ordered by value)
1. Fix the stale accessibility claims — the PR body still says "an accessible `role="progressbar"` bar" and the rendering test is titled "…accessible attributes…", but the shared `ProgressBar` has no progressbar role/`aria-value*`; adding those to the shared component is a clean follow-up. See the thread on the `<ProgressBar>` line.
2. Confirm the memory-baseline flow above (advisory vs. required; baseline regenerates on merge).
3. Decide consciously whether `cardTitle` should drop `cardInfo.name` — consistent with `Tag`/`Country`/`CardsGrid`, differs from `StyleReference`. See the thread on the `cardTitle` field.
4. Consider making the progress unit ("items") customizable or noun-free, since the card spans indexing/import/setup-flow jobs. See the thread on `progressLabel`.
5. Give `fitted` a compact variant only if it clips in the smallest tiles. See the thread on the format assignments.

**Adjacent, out of scope (minor, whoever touches this next)**
- `.process-card` declares `container-type: inline-size` but there are no `@container` queries, so it's inert (it does establish a query container + inline-size containment, which is harmless here).
- `data-status={{@model.statusLabel}}` on the status span is keyed by nothing — no CSS or test reads it. Fine as a deliberate styling/testing hook, otherwise removable.
- `startedAt` and `setupSurvey` are unused in this card's own template — expected, since they're contract fields for the forthcoming consumer, not dead code.

---

habdelra (approved):

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Applied — this gap uses the `--boxel-sp-3xs` token.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Applied — this text now uses `--muted-foreground`.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Applied — swapped the `font` shorthand for separate `font-weight` / `font-size` / `line-height` so the font-family stays inherited/themed.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Handled by adopting the shared `ProgressBar`, whose fill defaults to `--primary` internally.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Applied — now `--muted-foreground`.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Handled via the shared `ProgressBar` — its track background defaults to `--muted`.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Fixed centrally rather than at this call site: the shared `ProgressBar` now carries the progressbar ARIA (`role="progressbar"`, `aria-valuenow`/`valuemin`/`valuemax`/`valuetext`, plus an accessible-name fallback) in #5586 — your recommended "fix in the shared component" path, so every consumer benefits. On this PR I've corrected the two stale claims: the rendering test is retitled "renders the progress bar and copy" (it no longer implies it asserts ARIA — that's covered by #5586's own test, and asserting it here would create a merge-order dependency between the two PRs), and the PR body now attributes the progressbar semantics to the shared component / #5586 instead of this card.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Keeping the replace-outright shape (`listingName ?? 'Setup process'`) deliberately — it matches the sibling default-index card `CardsGrid` (and `Tag` / `Country`). These `Process` instances are machine-created with `listingName` populated, so a user-set `cardInfo.name` rename isn't an expected path here.

Also fixed the related circular doc nit you flagged: the `listingName` comment now reads "The name shown on the setup bar. `cardTitle` reads this, defaulting to 'Setup process' when it is unset."

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Keeping the "items" wording for now. The label is display-only and isn't part of the read contract — consumers read the `progressDone` / `progressTotal` fields, not this string — so a later job type that wants a different noun (steps, files) can be handled when its consumer lands, without a contract change here.

packages/base/process-card.gts
lukemelia: [Claude Code 🤖] Deferring as a low-priority follow-up. The card is folded away as "System" machinery, so it's rarely shown fitted; if it clips in the card chooser's smallest fitted tiles, a compact `fitted` variant can be added then. Not doing the visual pass in this PR.

PR #5581, opened 2026-07-22
Add RemixCard base card for realm remixes

lukemelia:
## What

Adds `RemixCard` (`packages/base/remix-card.gts`), a base CardDef for a realm or card that was remixed (cloned) from a source.

## Why

A remix runs through the same setup-progress lifecycle as any other setup job, and the Workspace index card's Home setup-bar discovers jobs by querying the realm for both `ProcessCard` and `RemixCard` instances. So `RemixCard` reuses that job-progress contract by extending `ProcessCard`, and adds the source it was remixed from. Its `displayName` (`'Remix'`) matches the fixed set the Workspace uses to fold machinery types behind a "System" link.

## Design

- Extends `ProcessCard` rather than re-declaring the shared `listingName` / `stage` / `progressDone` / `progressTotal` / `startedAt` / `processStatus` / `setupSurvey` contract — a remix *is* a specialized process. The shared progress template and getters are inherited unchanged; `cardTitle` falls back to "Remix".
- Adds `remixedFrom` (`linksTo` a `CardDef`) for the source it was cloned from. Richer remix-specific fields and presentation are deferred to the Activity-feed remix work.

## Stacking

This extends `ProcessCard`, so it's based on the `cs-12276-land-processcard-base-carddef` branch and will retarget to `main` once that merges.

## Tests

- `packages/host/tests/integration/components/remix-card-test.gts` — subtype relationship + `displayName`, inherited job-progress getters, rendering the inherited progress bar, and the title fallback.
- `packages/host/tests/helpers/base-realm.ts` — exports `RemixCard` from the base-realm test helper.

Test plan: ran locally against a full dev stack — process-card (10) + remix-card (4) = **14 passed, 0 failed**. Clean under glint (`ember-tsc --noEmit`), `ember-template-lint`, and prettier.

### DIFF
--- a/packages/base/remix-card.gts
+++ b/packages/base/remix-card.gts
@@ -21,5 +21,3 @@ export class RemixCard extends ProcessCard {
     },
   });
 }
-
-export default RemixCard;
--- a/packages/base/remix-card.gts
+++ b/packages/base/remix-card.gts
@@ -1,0 +1,25 @@
+import { CardDef, StringField, contains, field, linksTo } from './card-api';
+import { ProcessCard } from './process-card';
+
+import GitForkIcon from '@cardstack/boxel-icons/git-fork';
+
+// A remix: a realm (or card) cloned/forked from a source. A remix is a
+// specialized setup process — it runs through the same progress lifecycle —
+// so it extends ProcessCard and the Workspace surfaces it in the same Home
+// setup-bar job list (which queries for both types). It adds the source it
+// was remixed from; the shared progress template is inherited unchanged.
+export class RemixCard extends ProcessCard {
+  static displayName = 'Remix';
+  static icon = GitForkIcon;
+
+  // The card or realm this space was remixed (cloned) from.
+  @field remixedFrom = linksTo(() => CardDef);
+
+  @field cardTitle = contains(StringField, {
+    computeVia: function (this: RemixCard) {
+      return this.listingName ?? 'Remix';
+    },
+  });
+}
+
+export default RemixCard;
--- a/packages/host/tests/integration/components/remix-card-test.gts
+++ b/packages/host/tests/integration/components/remix-card-test.gts
@@ -4,6 +4,7 @@ import { module, test } from 'qunit';
 import type { Loader } from '@cardstack/runtime-common';
 
 import {
+  CardDef,
   ProcessCard,
   RemixCard,
   setupBaseRealm,
@@ -38,6 +39,16 @@ module('Integration | Card | remix-card', function (hooks) {
     assert.strictEqual(card.percentComplete, 50);
     assert.strictEqual(card.progressLabel, '4 of 8 items');
     assert.strictEqual(card.statusLabel, 'running');
+  });
+
+  test('links to the card it was remixed from', function (assert) {
+    let source = new CardDef({});
+    let card = new RemixCard({ remixedFrom: source });
+    assert.strictEqual(
+      card.remixedFrom,
+      source,
+      'remixedFrom holds the source the remix was cloned from',
+    );
   });
 
   module('rendering', function () {
--- a/packages/host/tests/integration/components/remix-card-test.gts
+++ b/packages/host/tests/integration/components/remix-card-test.gts
@@ -55,10 +55,8 @@ module('Integration | Card | remix-card', function (hooks) {
       assert.dom('.process-card__stage').hasText('Cloning cards');
       assert.dom('.process-card__count').hasText('4 of 8 items');
       assert
-        .dom('.process-card__bar')
-        .hasAttribute('role', 'progressbar')
-        .hasAttribute('aria-valuenow', '50');
-      assert.dom('.process-card__fill').hasAttribute('style', 'width: 50%');
+        .dom('[data-test-boxel-progress-bar]')
+        .exists('renders the inherited shared progress bar');
     });
 
     test('title falls back to "Remix" when listingName is unset', async function (assert) {
--- a/packages/host/tests/integration/components/remix-card-test.gts
+++ b/packages/host/tests/integration/components/remix-card-test.gts
@@ -1,0 +1,72 @@
+import { getService } from '@universal-ember/test-support';
+import { module, test } from 'qunit';
+
+import type { Loader } from '@cardstack/runtime-common';
+
+import {
+  ProcessCard,
+  RemixCard,
+  setupBaseRealm,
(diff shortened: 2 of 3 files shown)


packages/base/remix-card.gts
```diff
@@ -0,0 +1,25 @@
...
+  static icon = GitForkIcon;
+
+  // The card or realm this space was remixed (cloned) from.
+  @field remixedFrom = linksTo(() => CardDef);
+
+  @field cardTitle = contains(StringField, {
+    computeVia: function (this: RemixCard) {
+      return this.listingName ?? 'Remix';
+    },
+  });
```
habdelra, line 22: [Claude Code 🤖] **Confirmation + guard-rail (non-blocking).** Overriding `cardTitle` is the right way to change the fallback title — there is no clean way to reuse the parent's `computeVia` and swap only the default string, so redefining the field is the idiomatic mechanism. The render test ("title falls back to 'Remix'") pins that this override actually takes effect, so I'm confident it works rather than silently keeping the parent's `'Setup process'`.

The one thing to watch: this copies ProcessCard's entire `computeVia` body just to change `'Setup process'` → `'Remix'`. If ProcessCard's `cardTitle` ever grows real logic (trimming, reading a different field), this copy won't inherit it and the two will drift with nothing failing loudly. Nothing to change now — worth a shared helper only if that computation gets more complex.

---

packages/base/remix-card.gts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export default RemixCard;
```
habdelra: [Claude Code 🤖] **Minor consistency (optional, non-blocking).** The parent `ProcessCard` and other base cards such as `skill.gts` don't declare a default export, and nothing in this diff imports `RemixCard` as a default (the test helper and `base-realm.ts` both use the named export). The `export default` is harmless — it just makes this card inconsistent with its immediate sibling. Drop it or keep it; flagging only so the choice is deliberate.

---

packages/host/tests/integration/components/remix-card-test.gts
habdelra, line 1: [Claude Code 🤖] **Test coverage (non-blocking; a follow-up is fine).** `remixedFrom` is the only field `RemixCard` adds on top of `ProcessCard`, but no test sets or reads it. The suite covers the subtype relationship, the inherited progress getters, rendering, and the title fallback — all of which are ProcessCard behavior that its own test already exercises. So what makes this card a *remix* is currently unpinned.

A single assertion that constructs a `RemixCard` with a `remixedFrom` link and reads it back would close the gap. This mirrors the sibling `process-card-test.gts`, which also skips its linked `setupSurvey` field — so it's not a regression in the established pattern, just the one gap most worth closing here since `remixedFrom` is the card's reason to exist.

---

habdelra (commented):
[Claude Code 🤖] **Lens:** whether the `RemixCard` ↔ `ProcessCard` subtype relationship is faithful, whether the inherited progress contract is reused unchanged, and whether the tests pin what the card actually *adds*. I reviewed the head commit against the ProcessCard base branch and read the parent card, the `base-realm.ts` helper, the boxel-icons export map, and the index query-engine's `type`-filter semantics that the "query for both types" design leans on.

**Bottom line: no blocking issues.** This is a small, faithful extension. `RemixCard` reuses the ProcessCard progress contract unchanged, adds exactly one field (`remixedFrom`) plus a `'Remix'` title fallback, and registers itself in the base-realm test helper in the same three places ProcessCard did. Verified concretely: the `git-fork` icon import resolves through the boxel-icons `./*` wildcard export (`src/icons/git-fork.gts` exists), the `cardTitle` override is pinned by the render test, and the commit touches exactly the same file set as the ProcessCard commit — so there's no missed registration spot. All my notes are minor/optional.

**What lands right:**
- Extending `ProcessCard` instead of re-declaring the shared `listingName` / `stage` / `progress*` / `startedAt` / `processStatus` / `setupSurvey` contract is the right call — the progress getters, template, and field names stay in one home, and a remix genuinely *is* a specialized process.
- The comment on the class is evergreen and states the current contract without journey or tracker references. Good.

**One design point to confirm on the Workspace side — out of scope for this PR, since that code isn't in this branch or on the default branch yet.** The description says the setup-bar "queries the realm for both `ProcessCard` and `RemixCard` instances." Per the index query engine, a `type` filter matches an instance's whole adoptsFrom chain and results collapse by `GROUP BY url` — so a `type: ProcessCard` query *already* returns every `RemixCard` instance, which would make the second branch redundant (harmless, just a no-op). That redundancy disappears only if the Workspace matches on the exact `_cardType` synthetic key rather than a subtype `type` filter — in which case the `displayName = 'Remix'` string is what actually carries the "fold behind System" behavior. Worth confirming whichever way that query is written, because the two mechanisms distinguish the types very differently. Neither changes anything in this diff.

1. Add a test that sets and reads `remixedFrom` — it's the only field this card adds beyond ProcessCard and nothing exercises it (see the comment on the test file).
2. Keep the `cardTitle` fallback in sync with ProcessCard's, or extract a shared helper if that computation ever grows (see the inline note on the field).
3. Optional: drop the `export default` for consistency with `ProcessCard` (see the inline note).

---

habdelra (approved):

packages/host/tests/integration/components/remix-card-test.gts
lukemelia: [Claude Code 🤖] Added in 14d1ae15ce. New test links a `RemixCard` to a source card and reads `remixedFrom` back, pinning the one field that distinguishes this card from `ProcessCard`. Mirrors the constructor-set / synchronous read-back idiom used elsewhere (e.g. `computed-test.gts` reading `this.author.bestFriend`).

packages/base/remix-card.gts
lukemelia: [Claude Code 🤖] Dropped in 14d1ae15ce, for consistency with `ProcessCard` / `skill.gts`. Confirmed the sole importer (`base-realm.ts`) uses `import type * as RemixCardModule` + the named export, so nothing depended on the default.

packages/base/remix-card.gts
lukemelia: [Claude Code 🤖] Leaving as-is, per your recommendation — overriding `cardTitle` to swap only the fallback string is the idiomatic mechanism, and the render test pins that it takes effect. Worth a shared helper only if `ProcessCard.cardTitle` ever grows real logic; noted for that day.

PR #5582, opened 2026-07-22
Add Workspace base card (with test coverage) as the realm index card

lukemelia:
## What

Adds `Workspace` (`packages/base/workspace.gts`), a base CardDef built to replace `CardsGrid` as a realm's index card, together with its tests.

**This PR lands the card and its tests only — it does not change which index card any realm uses.** No realm-creation code or `index.json` is touched here, so new realms still get a `CardsGrid` index and existing realms are unchanged. Making Workspace the default for newly-created realms (`create-realm.ts`) and migrating existing realms are separate changes (see Scope).

The card bundles the existing CardsGrid experience under a **Library** tab and adds:

- **Home** — setup-progress bar, pinned entry-point "door" tiles, collapsible README hero, browse chips by card/file type, and a live space-details strip (visibility, counts, published sites, route count).
- **Activity** — a reverse-chron event feed with day headers, reveal-on-scroll (20 at a time, capped at 100), loaded lazily when the tab opens.
- **Frame** — a Cmd+K typeahead search, realm-scoped.
- **Edit** — a settings format including realm publish/republish.

## How it wires into existing infrastructure

- **Library** delegates to the existing `CardsGridLayout` (reusing its `VIEW_OPTIONS`/`SORT_OPTIONS`/`FilterOption`), so the current grid behavior is reachable unchanged.
- The `readme` field uses `MarkdownDef`; realm facts come from `RealmInfo` and the realm `_types` endpoint; realm change notifications come from `subscribeToRealm`.
- Publish wraps the host `publish-realm` tool, invoked with a typed `PublishRealmInput` instance.
- Setup-progress and remix jobs are discovered by querying the realm for `ProcessCard` / `RemixCard` via `codeRef` (not static imports), so this card compiles on its own and the Home job list populates once those base cards exist.

## Port notes

Ported from the proof-of-concept: dropped the edit-tracking header and the superscript annotation markers, removed opaque private design-doc references, converted absolute base-realm import URLs to relative, removed the unused first `VIEW_OPTIONS` binding, and replaced the `as any` casts with real types. The current-module URL uses the same `import.meta` + `@ts-ignore` idiom as the rest of `packages/base`.

## Tests

- **Unit** — `etaMinutes` (exported, with an injectable `now`) projection + every early-return + the >30-minute suppression; `classifyActivityVerb` across the created-window boundary and missing timestamps.
- **Integration** — segment switching across the Home/Library/Activity tabs and presence of the Frame search input; plus a render smoke test for the isolated and edit formats.
- **Acceptance** — a realm whose `index.json` adopts Workspace renders its shell in operator mode and switches segments, exercising the live realm-backed paths (`subscribeToRealm`, the `_types` fetch, and the Activity feed).

## Scope — what makes Workspace the actual replacement (separate changes)

- **Default index for new realms** — `create-realm.ts` currently hardcodes the `CardsGrid` index card; that must switch to Workspace.
- **Migrate existing realms** — flip committed realm `index.json` files off CardsGrid.
- **Realm-server recognition** — teach the publish handler + prerender fast-path to recognize Workspace adoption.
- Plus accessibility/i18n, magic-number configurability, missing-`RealmConfig` fallback, and publish/unpublish test coverage.

## Test plan

glint (`ember-tsc --noEmit`) clean over the full port; `ember-template-lint` and prettier clean. Run locally against a full dev stack, per module: unit **9 passed**, integration segments **2 passed**, render smoke **2 passed**, acceptance **1 passed** — 0 failed.


packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+      this.filterOptions.find(
+        (filter) => filter.displayName === this.activeFilter.displayName,
+      ) ?? this.filterOptions[0];
+  });
+
+  private loadJobs = restartableTask(async () => {
+    if (!this.args.context?.store) {
+      return;
+    }
+    let processRef = codeRef(here, './process-card', 'ProcessCard');
```
lukemelia, line 3147: [Claude Code 🤖] **The setup-progress feature is dead — these type refs resolve to nonexistent base-realm modules.**

`codeRef(here, './process-card', 'ProcessCard')` and `codeRef(here, './remix-card', 'RemixCard')` resolve `here` (the base-realm module URL) against `./process-card` / `./remix-card`, i.e. `<baseRealm>/process-card` and `<baseRealm>/remix-card`. Neither module exists in `packages/base` — `ProcessCard`/`RemixCard` live only as host test fixtures / in other realms.

So `loadJobs`'s query `{ any: [{ type: processRef }, { type: remixRef }] }` never matches anything, and the whole setup-progress UI is inert: the "Setting up …" strip on Home, the collapsing progress dock on Activity, and the Activity-tab attention dot never render even while a real setup job runs in the realm.

Worse case: if the realm-server *errors* on an unresolvable type ref (rather than returning empty), `loadFilterList`/`refreshOnIndex` re-fire `loadJobs` on every incremental index and each throws. Please confirm the intended module location for these two card types.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+      }
+      let ref = codeRefFromInternalKey(summary.id);
+      if (!ref) {
+        return;
+      }
+      let kind = summary.attributes.kind ?? 'instance';
+      if (kind === 'file') {
+        fileTotal += summary.attributes.total ?? 0;
+      } else {
+        cardTotal += summary.attributes.total ?? 0;
```
lukemelia: [Claude Code 🤖] **Rail counts are inflated because totals accumulate before the exclusion guards.**

`cardTotal += summary.attributes.total` (and the `fileTotal` sibling) runs *before* the `excludedTypeIds.includes(summary.id)` and `summary.id.endsWith('workspace/Workspace')` guards a few lines down. So the Workspace index card itself (always present — it's the very card being rendered) plus any `Cards Grid` / `CardDef` instances get added into `cardTotal`.

But `countFor` returns `cardTotal` for the "Cards" row and `cardTotal + fileTotal` for "Everything", while those rows' queries exclude `_cardType: 'Workspace'` and `_cardType: 'Cards Grid'`. Net result: the sidebar shows e.g. "Everything (6)" / "Cards (6)" but the grid those rows open lists only 5 — off by at least one (the Workspace card counting itself) on every workspace. Move the `+=` accumulation after the exclusion guards.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+        <div class='library'>
+          <nav class='rail scroll-container' aria-label='Library filters'>
+            <div class='rail-group'>
+              <h3 class='rail-label'>Library</h3>
+              {{#each this.libraryFilters as |option|}}
+                <button
+                  type='button'
+                  class='rail-row
+                    {{if
+                      (eq option.displayName this.activeFilter.displayName)
```
lukemelia: [Claude Code 🤖] **Rail selection + `countFor` key on `displayName`, which collides with user type names.**

Selected styling here uses `(eq option.displayName this.activeFilter.displayName)` (also at the card/file-type rows), and `countFor` switches on `option.displayName` — but `selectFilter` sets `activeFilter` by object identity. If a realm defines a user card/file type whose displayName happens to be `Cards`, `Files`, or `Everything`:

- both the library-group row and the user-type row highlight together when either is clicked, and
- `countFor` returns the aggregate `cardTotal`/`fileTotal` for the user type's row instead of that type's own count.

Key the selected check and `countFor` on the filter identity (or a stable id) rather than `displayName`.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+      this.jobComponents.push({
+        id: card.id!,
+        card,
+        status: card.processStatus ?? 'running',
+        component: (card.constructor as typeof BaseDef).getComponent(card),
+      });
+    }
+  });
+
+  private refreshOnIndex = (ev: RealmEventContent) => {
```
lukemelia, line 3179: [Claude Code 🤖] **(Plausible) `refreshOnIndex` reacts only to `incremental` index events, so full/copy reindexes leave the UI stale.**

This guard runs `loadFilterList` / `loadJobs` / `loadLatest` / `loadFeed` only when `ev.indexType === 'incremental'`. After a full reindex or a realm copy/remix — which emit index events with a non-`incremental` `indexType` — the rail type list, counts, running-jobs dock, Home "latest" preview, and Activity feed stay stale until the user navigates away and back.

Note the primary user-triggered full-reindex path may be shielded: `realm.ts` broadcasts full-reindex invalidations via `broadcastIncrementalInvalidationEvent` (indexType `incremental`), so this only bites the true `full`/`copy` event paths. Worth confirming which reindex flows actually emit non-incremental events here and widening the guard to cover them.

packages/host/tests/helpers/base-realm.ts
```diff
@@ -347,6 +351,10 @@ async function initialize() {
     await loader.import<typeof CardsGridModule>('@cardstack/base/cards-grid')
   ).CardsGrid;
 
+  Workspace = (
```
lukemelia: [Claude Code 🤖] **(Plausible) This unconditional import pulls the 3948-line `workspace.gts` graph into every `setupBaseRealm` test's `beforeEach`.**

`setupBaseRealm` registers `initialize()` as a `beforeEach`, so this `await loader.import('@cardstack/base/workspace')` now runs for every rendering/integration test that uses the helper — synchronously loading `workspace.gts` plus its transitive graph (CardsGridLayout, several boxel-icons, BoxelInput, and the `PublishRealmCommand` host tool with its `HostBaseTool`/config chain) into the loader on each test, even tests that never touch Workspace.

The host suite is already documented to OOM shards and is gated by a memory-baseline check, so this added per-test retained module graph is a plausible trigger for the baseline check tripping red or a shard OOMing. Consider importing Workspace lazily (only in the tests that need it) rather than in the shared `initialize()`.

packages/base/workspace.gts
lukemelia, line 3179: [Claude Code 🤖] **(Efficiency) `refreshOnIndex` fans out to 3–4 full realm searches on every incremental index event.**

Each incremental index event triggers `loadFilterList` (a `_types` fetch), `loadJobs`, `loadLatest`, and `loadFeed` (when Activity is open) — 3–4 independent realm round-trips per event. During active editing (which produces frequent incremental index events — e.g. an AI setup flow writing many cards) this multiplies index-event traffic 3–4×. The tasks are restartable, so this isn't a correctness bug, but consider debouncing/coalescing the refresh, or only refreshing the panels currently visible.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+
+  private get realms(): string[] {
+    return this.args.model[realmURL] ? [this.args.model[realmURL].href] : [];
+  }
+
+  private get primaryRealm(): string | undefined {
+    return this.realms[0];
+  }
+
+  private searchRealm = async (query: Query): Promise<CardDef[]> => {
```
lukemelia, line 2809: [Claude Code 🤖] **(Efficiency) `searchRealm` hard-forces `page: { size: 100 }`, so `loadLatest` hydrates up to 100 cards to read one title.**

`searchRealm` spreads `...query` then overrides `page: { size: 100 }`, ignoring any smaller page a caller passes. `loadLatest` only needs the single newest card (it sorts by `lastModified desc` and `.find()`s the first, reading just `.cardTitle`), but `store.search` hydrates up to 100 instances every time — on subscription setup and on every incremental index event. On a large realm that's 100 cards fetched and instantiated per index tick to display one line. Let callers pass their own `page` (default to 100 only when unspecified), and have `loadLatest` request `size: 1`.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+    this.segment = segment;
+    if (segment === 'activity') {
+      // Lazy: the whole-realm reverse-chron search only runs when the
+      // Activity segment is actually opened — keeps index-time prerenders
+      // (which render Home) off the expensive path.
+      this.loadFeed.perform();
+    }
+  };
+
+  get runningJobs() {
```
lukemelia, line 2277: [Claude Code 🤖] **(Efficiency) `runningJobs` is uncached and re-filters `jobComponents` on every template access.**

`get runningJobs()` runs `.filter` over `jobComponents` on each read, and the isolated template reads `this.runningJobs` / `this.runningJobs.length` at ~6 sites plus `dockSummary`. Since `jobComponents` is a `TrackedArray`, every re-render re-runs the filter multiple times. The sibling chip getters (`contentCardChips`, `fileChips`, `systemTypeCount`) are all `@cached` — mark this one `@cached` too so it recomputes only when its tracked inputs change.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+        realmURL: this.args.model[realmURL],
+      });
+    } else if (activeFilterRef) {
+      await this.args.createCard?.(activeFilterRef, undefined, {
+        realmURL: this.args.model[realmURL],
+      });
+    }
+  });
+
+  private loadFilterList = restartableTask(async () => {
```
lukemelia, line 3050: [Claude Code 🤖] **(Duplication) `loadFilterList` / `setupRealmSubscription` / `teardownRealmSubscription` / `createCard` are near-verbatim copies from `cards-grid.gts`.**

These methods are near-identical to the same methods in `packages/base/cards-grid.gts` (`loadFilterList` 331–438, `setupRealmSubscription` 245–265, `teardownRealmSubscription` 239–243 — byte-for-byte identical, `createCard` 279–329). The `_types` fetch, error-construction block, `excludedTypeIds` list, kind grouping, and realm-subscription lifecycle now live in two places. Any fix to `_types` parsing or subscription teardown (e.g. a rolling-deploy compatibility change) has to be made twice, or one card silently diverges. Consider extracting a shared mixin/util.

packages/base/workspace.gts
```diff
@@ -0,0 +1,3948 @@
...
+  // already-loaded links) instead of a query.
+  @cached
+  private get everythingFilter(): RailOption {
+    return {
+      displayName: 'Everything',
+      icon: LayoutGridIcon,
+      query: {
+        filter: {
+          every: [
+            { not: { eq: { _cardType: 'Cards Grid' } } },
```
lukemelia: [Claude Code 🤖] **(Duplication) The `Cards Grid` / `Workspace` exclusion clause pair is repeated across 7 query sites.**

This two-line `{ not: { eq: { _cardType: 'Cards Grid' } } }` / `'Workspace'` exclusion is repeated at 7 places (≈ lines 2301, 2406, 2456, 2650, 2818, 2853, 3233). If the Workspace displayName changes, or a third self-referential type needs hiding, all seven sites must change — miss one and the workspace/grid card leaks into the Library, search dropdown, Activity feed, or pin chooser. Extract a single `excludeSystemCardsFilter` (or const clause array) and spread it into each query.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Obsolete after merging main: `./process-card` and `./remix-card` now exist in `packages/base` (the ProcessCard/RemixCard cards landed on main), so `here` (the base-realm module URL) resolves the codeRefs and the setup-progress query matches. Resolving.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Already addressed: the `cardTotal`/`fileTotal` accumulation now runs after the exclusion guards (with a comment noting the counts must match the grid each row opens). Resolving.

packages/host/tests/helpers/base-realm.ts
lukemelia: [Claude Code 🤖] Already addressed: `initialize()` (the `setupBaseRealm` beforeEach) no longer imports Workspace — it moved to a dedicated `initializeWorkspace()`/`setupWorkspaceCard()` so the heavy graph loads only in tests that opt in. Resolving.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Already addressed: `searchRealm` now defaults to 100 but honors a smaller caller page, and `loadLatest` requests `page: { size: 1 }`. Resolving.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Already addressed: `runningJobs` is now `@cached`. Resolving.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Already addressed: extracted to `SELF_REFERENTIAL_CARD_TYPES` + `excludeSelfReferentialCards()`, spread into each query site. Resolving.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Fixed in 4e54093456. Filters now carry a stable `id` (library groups a fixed slug; type rows the card-type internal key that survives the loadFilterList rebuild), and rail selection, `countFor`, and the post-rebuild re-resolve all compare on `id` instead of `displayName`. `id` lives on the shared `FilterOption`. A user type named "Cards" no longer collides.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Fixed in 4e54093456. `refreshOnIndex` now reacts to `incremental`, `full`, and `copy` completions (still excluding the `incremental-index-initiation` pre-index signal), so a full reindex or realm copy/remix refreshes the rail, counts, jobs dock, and feed.

lukemelia: [Claude Code 🤖] Fixed in 4e54093456. The four panel searches now run through a debounced restartable task (`refreshAfterIndex`, 200ms), so a burst of index events during active editing coalesces into a single refresh instead of firing 3-4 searches per event.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Leaving this open as a follow-up rather than forcing an extraction now — the two implementations have meaningfully **diverged**, so they are no longer the near-verbatim copies this thread assumes:

- `loadFilterList`: the Workspace copy accumulates `cardTotal`/`fileTotal`, excludes the self-referential Workspace card, and stamps `id`/`count` on each row; the Cards Grid copy has the "All Files" root-group (`sawFileSummary`) handling and a rolling-deploy `kind`-missing compat branch. A shared version would have to be parameterized across all of those.
- `setupRealmSubscription`: identical lifecycle shell, but Workspace injects a multi-panel `refreshOnIndex` + a five-task initial load, vs Cards Grid's single `refreshFilterList` + one load.
- Only `teardownRealmSubscription` is byte-identical.

Because the shareable surface is now small and the risk is a shared abstraction over two intentionally-different behaviors touching the core Cards Grid card, this is better done as a dedicated refactor PR that can run both cards' suites. Suggest tracking it separately; leaving unresolved.

richardhjtan (commented):
[Claude Code 🤖] This pass re-derives every previously-flagged finding against the current head and re-scans the fix commits for regressions, rather than re-walking the whole 3,983-line file from zero — the prior round's inline threads already cover the file's mechanics in detail and are all marked resolved.

**Bottom line: no blocking issues.** All eight substantive findings from the earlier round are verified fixed at the current head, and the fix commits (`591d1e1f61`, `4e540934568c`) don't introduce new regressions. The one deferred finding (the `loadFilterList`/`setupRealmSubscription` overlap with `cards-grid.gts`) is correctly left open — reading both implementations confirms they've genuinely diverged, not just superficially.

**What lands right, verified against the checked-out branch:**
- `FilterOption.id` (added in `cards-grid-layout.gts`) is threaded consistently: every `RailOption` `workspace.gts` constructs (`everythingFilter`, `entryPointsFilter`, `cardsFilter`, `filesFilter`, the search pseudo-filter, and each per-type row pushed in `loadFilterList`) stamps an `id`, and all three rail-selection comparisons plus `countFor` plus the post-rebuild `activeFilter` re-resolve key off `option.id` / `filter.id` instead of `displayName`. This closes the collision a user-defined type named "Cards"/"Files"/"Everything" would have caused.
- `excludeSelfReferentialCards()` / `SELF_REFERENTIAL_CARD_TYPES` is spread at all 7 call sites, and each spread correctly preserves whether the original inline clause used `on: baseCardRef` or the bare form — verified site by site against the pre-fix diff.
- `loadFilterList` now accumulates `cardTotal`/`fileTotal` after both the `excludedTypeIds` guard and the `workspace/Workspace` self-exclusion, so the "Everything"/"Cards" rail totals match what each row's own query returns.
- `searchRealm`'s page merge is now `{ page: { size: 100 }, ...query }`, so a caller-supplied `page` (e.g. `loadLatest`'s `{ size: 1 }`) wins over the default instead of being clobbered by it; `loadLatest` no longer hydrates 100 instances to read one title.
- `refreshOnIndex` widens to `indexType` `'full'` and `'copy'` in addition to `'incremental'`, still excluding `'incremental-index-initiation'` — checked against the four `indexType` values `realm.ts` actually emits, so the branch isn't guessing at an enum. The four panel reloads now run through `refreshAfterIndex`, a 200ms-debounced `restartableTask`, coalescing an index-event burst into one refresh.
- `runningJobs` is `@cached`, matching its sibling getters.
- `base-realm.ts`'s `initialize()` no longer imports `workspace.gts`; the heavy import moved to an opt-in `initializeWorkspace()`/`setupWorkspaceCard()`, and both call sites (`workspace-test.gts`, `workspace-behavior-test.gts`) call the new hook alongside `setupBaseRealm`.

**Open discussion:** none outstanding — every existing thread already carries a specific fix-commit reference and resolution note, and each one checks out against the code at head.

**Adjacent, out of scope:** `packages/base/cards-grid.gts` (untouched by this PR) still resolves its own rail selection with `filter.displayName === this.activeFilter.displayName` (its `loadFilterList`'s post-rebuild re-resolve). That's the identical collision this PR fixed in `workspace.gts`'s copy, still live in the original it was copied from. Not this PR's bug to fix, but worth a note for whoever next touches that file's rail logic.

PR #5585, opened 2026-07-22
Recognize Workspace as a default realm index card

lukemelia:
## What

Both the publish handler and the prerender fast-path only recognized `CardsGrid` as a realm's default index card. This teaches them to recognize `Workspace` too, so Workspace-indexed realms get the same default-index handling.

## Changes

- **`handle-publish-realm.ts`** — the CardsGrid-specific `CARDS_GRID_MODULE_FORMS` + name check becomes a list of recognized default-index adoptions (CardsGrid + Workspace). Publishing a realm whose `index.json` adopts either now writes `includePrerenderedDefaultRealmIndex: true` so the published homepage gets real prerendered content.
- **`routes/render/html.ts`** — `#isDefaultRealmCardsGridIndex` → `#isDefaultRealmIndexCard`, comparing the index card's top type against a list of recognized default-index refs via `internalKeyFor`. A realm indexed by CardsGrid or Workspace (and not opted in) short-circuits the isolated render to the boilerplate placeholder.
- **`realm-index-boilerplate.ts`** — the placeholder serves both index cards now, so its cards-grid-specific class/data attributes are generalized to realm-index-neutral names (no other consumers reference them).

A realm with a bespoke (non-default) index card is still left alone in both paths.

## Why this matters for the project

Making Workspace the actual default for new realms (`create-realm.ts`) depends on this — otherwise a newly created Workspace-indexed realm would lose the prerendered-default-index fast-path on publish.

## Tests

- **Realm-server** (`publish-unpublish-realm-test.ts`) — publishing a realm whose index adopts Workspace writes `includePrerenderedDefaultRealmIndex`, mirroring the existing CardsGrid case.
- **Host** (`realm-indexing-test.gts`) — a Workspace-indexed realm that has not opted in gets the boilerplate placeholder as its isolated HTML, mirroring the existing CardsGrid case.

## Stacking

Based on the Workspace port branch — the recognition is meaningless without the card, and the tests set up Workspace-indexed realms. Retargets to `main` once the port merges. Draft until then.

## Test plan

realm-server `ember-tsc` + host glint clean; eslint + prettier clean. Host boilerplate test run locally against a full dev stack — **1 passed**. The realm-server publish test is static-verified and mirrors an existing passing test; it runs in CI (the realm-server publish harness needs test-pg + its own prerender, not wired up in this local session).

### DIFF
--- a/packages/host/app/routes/render/html.ts
+++ b/packages/host/app/routes/render/html.ts
@@ -20,14 +20,14 @@ import { getClass, getTypes } from './meta';
 import type { Model as ParentModel } from '../render';
 import type { BoxComponent, CardDef, Format } from '@cardstack/base/card-api';
 
-// Stable internal key for the base CardsGrid type. We compare against
-// the internalKeyFor representation of the cards-grid module + name so
-// the check tolerates whatever resolved form the host's identify path
-// produces (e.g. realm-aliased base URLs).
-const CARDS_GRID_REF = {
-  module: '@cardstack/base/cards-grid',
-  name: 'CardsGrid',
-} as ResolvedCodeRef;
+// Stable internal keys for the base card types recognized as a realm's
+// default index. We compare against the internalKeyFor representation of
+// each module + name so the check tolerates whatever resolved form the
+// host's identify path produces (e.g. realm-aliased base URLs).
+const DEFAULT_REALM_INDEX_REFS = [
+  { module: '@cardstack/base/cards-grid', name: 'CardsGrid' },
+  { module: '@cardstack/base/workspace', name: 'Workspace' },
+] as ResolvedCodeRef[];
 
 export interface Model {
   instance: CardDef;
@@ -107,20 +107,21 @@ export default class RenderHtmlRoute extends Route<Model> {
     let useRealmIndexBoilerplate =
       format === 'isolated' &&
       level === 0 &&
-      this.#isDefaultRealmCardsGridIndex(instance, types);
+      this.#isDefaultRealmIndexCard(instance, types);
 
     return { format, instance, Component, useRealmIndexBoilerplate };
   }
 
   // True when the card under render is the realm's default index card
-  // AND its type chain begins with the base CardsGrid AND the realm
-  // has NOT opted in to keeping its prerendered isolated HTML via
+  // AND its type chain begins with a recognized default index card
+  // (CardsGrid or Workspace) AND the realm has NOT opted in to keeping
+  // its prerendered isolated HTML via
   // `RealmInfo.includePrerenderedDefaultRealmIndex`. The orchestrator
   // substitutes a boilerplate placeholder for the captured HTML in
   // that case so the indexer doesn't pay for the (expensive) grid
   // fan-out render of every card in the realm. Published realms
   // receive the opt-in automatically from the publish handler.
-  #isDefaultRealmCardsGridIndex(instance: CardDef, types: CodeRef[]): boolean {
+  #isDefaultRealmIndexCard(instance: CardDef, types: CodeRef[]): boolean {
     let cardRealmURL = instance[realmURL];
     if (
       !cardRealmURL ||
@@ -138,8 +139,10 @@ export default class RenderHtmlRoute extends Route<Model> {
     }
     let vn = this.network.virtualNetwork;
     let topKey = internalKeyFor(topType, undefined, vn);
-    let cardsGridKey = internalKeyFor(CARDS_GRID_REF, undefined, vn);
-    if (topKey !== cardsGridKey) {
+    let isDefaultIndexType = DEFAULT_REALM_INDEX_REFS.some(
+      (ref) => internalKeyFor(ref, undefined, vn) === topKey,
+    );
+    if (!isDefaultIndexType) {
       return false;
     }
     let info = this.realm.info(cardRealmURL.href);
--- a/packages/host/app/routes/render/html.ts
+++ b/packages/host/app/routes/render/html.ts
@@ -36,7 +36,8 @@ export interface Model {
   // True when this render should be short-circuited to the realm-
   // index boilerplate placeholder instead of running through Glimmer.
   // Set only when `format === 'isolated'`, the card is the realm's
-  // default index, the type chain is base CardsGrid, and the realm
+  // default index, the type chain begins with a recognized default
+  // index card (CardsGrid or Workspace), and the realm
   // has not opted in via `includePrerenderedDefaultRealmIndex` on its
   // RealmConfig card. The orchestrator in `card-prerender.gts`
   // honours this flag by substituting the boilerplate string and
--- a/packages/host/tests/integration/realm-indexing-test.gts
+++ b/packages/host/tests/integration/realm-indexing-test.gts
@@ -5191,6 +5191,45 @@ posts/ignore-me.json
     );
   });
 
+  test('isolated HTML for a default Workspace realm-index card is replaced with the boilerplate when realm has not opted in', async function (assert) {
+    let { realm } = await setupIntegrationTestRealm({
+      mockMatrixUtils,
+      contents: {
+        'index.json': {
+          data: {
+            type: 'card',
+            attributes: {},
+            meta: {
+              adoptsFrom: {
+                module: '@cardstack/base/workspace',
+                name: 'Workspace',
+              },
+            },
+          },
+        },
+        'realm.json': {
+          data: {
+            type: 'card',
+            attributes: {},
+            meta: {
+              adoptsFrom: {
+                module: '@cardstack/base/realm-config',
+                name: 'RealmConfig',
+              },
+            },
+          },
+        },
+      },
+    });
+    let entry = await getInstance(realm, new URL(`${testRealmURL}index`));
+    assert.ok(entry, 'realm index card is indexed');
+    assert.strictEqual(
+      entry?.isolatedHtml,
+      REALM_INDEX_BOILERPLATE_HTML,
+      'isolated_html is the boilerplate constant when a Workspace-indexed realm has not opted in',
+    );
+  });
+
   test('isolated HTML for a non-CardsGrid index card is rendered normally regardless of opt-in', async function (assert) {
     class PlainHomePage extends CardDef {
       static displayName = 'Plain Home Page';
--- a/packages/realm-server/handlers/handle-publish-realm.ts
+++ b/packages/realm-server/handlers/handle-publish-realm.ts
@@ -54,11 +54,39 @@ const log = logger('handle-publish');
 // disk is the absolute URL. Listing both forms keeps the detection
 // future-proof: once `@cardstack/base/` becomes a live prefix
 // mapping, `index.json` files that use it are still caught.
-const CARDS_GRID_MODULE_FORMS = new Set<string>([
-  'https://cardstack.com/base/cards-grid',
-  '@cardstack/base/cards-grid',
-]);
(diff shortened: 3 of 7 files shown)


packages/realm-server/handlers/handle-publish-realm.ts
```diff
@@ -54,11 +54,39 @@ const log = logger('handle-publish');
...
+function isDefaultRealmIndexAdoption(adoptsFrom: {
+  module: string;
+  name: string;
+}): boolean {
+  return DEFAULT_REALM_INDEX_ADOPTIONS.some(
+    (adoption) =>
+      adoption.modules.has(adoptsFrom.module) &&
+      adoption.name === adoptsFrom.name,
+  );
+}
```
habdelra, line 89: [Claude Code 🤖] **Confirmation — the recognizer is correct and matches its twin.** Non-blocking.

`isDefaultRealmIndexAdoption` AND-checks module membership against name equality, so it won't misfire on a card *named* `Workspace` living in a different module, nor on a `CardsGrid`-named card in the workspace module. Both module spellings per card mirror the existing CardsGrid pair, and the base-realm URL for the new entry (`https://cardstack.com/base/workspace`) is the correct absolute form for `packages/base/workspace.gts`.

This is one of the two places that recognize a realm's default index card. I checked the other — `#isDefaultRealmIndexCard` in `packages/host/app/routes/render/html.ts` — and it was generalized the same way (comparing `types[0]` against `DEFAULT_REALM_INDEX_REFS` via `internalKeyFor`). The two agree in the common case that matters here: an `index.json` that directly adopts the base def produces an instance whose own type (`types[0]`) *is* that base def, so the disk-`adoptsFrom` read and the resolved-type read land on the same answer. I also confirmed there's no third recognizer downstream — the published-realm SSR path keys off the `includePrerenderedDefaultRealmIndex` flag this handler writes, not the card type — so generalizing these two is complete.

Minor, optional: the comment block just above (beginning "The CardsGrid CardDef can be referenced two equivalent ways…") still frames the two-module-forms rationale as CardsGrid-specific, even though it now applies to every entry in this list. Consider opening it with "A base CardDef can be referenced two equivalent ways…" so the note reads as general, since the generalized comment below points back to it with "see the note above."

---

packages/host/app/routes/render/html.ts
habdelra, line 1: [Claude Code 🤖] **Stale comment left behind by the generalization.** Non-blocking (doc drift).

The `Model.useRealmIndexBoilerplate` docstring near the top of this file still reads:

> Set only when `format === 'isolated'`, the card is the realm's default index, **the type chain is base CardsGrid**, and the realm has not opted in…

That sentence describes the pre-PR behavior. The change generalized the check to CardsGrid *or* Workspace and updated the sibling comments — the method's own docstring on `#isDefaultRealmIndexCard`, the comment in `packages/host/app/components/card-prerender.gts`, and the boilerplate constant's header all now say "CardsGrid or Workspace" — but this interface docstring was missed. Suggest matching it, e.g. "the type chain begins with a recognized default index card (CardsGrid or Workspace)", so the two comments in this file don't contradict each other.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with a focus on **completeness of the generalization** — the risk in a "recognize one more card type" change is a missed twin recognizer or a downstream consumer still hard-coded to the old type, so I traced every place that decides "is this the realm's default index card" and every reference to the renamed CSS hooks.

**Bottom line: this is a clean, correctly-scoped change with no blocking issues.** Both recognizers were updated symmetrically, the rename is safe, and the tests pin the new behavior. Only two minor comment/doc drifts to tidy (both non-blocking).

**What lands right**
- **Both recognizers updated, and they agree.** The publish handler (`isDefaultRealmIndexAdoption`, reading `index.json`'s `adoptsFrom`) and the host render route (`#isDefaultRealmIndexCard`, comparing `types[0]` via `internalKeyFor`) are the only two places that classify the default index card. Both were generalized the same way, and they resolve to the same answer for a directly-adopting `index.json`. Details in the inline note on `handle-publish-realm.ts`.
- **The CSS-hook rename is safe.** `boxel-cards-grid-shell` / `data-boxel-cards-grid-index` → `boxel-realm-index-shell` / `data-boxel-realm-index`. I grepped the tree: the old names have zero remaining references, and the new names appear only in `realm-index-boilerplate.ts`. Nothing selects on these (they exist for hydration-replaceable markup), so the rename is inert beyond readability.
- **No third recognizer downstream.** The published-realm SSR injection path keys off the `includePrerenderedDefaultRealmIndex` flag (read from RealmConfig on disk into `RealmInfo`), not the card type — so writing the flag from the publish handler is the whole story; nothing else needed generalizing.
- **Tests mirror the CardsGrid cases faithfully.** The host test asserts a not-opted-in Workspace realm gets `REALM_INDEX_BOILERPLATE_HTML`, and the realm-server test asserts publishing a Workspace-indexed realm writes the opt-in flag — each a line-for-line analogue of the existing CardsGrid test. Both are green on the head commit in CI.

**On the one red check (Host Memory Baseline)** — not caused by this diff. It fails on "modules not in baseline," and the modules it names are all the Workspace card's own test suites, which come from the base branch this PR stacks on, not from this PR (which adds no new test module and no new retained state). It should re-baseline once the stack retargets to `main`. Worth a glance to confirm, but not a regression here.

**Recommendations** (all non-blocking)
1. `html.ts` — update the stale `Model.useRealmIndexBoilerplate` docstring ("the type chain is base CardsGrid") to match the generalized sibling comments. See the file comment.
2. `handle-publish-realm.ts` — optionally generalize the opening of the "The CardsGrid CardDef can be referenced two equivalent ways…" comment so the shared rationale reads as card-agnostic. See the inline note.

**Adjacent, out of scope** — `RealmInfo.includePrerenderedDefaultRealmIndex` in `packages/runtime-common/realm.ts` is documented as opt-in "for the realm's default **CardsGrid** index card." This PR widens that field's meaning to Workspace-indexed realms too, so the field doc is now narrower than the behavior. That file isn't in this diff; a one-word tweak there (or in a follow-up) would keep the doc honest.

---

habdelra (approved):

packages/realm-server/handlers/handle-publish-realm.ts
lukemelia: [Claude Code 🤖] Done in 632a1b9 — opened the comment with "A base CardDef can be referenced two equivalent ways…" so the shared two-module-forms rationale reads as card-agnostic, matching the "see the note above" pointer below it.

packages/host/app/routes/render/html.ts
lukemelia: [Claude Code 🤖] Fixed in 632a1b9 — the `useRealmIndexBoilerplate` docstring now reads "the type chain begins with a recognized default index card (CardsGrid or Workspace)", consistent with the sibling comments. Also generalized the `RealmInfo.includePrerenderedDefaultRealmIndex` field doc in `packages/runtime-common/realm.ts` (the adjacent out-of-scope item) in the same commit.

richardhjtan (commented):
[Claude Code 🤖] Re-reviewed with a focus on **verifying the earlier thread's fixes as actual diffs, not as descriptions of them**, plus re-tracing the two default-index recognizers and the CSS-hook rename from scratch — since "recognize one more type" changes fail quietly when a twin call site or a downstream consumer is missed.

**Bottom line: confirmed correct, no blocking issues.** The two doc-drift fixes promised in the existing threads landed exactly as described, and my own independent trace of the mechanism agrees with the earlier review: both recognizers resolve to the same answer, the CSS rename is complete, and the new tests exercise real behavior rather than just shape.

**What I verified independently**
- **The doc fixes are byte-exact matches for what the reply comments promised.** Diffing the two commits on this branch, `handle-publish-realm.ts`'s "The CardsGrid CardDef can be referenced two equivalent ways…" comment now opens with "A base CardDef can be referenced two equivalent ways…", `routes/render/html.ts`'s `useRealmIndexBoilerplate` docstring now says "a recognized default index card (CardsGrid or Workspace)", and `runtime-common/realm.ts`'s `RealmInfo.includePrerenderedDefaultRealmIndex` doc was also generalized in the same commit (the item flagged as adjacent/out-of-scope in the prior review). All three match their thread replies exactly.
- **The two recognizers still agree, traced past the diff.** `getTypes` in `packages/host/app/routes/render/meta.ts` walks from the instance's own constructor upward, pushing each ancestor's ref and stopping at `baseRef` — so `types[0]` in `#isDefaultRealmIndexCard` (`html.ts`) is the instance's own most-derived type, not some intermediate ancestor. That's the same thing `isDefaultRealmIndexAdoption` in `handle-publish-realm.ts` checks on the raw `index.json` `adoptsFrom`. For a realm whose index directly adopts `CardsGrid` or `Workspace` (not a further subclass), both land on the same true/false answer, and a bespoke subclass is correctly excluded from both.
- **`Workspace` is a real target, not a placeholder ref.** `packages/base/workspace.gts` exports `class Workspace extends CardDef` — the module/name pair added to both `DEFAULT_REALM_INDEX_REFS` and `DEFAULT_REALM_INDEX_ADOPTIONS` resolves to an actual base CardDef.
- **The CSS rename left nothing behind.** Grepped the tree for `boxel-cards-grid-shell` and `data-boxel-cards-grid-index` — zero hits. The new names (`boxel-realm-index-shell`, `data-boxel-realm-index`) appear only in `realm-index-boilerplate.ts`, where they were introduced, confirming there was never a stylesheet rule keyed on either name (a pure rename, not a styling change).
- **The new tests assert real rendered/persisted content, not just shape.** The host test drives `getInstance` through actual realm indexing and checks `isolatedHtml` against the boilerplate constant; the realm-server test publishes a realm and reads `includePrerenderedDefaultRealmIndex` back off the written `realm.json` on disk. Both would fail if the recognizer regressed.

**Adjacent, out of scope** — the `percy/-cardstack-host` check on the head commit reports 6 visual changes needing review. This diff touches no stylesheet (confirmed above — the renamed CSS hooks have no CSS rules), so those diffs aren't from this change; they're most likely inherited from the Workspace-port branch this PR stacks on. Worth a glance before this retargets to `main`, but not something this diff introduced.

One minor, non-blocking observation in the inline thread on `realm-indexing-test.gts`.

packages/host/tests/integration/realm-indexing-test.gts
```diff
@@ -5191,6 +5191,45 @@ posts/ignore-me.json
...
+            },
+          },
+        },
+      },
+    });
+    let entry = await getInstance(realm, new URL(`${testRealmURL}index`));
+    assert.ok(entry, 'realm index card is indexed');
+    assert.strictEqual(
+      entry?.isolatedHtml,
+      REALM_INDEX_BOILERPLATE_HTML,
```
richardhjtan, line 5228: [Claude Code 🤖] **Follow-up, non-blocking — the "opted-in" case has no Workspace mirror.** This block adds the Workspace analogue of the existing "not opted in" CardsGrid test, but there's no Workspace analogue of `isolated HTML for the default CardsGrid realm-index card is rendered normally when realm has opted in` a few tests up. Risk is low: `#isDefaultRealmIndexCard` in `packages/host/app/routes/render/html.ts` only uses the card's type to decide *whether* the realm's index counts as a default index at all — once that's true, the opted-in branch (`info?.includePrerenderedDefaultRealmIndex !== true`) is identical, type-agnostic code already covered by the CardsGrid opted-in test. Still, a Workspace-opted-in test would close the loop the same way the "not opted in" test does here, and would catch a regression if that branch ever became type-aware.

richardhjtan (approved):

PR #5586, opened 2026-07-22
Give ProgressBar the progressbar ARIA role and value range

lukemelia:
## What

The shared `ProgressBar` (`@cardstack/boxel-ui/components`) rendered a plain `<div>`, so assistive technology couldn't identify it as a progress indicator. Add the standard progressbar semantics:

- `role="progressbar"`
- `aria-valuenow` — the value clamped into `[0, max]` (via a new `valueNow` getter, so it's always a valid value between min and max)
- `aria-valuemin="0"` and `aria-valuemax={{max}}`
- `aria-valuetext` — the rounded percentage (e.g. `"25%"`) for a friendly screen-reader readout
- keeps the existing `aria-label`

## Why

Every consumer of `ProgressBar` benefits, so the fix belongs in the shared component rather than in each caller. (Surfaced while adopting `ProgressBar` in a card, whose hand-rolled bar previously carried these attributes itself.)

## Tests

Adds `progress-bar-test.gts` (the component had none): asserts the role + aria value range, that `aria-valuenow` clamps into `[0, max]`, and that `@label` becomes the accessible name. Run locally — **3 passed**. Type-check, eslint, template-lint, prettier all clean.

### DIFF
--- a/packages/boxel-ui/addon/src/components/progress-bar/index.gts
+++ b/packages/boxel-ui/addon/src/components/progress-bar/index.gts
@@ -40,6 +40,14 @@ export default class ProgressBar extends Component<Signature> {
     return position ?? 'end';
   }
 
+  // A node with role="progressbar" must have an accessible name, and — unlike
+  // roles that take their name from content — the visible label text does not
+  // supply one. Fall back to a generic name when no label is passed so the
+  // progressbar always has one; callers with context pass a specific `@label`.
+  get accessibleLabel(): string {
+    return this.args.label || 'Progress';
+  }
+
   <template>
     <div
       class='boxel-progress-bar'
@@ -49,7 +57,7 @@ export default class ProgressBar extends Component<Signature> {
       aria-valuemin='0'
       aria-valuemax={{this.max}}
       aria-valuetext={{this.progressPercentage}}
-      aria-label={{@label}}
+      aria-label={{this.accessibleLabel}}
       ...attributes
     >
       <div class='progress-bar-container'>
--- a/packages/boxel-ui/addon/src/components/progress-bar/index.gts
+++ b/packages/boxel-ui/addon/src/components/progress-bar/index.gts
@@ -25,7 +25,10 @@ export default class ProgressBar extends Component<Signature> {
   }
 
   get progressPercentage(): string {
-    return Math.round((this.valueNow / this.max) * 100) + '%';
+    // Guard the degenerate max=0 case so aria-valuetext (announced by screen
+    // readers) and the CSS width don't become "NaN%".
+    const pct = this.max === 0 ? 0 : (this.valueNow / this.max) * 100;
+    return Math.round(pct) + '%';
   }
 
   get progressWidth(): ReturnType<typeof htmlSafe> {
--- a/packages/boxel-ui/addon/src/components/progress-bar/index.gts
+++ b/packages/boxel-ui/addon/src/components/progress-bar/index.gts
@@ -14,10 +14,18 @@ interface Signature {
 }
 
 export default class ProgressBar extends Component<Signature> {
+  get max(): number {
+    return this.args.max ?? 100;
+  }
+
+  // The current value, clamped into [0, max] so it's always a valid
+  // `aria-valuenow` between `aria-valuemin` and `aria-valuemax`.
+  get valueNow(): number {
+    return Math.min(Math.max(this.args.value ?? 0, 0), this.max);
+  }
+
   get progressPercentage(): string {
-    const max = this.args.max ?? 100;
-    const value = this.args.value ?? 0;
-    return Math.round(Math.min(Math.max((value / max) * 100, 0), 100)) + '%';
+    return Math.round((this.valueNow / this.max) * 100) + '%';
   }
 
   get progressWidth(): ReturnType<typeof htmlSafe> {
@@ -36,6 +44,11 @@ export default class ProgressBar extends Component<Signature> {
     <div
       class='boxel-progress-bar'
       data-test-boxel-progress-bar
+      role='progressbar'
+      aria-valuenow={{this.valueNow}}
+      aria-valuemin='0'
+      aria-valuemax={{this.max}}
+      aria-valuetext={{this.progressPercentage}}
       aria-label={{@label}}
       ...attributes
     >
--- a/packages/boxel-ui/test-app/tests/integration/components/progress-bar-test.gts
+++ b/packages/boxel-ui/test-app/tests/integration/components/progress-bar-test.gts
@@ -1,0 +1,44 @@
+import { render } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'test-app/tests/helpers';
+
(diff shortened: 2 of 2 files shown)


burieberry (approved):

packages/boxel-ui/addon/src/components/progress-bar/index.gts
```diff
@@ -14,10 +14,18 @@ interface Signature {
...
+  get valueNow(): number {
+    return Math.min(Math.max(this.args.value ?? 0, 0), this.max);
+  }
+
   get progressPercentage(): string {
-    const max = this.args.max ?? 100;
-    const value = this.args.value ?? 0;
-    return Math.round(Math.min(Math.max((value / max) * 100, 0), 100)) + '%';
+    return Math.round((this.valueNow / this.max) * 100) + '%';
   }
```
habdelra, line 32: [Claude Code 🤖] **`aria-valuetext="NaN%"` when `@max` is `0` — reachable, and a screen reader announces it literally.**

`progressPercentage` divides by `this.max`. When a caller passes `@max={{0}}`, `this.args.max ?? 100` keeps the `0` (nullish coalescing only replaces `null`/`undefined`), so `this.valueNow / this.max` is `0 / 0 = NaN`, `Math.round(NaN)` is `NaN`, and the getter returns `"NaN%"`. That string now feeds `aria-valuetext`, so assistive tech reads out "NaN percent"; it also lands in `progressWidth` as `width: NaN%;`.

**Verified** — old vs. new math for `max=0`:
```
value=0 max=0 → OLD width "NaN%"  | NEW valuetext "NaN%", valuemax "0"
value=5 max=0 → OLD width "100%"  | NEW valuetext "NaN%"
```

**Reachable today:** `packages/software-factory/realm/overview.gts` renders `<ProgressBar @value={{this.doneIssues}} @max={{this.totalIssues}} />`, where `totalIssues` is the length of `this.project?.issues ?? []` — `0` for an empty board (and `doneIssues` is `0` too). That consumer guards its own readout with `if (!this.totalIssues) return 0` in `progressPct`, but the ProgressBar still receives the raw `0`, so the empty-board bar announces "NaN%".

**Class:** the NaN arithmetic is pre-existing (the width already went to `NaN%`), but this PR newly routes it into a screen-reader-announced attribute — which cuts against the a11y goal of the change. Non-blocking, but it's a one-liner worth folding in here since the fix lives in the getter you're already touching:

```suggestion
  get progressPercentage(): string {
    const pct = this.max === 0 ? 0 : (this.valueNow / this.max) * 100;
    return Math.round(pct) + '%';
  }
```

This also fixes the pre-existing `width: NaN%;`.

---

packages/boxel-ui/test-app/tests/integration/components/progress-bar-test.gts
```diff
@@ -0,0 +1,54 @@
...
+  test('clamps aria-valuenow into [0, max]', async function (assert) {
+    await render(
+      <template><ProgressBar @value={{20}} @max={{12}} /></template>,
+    );
+
+    assert
+      .dom('[data-test-boxel-progress-bar]')
+      .hasAttribute('aria-valuenow', '12')
+      .hasAttribute('aria-valuetext', '100%');
+  });
```
habdelra, line 31: [Claude Code 🤖] **Coverage gap: this pins only the upper clamp.**

`valueNow` is `Math.min(Math.max(value, 0), max)` — two clamps. This case exercises the upper one (`20 → 12`), but nothing pins the lower one: a negative `@value` should land at `0`. A regression that dropped the `Math.max(..., 0)` would still pass every test in this file.

Consider adding a case like `@value={{-5}} @max={{12}}` asserting `aria-valuenow` is `'0'` (and `aria-valuetext` `'0%'`). If you take the `max=0` guard from the `progressPercentage` thread, a `@max={{0}}` case asserting `aria-valuetext` is not `'NaN%'` would pin that behavior too.

**Class:** follow-up / non-blocking test hardening.

---

habdelra (commented):
[Claude Code 🤖] Reviewed for ARIA correctness and whether the new progressbar semantics hold across the component's real consumers (I checked out the head and traced each `ProgressBar` call site).

**Bottom line: a clean, correctly-scoped a11y fix with real test coverage — no blocking issues.** Putting the semantics in the shared component rather than each caller is the right call, and the clamping design is exactly what assistive tech expects. One reachable edge case (`aria-valuetext="NaN%"` when `@max` is `0`) is worth folding in since the fix is a one-liner in a getter you're already changing.

**What lands right (and why it's safe):**
- **`valueNow` clamped into `[0, max]`.** `aria-valuenow` never escapes the declared `aria-valuemin`/`aria-valuemax` range — screen readers compute their percentage from `valuenow` within that range, so an out-of-range `valuenow` would be announced wrong. The clamp is the correct guard. I verified the display width is unchanged for every in-range and over-range input (only the degenerate `max=0` case differs — see the inline thread).
- **`accessibleLabel` fallback to `'Progress'`.** Correct: a `role="progressbar"` node needs an accessible name, and unlike name-from-content roles the visible label text doesn't supply one, so the fallback prevents an unnamed progressbar. The hardcoded English is idiomatic here — boxel-ui has no i18n layer (no `ember-intl`/`intl` usage in the addon), so sibling components hardcode strings the same way.
- **`...attributes` after the aria attributes.** Callers can still override `aria-label`/`role`/etc. via attribute splatting, so this doesn't lock consumers out.
- **Percentage as `aria-valuetext`.** A friendly readout ("25%") consistent with `valuenow`/`valuemax`; a reasonable choice.

**Recommendations:**
1. Guard the `max === 0` division in `progressPercentage` so `aria-valuetext` (and the CSS width) can't become `NaN%` on an empty-board bar — see the inline thread on `index.gts`. Non-blocking, one-liner.
2. Add a lower-clamp test (negative `@value` → `aria-valuenow` `'0'`); optionally a `@max={{0}}` case — see the inline thread on the test file. Non-blocking.

**Adjacent, out of scope:** `progressBarPosition` has a dead `return position ?? 'end'` after an earlier `if (!position) return ''` — the `??` branch is unreachable. Pre-existing, not this PR; flagging for whoever next touches that getter.

---

packages/boxel-ui/addon/src/components/progress-bar/index.gts
lukemelia: [Claude Code 🤖] Fixed in 0dd27b2db3. `progressPercentage` now treats `max === 0` as `0%`, so neither `aria-valuetext` nor the CSS width can become `NaN%` on an empty-board bar. Added a test for the `@max={{0}}` case.

packages/boxel-ui/test-app/tests/integration/components/progress-bar-test.gts
lukemelia: [Claude Code 🤖] Added in 0dd27b2db3: a lower-clamp test (`@value={{-5}}` → `aria-valuenow` `0`, `aria-valuetext` `0%`) and a `@max={{0}}` test asserting `aria-valuetext` is `0%` rather than `NaN%`.

PR #5587, opened 2026-07-22
Default new realms to a Workspace index card

lukemelia:
## What

`create-realm.ts` seeds every new realm's `index.json`. Point it at `Workspace` instead of `CardsGrid`, so a newly created realm opens to the Workspace experience. This is the "replace CardsGrid" keystone — the card, the recognition, and the migration are only meaningful once the default flips.

CardsGrid stays available for existing/bespoke realms; only the *default* for freshly created realms changes.

## Depends on

- The Workspace card must exist (so a new realm's index resolves).
- The publish handler + prerender fast-path must recognize Workspace (so a new realm still gets the prerendered default-index treatment on publish, and the realm-server tests that assumed a CardsGrid default stay correct). This PR is stacked on that recognition branch.

## Blast radius (audited)

The realm-server test helper `createRealm` seeds its own `fileSystem`, so it does **not** go through this handler — most realm-server tests are unaffected. Of the tests that hit the real `/_create-realm` endpoint:

- `index-responses` and `realm-lifecycle` seed their own `index.json` → unaffected.
- `realm-routing`'s deadlock guard uses a mock fixture, not a real create+prerender → unaffected (comment now slightly historical).
- `publish-unpublish` creates its source realm via the endpoint and *did* rely on the default being CardsGrid → **updated here**: the default-index case now exercises Workspace, and the explicit case is repurposed to cover the legacy CardsGrid path, so both recognized defaults stay tested; the bespoke-index negative case is unchanged.

Also fixed software-factory comments that described the old CardsGrid default (the factory overwrites the index regardless, so it's behavior-neutral).

## Test plan

realm-server `ember-tsc`, eslint, prettier clean. **Not run locally** — the realm-server publish/create suites need test-pg + their own prerender, which isn't wired up in this session; they run in CI. Given the default flip is behaviorally significant, CI's full realm-server + matrix suites (which create realms via `/_create-realm`) are the real confirmation. The publish test changes mirror the existing passing tests' structure.

### DIFF
--- a/packages/matrix/helpers/index.ts
+++ b/packages/matrix/helpers/index.ts
@@ -450,12 +450,20 @@ export async function enterWorkspace(
 }
 
 export async function showAllCards(page: Page) {
-  try {
-    await page
-      .locator(`[data-test-boxel-filter-list-button="All Cards"]`)
-      .click();
-  } catch (e) {
-    console.warn('all cards filter is not found');
+  // A realm's index is either the legacy CardsGrid — whose "All Cards" filter
+  // lists every card — or the default Workspace, whose Library tab renders the
+  // same live card grid. Use whichever this realm's index provides; do nothing
+  // if neither is present (a realm with a bespoke index card).
+  let allCardsFilter = page.locator(
+    `[data-test-boxel-filter-list-button="All Cards"]`,
+  );
+  if ((await allCardsFilter.count()) > 0) {
+    await allCardsFilter.click();
+    return;
+  }
+  let libraryTab = page.locator(`[data-test-workspace-tab="library"]`);
+  if ((await libraryTab.count()) > 0) {
+    await libraryTab.click();
   }
 }
 
--- a/packages/matrix/helpers/index.ts
+++ b/packages/matrix/helpers/index.ts
@@ -452,16 +452,23 @@ export async function enterWorkspace(
 export async function showAllCards(page: Page) {
   // A realm's index is either the legacy CardsGrid — whose "All Cards" filter
   // lists every card — or the default Workspace, whose Library tab renders the
-  // same live card grid. Use whichever this realm's index provides; do nothing
-  // if neither is present (a realm with a bespoke index card).
+  // same live card grid. `.count()` is point-in-time, so wait for the index to
+  // render one of its browse affordances before deciding; checking right after
+  // a `goto` would otherwise miss both and silently no-op. A bespoke index has
+  // neither — leave it be.
   let allCardsFilter = page.locator(
     `[data-test-boxel-filter-list-button="All Cards"]`,
   );
+  let libraryTab = page.locator(`[data-test-workspace-tab="library"]`);
+  try {
+    await allCardsFilter.or(libraryTab).first().waitFor({ timeout: 30_000 });
+  } catch {
+    return;
+  }
   if ((await allCardsFilter.count()) > 0) {
     await allCardsFilter.click();
     return;
   }
-  let libraryTab = page.locator(`[data-test-workspace-tab="library"]`);
   if ((await libraryTab.count()) > 0) {
     await libraryTab.click();
   }
--- a/packages/realm-server/handlers/create-realm.ts
+++ b/packages/realm-server/handlers/create-realm.ts
@@ -187,8 +187,8 @@ export async function createRealm(
         type: 'card',
         meta: {
           adoptsFrom: {
-            module: '@cardstack/base/cards-grid',
-            name: 'CardsGrid',
+            module: '@cardstack/base/workspace',
+            name: 'Workspace',
           },
         },
       },
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -249,6 +249,7 @@ class Isolated extends Component<typeof Workspace> {
   <template>
     <section
       class='card-grid'
+      data-test-workspace-index
       {{this.setupRealmSubscription this.primaryRealm}}
     >
       <header class='frame'>
@@ -256,16 +257,19 @@ class Isolated extends Component<typeof Workspace> {
           <button
             type='button'
             class='tab {{if (eq this.segment "home") "active"}}'
+            data-test-workspace-tab='home'
             {{on 'click' (this.setSegment 'home')}}
           ><HouseIcon class='tab-icon' /> Home</button>
           <button
             type='button'
             class='tab {{if (eq this.segment "library") "active"}}'
+            data-test-workspace-tab='library'
             {{on 'click' (this.setSegment 'library')}}
           ><LayoutGridIcon class='tab-icon' /> Library</button>
           <button
             type='button'
             class='tab {{if (eq this.segment "activity") "active"}}'
+            data-test-workspace-tab='activity'
             {{on 'click' (this.setSegment 'activity')}}
           ><ActivityIcon class='tab-icon' />
             Activity{{#if this.runningJobs.length}}<span
--- a/packages/matrix/tests/create-realm.spec.ts
+++ b/packages/matrix/tests/create-realm.spec.ts
@@ -32,9 +32,8 @@ test.describe('Create Realm via Dashboard', () => {
     await expect(
       page.locator(`[data-test-stack-card="${newRealmURL}index"]`),
     ).toBeVisible();
-    await expect(
-      page.locator(`[data-test-boxel-filter-list-button]`),
-    ).toHaveCount(3);
+    // A newly created realm defaults to a Workspace index card.
+    await expect(page.locator(`[data-test-workspace-index]`)).toBeVisible();
 
     await page.locator('[data-test-submode-switcher] button').click();
     await page.locator('[data-test-boxel-menu-item-text="Host"]').click();
--- a/packages/matrix/tests/live-cards.spec.ts
+++ b/packages/matrix/tests/live-cards.spec.ts
@@ -226,14 +226,15 @@ test.describe('Live Cards', () => {
     const realmURL = new URL(`${username}/${realmName}/`, serverIndexUrl).href;
     await createRealm(page, realmName);
 
-    // Land on the realm's grid so the create-new-card affordance is present.
+    // Open the card chooser via the operator-mode "New" button — an
+    // index-agnostic affordance, so this doesn't depend on whether the realm
+    // index is a CardsGrid or a Workspace. Its "Choose a card type..." item
+    // opens the same chooser CardsGrid's create button did.
     await page.goto(realmURL);
-    await showAllCards(page);
-    await expect(
-      page.locator(`[data-test-boxel-filter-list-button="All Cards"]`),
-    ).toHaveCount(1);
-
-    await page.locator('[data-test-create-new-card-button]').click();
+    await page.locator('[data-test-new-file-button]').click();
+    await page
+      .locator('[data-test-boxel-menu-item-text="Choose a card type..."]')
+      .click();
     await expect(page.locator('[data-test-card-chooser-modal]')).toBeVisible();
     await expect(page.locator('[data-test-item-button]').first()).toBeVisible();
 
--- a/packages/matrix/tests/live-cards.spec.ts
+++ b/packages/matrix/tests/live-cards.spec.ts
@@ -74,9 +74,7 @@ test.describe('Live Cards', () => {
     await page.goto(realmURL);
     await showAllCards(page);
 
-    await expect(
-      page.locator(`[data-test-boxel-filter-list-button="All Cards"]`),
-    ).toHaveCount(1);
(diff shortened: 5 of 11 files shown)


packages/realm-server/handlers/create-realm.ts
```diff
@@ -187,8 +187,8 @@ export async function createRealm(
         type: 'card',
         meta: {
           adoptsFrom: {
-            module: '@cardstack/base/cards-grid',
-            name: 'CardsGrid',
+            module: '@cardstack/base/workspace',
+            name: 'Workspace',
```
lukemelia, line 191: [Claude Code 🤖] **This flips the default realm index to Workspace, but the acceptance tests that assert the old CardsGrid index UI weren't updated — only the realm-server publish tests were.**

`packages/matrix/tests/create-realm.spec.ts:35` creates a realm and asserts its index renders `[data-test-boxel-filter-list-button]` with `toHaveCount(3)` — the CardsGrid sidebar (CardsGrid renders the boxel-ui `FilterList`). A Workspace index renders its own rail rather than that `FilterList`, so the count is no longer 3 and the spec fails (this is the deterministic Matrix Client Tests shard-1 failure).

Sweep every spec that creates a realm and then renders/asserts its index card and update the index-UI expectations to Workspace's markup: `create-realm.spec.ts:35` at minimum, plus check the other shard-1 failures (`live-cards.spec.ts`) and any host acceptance test that visits a freshly-created realm's `index`.

packages/realm-server/handlers/create-realm.ts
```diff
@@ -187,8 +187,8 @@ export async function createRealm(
         type: 'card',
         meta: {
           adoptsFrom: {
-            module: '@cardstack/base/cards-grid',
-            name: 'CardsGrid',
+            module: '@cardstack/base/workspace',
```
lukemelia, line 190: [Claude Code 🤖] **Weigh the cost of making the ~4000-line `Workspace` card the default realm index.** Every newly-created realm now prerenders a Workspace instance, and every test/flow that creates a realm pulls Workspace's transitive module graph (CardsGridLayout, several boxel-icons, BoxelInput, the `PublishRealmCommand` host tool chain, …) instead of the small CardsGrid.

The Host Memory Baseline check on this PR flags large increases (e.g. operator-mode tests), and shard 1 shows `correctness-checks.spec.ts` dying with "Target page, context or browser has been closed" — both consistent with memory pressure from loading the heavy card in more places. The baseline numbers look partly noisy (0.0 MB baselines, negative recent samples), so confirm against a clean baseline, but verify the prerender/host-test memory impact is acceptable before flipping the default.

lukemelia (commented):
[Claude Code 🤖] **Code review (high effort) — PR #5587**

The diff itself is correct and internally consistent: the `create-realm.ts` `adoptsFrom` swap is the one behavioral change, and the realm-server publish tests were updated in step (the bespoke-index case correctly uses `assert.notStrictEqual(..., true)`, and the publish handler's `isDefaultRealmIndexAdoption` already recognizes both CardsGrid and Workspace on the base branch).

The findings are about the **downstream reach** of flipping the default index — which is also why CI is red:

1. (inline) Acceptance tests that assert the created realm's CardsGrid sidebar weren't updated — `matrix/tests/create-realm.spec.ts:35` fails deterministically, and likely siblings.
2. (inline) Making the heavy Workspace card the default index broadens where its module graph is loaded — plausibly behind the Host Memory Baseline regression and the 'browser closed' crashes in the matrix shard.

Minor (not blocking): `realm-server/tests/realm-routing-test.ts:190`'s comment ("The CardsGrid in index.json fires _federated-search") is now stale — the seeded index is a Workspace. The test should still pass if Workspace fires the same federated search during its prerender, but the comment should be refreshed.

**CI status:** Matrix Client Tests (1,3)/(3,3) + Matrix Playwright fail — `create-realm.spec.ts` is the deterministic, PR-caused one; Host Memory Baseline fails (see finding 2).

lukemelia:
[Claude Code 🤖] **Migrated the matrix tests that assumed a CardsGrid index** — pushed in ed3f3c155e:

- Added `data-test` hooks to the Workspace card (index root + Home/Library/Activity tabs; it had none).
- `showAllCards` now handles both index types (CardsGrid "All Cards" filter, or the Workspace Library tab — same live `CardsGridLayout` grid).
- Fixed `create-realm.spec.ts` (asserts the Workspace index rendered) and `live-cards.spec.ts` test 1 (asserts the shared cards grid).

**Two items intentionally left for you (not fixed):**

1. **`live-cards.spec.ts` "can specify a card by URL in the card chooser"** — this one clicks `[data-test-create-new-card-button]` (CardsGrid's `AddButton`). The Workspace index doesn't render that button; its create affordances are the per-type rail `+` and a "New card" welcome button. So it's not a selector swap — it needs either a create-card affordance added to the Workspace index or a reworked flow. Decide which fits the product intent.
2. **`correctness-checks.spec.ts` "browser has been closed"** — this reads as memory pressure from loading the ~4000-line Workspace card as the default index (see review finding #2), not a selector issue. Worth confirming the prerender/host-test memory impact before this lands.

I couldn't run the matrix suite locally (needs the full synapse/docker stack), so the pushed fixes are CI-verified — watch the next Matrix Client Tests run.

lukemelia:
[Claude Code 🤖] **Update — the card-chooser coupling is fixed** (1781bfe9ed), superseding item 1 in my earlier note.

The `[data-test-create-new-card-button]` (CardsGrid's `AddButton`) was used to open the card chooser in `live-cards`, `skills`, and `tools` — all of which broke on the Workspace index. Rather than depend on any index affordance, they now open the chooser via the operator-mode **New** button (`[data-test-new-file-button]` → "Choose a card type…"), which calls the same `chooseCard()` and pops the same modal. `skills.spec.ts:352` and `tools.spec.ts` (30/87/141/296) were failing on this in shard 3 too and are covered by the same swap.

**Still open:**
- `correctness-checks.spec.ts` "browser has been closed" — reads as memory pressure from the heavy Workspace card as default index (review finding #2), not a selector issue.
- `registration-with-token.spec.ts:50` failed in shard 3 but is unrelated to the index change — likely a flake; worth a re-run to confirm.

All CI-verified on the next Matrix run.

packages/matrix/helpers/index.ts
```diff
@@ -450,12 +450,20 @@ export async function enterWorkspace(
...
+    `[data-test-boxel-filter-list-button="All Cards"]`,
+  );
+  if ((await allCardsFilter.count()) > 0) {
+    await allCardsFilter.click();
+    return;
+  }
+  let libraryTab = page.locator(`[data-test-workspace-tab="library"]`);
+  if ((await libraryTab.count()) > 0) {
+    await libraryTab.click();
   }
```
habdelra, line 474: [Claude Code 🤖] **This helper no longer waits for the index to render, so it can silently no-op and leave the caller asserting a grid that was never opened.** Regression introduced by this rewrite; non-blocking but worth fixing, because it's a latent flake in exactly the code paths this change touches.

**Mechanism.** The previous body called `.click()`, which auto-waits (up to the Playwright timeout) for the affordance to appear. This version probes with `await allCardsFilter.count()` / `await libraryTab.count()`, and `Locator.count()` is one of the few non-auto-waiting queries — it snapshots the DOM and returns immediately. So if `showAllCards` runs before the realm index has finished rendering, both counts are `0`, the function returns having clicked nothing, and no wait ever happened.

**Why the no-op then fails the caller.** For the default Workspace index the grid is behind the Library tab — the default segment is `home` (`packages/base/workspace.gts:2236`), and `CardsGridLayout` only renders under `{{else if (eq this.segment 'library')}}` (`workspace.gts:609-721`). So if the Library tab is never clicked, `[data-test-cards-grid-cards]` / `[data-test-cards-grid-item=…]` never appears, and the caller's follow-up `toBeVisible()` / `waitFor()` times out. The raciest caller is `live-cards.spec.ts:74-75` (`page.goto(realmURL)` immediately followed by `showAllCards`, with nothing in between to let the app boot). Under the added render/memory cost of the heavier default index, this window widens.

**The fix** — wait for whichever affordance the realm's index renders before probing, so a slow render can't turn this into a no-op:

```js
await page
  .locator(
    `[data-test-boxel-filter-list-button="All Cards"], [data-test-workspace-tab="library"]`,
  )
  .first()
  .waitFor({ state: 'attached' })
  .catch(() => {}); // bespoke index → neither ever appears; fall through to the no-op path
```

The `.count()` branches below then behave as written. (Tradeoff: a genuinely bespoke-index caller waits the full timeout before the no-op — but every current caller creates a default realm, so that path isn't exercised today.)

---

habdelra (commented):
[Claude Code 🤖] This review focused on how flipping the default realm index lands downstream — the seeding path, the recognizer it depends on, the acceptance-test migration, and the memory/render cost — since the behavioral change itself is a two-line swap.

**Bottom line: the core change is correct and well-covered; no blocking correctness issue in the diff. One new latent test-flake (inline on `showAllCards`), plus the memory question — now the only red CI signal — still needs a confirm-benign before merge.**

**What lands right**
- The seed swap in `create-realm.ts` matches the base recognizer exactly: `isDefaultRealmIndexAdoption` (`handle-publish-realm.ts:62-89`) accepts both CardsGrid and Workspace, so a freshly created realm still gets `includePrerenderedDefaultRealmIndex` on publish. `realm-index-boilerplate.ts` already speaks "CardsGrid or Workspace." Nothing on the seeding path was left assuming CardsGrid.
- The publish suite is real regression coverage, not shape-only: the source realm is created through `/_create-realm` in `beforeEach`, so the "default Workspace index" test exercises the handler default rather than a hand-seeded fixture. All three cases stay covered — default Workspace → flag set, bespoke `CardDef` → flag *not* set (`assert.notStrictEqual(..., true)`), legacy CardsGrid → flag set.
- The `showAllCards` Library path is semantically equivalent to CardsGrid's "All Cards": Workspace's default `everythingFilter` returns all cards except `Cards Grid`/`Workspace` index instances, and every migrated caller asserts only non-index items, so the visible set matches. `[data-test-cards-grid-cards]` lives in the shared `cards-grid-layout.gts`, so the assertion holds against either index.
- The card-chooser rework is genuinely index-agnostic — the operator-mode New button → "Choose a card type…" calls the same `chooseCard()` the CardsGrid `+` did. The matrix migration is complete; no spec still references `[data-test-create-new-card-button]`.

**Findings**
1. (inline, `packages/matrix/helpers/index.ts`) `showAllCards` swapped an auto-waiting `.click()` for a non-waiting `.count()`, so a slow index render makes it a silent no-op and the caller times out waiting for a grid that was never opened. Non-blocking, but it's a latent flake in the exact paths this change touches — see the thread for the mechanism and a one-line wait fix.
2. (non-blocking, needs a confirm) Making the heavier Workspace card the default index is the concern already open in the thread on `create-realm.ts`. Fresh signal in its favor: **Host Memory Baseline is now green** on the head commit, which softens the host-side worry. What remains is the matrix "browser has been closed" crash — Matrix Client Tests (1,3) and (3,3) are the only red checks on the head commit. Worth attributing that crash from the uploaded Playwright trace (memory-driven vs. the separate `registration-with-token` flake) before merging, since it's the last thing between this and green.
3. (minor, follow-up) `packages/realm-server/tests/realm-routing-test.ts` — the deadlock-guard comment still reads "The CardsGrid in index.json fires `_federated-search`…", but the seeded default is now Workspace. The test uses a mock fixture so it still passes; the comment is just historical and reads as describing current behavior.

**Recommendations**
1. Add the `waitFor` guard to `showAllCards` (finding 1) so the helper can't no-op under a slow render.
2. Confirm the matrix "browser closed" crash is not caused by the heavier default index before merge (finding 2); Host Memory Baseline being green is reassuring but not conclusive for the matrix runner.
3. Refresh the `realm-routing-test.ts` comment to name Workspace (finding 3).

---

packages/matrix/helpers/index.ts
lukemelia: [Claude Code 🤖] Resolved in 1d5c822 — `showAllCards` now waits for whichever browse affordance the index renders before probing, matching the fix suggested here:

```js
try {
  await allCardsFilter.or(libraryTab).first().waitFor({ timeout: 30_000 });
} catch {
  return; // bespoke index → neither appears; fall through to the no-op path
}
```

Confirmed against the latest full-shard logs: the `live-cards.spec.ts:77` and `skills.spec.ts:431` failures that this no-op was causing are gone on HEAD (the analyzed run predated the fix). `[data-test-filter-list-item]` no longer appears in any matrix test either.

packages/realm-server/handlers/create-realm.ts
lukemelia: Considered. And OK'd.

habdelra (approved):

jurgenwerk:
I tested this locally and it works great 

image

PR #5588, opened 2026-07-23
Fix Rich Markdown embeds becoming broken links in isolated view after save

FadhlanR:
## Background and Goal

A card/file embed inserted into a `RichMarkdownField` previews correctly in the compose editor but renders as a **broken link** in the saved isolated/embedded view (CS-12111). This fixes that so a just-added embed resolves in the saved view exactly as it did in compose.

## Where to start

- `packages/base/query-field-support.ts` — `ensureQueryFieldSearchResource`. The reuse branch now rebuilds an inert cached resource; provenance is recorded at creation.
- `packages/base/card-api.gts` — `FallbackCardStore` gains an `isFallbackCardStore` brand so the query-field layer can recognize the storeless fallback.

- **Root cause:** the compose editor resolves refs via a live `getCards` search, while the saved isolated view resolves only through the query-backed `linkedCards` / `linkedFiles` relationships. That query field's search resource is cached per field-instance and reused unconditionally. When the field is first read in edit mode, `getStore` returns a `FallbackCardStore` whose `getSearchResource` hands back a permanently-empty, non-updating resource (and a just-created card's refs aren't resolved yet). The isolated render then reuses that inert resource → broken link.
- **Fix:** record whether a resource was built against the fallback store or before its `$this.*` query resolved, and rebuild it once a real store / resolvable query is available. It's a one-shot transition — it never fires while still on the fallback store (no loop), and the prerender seed path is untouched.
- **Verified in-app** on the dev stack: edit a card with a Rich Markdown body → insert `:card[...]` → save → isolated now renders the embed resolved (no broken-link box). Regression test added at `packages/host/tests/acceptance/markdown-embed-save-isolated-test.gts` covering both card and file embeds.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -4891,6 +4891,12 @@ function myLoader(): Loader {
 }
 
 class FallbackCardStore implements CardStore {
+  // Marks this as the storeless fallback (`getStore` hands one back when an
+  // instance isn't registered in `stores`). Its `getSearchResource` returns a
+  // permanently-empty, non-updating resource, so query-backed relationships
+  // resolved against it can never populate. Query-field support reads this brand
+  // to rebuild such a resource once a real store is available. [CS-12111]
+  readonly isFallbackCardStore = true;
   #instances: Map<string, CardDef> = new Map();
   #fileMetaInstances: Map<string, FileDef> = new Map();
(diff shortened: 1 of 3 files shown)


FadhlanR:
Cannot reproduce it anymore.

PR #5592, opened 2026-07-23
Add homepage realm test suite (CS-12299)

burieberry:
Linear: [CS-12299](https://linear.app/cardstack/issue/CS-12299/add-homepage-realm-test-suite-realm-integrity-pure-logic-source)

25-test suite for the homepage realm, built during the CS-12248 hardening pass (cardstack/boxel-home#63). Plain `node --test` over `packages/boxel-homepage-realm/tests/` — no build step, Node ≥23 strips types natively, runs in ~150ms via `pnpm test` in the package.

## What it covers

- **`realm-integrity.test.ts`** — 8 structural invariants over `contents/`: every `adoptsFrom.module` and relationship `links.self` resolves to a file; every instance is reachable from a `realm.json`-routed page (or allowlisted in `allowed-orphans.txt`, currently one entry); JSON attribute keys match `@field` declarations; every screenshot asset is referenced; no extensionless files shadow card-instance URLs; navbar `navLinks` anchors exist in the layout; frame/CTA design tokens stay adopted (no literal `80rem`, no per-use token fallbacks). Five of the CS-12248 review findings (dangling theme link, unreachable sections, stray extensionless card, orphaned screenshots, phantom attribute key) would have been caught automatically by this file.
- **`grid-cards.test.ts`** — pure-logic units for the hero grid layout (`lib/grid-cards.ts`: deterministic seeded shuffle, adjacency dedupe, unresolved-link dropping) plus source contracts on `animated-grid.gts` (`@cached` getter, IntersectionObserver gating, cancellation cleanup, data-attribute state).
- **`terminal-scroll.test.ts`** — units for the CLI terminal auto-scroll decisions (`lib/terminal-scroll.ts`) plus the visibility-gating contract on `cli-section.gts`.
- **`serial-list.test.ts`** — Oxford-comma list joins (`lib/serial-list.ts`), including the 2-item regression case.

## Notes

- The `lib/` modules under test live in the contents repo (cardstack/boxel-home) and are imported by relative path; the source-contract tests read `.gts` sources as text, so nothing here needs the host build.
- The suite pairs with the contents-side changes in cardstack/boxel-home#63 — the integrity tests assert the post-cleanup state, so this PR should land at or after that one (they pass against boxel-home `cs-12248-update-website`).
- CI wiring for `pnpm test` in this package is not included here; follow-up on CS-12299: CS-12317

### DIFF
--- a/packages/boxel-homepage-realm/package.json
+++ b/packages/boxel-homepage-realm/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "scripts": {
-    "test": "node --test 'tests/*.test.ts'",
+    "test": "pnpm boxel-homepage:setup && node --test 'tests/*.test.ts'",
     "boxel-homepage:setup": "sh ./scripts/boxel-homepage-setup.sh",
     "boxel-homepage:update": "pnpm boxel-homepage:setup && cd contents && git pull",
     "boxel-homepage:reset": "rm -rf contents && pnpm boxel-homepage:setup"
--- a/packages/boxel-homepage-realm/package.json
+++ b/packages/boxel-homepage-realm/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "scripts": {
+    "test": "node --test 'tests/*.test.ts'",
     "boxel-homepage:setup": "sh ./scripts/boxel-homepage-setup.sh",
     "boxel-homepage:update": "pnpm boxel-homepage:setup && cd contents && git pull",
     "boxel-homepage:reset": "rm -rf contents && pnpm boxel-homepage:setup"
--- a/packages/boxel-homepage-realm/tests/grid-cards.test.ts
+++ b/packages/boxel-homepage-realm/tests/grid-cards.test.ts
@@ -120,3 +120,60 @@ test('gridCards getter in animated-grid.gts stays @cached', () => {
     'gridCards must delegate to the tested lib/grid-cards module',
   );
 });
+
+test('hero grid animation contracts in animated-grid.gts', () => {
+  const source = readFileSync(
+    fileURLToPath(
+      new URL(
+        '../contents/boxel-ai-website/animated-grid.gts',
+        import.meta.url,
+      ),
+    ),
+    'utf8',
+  );
+  assert.match(
+    source,
+    /captureGridPlane = modifier[\s\S]*?IntersectionObserver/,
+    'animation loop must be gated behind an IntersectionObserver',
+  );
+  assert.match(
+    source,
+    /\} finally \{[\s\S]*?delete activeCard\.dataset\.active;[\s\S]*?delete popup\.dataset\.visible;/,
+    'cancellation must clear the highlighted tile and popup (regression: ' +
+      'stranded data-active tiles accumulated after scroll-out cancels)',
+  );
+  assert.match(
+    source,
+    /for \(const card of cards\) \{\s*delete card\.dataset\.active;/,
+    'each run must sweep stale highlights from prior cancelled runs',
+  );
+  assert.match(
+    source,
+    /resumeSignal/,
+    'pause must park on a promise resolved by resumeAnimation',
+  );
+  assert.doesNotMatch(
+    source,
+    /await timeout\(200\)/,
+    'pause must not busy-poll',
+  );
+  assert.doesNotMatch(
+    source,
+    /classList\./,
+    'JS-driven state uses data attributes, not classnames',
+  );
+});
+
+test('grid tile active state is a data attribute, not a class', () => {
+  const source = readFileSync(
+    fileURLToPath(
+      new URL(
+        '../contents/boxel-ai-website/components/animated-grid-card.gts',
+        import.meta.url,
+      ),
+    ),
+    'utf8',
+  );
+  assert.match(source, /\.hero-mini-card\[data-active\]/);
+  assert.doesNotMatch(source, /\.hero-mini-card\.active/);
+});
--- a/packages/boxel-homepage-realm/tests/grid-cards.test.ts
+++ b/packages/boxel-homepage-realm/tests/grid-cards.test.ts
@@ -1,0 +1,122 @@
+// Regression tests for the AnimatedGrid hero-grid layout logic
+// (contents/boxel-ai-website/lib/grid-cards.ts) and for the caching
+// contract of the gridCards getter in animated-grid.gts.
+//
+// Run with: pnpm test (node --test; Node >= 23 strips types natively)
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+import {
+  buildGridPool,
+  buildGridSlots,
+  makeSeededRandom,
+} from '../contents/boxel-ai-website/lib/grid-cards.ts';
+
+// The production dimensions: 25 cols x 15 rows, 28 noun sample cards.
+const CELLS = 375;
+const SAMPLES = 28;
+
+test('pool fills every cell with a valid sample index', () => {
+  const pool = buildGridPool(CELLS, SAMPLES, makeSeededRandom(42));
+  assert.equal(pool.length, CELLS);
+  for (const idx of pool) {
+    assert.ok(Number.isInteger(idx) && idx >= 0 && idx < SAMPLES);
+  }
+});
+
+test('pool is deterministic for a given seed', () => {
+  const a = buildGridPool(CELLS, SAMPLES, makeSeededRandom(7));
+  const b = buildGridPool(CELLS, SAMPLES, makeSeededRandom(7));
+  assert.deepEqual(a, b);
+});
+
+test('different seeds produce different layouts', () => {
+  const a = buildGridPool(CELLS, SAMPLES, makeSeededRandom(1));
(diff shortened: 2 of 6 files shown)


burieberry:
[Claude Code 🤖] Closing in favor of moving this suite into the boxel-home repo itself. The tests turned out to have zero monorepo dependencies — every import is a Node builtin or a module under contents/ — so relocating them lets contents and tests version together in one repo, and the CI wiring (CS-12317) becomes a single-repo workflow with no cross-repo checkout, token, or ref-pinning concerns.

packages/boxel-homepage-realm/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "scripts": {
+    "test": "pnpm boxel-homepage:setup && node --test 'tests/*.test.ts'",
```
habdelra, line 6: [Claude Code 🤖] **The suite asserts the state of a moving external branch, cloned non-deterministically — worth deciding deliberately before this gets wired into CI.** Non-blocking design concern.

**What the wiring does.** `pnpm test` chains `boxel-homepage:setup`, which runs `scripts/boxel-homepage-setup.sh`. That script does `git clone https://github.com/cardstack/boxel-home.git contents` with no `--branch` and no pinned SHA, and — critically — early-returns (`if [ -d contents ]; then exit 0`) whenever `contents/` already exists. It never pulls. So the content under test is "whatever boxel-home's default branch was the first time this workspace cloned it."

**Three consequences follow:**

1. **Coordination is manual-only.** The integrity tests assert the *post-cleanup* realm state, which (per the description) currently lives on the paired boxel-home content branch, not boxel-home's default branch. Until that content merges to default, a fresh `pnpm test` clones default and the reachability / orphan / token invariants fail. Nothing enforces the ordering — it's a note in the description.

2. **The failure signal lands on the wrong repo.** boxel-home is independently versioned and deployed. Once these run in boxel CI, any future content edit there — a new unlinked instance, a renamed token, a deleted screenshot, a restructured layout file — reddens *boxel* CI on an unrelated PR, while the boxel-home author who made the change gets no pre-merge signal.

3. **Stale `contents/` gives silent false results.** On a dev machine or a cached CI workspace, `contents/` already exists, so `setup` no-ops and the tests run against whatever was cloned earlier. `boxel-homepage:update` (which pulls) exists but `test` doesn't call it.

**Ways out** (pick per intent):
- If these gate *this* repo: pin a known-good boxel-home revision the setup script checks out (`--branch <tag>` or a SHA) so the suite is reproducible and an intentional bump is a visible diff here; consider having `test` run `:update` rather than the no-op `setup`.
- If they really guard *boxel-home* content: the more durable home is boxel-home's own CI, where the signal reaches the author who can act on it.

Flagging now because the CI-wiring follow-up is where this bites; the description already half-acknowledges it ("should land at or after" the content change).

---

packages/boxel-homepage-realm/tests/realm-integrity.test.ts
```diff
@@ -0,0 +1,317 @@
...
+        // in this realm's sources.
+        if (key === 'cardInfo') continue;
+        keys.push(key);
+        collectKeys(value);
+      }
+    };
+    collectKeys(attributes);
+    for (const key of new Set(keys)) {
+      if (!new RegExp(`@field\\s+${key}\\b`).test(allSource)) {
+        failures.push(`${rel(f)}: "${key}"`);
```
habdelra, line 224: [Claude Code 🤖] **This recursion reports false failures for any base/external *compound* field, not just `cardInfo` — verified by running the test.** Latent robustness bug; non-blocking only because whether it bites depends on what the content currently contains.

**Mechanism.** `collectKeys` descends into *every* nested attribute value (line 218) and pushes every key at every depth. The final check (line 223) then requires each collected key to appear as `@field <key>` somewhere in this realm's own sources. `cardInfo` is special-cased out (line 216) precisely because its subtree (`name`, `cardThumbnailURL`, …) is declared in base card-api, not here — but that reasoning applies to *any* field whose type is defined outside the realm. The moment an instance sets a base or external compound field, its subkeys have no local `@field` and the test fails.

**Verified.** I ran this file against a synthetic realm where a locally-adopted card sets a compound field `address: { street, city }` whose subfields are declared outside the realm sources:

```
not ok 4 - instance attribute keys match a declared @field
    attribute keys with no matching @field declaration:
    about.json: "street"
    about.json: "city"
```

The top-level `address` key passes (it's a local `@field`); its subkeys wrongly fail — exactly the `cardInfo` problem, one field over.

**Way out.** Don't blanket-descend. Either stop recursing once you enter a field whose declaring type isn't local (the `cardInfo` skip, generalized), or match each *top-level* attribute key against the adopted card's own declared field set instead of a global regex over all sources. The global `@field <key>` match is loose in the other direction too — a key that matches a *different* card's field passes — so scoping to the card's own fields fixes both the false positive and the false acceptance.

---

packages/boxel-homepage-realm/tests/grid-cards.test.ts
```diff
@@ -0,0 +1,179 @@
...
+  assert.match(
+    source,
+    /@cached\s+private get gridCards\(\)/,
+    'gridCards must keep the @cached decorator',
+  );
+  assert.match(
+    source,
+    /buildGridSlots\(/,
+    'gridCards must delegate to the tested lib/grid-cards module',
+  );
```
habdelra, line 121: [Claude Code 🤖] **These source-contract regexes over-specify the *shape* of code in the external repo, so benign refactors there will redden boxel CI.** Non-blocking design note; ties to the coupling concern on the `test` script.

The intent — guarding behavioral contracts that can't be unit-tested because the component only runs in the host — is reasonable and worth keeping. But several patterns pin incidental syntax, not the contract:

- `/@cached\s+private get gridCards\(\)/` (line 114) fails if someone drops `private`, adds a return-type annotation, or reorders decorators — none of which breaks caching.
- `doesNotMatch(/classList\./)` (line 162) is a blanket ban across the whole file; one unrelated legitimate `classList` read anywhere reddens it.
- The `finally`-block regex (line 141) pins the exact statement order and spelling of the cleanup.

Because `contents/` is a separately-versioned repo (see the note on the `test` script in package.json), each of these is a tripwire a boxel-home author can trip without ever touching boxel — and the red build shows up here, not there. Recommend anchoring on the semantic essential only: e.g. assert `@cached` and `gridCards` appear together, and that no `classList` toggles the *active/visible* state specifically, rather than exact modifiers/ordering/whitespace. Same applies to the analogous contracts in `terminal-scroll.test.ts`.

---

habdelra (commented):
[Claude Code 🤖] **Lens:** this review focuses on the two things a test-only PR lives or dies by — whether the suite is *reproducible* (it exercises code in the separately-versioned boxel-home repo) and whether each assertion actually pins the invariant it claims. I checked out the head and ran `realm-integrity.test.ts` against a synthetic realm to verify its graph logic empirically; the `lib/` unit tests and the `.gts` source-contract tests I could only read, since boxel-home isn't reachable from this session.

**Bottom line: no blocking correctness bug in the test logic — orphan/allowlist detection, the reachability BFS, and link/module resolution all behave as claimed when I exercised them. The substantive items are design-level: the suite's reproducibility is coupled to a moving external default branch, and one assertion (`@field` nesting) has a verified false-positive class. Both are non-blocking; the verdict is COMMENT.**

**What lands right.** The split between pure-logic units (`grid-cards`, `terminal-scroll`, `serial-list`) and structural invariants is the right shape, and the invariants target real, previously-shipped defect classes rather than vanity coverage. The `allowed-orphans.txt` mechanism is clean — single source of truth, self-documenting, fails closed on *new* orphans (I confirmed: an unlisted orphan reddens the reachability test, allowlisting it greens it again). `node --test` with native type-stripping and no build step is a good fit for a leaf package; I confirmed the `.ts` glob and type-stripping run on the Node here, so the pinned Node 24 is safe.

**On the two resolved Codex threads — both fixes hold.** The setup-before-test chaining resolves the fresh-checkout `ERR_MODULE_NOT_FOUND`. The `cardInfo` skip is a *better* fix than the `cardThumbnailURL` allowlist Codex suggested, because it drops the whole free-form subtree rather than enumerating base fields — though see my inline note: the same reasoning needs generalizing beyond `cardInfo`.

1. Decide deliberately whether these invariants gate *this* repo or boxel-home, and either pin a boxel-home revision in the setup script or move the content-invariant tests to boxel-home CI — see the note on the `test` script in `package.json`. *(design, non-blocking)*
2. Generalize the `@field` nesting skip beyond `cardInfo`, or scope the match to the card's own field set — verified false-positive, see the note in `realm-integrity.test.ts`. *(robustness, non-blocking)*
3. Loosen the source-contract regexes to the semantic essentials so boxel-home refactors don't redden boxel CI — see the note in `grid-cards.test.ts`. *(brittleness, non-blocking)*

**Adjacent, out of scope** (noticed while reading; not asked of this PR):
- `boxel-home-layout.gts` is read with `readFileSync` and no `existsSync` guard (navbar-anchors test), so if boxel-home renames or removes that file the test throws an opaque `ENOENT` instead of a clean assertion message. The screenshot test right above it guards its dir; this one doesn't.
- The extensionless-file check allowlists only `LICENSE`; a legitimate `CODEOWNERS`, `Makefile`, or `Dockerfile` landing in boxel-home would fail it.
- The screenshot-reference check uses substring `includes`, so a shorter filename that is a substring of a referenced longer one (`hero.png` vs `super-hero.png`) reads as referenced even when orphaned.

---

PR #5603, opened 2026-07-24
Render BFM embeds across the full format × placement matrix

FadhlanR:
## Summary

Embedding a card or file in a markdown field only rendered correctly for the **default** format+placement combos. Notably `isolated` displayed **nothing** in either placement, inline `embedded` collapsed, and several other combos were unverified.

Root cause: across every render surface, a resolved embed slot was only given a sizing style when `format === 'fitted'`. The `isolated` and inline `embedded` default templates lay out at `width/height: 100%`, which resolves against a shrink-wrapping inline-block (or an unbounded block) and collapses to zero — the instance was in the DOM but invisible.

## Change

- **New shared helper** `bfmResolvedEmbedStyle(format, kind, fittedSizeStyle)` in `packages/runtime-common/bfm-card-references.ts` hands every non-atom embed slot a definite footprint — the resolved instance, the loading shimmer, and the broken-link box all take the same style, so the layout doesn't jump as a slot transitions between states:
  - inline `embedded` → `16rem × 9.375rem`
  - inline `isolated` → `24rem × 18.75rem`
  - block `isolated` → growable `min-height: 18.75rem`
  - `fitted` → its requested dimensions + `overflow: hidden`
- **Applied in all four render surfaces** so footprints stay in lockstep, and the embed-chooser preview (which previously carried its own copy of this logic) now delegates to the helper — no duplicated code:
  - `packages/base/default-templates/markdown.gts` (saved / isolated / embedded-nested)
  - `packages/host/app/components/operator-mode/preview-panel/rendered-markdown.gts` (preview panel)
  - `packages/host/app/lib/codemirror-context.ts` (CodeMirror source editor)
  - `packages/host/app/components/markdown-embed-chooser/preview/index.gts` (chooser preview)
- **`CardWidgetTarget.style` docs** (host `codemirror-context.ts` and the duplicate interface in `packages/base/codemirror-editor.gts`) describe the new contract: fitted dimensions plus `overflow: hidden`, or the shared non-atom footprint; undefined only for atom and block embedded.
- **Plain `MarkdownField`** (a `StringField`) has no relationship to resolve embeds, so its directives stay unresolved by design — documented at the class.

## Matrix now covered

For **card** and **file**, each of `atom` / `embedded` / `fitted` / `isolated`, in both **inline** and **block** placement (16 combinations).

## Tests

- Unit test for `bfmResolvedEmbedStyle`.
- Full card matrix on the saved `MarkdownTemplate` render path + isolated/embedded footprint assertions (`rich-markdown-field-test.gts`).
- Full file matrix in code-mode preview, including block-isolated `min-height` and inline-isolated `width` footprint pins so the file path guards the collapse fix independently (`markdown-file-def-test.gts`).
- Updated/added CodeMirror-editor footprint assertions (`codemirror-editor-test.gts`).

## Verification

`ember-tsc` (glint) + eslint clean on all changed files. Affected suites run locally against the dev stack: `Integration | rendered-markdown` 27/27, `Integration | RichMarkdownField` 28/28, `Acceptance | markdown BFM card references` 11/11, `Integration | codemirror-context` 52/52. Verified live in the dev stack against `experiments/bfm-showcase.md`: block `isolated` renders the full Author card (name/title/quote/portrait) instead of nothing; inline atom, block embedded, all fitted variants (strip / double-wide-strip / tile / compact-card), and file embeds (inline atom + block embedded image) all render correctly, and inline embedded/isolated placeholders occupy the same width as the card they resolve into.

### DIFF
--- a/packages/base/default-templates/markdown.gts
+++ b/packages/base/default-templates/markdown.gts
@@ -10,6 +10,7 @@ import LinkOffIcon from '@cardstack/boxel-icons/link-off';
 
 import {
   bfmRefFormatAndSize,
+  bfmResolvedEmbedStyle,
   buildWaiter,
   cardTypeName,
   fileNameFromUrl,
@@ -271,14 +272,14 @@ export default class MarkDownTemplate extends GlimmerComponent<{
           let format: CardSlotFormat = derived.format;
           let sizeStyle: string | undefined = derived.sizeStyle;
 
-          // Fitted slots carry an inline width/height plus `overflow: hidden`
-          // so the resolved instance occupies the requested footprint.
-          let resolvedStyle: ReturnType<typeof htmlSafe> | undefined;
-          if (format === 'fitted') {
-            resolvedStyle = htmlSafe(
-              sizeStyle ? `${sizeStyle}; overflow: hidden` : 'overflow: hidden',
-            );
-          }
+          // Non-atom resolved slots carry a footprint so the instance occupies
+          // a definite box instead of collapsing (isolated/inline-embedded
+          // default templates lay out at 100%). Fitted uses its requested
+          // dimensions; embedded/isolated get shared defaults. Same helper as
+          // the other render surfaces so footprints stay in lockstep.
+          let resolvedStyleRaw = bfmResolvedEmbedStyle(format, kind, sizeStyle);
+          let resolvedStyle: ReturnType<typeof htmlSafe> | undefined =
+            resolvedStyleRaw ? htmlSafe(resolvedStyleRaw) : undefined;
 
           let resolvedUrl = resolveUrl(rawUrl, baseUrl);
 
--- a/packages/base/default-templates/markdown.gts
+++ b/packages/base/default-templates/markdown.gts
@@ -272,14 +272,18 @@ export default class MarkDownTemplate extends GlimmerComponent<{
           let format: CardSlotFormat = derived.format;
           let sizeStyle: string | undefined = derived.sizeStyle;
 
-          // Non-atom resolved slots carry a footprint so the instance occupies
-          // a definite box instead of collapsing (isolated/inline-embedded
+          // Non-atom slots carry a footprint so the instance occupies a
+          // definite box instead of collapsing (isolated/inline-embedded
           // default templates lay out at 100%). Fitted uses its requested
-          // dimensions; embedded/isolated get shared defaults. Same helper as
-          // the other render surfaces so footprints stay in lockstep.
-          let resolvedStyleRaw = bfmResolvedEmbedStyle(format, kind, sizeStyle);
-          let resolvedStyle: ReturnType<typeof htmlSafe> | undefined =
-            resolvedStyleRaw ? htmlSafe(resolvedStyleRaw) : undefined;
+          // dimensions; embedded/isolated get shared defaults. The same style
+          // goes on the loading shimmer and broken-link box so the layout
+          // doesn't jump as the slot transitions between states, and the same
+          // helper drives the other render surfaces so footprints stay in
+          // lockstep.
+          let styleRaw = bfmResolvedEmbedStyle(format, kind, sizeStyle);
+          let style: ReturnType<typeof htmlSafe> | undefined = styleRaw
+            ? htmlSafe(styleRaw)
+            : undefined;
 
           let resolvedUrl = resolveUrl(rawUrl, baseUrl);
 
@@ -295,7 +299,7 @@ export default class MarkDownTemplate extends GlimmerComponent<{
               state: 'resolved',
               format,
               instance,
-              style: resolvedStyle,
+              style,
             });
             continue;
           }
@@ -304,7 +308,6 @@ export default class MarkDownTemplate extends GlimmerComponent<{
           // linked instances have settled (showFallback), then fall back to the
           // broken-link box. Skipping the broken state on the first modifier
           // run avoids flashing it for refs that will soon resolve.
-          let style = sizeStyle ? htmlSafe(sizeStyle) : undefined;
           if (!showFallback) {
             slots.push({
               element: el,
@@ -991,8 +994,9 @@ export default class MarkDownTemplate extends GlimmerComponent<{
 
         /* Placeholder footprint shared by loading + broken states. The
            default block sizes approximate the eventual card so the layout does
-           not jump when the card resolves; explicit fitted dimensions arrive
-           as an inline style that overrides these. */
+           not jump when the card resolves; the slot's shared inline footprint
+           (fitted dimensions, inline embedded/isolated defaults, block
+           isolated min-height) overrides these when present. */
         .markdown-bfm-loading--embedded,
         .markdown-bfm-broken--embedded {
           width: 100%;
--- a/packages/host/app/lib/codemirror-context.ts
+++ b/packages/host/app/lib/codemirror-context.ts
@@ -48,6 +48,7 @@ import {
   type BfmRefFormat,
   type BfmRefRange,
   bfmRefFormatAndSize,
+  bfmResolvedEmbedStyle,
   extractBfmRefRanges,
   parseBfmSizeSpec,
 } from '@cardstack/runtime-common/bfm-card-references';
@@ -941,14 +942,10 @@ function createCardTargetNotifier(
                 el.getAttribute('data-boxel-bfm-height') ?? undefined,
                 kind === 'inline' ? 'atom' : 'embedded',
               );
-              // Fitted slots carry the width/height plus `overflow: hidden` so
-              // the resolved instance occupies the requested footprint.
-              let style =
-                format === 'fitted'
-                  ? sizeStyle
-                    ? `${sizeStyle}; overflow: hidden`
-                    : 'overflow: hidden'
-                  : undefined;
+              // Non-atom slots carry a footprint so the resolved instance
+              // occupies a definite box instead of collapsing. Same helper as
+              // the saved/preview renderers so footprints stay in lockstep.
+              let style = bfmResolvedEmbedStyle(format, kind, sizeStyle);
               targets.push({
                 element: el as HTMLElement,
                 cardId,
--- a/packages/host/app/lib/codemirror-context.ts
+++ b/packages/host/app/lib/codemirror-context.ts
@@ -63,9 +63,10 @@ export interface CardWidgetTarget {
   // 'card' refs (`:card[URL]`) resolve to CardDef instances; 'file' refs
   // (`:file[URL]`) resolve to FileDef instances.
   refType: 'card' | 'file';
-  // Inline sizing (`width`/`height`, plus `overflow: hidden` for fitted) derived
-  // from the directive's size specifier. Undefined for non-fitted formats.
-  // Mirrors the style the saved/preview markdown renderers apply.
+  // Inline sizing derived from the directive's format — fitted dimensions plus
+  // `overflow: hidden`, or the shared non-atom footprint from
+  // `bfmResolvedEmbedStyle`. Undefined for atom and block embedded. Mirrors
+  // the style the saved/preview markdown renderers apply.
   style?: string;
 }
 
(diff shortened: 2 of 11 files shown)


packages/runtime-common/bfm-card-references.ts
```diff
@@ -237,6 +237,55 @@ export function bfmRefFormatAndSize(
...
+ *
+ * Footprints:
+ *  - `atom`                 — none (intrinsic pill).
+ *  - `fitted`               — the requested dimensions + `overflow: hidden`.
+ *  - inline `embedded`      — 16rem × 9.375rem (matches `--embedded` placeholder height).
+ *  - inline `isolated`      — 24rem × 18.75rem (matches `--isolated` placeholder height).
+ *  - block `embedded`       — none (flows naturally at full width).
+ *  - block `isolated`       — `min-height: 18.75rem`, growable (matches `--isolated` placeholder).
+ */
+export function bfmResolvedEmbedStyle(
```
FadhlanR, line 264: [Claude Code 🤖] **Confirmation — the helper's contract holds at every call site.** Three things verified rather than assumed:

**`fittedSizeStyle` can never carry a stale value for non-fitted formats.** Every caller passes `bfmRefFormatAndSize(...).sizeStyle` straight through, and `bfmRefFormatAndSize` early-returns `{ format }` for anything but `fitted` (the `if (format !== 'fitted')` branch above), so the "ignored for the other formats" clause is never exercised with a live value — the parameter is `undefined` whenever it would be ignored.

**The documented footprints match the placeholder CSS.** `.markdown-bfm-loading--embedded` / `.markdown-bfm-broken--embedded` is `min-height: 9.375rem` and the `--isolated` pair is `min-height: 18.75rem`, in both `packages/base/default-templates/markdown.gts` and its preview-panel twin `rendered-markdown.gts` — the height claims in this doc block are accurate.

**No fifth surface was missed.** The only product-code call sites of `bfmRefFormatAndSize` are the four files this change touches, and a sweep for the old fitted-only `overflow: hidden` conditional turns up nothing outside this module.

Safe-until: this stops being safe if `bfmRefFormatAndSize` ever starts returning a `sizeStyle` for non-fitted formats (e.g. honoring `w:`/`h:` on `embedded`) — at that point each call site's pass-through would silently feed it in here. Non-blocking; nothing to change.

packages/base/default-templates/markdown.gts
```diff
@@ -271,14 +272,14 @@ export default class MarkDownTemplate extends GlimmerComponent<{
...
-            resolvedStyle = htmlSafe(
-              sizeStyle ? `${sizeStyle}; overflow: hidden` : 'overflow: hidden',
-            );
-          }
+          // Non-atom resolved slots carry a footprint so the instance occupies
+          // a definite box instead of collapsing (isolated/inline-embedded
+          // default templates lay out at 100%). Fitted uses its requested
+          // dimensions; embedded/isolated get shared defaults. Same helper as
+          // the other render surfaces so footprints stay in lockstep.
+          let resolvedStyleRaw = bfmResolvedEmbedStyle(format, kind, sizeStyle);
```
FadhlanR: [Claude Code 🤖] **Follow-up (not this PR, non-blocking): the loading/broken placeholders still get a different width than the resolved slot for inline embedded/isolated, so the layout jumps at resolve time.**

**Mechanism.** The resolved slot now carries `width: 16rem`/`24rem` from `bfmResolvedEmbedStyle`, but the loading/unresolved branches below (`let style = sizeStyle ? htmlSafe(sizeStyle) : undefined`) only apply fitted dimensions — inline embedded/isolated placeholders fall through to the CSS classes `.markdown-bfm-loading--embedded` / `--isolated`, which are `width: 100%`. So an inline embedded ref shows a paragraph-wide shimmer that snaps to 16rem when the card resolves — the exact jump the placeholder CSS comment ("so the layout does not jump when the card resolves") exists to prevent. Heights match; widths don't.

**The way out, when someone picks it up.** Route the placeholder style through the same helper (falling back to `sizeStyle` for fitted) here and in the twin `packages/host/app/components/operator-mode/preview-panel/rendered-markdown.gts`. Fitted output would be unchanged (`overflow: hidden` is already on `.markdown-bfm-loading`/`.markdown-bfm-broken`), block embedded/isolated are unchanged (the helper returns `undefined`/`min-height`), so the only visible change is inline placeholders matching the resolved width.

**Classification.** Pre-existing in spirit — before this change the resolved slot collapsed entirely, which was worse — but the change makes the resolved footprint definite, so the placeholder is now the only odd one out.

packages/host/app/lib/codemirror-context.ts
```diff
@@ -941,14 +942,10 @@ function createCardTargetNotifier(
...
-              let style =
-                format === 'fitted'
-                  ? sizeStyle
-                    ? `${sizeStyle}; overflow: hidden`
-                    : 'overflow: hidden'
-                  : undefined;
+              // Non-atom slots carry a footprint so the resolved instance
+              // occupies a definite box instead of collapsing. Same helper as
+              // the saved/preview renderers so footprints stay in lockstep.
+              let style = bfmResolvedEmbedStyle(format, kind, sizeStyle);
```
FadhlanR, line 949: [Claude Code 🤖] **Regression (doc drift; non-blocking but worth fixing here): `CardWidgetTarget.style`'s doc comment now contradicts this line — in both of its homes.**

The interface doc near the top of this file still reads "Undefined for non-fitted formats." This line now sets it for inline `embedded`, inline `isolated`, and block `isolated` too — exactly what the new codemirror-editor tests pin. And the interface has a twin: `packages/base/codemirror-editor.gts` declares its own structurally-identical `CardWidgetTarget` (the base editor only sees the host module through the lazy-load boundary, per the comment there) with the same stale sentence on its `style` field. Both copies need the update, or the next reader of either one trusts the wrong contract.

Suggested wording for both: "Inline sizing derived from the directive's format — fitted dimensions plus `overflow: hidden`, or the shared non-atom footprint from `bfmResolvedEmbedStyle`. Undefined for atom and block embedded."

packages/host/tests/acceptance/markdown-file-def-test.gts
```diff
@@ -472,6 +495,35 @@ module('Acceptance | markdown BFM card references', function (hooks) {
       );
   });
 
+  test('every file format renders in both inline and block placement', async function (assert) {
```
FadhlanR, line 498: [Claude Code 🤖] **Test-coverage note (non-blocking): the `.exists()` matrix here doesn't pin the collapse fix — per the PR description the pre-fix instance was "in the DOM but invisible", so these assertions would have passed before the change too.**

What they *do* pin is format threading: each combo renders the requested `data-test-markdown-*` template rather than falling back, which is real coverage this suite didn't have. The actual collapse regression guard lives in the style assertions elsewhere in the change — `rich-markdown-field-test.gts` asserts the slot's `style` attribute on the card path, and `codemirror-editor-test.gts` asserts `target.style` including the file-ref case — and file refs flow through the same slot code in `MarkdownTemplate.captureCardSlots` as card refs, so the file path is transitively covered.

If you want this test to stand alone, one `hasAttribute('style', /min-height: 18\.75rem/)` on the block-isolated wrapper would do it; otherwise fine as is.

FadhlanR (commented):
[Claude Code 🤖] Review lens: this change's core claim is "one shared footprint helper, applied identically on every render surface" — so the review focused on verifying the helper's contract at each call site, hunting for a missed surface, and checking the claimed CSS mechanics against the actual stylesheets.

**Bottom line: no blocking issues. The fix is correct, the four surfaces are genuinely in lockstep, and the tests pin the new contract. One small thing is worth doing before merge: refresh the now-stale `CardWidgetTarget.style` doc comment in both of its homes.**

**What lands right**
- Centralizing the footprint in `bfmResolvedEmbedStyle` removes the one hand-rolled copy that had already forked (the chooser preview), and no other call site of `bfmRefFormatAndSize` still carries the old fitted-only `overflow: hidden` conditional — I enumerated the call sites; the four touched files are all of them.
- The collapse mechanism the description names is real: `.field-component-card.isolated-format { height: 100% }` in `packages/base/field-component.gts` is what dissolves against a shrink-wrapping inline-block wrapper. Using `min-height` (rather than `height`) for block isolated is the right call — percentage heights don't resolve against a parent's `min-height`, so the card lays out at its natural height and the box stays growable, which matches the live verification described.
- The `MarkdownField` doc comment added in `card-api.gts` is accurate as written: `RichMarkdownField` (`packages/base/rich-markdown.gts`) owns the `linkedCards`/`linkedFiles` `linksToMany` relationships, and plain `MarkdownField` renders `MarkdownTemplate` with no linked args — directives staying unresolved there is structural, not an oversight.

**CI**: the single failing host test on this commit is `Integration | modifier | persist-scroll-position: records the offset only after a genuine user scroll gesture` — a file this diff doesn't touch, and a gesture-timing test; re-run rather than chase. Realm server tests are green.

**Recommendations**
1. Update the stale "Undefined for non-fitted formats" doc on `CardWidgetTarget.style` — in `codemirror-context.ts` *and* its duplicate in `packages/base/codemirror-editor.gts` (see the inline comment in `codemirror-context.ts`).
2. Follow-up, not this PR: route the loading/broken placeholder style through the same helper so inline embedded/isolated placeholders match the resolved width instead of `width: 100%` (see the inline comment in `default-templates/markdown.gts`).
3. Optional: one footprint assertion in the file-matrix acceptance test so it pins the collapse fix independently (see the inline comment there).

**Adjacent, out of scope**: (a) `default-templates/markdown.gts` and the preview panel's `rendered-markdown.gts` remain wholesale twin implementations — pre-existing, and this change keeps them synchronized, but that duplication is where the next drift will happen; (b) the `markdown-bfm-card-slot--fitted` class (and the chooser's `markdown-embed-preview--fitted`) is bound to "has any inline style", so it now also applies to isolated/inline-embedded slots — visually harmless today (it only adds border-radius alongside the new `overflow: hidden`), but the name no longer means "fitted".

packages/host/app/lib/codemirror-context.ts
FadhlanR: [Claude Code 🤖] Resolved in 45fa5b7cbb: both copies of the doc now read "fitted dimensions plus `overflow: hidden`, or the shared non-atom footprint from `bfmResolvedEmbedStyle`; undefined for atom and block embedded" — the host interface here and the duplicate in `packages/base/codemirror-editor.gts`.

packages/base/default-templates/markdown.gts
FadhlanR: [Claude Code 🤖] Taken in this PR rather than deferred (45fa5b7cbb): the loading/broken slots now take the same `bfmResolvedEmbedStyle` output as the resolved slot — the separate `sizeStyle`-only placeholder style is gone in both this file and `rendered-markdown.gts`, so inline embedded/isolated placeholders occupy the resolved width and the resolve-time jump is gone. The placeholder-footprint CSS comments in both files were updated to match. Verified against the tests that pin the old edges: `Integration | rendered-markdown` (27/27 — including the block-embedded `doesNotHaveAttribute('style')` assertion, which still holds because the helper returns `undefined` there) and `Integration | RichMarkdownField` (28/28).

packages/host/tests/acceptance/markdown-file-def-test.gts
FadhlanR: [Claude Code 🤖] Resolved in 45fa5b7cbb: the matrix test now pins the block-isolated slot's `min-height: 18.75rem` and the inline-isolated slot's `width: 24rem` on the file path, so it guards the collapse fix independently of the card-path tests. Passing locally (`Acceptance | markdown BFM card references`, 11/11, 51 assertions).

packages/runtime-common/bfm-card-references.ts
FadhlanR: [Claude Code 🤖] One update since this was written: as of 45fa5b7cbb the helper is no longer resolved-slot-only — the loading shimmer and broken-link box in `default-templates/markdown.gts` / `rendered-markdown.gts` now take its output too, and its doc block was reworded to match. The three verified claims above still hold; the safe-until condition is unchanged (and now also covers the placeholder styles, since they flow through the same pass-through).

lukemelia (approved):

habdelra (approved):

PR #5611, opened 2026-07-27
Wire unpublish into the Workspace edit format

lukemelia:
## What

The hosting section of the Workspace edit format could republish an existing destination but had no way to take one down. Each published site now carries an Unpublish button backed by the registered `unpublish-realm` host tool, invoked with a typed `UnpublishRealmInput` exactly like the existing republish path.

Both mutations report what they actually got back. Neither tool signals failure by rejecting: `UnpublishRealmTool` maps a rejected request onto `UnpublishRealmResult({ status: 'error', error })`, and publish reports per-destination outcomes as `PublishTargetResult`s — so an action that didn't happen is only visible to a caller that inspects the result, and previously nothing did.

- **On success**, the list is re-read through the registered `get-published-realms` tool, which reports the `RealmService` state both mutations write through (`RealmResource.unpublish` deletes the key; `RealmResource.publish` merges fresh timestamps). `meta.realmInfo` remains the initial view — it is a snapshot taken when the card was (de)serialized, and unpublish triggers no reindex, so nothing would otherwise rebuild it in-session and an unpublished site would keep being listed as live until the card was reloaded.
- **On failure**, the row stays and the reason is rendered. This is where re-reading beats optimistic removal: an unpublish the server rejects leaves the list exactly as it was.

Republish and Unpublish also disable each other while either runs. Republish acts on the whole list, so republishing a destination that was mid-unpublish would put the site back up — the same race the host submode's modal closes with `isUnpublishingAnyRealms`.

Two smaller corrections alongside it:

- The local binding for the publish tool is renamed to its post-rename export name, `PublishRealmTool` (the module's default export; `PublishRealmCommand` survives only as a back-compat alias).
- The not-yet-published help text said "Use Publish in the workspace menu". There is no workspace-menu publish item — `data-test-publish-realm-button` lives in the Host submode top bar, so the text now points there.

Hosting state lives in a `TrackedObject` bag (`{ sites, error }`) and the in-flight row in a `TrackedSet`, rather than `@tracked` fields: the edit format is a class *expression*, where decorators are rejected (`TS1206`), which is why that class carries no tracked fields today. The task's own `isRunning` is true for every row at once, so it can't label just the row being unpublished.

## Why this needed verifying, not just reading

`workspace.gts` is currently the only card-side importer of a host tool, so `@cardstack/boxel-host/tools/*` from base-realm card code had no prior traffic. Two things had to hold, and one of them was not obvious:

**Runtime.** These specifiers are not a package export. They resolve because `shimHostTools` registers the tool modules on the loader's virtual network, and `shimExternals` (called by the `network` service) runs that — which is what makes them available in operator mode *and* under prerender.

**Types.** `packages/base/types/@cardstack/boxel-host/index.d.ts` declares the wildcard as:

```ts
declare module '@cardstack/boxel-host/tools/*' { const mod: any; export default mod }
```

Had that ambient declaration won over the tsconfig `paths` mapping, the import would be `any`, the tool construction would be entirely unchecked, and "no `as any` at the call site" would mean nothing. It does not win under the host's project — probing with a deliberate bad member access reports `Property '__nope__' does not exist on type 'typeof PublishRealmTool'`, and likewise for `UnpublishRealmTool`. Both imports carry their real types.

It *does* win elsewhere, which is worth recording: `packages/base` is also compiled by `packages/realm-server`'s project (`lint:types`, part of the Lint job), whose tsconfig has no such `paths` entry. There the tools resolve to the ambient `any`, so the two result callbacks are annotated explicitly rather than left to inference — otherwise `noImplicitAny` fails there while the host project stays clean.

## Test plan

`Acceptance | workspace hosting` (6 tests, 14 assertions): renders a realm whose index card is a `Workspace` in edit format, then pins the hop this change is about — base-realm card code → shimmed tool module → `RealmService` → realm-server — end to end in the app.

- Republish reaches `realm-server.publishRealm` and Unpublish reaches `realm-server.unpublishRealm` with exactly the clicked destination.
- A successful unpublish drops the row, takes Republish with the last destination, reports no error, and falls back to the not-published copy.
- An unpublish the server rejects keeps the row and renders the reported reason.
- A republish leaves the destination listed.
- While an unpublish is in flight, Republish and the row's own button are both disabled — asserted mid-flight against a stub that blocks on a promise the test releases, rather than after settling.

Shard 5 of an earlier run failed on `Integration | modifier | persist-scroll-position: records the offset only after a genuine user scroll gesture`, unrelated to this work. That flake is already fixed on `main` by #5607, which this branch now has merged in; no change to that test is carried here.

Local verification: `ember-tsc --noEmit` clean for both the host and realm-server projects; eslint, ember-template-lint, and prettier clean on the changed files; `Integration | Card | workspace` 19/19.

The realm-server's own `_unpublish-realm` HTTP handler is stubbed at the service boundary, matching the existing host-submode publish tests; it has its own realm-server coverage. A live publish→unpublish round trip wasn't possible in this dev stack — no realm is published locally, and host submode never finished loading for an unpublished dev realm (a CardsGrid-index realm hangs identically, so that is environmental rather than anything to do with the Workspace card).

Note: this touches `packages/base/workspace.gts`, as does #5610, in a different region of the file — whichever lands second needs a trivial rebase.

### DIFF
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -5,14 +5,17 @@ import { htmlSafe, type SafeString } from '@ember/template';
 import { cached, tracked } from '@glimmer/tracking';
 import { restartableTask, timeout } from 'ember-concurrency';
 import { modifier } from 'ember-modifier';
-import { TrackedArray } from 'tracked-built-ins';
+import { TrackedArray, TrackedSet } from 'tracked-built-ins';
 
 import { BoxelInput } from '@cardstack/boxel-ui/components';
 import { eq } from '@cardstack/boxel-ui/helpers';
 import BooleanField from './boolean';
-// host-mode mutation: publish is a registered host tool (tools/index.ts)
-import PublishRealmCommand from '@cardstack/boxel-host/tools/publish-realm';
-import { PublishRealmInput } from './command';
+// host-mode mutation: publish and unpublish are registered host tools
+// (tools/index.ts shims them onto the loader's virtual network, so these
+// specifiers resolve for card code in operator mode and under prerender)
+import PublishRealmTool from '@cardstack/boxel-host/tools/publish-realm';
+import UnpublishRealmTool from '@cardstack/boxel-host/tools/unpublish-realm';
+import { PublishRealmInput, UnpublishRealmInput } from './command';
 
 import LayoutGridPlusIcon from '@cardstack/boxel-icons/layout-grid-plus';
 import Captions from '@cardstack/boxel-icons/captions';
@@ -3403,9 +3406,10 @@ export class Workspace extends CardDef {
       return this.args.model.pinnedSize === 'compact' ? 'compact' : 'regular';
     }
 
-    // Hosting: present published sites (free on realmInfo) and mutate
-    // via the registered publish-realm host command. First-time publish
-    // (domain choice) stays in the workspace menu's publish flow.
+    // Hosting: present published sites (free on realmInfo) and mutate via the
+    // registered publish-realm / unpublish-realm host tools. First-time publish
+    // (domain choice) stays in the host submode's publish flow; this format
+    // acts on destinations that already exist.
     get publishedSites() {
       return publishedSitesOf(this.args.model);
     }
@@ -3425,7 +3429,7 @@ export class Workspace extends CardDef {
       if (!commandContext || !realm || !urls.length) {
         return;
       }
-      await new PublishRealmCommand(commandContext).execute(
+      await new PublishRealmTool(commandContext).execute(
         new PublishRealmInput({
           realmURL: realm,
           publishedRealmURLs: urls,
@@ -3435,6 +3439,44 @@ export class Workspace extends CardDef {
 
     get publishBusy() {
       return this.republishTask.isRunning;
+    }
+
+    unpublish = (publishedRealmURL: string) => () => {
+      this.unpublishTask.perform(publishedRealmURL);
+    };
+
+    // Which site is in flight, so one row can label itself while the others
+    // stay idle — `unpublishTask.isRunning` is true for all of them. A
+    // TrackedSet rather than a `@tracked` field because this is a class
+    // expression, where decorators aren't allowed.
+    private unpublishing = new TrackedSet<string>();
+
+    private unpublishTask = restartableTask(
+      async (publishedRealmURL: string) => {
+        let commandContext = this.args.context?.commandContext;
+        let realm = this.args.model[realmURL]?.href;
+        if (!commandContext || !realm) {
+          return;
+        }
+        this.unpublishing.add(publishedRealmURL);
+        try {
+          await new UnpublishRealmTool(commandContext).execute(
+            new UnpublishRealmInput({
+              realmURL: realm,
+              publishedRealmURL,
+            }),
+          );
+        } finally {
+          this.unpublishing.delete(publishedRealmURL);
+        }
+      },
+    );
+
+    isUnpublishing = (publishedRealmURL: string) =>
+      this.unpublishing.has(publishedRealmURL);
+
+    get unpublishBusy() {
+      return this.unpublishTask.isRunning;
     }
 
     <template>
@@ -3707,6 +3749,17 @@ export class Workspace extends CardDef {
                     {{#if site.when}}
                       <span class='site-when'>{{site.when}}</span>
                     {{/if}}
+                    <button
+                      type='button'
+                      class='publish-btn unpublish-btn'
+                      disabled={{this.unpublishBusy}}
+                      data-test-unpublish-site={{site.url}}
+                      {{on 'click' (this.unpublish site.url)}}
+                    >{{if
+                        (this.isUnpublishing site.url)
+                        'Unpublishing…'
+                        'Unpublish'
+                      }}</button>
                   </div>
                 </div>
               {{/each}}
@@ -3730,7 +3783,7 @@ export class Workspace extends CardDef {
                 <div class='setting-text'>
                   <span class='setting-label'>Publishing</span>
                   <p class='setting-help'>Not published. Use Publish in the
-                    workspace menu to put this space on the web.</p>
+                    Host submode toolbar to put this space on the web.</p>
                 </div>
               </div>
             {{else}}
@@ -3940,6 +3993,18 @@ export class Workspace extends CardDef {
         .publish-btn:disabled {
           opacity: 0.6;
           cursor: default;
+        }
+        /* Sits at the end of the site row, and reads as the destructive
+          counterpart to Republish rather than a second primary action. */
+        .unpublish-btn {
+          margin-left: auto;
+          padding: 5px 10px;
+          font-size: 11.5px;
+          color: var(--grid-ink-quiet);
+        }
+        .unpublish-btn:hover:not(:disabled) {
+          border-color: var(--grid-broken);
+          color: var(--grid-broken);
         }
         .choice-opt:focus-visible,
         .order-move:focus-visible,
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -5,7 +5,7 @@ import { htmlSafe, type SafeString } from '@ember/template';
 import { cached, tracked } from '@glimmer/tracking';
 import { restartableTask, timeout } from 'ember-concurrency';
 import { modifier } from 'ember-modifier';
-import { TrackedArray, TrackedSet } from 'tracked-built-ins';
+import { TrackedArray, TrackedObject, TrackedSet } from 'tracked-built-ins';
 
 import { BoxelInput } from '@cardstack/boxel-ui/components';
 import { eq } from '@cardstack/boxel-ui/helpers';
@@ -13,9 +13,16 @@ import BooleanField from './boolean';
 // host-mode mutation: publish and unpublish are registered host tools
 // (tools/index.ts shims them onto the loader's virtual network, so these
 // specifiers resolve for card code in operator mode and under prerender)
+import GetPublishedRealmsTool from '@cardstack/boxel-host/tools/get-published-realms';
 import PublishRealmTool from '@cardstack/boxel-host/tools/publish-realm';
 import UnpublishRealmTool from '@cardstack/boxel-host/tools/unpublish-realm';
-import { PublishRealmInput, UnpublishRealmInput } from './command';
+import {
+  GetPublishedRealmsInput,
+  PublishRealmInput,
+  UnpublishRealmInput,
+  type PublishedRealmInfo,
+  type PublishTargetResult,
+} from './command';
 
 import LayoutGridPlusIcon from '@cardstack/boxel-icons/layout-grid-plus';
 import Captions from '@cardstack/boxel-icons/captions';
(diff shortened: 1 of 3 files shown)


packages/base/workspace.gts
```diff
@@ -3437,6 +3441,44 @@ export class Workspace extends CardDef {
...
+        try {
+          await new UnpublishRealmTool(commandContext).execute(
+            new UnpublishRealmInput({
+              realmURL: realm,
+              publishedRealmURL,
+            }),
+          );
+        } finally {
+          this.unpublishing.delete(publishedRealmURL);
+        }
```
habdelra, line 3548: [Claude Code 🤖] **Non-blocking follow-up:** after a successful unpublish the row stays put, with no success or error feedback in this format.

**Mechanism.** `publishedSites` → `publishedSitesOf(this.args.model)` reads `model[realmInfo]`, which is `getCardMeta(this, 'realmInfo')` → `card[meta].realmInfo` — a snapshot captured at (de)serialization time, not a live view of `RealmService`. This tool calls `RealmService.unpublish`, which deletes the key from the *resource's* separate `info.lastPublishedAt` copy; it never touches `card[meta]`. Unpublish also does no reindex (per the tool's own comment), so no `index` event fires — and this `Edit` component (`static edit = class Edit`, line 3327) has no `subscribeToRealm` wiring at all (that lives on `Isolated`). So nothing rebuilds `publishedSites` in-session.

**What the user sees.** The button flips `Unpublish → Unpublishing… → Unpublish`, the row stays listed, and clicking again re-hits the server (now a no-op/404 whose error result is swallowed by `run`). The site is actually down server-side, but the format keeps showing it as published until the card is reloaded.

**Contrast with the sibling surface.** `host-submode.gts`/`publish-realm-modal.gts` read unpublish state from `RealmService` (`isUnpublishingRealm`, plus the resource's live `info`), so their list reflects the removal. Base-realm card code can't reach that service — a fair reason for the local `TrackedSet` — but it also means this list can't self-update the way the modal's does.

**Ways out** (pick per scope): optimistically drop the URL from a local override of `publishedSites` on success; or have the tool/`RealmService` surface completion the card can observe; or explicitly document this format as fire-and-reload for v1. Republish shares the no-refresh property, but that's benign there — the set of destinations is unchanged, only the content.

Scope: follow-up, not blocking — but it undercuts the visible effect of the feature, so worth an explicit decision rather than leaving it implicit.

---

packages/base/workspace.gts
```diff
@@ -3707,6 +3749,17 @@ export class Workspace extends CardDef {
                     {{#if site.when}}
                       <span class='site-when'>{{site.when}}</span>
                     {{/if}}
+                    <button
+                      type='button'
+                      class='publish-btn unpublish-btn'
+                      disabled={{this.unpublishBusy}}
```
habdelra: [Claude Code 🤖] **Non-blocking (parity):** publish and unpublish don't cross-disable here. `unpublishBusy` and `publishBusy` each gate only their own task, so during an in-flight unpublish the Republish button (gated on `publishBusy`, still `false`) stays clickable — and Republish acts on `this.publishedSites`, which still includes the destination being removed. A fast click could republish a site mid-unpublish.

The host-submode modal closes this by disabling its publish controls whenever any unpublish is running (`disabled={{this.isUnpublishingAnyRealms}}` in `publish-realm-modal.gts`). Cheap to match: gate Republish on `this.unpublishBusy` as well (and, for symmetry, the unpublish rows on `this.publishBusy`).

For the record, the per-row disable is otherwise correct: `unpublishBusy` disables every unpublish row while one is in flight, so under `restartableTask` only one destination is ever mid-unpublish — the `unpublishing` set exists purely to label that one row, which the code does right.

---

packages/host/tests/acceptance/workspace-hosting-test.gts
```diff
@@ -0,0 +1,148 @@
...
+      unpublishCalls,
+      [publishedRealmURL],
+      'the tool unpublished exactly the clicked destination',
+    );
+  });
+
+  test('Republish reaches the publish-realm tool', async function (assert) {
+    await visitWorkspaceSettings();
+
+    await click('.publish-btn:not(.unpublish-btn)');
```
habdelra: [Claude Code 🤖] **Non-blocking (test quality):** the Republish click keys off `.publish-btn:not(.unpublish-btn)` — a structural CSS selector rather than a `data-test-*` hook. It's unambiguous today (only two `.publish-btn` nodes: the Republish button and the unpublish button, which also carries `.unpublish-btn`), but it breaks silently the moment the settings section grows another `.publish-btn`. The unpublish button already has `data-test-unpublish-site`; a matching `data-test-republish-site`/`data-test-republish` on the Republish button would make this selector as robust as its counterpart.

Separately: these three tests pin the click→tool→service hop well, but none assert what the format shows *after* the action completes — see the `unpublishTask` thread on `workspace.gts` for why the row doesn't currently clear. Once that behavior is settled, an assertion on the post-unpublish list state would pin it.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with two lenses: (1) does the click→shimmed-tool→`RealmService`→realm-server hop this PR set out to wire and pin actually hold, and (2) how the new UI behaves once the action completes.

**Bottom line: the wiring is correct and honestly tested at the hop level, and the typed-import claim holds. No blocking issues. One non-blocking UX gap is worth an explicit decision: after a successful unpublish the row doesn't clear and there's no success/error feedback in this format.**

**What lands right (mechanistically, not as courtesy):**
- The chain is real end to end: `new UnpublishRealmTool(commandContext).execute(new UnpublishRealmInput({ realmURL, publishedRealmURL }))` → `RealmService.unpublish(realmURL, publishedRealmURL)` → `resource.unpublish` → `realmServer.unpublishRealm(publishedRealmURL)`, which is exactly the stub the acceptance test captures. `UnpublishRealmInput` (`realmURL` + `publishedRealmURL`) and the default export `UnpublishRealmTool` (registered by `shimHostTools`) all exist and match.
- The `PublishRealmCommand → PublishRealmTool` rename is complete — the only surviving `PublishRealmCommand`/`UnpublishRealmCommand` references are the deliberate back-compat aliases in the tool modules; no other card-side importer still uses the old binding.
- On the "why this needed verifying" narrative: the typed import genuinely resolves (not `any`), and it's de-risked more than the description implies. `tools/*` is only new as a *spelling* — `packages/base/tsconfig.json` maps both `@cardstack/boxel-host/tools/*` and `@cardstack/boxel-host/commands/*` to `../host/app/tools/*`, and `commands/*` is already imported by many base cards (`card-api.gts`, `menu-items.ts`, `system-card.gts`, `file-menu-items.ts`, `spec.gts`, `realm.gts`, …). So the paths-wins-over-ambient-`any` behavior is the same mechanism already carrying real traffic; this PR just exercises the other spelling of it.
- The help-text fix is right: there is no workspace-menu publish item, and `data-test-publish-realm-button` lives in the host submode toolbar.
- CSS is sound: `--grid-ink-quiet`/`--grid-broken` are in scope, and `.unpublish-btn:hover:not(:disabled)` (0,3,0) correctly wins over `.publish-btn:hover` (0,2,0), so the destructive red hover actually takes effect.

**Recommendations (all non-blocking; detail in the inline threads):**
1. Decide-and-document the post-unpublish behavior — optimistic row removal, a completion signal the card can observe, or an explicit fire-and-reload note for v1. (`unpublishTask` thread on `workspace.gts`.)
2. Cross-disable publish/unpublish while either is in flight, matching the modal's `isUnpublishingAnyRealms` gating, to close the republish-mid-unpublish race. (`disabled={{this.unpublishBusy}}` thread.)
3. Give the Republish button a `data-test-*` hook so the test doesn't rely on `.publish-btn:not(.unpublish-btn)`, and add post-action list assertions once (1) is settled. (Test-file thread.)

**Adjacent, out of scope:** the ambient `declare module '@cardstack/boxel-host/tools/*' { … any }` in `packages/base/types/@cardstack/boxel-host/index.d.ts` is inert here (paths win) — the PR already calls its removal a separate cleanup, and that's the right call. This format also surfaces no tool errors at all (the modal has `handlePublishError`); that's a pre-existing v1 characteristic of the card surface, not something this PR introduced.

---

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Fixed in d2cb3250f5 — and it turned out an observable completion signal *was* available, so it went that way rather than optimistic removal.

Two things the previous version had in hand and discarded:

- `UnpublishRealmTool.run` maps a rejected request onto `new UnpublishRealmResult({ status: 'error', error })` instead of throwing, so `await …execute(…)` resolved successfully whether or not the site came down. The task now reads `result.status`, and on `'error'` keeps the row and reports the message.
- `GetPublishedRealmsTool` reports `RealmService.info(realmURL).lastPublishedAt` — the same live state `RealmResource.unpublish` deletes the key from and `RealmResource.publish` merges fresh timestamps into. That is exactly the state the sibling surface reads, reachable from card code because the tool is already shimmed (`tools/index.ts`, `'get-published-realms'`). So the card can re-read server truth without touching `RealmService` directly.

So after a mutation lands, `reloadPublishedSites` re-reads through that tool and the result becomes the list from then on:

```ts
get publishedSites(): PublishedSite[] {
  return this.hosting.sites ?? publishedSitesOf(this.args.model);
}
```

`meta.realmInfo` stays the initial view (still free, still correct on first render); `hosting.sites` takes over once there is a newer authoritative answer. That beats optimistic removal on the failure path in particular — an unpublish the server rejects leaves the row exactly where it was, which optimism would have removed.

Republish refreshes too, so the `lastPublishedAt` timestamps update rather than staying at their load-time values, and its per-destination `PublishTargetResult` errors surface the same way — previously also swallowed.

`hosting` is a single `TrackedObject` state bag (`{ sites, error }`) for the reason the `TrackedSet` was already there: no decorators in a class expression.

Covered by three new tests: the row disappears on success and the section falls back to its not-published copy; a rejected unpublish keeps the row and renders `Could not unpublish testuser.localhost: destination is locked`; and a republish leaves the destination listed.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Fixed in d2cb3250f5. Both controls now gate on one getter:

```ts
// Either mutation blocks both controls: Republish acts on the whole list,
// so republishing a destination that is mid-unpublish would put the site
// back up. Matches the host submode's modal, which disables its publish
// controls whenever an unpublish is running.
get hostingBusy() {
  return this.publishBusy || this.unpublishBusy;
}
```

Applied to the Republish button and to every unpublish row, so the race closes from both directions — a republish can no longer re-publish a destination whose removal is in flight, and an unpublish can no longer start mid-republish against a list that is about to change.

Pinned by a test that inspects the buttons *while* the request is in flight rather than after settling: `realmServer.unpublishRealm` is stubbed to block on a promise the test holds the resolver for, the click is deliberately not awaited, and both Republish and the row's own button are asserted `isDisabled` before the stub is released.

Your reading of the per-row disable was right, and it survives unchanged: `unpublishBusy` (now folded into `hostingBusy`) disables all rows, `restartableTask` means only one destination is ever mid-flight, and the `unpublishing` set exists only to label that row.

packages/host/tests/acceptance/workspace-hosting-test.gts
lukemelia: [Claude Code 🤖] Both parts done in d2cb3250f5.

The Republish button now carries `data-test-republish`, and the test selects that instead of `.publish-btn:not(.unpublish-btn)` — so it no longer breaks silently if the settings section grows another `.publish-btn`.

Post-action assertions came with it, now that the behavior they would pin is settled (see the `unpublishTask` thread — the list re-reads authoritative state after a mutation instead of going stale):

- **success** — the row for the unpublished destination is gone, Republish goes away with the last destination, no error is reported, and the section falls back to its not-published copy.
- **failure** — a rejected unpublish keeps the row and renders `Could not unpublish testuser.localhost: destination is locked`. This is the case optimistic removal would have gotten wrong, which is why it is asserted separately.
- **republish** — the destination is still listed afterwards.
- **cross-disable** — Republish and the row's button are both disabled while an unpublish is in flight.

The module is 6 tests / 14 assertions, all passing locally, alongside `Integration | Card | workspace` at 19/19.

habdelra (approved):

PR #5614, opened 2026-07-27
Align the Delete Workspace modal with the shared host modal style

FadhlanR:
## What

Bring the whole Delete Workspace confirmation modal in line with the app's other host modals (New/Create-file, Edit-field, Publish-realm, Profile-settings) — structure, style, and how it surfaces errors — instead of the bespoke look it had.

## Details

**Structure**
- Renders through `ModalContainer`'s built-in header (`@title='Delete Workspace'`, standard X close button, standard overlay alignment) rather than hiding that header and drawing a custom title row. The X and overlay-dismiss lock while a delete is in flight (`@isCloseDisabled`).
- Confirm/Cancel are boxel-ui `<Button>`s (`destructive` / `secondary`, `@size='tall'`) with the built-in `@loading` state, replacing the raw buttons and the manual loading spinner.
- Footer is the standard bar with the "not reversible" note on the left and the right-aligned action group (`margin-left: auto`), matching the profile-settings footer pattern. The custom max-width / border-radius / content-padding overrides that fought the wrapper are gone.

**Warning & error — follows the Publish-realm modal**
- The consequences notice is now the publish modal's `.publish-warning` callout pattern: a `warning` box with a `WarningIcon` + body over the `--boxel-warning-200` fill.
- The delete failure renders in a matching `error` box — `--boxel-error-200` border over an 8% tint of the same color — instead of a bare line of red text, so the two modals read as a family.

**Style**
- Typography uses the design-system `font:` shorthand tokens (`var(--boxel-font-sm)` / `var(--boxel-font-xs)`) with the family's 500/600 emphasis weights instead of hardcoded `font-size` + `font-weight: 700`; secondary copy uses the muted `--boxel-450`.
- Spacing and gaps use `--boxel-sp*` tokens throughout (no stray `rem` values).

## Testing

- `eslint` and `lint:types` (glint) clean.
- Existing delete acceptance tests (`workspace-chooser-delete-test`, `workspace-chooser-test`) assert only on the preserved `data-test-*` selectors, so behavior is unchanged.

## Note

The sibling **Archive** and **Duplicate** modals in the same file still use the old hidden-header + custom-header pattern. They're left as-is here; giving them the same treatment for full internal consistency is a good follow-up.

### Before

Screenshot 2026-07-27 at 18 48 26

### After

Screenshot 2026-07-27 at 19 00 32

### DIFF
--- a/packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
+++ b/packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
@@ -885,19 +885,17 @@ export default class Workspace extends Component<Signature> {
         gap: var(--boxel-sp-4xs);
       }
       .delete-modal__workspace-name {
-        font-size: var(--boxel-font-size-sm);
-        font-weight: 700;
+        font: 600 var(--boxel-font-sm);
         color: var(--boxel-dark);
       }
       .delete-modal__workspace-meta {
-        font-size: var(--boxel-font-size-sm);
-        font-weight: 400;
-        color: var(--boxel-dark);
+        font: var(--boxel-font-sm);
+        color: var(--boxel-450);
       }
       .delete-modal__warning-box {
         background: var(--boxel-danger-bg);
         border-radius: var(--boxel-border-radius-lg);
-        padding: var(--boxel-sp-lg);
+        padding: var(--boxel-sp);
         display: flex;
         flex-direction: row;
         align-items: flex-start;
@@ -910,8 +908,7 @@ export default class Workspace extends Component<Signature> {
       }
       .delete-modal__warning-text {
         margin: 0;
-        font-size: var(--boxel-font-size-sm);
-        font-weight: 700;
+        font: 500 var(--boxel-font-sm);
         color: var(--boxel-dark);
         display: flex;
         flex-direction: column;
@@ -920,13 +917,12 @@ export default class Workspace extends Component<Signature> {
       .delete-modal__realms {
         display: flex;
         flex-direction: column;
-        gap: 0.5rem;
+        gap: var(--boxel-sp-xs);
         margin-top: var(--boxel-sp-4xs);
       }
       .delete-modal__realms-title {
         margin: 0;
-        font-size: var(--boxel-font-size-sm);
-        font-weight: 700;
+        font: 600 var(--boxel-font-sm);
         color: var(--boxel-dark);
       }
       .delete-modal__realms-list {
@@ -934,19 +930,17 @@ export default class Workspace extends Component<Signature> {
         padding-left: var(--boxel-sp-lg);
         display: flex;
         flex-direction: column;
-        gap: 0.5rem;
+        gap: var(--boxel-sp-xs);
       }
       .delete-modal__realms-list li {
-        font-size: var(--boxel-font-size-sm);
-        font-weight: 500;
+        font: var(--boxel-font-sm);
         color: var(--boxel-dark);
         list-style: disc;
       }
       .delete-modal__error {
-        color: var(--boxel-danger);
-        font-size: var(--boxel-font-size-sm);
-        font-weight: 600;
         margin: 0;
+        font: 500 var(--boxel-font-sm);
+        color: var(--boxel-danger);
       }
       .delete-modal__footer {
         display: flex;
@@ -962,8 +956,7 @@ export default class Workspace extends Component<Signature> {
         margin-left: auto;
       }
       .delete-modal__disclaimer {
-        font-size: var(--boxel-font-size-xs);
-        font-weight: 700;
+        font: 500 var(--boxel-font-xs);
         color: var(--boxel-danger);
       }
       .workspace-chooser-archive-modal-container > :deep(.boxel-modal__inner) {
--- a/packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
+++ b/packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
@@ -6,7 +6,6 @@ import Component from '@glimmer/component';
 import { cached, tracked } from '@glimmer/tracking';
 
 import ArchiveIcon from '@cardstack/boxel-icons/archive';
-import CircleAlert from '@cardstack/boxel-icons/circle-alert';
 import CopyIcon from '@cardstack/boxel-icons/copy';
 import FileSettingsIcon from '@cardstack/boxel-icons/file-settings';
 import Home from '@cardstack/boxel-icons/home';
@@ -33,6 +32,7 @@ import {
   Lock,
   Star,
   StarFilled,
+  Warning as WarningIcon,
 } from '@cardstack/boxel-ui/icons';
 
 import {
@@ -242,18 +242,19 @@ export default class Workspace extends Component<Signature> {
               </div>
             </div>
 
-            <div class='delete-modal__warning-box'>
-              <CircleAlert class='delete-modal__warning-icon' />
-              <div class='delete-modal__warning-content'>
-                <p class='delete-modal__warning-text'>
-                  <strong>
-                    This permanently deletes the workspace and any custom
-                    domains tied to it.
-                  </strong>
-                  <strong>
-                    Links to cards in this workspace may stop working elsewhere.
-                  </strong>
-                </p>
+            <div class='delete-modal__warning warning'>
+              <WarningIcon
+                class='delete-modal__warning-icon'
+                width='20'
+                height='20'
+                role='presentation'
+              />
+              <div class='delete-modal__warning-body'>
+                <div>
+                  This permanently deletes the workspace and any custom domains
+                  tied to it. Links to cards in this workspace may stop working
+                  elsewhere.
+                </div>
                 {{#if this.hasPublishedRealms}}
                   <div class='delete-modal__realms'>
                     <p class='delete-modal__realms-title'>
@@ -272,7 +273,9 @@ export default class Workspace extends Component<Signature> {
             </div>
 
             {{#if this.deleteError}}
-              <p class='delete-modal__error'>{{this.deleteError}}</p>
+              <div class='delete-modal__warning error' data-test-delete-error>
+                {{this.deleteError}}
+              </div>
             {{/if}}
           </:content>
           <:footer>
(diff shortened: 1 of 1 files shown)


packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
```diff
@@ -246,62 +240,65 @@ export default class Workspace extends Component<Signature> {
...
+                @disabled={{this.deleteWorkspaceTask.isRunning}}
+                {{on 'click' this.closeDeleteModal}}
+                data-test-cancel-delete-button
+              >
+                Cancel
+              </Button>
+              <Button
+                @kind='destructive'
+                @size='tall'
+                @loading={{this.deleteWorkspaceTask.isRunning}}
```
habdelra, line 297: [Claude Code 🤖]

**Verified, and the one item I'd fix in this PR (non-blocking).** Codex flagged this thread too — here's the mechanism and the exact fix.

The destructive `Button` gets `@loading` but not `@disabled`, while Cancel just above correctly gets `@disabled`. In boxel-ui `Button` (`packages/boxel-ui/addon/src/components/button/index.gts`), the native `disabled`/`aria-disabled` attributes come *only* from `@disabled` (lines 88–89); `@loading` only sets `aria-label='loading'` (line 87) and `.loading { pointer-events: none }` (line 221). `pointer-events: none` blocks the mouse but not the keyboard — a focused native `<button>` still activates on Space/Enter — so during a delete the confirm control stays keyboard-activatable and its accessible name flips from "Delete this workspace" to "loading".

There's no data-safety bug behind this: `deleteWorkspaceTask` is a `dropTask` (line 1683), so a re-press mid-flight is dropped. The cost is a11y — the control misrepresents its state to assistive tech and loses its label, and it's inconsistent with Cancel.

Fix — add `@disabled` alongside `@loading`:
```suggestion
                @loading={{this.deleteWorkspaceTask.isRunning}}
                @disabled={{this.deleteWorkspaceTask.isRunning}}
```
The Button CSS is written for exactly this combination: `.boxel-button:disabled:not(.loading)` (line 191) means the dimmed-disabled styling is suppressed while `.loading` is also present, so the button keeps its red fill + spinner and becomes genuinely non-activatable.

Class: regression — the previous code hid both buttons behind the spinner during a delete, so there was no activatable confirm then. Non-blocking.

---

packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
```diff
@@ -822,203 +819,110 @@ export default class Workspace extends Component<Signature> {
...
+        padding: var(--boxel-sp-sm);
       }
       .delete-modal__realm-icon {
-        --boxel-realm-icon-size: 2.625rem;
-        --boxel-realm-icon-border-radius: calc(
-          var(--boxel-border-radius-xs) + 6px
-        );
-        --boxel-realm-icon-background-color: var(--boxel-light);
+        flex-shrink: 0;
+        --boxel-realm-icon-size: 30px;
```
habdelra: [Claude Code 🤖]

**Valid nit (non-blocking) — Codex raised this too.** `--boxel-realm-icon-size: 30px` is the only hardcoded px dimension in the delete-modal CSS. The repo convention (AGENTS.md / the `.gts` styling guide) is rem for dimensional values, and this is specifically a rem→px regression *within this file*: the delete modal used `2.625rem` before this change, and the sibling Archive/Duplicate modals here still use `2.625rem`.

The intent — matching the publish modal's icon scale — is right (`publish-realm-modal.gts:1735` uses `30px`), so keep the size but express it in rem:
```suggestion
        --boxel-realm-icon-size: 1.875rem;
```
(1.875rem = 30px at the 16px root.) If you'd rather the two modals stay byte-identical, the cleaner move is to switch the publish modal to rem as well — but that's out of scope here.

---

packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
```diff
@@ -211,26 +211,20 @@ export default class Workspace extends Component<Signature> {
       </div>
       {{#if this.showDeleteModal}}
         <ModalContainer
-          @title=''
+          @title='Delete Workspace'
```
habdelra, line 214: [Claude Code 🤖]

**Real, but pre-existing and app-wide — not this PR's to fix (non-blocking). Codex raised it here too.** Moving the title into `@title` does drop this modal's only heading: `ModalContainer` renders `<Header @title=…>`, and `Header` (`packages/boxel-ui/addon/src/components/header/index.gts:36`) prints the title as a plain `<div class='title'>` — not a heading, no `id`. The underlying `<dialog>` (`modal/index.gts:59`) has `aria-modal='true'` but no `aria-label`/`aria-labelledby`, so the dialog is also unnamed. Strictly, this modal loses the semantic `<h2>` it used to carry.

I think the trade-off is correct, though: this is the exact shared chrome every other `ModalContainer` modal uses (publish-realm, profile-settings, …), all of which have the same gap — re-adding a bespoke heading here would undo the consistency this PR is going for. The proper fix belongs in `ModalContainer`/`Header` (give the title an `id` + wire `aria-labelledby`, or render it as a heading) so every modal benefits at once.

If you want a named dialog cheaply *without* touching shared components: `ModalContainer` spreads `...attributes` down to `Modal` → the `<dialog>`, so `aria-label='Delete Workspace'` on this `<ModalContainer>` invocation reaches the dialog. Optional.

---

packages/host/app/components/operator-mode/workspace-chooser/workspace.gts
```diff
@@ -822,203 +819,110 @@ export default class Workspace extends Component<Signature> {
...
-        border-radius: var(--boxel-border-radius-lg);
-        padding: var(--boxel-sp-lg);
+      .delete-modal__warning {
         display: flex;
         flex-direction: column;
-        gap: var(--boxel-sp-sm);
+        gap: var(--boxel-sp-xs);
+        padding: var(--boxel-sp-sm);
+        border-radius: var(--boxel-border-radius-lg);
+        font: var(--boxel-font-sm);
```
habdelra: [Claude Code 🤖]

**Minor convention nit (non-blocking).** The `.gts` styling guide (and AGENTS.md) discourage the `font:` shorthand because it also resets `font-family` off the inherited themed family (and resets weight/style). `--boxel-font-sm` expands to `<size>/<line-height> var(--boxel-font-family)` (`packages/boxel-ui/addon/src/styles/variables.css:38`), so the shorthand pins the family instead of inheriting it. In this host-chrome modal the inherited family already *is* `--boxel-font-family`, so there's no visible regression — hence just a nit.

Two things worth noting: (1) it's a step away from the delete modal's previous code, which used the compliant split form (`font-size` + `font-weight`); and (2) the analogous element in the modal you're aligning to — `.publish-warning` in `publish-realm-modal.gts:1634` — uses split `font-size: var(--boxel-font-size-sm)`. Matching that (here and on `__workspace-name` / `__workspace-meta` / `__realms-title` / `__realms-list li`) would be the most consistent:
```
        font-size: var(--boxel-font-size-sm);
        line-height: var(--boxel-line-height-sm);
```
Or, if you keep the shorthand, follow it with `font-family: inherit` the way boxel-ui `Button` does (`button/index.gts:154–155`). This file already uses `font:` in the `.reindex-error__*` styles, so the shorthand is consistent with the file as it stands — low priority either way.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** This is a pure presentation refactor of the Delete Workspace modal, so I reviewed it against the shared host-modal chrome it adopts (`ModalContainer` + boxel-ui `Button` + the publish-realm `.warning`/`.error` callouts) and the repo's `.gts`/CSS conventions — verifying each claim against the checked-out branch and the sibling components rather than reading the diff in isolation.

**Bottom line: no blocking issues.** The refactor is a genuine simplification that lands on the shared primitives correctly, and CI is green (host 3,656 ✅, realm 1,935 ✅). Four non-blocking items below — one worth fixing here (the confirm button's disabled state), three pre-existing/convention nits.

**What lands right (verified):**
- Reuses the shared chrome instead of the bespoke look: `ModalContainer`'s built-in header + `@isCloseDisabled` (which disables both the X *and* overlay-dismiss during a delete — it flows to `Modal`'s overlay `disabled` and the close button), boxel-ui `Button` `destructive`/`secondary` with `@loading`, and the `.warning`/`.error` callout copied faithfully from `publish-realm-modal.gts` — including the exact `rgb(from var(--boxel-error-200) r g b / 8%)` relative-color, so the two modals read as one family. Net removal of a lot of hand-rolled CSS (custom buttons, manual spinner, hidden-header hack).
- Imports are clean: `CircleAlert` is fully removed (0 refs), and `LoadingIndicator` is correctly *retained* — still used by the Archive/Duplicate modals in the same file.
- The re-entrancy model is sound: `deleteWorkspaceTask` is a `dropTask`, `closeDeleteModal` early-returns while it runs, and `@isCloseDisabled` locks the close affordances — a delete in flight can't be closed or double-submitted via the mouse.

**Recommendations (all non-blocking; detail in each inline thread):**
1. Add `@disabled={{this.deleteWorkspaceTask.isRunning}}` to the confirm `Button` — the one I'd fix here. Right now it's keyboard-activatable during the delete and loses its accessible name; `dropTask` prevents an actual double-delete, so this is an a11y-state issue, not data loss.
2. `--boxel-realm-icon-size: 30px` → `1.875rem` — keeps the size while staying rem (and matching the Archive/Duplicate siblings in this file).
3. The dialog has no accessible name and the title is no longer a heading — real, but pre-existing to `ModalContainer`/`Header` and shared by every modal; the right fix is there, not here.
4. `font:` shorthand — minor convention nit; the split-token form matches both the guide and the analogous `.publish-warning` sibling.

**Adjacent, out of scope (for whoever touches this next):**
- The unnamed-dialog / non-heading-title gap lives in `ModalContainer`/`Header` and affects every modal in the app — a fix there fixes all of them at once.
- Neither this `.error` box nor the publish modal's marks the failure `role='alert'`, so a delete failure isn't announced to screen-reader users — app-wide, low-cost follow-up.
- As the PR notes, the Archive/Duplicate modals in this file still use the old hidden-header pattern.

---

habdelra (approved):
a couple nits, but otherwise looks good

PR #5615, opened 2026-07-27
Refresh the realm list when a realm is created outside the session

jurgenwerk:
Realms created or deleted outside the session (boxel-cli, the software factory, another tab) don't show up — or linger — in an open session's workspace chooser until you reload. (Archiving doesn't write `app.boxel.realms` today, so an external archive still needs a reload; wiring that signal is a follow-up.)

Cause: the `app.boxel.realms` matrix event used to both carry the realm list and announce changes. Copying the carried list caused a stale-echo bug on page load, which was fixed by ignoring the event entirely — silently discarding the only live change signal along with it.

Fix: never copy the event's contents, only compare them. Same realms as shown → echo, ignore. Different → re-fetch the authoritative list from the realm server.

```mermaid
flowchart LR
    A["realm changed<br/>outside the session"] --> B["app.boxel.realms<br/>event pushed"]
    B --> C{"list actually<br/>changed?"}
    C -- "no (echo)" --> D["ignore"]
    C -- yes --> E["re-fetch from<br/>realm server"]
    E --> F["chooser updates,<br/>no reload"]
```

One re-fetch per genuine change is sufficient — the server commits permissions before writing the event, so the answer is always fresh. The re-fetch runs as a `restartableTask` so rapid changes coalesce and logout cancels in-flight work. The assembly is renamed `assembleRealmsFromTrustedServers`; the listener documents that the server-side write of `app.boxel.realms` must not be removed without replacing this signal.

- New integration test: external create appears / external delete disappears, no reload; existing coverage keeps the echo case pinned.
- Test helpers: `simulateAccountDataEvent` (mock matrix), `removeRealmPermissions` (realm-server mock).

### DIFF
--- a/packages/host/app/services/matrix-service.ts
+++ b/packages/host/app/services/matrix-service.ts
@@ -8,6 +8,7 @@ import { isTesting } from '@embroider/macros';
 import { cached, tracked } from '@glimmer/tracking';
 
 import {
+  didCancel,
   dropTask,
   rawTimeout,
   restartableTask,
@@ -218,12 +219,6 @@ export default class MatrixService extends Service {
   // gains content at runtime. Login-related side effects (`loginToRealms`,
   // `loadMoreAuthRooms`) still run regardless.
   private trustedRealmServersAuthoritative = false;
-  // The trusted server list backing the authoritative assembly, kept
-  // current by boot and by the realm-servers listener. Needed at
-  // `app.boxel.realms` event time: a realm created mid-session announces
-  // itself only through that legacy event, and re-running the trusted
-  // assembly requires the server list.
-  private trustedRealmServers: string[] = [];
   // Sticky for the lifetime of this instance once a boot assembles from the
   // legacy `app.boxel.realms` list. Keeps later start() calls on the legacy
   // path even after the lazy migration writes `app.boxel.realm-servers`, so
@@ -480,12 +475,14 @@ export default class MatrixService extends Service {
                 // like the realm-servers listener: a reconcile failure must
                 // not crash the app from an async event handler.
                 try {
-                  await this.reconcileRealmsWithAnnouncedList(legacyRealms);
+                  await this.reconcileRealmsTask.perform(legacyRealms);
                 } catch (err) {
-                  console.error(
-                    'Failed to reconcile realms after app.boxel.realms account-data event',
-                    err,
-                  );
+                  if (!didCancel(err)) {
+                    console.error(
+                      'Failed to reconcile realms after app.boxel.realms account-data event',
+                      err,
+                    );
+                  }
                 }
               }
               // Only do this after we've completed our overall login
@@ -510,9 +507,6 @@ export default class MatrixService extends Service {
               }
               let realmServers = e.event.content.realmServers as string[];
               this.trustedRealmServersAuthoritative = realmServers.length > 0;
-              if (this.trustedRealmServersAuthoritative) {
-                this.trustedRealmServers = realmServers;
-              }
               if (this.trustedRealmServersAuthoritative) {
                 // A server-pushed account-data event must not crash the app:
                 // assembly can reject (e.g. fetchUserRealmsFromTrustedServers
@@ -1104,9 +1098,6 @@ export default class MatrixService extends Service {
         // this flag here makes that re-emission a no-op for the available-
         // realms list — the realm-servers path is the authoritative source.
         this.trustedRealmServersAuthoritative = useTrustedServers;
-        if (useTrustedServers) {
-          this.trustedRealmServers = trustedServers;
-        }
         let userRealmURLs: string[];
         if (useTrustedServers) {
           if (isTesting())
@@ -1404,7 +1395,13 @@ export default class MatrixService extends Service {
   //   advertises is dropped on the first pass. If the announced list
   //   still disagrees after the budget, the trusted result stands — the
   //   next signal or boot converges it.
-  async reconcileRealmsWithAnnouncedList(announced: string[]): Promise<void> {
+  //
+  // Restartable on purpose: a newer signal supersedes an in-flight
+  // reconcile (its retry waits are cancelled and the latest announced
+  // list wins), so back-to-back realm creations coalesce instead of
+  // interleaving assemblies — and destroying the service cancels any
+  // pending retry outright.
+  reconcileRealmsTask = restartableTask(async (announced: string[]) => {
     let announcedSet = new Set(announced.map((u) => ensureTrailingSlash(u)));
     let currentSet = () =>
       new Set<string>(
@@ -1420,13 +1417,17 @@ export default class MatrixService extends Service {
     if (setsEqual(announcedSet, currentSet())) {
       return;
     }
-    await this.applyTrustedRealmServersAccountData(this.trustedRealmServers);
+    let trustedServers = await this.getRealmServersFromAccountData();
+    if (trustedServers.length === 0) {
+      return;
+    }
+    await this.applyTrustedRealmServersAccountData(trustedServers);
     for (let delayMs of MatrixService.RECONCILE_CATCHUP_DELAYS_MS) {
       if (missingAnnounced().length === 0) {
         return;
       }
-      await new Promise((resolve) => setTimeout(resolve, delayMs));
-      await this.applyTrustedRealmServersAccountData(this.trustedRealmServers);
+      await timeout(delayMs);
+      await this.applyTrustedRealmServersAccountData(trustedServers);
     }
     let missing = missingAnnounced();
     if (missing.length > 0) {
@@ -1434,7 +1435,7 @@ export default class MatrixService extends Service {
         `Announced realm(s) still absent from the trusted-servers assembly after retries: ${missing.join(', ')} — keeping the trusted result`,
       );
     }
-  }
+  });
 
   // Re-assemble the available-realms list from a runtime
   // `app.boxel.realm-servers` account-data event. Unlike the fail-loud boot
@@ -2218,7 +2219,6 @@ export default class MatrixService extends Service {
     }
     this.setPostLoginCompleted(false, 'resetState');
     this.bootedFromLegacyRealmsList = false;
-    this.trustedRealmServers = [];
     this._isLoadingMoreAIRooms = false;
     this.messagesToSend.clear();
     this.cardsToSend.clear();
--- a/packages/host/app/services/matrix-service.ts
+++ b/packages/host/app/services/matrix-service.ts
@@ -516,7 +516,7 @@ export default class MatrixService extends Service {
                 // start(); here we log and leave the available-realms list as
                 // it was.
                 try {
-                  await this.applyTrustedRealmServersAccountData(realmServers);
+                  await this.assembleRealmsFromTrustedServers(realmServers);
                 } catch (err) {
                   console.error(
                     'Failed to assemble realms from trusted servers in app.boxel.realm-servers account data',
@@ -1409,11 +1409,15 @@ export default class MatrixService extends Service {
     if (trustedServers.length === 0) {
       return;
     }
-    await this.applyTrustedRealmServersAccountData(trustedServers);
+    await this.assembleRealmsFromTrustedServers(trustedServers);
   });
 
-  // Re-assemble the available-realms list from a runtime
-  // `app.boxel.realm-servers` account-data event. Unlike the fail-loud boot
+  // The one procedure that builds the available-realms list: ask each
+  // trusted realm server via `_realm-auth` which realms this user has
+  // (the answer carries the realm URLs AND a JWT per realm), then
+  // replace the user entries and log in. Boot runs this, the
+  // `app.boxel.realm-servers` listener runs it, and the reconcile task
+  // re-runs it when a change signal arrives. Unlike the fail-loud boot
   // assembly, an event-time refresh must be conservative: because
   // `fetchUserRealmsFromTrustedServers` now returns a partial list when a
   // trusted server is unreachable (rather than throwing), replacing the list
@@ -1422,7 +1426,7 @@ export default class MatrixService extends Service {
   // merge (add newly-discovered realms, never remove) and let the retry
   // reconcile; only a fully reachable assembly is authoritative enough to
   // remove realms. Called by the AccountData listener and directly by tests.
-  async applyTrustedRealmServersAccountData(realmServers: string[]) {
+  async assembleRealmsFromTrustedServers(realmServers: string[]) {
     let realmURLs =
       await this.realmServer.fetchUserRealmsFromTrustedServers(realmServers);
     if (this.realmServer.unreachableRealmServers.length > 0) {
--- a/packages/host/app/services/matrix-service.ts
+++ b/packages/host/app/services/matrix-service.ts
@@ -1430,7 +1430,6 @@ export default class MatrixService extends Service {
   reconcileRealmsTask = restartableTask(async (announced: string[]) => {
     let announcedSet = new Set(announced.map((u) => ensureTrailingSlash(u)));
     let previous = this.lastAnnouncedRealms;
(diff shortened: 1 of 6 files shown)


packages/host/app/services/matrix-service.ts
```diff
@@ -1349,6 +1375,43 @@ export default class MatrixService extends Service {
...
+  // cancels any run in progress.
+  reconcileRealmsTask = restartableTask(async (announced: string[]) => {
+    let announcedSet = new Set(announced.map((u) => ensureTrailingSlash(u)));
+    let currentSet = new Set<string>(
+      this.realmServer.userRealmIdentifiers.map((u) => ensureTrailingSlash(u)),
+    );
+    let setsEqual =
+      announcedSet.size === currentSet.size &&
+      [...announcedSet].every((x) => currentSet.has(x));
+    if (setsEqual) {
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 2cb6413 — echo detection now compares the announcement against the previous announcement (seeded from `app.boxel.realms` account data during the trusted-servers boot, cleared on resetState), never against the assembled list. A steady-state divergence between the legacy key and the `_realm-auth` assembly no longer triggers re-fetches; only a changed announcement does.

packages/host/app/services/matrix-service.ts
```diff
@@ -458,6 +460,30 @@ export default class MatrixService extends Service {
...
+                  await this.reconcileRealmsTask.perform(legacyRealms);
+                } catch (err) {
+                  if (!didCancel(err)) {
+                    console.error(
+                      'Failed to reconcile realms after app.boxel.realms account-data event',
+                      err,
+                    );
+                  }
+                }
               }
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 2cb6413 — the trailing `loginToRealms`/`loadMoreAuthRooms` block moved into the legacy (non-authoritative) branch, which is the only path that assembles the list without them. On the trusted path, `assembleRealmsFromTrustedServers` is now the single place login side effects happen: once per genuine change, zero on echoes.

packages/host/tests/integration/matrix-service-boot-assembly-test.ts
```diff
@@ -150,6 +153,102 @@ module(
...
+
+      // …and the reverse: the realm is archived externally. `_realm-auth`
+      // stops advertising it and the server rewrites `app.boxel.realms`
+      // without it. The signal set differs from ours (a removal), so the
+      // assembly re-runs and drops it — no retry needed, the assembly is
+      // authoritative.
+      setRealmArchived(newRealmURL, true);
+      mockMatrixUtils.simulateAccountDataEvent(APP_BOXEL_REALMS_EVENT_TYPE, {
+        realms: [testRealmURL],
+      });
```
habdelra: [Claude Code 🤖] **The archive half of this test asserts a signal that production never emits — it passes only because `simulateAccountDataEvent` manufactures the event.** Blocking for the archive claim (test + description), not for the create feature.

**What I traced.** The comment here says archiving makes the server "rewrite `app.boxel.realms` without it." I followed every archive path and that write does not exist:
- The host archive action (`archiveWorkspaceTask` in `components/operator-mode/workspace-chooser/workspace.gts`) calls `realmServer.archiveRealm()` and `realm.removeRealm()` for local cleanup, but never `matrixService.removeRealmFromAccountData()`. Compare the **delete** action (`deleteWorkspaceTask`, same file) which calls `removeRealmFromAccountData()` right after `deleteRealm()`.
- `realmServer.archiveRealm()` → `POST /_archive-realm` → `handleArchiveRealm` (`realm-server/handlers/handle-archive-realm.ts`) → `archiveRealm(dbAdapter, …)` + `cancelAllJobsInConcurrencyGroup(…)`. `archiveRealm` (`runtime-common/db-queries/realm-metadata-queries.ts`) only sets `realm_metadata.archived_at`. No account-data write on the path.
- The only server-side writer of `app.boxel.realms` is `appendRealmToUserAccountData` (append-only), called solely from `handle-upsert-realm-user-permission.ts`. There is no server code that removes a realm from `app.boxel.realms`.

So in production an externally-archived realm produces no `app.boxel.realms` event, `reconcileRealmsTask` never fires, and the realm stays in the open session's chooser until reload — the exact bug this PR fixes, still unfixed for archive.

**Two things here are genuinely correct and worth keeping:**
- The mock's `setRealmArchived` faithfully models `_realm-auth`: `fetchUserPermissions` (`runtime-common/db-queries/realm-permission-queries.ts`) excludes archived realms by default via its `excludeArchivedSql`, so a reconcile that *did* fire would correctly drop the realm.
- The removal **direction** of the reconcile is real and reachable — via **delete**, not archive: `deleteWorkspaceTask` → `removeRealmFromAccountData` → `setAccountData('app.boxel.realms', smallerList)` → this session's reconcile sees a changed announcement and re-assembles; `_realm-auth` no longer lists the deleted realm (its permission rows are gone), so it drops. The machinery isn't dead — it's just labelled "archive" when the real trigger is delete.

**Way out (pick one):**
- Retarget this removal assertion at delete-from-another-session (a real signal, no production change) and reword the description's "external archive disappears." Lowest risk.
- Or, if archive is meant to propagate live across sessions, wire archive to emit the signal — host-side (`archiveWorkspaceTask` calling `removeRealmFromAccountData`) or server-side (`handleArchiveRealm` writing the user's `app.boxel.realms`) — and keep this test. That's a follow-up, not a one-liner.

**Class:** the create half is verified-correct and is the PR's headline; this archive half tests behavior absent from production.

---

packages/host/app/services/matrix-service.ts
```diff
@@ -2131,6 +2231,7 @@ export default class MatrixService extends Service {
     }
     this.setPostLoginCompleted(false, 'resetState');
     this.bootedFromLegacyRealmsList = false;
+    this.lastAnnouncedRealms = undefined;
```
habdelra, line 2244: [Claude Code 🤖] **`resetState()` clears the echo baseline but doesn't cancel `reconcileRealmsTask`, so an in-flight reconcile can outlive logout and repopulate the just-cleared realm list.** Non-blocking, but a cheap and worthwhile guard.

`MatrixService` is an application-lifetime singleton, so ember-concurrency's owner-destruction cancellation fires only on app teardown, never on logout. `logout()` calls `resetState()`, which sets `lastAnnouncedRealms = undefined` here but leaves any running `reconcileRealmsTask` alone. If the task is mid-flight — awaiting `getRealmServersFromAccountData()`, or `fetchUserRealmsFromTrustedServers` inside `assembleRealmsFromTrustedServers` — when logout runs, it resumes afterward and calls `setAvailableRealmIdentifiers(…)`. That write is **not** guarded by `session.isAuthenticated` (only the trailing `loginToRealms`/`loadMoreAuthRooms` are). Result: the logged-out session's `RealmServerService` gets the previous user's realms written back; and in a fast logout→login-as-different-user sequence the stale run can clobber the new user's list.

The window is one network round-trip and the login side-effects are guarded, so the blast radius is limited — but the stale `setAvailableRealmIdentifiers` write is real, and this task is a new writer this PR introduces, so it's a new post-reset exposure.

Way out — cancel at the session boundary, next to the baseline clear:
```suggestion
    this.reconcileRealmsTask.cancelAll();
    this.lastAnnouncedRealms = undefined;
```

**Class:** regression (new task, not cancelled on reset). Non-blocking.

---

packages/host/app/services/matrix-service.ts
```diff
@@ -1349,8 +1397,60 @@ export default class MatrixService extends Service {
...
+      previous.size === announcedSet.size &&
+      [...announcedSet].every((x) => previous.has(x))
+    ) {
+      return;
+    }
+    let trustedServers = await this.getRealmServersFromAccountData();
+    if (trustedServers.length === 0) {
+      return;
+    }
+    await this.assembleRealmsFromTrustedServers(trustedServers);
```
habdelra, line 1444: [Claude Code 🤖] **`lastAnnouncedRealms` is committed before the assembly runs, so a superseded run can record a change it never applied.** Non-blocking edge case.

The assignment on line 1433 runs synchronously at task start, before the two `await`s below. Because this is a `restartableTask`, a second `app.boxel.realms` event carrying the *same* realm set cancels this run and starts a fresh one; the fresh run reads `previous = this.lastAnnouncedRealms` — now already equal to its own `announcedSet`, because this line set it — takes the echo branch on line 1434, and returns without assembling. The first run was cancelled at its first `await` (before `assembleRealmsFromTrustedServers`), so the change is applied by neither run, and the realm doesn't appear until the next *distinct* event or a reload.

Trigger: the same account-data content delivered as two events within the assembly's network window (`getRealmServersFromAccountData` + `fetchUserRealmsFromTrustedServers`). Matrix re-emits account data on re-sync / gappy sync, so identical redelivery isn't exotic — and this is a feature whose whole point is reliability of the live signal.

Way out — record the baseline only after the assembly commits, so a cancelled run leaves it untouched:
```js
reconcileRealmsTask = restartableTask(async (announced: string[]) => {
  let announcedSet = new Set(announced.map((u) => ensureTrailingSlash(u)));
  let previous = this.lastAnnouncedRealms;
  if (
    previous !== undefined &&
    previous.size === announcedSet.size &&
    [...announcedSet].every((x) => previous.has(x))
  ) {
    return;
  }
  let trustedServers = await this.getRealmServersFromAccountData();
  if (trustedServers.length === 0) {
    return;
  }
  await this.assembleRealmsFromTrustedServers(trustedServers);
  this.lastAnnouncedRealms = announcedSet;
});
```

Echo detection still holds: the boot seed sets the baseline before `startClient()` re-emits, so the initial-sync echo is still caught. Distinct rapid changes are unaffected — the latest run re-fetches the authoritative `_realm-auth` list, which reflects them all. As a bonus this also stops recording a baseline when the assembly throws or `trustedServers` is empty, so those cases retry on the next event instead of being swallowed.

**Class:** regression (new task), non-blocking edge case.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I focused on whether the `app.boxel.realms` "change signal" reliably fires *and* reconciles in **production** for each lifecycle the PR claims (create / archive / delete, from another tab, boxel-cli, or the software factory), plus the task's lifecycle safety (logout, rapid duplicate events). I checked out the head and traced each path through the host `MatrixService`, the realm-server handlers, and the shared permission queries.

**Bottom line.** The headline — a realm *created* outside the session appearing without a reload — is correctly implemented, and I verified the signal path end to end. **One blocking issue, and it is not in the create behavior:** the **archive** half of the new test asserts a removal signal that production never emits, so that assertion passes on a manufactured event. Two smaller, non-blocking robustness gaps (reconcile task not cancelled on logout; echo baseline committed before the assembly runs).

**What lands right.**
- Demoting `app.boxel.realms` from *state* to *change signal* while keeping `_realm-auth` as the sole authority for the list is the right shape — it fixes the stale-echo regression without re-opening it, and the listener comment documents the server-side coupling clearly.
- The echo-vs-previous-announcement fix is correct. I confirmed the assembled list legitimately diverges from the legacy key (`appendRealmToAccountData` merges in `_realm-auth` realms the legacy list never held), so comparing the announcement against the *assembly* — as the earlier revision did — would have re-fetched on every echo. Comparing announcement-to-previous-announcement, seeded at boot before `startClient()` re-emits, is what makes the initial-sync echo a no-op.
- Login side-effects now happen in exactly one place (`assembleRealmsFromTrustedServers`); the earlier double-run is gone.

**The two resolved bot threads check out** against the head: echo detection compares announcement-to-previous-announcement (not to the assembled list), and `loginToRealms`/`loadMoreAuthRooms` run once per genuine change from a single site.

1. Archive emits no `app.boxel.realms` signal in production — retarget the removal assertion at **delete** (a real signal) or wire archive to emit one; correct the description either way. *(Blocking — comment on the test's archive block.)*
2. Cancel `reconcileRealmsTask` in `resetState()` so an in-flight reconcile can't repopulate the realm list after logout. *(Non-blocking — comment on `resetState`.)*
3. Record `lastAnnouncedRealms` after the assembly commits, not at task start, so a superseded identical event can't swallow a change. *(Non-blocking edge case — comment on `reconcileRealmsTask`.)*

**Adjacent, out of scope.** Archive has no live cross-session propagation story on any path today, whereas create and delete do. If live archive propagation is desired product behavior, that's its own change (a server- or host-side `app.boxel.realms` write on archive), not this PR's job.

This is a COMMENT, not an approval or change request — the verdict above is advisory.

---

packages/host/app/services/matrix-service.ts
```diff
@@ -1349,8 +1397,60 @@ export default class MatrixService extends Service {
...
+  // rows `_realm-auth` reads BEFORE it writes `app.boxel.realms` (see
+  // handle-upsert-realm-user-permission), and the event can't arrive
+  // before the write that caused it — so by signal time the enumeration
+  // is consistent. If that ordering is ever violated, the trusted result
+  // stands and the next signal or boot converges it.
+  //
+  // Restartable on purpose: a newer signal supersedes an in-flight
+  // reconcile with the latest announced list, and destroying the service
+  // cancels any run in progress.
+  reconcileRealmsTask = restartableTask(async (announced: string[]) => {
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Fixed in 81f3d89 — `resetState()` now calls `reconcileRealmsTask.cancelAll()` right before clearing the baseline, so an in-flight reconcile is cancelled at the session boundary instead of resuming after logout.

packages/host/app/services/matrix-service.ts
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 81f3d89 — `resetState()` cancels `reconcileRealmsTask` before clearing `lastAnnouncedRealms`, with a comment noting why owner-destruction cancellation is not enough for an app-lifetime service.

packages/host/app/services/matrix-service.ts
jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 81f3d89 — `lastAnnouncedRealms` is now committed only after `assembleRealmsFromTrustedServers` lands, so a superseded restartable run leaves the baseline untouched and identical redelivery can no longer swallow an unapplied change. As a side effect, a thrown assembly or an empty trusted-servers list no longer records a baseline either, so those retry on the next signal.

packages/host/tests/integration/matrix-service-boot-assembly-test.ts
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Fixed in 81f3d89 by retargeting the removal half at the real production signal: external **delete**, where the permission rows are removed (so `_realm-auth` stops advertising the realm) and the deleting session rewrites `app.boxel.realms` without it via `removeRealmFromAccountData`. The test now models exactly that (`removeRealmPermissions` replaces `setRealmArchived`, which is deleted). Making archive itself emit the signal is a production change (neither the archive handler nor the host archive action touches `app.boxel.realms` today) and is left as a follow-up.

jurgenwerk: [Claude Code 🤖] Confirmed and fixed in 81f3d89, taking the lowest-risk way out: the removal assertion is retargeted at delete-from-another-session — permission rows dropped from the mock (`removeRealmPermissions`, replacing the now-deleted `setRealmArchived`) plus the shrunken `app.boxel.realms` announcement that `removeRealmFromAccountData` produces in production. The test name and comments now say delete, and the PR description is reworded to match. Wiring archive to emit the signal remains a follow-up if live cross-session archive propagation is wanted.

habdelra (approved):

richardhjtan (approved):

jurgenwerk:
Closing in favor of https://github.com/cardstack/boxel/pull/5638

PR #5617, opened 2026-07-27
Run submission linter as lint-source worker jobs

richardhjtan:
## What this does

The listing submission flow lints the collected files before any PrCard or GitHub activity. The bot-runner publishes one `lint-source` queue job per lintable file (`.js` `.ts` `.gjs` `.gts` `.hbs`), sharded 10 ways at user-initiated priority. The jobs execute in the worker with the same cached engines that serve the realm `_lint` endpoint (ESLint autofix → Prettier → ember-template-lint), so no lint tooling runs inside the bot-runner process — running those engines in-process there is memory-unsafe, which is why the worker owns this execution path.

Gate semantics:

- Error-severity findings **block** the workflow: the card is patched with `lintStatus: 'failed'` and the per-file findings in `lintErrors`; `prCreationError` carries only a one-line count so the findings aren't duplicated on the card.
- Auto-fixed output (import injection, formatting) is what lands on the PrCard and the GitHub branch; `lintFixedCount` records how many files changed.
- Retry clears the previous attempt's `prCreationError`/`failedStep` up front so an active attempt never displays a stale failure; each step re-records its own error if the attempt fails again.

The spawn-based submission linter (`runtime-common/lint/submission-lint.ts`, temp-dir management, child `ember-template-lint`/`ember-tsc` processes) is deleted; `SubmissionFile`/`LintOutcome` live next to the new fan-out in `packages/bot-runner/lib/pr-listing/lint-submission-files.ts`.

## **Unhappy path (unfixable lint error)**

https://github.com/user-attachments/assets/91768cb3-dbbd-4ca7-a2a3-da70b34089cb

## **Happy path (with/without fixable file & lint passed)**

https://github.com/user-attachments/assets/c1f0f694-f8ac-419b-ba15-4e8a7cf93816

### Why the retry-path change

`fetch-card-json` takes its URL as `cardIdentifier`. The workflow handler passed `url`, so the retry path's first step — reloading the workflow card to recover its context — fetched `undefined` and 404'd, failing every retry before it reached any step. It now passes `cardIdentifier`.

### Why a base-realm file is in a bot-runner PR

`collect-submission-files` gathers more than the listing folder — it also vendors the source of the modules the listing's Spec cards adopt from, so every spec-bearing submission carries `base/spec.gts` in its file set. The lint gate runs the host ESLint config over each collected file, and `getter-return` is error severity there. Both `defaultIcon` getters return a value on one branch and bare-`return` on the no-model branch, which is exactly what that rule flags — two error-severity findings in framework code the submitter cannot edit, blocking every spec-bearing submission. Returning `undefined` explicitly is behavior-identical (a bare `return` already yields `undefined`) and clears the gate. Nothing else applies `getter-return` to these files: the base package's own lint is ember-template-lint plus a custom field-option check, with no ESLint pass.

## Testing

- Unit: job fan-out payloads (type/priority/timeout/shard keys), extension filtering, fixed-output aggregation, error formatting, rejected-job handling; workflow paths for pass (fixed files forwarded to `create-pr-card`), blocking failure (`failedStep: 'lint'`, no PrCard/GitHub jobs), lint crash, and retry clearing stale errors.
- Verified end-to-end against a local dev stack. A listing carrying an unfixable error runs the lint jobs in the worker, blocks with the per-file findings rendered on the workflow card, and creates no branch or PR. Clearing the errors and retrying re-lints the fresh file contents, reports `lintStatus: passed` with `lintFixedCount: 3`, and proceeds through PrCard creation to the GitHub step.

## Deferred

- Lint-rule parity: the worker's config (host `.eslintrc` + extra `@cardstack/boxel` rules + ember-template-lint `recommended`) is stricter than the catalog repo's CI, so some listings that pass CI fail this gate (e.g. `<style scoped>` blocks trip `no-whitespace-for-layout`). Rule alignment, and whether vendored base-realm files should be linted as part of a submission at all, are tracked internally.
- Typecheck (`ember-tsc`) is not part of the gate; the catalog repo's CI remains the typecheck gate.


packages/bot-runner/lib/pr-listing/lint-submission-files.ts
```diff
@@ -0,0 +1,130 @@
...
+
+async function lintFileViaJob(
+  queuePublisher: QueuePublisher,
+  file: SubmissionFile,
+  index: number,
+): Promise<LintResult> {
+  let job = await queuePublisher.publish<LintResult>({
+    jobType: 'lint-source',
+    // Same 10-bucket spread the realm _lint endpoint uses, so concurrent
+    // submissions can lint in parallel without flooding the worker pool.
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Reworded — the comment now describes the group as a 10-way shard keyed by file index, mirroring the endpoint's spread without implying an identical key.

packages/bot-runner/tests/lint-submission-files-test.ts
```diff
@@ -0,0 +1,166 @@
...
+        concurrencyGroup: 'lint:submission:0',
+        timeout: 30,
+        priority: 10,
+        args: {
+          source: 'let a = 1',
+          filename: 'catalog/MyListing/component.gts',
+        },
+      },
+      'publishes the same payload shape the realm _lint endpoint uses',
+    );
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Assertion message now says "publishes the lint-source payload the worker task consumes" instead of claiming endpoint parity.

packages/bot-runner/lib/pr-listing/pr-listing-workflow-handler.ts
```diff
@@ -313,20 +316,58 @@ export class PrListingWorkflowHandler implements BotCommandHandler {
...
+        listingId: ctx.listingId,
+      });
+    } catch (lintError: any) {
+      let message = lintError?.message ?? 'lint runner crashed';
+      log.error('pr-listing-create: lint runner crashed', { error: message });
+      await this.patchWorkflowCard(ctx, {
+        attributes: { lintStatus: 'failed', lintErrors: [message] },
+      });
+      throw new Error(message);
+    }
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] The rethrow now carries `{ cause: lintError }`, so the original stack survives into logs and Sentry.

packages/base/spec.gts
```diff
@@ -131,7 +131,7 @@ interface SpecHeaderSignature {
 export class SpecHeader extends GlimmerComponent<SpecHeaderSignature> {
   get defaultIcon() {
     if (!this.args.model) {
-      return;
+      return undefined;
```
richardhjtan, line 134: [Claude Code 🤖] Why a base-realm file is in a bot-runner PR:

`collect-submission-files` gathers more than the listing folder — it also vendors the source of the modules the listing's Spec cards adopt from, so every spec-bearing submission carries `base/spec.gts` in its file set. The lint gate this PR adds runs the host ESLint config over each collected file, and `getter-return` is error severity there. Both `defaultIcon` getters return a value on one branch and bare-`return` on the no-model branch, which is exactly what that rule flags — two error-severity findings in framework code the submitter cannot edit, blocking every spec-bearing submission.

Returning `undefined` explicitly is behavior-identical (a bare `return` already yields `undefined`) and clears the gate. Nothing else applies `getter-return` to these files: the base package's own lint is ember-template-lint plus a custom field-option check, with no ESLint pass.

The broader question — whether vendored base-realm files should be linted as part of a submission at all, and how the worker's rule set should line up with the catalog repo's CI — is the rule-parity item under Deferred.

packages/bot-runner/lib/pr-listing/pr-listing-workflow-handler.ts
```diff
@@ -313,20 +316,58 @@ export class PrListingWorkflowHandler implements BotCommandHandler {
...
     }
+
+    log.info('pr-listing-create: lint passed', {
+      fixedFileCount: outcome.fixedFileCount,
+    });
+    await this.patchWorkflowCard(ctx, {
+      attributes: {
+        lintStatus: 'passed',
+        lintFixedCount: outcome.fixedFileCount,
+      },
```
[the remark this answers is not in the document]
richardhjtan: [Claude Code 🤖] Valid, and reproduced on a real run before fixing — a workflow card from local end-to-end testing ended up with `lintStatus: "passed"`, `lintFixedCount: 3`, and two `lintErrors` entries left over from the attempt before it.

Not visible in the UI today: the card guards both the error list and the count on `lintStatus === "failed"`, and every path that sets `failed` writes its own findings in the same patch. So the damage is the persisted state being self-contradictory for anything that reads the card without replicating that guard, and an implicit coupling that a future template change could break.

Fixed by taking your parenthetical as the primary fix: the `in-progress` patch now clears `lintErrors`, so each run starts from a clean slate and every terminal state below it writes only the findings it produced. The pass patch clears them too, so it stands on its own rather than relying on the earlier patch having landed. Test assertions updated to pin both.

packages/bot-runner/lib/pr-listing/lint-submission-files.ts
```diff
@@ -0,0 +1,131 @@
...
+  lintErrors: string[];
+  fixedFileCount: number;
+}
+
+const log = logger('bot-runner');
+
+// Union of the lint-source task's ESLINT_EXTENSIONS and
+// TEMPLATE_LINT_EXTENSIONS (runtime-common/tasks/lint.ts); other files pass
+// through the lint step untouched.
+const LINTABLE_EXTENSIONS = new Set(['.js', '.ts', '.gjs', '.gts', '.hbs']);
```
habdelra: [Claude Code 🤖] **Non-blocking (follow-up): the lintable-extension list is a hand-copied twin of the task's own routing sets, and nothing keeps them in sync.**

`LINTABLE_EXTENSIONS` here restates the union of `ESLINT_EXTENSIONS` (`.js .ts .gjs .gts`) and `TEMPLATE_LINT_EXTENSIONS` (`.hbs .gts .gjs`) in `runtime-common/tasks/lint.ts`. Both of those are module-private `const`s (not exported), so this set can't derive from them and instead duplicates them by hand — the comment above documents the coupling but doesn't enforce it.

The failure mode is one-directional and silent: if the task later routes a new extension to an engine, a submission file of that type fails `isLintable()` here, skips the fan-out entirely, and lands on the PrCard/branch without ever hitting the gate — even though the same file *is* linted at the realm `_lint` endpoint. The reverse (sending an extra extension the task doesn't recognize) is harmless: `lintOne` no-ops on an unrouted extension and returns `passed`. So only the missing-extension direction bites.

Suggest exporting the two sets from `tasks/lint.ts` and deriving the union here (`new Set([...ESLINT_EXTENSIONS, ...TEMPLATE_LINT_EXTENSIONS])`) so gate coverage tracks engine routing automatically. Not this PR's to fix — worth filing.

---

packages/bot-runner/lib/pr-listing/lint-submission-files.ts
```diff
@@ -0,0 +1,131 @@
...
+  queuePublisher: QueuePublisher,
+  file: SubmissionFile,
+  index: number,
+): Promise<LintResult> {
+  let job = await queuePublisher.publish<LintResult>({
+    jobType: 'lint-source',
+    // 10-way shard (mirroring the realm _lint endpoint's spread, keyed by
+    // file index here) so concurrent submissions can lint in parallel
+    // without flooding the worker pool.
+    concurrencyGroup: `lint:submission:${index % 10}`,
```
habdelra, line 103: [Claude Code 🤖] **Confirmation (non-blocking): the shard key is correct, and the two properties it rests on are worth stating for the next editor.**

- **Disjoint namespace from interactive lint.** The realm `_lint` endpoint keys on `lint:${this.url}:<digit>` (`runtime-common/realm.ts`), so `lint:submission:<n>` never shares a group with an editor-driven lint. A large submission therefore can't serialize behind — or starve — someone's live editor linting, and vice-versa. Good call keeping them separate.
- **The group is a hard serialization key, not a coalescing hint.** The worker's claim query only picks a job whose `concurrency_group NOT IN (active_concurrency_groups)` (`postgres/pg-queue.ts`) — at most one job per group runs at a time across the whole pool. With 10 groups that caps *system-wide* submission-lint concurrency at 10, which is exactly the "without flooding the worker pool" property the comment claims.

One implication the comment doesn't spell out: because the prefix is a global constant, two concurrent submissions contend pairwise on the same 10 buckets (both files at index 0 → `lint:submission:0`), so 10 is a global ceiling, not per-submission. That's the intended load cap and it's safe here because `lint-source` registers no coalesce handler (`tasks/lint.ts` never calls `registerQueueJobDefinition` with one), so two distinct sources sharing a group never collapse into one job — they serialize, each getting its own result. Nothing to change.

---

packages/bot-runner/lib/pr-listing/lint-submission-files.ts
```diff
@@ -0,0 +1,131 @@
...
+      `lint-source job failed for ${file.filename}: ${
+        err instanceof Error ? err.message : String(err)
+      }`,
+      { cause: err },
+    );
+  }
+}
+
+function formatLintError(filename: string, message: LintMessage): string {
+  let rule = message.ruleId ?? message.source ?? 'lint';
```
habdelra: [Claude Code 🤖] **Nit (non-blocking): `message.source` is the engine, not a rule id, so this fallback prints `[eslint]`/`[template-lint]` in the rule slot.**

`LintMessage.source` is typed `'eslint' | 'template-lint'` (`tasks/lint.ts`) — it names which engine produced the finding, not a rule. When `ruleId` is null but `source` is set (a rule-less finding that still came from an engine), the output reads e.g. `foo.gts:1:1 [eslint] …`, which looks like a rule literally named `eslint`. Only the validate/parse-error paths, which set neither field, fall through to `'lint'`. Minor cosmetic; if you want the distinction, drop `?? message.source` and let it fall straight to `'lint'`.

---

packages/bot-runner/lib/pr-listing/pr-listing-workflow-handler.ts
```diff
@@ -377,7 +421,7 @@ export class PrListingWorkflowHandler implements BotCommandHandler {
       runAs: this.submissionBotUserId,
       realmURL: ctx.submissionRealm,
       command: FETCH_CARD_JSON_COMMAND,
-      commandInput: { url: ctx.existingPrCardUrl },
+      commandInput: { cardIdentifier: ctx.existingPrCardUrl },
```
habdelra, line 424: [Claude Code 🤖] **Confirmation (non-blocking): the `url` → `cardIdentifier` switch is the right fix, and it explains why the retry path failed at its first step.**

Verified against the command: `fetch-card-json` declares `requireInputFields = ['cardIdentifier']` and reads `input.cardIdentifier` (`host/app/tools/fetch-card-json.ts`). Passing `url` left `cardIdentifier` unset, so every fetch on the retry path — starting with `buildRetryContext` reloading the workflow card — resolved against `undefined`. Both call sites (`loadExistingPrCard` here and `fetchCardDocument`) now pass `cardIdentifier`; consistent and correct.

Adjacent, not this PR: the retry test's queue mock branches on `commandInput.cardIdentifier`, but it's a hand-rolled stand-in that echoes whatever field the workflow sends — it can't detect a mismatch between the field the workflow passes and the field the real command requires. That fidelity gap is exactly what let the `url` spelling satisfy the test while failing against the real command. A lightweight guard (assert the workflow sends the command's declared `requireInputFields`) would close that class of blind spot.

---

habdelra (commented):
[Claude Code 🤖] ## Review: submission lint moved to worker `lint-source` jobs

Focused on the execution-path move (bot-runner → worker jobs), the gate/retry semantics wired onto the workflow card, and whether deleting the spawn-based linter left anything orphaned.

**Bottom line: solid, no blocking issues.** The fan-out is correct, the gate semantics are coherent, the spawn-based linter is removed cleanly (no remaining importers of `submission-lint.ts`, `cleanupOrphanedSubmissionTemps`, or the temp-dir exports; `main.ts` startup cleanup and the `.gitignore` entry go with it), and every earlier bot finding is addressed in-thread. Tests pin the behavior they claim — fan-out payloads, extension filtering, fixed-output forwarding, error formatting, the blocking path, the crash path, and stale-error clearing.

**What lands right:**
- **Worker owns execution.** Fanning each file out as a `lint-source` job reuses the worker's cached ESLint/Prettier/template-lint engines and keeps the memory-heavy tooling out of the bot-runner — the stated motivation, and the right one.
- **Namespace separation.** `lint:submission:*` is disjoint from the realm endpoint's `lint:<realmUrl>:*`, so submission linting and interactive editor linting never contend. Detail in the thread on `concurrencyGroup`.
- **Error cause preserved** through the rethrow (`{ cause }`), and the blocking path keeps full findings on `lintErrors` while `prCreationError` carries only a count — no duplication on the card.
- **Retry correctness.** The `cardIdentifier` fix repairs a retry path that failed at its first step; the up-front stale-error clear plus the `lintErrors: []` reset at each run start keep the card from reporting `passed` alongside a prior attempt's findings.

**One thing I couldn't verify from the tree:**
- **`lintStatus: 'in-progress'` is a new value** this PR introduces (the code it replaces only ever wrote `passed`). The `SubmissionWorkflowCard` schema and its templates live in catalog realm content, not in this repo, so I can't confirm the field accepts `in-progress` and renders sensibly for it. The end-to-end verification in the description covers this in practice — just flagging that the guarantee rests on that manual check, not on anything in the diff or tests.

1. Derive `LINTABLE_EXTENSIONS` from the task's own routing sets rather than hand-copying them — thread on `lint-submission-files.ts` line 28.
2. `formatLintError` prints the engine name as a rule when `ruleId` is null — thread on line 129.
3. Consider a contract check so the retry test can't echo a wrong input-field spelling — thread on `pr-listing-workflow-handler.ts` line 424.

**Adjacent, out of scope (already under Deferred):** the worker's rule set (host `.eslintrc` + extra `@cardstack/boxel` rules + template-lint `recommended`) is stricter than the catalog repo's CI, so listings that pass CI can be blocked by this gate. That's a real submitter-facing surprise; when it's picked up, a block message that reads as "stricter than CI" (rather than a generic lint failure) would help. `base/spec.gts` is the concrete instance — vendored framework files the submitter can't edit getting linted — and is correct as a point fix, since `return;` and `return undefined;` are behavior-identical.

---

packages/bot-runner/lib/pr-listing/lint-submission-files.ts
richardhjtan: [Claude Code 🤖] Agreed on the analysis, and the asymmetry you point out is the reason I fixed it here rather than filing it: the dangerous direction fails open — a newly-routed extension silently bypasses the gate and lands on the branch unlinted — and a gate that can silently stop covering a file type is worth closing now.

Fixed slightly differently than suggested. Rather than exporting both sets and re-deriving the union at this call site (which still leaves the union logic duplicated, just one layer down), `tasks/lint.ts` now exports `isLintableFilename(filename)` and this module asks it. The routing decision stays co-located with the routing it governs, and `LINTABLE_EXTENSIONS` here is gone.

One constraint worth recording for anyone reaching for this from elsewhere: it has to be imported from the node path (`@cardstack/runtime-common/tasks/lint`), not the `tasks` barrel — the barrel’s value exports resolve through `#lint-task`, which swaps to `lint-browser.ts` in browser builds, and that variant exports only a throwing `lintSource`.

packages/bot-runner/lib/pr-listing/lint-submission-files.ts
richardhjtan: [Claude Code 🤖] Thanks — the global-vs-per-submission distinction is the part worth having written down, and your reading matches what the queue does: the group is a hard serialization key, so 10 buckets cap submission linting at 10 concurrent jobs pool-wide, not per submission.

Two things that make that ceiling the right trade here rather than a bottleneck. The jobs are short and independent, so pairwise contention between two submissions costs queueing latency, never correctness. And the disjoint prefix is doing load-bearing work in both directions: submission linting cannot starve interactive `_lint`, and a listing large enough to saturate its buckets cannot push an editor’s lint behind it.

Also confirming your coalescing note, since it is the property that makes sharing a group safe at all: `lint-source` registers no coalesce handler, and `tasks/lint.ts` documents why — each enqueue carries its own `source`, so collapsing two would hand one caller the other’s output. Two sources in the same group serialize and each gets its own result. No change made.

packages/bot-runner/lib/pr-listing/lint-submission-files.ts
richardhjtan: [Claude Code 🤖] Taken — `?? message.source` is gone, so a rule-less finding now reads `[lint]` and the bracket only ever holds an actual rule id.

Nothing is lost by dropping it: the paths that produce a null `ruleId` alongside a set `source` are the engine-failure ones, and their message text already names the engine (`ESLint failed to process source: …`), so the engine was never carried solely by that slot. Added a case to the formatting test pinning `[lint]` for that shape.

packages/bot-runner/lib/pr-listing/pr-listing-workflow-handler.ts
richardhjtan: [Claude Code 🤖] Thanks for verifying it against the command declaration.

On the fidelity gap — worth noting the retry test does now catch this specific regression, though by accident of shape rather than by design: the mock keys its workflow-card and PrCard branches off `commandInput.cardIdentifier`, so reverting the workflow to `url` leaves both branches unmatched, the fetches fall through to the generic patch response, and the retry assertions fail. What it cannot catch is the general case you describe — the mock still echoes whatever field the workflow sends, so it validates internal consistency, not conformance to the command contract.

Agreed that a guard asserting the workflow supplies each command’s declared `requireInputFields` is the right shape for closing that class. It needs the declarations to be reachable from a bot-runner test, and these tools live in the host app behind ember service injection, so it is a follow-up rather than an addition here. Filing it alongside the other deferred items.

habdelra (approved):

PR #5619, opened 2026-07-27
Add Workspace indexing/prerender safety regression test

lukemelia:
## What

Adds integration coverage that locks in the defensive patterns keeping the `Workspace` isolated render safe when the indexer/prerender renders a realm's index card. Production's prerender renders the isolated shell through the full Ember lifecycle (constructor + modifiers + `afterRender` settle), so these properties must hold at index time:

- **The Activity feed stays lazy.** `loadFeed` runs only under the Activity segment; the default segment is Home, so the whole-realm reverse-chron search never runs during indexing.
- **`linksToMany` entry points resolve at render time.** `.constructor` is never touched on an unresolved link — the `entryPointsFilter` getter defers it behind a lazy `get cards()`, and the Home `doorKind` reads it only once links have resolved.

## How it's tested

`packages/host/tests/integration/workspace-indexing-test.gts` indexes a `Workspace` **off the index path** (`ws.json`), because the in-browser indexer substitutes boilerplate HTML for a realm's default index card and would otherwise skip the Glimmer render entirely. Off the index path, the real isolated render runs through the indexer — the same path production's puppeteer prerender takes for the index card itself.

It then asserts on the indexed `isolatedHtml`:

- the shell renders (no crash on entry-point resolution),
- both pinned entry points resolve and render as doors,
- the Home segment is what rendered and **no** Activity pane or feed rows appear.

A second test covers the empty-entry-points case.

## Verification

`ember-tsc` (glint) and eslint/template-lint pass. The host integration suite requires the full local stack, so functional execution runs in CI.

### DIFF
--- a/packages/host/tests/integration/workspace-indexing-test.gts
+++ b/packages/host/tests/integration/workspace-indexing-test.gts
@@ -1,0 +1,175 @@
+import { module, test } from 'qunit';
+
+import type { IndexedInstance, Realm } from '@cardstack/runtime-common';
+
+import {
+  testRealmURL,
+  setupLocalIndexing,
+  setupIntegrationTestRealm,
+} from '../helpers';
+import { setupBaseRealm, setupWorkspaceCard } from '../helpers/base-realm';
+import { setupMockMatrix } from '../helpers/mock-matrix';
+import { setupRenderingTest } from '../helpers/setup';
+
+// The Workspace card is a realm's default index card, so its isolated shell is
+// rendered by the indexer/prerender. This suite locks in the defensive patterns
+// that keep that render cheap and crash-free: the Activity feed must stay lazy,
+// and `linksToMany` entry points must not have `.constructor` called on them
+// before they resolve.
+//
+// The Workspace is authored at a NON-index path (`ws.json`, not `index.json`) on
+// purpose. The in-browser indexer substitutes boilerplate HTML for a realm's
+// default index card (routes/render/html.ts `#isDefaultRealmIndexCard`), which
+// would skip the Glimmer render entirely. Rendering it off the index path forces
+// the real isolated render through the indexer so we actually exercise the
+// component's constructor, modifiers, and template — the same path the
+// production puppeteer prerender takes for the index card itself.
+module(`Integration | Workspace indexing/prerender safety`, function (hooks) {
+  setupRenderingTest(hooks);
+  setupBaseRealm(hooks);
+  setupWorkspaceCard(hooks);
+  setupLocalIndexing(hooks);
+  let mockMatrixUtils = setupMockMatrix(hooks);
+
+  async function getInstance(
+    realm: Realm,
+    url: URL,
+  ): Promise<IndexedInstance | undefined> {
+    let maybeInstance = await realm.realmIndexQueryEngine.instance(url);
+    if (maybeInstance?.type === 'instance-error') {
+      return undefined;
+    }
+    return maybeInstance as IndexedInstance | undefined;
+  }
+
+  function countOccurrences(haystack: string, needle: string): number {
+    return haystack.split(needle).length - 1;
+  }
+
+  test('indexing a Workspace renders its isolated shell and resolves pinned entry points without eagerly loading the Activity feed', async function (assert) {
+    let { realm } = await setupIntegrationTestRealm({
+      mockMatrixUtils,
+      contents: {
+        // Two plain cards pinned as entry points (a `linksToMany(CardDef)`).
+        'pinned-a.json': {
+          data: {
+            meta: {
+              adoptsFrom: {
+                module: '@cardstack/base/card-api',
+                name: 'CardDef',
+              },
+            },
+          },
+        },
+        'pinned-b.json': {
+          data: {
+            meta: {
+              adoptsFrom: {
+                module: '@cardstack/base/card-api',
+                name: 'CardDef',
+              },
+            },
+          },
+        },
+        // A Workspace pinning both cards. Off the index path so the real
+        // isolated render runs (see the module comment).
+        'ws.json': {
+          data: {
+            relationships: {
+              'entryPoints.0': { links: { self: './pinned-a' } },
+              'entryPoints.1': { links: { self: './pinned-b' } },
+            },
+            meta: {
+              adoptsFrom: {
+                module: '@cardstack/base/workspace',
+                name: 'Workspace',
+              },
+            },
+          },
+        },
+      },
+    });
+
+    let instance = await getInstance(realm, new URL(`${testRealmURL}ws`));
+    assert.ok(
+      instance,
+      'the Workspace indexed cleanly (no render error on entry-point resolution)',
+    );
+
+    let isolatedHtml = instance?.isolatedHtml ?? '';
+    assert.ok(
+      isolatedHtml.includes('data-test-workspace-index'),
+      'the Workspace isolated shell rendered (not boilerplate, no crash)',
+    );
+    assert.ok(
+      isolatedHtml.includes('data-test-workspace-tab="home"'),
+      'the Home tab is present',
+    );
+    assert.ok(
+      isolatedHtml.includes('data-test-workspace-tab="activity"'),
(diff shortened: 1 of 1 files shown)


packages/host/tests/integration/workspace-indexing-test.gts
```diff
@@ -0,0 +1,175 @@
...
+    // The Activity feed must NOT hydrate at index time: its pane and rows only
+    // render under the Activity segment, which prerender never opens.
+    assert.notOk(
+      isolatedHtml.includes('activity-pane'),
+      'the Activity pane did not render at index time',
+    );
+    assert.notOk(
+      isolatedHtml.includes('feed-row'),
+      'no Activity feed rows were eagerly loaded during indexing',
+    );
```
habdelra, line 139: [Claude Code 🤖]

**These two assertions confirm "the Activity segment didn't render," not "the feed search didn't run" — non-blocking, but narrower than the description claims.**

**The two `notOk`s are the same fact twice.** Both `activity-pane` and `feed-row` live *exclusively* inside the `{{else}}` (Activity) branch of the segment conditional in `packages/base/workspace.gts` — the `.activity-pane` wrapper and the `feed` / `feed-row` markup are all under that branch (~lines 740–901). The Home branch emits neither string. So once the segment is `home`, both `notOk`s are already guaranteed; they're two views of one condition, not independent signals.

**Neither observes whether the search ran.** The reverse-chron query is gated separately from the markup. `loadFeed.perform()` has three call sites in `workspace.gts` — `setSegment` (~2276), the `setupRealmSubscription` modifier (~2970), and `refreshAfterIndex` (~3208) — and only the `if (this.segment === 'activity')` guard keeps it from firing. `setupRealmSubscription` *does* run at index time (it's attached to the root `<section>` and performs `loadFilterList` / `loadJobs` / `loadLatest` / `loadConfig` during the prerender). If a refactor dropped that one guard, the whole-realm `store.search` would run during indexing — the exact cost this PR guards against — yet it would still render zero `feed-row`s on Home, so both tests stay green.

**Way out:** to pin the property the description states ("the whole-realm reverse-chron search never runs during indexing"), assert on the search rather than the markup — e.g. count `store.search` calls (or observe the `loadFeed` task state) across the index and assert the feed query never ran. As written, the suite is still a real guard that *Home is what prerenders and it doesn't crash* — it just doesn't lock laziness.

Class: asserted-but-misleading coverage / follow-up. Non-blocking.

---

packages/host/tests/integration/workspace-indexing-test.gts
```diff
@@ -0,0 +1,175 @@
...
+      'the Home segment rendered (the Pinned zone is present)',
+    );
+
+    // Entry points resolved and rendered as doors — the lazy `.constructor`
+    // access (doorKind / entryPointsFilter getter) survived prerender.
+    assert.strictEqual(
+      countOccurrences(isolatedHtml, 'class="door"'),
+      2,
+      'both pinned entry points rendered as doors',
+    );
```
habdelra, line 128: [Claude Code 🤖]

**This pins the `doorKind` path well; it does not exercise the `entryPointsFilter` lazy `get cards()` the comment also names — worth knowing so the coverage isn't over-read.** Non-blocking.

Two distinct sites touch `.constructor` on an entry point:

- **`doorKind`** (`workspace.gts` ~2290) reads `card.constructor` for every pinned door and runs during the **Home** render (`{{this.doorKind index}}`, template ~line 410). This test renders Home with two resolved entry points, so a throw here surfaces as an `instance-error` (→ `getInstance` returns `undefined` → the `assert.ok(instance)` above fails) or as a door count ≠ 2. **That half is genuinely locked in.** 👍

- **`entryPointsFilter`'s `get cards()`** (`workspace.gts` ~2854) is the getter the comment on line 123 (and the PR description) specifically call out — "touching `.constructor` on an unresolved entry … crashes prerender." But `cards` only executes when the **Library** "Entry points" filter renders through `CardsGridLayout`; this test only ever renders the Home segment, so `cards` is never invoked. A regression making it eager would be caught only incidentally (if the option object were built with present-but-unresolved entries during construction), not by design.

**Way out:** if the lazy-`cards` pattern is worth pinning, either drive the render to the Library entry-points filter or unit-test `entryPointsFilter` directly. Otherwise consider narrowing the comment on line 123 so it doesn't read as if this assertion covers the `entryPointsFilter` getter — as written it covers `doorKind`.

Class: confirmation (doorKind) + coverage gap (entryPointsFilter). Non-blocking.

---

packages/host/tests/integration/workspace-indexing-test.gts
```diff
@@ -0,0 +1,175 @@
...
+      },
+    });
+
+    let instance = await getInstance(realm, new URL(`${testRealmURL}ws`));
+    assert.ok(
+      instance,
+      'the Workspace indexed cleanly (no render error on entry-point resolution)',
+    );
+
+    let isolatedHtml = instance?.isolatedHtml ?? '';
```
habdelra, line 99: [Claude Code 🤖]

Minor / convention: every other `isolatedHtml` assertion in the repo runs the string through `stripScopedCSSAttributes(...)` first — see the sibling `realm-indexing-test.gts` (imported at its top and used ~line 81+) and `realm-server/tests/indexing-test.ts` (~line 634+). The raw `instance.isolatedHtml` used here still carries the ` data-scopedcss-<hash>` attributes glimmer-scoped-css stamps on each element.

It's harmless for these particular assertions — the scope attribute is appended *after* the class attribute, so `class="door"` still matches, and the `<style>` block is injected into `document.head` (not the card DOM), so `.activity-pane` / `.feed-row` CSS selectors don't leak into `isolatedHtml`. But the `class="door"` count silently depends on that attribute ordering. Running the html through `stripScopedCSSAttributes` first would de-couple the assertions from scope-attribute placement and match the established convention.

Class: convention / robustness. Non-blocking.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** this is a test-only PR (one new integration spec, no production change), so the review question is narrow — does the test *pin* the two safety properties it names, or does it pass for reasons that would survive a regression in those properties? I checked each assertion against what the `Workspace` component (`packages/base/workspace.gts`) actually emits at index time, on the checked-out head.

**Bottom line: no blocking issues.** The test is a genuine, valuable guard: it proves the real isolated shell prerenders (off the index path, so it isn't the boilerplate), doesn't crash on entry-point resolution, and resolves pinned links into doors. Two properties are pinned less tightly than the description implies — both non-blocking, flagged so the coverage isn't over-read.

**What lands right (verified):**
- **Off-index-path rationale is correct.** `#isDefaultRealmIndexCard` (`host/app/routes/render/html.ts` ~111/125) sets `useRealmIndexBoilerplate`, and `card-prerender.gts` (~728–731) returns `REALM_INDEX_BOILERPLATE_HTML` and skips Glimmer for the default index card. Authoring at `ws.json` genuinely forces the real render — the module comment is accurate.
- **Crash-safety on entry-point resolution is pinned.** A throw during `.constructor` access on the doors would make the instance an `instance-error` and fail `assert.ok(instance)`; the `class="door"` count == 2 further proves both links resolved. This is the strongest part of the suite.
- **Selectors are real.** `data-test-workspace-index`, `data-test-workspace-tab="home"/"activity"`, the `Pinned` heading, and `<article class="door">` all exist in the template and match as written.

1. The `activity-pane` / `feed-row` `notOk`s confirm the Home segment rendered, not that the feed search stayed lazy — a dropped `segment === 'activity'` guard would still pass. To pin the description's claim, assert on `store.search` not running. *(thread on the two `notOk` assertions)*
2. The `entryPointsFilter` lazy `get cards()` the comment calls out is never executed (the Library segment isn't rendered); only `doorKind` is. Add Library coverage or narrow the comment. *(thread on the doors assertion)*
3. Optional: run `isolatedHtml` through `stripScopedCSSAttributes` to match the repo convention and de-couple the `class="door"` count from scope-attribute ordering. *(thread on line 99)*

This is a COMMENT review — approve / request-changes is the maintainer's call.

---

habdelra (approved):

PR #5620, opened 2026-07-27
Workspace: extract tuning constants and lock RealmConfig fallback

lukemelia:
Replaces the Workspace card's hard-coded literals with named module constants and adds regression coverage for rendering when the linked `RealmConfig` card is missing. (CS-12282; builds on the Workspace extraction in #5582.)

## Extract magic numbers

Each value now comes from a named constant. They live as module constants because they are internal layout/bound tuning rather than per-realm settings — the one operator-facing lever, tile density, stays a `pinnedSize` edit-format setting that selects between the two door heights:

| Constant | Value | Was |
| --- | --- | --- |
| `ETA_IMPLAUSIBLE_MINUTES` | 30 | honest-ETA suppression threshold |
| `ACTIVITY_FEED_CAP` | 100 | activity-feed query bound + "last N changes" note |
| `FEED_REVEAL_CHUNK` | 20 | reveal-on-scroll page increment |
| `SEARCH_PAGE_SIZE` | 100 | generic realm-local search / chooser page bound |
| `SEARCH_RESULTS_CAP` | 8 | inline header (⌘K) search hits |
| `DOOR_TILE_HEIGHT_PX` / `_COMPACT_PX` | 300 / 220 | pinned tile heights |
| `TYPE_RAIL_EXCLUDED_IDS` | — | base machinery hidden from the `_types` rail |

The door heights are now sourced entirely from the `doorsStyle` getter; the CSS custom-property fallback literal (`var(--door-h, 300px)`) is gone, and the "last N changes" note interpolates `ACTIVITY_FEED_CAP` instead of a literal.

## Graceful fallback

A Workspace whose `workspace` link (the realm's `RealmConfig` card) is absent already renders Home and settings without erroring — the config-owned identity inputs are guarded and the getters default. Two new tests lock that behavior:

- Home (isolated) renders its shell with no linked config.
- Settings (edit) render, omitting the config-owned Name/Icon inputs rather than erroring.

## Verification

- glint (`ember-tsc --noEmit`): 0 errors
- `ember-template-lint`: clean
- `prettier --check`: clean

### DIFF
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -93,8 +93,12 @@ type JobCard = CardDef & {
   setupSurvey?: CardDef; // the separate, optional themed survey card
 };
 
-// honest ETA (same rules as the job card): linear from arrival rate,
-// only once 3 pieces are in, suppressed when implausible (> 30 min).
+// A projected ETA further out than this many minutes reads as noise rather
+// than signal, so it is suppressed.
+const ETA_IMPLAUSIBLE_MINUTES = 30;
+
+// honest ETA (same rules as the job card): linear from arrival rate, only
+// once 3 pieces are in, suppressed when implausible (see the threshold above).
 // The subset of a job card the ETA reads. A full `JobCard` satisfies it.
 export interface EtaJob {
   progressDone?: number;
@@ -117,7 +121,7 @@ export function etaMinutes(
     return undefined;
   }
   let mins = Math.round(((elapsed / done) * (total - done)) / 60000);
-  return mins > 30 ? undefined : mins;
+  return mins > ETA_IMPLAUSIBLE_MINUTES ? undefined : mins;
 }
 
 // A card modified within this window of its creation reads as "Created" in
@@ -208,6 +212,37 @@ function excludeSelfReferentialCards(on?: CodeRef): Filter[] {
 // flow writing many cards) emits a burst of index events; coalescing them into
 // one refresh avoids firing the panel searches several times per keystroke.
 const INDEX_REFRESH_DEBOUNCE_MS = 200;
+
+// Layout / bound tuning. These are internal tuning knobs rather than per-realm
+// settings, so they live as module constants — the one operator-facing lever,
+// tile density, is the `pinnedSize` edit-format setting that selects between
+// the two door heights below.
+
+// Activity feed: both the server query and the reveal-on-scroll pager are
+// bounded to this many of the most recently modified cards.
+const ACTIVITY_FEED_CAP = 100;
+// The feed reveals this many rows at a time as the sentinel scrolls into view.
+const FEED_REVEAL_CHUNK = 20;
+
+// Generic upper bound for realm-local search / chooser result pages. Callers
+// that need fewer rows may request a smaller page.
+const SEARCH_PAGE_SIZE = 100;
+
+// The header's ⌘K search lists this many hits inline; the rest live behind the
+// "See all" row in Library.
+const SEARCH_RESULTS_CAP = 8;
+
+// Pinned-card tile heights (px) for the two densities `pinnedSize` selects.
+const DOOR_TILE_HEIGHT_PX = 300;
+const DOOR_TILE_HEIGHT_COMPACT_PX = 220;
+
+// Base machinery the `_types` rail never lists: these are not realm content.
+const TYPE_RAIL_EXCLUDED_IDS = [
+  `${baseRealmRRI}card-api/CardDef`,
+  `${baseRealmRRI}cards-grid/CardsGrid`,
+  `${baseRealmRRI}card-api/FieldDef`,
+  `${baseRealmRRI}card-api/FileDef`,
+];
 
 function toMs(value: unknown): number | undefined {
   let ms =
@@ -632,10 +667,7 @@ class Isolated extends Component<typeof Workspace> {
                 <button
                   type='button'
                   class='rail-row
-                    {{if
-                      (eq option.id this.activeFilter.id)
-                      "selected"
-                    }}'
+                    {{if (eq option.id this.activeFilter.id) "selected"}}'
                   {{on 'click' (this.selectFilter option)}}
                 >
                   {{#let (this.iconComponent option) as |Icon|}}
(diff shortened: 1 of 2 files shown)


packages/base/workspace.gts
```diff
@@ -2662,10 +2691,12 @@ class Isolated extends Component<typeof Workspace> {
...
-  private get doorsStyle(): SafeString | undefined {
-    return this.args.model.pinnedSize === 'compact'
-      ? htmlSafe('--door-h: 220px')
-      : undefined;
+  private get doorsStyle(): SafeString {
+    let height =
+      this.args.model.pinnedSize === 'compact'
+        ? DOOR_TILE_HEIGHT_COMPACT_PX
+        : DOOR_TILE_HEIGHT_PX;
+    return htmlSafe(`--door-h: ${height}px`);
```
habdelra, line 2699: [Claude Code 🤖]

**Confirmation — this is correct, and here's the invariant it now rests on (non-blocking).**

This is the one change in the PR with a real behavioral surface, and it's safe. Making `doorsStyle` return `SafeString` unconditionally is what lets the CSS drop its literal fallback (`grid-auto-rows: var(--door-h, 300px)` → `var(--door-h)` at line 1701). I traced both halves:

- **Binding is unconditional.** `--door-h` is consumed only in the `.doors` rule (line 1701), and the only element carrying that class is `<div class='doors' style={{this.doorsStyle}}>` (line 463). Since the getter now always returns a value, `--door-h` is always defined on that element, so the removed `300px` fallback can never be reached.
- **Rendered output is unchanged.** Non-compact: old path was `undefined` inline → CSS fallback `300px`; new path is `--door-h: 300px` inline → `300px`. Compact: `220px` either way. Identical in both densities.

The thing worth naming for the next editor: the CSS fallback removal makes this getter **load-bearing**. If a future change lets `doorsStyle` return `undefined`/empty again (or the inline `style` binding is dropped from line 463), `grid-auto-rows` silently resolves to `auto` — collapsed tiles — with nothing in the CSS to catch it. Today the return type (`SafeString`, not `SafeString | undefined`) guards that at the type level, which is the right place for it. No change requested.

---

packages/host/tests/integration/components/workspace-test.gts
```diff
@@ -44,4 +44,38 @@ module('Integration | Card | workspace', function (hooks) {
...
+        .dom('.settings-title')
+        .hasText('Workspace settings', 'the settings form still renders');
+      // Name/Icon live on the RealmConfig card, so they fall away when it is
+      // missing — the rest of the form (including the config-card picker) stays.
+      assert
+        .dom('input[placeholder="Workspace name"]')
+        .doesNotExist('the config-owned name input is omitted, not errored');
+      assert
+        .dom('input[placeholder="https://…"]')
+        .doesNotExist('the config-owned icon input is omitted, not errored');
```
habdelra, line 78: [Claude Code 🤖]

**These assertions do pin the guard — one narrow gap worth knowing about (non-blocking, minor).**

I checked what these `doesNotExist` assertions actually catch. The Name/Icon inputs are gated behind `{{#if @model.workspace}}` (workspace.gts:3836), and with `new Workspace({})` the `workspace` link (`@field workspace = linksTo(CardDef)`, workspace.gts:4176) is undefined, so the block never renders. If someone deletes that `{{#if}}` guard, the inputs *would* render (their getters — `workspaceName`/`workspaceIcon` at workspace.gts:3395-3404 — default to `''` via `?.`/`??` rather than throwing), and these assertions would then fail. So the guard-removal regression **is** covered. Good.

The one thing that would slip through: the selectors key off the placeholder strings `"Workspace name"` / `"https://…"`. Rename a placeholder and the selector matches nothing, so `doesNotExist` stays green whether or not the input is really gone — the assertion decouples silently from the thing it's meant to watch. Since there's no positive-case test in this PR (config present → inputs *do* render with those placeholders) to anchor the strings, that decoupling wouldn't be noticed.

Not asking you to add the positive test here — it's a reasonable follow-up. If you want to harden these two lines cheaply now, a `data-test-*` marker on each input (or on the `{{#if @model.workspace}}` block) gives the assertions a stable hook that a copy rename can't rot.

---

habdelra (commented):
[Claude Code 🤖]

Reviewed as a behavior-preserving refactor: for the constant extraction the lens is "did any literal's meaning or value change at any call site," and for the new tests it's "would they fail if the graceful-degradation path broke."

**Bottom line: no blocking issues. This is a clean, well-scoped refactor and the two new tests genuinely pin the behavior they claim.** The one change with a real behavioral surface (dropping the CSS `--door-h` fallback) is safe; details in the inline thread.

**What lands right**

- **The extraction is complete and value-exact.** I enumerated every replaced literal against its constant and swept the file for stragglers — no `size: 100`, `slice(0, 8)`, `> 30`, `>= 100`, or `+ 20` left behind. The lone remaining `220px` (workspace.gts:3992) is an unrelated settings-column width, correctly left alone (coincidental value collision, not a door height).
- **The right semantic split.** The old literal `100` appeared as both the activity-feed bound and the search/chooser page size; splitting it into `ACTIVITY_FEED_CAP` (feed query + `feedAtCap` + the "last N changes" note) and `SEARCH_PAGE_SIZE` (the four search/chooser sites) is the correct call — same value today, different roles, free to diverge later. The feed query (workspace.gts:3347) correctly uses `ACTIVITY_FEED_CAP`, not the search constant.
- **`TYPE_RAIL_EXCLUDED_IDS` at module scope is safe.** `baseRealmRRI` is `export const baseRealmRRI = ri('@cardstack/base/')` (a module constant), so the module-level template-literal init is fully deterministic and byte-identical to the old method-local array; ESM import order guarantees it's initialized first. Hoisting it also stops rebuilding the array on every `loadTypes` run.
- **`doorsStyle` / CSS fallback removal is behavior-preserving** in both densities — see the inline confirmation, including the invariant this now rests on.
- No content-tag `<template>` hazards introduced (the new backtick literals are all module-scope, outside `<template>`); CI on the work commit was fully green (glint 0, template-lint/prettier clean, all host + realm tests passing). The current re-run is just against the merge-in-of-main commit.

**Recommendations**

1. (non-blocking, minor) Consider a `data-test-*` hook on the config-owned Name/Icon inputs so the settings test's `doesNotExist` assertions can't silently decouple from a placeholder rename — see the inline thread on the test file. A positive-case test (config present → inputs render) is a fine follow-up rather than part of this PR.

**Adjacent, out of scope (not asking for it here)**

- The three `class='rail-row {{if (eq option.id this.activeFilter.id) "selected"}}'` sites (workspace.gts:~692/714/752) are the inline-`{{if}}`-in-class-string pattern the repo's `.gts` conventions steer toward the `cn` helper. This PR only reflowed their whitespace via prettier — pre-existing, and converting them would add churn unrelated to the constant extraction. Flagging for whoever next touches that rail markup.

---

habdelra (approved):

PR #5621, opened 2026-07-27
Announce the Workspace chrome's self-updating surfaces

lukemelia:
## What

An accessibility pass over the Workspace card's chrome, plus a platform-correct hotkey hint.

### Status regions for the surfaces that update on their own

Search results and setup progress both change with no interaction to hang an announcement off, so assistive tech got nothing when they did. Two `role='status'` regions now carry them. Both are rendered up front and left in place — a live region only announces text that changes while it is already in the DOM — and both are deliberately coarser than the thing they describe:

- **Search** reports a count, not the matches. It reruns on every keystroke, and reading titles back would talk over the user's typing. It also stays quiet while the debounced search is in flight, so a prefix of a term that *does* match can't be announced as "no matches" on the way there.
- **Progress** reports the job and a percentage quantised to quarters. The meters advance continuously, and a region tracking them would interrupt constantly. `progressMilestone` is where the quantising lives, with its own unit coverage, so the intent survives a later "simplification" back to announcing every tick.

The progress region sits at the card root rather than inside a segment: the Activity tab's dot is always on screen, while the dock that details it renders only on that one tab.

### Decoration marked as decoration

The rings, meters and dots are now `aria-hidden` — they restate percentages the surrounding text already spells out (`.setup-pct`, `dockSummary`), so exposing them just doubled up.

One of these was a real defect rather than noise: the collapsed dock's button carried `aria-label='Show progress details'`, which **replaces** the button's own text as its accessible name. That text is the live summary of what is being set up, so a screen reader heard the action and never what was progressing. The label is gone and the action is appended as hidden text instead, leaving the accessible name with both.

### Platform-correct hotkey

The Frame hint was a hardcoded `⌘K`. It now reads `Ctrl+K` off macOS, via an exported `searchHotkeyLabel` matching on `Mac` exactly as `codemirror-editor.gts`'s own mod-key label does. The input carries `aria-keyshortcuts='Meta+K Control+K'` — both spellings, because `setupSearchHotkey` binds `metaKey || ctrlKey` — and the visible hint is `aria-hidden` now that the shortcut is on the input itself.

Two caveats worth knowing rather than hiding:

- The label resolves once, against whichever browser evaluates the module. In the app that is the user's own, which is the case this is for. A **prerender** pass resolves it against the prerender machine, so generated HTML carries that spelling until the app renders the card live.
- Snapshots taken on CI (Linux) will now show `Ctrl+K` where they showed `⌘K`. That is the correct label for the machine taking them.

## The `.buried` rule is live, not dead

The issue's third item assumed `.operator-mode .buried` was dead CSS and asked to wire it or delete it. It is already driven by real operator-mode state, so this documents and pins it instead.

Both ancestors are host chrome that really renders: `container.gts` puts `class='operator-mode'` on the operator-mode container, and `stack-item.gts` applies `buried` via `cn` to a stack item that isn't on top. They resolve from inside a card's scoped CSS because the scoping transform attaches this card's attribute only to the selector's **last** compound (`rewriteSelector` in glimmer-scoped-css walks to the last non-pseudo, non-combinator node), leaving the ancestor part to match freely outside the card's own DOM.

Verified rather than reasoned about: two tests render the card under the ancestors the host really uses, one with `.buried` and one without, and assert `.frame-actions` is hidden in the first and visible in the second — the second being the control that rules out the wrapper hiding it for some unrelated reason.

**One gap this surfaced, left alone deliberately:** host mode marks its stack items `buried` too (`host-mode/stack-item.gts`) but has no `.operator-mode` ancestor, so a buried Workspace keeps its chrome there. Dropping the `.operator-mode` prefix would fix that in one line, but it changes visual behavior on published sites, which is beyond an accessibility pass. Noted in a comment next to the rule; happy to split it out if wanted.

## Test plan

`Integration | Card | workspace` — **46 assertions, 0 failures** (was 19 before this branch).

New `Integration | Card | workspace | accessibility` (8 tests):

- The hotkey hint is `aria-hidden` and the input carries both shortcut spellings.
- The hint's text matches what the module computed for the running browser, so the test reads the same on any platform rather than asserting a fixed glyph.
- Both status regions exist up front, carry `role='status'`, and are empty when there is nothing to say.
- A search with no matches announces "No matching cards" rather than staying silent — waited for rather than asserted into the debounce window.
- That same announcement is asserted to be **visually** hidden via computed style (`position: absolute`, `clip-path: inset(50%)`). `.boxel-sr-only` lives in `@layer utilities` and this card's scoped CSS is unlayered, so any scoped rule matching those spans would win and put the text on screen; the assertion catches that if it ever happens.
- The chrome is hidden on a buried operator-mode card, and visible when not buried.

Extended `Integration | Card | workspace | pure functions`: `progressMilestone` holds steady between quarters (26 and 49 both report 25 — the anti-interruption property), advances on each quarter, and clamps negatives; `searchHotkeyLabel` covers macOS, Windows, Linux and the no-`navigator` fallback.

`ember-tsc --noEmit` clean for both the host and realm-server projects; eslint, ember-template-lint and prettier clean on the changed files.

Note: `packages/base` is not prettier-enforced in CI and `workspace.gts` is not currently prettier-clean, so formatting here is confined to the touched lines rather than swept file-wide.

### DIFF
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -143,6 +143,22 @@ export type ActivityVerb = 'Created' | 'Updated' | 'Remixed';
 // the Activity feed rather than "Updated".
 const CREATED_WINDOW_MS = 120000;
 
+// Setup progress is announced on crossing a multiple of this, rather than on
+// every change, so a long-running job reports in a few times instead of
+// continuously.
+const PROGRESS_ANNOUNCE_STEP = 25;
+
+// Rounds a live percentage down to the last announced milestone. The status
+// region's text is derived from this, so it only changes when a milestone is
+// crossed — that is what keeps a job whose meter advances every few seconds from
+// interrupting a screen reader every few seconds.
+export function progressMilestone(pct: number): number {
+  return (
+    Math.floor(Math.max(0, pct) / PROGRESS_ANNOUNCE_STEP) *
+    PROGRESS_ANNOUNCE_STEP
+  );
+}
+
 export function classifyActivityVerb(
   modMs: number | undefined,
   createdMs: number | undefined,
@@ -173,6 +189,22 @@ export function activityVerbFor(
     ? 'Remixed'
     : classifyActivityVerb(modMs, createdMs);
 }
+
+// How the Frame typeahead's hotkey is spelled on the platform the user is
+// actually on. `setupSearchHotkey` binds `metaKey || ctrlKey`, so both spellings
+// work everywhere and this is purely about naming the one they'd reach for.
+// Matches on `Mac`, as codemirror-editor.gts's own mod-key label does.
+export function searchHotkeyLabel(platform: string): string {
+  return /Mac/i.test(platform) ? '⌘K' : 'Ctrl+K';
+}
+
+// Resolved once, against whichever browser evaluates this module. In the app
+// that is the user's own, which is the case this is for. A prerender pass
+// resolves it against the prerender browser instead, so generated HTML carries
+// that machine's spelling until the app renders the card live.
+const SEARCH_HOTKEY_LABEL = searchHotkeyLabel(
+  typeof navigator === 'undefined' ? '' : navigator.platform,
+);
 
 type RealmConfigCard = CardDef & { iconURL?: string }; // RealmConfig shape
 
@@ -351,6 +383,16 @@ class Isolated extends Component<typeof Workspace> {
       data-test-workspace-index
       {{this.setupRealmSubscription this.primaryRealm}}
     >
+      {{! Setup progress runs on its own — jobs start, advance and finish with no
+        interaction to hang an announcement off. Lives at the root, outside every
+        segment: the Activity tab's dot is always on screen while the dock that
+        details it is only rendered on that one tab, and a live region has to be
+        in the DOM before its text changes to be announced at all. }}
+      <span
+        class='boxel-sr-only'
+        role='status'
+        data-test-progress-announcement
+      >{{this.progressAnnouncement}}</span>
       <header class='frame'>
         <nav class='tabs' aria-label='Sections'>
           <button
@@ -373,7 +415,9 @@ class Isolated extends Component<typeof Workspace> {
           ><ActivityIcon class='tab-icon' />
             Activity{{#if this.runningJobs.length}}<span
                 class='attention-dot'
-              />{{/if}}</button>
+                aria-hidden='true'
+              /><span class='boxel-sr-only'>({{this.runningJobs.length}}
+                in progress)</span>{{/if}}</button>
         </nav>
         {{#if @model.signage}}
           {{! workspace signage — hover reveals the purpose annotation }}
@@ -388,15 +432,32 @@ class Isolated extends Component<typeof Workspace> {
               type='text'
               placeholder='Search'
               aria-label='Search this space'
+              aria-keyshortcuts='Meta+K Control+K'
+              aria-controls='workspace-search-results'
+              aria-expanded={{if this.searchResults.length 'true' 'false'}}
               value={{this.searchTerm}}
               {{on 'input' this.onSearchInput}}
               {{on 'keydown' this.onSearchKeydown}}
               {{on 'focus' this.onSearchFocus}}
               {{on 'focusout' this.onSearchBlur}}
             />
-            <span class='search-kbd'>⌘K</span>
+            {{! the visible hint is decorative — aria-keyshortcuts above carries
+              the same thing to assistive tech, in both spellings }}
+            <span
+              class='search-kbd'
+              aria-hidden='true'
+            >{{SEARCH_HOTKEY_LABEL}}</span>
+            {{! Results appear and refresh without any focus change, so nothing
+              would reach a screen reader on its own. Announced as a count
+              rather than a list of titles: the search reruns on each keystroke,
+              and reading matches back would talk over the user's typing. }}
+            <span
+              class='boxel-sr-only'
+              role='status'
+              data-test-search-announcement
+            >{{this.searchAnnouncement}}</span>
             {{#if this.searchResults.length}}
-              <div class='search-results'>
+              <div class='search-results' id='workspace-search-results'>
                 {{#each this.searchResults as |result|}}
                   <button
                     type='button'
@@ -441,7 +502,13 @@ class Isolated extends Component<typeof Workspace> {
                 class='setup-bar'
                 {{on 'click' (this.setSegment 'activity')}}
               >
-                <span class='setup-ring' style={{this.ringStyle job}}>
+                {{! ring and meter both restate the percentage this button
+                  already spells out in `.setup-pct`, so they are decoration }}
+                <span
+                  class='setup-ring'
+                  style={{this.ringStyle job}}
+                  aria-hidden='true'
+                >
                   <span class='setup-ring-hole' /></span>
                 <span class='setup-lines'>
                   <span class='setup-name'>Setting up
@@ -452,7 +519,7 @@ class Isolated extends Component<typeof Workspace> {
                       ·
                       {{this.jobEta job}}{{/if}}</span>
                 </span>
-                <span class='setup-track'><span
+                <span class='setup-track' aria-hidden='true'><span
                     class='setup-fill'
                     style={{this.jobFillStyle job}}
                   /></span>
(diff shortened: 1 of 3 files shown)


packages/base/workspace.gts
```diff
@@ -306,15 +350,32 @@ class Isolated extends Component<typeof Workspace> {
               type='text'
               placeholder='Search'
               aria-label='Search this space'
+              aria-keyshortcuts='Meta+K Control+K'
+              aria-controls='workspace-search-results'
```
FadhlanR: [Claude Code 🤖] Small wiring nit: `aria-controls='workspace-search-results'` is always present on the input, but the element carrying that id is only rendered inside the `{{#if this.searchResults.length}}` block below. So whenever there are no results — the same state where `aria-expanded` is `'false'` — this IDREF points at an element that isn't in the DOM. Worth gating the attribute on `this.searchResults.length` (or otherwise keeping the reference and its target in sync) so it never dangles. Non-blocking.

packages/base/workspace.gts
```diff
@@ -2571,6 +2669,30 @@ class Isolated extends Component<typeof Workspace> {
     return this.runningJobs[0]!; // Read only under the runningJobs.length guard.
   }
 
+  // What the progress status region says. Deliberately coarser than the meters
+  // it stands in for: those advance continuously, and a region that changed with
+  // them would talk over everything else on the page. Quantising to quarters
+  // means the text changes at most four times per job, plus whenever the set of
+  // running jobs changes.
+  get progressAnnouncement(): string {
```
FadhlanR, line 2810: [Claude Code 🤖] Gap worth noting on the progress region: when the last running job finishes, `runningJobs` empties and this getter returns `''`. A live region going from text to empty announces nothing, so completion is never spoken — and whether the final `100% complete` milestone is even reached depends on timing (if the job leaves `runningJobs` in the same tick it hits 100%, the region jumps straight from `75% complete` to silent). If hearing that setup finished matters, consider a brief terminal announcement (e.g. `Setup complete`) before the region clears. Non-blocking follow-up, not a merge blocker.

FadhlanR (approved):
[Claude Code 🤖] Approving. Careful, well-tested accessibility pass — the status regions are correctly rendered up front and coarsened to avoid chatter, the quarter-quantised progress has dedicated unit tests pinning the anti-interruption intent, and the collapsed dock button's `aria-label` shadowing its own live summary text was a real defect that's now fixed. Verifying the `.buried` rule with tests that render under the actual host ancestors, rather than assuming it dead, is a nice touch too.

Two non-blocking follow-ups left inline.

burieberry (commented):
Reviewed the accessibility pass. The status-region approach and the `aria-label`-replacing-accessible-name fix are both right, and the `.buried` investigation is well evidenced. Four findings, all in `packages/base/workspace.gts`.

### 1. Blurring the search box announces "No matching cards" for a search that did match

`workspace.gts:2444` + `workspace.gts:2500`

`onSearchBlur` → `hideResults` waits 200ms then splices `searchResults` empty, but leaves `searchTerm` intact. `searchAnnouncement` then sees a non-empty term, `runSearch` not running, and zero results, so it lands on the `'No matching cards'` branch:

```js
if (!this.searchTerm.trim() || this.runSearch.isRunning) return '';
let shown = this.searchResults.length;
if (!shown) return 'No matching cards';
```

Repro: type `de`, hear "3 results", then click or tab away. 200ms later the still-mounted `role='status'` region changes to "No matching cards" and a screen reader reads it — for a search that matched.

Escape (`onSearchKeydown`) and `openResult` both clear `searchTerm`, so blur is the only path that leaves this state, which is why the new tests don't catch it. The `!term || !store` early return in `runSearch` reaches the same branch, so a missing `context.store` also announces "No matching cards" for a search that never ran.

The getter needs to distinguish *results dismissed* from *search returned nothing* — e.g. a tracked `lastSearchedTerm` (or `hasSettledResults`) set only on `runSearch`'s success path and cleared by `hideResults`.

### 2. Hardcoded `id='workspace-search-results'` breaks when the card renders twice

`workspace.gts:378`

This is the only literal `id` in the file. Two Workspace cards side by side in operator mode — or the same one open in two stacks — put two elements with this id in the document, and every `aria-controls='workspace-search-results'` resolves to the first. The right-hand card's input then points at the left-hand card's results list. It's also a `duplicate-id-aria` violation of exactly the kind this PR is fixing. Generating it per instance (`guidFor(this)`) and interpolating into both places fixes it.

### 3. `aria-controls` / `aria-expanded` on a plain textbox

`workspace.gts:354-355`

`aria-expanded` is not a global ARIA attribute and is not supported on the implicit `textbox` role of `<input type='text'>`, so the expanded state is dropped by assistive tech and `aria-allowed-attr` flags it. `aria-controls` additionally dangles whenever `searchResults.length` is 0, since `#workspace-search-results` only exists inside the `{{#if}}` — including in the "No matching cards" state.

Either commit to the combobox pattern (`role='combobox'`, `aria-autocomplete='list'`, results as `listbox`/`option`, active-descendant management) or drop both attributes and rely on the status region this PR already adds. The status region alone is a defensible choice here given the results are click-only.

### 4. The `isRunning` guard re-announces an unchanged count on every keystroke

`workspace.gts:2444`

Each keystroke sets `isRunning`, so the region's text goes `'3 results'` → `''` → `'3 results'`. A live region treats the return to the same string as a change and announces it again — which is the talking-over-the-user's-typing the comment says it wants to avoid; without the blanking, an unchanged count would produce no announcement at all. Returning the *previous settled* announcement while a search is in flight keeps the debounce protection without the repeat.

### Checked and clean

- The `boxel-sr-only` spans can't shift layout: the utility is `position: absolute`, and the file has no `:first-child`/`nth-child`/sibling selectors over the affected containers, nor bare element selectors in the scoped CSS that could out-specify `@layer utilities`.
- `SEARCH_HOTKEY_LABEL` as a module-scope const does resolve in the strict-mode template, same mechanism as the existing `SearchIcon`/`ActivityIcon` imports.
- `progressMilestone` clamps correctly at both ends, including `jobPct > 100`.
- Test helpers follow the existing conventions.

🤖 Reviewed by Claude (Opus 5)

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Fixed in b813a63 — dropped `aria-controls` (and `aria-expanded`, which a textbox role doesn't honour anyway) together with the now-orphaned `id`. The results are click-only, so the input stays a plain textbox and the `role='status'` region carries the state; nothing left to dangle.

packages/base/workspace.gts
lukemelia: [Claude Code 🤖] Addressed in b813a63. `loadJobs` re-runs on every index event, so it now detects the running→idle edge and sets a one-shot `Setup complete` that the region reads once `runningJobs` is empty. It's tracked state rather than derived from the job set, so it only fires on a transition observed this session — never on the initial load of a space whose jobs finished in a past session (which a purely-derived version would have mis-announced).

lukemelia:
[Claude Code 🤖] Thanks for the thorough pass, @burieberry — all four addressed in b813a63:

1. **Blur announcing "No matching cards" for a match.** The announcement is now settled tracked state, set only where a search actually settles and cleared on every dismissal path (blur/`hideResults`, `Escape`, `openResult`, and the never-ran `!term || !store` branch you flagged). Blur now falls silent instead of reporting a no-match it never checked. Added a live-search test that renders the card against a real realm and pins match → `1 result` → blur → silent end to end.

2. **Hardcoded `id`.** Dropped the literal `id` entirely (see #3), so there's nothing to collide when two Workspaces render at once.

3. **`aria-controls` / `aria-expanded` on a textbox.** Dropped both. Since the results are click-only I went with the "rely on the status region" option rather than building out a full combobox — and removing them also removed the `id`'s only referrer, which is what let me delete it for #2.

4. **`isRunning` re-announcing an unchanged count.** Same settled-field refactor. In-flight searches no longer blank-and-restore the region, so an unchanged count isn't re-spoken; the previous settled answer stays put until the next settle.

Also extracted `describeSearchResults` as an exported pure function with unit coverage, matching the file's existing `progressMilestone` / `searchHotkeyLabel` style.

`ember-tsc --noEmit` (host), eslint/prettier, and ember-template-lint are all clean. I couldn't run the browser suite locally (it needs the realm-server/matrix stack), so the live-search test's runtime behavior is on CI to confirm.

burieberry (approved):

PR #5624, opened 2026-07-27
Workspace: surface realm indexing as first-class Activity events

lukemelia:
Surfaces **realm indexing** passes in the Workspace **Activity** feed as first-class events, so an operator watching a realm come online sees indexing reflected live. (CS-12121, sibling of CS-12123.)

> **Stacked on #5622** (CS-12123). Review/merge that first; this PR targets `cs-12123-reflect-remix-activity` and its diff is only the indexing work on top.

## What changed

The feed previously showed only card saves. It now **merges two event sources**, reverse-chron by timestamp:

1. **Card saves** — from the realm search, as before.
2. **Completed indexing passes** — from the realm subscription. These are *not* indexed cards, so they accumulate in their own bounded log rather than coming from a query.

Day headers, `when`/`absolute`, the 20-item reveal, and the 100-item cap are all computed over the merged, capped list. `FeedRow.kind` discriminates a card row (renders the embedded card + click-to-open) from a card-less index row (renders an icon tile).

- **`indexEventTitle`** maps an index event to its feed title: incremental → `N cards reindexed`, full → `Full reindex`, copy → `Copied from <host>`. The pre-index `incremental-index-initiation` signal (no committed state yet) and any non-index event are skipped.
- Index rows carry a new **`Indexed`** verb, a database icon tile, and distinct styling.

### Event-source decision
Per the ticket, the source was the open question. This reuses the **existing realm subscription** — the same `subscribeToRealm` → `refreshOnIndex` path that already refreshes the panels now also logs the pass (stamped at receipt, since the event carries no timestamp). No new plumbing, and it composes with CS-12123's remix events through the same merged view-model.

### Preserved behavior
The feed still defers loading until the Activity tab opens and honors the cap; card-save rows are unchanged.

## Tests
- **Unit** (`workspace-unit-test`): `indexEventTitle` for incremental (plural/singular), full, copy, and the initiation-signal/no-op case.
- **Acceptance** (`workspace-test`): delivers a realm index event over the realm's matrix room and asserts the feed renders it as a first-class `Indexed` event (verb, event tile, and "2 cards reindexed" title).

## Verification
- glint (`ember-tsc --noEmit`): 0 errors
- `ember-template-lint`: clean
- prettier: test files clean (a few `{{if}}` blocks in `workspace.gts` are kept multi-line to match `main`, as `packages/base` has no CI prettier check and the plugin version drifts)

The host acceptance/integration suites need the env-mode realm+matrix stack, which wasn't run locally — CI exercises them. The acceptance test's event-routing (matrix room → `subscribeToRealm` → feed) is the piece I'd most want CI to confirm.

### DIFF
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -32,6 +32,7 @@ import LayoutGridIcon from '@cardstack/boxel-icons/layout-grid';
 import ActivityIcon from '@cardstack/boxel-icons/activity';
 import DoorOpenIcon from '@cardstack/boxel-icons/door-open';
 import SearchIcon from '@cardstack/boxel-icons/search';
+import DatabaseIcon from '@cardstack/boxel-icons/database';
 
 import {
   chooseCard,
@@ -72,7 +73,7 @@ import {
   type BoxComponent,
 } from './card-api';
 import { MarkdownDef } from './markdown-file-def'; // realm README
-import type { RealmEventContent } from './matrix-event';
+import type { RealmEventContent, IndexRealmEventContent } from './matrix-event';
 import { Spec } from './spec';
 
 // This file is always loaded through the Boxel loader, which supplies
@@ -132,8 +133,8 @@ export function etaMinutes(
 
 // The verbs the Activity feed labels an event with. A card save is classified
 // as Created/Updated by timing; a RemixCard instance is a first-class Remixed
-// event regardless of when it was written.
-export type ActivityVerb = 'Created' | 'Updated' | 'Remixed';
+// event; a completed realm indexing pass is a first-class Indexed event.
+export type ActivityVerb = 'Created' | 'Updated' | 'Remixed' | 'Indexed';
 
 // A card modified within this window of its creation reads as "Created" in
 // the Activity feed rather than "Updated".
@@ -169,6 +170,54 @@ export function activityVerbFor(
     ? 'Remixed'
     : classifyActivityVerb(modMs, createdMs);
 }
+
+// The type label shown beside the Indexed verb in the feed.
+const INDEX_TYPE_NAME = 'Indexing';
+
+// A completed realm indexing pass, described for the Activity feed. Returns
+// undefined for realm events that are not first-class feed entries — the
+// pre-index 'incremental-index-initiation' signal (no committed state yet) and
+// any non-index event — so callers can skip them.
+export function indexEventTitle(ev: RealmEventContent): string | undefined {
+  if (ev.eventName !== 'index') {
+    return undefined;
+  }
+  switch (ev.indexType) {
+    case 'incremental': {
+      let n = ev.invalidations?.length ?? 0;
+      return n === 1 ? '1 card reindexed' : `${n} cards reindexed`;
+    }
+    case 'full':
+      return 'Full reindex';
+    case 'copy': {
+      let host = hostOf(ev.sourceRealmURL);
+      return host ? `Copied from ${host}` : 'Copied from another realm';
+    }
+    default:
+      return undefined; // initiation signal and any future index types
+  }
+}
+
+// A single Activity-feed row. `kind` discriminates a card save (which carries a
+// `component`/`card` to render and open) from a card-less index event (rendered
+// as an icon tile). The time fields are derived from `ms` when the merged feed
+// is assembled.
+type FeedRow = {
+  kind: 'card' | 'index';
+  id: string;
+  ms: number | undefined;
+  when: string | undefined;
+  absolute: string | undefined;
+  dayLabel: string;
+  showDay: boolean;
+  verb: ActivityVerb;
+  title: string | undefined;
+  typeName: string;
+  typeIcon: typeof CardDef.icon;
+  note: string | undefined;
+  component: BoxComponent | undefined; // card kind only
+  card: CardDef | undefined; // card kind only
+};
 
 type RealmConfigCard = CardDef & { iconURL?: string }; // RealmConfig shape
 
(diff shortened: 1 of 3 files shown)


packages/base/workspace.gts
```diff
@@ -3291,48 +3391,88 @@ class Isolated extends Component<typeof Workspace> {
...
+      prevDay = day || prevDay;
+      return {
+        ...row,
+        when: row.ms !== undefined ? relativeTime(row.ms) : undefined,
+        absolute:
+          row.ms !== undefined ? new Date(row.ms).toLocaleString() : undefined,
+        dayLabel: day,
+        showDay,
+      };
+    });
```
habdelra, line 3456: [Claude Code 🤖]

**Non-blocking (pre-existing, now load-bearing): the feed `{{#each}}` has no `key`, so every recompute re-renders all visible embedded cards.**

**What the getter emits.** This trailing `.map` mints a brand-new row object for every row on each recompute (`{ ...row, when, absolute, dayLabel, showDay }`), so `feed` never returns stable object identities — not even for card rows whose underlying `cardFeed` entry is unchanged. The template iterates it with `{{#each this.visibleFeed as |item|}}` and no `key`, so Ember reconciles by object identity and tears down + rebuilds every visible row — including each `<Embedded @format='embedded' />` — whenever `feed` recomputes.

**Why this PR makes it bite.** Before, the rows were rebuilt roughly once per index burst, because the only thing that rebuilt them was the debounced `loadFeed` (`refreshAfterIndex` coalesces a burst via `INDEX_REFRESH_DEBOUNCE_MS`). This PR adds `recordIndexEvent`, which `unshift`s the tracked `indexLog` once per event, outside that debounce. So a burst of index events — exactly the "operator watching a realm come online" case this PR targets — recomputes `feed` and re-renders every visible embedded card once per event (up to `feedShown`, i.e. 20–100 cards).

**The fix.** Give the each a stable key, which this repo already does for card lists (`packages/base/components/card-list.gts:173,266`): `{{#each this.visibleFeed key='id' as |item|}}`. Every `FeedRow` has a unique `id`, so Ember reconciles by id and reuses the DOM for unchanged rows across recomputes; the extra recomputes then become cheap regardless of burst size.

Non-blocking; the merge logic itself is correct.

---

packages/base/workspace.gts
```diff
@@ -170,6 +171,54 @@ export function activityVerbFor(
...
+// any non-index event — so callers can skip them.
+export function indexEventTitle(ev: RealmEventContent): string | undefined {
+  if (ev.eventName !== 'index') {
+    return undefined;
+  }
+  switch (ev.indexType) {
+    case 'incremental': {
+      let n = ev.invalidations?.length ?? 0;
+      return n === 1 ? '1 card reindexed' : `${n} cards reindexed`;
+    }
```
habdelra, line 192: [Claude Code 🤖]

**Non-blocking accuracy: "N cards reindexed" counts every invalidated URL, not just cards.**

`n` is `invalidations.length`, but `invalidations` is the realm's full invalidation set, not a card count. Traced server-side: in `packages/runtime-common/realm.ts`, `broadcastIncrementalInvalidationEvent` sends whatever `invalidateURLs` accumulated in its `onInvalidation` callback, which adds **every** invalidated URL to the set — `handleExecutableInvalidations(invalidatedURLs)` runs over that same set (realm.ts ~1470–1476). Upstream, `realm-index-updater.ts:295` only strips `.json` from instance hrefs; module hrefs (`.gts`/`.ts`) pass through unchanged. So saving a shared module invalidates the module **plus** its instances, and the module URL gets counted as a "card" (the count is off by the number of modules touched). A no-op incremental pass with an empty set renders "0 cards reindexed".

Minor — the number is indicative and this is an activity line, not a metric. If you want it exact, filter to card URLs before counting; otherwise softening the noun to "N items reindexed" sidesteps the mislabel. Follow-up, not blocking.

---

packages/base/workspace.gts
```diff
@@ -3373,11 +3513,10 @@ class Isolated extends Component<typeof Workspace> {
       } as Query,
       [realm],
     ); // Search only the Card Grid instance's realm.
-    this.feedItems.splice(0, this.feedItems.length);
+    this.cardFeed.splice(0, this.cardFeed.length);
```
habdelra, line 3519: [Claude Code 🤖]

**Confirmation / guard rail: clearing only `cardFeed` here is load-bearing — leave it that way.**

This deliberately resets `cardFeed` and leaves `indexLog` untouched, which is exactly what lets a completed indexing pass survive every subsequent card refresh: `loadFeed` re-runs on each index burst (via `refreshAfterIndex`) and on tab open, and the `feed` getter re-merges the two sources each time. If a later change "tidies up" by also clearing `indexLog` here (or by folding the index log into `loadFeed`), index rows would silently vanish on the next refresh. Nothing to change — noting the invariant for the next editor.

---

habdelra (commented):
[Claude Code 🤖]

Reviewed for how the new second event source composes with the existing card feed — merge ordering, the cap/persistence invariants, event filtering, and the render cost of a live burst — plus the accuracy of the user-facing labels. Checked out the head branch and read the feed getter, the realm-subscription path, and the realm-server side that stamps `invalidations`.

**Bottom line: no blocking issues. The design is sound — two sources merged reverse-chron, the index log kept separate so it survives card refreshes, both bounded to the cap.** Two non-blocking notes (a live-burst re-render cost with a one-line fix, and a label-count accuracy nit) plus one guard-rail confirmation, all inline.

**What lands right:**
- Keeping index passes in their own `indexLog` rather than the card search is the right call — they aren't indexed instances, they survive `loadFeed` reloads, and the `@cached feed` getter re-derives day headers / `when` / `absolute` over the final merged order so ordering stays correct as either source changes.
- Event filtering is tight and doubly guarded: `refreshOnIndex` only records `incremental`/`full`/`copy`, and `indexEventTitle` independently returns `undefined` for the `incremental-index-initiation` signal and any non-`index` event (`prerender_html`, `update`), so pre-commit signals never surface a row.
- The acceptance test delivers the event the way the realm-server does (realm room → `subscribeToRealm`) and asserts on `.feed-verb.indexed` with `exists` rather than a count, so a stray real incremental pass can't flake it.

**Recommendations:**
1. Add `key='id'` to the feed `{{#each}}` so a live burst of index events doesn't re-render every visible embedded card — detail on the `feed` getter. (non-blocking)
2. Decide whether "N cards reindexed" should exclude non-card invalidations — detail on `indexEventTitle`. (non-blocking)

**Adjacent, out of scope:** `feed`'s `when` / `dayLabel` are computed with `Date.now()` inside a `@cached` getter, so relative times ("2m ago", the Today→Yesterday rollover) don't refresh until the next event or reload. That's pre-existing — the old `loadFeed` snapshotted the same fields — so not this PR's to fix; noting it for whoever revisits the feed.

Note on stacking: this targets another open branch rather than the default branch, so the review covers only the indexing commits on top; that base should land first.

---

habdelra (approved):

PR #5626, opened 2026-07-28
Add a migration for realm index cards still on the legacy default index

lukemelia:
## What

`packages/realm-server/scripts/migrate-index-to-workspace.ts` moves a realm's index card off the legacy default index and onto `Workspace`. The committed realms in this repo were migrated by hand in #5610; this is the tool for the realms that only exist on a realm server's disk.

A realm's index card is the `index.json` at its root. Three adoptions are treated as the legacy default:

- `CardsGrid`, as `@cardstack/base/cards-grid` or an absolute `…/base/cards-grid` URL.
- `IndexCard`, as `@cardstack/base/index` or an absolute `…/base/index` URL — `packages/base/index.gts` is a one-line re-export of `CardsGrid` under that historical name, so a realm adopting it is on CardsGrid too.

Absolute URLs are matched on the `/base/<export>` path rather than a fixed host, so the deployment-URL spellings that predate the RRI migration are caught alongside `https://cardstack.com/base/…`.

Anything else is reported and left alone. Relative specifiers get their own verdict rather than being folded in with the rest: inside the base realm `./cards-grid` *is* the default index, but in a user realm it names that realm's own module, and nothing on disk distinguishes the two.

## What it preserves

Only `data.meta.adoptsFrom` is rewritten. Attributes and relationships carry over untouched, and anything a file holds beyond the adoption is named in the report so it can be eyeballed before the run is applied.

Each realm keeps whichever module form it already used — the publish handler's `DEFAULT_REALM_INDEX_ADOPTIONS` recognizes both the `@cardstack/base/` prefix and the absolute base-realm URL, so normalizing them would be churn.

Formatting is preserved too, which matters more than it sounds: card JSON on disk is one line plus a trailing newline, and a naive "does this file contain a newline" check reads that as pretty-printed and reflows the whole file. The rewrite reflows nothing.

## Rollback

An applied run writes a manifest holding each changed file's original bytes; `--rollback <manifest>` restores them. It is plain JSON read and written by node, so undoing a run needs nothing installed in the container beyond the runtime that produced it.

## Publish snapshots

`_published/<disk-id>` is skipped by default, the way `lib/realm-registry-backfill.ts` skips that directory when walking owners — it is a flat snapshot tree, not a username. A snapshot is regenerated from its source realm on the next publish, so migrating the source and republishing is what keeps the two in step. `--published` opts into them for the case where that isn't wanted.

## Hand-rolled workspaces

Some realms have an index card that is a copy of the Workspace card living in the realm itself. Those adopt a realm-local module, so no rule distinguishes them from a genuinely bespoke index — `--include <realm>` names them explicitly, and an `--include` that matches no scanned realm fails the run rather than passing silently.

## Reindex

The deployed server does not watch EFS, so an on-disk edit is invisible to the index until a reindex runs. Every report ends with the realm paths to feed `/_grafana-reindex?realm=<path>`.

## Test plan

`Unit | migrate-index-to-workspace` — 19 passed, 0 failed. Covers module-form recognition, each classification, the preserved-data and preserved-formatting guarantees, the two-level owner walk vs. the flat `_published` walk, apply/rollback round-trip, `--include`, and unreadable or missing index JSON.

Verified end-to-end against a live `dev-all` stack on two local realms — one plain `CardsGrid`, one `IndexCard` carrying a leftover `cardsGrid` relationship for a field neither card defines, which was the case most likely to error-index. Both reindexed with 0 file errors and 0 instance errors, landed in `boxel_index` with `_cardType` `Workspace`, and produced isolated and fitted prerendered HTML with no error doc. Rolling back and reindexing restored both to `Cards Grid`, and the restored files were byte-identical to the originals.

eslint, prettier and `tsc --noEmit` clean on the changed files.

### DIFF
--- a/packages/realm-server/scripts/migrate-index-to-workspace.ts
+++ b/packages/realm-server/scripts/migrate-index-to-workspace.ts
@@ -72,7 +72,7 @@
  *     --rollback ./migrate-index-to-workspace-manifest.json
  */
 import { readFileSync, writeFileSync, readdirSync, existsSync } from 'fs';
-import { join, resolve, sep } from 'path';
+import { join, relative, resolve, sep } from 'path';
 import { fileURLToPath } from 'url';
 import { PUBLISHED_DIRECTORY_NAME } from '@cardstack/runtime-common/constants';
 
@@ -250,6 +250,13 @@ type RealmReport = {
   forced: boolean;
   willMigrate: boolean;
   extras: string[];
+  /**
+   * The realm's path relative to the realms root it was discovered under —
+   * `<user>/<realm>`, which is what `/_grafana-reindex?realm=` wants. Absent
+   * for a realm named as an explicit directory argument, where the disk path
+   * says nothing about the URL the realm is served at.
+   */
+  reindexPath?: string;
 };
 
 type ManifestEntry = { path: string; original: string };
@@ -301,56 +308,42 @@ export function discoverPublishedRealms(realmsRoot: string): string[] {
   return childDirs(publishedRoot).map((diskId) => join(publishedRoot, diskId));
 }
 
-function classifyRealm(realmDir: string): RealmReport {
+function classifyRealm(
+  realmDir: string,
+  reindexPath: string | undefined,
+): RealmReport {
   let indexPath = join(realmDir, 'index.json');
+  let base = {
+    realmDir,
+    indexPath,
+    forced: false,
+    willMigrate: false,
+    extras: [] as string[],
+    reindexPath,
+  };
+  let unusable = (reason: string): RealmReport => ({
+    ...base,
+    classification: { kind: 'unusable', reason },
+  });
+
   if (!existsSync(indexPath)) {
-    return {
-      realmDir,
-      indexPath,
-      classification: { kind: 'unusable', reason: 'no index.json' },
-      forced: false,
-      willMigrate: false,
-      extras: [],
-    };
+    return unusable('no index.json');
   }
   let source: string;
   try {
     source = readFileSync(indexPath, 'utf8');
   } catch (err) {
-    return {
-      realmDir,
-      indexPath,
-      classification: {
-        kind: 'unusable',
-        reason: `unreadable: ${(err as Error).message}`,
-      },
-      forced: false,
-      willMigrate: false,
-      extras: [],
-    };
+    return unusable(`unreadable: ${(err as Error).message}`);
   }
   let doc: any;
   try {
     doc = JSON.parse(source);
   } catch (err) {
-    return {
-      realmDir,
-      indexPath,
-      classification: {
-        kind: 'unusable',
-        reason: `invalid JSON: ${(err as Error).message}`,
-      },
-      forced: false,
-      willMigrate: false,
-      extras: [],
-    };
+    return unusable(`invalid JSON: ${(err as Error).message}`);
   }
   return {
-    realmDir,
-    indexPath,
+    ...base,
     classification: classify(doc?.data?.meta?.adoptsFrom),
-    forced: false,
-    willMigrate: false,
     extras: extraKeys(source),
   };
 }
@@ -479,14 +472,29 @@ export function main(argv: string[]): number {
     return rollback(args.rollback);
   }
 
-  let realmDirs = [...args.dirs];
+  // Each discovered realm remembers the realms root it came from, so its
+  // `<user>/<realm>` path — the value `/_grafana-reindex?realm=` takes — can be
+  // printed alongside the disk path. An explicit directory argument gets none:
+  // its disk path implies nothing about the URL the realm is served at.
+  let discovered = new Map<string, string | undefined>();
+  for (let dir of args.dirs) {
+    discovered.set(dir, undefined);
+  }
   for (let root of args.realmsRoots) {
-    realmDirs.push(...discoverUserRealms(root));
-    if (args.published) {
-      realmDirs.push(...discoverPublishedRealms(root));
+    let fromRoot = [
+      ...discoverUserRealms(root),
+      ...(args.published ? discoverPublishedRealms(root) : []),
+    ];
+    for (let dir of fromRoot) {
+      if (!discovered.has(dir)) {
+        discovered.set(dir, relative(root, dir));
+      }
     }
   }
-  realmDirs = [...new Set(realmDirs)].filter((dir) => isDirectory(dir)).sort();
+
+  let realmDirs = [...discovered.keys()]
+    .filter((dir) => isDirectory(dir))
+    .sort();
 
   if (realmDirs.length === 0) {
     console.error(
@@ -495,7 +503,7 @@ export function main(argv: string[]): number {
     return 1;
   }
 
-  let reports = realmDirs.map(classifyRealm);
+  let reports = realmDirs.map((dir) => classifyRealm(dir, discovered.get(dir)));
   let unmatchedIncludes = new Set(args.include);
   for (let report of reports) {
     let hit = args.include.find((include) =>
@@ -547,6 +555,7 @@ export function main(argv: string[]): number {
           manifest: args.dryRun || manifest.length === 0 ? null : args.manifest,
           realms: reports.map((report) => ({
             realmDir: report.realmDir,
+            reindexPath: report.reindexPath ?? null,
             classification: report.classification,
             forced: report.forced,
             willMigrate: report.willMigrate,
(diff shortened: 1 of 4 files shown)


lukemelia:
[Claude Code 🤖]

Correcting the test-plan claim in the description above: the new unit suite was **not** running in CI when I wrote that.

CI shards this package by setting `TEST_MODULES`, which `tests/index.ts` turns into a QUnit filter anchored on each file's own name:

```js
const pattern = `^(?:${escaped.join('|')})(?:\\s>\\s|:)`;
```

The suite's outermost module was `Unit | migrate-index-to-workspace`, which matches nothing, so all 20 tests were filtered out. It passed locally the whole time because the local loop uses `TEST_FILES`, which applies no filter.

Evidence: shard 5 listed `migrate-index-to-workspace-test.ts` in its `TEST_MODULES` and produced zero result lines, while the other 26 files in that shard all reported.

Fixed in 4631d34 by naming both suites after their file, as the rest of the package does:

```ts
module(basename(import.meta.filename), function () { … });
```

`searchable-parity-diff-test.ts` is in the same commit — it had the identical problem in shard 2, and is the file this one was modeled on, which is how the mistake propagated. Together the two contribute 37 tests that were running nowhere but a laptop.

Verified through the CI code path rather than the local one:

```
TEST_FILES=migrate-index-to-workspace-test,searchable-parity-diff-test \
TEST_MODULES='migrate-index-to-workspace-test.ts|searchable-parity-diff-test.ts' \
  node tests/index.ts
# pass 37, # fail 0
```

Four more files in this package have the same defect (prerender, CPU profiler, network in-flight tracker). They are **not** in this PR — enabling tests that haven't run in CI for an unknown period could surface real failures, which belongs somewhere a prerender failure is expected signal. Filed as CS-12357 with the evidence and a suggestion to make `tests/index.ts` fail when a `TEST_MODULES` entry matches zero tests, so the class becomes self-detecting.

packages/realm-server/scripts/migrate-index-to-workspace.ts
```diff
@@ -0,0 +1,634 @@
...
+  let prefixMatch = /^(@cardstack\/base\/)([^/]+)$/.exec(module);
+  if (prefixMatch) {
+    return { prefix: prefixMatch[1], segment: prefixMatch[2] };
+  }
+  let urlMatch = /^(https?:\/\/.*\/base\/)([^/]+)$/.exec(module);
+  if (urlMatch) {
+    return { prefix: urlMatch[1], segment: urlMatch[2] };
+  }
+  return undefined;
+}
```
habdelra, line 137: [Claude Code 🤖]

**Non-blocking, and not a regression — but a boundary worth pinning down, because the header comment slightly over-promises it.** The doc up top says the preserved absolute form is safe "because the publish handler … recognize[s] both." That holds only for the two *canonical* spellings, not for every absolute form this regex accepts.

**What the publish handler actually recognizes.** `DEFAULT_REALM_INDEX_ADOPTIONS` in `handlers/handle-publish-realm.ts` is a closed set of exactly four strings:
- `CardsGrid`: `https://cardstack.com/base/cards-grid` | `@cardstack/base/cards-grid`
- `Workspace`: `https://cardstack.com/base/workspace` | `@cardstack/base/workspace`

`isDefaultRealmIndexAdoption` gates whether a published realm gets `includePrerenderedDefaultRealmIndex = true` (handle-publish-realm.ts around the `if (!isDefaultRealmIndexAdoption(adoptsFrom)) return;` check).

**What this regex accepts.** `^(https?:\/\/.*\/base\/)([^/]+)$` is deliberately broader — it catches the deployment-URL spellings (`https://realms-staging.stack.cards/base/cards-grid`, `https://app.boxel.ai/base/cards-grid`), which is the stated point. A realm migrated from one of those lands on e.g. `https://realms-staging.stack.cards/base/workspace`, and that string is **not** in `DEFAULT_REALM_INDEX_ADOPTIONS`, so on publish it won't get the fast-path flag.

**Why it's still fine.** This is not a regression: the `cards-grid` deployment-URL spelling wasn't recognized *before* the migration either, so those realms were already outside the publish fast-path — the migration leaves them exactly where they were, and it's symmetric. Runtime resolution is unaffected (the same `/base/` prefix that resolved `cards-grid` resolves `workspace`).

Verified empirically against a temp realm tree: `https://realms-staging.stack.cards/base/cards-grid` → `https://realms-staging.stack.cards/base/workspace` (host preserved, in place), and that output is absent from the recognized set.

**Two things worth a word in the comment so the next editor doesn't trip:**
1. Narrow "recognize both" to the `@cardstack/base/` and `https://cardstack.com/base/` forms — deployment-URL-spelled realms migrate correctly but stay outside the publish fast-path (as they were).
2. Matching on the `/base/` path rather than a host allowlist also means a *non-base* URL that merely contains `/base/` (e.g. `https://app.boxel.ai/some-realm/base/cards-grid`) would be treated as a base module. Contrived, and the dry-run-first workflow surfaces it, but it's the flip side of the path match.

Scope: pre-existing / confirmation. No code change required.

---

habdelra (commented):
[Claude Code 🤖]

Reviewed as a one-time, on-disk data-mutation tool: the lens is (1) does it ever rewrite a realm it shouldn't, (2) is the apply/rollback safety net bulletproof, and (3) is the CI-visibility fix real. I checked out the head branch and verified the behavioral claims by running the script against a temp realm tree and by reading the referenced handlers.

**Bottom line: no blocking issues.** This is a careful, well-tested migration with a genuine safety net, and the "these tests weren't actually running in CI" fix is correct and independently verified. Everything below is non-blocking.

**What lands right (verified, not skimmed):**
- **Classification is defensive in the ways that matter.** `classify` requires *both* module segment and export name to match before calling something legacy, so `@cardstack/base/cards-grid` + name `Renderer` falls through to `bespoke` and is left alone. Relative specifiers get their own verdict and are never rewritten — correct, since `./cards-grid` means the base default inside the base realm but a realm-local module elsewhere, and nothing on disk tells them apart.
- **The safety net is byte-perfect.** Applied a run over four temp realms (canonical CardsGrid, a deployment-URL-spelled CardsGrid, the `IndexCard` alias, and a bespoke `@cardstack/catalog/blog`), then rolled back: all four files were sha256-identical to their originals. Only `data.meta.adoptsFrom` changed; a compact single-line-plus-newline file stayed exactly that (no reflow), confirming the `trimEnd().includes('\n')` formatting guard.
- **The CI-filter fix is the real bug fix here.** `buildModuleFilter` (tests/index.ts) anchors the shard filter as `^(?:<module>)(?:\s>\s|:)`, so the outermost `module()` name must equal the file's basename for `TEST_MODULES` to match it. The old top module `Unit | migrate-index-to-workspace` matched nothing → the whole file was silently filtered out in CI while passing locally (where `TEST_FILES` applies no filter). `module(basename(import.meta.filename), …)` fixes the anchor. I confirmed exactly four files still carry this defect — `cpu-profiler-affinity-gate-test`, `network-inflight-tracker-test`, `prerender-artifact-sink-test`, `prerender-v8-prof-test` — matching the "prerender, CPU profiler, network in-flight tracker" set scoped out as a follow-up. 118 files already use the convention.
- **The supporting claims all check out.** `PUBLIC_REALM_DIRS` matches the `--path` args in both `start-staging.sh` and `start-production.sh` exactly (8/8, `submissions` via `SUBMISSION_REALM_PATH`); the `_published` skip mirrors `realm-registry-backfill.ts`; `forcedTarget()` returns `@cardstack/base/workspace`, the form `create-realm.ts` stamps on new realms; and `packages/base/index.gts` is indeed a one-line re-export of `CardsGrid` as `IndexCard`.

**Non-blocking notes:**
1. The `splitBaseModule` `/base/`-path match is broader than the publish handler's recognized set — deployment-URL-spelled realms migrate correctly but stay outside the publish fast-path (pre-existing, symmetric, no regression), and the path match will also accept a non-base URL containing `/base/`. Mechanism + evidence in the inline thread on `splitBaseModule`. A one-sentence narrowing of the header comment's "recognize both" is all that's suggested.
2. **Rollback manifest, narrow edge:** if the same realm is reached both via `--realms-root` discovery and as an explicit arg with a different spelling (e.g. a trailing slash), it survives the `discovered` Map dedup (keyed by exact string) as two entries pointing at one `index.json`. The apply loop would then push a second manifest entry whose "original" is the already-migrated bytes, so rollback wouldn't restore that one path. The dry-run lists the realm twice (a visible tell) and it's an unusual invocation, but since the manifest is the whole safety story, de-duping `realmDirs` by resolved path would close it cheaply. Low priority.
3. **Apply-by-default:** for a tool that mutates realm data across potentially hundreds of realms, defaulting to apply (dry-run opt-in) is a surprising default even though it's documented and recoverable via the manifest. Flagging only to confirm the tradeoff is intended — the reindex being a separate manual step means an accidental apply sits invisible-to-the-index until then. Not asking for a change.

**Adjacent, out of scope:** the four test files above have been silently unfiltered-nowhere in CI for an unknown period; the author has already filed that as a follow-up with a good suggestion (make `tests/index.ts` fail when a `TEST_MODULES` entry matches zero tests, so the class self-detects). Verified the claim; leaving it to that follow-up.

---

PR #5627, opened 2026-07-28
Add StlDef FileDef subclass for .stl 3D meshes

FadhlanR:
Resolves [CS-12053](https://linear.app/cardstack/issue/CS-12053/add-stl-filedef-subclass).

Registers `.stl` (3D-print mesh) as a first-class file type, the way every other file extension already is.

## What changed
- **`packages/base/stl-file-def.gts`** — `StlDef extends FileDef`: cube icon, `'3D Model (STL)'` display name, isolated/embedded/fitted/atom formats, and a single `format` field (`binary` | `ascii`).
- **`packages/base/stl-meta-extractor.ts`** — `extractStlFormat(bytes, contentSize)`. Classifies the encoding from the first 84 bytes plus the file size: the binary size identity `84 + 50·count === contentSize`, else an ASCII `solid …` header. Non-STL bytes throw `FileContentMismatchError` so a mislabeled/truncated file indexes as a bare `FileDef` instead of erroring the row. Isolated from the component for unit testing, mirroring `mp3-meta-extractor.ts`.
- **`packages/runtime-common/file-def-code-ref.ts`** — maps `.stl → StlDef`, so `urlNamesFile()` / `resolveFileDefCodeRef()` route a `linksTo` a `.stl` through the file-meta MIME path rather than mis-dispatching it as a card.
- **`packages/runtime-common/infer-content-type.ts`** — `isBinaryFilename()` now treats `model/stl` (and legacy `application/vnd.ms-pki.stl`) as binary. `.stl` matched no MIME prefix before, so binary STL was read as UTF-8 text and its bytes mangled in boxel-cli push/pull/read.

## Extract scope
Only `format` is stored. The extractor reads the header triangle count *internally* to run the binary/ASCII sniff but does not persist it — `triangleCount` isn't a real search key and carries no display value on its own. `extractAttributes` runs per-file on the indexing hot path, so the read is bounded to an 84-byte `readFirstBytes` (like `mp3-audio-def`); a full-mesh parse (bounding box, volume) is deliberately deferred off the index path.

## Tests
- `packages/host/tests/integration/components/stl-file-def-test.gts` — `extractStlFormat` (binary / ASCII / whitespace-ASCII / non-STL) and `StlDef.extractAttributes` (format stamping, inherited base fields, extension guard).
- `packages/host/tests/acceptance/stl-file-def-test.gts` — drives the `file-extract` render route for binary + ASCII fixtures, the mismatch fallback, and the indexing → file-meta path resolving to `StlDef` (mirrors `csv-file-def-test.gts`).
- `packages/host/tests/unit/file-def-code-ref-test.ts` — extends `isFileDefCodeRef` coverage for `StlDef`.

## Follow-ups (out of scope)
- Interactive 3D viewer (needs a WebGL/model-viewer dep; the isolated format also renders during prerender).
- Bounding-box dimensions for size-range filtering — deferrable behind an opt-in flag or a lazy viewer path, off the index hot path.
- Text-search value from 3D content is a **3MF** story, tracked separately.

### DIFF
--- a/packages/base/stl-file-def.gts
+++ b/packages/base/stl-file-def.gts
@@ -1,0 +1,326 @@
+import { readFirstBytes } from '@cardstack/runtime-common';
+import CubeIcon from '@cardstack/boxel-icons/cube';
+import {
+  BaseDefComponent,
+  Component,
+  StringField,
+  contains,
+  field,
+} from './card-api';
+import {
+  FileContentMismatchError,
+  FileDef,
+  type ByteStream,
+  type SerializedFile,
+} from './file-api';
+import { STL_SNIFF_BYTES, extractStlFormat } from './stl-meta-extractor';
+
+function getExtension(url: string): string {
+  try {
+    let parsed = new URL(url);
+    let name = parsed.pathname.split('/').pop() ?? '';
+    let dot = name.lastIndexOf('.');
+    return dot === -1 ? '' : name.slice(dot).toLowerCase();
+  } catch {
+    let dot = url.lastIndexOf('.');
+    return dot === -1 ? '' : url.slice(dot).toLowerCase();
+  }
+}
+
+function stlTitle(model: { name?: string | null } | null | undefined): string {
+  return model?.name ?? 'Untitled 3D Model';
+}
+
+// Human label for the stored `format` value; blank when the encoding is
+// unknown (e.g. a bare FileDef fallback that never ran STL extraction).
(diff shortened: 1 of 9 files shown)


FadhlanR:
[Claude Code 🤖] Superseded by #5658, which unifies StlDef (CS-12053) and ThreeMfDef (CS-12054) under a shared `ModelDef` base. Closing in favor of that combined PR.

PR #5628, opened 2026-07-28
Add ThreeMfDef FileDef subclass for 3MF (3D printing) files

FadhlanR:
## Background and Goal

Adds `ThreeMfDef`, a `FileDef` subclass for `.3mf` (3D Manufacturing Format / 3D printing) files, so they index as a first-class file type. A `.3mf` is an OPC package (a ZIP) whose `3D/3dmodel.model` XML carries `title` / `designer` / `description` / `license` and a `unit` — searchable text that is **not** present in the filename — so the subclass extracts it and feeds it into full-text search. Distinct icon + light metadata templates; no 3D viewer.

Resolves CS-12054.

## Where to start

- **`packages/base/3mf-meta-extractor.ts`** — the parse. `fflate.unzipSync` with a `filter` that records every entry name (→ `hasThumbnail`) while inflating only the `*.model` part; then decodes ≤256 KB of the model, cuts at `<resources>`/`<build>`, and regex-extracts the metadata + `unit`. Throws `FileContentMismatchError` on a non-ZIP or a ZIP with no model part.
- **`packages/base/3mf-file-def.gts`** — `ThreeMfDef extends FileDef`: the six searchable fields, `extractAttributes` (`super` + `extract3mfMetadata`), `cube-3d-sphere` icon, `acceptTypes`, and metadata-display `isolated`/`embedded` templates.
- **`packages/runtime-common/file-def-code-ref.ts`** — the one-line registry entry that maps `.3mf → ThreeMfDef` and lights up `urlNamesFile` / `isFileDefCodeRef`.

- **Bounded on the indexing hot path.** `extractAttributes` runs for every file during the index/prerender visit and 3MF files can be large, so we inflate only the model part and cap the XML parse at the header (metadata lives above `<resources>`; geometry lives below). Counting objects/triangles or reading the bounding box would require scanning past all vertices, so those are deferred.
- **Graceful fallback.** A malformed/non-3MF file throws `FileContentMismatchError`, so the extractor falls back to a bare `FileDef` and the file still indexes — same pattern as `Mp3Def`/`PngDef`.
- **New dependency: `fflate`** (~8 KB, isomorphic, added to the catalog + `base`). No ZIP/deflate library existed in the repo; `base` already imports npm runtime deps this way (`super-fast-md5`, `awesome-phonenumber`). Also added to `host` so the acceptance test can build fixtures.
- **MIME + binary.** `.3mf → model/3mf` (IANA-registered; `mime-db` has no entry) and classified binary, so the CLI/bot don't treat the ZIP as text.
- **Deferred to follow-ups:** a WebGL/three.js viewer (greenfield + prerender-render risk), the expensive geometry fields (triangle/vertex counts, bounding box, volume, object count), and surfacing the embedded thumbnail as the preview image.

### DIFF
--- a/packages/base/3mf-file-def.gts
+++ b/packages/base/3mf-file-def.gts
@@ -1,0 +1,210 @@
+import { byteStreamToUint8Array } from '@cardstack/runtime-common';
+import Cube3dSphereIcon from '@cardstack/boxel-icons/cube-3d-sphere';
+import {
+  BaseDefComponent,
+  Component,
+  StringField,
+  contains,
+  field,
+} from './card-api';
+import BooleanField from './boolean';
+import {
+  FileDef,
+  type ByteStream,
+  type SerializedFile,
+} from './file-api';
+import { extract3mfMetadata } from './3mf-meta-extractor';
+
+type ThreeMfExtra = {
+  title?: string;
+  designer?: string;
+  description?: string;
+  license?: string;
+  unit?: string;
+  hasThumbnail: boolean;
+};
+
+function threeMfTitle(
+  model: { title?: string | null; name?: string | null } | null | undefined,
+): string {
+  return model?.title ?? model?.name ?? 'Untitled 3MF model';
+}
+
+class Isolated extends Component<typeof ThreeMfDef> {
+  get title() {
+    return threeMfTitle(this.args.model);
+  }
+  <template>
+    <article class='threemf-isolated' data-test-threemf-isolated>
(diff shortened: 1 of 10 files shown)


FadhlanR:
[Claude Code 🤖] Superseded by #5658, which unifies StlDef (CS-12053) and ThreeMfDef (CS-12054) under a shared `ModelDef` base. Closing in favor of that combined PR.

PR #5632, opened 2026-07-28
Add memoisation for `VirtualNetwork#unresolveURL`

backspace:
This is used a lot and can be cached until the mappings change.

### DIFF
--- a/packages/runtime-common/virtual-network.ts
+++ b/packages/runtime-common/virtual-network.ts
@@ -34,6 +34,11 @@ export class VirtualNetwork {
   // a pure function of the realm mappings, so entries stay valid until a
   // mapping is added or removed (both clear the cache).
   private toURLHrefCache = new Map<string, string>();
+  // Memo for unresolveURL, the inverse of toURLHref. It runs on hot store and
+  // indexing paths and each miss pays a native `new URL()` in the virtual→real
+  // mapping chase. Same pure-function-of-mappings contract as toURLHrefCache:
+  // entries stay valid until a realm or URL mapping changes.
+  private unresolveURLCache = new Map<string, RealmResourceIdentifier>();
 
   // Notified whenever a realm-prefix mapping changes — added, removed, or
   // re-registered against a new target. Consumers that key caches by the RRI
@@ -88,6 +93,10 @@ export class VirtualNetwork {
 
   addURLMapping(from: URL, to: URL) {
     this.urlMappings.push([from.href, to.href]);
+    // Both memos resolve through urlMappings (toURLHref via toURL, unresolveURL
+    // via its virtual→real chase), so a new URL mapping invalidates them.
+    this.toURLHrefCache.clear();
+    this.unresolveURLCache.clear();
   }
 
   mapURL(
@@ -118,6 +127,7 @@ export class VirtualNetwork {
     let normalizedTarget = ensureTrailingSlash(targetURL);
     this.realmMappings.set(normalizedId, normalizedTarget);
     this.toURLHrefCache.clear();
+    this.unresolveURLCache.clear();
     this.addImportMap(
       normalizedId,
       (rest) => new URL(rest, normalizedTarget).href,
@@ -136,6 +146,7 @@ export class VirtualNetwork {
     this.realmMappings.delete(normalizedId);
     this.importMap.delete(normalizedId);
     this.toURLHrefCache.clear();
+    this.unresolveURLCache.clear();
     this.notifyMappingChange();
   }
 
@@ -171,6 +182,16 @@ export class VirtualNetwork {
    * Inputs that match no prefix and no URL mapping are returned as-is.
    */
   unresolveURL(url: string): RealmResourceIdentifier {
+    let cached = this.unresolveURLCache.get(url);
+    if (cached !== undefined) {
+      return cached;
+    }
+    let result = this.computeUnresolveURL(url);
+    this.unresolveURLCache.set(url, result);
+    return result;
+  }
+
+  private computeUnresolveURL(url: string): RealmResourceIdentifier {
     for (let [prefix, target] of this.realmMappings) {
       if (url.startsWith(target)) {
         return (prefix + url.slice(target.length)) as RealmResourceIdentifier;
--- a/packages/runtime-common/virtual-network.ts
+++ b/packages/runtime-common/virtual-network.ts
@@ -39,6 +39,24 @@ export class VirtualNetwork {
   // mapping chase. Same pure-function-of-mappings contract as toURLHrefCache:
   // entries stay valid until a realm or URL mapping changes.
   private unresolveURLCache = new Map<string, RealmResourceIdentifier>();
+  // Cap on the URL memos above. A VirtualNetwork is process-long-lived (notably
+  // in the realm server), and toURLHref/unresolveURL are called with distinct
+  // card, index, and request URLs — unique or nonexistent inputs would
+  // otherwise accumulate for the lifetime of the process. On overflow the
+  // oldest entry is evicted (Map preserves insertion order), which suits the
+  // "same identifiers resolved repeatedly" pattern these memos target.
+  private static readonly MAX_URL_CACHE = 20_000;
+
+  private setBoundedCache<V>(cache: Map<string, V>, key: string, value: V): V {
+    if (cache.size >= VirtualNetwork.MAX_URL_CACHE) {
+      let oldest = cache.keys().next().value;
+      if (oldest !== undefined) {
+        cache.delete(oldest);
+      }
+    }
+    cache.set(key, value);
+    return value;
+  }
 
   // Notified whenever a realm-prefix mapping changes — added, removed, or
   // re-registered against a new target. Consumers that key caches by the RRI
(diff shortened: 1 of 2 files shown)


packages/runtime-common/virtual-network.ts
```diff
@@ -171,6 +182,16 @@ export class VirtualNetwork {
    * Inputs that match no prefix and no URL mapping are returned as-is.
    */
   unresolveURL(url: string): RealmResourceIdentifier {
+    let cached = this.unresolveURLCache.get(url);
+    if (cached !== undefined) {
+      return cached;
+    }
+    let result = this.computeUnresolveURL(url);
+    this.unresolveURLCache.set(url, result);
```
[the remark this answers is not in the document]
backspace: 1311c81 adds an LRU cache

packages/runtime-common/virtual-network.ts
```diff
@@ -88,6 +111,10 @@ export class VirtualNetwork {
 
   addURLMapping(from: URL, to: URL) {
     this.urlMappings.push([from.href, to.href]);
+    // Both memos resolve through urlMappings (toURLHref via toURL, unresolveURL
+    // via its virtual→real chase), so a new URL mapping invalidates them.
+    this.toURLHrefCache.clear();
+    this.unresolveURLCache.clear();
```
habdelra: [Claude Code 🤖]

**The stated mechanism is wrong: `toURLHref` does not resolve through `urlMappings`, so clearing `toURLHrefCache` here is unnecessary (harmless, but the reasoning is what a future editor will trust).** Non-blocking.

**Traced.** `toURLHref` → `toURL` → `resolveRRIToURL`. `resolveRRIToURL` iterates `this.realmMappings` only; on no match, `toURL` falls back to `new URL(rri)`. Neither step reads `urlMappings` — that array is read exclusively by `resolveURLMapping`, which `toURL` never calls. So `toURLHref` is a pure function of `realmMappings` alone.

`unresolveURL` is the opposite: it genuinely chases through `urlMappings` (`computeUnresolveURL` → `resolveURLMapping`), so **its** clear on this path is correct and required.

Net: the `unresolveURLCache.clear()` here is necessary; the `toURLHrefCache.clear()` is a safe no-op for correctness (over-invalidation can't return a wrong answer, and `addURLMapping` is setup-time, not hot). Only the comment misleads — it says both memos "resolve through urlMappings (toURLHref via toURL …)", and someone refactoring `toURL`'s resolution or pruning these clears on that premise would reason from a false fact.

Suggested wording (keep the defensive clear if you want it, just say what's true):

```js
this.urlMappings.push([from.href, to.href]);
// unresolveURL chases through urlMappings (via resolveURLMapping), so a new
// URL mapping invalidates its memo. toURLHref resolves only through
// realmMappings, so clearing its cache here is purely defensive.
this.toURLHrefCache.clear();
this.unresolveURLCache.clear();
```

Scope: documentation regression introduced here (the code is safe). Non-blocking.

---

packages/runtime-common/virtual-network.ts
```diff
@@ -34,6 +34,29 @@ export class VirtualNetwork {
...
+  private setBoundedCache<V>(cache: Map<string, V>, key: string, value: V): V {
+    if (cache.size >= VirtualNetwork.MAX_URL_CACHE) {
+      let oldest = cache.keys().next().value;
+      if (oldest !== undefined) {
+        cache.delete(oldest);
+      }
+    }
+    cache.set(key, value);
+    return value;
+  }
```
habdelra, line 59: [Claude Code 🤖]

**Confirmation + one caveat: this is FIFO eviction, not LRU.** `setBoundedCache` evicts `cache.keys().next().value` (the oldest *inserted* key), and the read fast-paths in `unresolveURL`/`toURLHref` return on a hit without re-inserting, so a hit never refreshes recency. The comment above is accurate — "the oldest entry is evicted (Map preserves insertion order)" — so nothing to change here; flagging only because "LRU" is easy to assume for a bounded map and the two behave differently at the margin.

Where it matters: the memos target "same identifiers resolved repeatedly," and the 20k cap is a safety valve against unbounded growth from unique/nonexistent inputs — not a working-set cache. FIFO and LRU diverge only once the *distinct* live working set exceeds 20k, where FIFO can evict a hot early-inserted key that LRU would keep, degrading hit rate silently. Fine for this PR; worth knowing if a single realm-server process ever resolves >20k distinct URLs faster than the hot set recurs.

Bound check (verified): evict-one-before-insert pins size at `MAX_URL_CACHE`, and the `oldest !== undefined` guard is dead-defensive (`size >= 20_000` ⇒ non-empty) but harmless. Scope: confirmation, non-blocking.

---

packages/runtime-common/virtual-network.ts
```diff
@@ -171,6 +200,15 @@ export class VirtualNetwork {
    * Inputs that match no prefix and no URL mapping are returned as-is.
    */
   unresolveURL(url: string): RealmResourceIdentifier {
+    let cached = this.unresolveURLCache.get(url);
+    if (cached !== undefined) {
+      return cached;
+    }
+    let result = this.computeUnresolveURL(url);
+    return this.setBoundedCache(this.unresolveURLCache, url, result);
```
habdelra, line 209: [Claude Code 🤖]

**The memo is correct — here's the why for the next editor, plus the one test that would pin it.**

**Why it's safe.** `unresolveURL` is a pure function of `realmMappings` + `urlMappings`: `computeUnresolveURL` reads `realmMappings` directly and `urlMappings` via `resolveURLMapping`. Both are `private` and mutated at exactly three sites — `addURLMapping`, `addRealmMapping`, `removeRealmMapping` — each of which now clears `unresolveURLCache`. I enumerated them; there is no fourth mutation path. So between mapping changes the result is stable, which is the contract memoization needs.

**Caching misses is safe too.** The no-match branch returns the input `as RealmResourceIdentifier`, and that value is cached like any hit. That's fine precisely because any later mapping that would start matching it also clears the cache. And because `RealmResourceIdentifier` is a branded string (`string & { __rriBrand }`) — an immutable primitive — handing the same cached reference to many callers carries no aliasing hazard.

**Test-coverage gap (follow-up, non-blocking).** Nothing executable guards invalidation. The existing `unresolveURL` tests register all mappings first, then call once — none do call → mutate-a-mapping → call-again, so a future mutation method that forgets a clear (or an accidental removal of one of the three clears) would still pass CI green. A few lines in `canonical-url-memo-test.ts` (its natural home) would close it: `unresolveURL(u)` returns `u` as-is → `addRealmMapping` a prefix that now matches `u` → assert the second call reflects the new prefix; symmetric cases for `removeRealmMapping` and `addURLMapping`. That test is the guard rail for this whole change's correctness.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** the invalidation contract and the eviction/bound. For a pure-function-of-mappings memo those are the only two ways it can go wrong — a stale hit, or unbounded growth — so that's where the review concentrated.

**Bottom line: sound, no blocking issues.** This is a genuine pure-function-of-mappings cache; invalidation is complete (verified — `realmMappings`/`urlMappings` are `private` and mutated at exactly three sites, all three now clear both memos); the cached value is an immutable branded string, so sharing the reference across callers is safe; and the added bound keeps a process-long-lived `VirtualNetwork` from growing without limit. Three non-blocking notes below, none needed before merge.

**What lands right.** Reusing the existing `toURLHrefCache` contract verbatim — same "pure function of mappings, cleared on every mapping change" model — is the right call: one invalidation story for both memos rather than a second bespoke one. Splitting out `computeUnresolveURL` so the public method is just read/compute/write keeps the hot path obvious. And caching the no-match "return as-is" result, not just hits, is correct and is the case that dominates for unregistered URLs.

**On the resolved bound discussion.** The unbounded-growth concern is genuinely addressed — I re-derived the bound and it holds (evict-oldest-before-insert pins size at the cap). One nuance for the record: the eviction is FIFO, not LRU (detail in the `setBoundedCache` thread). Fine for this access pattern.

**Recommendations** (all non-blocking):
1. Correct the `addURLMapping` comment — `toURLHref` resolves only through `realmMappings`, not `urlMappings`; that clear is defensive, not a dependency. See the `addURLMapping` thread.
2. Note the eviction is FIFO, not LRU — the code comment already says so correctly; only the loose "LRU" framing is off. See the `setBoundedCache` thread.
3. Add a call → mutate-mapping → call-again invalidation test; nothing executable guards the three clears today. See the `unresolveURL` thread.

**Adjacent, out of scope.** `computeUnresolveURL` scans `realmMappings` linearly twice (directly, then again after the virtual→real chase). With the memo in front this is now paid once per distinct URL, so it's moot here — noting only for whoever revisits that function next.

---

habdelra (approved):

richardhjtan (approved):

packages/runtime-common/virtual-network.ts
backspace: comment fixed in 3027e3af63

packages/runtime-common/virtual-network.ts
backspace: 665f649975 has new tests for this

PR #5634, opened 2026-07-28
Ellipsize long card-type names in the workspace.gts Library rail

lukemelia:
## The bug

In the Workspace Library's left rail, a long card-type name pushed the row's count and its hover `+` past the rail's right edge, where they were clipped away entirely. `.rail-name` already carried `overflow: hidden; white-space: nowrap; text-overflow: ellipsis`, but the ellipsis never engaged.

## Why the existing ellipsis did nothing

Card-type rows are the only rail rows wrapped in `.rail-row-wrap`, the extra grid level that positions the hover `+`. Nested in that column, `.rail-row` kept its automatic minimum width, so the row sized to its content rather than to the rail. Text can only ellipsize once something forces it to give up space; nothing did, because the row was free to grow.

Measured on an isolated reproduction of the rail's markup and CSS, with a 214px rail (`clientWidth` 213):

| | before | after |
|---|---|---|
| row width | 320px | 193px |
| rail `scrollWidth` | 330px | 213px |
| name ellipsized | no | yes |
| count right edge | 300px | 173px |
| `+` right edge | 325px | 198px |

Everything past 214px was invisible, which is why both the count and the `+` disappeared — including on short rows, since the whole grid column had taken the widest row's width.

## The fix

`min-width: 0` on `.rail-row`. That lets the row take the rail's width instead of its content's, which is the condition the existing ellipsis was waiting on.

Worth noting what did *not* work: `min-width: 0` on `.rail-name` changes nothing, because `overflow: hidden` already zeroes that element's automatic minimum. The constraint has to be on the row.

This is a containment fix rather than a fixed `max-width` on the name, so it holds at any rail width and adapts to the count: the name shrinks further when the count is several digits wide, keeping the count and `+` inside the rail. A hard `max-width` would need re-tuning whenever the rail width, icon size, or count width changed.

## Scope

The rail's other two groups — Library filters and File types — render `.rail-row` directly with no wrapper, and already ellipsized correctly. Measured before and after: unchanged at 193px and ellipsizing in both cases, so the added property is a no-op there. It's on `.rail-row` rather than scoped to `.rail-row-wrap` so a future wrapped group can't reintroduce this.

## Test plan

Verified in Chrome against an isolated page built from this file's actual rail CSS — extracted from the edited source, not hand-copied — across four rows: a short name, two long names, and a long name paired with a five-digit count. In every case there is no horizontal overflow, long names ellipsize, and the count and `+` are fully within the rail with no overlap between them. Before/after screenshots confirm it visually.

No automated regression test. Card-type rows only render once `cardTypeFilters` is populated from the realm's `_types` endpoint, and the host test helpers don't serve it — `Integration | Card | workspace | segments` exercises only the unwrapped rows, which don't reproduce this. Adding one means stubbing `_types` in the host helpers, which is test-infrastructure work rather than part of a one-line CSS fix; it fits better alongside the broader Workspace coverage work. Happy to add it here if preferred.

`ember-template-lint` and `prettier` clean. `packages/base/workspace.gts` trips an eslint parse error at line 348 — verified identical on an untouched copy of `main`, roughly 950 lines from this change, and `packages/base` is template-lint-only in CI.

### DIFF
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -1311,6 +1311,12 @@ class Isolated extends Component<typeof Workspace> {
         font: 500 13px var(--grid-sans);
         color: var(--grid-nav-ink);
         cursor: pointer;
+        /* Without this the row keeps its max-content width instead of the
+           rail's, so a long type name pushes the count and the + past the
+           rail's right edge rather than ellipsizing. The ellipsis on
+           .rail-name only engages once the row itself is allowed to be
+           narrower than its content. */
+        min-width: 0;
       }
       .rail-row.type {
         padding: 5px 10px;


habdelra (approved):

packages/base/workspace.gts
```diff
@@ -1311,6 +1311,12 @@ class Isolated extends Component<typeof Workspace> {
         font: 500 13px var(--grid-sans);
         color: var(--grid-nav-ink);
         cursor: pointer;
+        /* Without this the row keeps its max-content width instead of the
+           rail's, so a long type name pushes the count and the + past the
+           rail's right edge rather than ellipsizing. The ellipsis on
+           .rail-name only engages once the row itself is allowed to be
+           narrower than its content. */
+        min-width: 0;
```
habdelra, line 1319: [Claude Code 🤖]

**Confirmation — the fix is correct and the broad placement is the right call. Non-blocking; nothing to change.** Recording the mechanism as a guard rail for whoever edits this rail next.

Traced against the full rail markup and CSS on this branch. The key fact the PR body leans on but doesn't spell out is that `.rail-group` is itself `display: grid` (a few rules up), so `.rail-row` is a grid item in *every* group, not only the wrapped one:

- **Library / File-type groups** render `.rail-row` directly as a grid item of the definite-width `.rail-group` track. `.rail-name`'s `overflow: hidden` already zeroes that row's min-content contribution, so the `auto` track collapses and then maximizes to the rail width, and the name ellipsizes — with or without this property. `min-width: 0` is a genuine no-op here, which is exactly the "193px before and after" the PR measured.
- **Card-types group** is the only one carrying the extra `.rail-row-wrap` grid level. Nested one grid deeper, `.rail-row`'s default `min-width: auto` kept the inner track — and therefore the wrapper's contribution to the outer track — sized to the row's content, which is what defeated the ellipsis. `min-width: 0` drops that automatic minimum so the collapse propagates through both levels.

So putting it on `.rail-row` rather than scoping to `.rail-row-wrap .rail-row` is the durable choice: redundant-but-harmless where the rows already collapse, and it pre-empts the identical bug if a future group ever gains a wrapper. The inline comment is accurate and evergreen.

---

habdelra (commented):
[Claude Code 🤖]

**Reviewed as a CSS-containment change** — the lens is whether `min-width: 0` fully fixes the clipping, whether it's safe on the rows that already worked, and whether there's a cleaner home for the property.

**Bottom line: correct, minimal, and safe — no blocking issues.** The mechanism holds when traced against the full rail markup, and the property is a verified no-op on the two unwrapped groups. The mechanism detail and the "why it's safe on the other groups" guard rail are in the inline thread on the changed lines.

**What lands right.** Relaxing the row's min-width (containment) instead of capping `.rail-name` with a fixed `max-width` is the durable choice: it adapts to rail width and to multi-digit counts with no magic number to re-tune when the icon size, count width, or rail width changes. The code comment is a model of what an evergreen CSS comment should be — it explains the *why*, not the history.

**Adjacent, out of scope.** Now that long names deliberately truncate with an ellipsis, there's no way to read the full name — the `.rail-name` spans carry no `title`, even though other truncated/meaningful elements in this same file do (`title={{@model.purpose}}`, `title={{item.absolute}}`). A `title={{option.displayName}}` on `.rail-name` would restore the full text on hover. Genuinely a follow-up, not this PR.

**Test coverage.** Agree with the PR's reasoning: the wrapped card-type rows only render once `cardTypeFilters` is populated from `_types`, which the host helpers don't serve, so a regression assertion here is test-infrastructure work out of proportion to a one-line CSS fix. Reasonable to defer to the broader Workspace coverage work.

---

burieberry (approved):

PR #5636, opened 2026-07-28
host: Remove noisy diff editor Percy snapshot

backspace:
This produces a lot of Percy noise:

s 2026-07-28 at 15 36 31@2x

While it would be nice to have visual diffs for this, it’s too unreliable.


lukemelia (approved):
My hero! 😻

PR #5637, opened 2026-07-28
Report write-time rebuilds and make the client-perf panels say why

habdelra:
A tab rebuilds its card graph from two places, and only one of them reported. The store's event-driven rebuild runs when an incremental index event names a loaded executable module — but only in a tab whose store is subscribed to that realm, which requires something to hold a card instance from it. A code-mode editing session with no instance loaded takes the other path entirely: the save itself re-establishes the graph at write time (`refreshReferencesForCodeChange`), and the index event that follows reaches only the file resource, never the store. That write-time pass emitted no telemetry, so the session doing all of its rebuilding there put nothing on the dashboard — an empty rebuild row next to a session that is visibly absorbing code changes.

Observed on a deployed environment: a session on a build with rebuild reporting in place edited an executable module repeatedly (server-side incremental indexing confirms each write), and the board showed zero rebuilds — the session's store held no instance from that realm, so every rebuild it paid was the unreported write-time kind.

## Every rebuild reports, tagged by what drove it

`rebuild` events carry a `rebuild_source`: `realm-event` for the store's event-driven rebuild, `write` for the save-time re-establishment. A save of a loaded module in a subscribed realm pays both — one event per source, describing two complete graph re-fetches. That double pass is a known cost rather than an intent: the index event cannot tell that the write-time pass already ran, and collapsing them is open work this telemetry makes legible. (The field is not named `source` because the firelens log wrapper carries its own `source` key, which wins the wrapper-unwrap parse and silently shadows the event's — parsed-label conflicts keep the first extraction, without the `_extracted` suffix stream-label conflicts get.) and makes the documented double cost visible rather than half-hidden. Write-sourced rebuilds carry the saved module as their trigger, the realm written to, and `coalesced_events: 0`, since no index event drove them. The rebuild-rate panel splits by source.

## Realm events carry more of what happened

`event_args` rides on every realm-event: the raw event payload minus its invalidation list (tracked separately, bounded, as `invalidated_ids`). Not surfaced on any panel; it is there so a raw-line read has the full payload the tab acted on, and any field the realm adds to its events later arrives in the tracked data without a client change.

A per-file churn ranking deliberately does not ship here: the realm sorts the broadcast invalidation list (realm config first, then executables, then instances, lexically), so no element of it identifies the file whose write caused the event, and a ranking keyed to any of them would attribute churn to whichever dependency sorts first. Ranking written files needs the realm event to carry the written URLs, which is realm-server surface with its own tests — split out rather than bolted on.

## Panels say why a signal is notable

- **Wedges by user** names the blocking frame and its worst blocked time — the frame is what makes a row actionable, and a blank frame is a longtask-corroborated freeze whose breadcrumb lives on the raw line.
- **Worst card-to-interactive by user** names the card behind the number, so the row says what to open — and the same card recurring across users points at the card, not the network.
- **Signal events by user** becomes **Signal mix by user**: one column per event type. Volume alone is activity, not a problem; weight in the wedge/rebuild columns is the signal.
- The Overview stats all read as rolling 5-minute windows with sparklines — card loads, wedges, and rebuilds show the current count and its movement over the range, rather than a rollup that only grows with the time picker.

## Who is on which build

The session-history panel carries the build story, since a stale tab is the first thing to rule out when a client-side change "does not work" for one person. Grey `build in use: <sha>` legend rows name each host build the active sessions run (every beacon, keepalives included, stamps `app_version`): one row is the quiet steady state, a second row right after a deploy is a tab still running the previous code. Hovering the chart names which user is on which build:

Hovering the session-history chart in a divergent window: the tooltip lists user → build for each active account, showing one account running two builds at once

In the frame above, hovering the divergent window shows one account running two builds at once — two tabs, one of them stale — while the other account is on the current build only. The legend's Last column is strict-last, so a band that has aged out reads blank rather than holding its final value, and a user currently split across builds shows as two same-named live bands.

Two identification rules run through the panels. A card **type** is an address, not a class name: `deserialize` events carry `card_type` as `<moduleURL>/<ExportName>` (a relative module specifier resolves against the instance's own id), so a row on the deserialize panels says where the type's code lives rather than leaving a bare name to grep for. And a stat that is not a cross-user aggregate names its user: the slowest-card-load and slowest-endpoint rows carry the account the number came from, and the wedge blocked-time series splits per user so the hover names whose thread froze — a max is one person's worst moment, and knowing whose is what makes it actionable. Aggregates (percentiles, rates, sums) stay aggregates.

The dashboard is titled **Host App Client Performance**: the host app is the instrumented client, and other clients of the platform (boxel-cli among them) do not beacon, so a person absent from these panels is not necessarily inactive.

## Tests

The write-path test drives `refreshReferencesForCodeChange` — the entry point a code-mode save actually takes — and asserts a `write`-sourced rebuild comes out; this is deliberately a workflow-altitude test, since the store-handler-altitude tests all presuppose the realm subscription that a code-mode session lacks. Further tests pin `rebuild_source` on the event-driven rebuild, the `event_args` shape, that a write-sourced rebuild reports no module re-fetches, and that a write-time rebuild with a rendered card counts the graph it re-established.

| check | |
| --- | --- |
| `Integration \| Store` | green |
| host typecheck | matches main |
| observability lint | green |

All new and changed panels verified rendering against a local Grafana with synthetic data spanning both rebuild sources, multiple builds, wedge frames, and invalidation fan-out.


packages/host/app/services/store.ts
```diff
@@ -1946,12 +1987,14 @@ export default class StoreService extends Service implements StoreInterface {
         index_type: 'incremental',
         invalidations_count: invalidations.length,
         invalidated_ids: invalidations.slice(0, 50),
+        first_invalidated_id: invalidations[0] ?? '',
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed and removed in 86f7ec4 — `sortInvalidations` orders the broadcast list realm-config → executables → instances, lexically within each group, so element zero identifies whichever dependency sorts first, not the written file. The grouping key and the panel ranked on it are gone rather than relabeled: a churn ranking that attributes writes to arbitrary dependencies misleads more than it informs. Carrying the written URLs on the realm event is the honest basis for that panel; it spans several `broadcastIncrementalInvalidationEvent` call sites with distinct semantics (writes, deferred flushes, failure re-broadcasts), so it is split out as its own change. The `event_args` field spreads the raw event, so once the event carries the written URLs they arrive in the tracked data with no client change.

packages/host/app/services/store.ts
```diff
@@ -410,11 +410,41 @@ export default class StoreService extends Service implements StoreInterface {
...
+            trigger_module: triggerModules[0] ?? '',
+            modules_refetched: triggerModules.length,
+            cards_reloaded: cardsReloaded ?? 0,
+            coalesced_events: 0,
+          });
+        })
+        .catch(() => {
+          // The task can be cancelled by a reset racing it; a cancelled
+          // re-establishment is not a rebuild worth reporting.
+        });
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in 86f7ec4. Only a cancellation (a reset racing the re-fetch) goes unreported now, gated on `didCancel`; a genuine rejection reports the rebuild with its elapsed cost, matching the event-driven path's deliberate report-in-finally.

packages/host/app/services/store.ts
```diff
@@ -410,11 +410,41 @@ export default class StoreService extends Service implements StoreInterface {
...
+      refetch
+        .then((cardsReloaded) => {
+          telemetry.recordEvent({
+            event_type: 'rebuild',
+            source: 'write',
+            realm: opts?.realm ?? null,
+            duration_ms: Math.round(performance.now() - start),
+            trigger_modules: triggerModules,
+            trigger_module: triggerModules[0] ?? '',
+            modules_refetched: triggerModules.length,
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in 86f7ec4. Write-sourced rebuilds report `modules_refetched: 0` — the re-establishment re-fetches instances, not modules, so the count is meaningful on the realm-event source only, and the field's type comment now says so. A test pins the zero.

richardhjtan (commented):
[Claude Code 🤖] Reviewed with one question in front: **does every rebuild now report, and does the reported number mean what the panel says it means?** That is the lens because this change is measurement code — a wrong panel is worse than a missing one, since it gets trusted. So each new field was traced to its emitter, each changed query read against the emitter's actual output, and the deferred-work claims traced to the realm code they defer to.

**Bottom line: the write-path reporting is the right fix in the right place, and the reasoning about why a code-mode session was invisible checks out against the code. One regression blocks: attaching `.then` to the task instance silently switches off ember-concurrency's uncaught-error reporting for the write path, and only when telemetry is enabled. Everything else is non-blocking — one panel that can now omit users it used to always show, a doc/comment mechanism that is wrong in a way that will mislead the next editor, and a deferred-scope claim that is more tractable than the description says.**

## What lands right

The diagnosis is verified, not asserted. `handleInvalidations` only ever reaches the store for a realm the store subscribed to, and `#reloadInvalidatedInstances` / `rebuildForCodeChange` are the only things downstream of it — so a code-mode session holding no instance from the edited realm genuinely never enters that path, and `file.ts`'s `moduleWasLoaded` branch is genuinely the only rebuild it pays. Reporting from `refreshReferencesForCodeChange` rather than from the file resource is also the right seam: the store is what knows the cost, and the one production caller passes what the store cannot know (which module, which realm).

The three concessions to honesty are the best part of the change. `modules_refetched: 0` on the write source rather than a plausible-looking 1 — correct, since `reestablishReferences` walks `referenceCount` and re-fetches instances, never modules. `coalesced_events: 0` rather than 1. And dropping the invalidation-keyed churn ranking outright instead of relabeling it: `sortInvalidations` really does order the broadcast list config → executables → instances, so element zero is whichever dependency sorts first, and a panel built on it would have been confidently wrong. Deleting a panel is a harder call than shipping it; it was the right one.

Privacy checked and clean: the screenshot embedded in the description renders synthetic accounts (`@probe_a:localhost`, `@probe_b:localhost`) against placeholder shas — verified by fetching the blob, not by trusting the caption. No user realm paths in the diff, the fixtures, or the panel definitions.

## The deferred churn ranking is smaller than the description says

The description defers per-written-file ranking on the grounds that "ranking written files needs the realm event to carry the written URLs, which is realm-server surface with its own tests". The realm already carries them. `sendIndexInitiationEvent` in `packages/runtime-common/realm.ts` broadcasts `{eventName: 'index', indexType: 'incremental-index-initiation', updatedFile, realmURL}` — one per written URL, from four call sites, on the same channel, immediately before the invalidation broadcast that follows indexing. `updatedFile` is exactly the written URL the ranking needs.

The client drops it: `handleInvalidations` returns at the `indexType !== 'incremental'` guard, so the initiation event never reaches telemetry. (`realm.ts` in the host consumes it for `isIndexing`, so the subscription already delivers it.) That makes the follow-up a client-side change — record a thin event for the initiation, or stash `updatedFile` and attach it to the invalidation event that follows — with no realm-server surface and no new server tests. Worth correcting in the description so whoever picks it up scopes it as an afternoon rather than a cross-package change. Class: follow-up, non-blocking, and this PR is right not to do it here.

## Recommendations

1. **Rethrow genuine failures on the write path** — the `.then` consumption disables EC's uncaught-error reporting, and only under telemetry. Blocking. See the thread on `refetch.then` in `store.ts`.
2. **Pin the two branches the reporting added** — cancel-vs-failure and a non-zero `cards_reloaded` are both unexercised. See the thread on the new test in `store-test.gts`.
3. **Fix the `_extracted` mechanism in the comment and the skill** — the naming decision is right, the stated reason is not. See the thread on `rebuild_source` in `client-telemetry.ts`.
4. **Raise the limit on Wedges by user, or add a count column** — grouping by frame inside `topk(25)` lets a few users evict a whole account from a panel whose job is per-user coverage. See the thread on that panel's `expr`.
5. **Say in the panel description why refC duplicates refA** — otherwise the next editor deletes the duplicate as dead weight and the build hover disappears with it. See the thread on refC.
6. **Consider whether the two-full-rebuilds-per-save cost should be normalized in the docs** — see the thread on the `file.ts` call site. Not this PR's bug; this PR is what makes it legible.

## Adjacent, out of scope

- **`trigger_module` can be spelled two ways for one module.** The write source labels with `this._url`, the realm-event source labels with the invalidation URL. `file.ts` documents these as capable of diverging — an extensionless request canonicalizing to `.gts` is the example in its own comment — in which case one module change produces two rows on *rebuild cost by trigger_modules*. Low probability for a code-mode save (the path usually already carries the extension) and `_url` is still the better of the two available spellings, since `innerState.url` is an RRI that may be a scoped identifier rather than a URL. Flagging only so it is not a surprise later.
- **`event_args`'s virtue is also its exposure.** Bounded today — an incremental payload minus invalidations is four or five scalars. But "any field the realm adds to its events later arrives in the tracked data without a client change" means a future large field ships to Loki with no review, and the flush's size guard (`while (chunk.length > 1 && ...)` in `client-telemetry.ts`) halves a chunk but sends a single oversized event as-is, having already spliced it out of the buffer. A byte cap on `event_args` at record time would keep the property without the tail risk.
- **Signal mix by user lost its `topk`.** It now returns a row per account seen anywhere in the range. Fine at a handful of users; on a 30-day picker it is every account that ever opened the app.
- **Panel ids 3 and 91 are retired and 4 now names a different panel.** Any bookmark or annotation carrying `?viewPanel=3` or `viewPanel=4` on this board resolves to something other than what it was saved against. Nothing in the repo links by panel id — checked — so this only touches whatever people have in their own tabs.

packages/host/app/services/store.ts
```diff
@@ -410,11 +410,46 @@ export default class StoreService extends Service implements StoreInterface {
...
+        (cardsReloaded) => report(cardsReloaded ?? 0),
+        (e) => {
+          // A cancelled re-establishment (the owner tearing down mid-flight)
+          // is not a rebuild worth reporting — but a genuine failure is a
+          // rebuild the tab paid for, and the one most worth seeing.
+          if (!didCancel(e)) {
+            report(0);
+          }
+        },
+      );
```
richardhjtan, line 457: [Claude Code 🤖] **Consuming the task instance's promise switches off ember-concurrency's uncaught-error reporting — so a genuine failure here now produces a dashboard row and nothing else.** And it does so only when telemetry is enabled, which forks error visibility on a feature flag.

**The mechanism.** In ember-concurrency 5.2.0, `TaskInstance#then` is not a passive observer:

```js
// dist/-private/external/task-instance/base.js
then(...args) { return this.executor.promise().then(...args); }

// dist/-private/external/task-instance/executor.js
promise() {
  if (!this.defer) {
    this.defer = this.env.defer();
    this.asyncErrorsHandled = true;   // <-- side effect
    this.maybeResolveDefer();
  }
  return this.defer.promise;
}

maybeThrowUnhandledTaskErrorLater() {
  if (!this.asyncErrorsHandled && this.state.completionState === COMPLETION_ERROR && !didCancel(this.state.error)) {
    this.env.async(() => {
      if (!this.asyncErrorsHandled) { this.env.reportUncaughtRejection(this.state.error); }
    });
  }
}
```

`runFinalizeCallbacks` calls `maybeThrowUnhandledTaskErrorLater` on every finish. Touching `.then` sets `asyncErrorsHandled` first, so that guard is already false by the time the task finishes and the error is never routed to `reportUncaughtRejection`. The host sets no `Ember.onerror` (grepped: no assignment anywhere in `packages/host/app`), so `reportUncaughtRejection` was doing `throw error` inside a `next` — a real uncaught exception, visible in the console and fatal to a QUnit test.

**What changes, concretely.** A card re-fetch inside `reestablishReferences` throws — `getCardInstance` rejecting on a 500, `startAutoSaving` failing:

| | before | after |
| --- | --- | --- |
| telemetry off | uncaught error surfaces | uncaught error surfaces |
| telemetry on | uncaught error surfaces | one `rebuild` row, `cards_reloaded: 0`, silence |

The realm-event path is untouched — `rebuildForCodeChange` awaits `reestablishReferences` inside its own task body, so `onYielded` propagates the error to the parent instance, which nobody consumes, and the parent still reports. So the two sources now diverge on error visibility too, which is the opposite of what this PR is otherwise doing.

**The way out.** The `didCancel` gate is right and the report-on-failure instinct is right; the missing half is putting the error back:

```suggestion
      refetch.then(
        (cardsReloaded) => report(cardsReloaded ?? 0),
        (e) => {
          // A cancelled re-establishment (the owner tearing down mid-flight)
          // is not a rebuild worth reporting — but a genuine failure is a
          // rebuild the tab paid for, and the one most worth seeing. Reporting
          // it must not also swallow it: reading this task instance's promise
          // marks its errors handled, so ember-concurrency stops reporting
          // them, and rethrowing restores the surface.
          if (didCancel(e)) {
            return;
          }
          report(0);
          throw e;
        },
      );
```

The rethrow rejects the derived promise with no handler, so the failure lands as an unhandled rejection — still loud in the console, still fatal in tests. If you would rather it keep the exact old shape (a thrown exception rather than a rejection), hand it to the same place EC did instead of rethrowing.

**Scope: regression introduced by this PR, blocking.** It arrived with the fix for the failed-rebuild-reporting thread above, which was the right change to make — this is the part of that change that needs a second pass.

packages/host/tests/integration/store-test.gts
```diff
@@ -3564,6 +3569,99 @@ module('Integration | Store', function (hooks) {
     );
   });
 
+  test('a code-mode save reports the rebuild it performs at write time', async function (assert) {
```
richardhjtan, line 3610: [Claude Code 🤖] **This test pins the happy path's shape but not the two behaviours the reporting actually had to get right, and not the wiring that supplies its most important field.** Nothing here would fail if any of the three broke.

**What the test exercises.** `loader.import(personModule)` loads a module, but nothing holds a card, so `referenceCount` is empty and `reestablishReferences` (`store.ts`) builds an empty `remoteIds`, awaits `Promise.all([])`, and returns `0` synchronously-ish. That is the intended shape of the motivating session, and asserting `rebuild_source`, `trigger_module`, `realm`, `coalesced_events: 0`, `modules_refetched: 0` against it is right.

**Three things it leaves open.**

1. **The cancel-vs-failure branch is never entered.** `didCancel(e)` versus `report(0)` is the subtlest logic the change added, and the resolve path is the only one a test walks. A rejecting re-fetch would pin the report-on-failure behaviour and, as it happens, would also have caught the error-swallowing problem in the thread on `refetch.then` — that test fails on the missing uncaught error.
2. **`cards_reloaded` is only ever plumbed as zero.** `(cardsReloaded) => report(cardsReloaded ?? 0)` is the one place a real number crosses from the task's return value into the event, and with no live references the assertion `modules_refetched: 0` is the only count checked — `cards_reloaded` is not asserted at all. `renderCard(hassan)` before the call would make `remoteIds.size` non-zero and pin it; the sibling test above already establishes that setup.
3. **The `file.ts` wiring is invisible.** The test calls `refreshReferencesForCodeChange` directly and supplies `opts` itself. `opts` is optional and `file.ts` is the only production caller, so if that call site ever drops `triggerModule` or `realm`, nothing fails: the code compiles, this test passes, and the dashboard quietly starts collecting `trigger_module: ''` / `realm: null` rows — invisible precisely on *rebuild cost by trigger_modules*, which groups by that field. An assertion at the file-resource altitude (write a `.gts` through a `Ready` file resource with `flushLoader: true`, assert the emitted trigger equals the written URL) closes it.

The workflow-altitude choice for this test is right and the reasoning in the description holds — the store-handler-altitude tests really do presuppose the realm subscription a code-mode session lacks. This is about the negative space around it, not the altitude.

**Scope: test coverage on new code, non-blocking on its own — but (1) is what would have caught the blocking finding, so it is worth doing in the same pass.**

packages/host/app/services/client-telemetry.ts
```diff
@@ -98,17 +98,28 @@ export interface WedgeEvent extends BaseEvent {
...
 export interface RebuildEvent extends BaseEvent {
   event_type: 'rebuild';
+  // What drove the rebuild: an incoming realm index event, or this tab's own
+  // write re-establishing the graph at save time. A save of a loaded module in
+  // a realm the store is subscribed to pays both — one rebuild per source.
+  // Named rebuild_source, not source: the firelens log router wraps each
+  // deployed log line in an envelope that carries its own `source` key, and
+  // the dashboard's wrapper-unwrap parse keeps the first occurrence of a
+  // duplicated label — the wrapper's — silently dropping the event's.
+  rebuild_source: 'realm-event' | 'write';
```
richardhjtan, line 112: [Claude Code 🤖] **The naming decision is right; the mechanism given for it is not, and the wrong version is the one that will send the next editor looking for something that isn't there.**

Loki does not drop a duplicated extracted label. When `| json` extracts a key that already exists in the pipeline, it keeps both and renames the *new* one with an `_extracted` suffix. So under the dashboard's `| json | line_format … | json` unwrap, a field named `source` on the event would not vanish — it would be queryable as `source_extracted`, while bare `source` resolved to the firelens wrapper's value.

That difference matters for whoever reads this comment next. As written, it says the value is gone, so a future editor who wants `source` back concludes it is impossible. What is actually true is sharper and makes the same case: a bare `source` filter would silently match the wrapper's value instead of the event's, and the event's would only be reachable under a name nobody would think to write. Same conclusion — don't name it `source` — for a reason that survives contact with the data.

Suggested rewording, and the same fix applies to the sentence in `.claude/skills/client-perf-diagnosis/SKILL.md` ("keeps the first occurrence of a duplicated label — the wrapper's — silently dropping the event's"), which is the copy a diagnosing reader will actually hit:

```suggestion
  // Named rebuild_source, not source: the firelens log router wraps each
  // deployed log line in an envelope carrying its own `source` key, and the
  // dashboard's wrapper-unwrap parse extracts the wrapper before the event.
  // Loki resolves the collision by suffixing the later extraction, so a
  // `source` filter would match the wrapper's value and the event's would be
  // reachable only as `source_extracted`.
  rebuild_source: 'realm-event' | 'write';
```

While in here: the `event_args` comment a few lines down says the field is "Not surfaced on the dashboard". True of the panels, but `| json` flattens nested objects, so the payload does arrive as queryable `event_args_*` labels — a raw-line read is not the only way to get at it. Understating that is harmless, but stating it would make the field more useful to the person reading the glossary.

**Scope: documentation correctness on new comments, non-blocking.**

packages/host/app/resources/file.ts
```diff
@@ -460,7 +460,10 @@ class _FileResource extends Resource<Args> {
       // belongs here: a write to a module nothing imported has no rendered
       // consumers to refresh.
       if (moduleWasLoaded) {
-        this.store.refreshReferencesForCodeChange('file write');
+        this.store.refreshReferencesForCodeChange('file write', {
+          triggerModule: this._url,
+          realm: state.realmURL,
+        });
```
richardhjtan, line 466: [Claude Code 🤖] **Confirmation, plus one thing worth not normalizing.** This is the right call site and the right two fields: `triggerModule` and `realm` are exactly what the store cannot derive for itself, and passing them here keeps the store from having to guess.

**What the new telemetry makes visible.** Traced end to end, a code-mode save of a loaded module in a realm the store *is* subscribed to pays two complete rebuilds, not two events describing one:

1. `write` — this branch: `store.reset()` plus `reestablishReferences.perform()`, re-fetching every live reference.
2. `realm-event` — the index event lands later; `wasModuleFlushedForCodeChange` reports the flush `saveSource({resetLoader: true})` performed, so `alreadyFlushed.size > 0` makes `needsRebuild` true, and `rebuildForCodeChange` does `resetLoader()` + `store.reset()` + `reestablishReferences.perform()` again.

`reestablishReferences` is a plain `task`, not `restartable` or `keepLatest`, so the second pass neither cancels nor coalesces with the first — the graph is re-fetched twice per save, and every open card is torn down and rebuilt twice.

**Why that is worth a word.** The cost predates this PR and the instrumentation is right to expose it. But the description and the skill now present it as settled shape — "pays both — one rebuild per source is the expected shape, not a double-count", "A save of a loaded module in a subscribed realm pays both". Both sentences are accurate about the *events*, and both read as endorsing the *work*. The next person to see two rebuild rows per save on the dashboard will read the skill, conclude it is by design, and move on — which is exactly the outcome this PR exists to prevent for the write path.

Nothing to change in the code here. What would help is one clause in the skill's rebuild section marking the double pass as a known cost rather than an intent — something to the effect that a subscribed code-mode session pays the graph re-fetch twice, that the second pass exists because the index event cannot tell that the write-time pass already ran, and that collapsing them is open work. Then the panel reads as evidence for a fix rather than as documentation of a feature.

**Scope: pre-existing, now load-bearing — the PR does not introduce the double rebuild but is what makes it legible, and the wording choice decides whether it gets fixed. Collapsing the two passes is a follow-up, not this PR. Non-blocking.**

packages/observability/grafanactl/resources/dashboards/boxel-status/client-performance.json
```diff
@@ -3018,18 +2936,19 @@
                   "uid": "loki"
                 },
                 "editorMode": "code",
-                "expr": "topk(25, sum by (matrix_user_id) (count_over_time({service=\"realm-server\", env=\"$env\"} |= \"boxel:client-perf\" | json | line_format \"{{ if .log }}{{ .log }}{{ else }}{{ __line__ }}{{ end }}\" | json | channel=\"boxel:client-perf\" | matrix_user_id=~\"$matrix_user_id\" | session_id=~\".*${session_id}.*\" | event_type=\"wedge\" [$__range])))",
+                "expr": "topk(25, max by (matrix_user_id, top_frame_function, top_frame_url) (max_over_time({service=\"realm-server\", env=\"$env\"} |= \"boxel:client-perf\" | json | line_format \"{{ if .log }}{{ .log }}{{ else }}{{ __line__ }}{{ end }}\" | json | channel=\"boxel:client-perf\" | matrix_user_id=~\"$matrix_user_id\" | session_id=~\".*${session_id}.*\" | event_type=\"wedge\" | unwrap blocked_ms [$__range])))",
```
richardhjtan: [Claude Code 🤖] **Adding the frame dimension inside the `topk` changes what the limit selects, and this panel can now omit a wedging user entirely — in the one row of the board whose job is telling you which users are suffering.**

**Before and after.** The old expression ranked accounts: `topk(25, sum by (matrix_user_id) (count_over_time(… event_type="wedge" …)))` — at most one row per user, so 25 covered every user this deployment will plausibly have. The new one ranks `(matrix_user_id, top_frame_function, top_frame_url)` triples by `max blocked_ms`. Row count is now users × distinct blocking frames, and `topk` keeps the 25 highest *values*, not the 25 highest users. A handful of accounts each hitting several different blocking frames across a day exceeds 25 triples, and what falls off the bottom is whole (user, frame) pairs ordered by blocked time — which can be every row a given user has.

**Which rows sort last is the unlucky part.** Read the emitter: `blocked_ms: Math.round(blockedMs)` where `blockedMs` accumulates LoAF `blocking_duration` and falls back to longtask durations only when no LoAF entry is in the window. Neither present → `blocked_ms: 0`. Those are exactly the freezes this panel's own description calls out as real ("A blank frame is a freeze corroborated only by a longtask, with no LoAF script attribution — real, but read its raw line"). They now rank at the very bottom of the `topk` and are the first thing evicted. A user whose only wedges are uncorroborated gaps can disappear from the panel while wedging.

**There is also now overlap with an existing panel.** *Top wedging frames (by blocking script)* in the wedge row already does `topk(15, sum by (top_frame_function, top_frame_url) (… | top_frame_function != "" …))` — the frame-centric view, with blanks deliberately filtered out. The frame dimension was covered; the per-user completeness was this panel's contribution, and that is what the change trades away.

**The way out**, in preference order:

1. Keep the per-user row and carry the frame as detail, ranking by user so the limit means what it used to. If the frame must be a grouping label, then
2. raise the limit well past the plausible triple count (`topk(200, …)`) — the table paginates anyway, and no query cost changes since the grouping happens after the range scan; and
3. add a second target `sum by (matrix_user_id) (count_over_time(… event_type="wedge" …))` for a wedge-count column, merged the way *rebuild cost by trigger_modules* merges its three targets. That restores "how often" alongside "how bad", and gives the panel a value that is non-zero for every wedging user.

The intent behind the change is right — a row naming the blocking frame is actionable and a bare count is not. It is the interaction with `topk` that bites.

**Scope: regression in this PR, non-blocking — it degrades a diagnostic panel rather than the data behind it, and the fix is a one-line expression change.**

packages/observability/grafanactl/resources/dashboards/boxel-status/client-performance.json
```diff
@@ -743,11 +631,35 @@
...
+            "legendFormat": "build in use: {{app_version}}",
+            "refId": "B"
+          },
+          {
+            "datasource": {
+              "type": "loki",
+              "uid": "loki"
+            },
+            "editorMode": "code",
+            "expr": "count by (matrix_user_id, app_version) (count by (matrix_user_id, app_version, session_id) (count_over_time({service=\"realm-server\", env=\"$env\"} |= \"boxel:client-perf\" | json | line_format \"{{ if .log }}{{ .log }}{{ else }}{{ __line__ }}{{ end }}\" | json | channel=\"boxel:client-perf\" | matrix_user_id=~\"$matrix_user_id\" | session_id=~\".*${session_id}.*\" [5m])))",
```
richardhjtan, line 658: [Claude Code 🤖] **refC is byte-identical to refA and exists only so the tooltip can carry a different display name — which works, and is invisible to the next person to open this JSON.**

Verified against the screenshot in the description: refA is `hideFrom.tooltip`, refC is `hideFrom.viz` + `hideFrom.legend`, so the legend shows `{{matrix_user_id}}` per band while the hover shows `user → build`. Grafana has no per-series legend/tooltip name split, so running the query twice is genuinely the mechanism, not an oversight.

Two consequences worth writing down.

**Cost.** This panel's query has no `event_type` filter, so it scans every client-perf line including keepalives, over the full picker range, with a 5m rolling window — the broadest query on the board. It went from one such scan to three (refA, refB, refC), with refC's contribution being purely presentational. Grafana does not deduplicate identical targets. On a wide picker the skill's own warning about unfiltered parses applies here more than anywhere else on the board.

refB at least computes something refA does not surface (sessions per build, independent of user). refC computes nothing new: it is refA's series with a different name. If the scan cost matters more than the hover, the hover is the cheaper thing to give up.

**Durability.** Nothing in the JSON records why the duplicate exists. Two identical `expr` strings in one panel read as a copy-paste mistake, and the obvious tidy-up — delete one — silently removes the user-to-build hover that the description is built around. The panel's `description` field is the only durable place to say so, and it already explains the grey legend rows; one clause there ("the hover is a third query of the same expression, named differently — Grafana cannot vary legend and tooltip names independently") makes the duplication legible.

**Scope: efficiency and maintainability on new panel config, non-blocking. No correctness problem — the panel renders what the description claims.**

packages/host/app/services/store.ts
richardhjtan: [Claude Code 🤖] **Correction to my own comment: downgrading this from blocking to non-blocking. The mechanism above is right, but I did not verify that the rejection it suppresses is reachable, and it very nearly is not.**

`getCardInstance` wraps everything after its first few lines in a `try` whose `catch` converts *every* throw into a returned `CardErrorJSONAPI` and logs it:

```js
} catch (error: any) {
  let errorResponse = processCardError(id, error);
  let cardError = errorResponse.errors[0];
  deferred?.fulfill(cardError);
  ...
  storeLogger.error(message);   // or .debug for suppressed 404s
  return cardError;
}
```

`createFromSerialized`, the fetch, deserialization, `setIdentityContext` and `startAutoSaving` are all inside that `try`. So the three causes the original thread named — "a referenced card fetch, deserialization, or autosave setup fails" — do not reject; they come back as values, already logged, and `Promise.all` in `reestablishReferences` resolves normally. Only two things can actually reject: a throw from `asURL` or the `inflightGetCards` bookkeeping, which sit above the `try`, and `processCardError` itself throwing inside the `catch`. Both are narrow enough that I would not design around them.

That changes three things:

1. **This is not blocking.** The suppression is real but sits on a path that essentially does not fire, and where it did fire the error had already been logged by `storeLogger.error` inside `getCardInstance`. Keeping the rethrow is still worth the one line as belt-and-braces — it costs nothing and removes the telemetry-on/telemetry-off divergence — but nothing here needs to hold the PR.
2. **The premise of the thread this fix answered did not hold.** `report(0)` is close to dead code. Worth knowing before anyone invests further in that branch — my own recommendation to test it was over-weighted for the same reason.
3. **My recommendation 2 in the review body is partly wrong.** The `cards_reloaded` half stands — that value really is only ever plumbed as zero, and a test with a rendered card would pin it. The cancel-vs-failure half does not: there is little point pinning a branch that cannot be driven without mocking a throw that production cannot produce.

Apologies for the mislabel — I verified the `asyncErrorsHandled` mechanism in the ember-concurrency source and then asserted the consequence without tracing whether `reestablishReferences` can reject at all, which is exactly the check I flagged as owed for the bot findings on this PR.

richardhjtan (approved):

packages/host/app/services/store.ts
habdelra: [Claude Code 🤖] Fixed in d647f0c — the rejection handler reports and then rethrows, so a genuine failure surfaces as an unhandled rejection and error visibility no longer forks on whether telemetry is armed. The correction on this thread shaped the disposition: with `getCardInstance` converting nearly every failure into a returned error value, the rethrow is belt-and-braces on a branch that rarely fires — kept because it costs one line and removes the divergence, not because the branch is hot.

packages/host/tests/integration/store-test.gts
habdelra: [Claude Code 🤖] Partially addressed in d647f0c: a companion test renders a card first and pins `cards_reloaded >= 1` — the field whose value crosses from the re-establishment task into the event, previously only ever plumbed as zero in tests. The cancel-vs-failure branch stays unpinned, per the correction on the sibling thread: the rejection it guards is near-unreachable without mocking a throw production cannot produce. The `file.ts` opts-plumbing gap is real and acknowledged — the degenerate outcome is `trigger_module: ""` rows collecting silently on the panel that groups by that field — and a file-resource-altitude test (write a `.gts` through a Ready resource, assert the emitted trigger equals the written URL) is the follow-up shape for it.

packages/host/app/services/client-telemetry.ts
habdelra: [Claude Code 🤖] The naming conclusion stands, but the suggested mechanism doesn't match what this Loki version does — tested on 3.4.4 (the deployed version) with a firelens-shaped line (`{"log": <event with source:"write">, "source": "stdout", ...}`) through the dashboard's exact parse:

```
sum by (source)            -> [({source: "stdout"}, 1)]
sum by (source_extracted)  -> [({}, 1)]   # no such label exists
```

The `_extracted` suffix is appended when an extracted key conflicts with a *stream* label; a parsed-vs-parsed conflict — which is what the double `| json` produces — keeps the first extraction and drops the second, so the event's value is genuinely unreachable, not reachable under a surprising name. The comment now states that distinction explicitly (bbba294), so a doc-aware reader isn't sent looking for `source_extracted`. The `event_args` point is taken as-is: the comment now says its members are queryable as flattened `event_args_*` labels, not raw-line-only.

packages/host/app/resources/file.ts
habdelra: [Claude Code 🤖] Fixed in bbba294. The skill's rebuild section and the PR description now mark the double pass as a known cost rather than an intent: the index event cannot tell that the write-time pass already ran, the tab tears down and re-fetches every open card twice per save, and collapsing the two passes is open work. Two rebuild rows per save in a subscribed session now reads as evidence for a fix, not documentation of a feature — which was the point of the thread.

packages/observability/grafanactl/resources/dashboards/boxel-status/client-performance.json
habdelra: [Claude Code 🤖] Fixed in bbba294, options 2 and 3 combined: the limit sits at 200 (grouping happens after the range scan, so no query cost changes) and rows carry a wedge count from a second target with the identical grouping, merged the way the rebuild-cost panel merges its targets. Verified against synthetic data that the eviction case is gone: a user whose only freeze is longtask-corroborated (blank frame, blocked time 0) keeps a row — blank frame, 0 worst blocked ms, count 1 — instead of sorting off the bottom. The description now states why blank-frame rows can read 0 and why the count column exists.

packages/observability/grafanactl/resources/dashboards/boxel-status/client-performance.json
habdelra: [Claude Code 🤖] Fixed in bbba294 — the panel description now states that the hover attribution is a third query byte-identical to the bands query, differing only in display name, and that deleting the apparent duplicate removes the user → build hover. On cost: the duplication stays, since the hover is the panel's attribution mechanism and the scan-cost concern is bounded by the panel's 1m interval floor; if the board ever needs the scans back, the description now says exactly what giving up the hover buys.

PR #5638, opened 2026-07-29
Refresh workspace chooser live when realms change out of band

lukemelia:
## Problem

Fixes [CS-12271](https://linear.app/cardstack/issue/CS-12271). If a user is viewing the workspace chooser and creates a realm out of band (another tab, the CLI, an AI agent), the chooser didn't update until a reload.

A modern **trusted-realm-server session** assembles its workspace list once at boot from the realm server's `_realm-auth` endpoint. Out-of-band realm changes had no live signal reaching that session — account data is ignored on those sessions, and no realm-lifecycle event was ever emitted. (Legacy sessions already refresh via `app.boxel.realms` account-data sync.)

## Fix — a `realms-list-updated` push

A new bare-signal event, `REALMS_LIST_UPDATED_EVENT_TYPE = 'realms-list-updated'` (`runtime-common/matrix-constants.ts`).

**Server** emits it to the owner's session room after success in all four list-mutating handlers — `_create-realm`, `_delete-realm`, `_archive-realm`, `_unarchive-realm` — reusing the same session-room channel billing notifications already use. The session room is per-user, so the push fans out to every one of the owner's tabs. Emits are best-effort: a failed notify never turns a successful mutation into an error. `resolveAndAuthorizeArchiveTarget` now returns `ownerUserId` for the archive/unarchive handlers.

**Host** (`matrix-service.ts`) subscribes once via the existing `realmServer.subscribeEvent` channel. On the event, a trusted-server session re-derives the list from `_realm-auth` (`applyTrustedRealmServersAccountData` → `setAvailableRealmIdentifiers`), which drives the chooser's `userRealmIdentifiers`. Legacy sessions are deliberately left to their account-data path.

## Design decisions

- **Bare signal, not a payload** — the host re-fetches authoritatively from `_realm-auth` rather than trusting event data, so the list can't drift from what the server actually returns.
- **All four endpoints** emit the same generic "list changed" signal, keeping the chooser consistent cross-session for every change, not just creation.

## Tests

`packages/host/tests/integration/matrix-service-realms-list-updated-test.ts`:
- a trusted-server session pulls in an out-of-band realm on the event;
- a legacy session ignores the event (guard), staying on its account-data path.

## Verification status

Typecheck (`lint:types`) and eslint are clean on all changed files. The host/realm-server suites weren't run locally — they need the full matrix-synapse + multi-realm + postgres harness. The server emit reuses the identical `sendEvent` plumbing as the existing production billing-notification path. Relying on CI for suite execution.


jurgenwerk:
Looks like https://github.com/cardstack/boxel/pull/5615 is a duplicate of this bug fix, but I think this is a better and simpler solution.

packages/host/tests/integration/matrix-service-realms-list-updated-test.ts
```diff
@@ -0,0 +1,160 @@
...
+    },
+  } as any);
+}
+
+// CS-12271: a realm created out of band (another tab, the CLI, an AI agent)
+// has no signal that reaches a trusted-server session — its realm list is
+// assembled authoritatively from `_realm-auth` at boot. The realm server now
+// pushes a `realms-list-updated` event to the owner's session room, and the
+// host re-derives the list so a session viewing the workspace chooser updates
+// without a reload.
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Fixed in fc7f3ec — dropped the `CS-12271:` prefix so the comment stands on its own describing the behavior.

packages/host/app/services/matrix-service.ts
```diff
@@ -1349,6 +1356,54 @@ export default class MatrixService extends Service {
...
+  private async refreshRealmsList() {
+    if (!this.trustedRealmServersAuthoritative) {
+      return;
+    }
+    try {
+      let realmServers = await this.getRealmServersFromAccountData();
+      if (realmServers.length === 0) {
+        return;
+      }
+      await this.applyTrustedRealmServersAccountData(realmServers);
```
habdelra, line 1407: [Claude Code 🤖] **The refresh re-derives only the active list, so the archive/unarchive handler comments overstate what the chooser does.** (Non-blocking correctness gap.)

`refreshRealmsList` re-derives the available-realms list from `_realm-auth` via `applyTrustedRealmServersAccountData` → `fetchUserRealmsFromTrustedServers` → `setAvailableRealmIdentifiers`. That path only ever touches the *active* "Your Workspaces" list. The **Archived** section is a separate list (`RealmServerService.archivedRealmsList`), populated from a different endpoint (`GET /_archived-realms`) and refreshed only by `fetchArchivedRealms({ force: true })` — which this method never calls.

`_realm-auth` excludes archived realms: `handle-realm-auth.ts` calls `fetchUserPermissions(..., { onlyOwnRealms: false })`, and `realm-permission-queries.ts` renders `AND realm_url NOT IN (SELECT url FROM realm_metadata WHERE archived_at IS NOT NULL)` when `includeArchived` is false (the default). So on this generic `realms-list-updated` push:

- **Out-of-band archive:** the realm correctly leaves "Your Workspaces" (good), but it does *not* appear in "Archived". If the owner already expanded that section (`archivedRealmsFetched` is true, so a plain `fetchArchivedRealms()` no-ops), it won't show up there until a reload. The comment on `handle-archive-realm.ts` — "moves the realm into Archived without a reload" — isn't what happens.
- **Out-of-band unarchive:** the realm correctly returns to "Your Workspaces", but the stale `archivedRealmsList` still contains it. The chooser renders `archivedRealms` with no filter against the active list (`operator-mode/workspace-chooser/index.gts`), so a previously-expanded Archived section shows the realm in *both* places until a reload.

Create and delete don't touch the archived list, so they're fully correct — this is specific to archive/unarchive.

**Fix:** after re-deriving the active list here, also refresh the archived list when it's already loaded — e.g. `if (this.realmServer.archivedRealmsFetched) await this.realmServer.fetchArchivedRealms({ force: true });` (expose a public getter/method for the flag; keep it guarded so an owner who never opened Archived still doesn't pay the fetch). Since one generic signal serves all four mutations, this handler is the single place that has to keep both lists honest. Either make that change, or soften the two handler comments to describe the actual "active list only" behavior.

---

packages/realm-server/handlers/handle-archive-realm.ts
```diff
@@ -70,6 +72,16 @@ export default function handleArchiveRealm({
         },
       });
       await setContextResponse(ctxt, response);
+
+      // Tell the owner's other sessions their realm list changed so a session
+      // viewing the workspace chooser moves the realm into Archived without a
+      // reload. Best-effort: the realm is already archived, so a failed notify
+      // must not turn a successful archive into an error.
```
habdelra: [Claude Code 🤖] This says the chooser "moves the realm into Archived", but the host's `realms-list-updated` handler (`matrix-service.ts` `refreshRealmsList`) only re-derives the *active* list from `_realm-auth`; it never refreshes the Archived section. The realm leaves "Your Workspaces" (correct) but doesn't land in "Archived" without a reload. See the inline on `refreshRealmsList` for the mechanism and fix. Non-blocking — the emit itself is right; it's the comment's promise that outruns the host.

---

packages/realm-server/handlers/handle-unarchive-realm.ts
```diff
@@ -77,6 +79,16 @@ export default function handleUnarchiveRealm({
         },
       });
       await setContextResponse(ctxt, response);
+
+      // Tell the owner's other sessions their realm list changed so a session
+      // viewing the workspace chooser restores the realm to Your Workspaces
+      // without a reload. Best-effort: the realm is already restored, so a
+      // failed notify must not turn a successful unarchive into an error.
```
habdelra: [Claude Code 🤖] The "restores the realm to Your Workspaces" half is accurate, but the refresh leaves the *archived* list untouched. If the owner had expanded Archived before this fired, the realm now shows in both "Your Workspaces" and "Archived" until a reload (the chooser renders `archivedRealms` without filtering against the active list). See the inline on `refreshRealmsList`. Non-blocking.

---

packages/host/app/services/realm-server.ts
```diff
@@ -859,9 +859,25 @@ export default class RealmServerService extends Service {
...
+            await subscriber(realmServerEvent.data);
+          } catch (err) {
+            console.error(
+              `realm-server event subscriber for ${realmServerEvent.eventType} failed`,
+              err,
+            );
+          }
+        }),
+      );
+    }
```
habdelra, line 888: [Claude Code 🤖] **Confirmation — this await change is correct and actually necessary for the feature, not just tidier.** The new `refreshRealmsList` subscriber does real awaited work (a `_realm-auth` round-trip) before it settles, so the old `forEach(async …)` returned from `handleEvent` before the refresh completed — which is exactly what the integration tests (and any `await handleEvent(...)` caller) need to observe. Isolating each subscriber in its own try/catch preserves the best-effort dispatch the `forEach` had (one rejection can't take down its siblings), and it's a strict improvement — under the old code a throwing subscriber became an unhandled rejection.

Worth recording why this doesn't regress production latency: the only live caller, `processDecryptedEventFromAuthRoom`, is invoked fire-and-forget from `processDecryptedEvent` (no `await`), so awaiting subscribers here does not block Matrix timeline processing. Guard rail for the next editor — keep that call site fire-and-forget, or a slow subscriber would start stalling event processing.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I focused on whether the host's single refresh handler actually produces the workspace-chooser state each of the four emitting endpoints promises in its comment, and on the safety of the new event-dispatch plumbing.

**Bottom line — solid, well-documented fix; no blocking issues.** The create and delete paths are correct end to end. One non-blocking correctness gap: the host refresh re-derives only the *active* realm list, never the *archived* list, so the archive/unarchive handler comments overstate what the chooser does — an out-of-band unarchive can leave a realm shown in both "Your Workspaces" and "Archived" until a reload. Detail inline on `refreshRealmsList`.

**What lands right.**
- The bare-signal design (re-fetch authoritatively from `_realm-auth` rather than trusting a payload) is the right call — the active list can't drift from what the server returns, and it lets one generic event serve all four mutations.
- Best-effort emits placed after `setContextResponse`, each in its own try/catch — a failed notify never turns a successful mutation into an error. Verified in all four handlers, and `sendEvent` is threaded through `CreateRoutesArgs` to delete/archive/unarchive and explicitly to `handleCreateRealmRequest`.
- `subscribeToRealmsListUpdatesOnce` wired from the post-login path with an app-lifetime guard matches the `resetState()` "subscriber wiring is app-scoped" contract, and correctly avoids forcing `RealmServerService` to instantiate merely by constructing `MatrixService`.
- The `handleEvent` await change is correct and actually necessary for this feature (confirmation inline on `realm-server.ts`).

**Recommendations.**
1. Refresh the archived list in `refreshRealmsList` (or soften the archive/unarchive comments to match active-list-only behavior). Non-blocking — see inline on `refreshRealmsList`.
2. Extend the new integration test to cover archive/unarchive. The two tests today exercise only create (trusted server) and the legacy no-op, so the archived-section behavior the handler comments claim is untested — CI is green precisely because nothing asserts it.

**Adjacent, out of scope.**
- `applyTrustedRealmServersAccountData` takes its "merge, never remove" branch whenever *any* trusted server was unreachable that round, and the retry only adds. So an out-of-band delete/archive won't drop the realm from the active list if a *different* trusted server is transiently down at refresh time. Harmless in the common single-trusted-server case (the server that just processed the mutation is reachable), but worth knowing for the multi-server future.
- The event reaches only the realm *owner's* session room. A realm shared with another user that gets archived/deleted out of band won't refresh that other user's chooser live. Consistent with this PR's stated scope; noting the boundary.
- `send-event.ts` asserts `roomId!` immediately after logging "cannot find session room" — a null session room would throw rather than no-op. Pre-existing (billing already routes through this), caught best-effort here, so not this PR's bug; flagging because this PR adds four more callers of that path.

---

packages/host/app/services/matrix-service.ts
lukemelia: [Claude Code 🤖] Fixed in 6b025e75b8. `refreshRealmsList` now also refreshes the archived list after re-deriving the active one — guarded on `isArchivedRealmsFetched` (a new public getter on `RealmServerService`) so an owner who never opened Archived doesn't pay the fetch, and inside the existing try/catch so it stays best-effort. Added integration coverage for out-of-band archive, unarchive, and the never-opened guard (plus a `setRealmArchived` mock helper). Thanks for the precise mechanism writeup — it pointed straight at the fix.

packages/realm-server/handlers/handle-archive-realm.ts
lukemelia: [Claude Code 🤖] Fixed in 6b025e75b8 — the host now refreshes the archived list on this event (when it's already open), so the realm actually lands in Archived without a reload. Tightened this comment to say it drops from Your Workspaces and, if that session already has Archived open, moves it there.

packages/realm-server/handlers/handle-unarchive-realm.ts
lukemelia: [Claude Code 🤖] Fixed in 6b025e75b8 — the refresh now also drops the realm from an already-open Archived section, so it no longer shows in both places until a reload. Comment updated to match, and there's an integration test asserting the realm ends up in exactly one section.

habdelra (approved):

PR #5640, opened 2026-07-29
Report uncaught exceptions and unhandled rejections as a client-performance beacon event

habdelra:
The client-performance instrument reports what a browser tab *did* — card loads, server round-trips, deserialization, loader rebuilds, index events, main-thread freezes — but nothing about what *broke*. A tab that throws rather than freezes leaves no beacon at all: the wedge channel only catches main-thread stalls, so a session killed by an uncaught error goes quiet and reads exactly like a closed tab. When a user says "the app broke" and it didn't freeze, there is nothing to pivot to from their `matrix_user_id`.

The same gap swallows a signal the code deliberately raises. `refreshReferencesForCodeChange` rethrows genuine re-establishment failures specifically so they stay loud — and loud means loud in that one user's console, where no operator will ever see it.

This adds a `client-error` event to the instrument, an errors row to the board, and the matching mode to the `client-perf-diagnosis` skill. The realm-server ingest route needs no change: it validates the envelope and re-emits each event verbatim, so a new event type rides the existing pipeline.

## The event

`window` `error` and `unhandledrejection` listeners join the instrument's existing arm/teardown lifecycle alongside `pagehide` / `visibilitychange`, so gating and release come from machinery that is already there. Both handlers are passive — neither consumes the event, so the browser's own reporting is unaffected.

Building the report lives in `app/lib/client-error-report.ts` rather than in the service, split along what the two halves actually are. What a throw *is* — its message, its stack, where it happened, and the identity repeats coalesce under — is pure: the same event always yields the same report. How many events to emit and when is stateful, owned by the ring buffer, the dedup slots, and the window budget. The module exposes one entry point per channel returning one `ErrorReport`, so each handler in the service is a line, and the pure half is directly unit-testable without a window listener, an armed instrument, or a realm.

An `ErrorEvent` prefers its `error` object and falls back to `event.message`, so a cross-origin `Script error.` is still reported rather than dropped. A rejection's `reason` can be anything: an `Error`-like reason gives name, message, and stack, and anything else gives a bounded string conversion. Everything runs inside a try/catch and a re-entrancy guard, because a throw from inside an error handler would dispatch `error` again and recurse.

Fields are `kind`, `message`, `message_key`, `stack`, `top_frame_function`, `source_url`, `line`, `col`, `dedup_count`, `folded_signatures`, plus the usual envelope. `source_url` rather than `source`: the FireLens envelope wrapping every deployed log line carries `log`, `source`, `container_id`, `container_name`, and `ecs_*`, and the board's wrapper-unwrap parse keeps the wrapper's value for a duplicated key and silently drops the event's.

### The stack

The stack is the payload the event exists to carry — the message says the app broke, the stack says where — so it gets the budget and the message is what yields. Sixteen frames and 2000 characters: eight frames does not clear the framework frames between a throw and the app code that caused it. Truncation runs from the **deep end**, so the throw site always survives and it is the least informative tail that goes.

Everything above the first frame is the message the engine opened the stack with, and a message is unbounded — an error carrying a response body or a serialized document runs to thousands of characters, across as many lines as it likes. That header is bounded as a unit against a budget of its own (500 characters) before deep frames are dropped, because it competes with the frames otherwise: it takes the whole allowance, dropping deep frames cannot shrink it, and every frame is discarded — losing exactly what the stack is here for.

Where the header ends is decided by *parsing*: the first line carrying an engine marker **and** a trailing `:line:col`. Both tests are needed and neither is sufficient. "Looks frame-shaped" alone accepts a message that merely mentions a scoped module (`@cardstack/…`) or an account (`@user:localhost`), because the marker test looks for a bare `@` — and a message essentially never carries a location, so requiring one separates the two reliably. `MAX_ERROR_STACK_FRAMES` counts frames, not lines, and truncation counts its own `…` marker, so a budget means what it says and a cut value always shows that it was cut.

`top_frame_function` lifts the throwing function out as a scalar so a table can group by where errors come from, mirroring the wedge event's field of the same name. `source_url` / `line` / `col` come from the error event when it provides a location and otherwise from the stack's first frame, so a rejection — which carries no location of its own — reports one too. All four are read from the stack **as the engine gave it** rather than from the bounded copy: where a throw happened is a fact about the error and must not depend on how much of the stack fit in a budget.

Frames are split structurally rather than by matching a tail pattern, and *which* delimiter applies is decided by the engine that wrote the frame rather than by which character appears first. V8 writes `at fn (location)` or a bare `at location`; SpiderMonkey and JSC write `fn@location`, leading with the `@` when the frame is anonymous. Keying off the `at` prefix matters because a `@` inside a V8 location is part of the URL — `@cardstack/…` and `@ember/…` are ordinary path segments in this app's bundle URLs — so treating it as a delimiter would cut the URL in half. Splitting on the *first* `' ('` likewise keeps the function name intact when the location itself carries parens (a bundler URL with escapes, an eval frame). A line counts as a frame only if it carries an engine's marker, and the `@` form is matched as a **shape** rather than as a bare character: a function name holds no whitespace and neither does a location, so `name@location` has none at all. That distinction is load-bearing — searching for the character alone accepts any message naming a scoped module or an account, and such a message ending in a parseable `url:line:col` would then be read as frame zero, costing the header its budget and the reader every frame. Belt and braces on the same hazard: no single line may spend the whole stack budget, and two lines are kept when the stack has them, so the split between message and frames never has to be perfect. A frame's function and URL are bounded too, since both land in dashboard labels.

An empty `stack` is not a missing capture: it means the throw carried no error object, which happens for a cross-origin script error and for a promise rejected with a non-`Error`. Reporting empty is deliberate — a synthesized stack would name this instrument rather than the throw.

## Why a storm can't drown the other signals

Identical errors coalesce in place. An event's `dedup_count` is incremented while that event is still in the ring buffer, and the flush serializes the buffer at send time, so the count is accurate on the wire without a second event. A render loop throwing 300 times reports one event carrying 300.

Distinct errors are bounded too: past a per-window event budget, further new signatures fold into one shared slot, and `folded_signatures` says so — `0` on an ordinary event, and on a folded one the number of *distinct* errors the slot stands for, while `dedup_count` stays the occurrence total. The two have to be able to differ, because that difference is the whole signal: without it, four errors seen 25 times each is byte-identical to one error looping a hundred times. So a window emits at most 12 error events however many distinct errors arrive, against a 400-event ring buffer — that ratio is what keeps an error storm from crowding out card-load, server-request, or rebuild events. (The bound is on emissions per window, not a hard cap on buffered error events: a buffer already at capacity evicts the slot an event is tracked by, so a saturated window can run somewhat over before the buffer drains.)

Dedup slots are pruned at each transmission, since a sent event can no longer absorb a repeat. The shared slot is dropped unconditionally there, even while still buffered: it is the one slot mixing unrelated errors, so letting it span windows would keep folding new ones onto a message and timestamp from an older window.

The dedup signature is `kind` + `message_key` + the top frame, where `message_key` is the message with the parts that vary per occurrence — uuids, hashes, tokens of four or more characters containing a digit — collapsed to `*`. That is what makes dedup work on a message carrying an instance id, and it doubles as the board's grouping key so one logical failure is one row instead of one row per card. Short numbers survive, so an HTTP status stays readable; a short or digit-free id does not collapse, so the key is a grouping aid rather than a guarantee of one row per failure.

The key derives from the whole message, not the truncated one, and past its own budget it ends in a digest of the full text. Two failures carrying a serialized document share a long prefix and differ only near the end — keying off either truncation would merge them and report only whichever arrived first. Keys also collapse whitespace, since they are read in a table cell. `message` keeps the verbatim text.

## Never during indexing

A prerender tab throws routinely — a card with a broken link or a bad module is an expected render outcome — so this channel has to be silent there or every indexing job would beacon its own render failures. Two gates:

- **Arm time**, which the instrument already had: it refuses to arm when `__boxelRenderContext` is set, and the page pool injects that flag with `evaluateOnNewDocument`, before any page script runs. So the flag is already true at app boot; the eager initializer skips, a later service lookup constructs an instrument whose `start()` refuses, and `sessionStarted()` re-checks so a Matrix session arriving later cannot arm it either. The render surfaces only ever clear the flag under `isTesting()`, so a production prerender tab never drops it between the many renders one job performs.
- **Capture time**, added here: both handlers bail on the same flag regardless of how the instrument came to be armed. That makes the guarantee local to this path rather than dependent on boot ordering, and it also covers a live tab that navigates onto a render-surface route, where the instrument armed legitimately at boot and the route then sets the flag.

Under tests the existing gate already means "unless this is what's being tested": dormant under `isTesting()` until a test calls `enableForTest()`.

## The board

Overview row

The Overview row's three 5-minute tiles drop to `w=6` and a fourth joins them, and the Signal event rate panel gains the `client-error` series — so a board that is only ever glanced at still says the app is throwing. The tile counts events rather than occurrences, which its description says, because a looping error reads small there by design.

client-error row

The row sits after `wedge` — its sibling in "what went wrong" — with the rows below shifted by two, headers and nested children together, so every row keeps `firstChild.y == row.y + 1` and a `pull.sh` round-trip after someone expands and saves a row doesn't renumber the rest.

- **client-error rate & occurrences** plots events by `kind` against summed `dedup_count` on the right axis. Those diverge under a storm and the divergence is the diagnosis: a flat event line over a climbing occurrence line is one error looping, not a quiet period. In the screenshot, 2 events and 302 occurrences.
- **Top errors by message** groups by `(message_key, kind, top_frame_function)`, so a row names what broke, whether it was a throw or a rejection, and the function it came from — the part that makes a row actionable.
- **Errors by user** is the pivot from "a user says the app broke" to their account and what broke for them.
- **client-error detail** is raw lines, because a stack cannot be aggregated and reading it is the whole point.

Every query keeps the `|= "boxel:client-perf"` line filter ahead of the first parser and carries the user/session filters inside the range selector it aggregates over, matching the rest of the board.

## Verification

Local dev stack with the observability stack up, Grafana 12.4.3, beacons reaching Loki through the realm-server's stdout:

- A real uncaught throw and a real unhandled rejection from a logged-in tab each produced one event, with the correct `kind`, message, `top_frame_function` (`computeTitle`, `reestablishReferences`), `source_url` / `line` / `col`, and the full stack — read off the raw Loki line and checked against the actual throw site, not just for non-emptiness.
- A 240-iteration identical-throw loop plus a standalone throw produced **one** event with `dedup_count: 241`, while `deserialize` and `server-request` events continued to flow in the same window.
- No `client-error` line came from the boot reindex, which failed 10 renders in the prerender log over the same period.
- Field survival through the wrapper-unwrap parse, checked by pushing a FireLens-shaped line (`{"log": …, "source": "stdout", "container_id": …, "ecs_*": …}`) into Loki and running the board's exact parse: every new field extracts with the event's own value, including the multi-line `stack`. The wrapper's own keys land as labels beside them and none collide — only `env` does, against the stream label, as it already did.
- All four new panels and the two edited Overview panels run against real data (queries executed directly against Loki, then confirmed rendering in Grafana).
- Expanding every row yields zero overlapping grid items in the collapsed view and none within any row's own layout.
- 35 unit tests over the report builder and 13 integration tests over the instrument, split by what each can actually prove. The unit tests own the pure surface: the frame parse across seven engine and bundler shapes, a message shaped like a frame, all four shapes of oversized message, the frame/header/character budgets, the grouping key and its digest tail, non-`Error` and unstringifiable reasons, and a cross-origin `Script error.`. The integration tests own the live surface: both listeners wired, dedup coalescing across the buffer, the window budget and its folded-group accounting, `isTesting()` dormancy, the render-context gate, and teardown. Run against the built dist under testem — the harness CI uses — not only against the dev server: dispatching a synthetic `ErrorEvent` at the window reaches `window.onerror`, which testem's client reports as a failure independently of QUnit, so the two harnesses do not agree unless the property itself is detached for the dispatch. A probe asserts no global handler observes it, which holds under either runner.
- `packages/observability` lint passes; ESLint and prettier clean on the changed files.

Known limitation: a message that opens with `at ` *and* ends in a parseable `url:line:col` would still be read as a frame. The line caps mean the frames survive either way, but the location fields would name what the message quoted.


packages/host/app/services/client-telemetry.ts
```diff
@@ -1233,6 +1415,170 @@ function codeRefURL(
...
+  if (typeof stack !== 'string' || stack.length === 0) {
+    return '';
+  }
+  let frames = stack
+    .split('\n')
+    .map((frame) => frame.trim())
+    .filter(Boolean)
+    .slice(0, MAX_ERROR_STACK_FRAMES);
+  let out = frames.join('\n');
+  while (out.length > MAX_ERROR_STACK_CHARS && frames.length > 1) {
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed and fixed in `d85155ce06`.

Reproduced the exact path: a 2,500-character message line with two frames below it yielded a 2,000-character `stack` holding **one** line — the truncated header — and `topStackFrame` then found no frame, so `top_frame_function` / `source_url` / `line` / `col` all came back empty. The message field's own 300-character cap doesn't help, because the stack's header comes straight off `error.stack`.

Two changes, since the finding has two halves:

- The stack's opening line is now bounded against a header budget of its own (500 characters) before deep frames are dropped, so the header can no longer consume the whole allowance. Truncation appends `…` so a cut value doesn't read as the whole one. A 4,000-character message with three frames now emits header + all three frames at 587 characters.
- The location fields are read from the stack **as the engine gave it** rather than from the bounded copy. Where a throw happened is a fact about the error; it shouldn't depend on how much of the stack fit in a budget. That makes those four fields correct even if a future budget change truncates differently.

Covered by a test that asserts the header is cut and marked, the throw site is still the first frame, the frames below it survive, and the location fields are populated.

packages/host/app/services/client-telemetry.ts
```diff
@@ -1233,6 +1415,170 @@ function codeRefURL(
...
+  let rest = candidate.replace(/^at\s+/, '');
+  let functionName = '';
+  let location = rest;
+  let open = rest.indexOf(' (');
+  if (open !== -1 && rest.endsWith(')')) {
+    functionName = rest.slice(0, open);
+    location = rest.slice(open + 2, -1);
+  } else {
+    let at = rest.indexOf('@');
+    if (at > 0) {
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed and fixed in `d85155ce06`. `@https://example/app.js:12:34` parsed its url as `@https://example/app.js`, exactly as described.

Chasing it surfaced a second defect in the same branch, and a more reachable one. The `@` fallback was reached for **V8** frames too, and a `@` inside a V8 location is part of the url — `@cardstack/…` and `@ember/…` are ordinary path segments in this app's bundle urls. So `at https://host/assets/@cardstack/boxel-ui.js:1:2` split at that `@` and produced `top_frame_function: "at https://host/assets/"` with `source_url: "cardstack/boxel-ui.js"`. Every anonymous frame in a scoped-package module was landing wrong, which on a minified build is most of them.

Both come from the same wrong premise — that the delimiter is whichever character appears first. It isn't; it's decided by the engine that wrote the frame. The parse now keys off the `at` prefix: V8 frames never take the `@` split, and for the frames that do, index 0 is accepted as a valid position meaning the function name is empty.

The frame-shape test gained both cases (scoped package in a V8 location, anonymous SpiderMonkey frame) alongside the existing V8 / paren-url / `@`-in-url / async-method ones.

habdelra:
[Claude Code 🤖] `Host Tests (3, 20)` was red on this branch and it was mine, not a flake — fixed in `248909175a`.

```
Global error: Uncaught boom at https://realm.example/my-realm/person.gts, line 42
 While executing test: … records a client-error event for an uncaught window error
```

An `ErrorEvent` dispatched at the window invokes `window.onerror`, and two harnesses hook it independently. The module suspends QUnit's uncaught-exception path, but `testem_client.js` wraps `window.onerror` separately and reports anything reaching it to its server as a `Global error`. Suspending QUnit's hook does nothing about testem's, so the test failed under the runner CI uses while passing against the vite dev server, which has no testem client. Only one test failed because every other uncaught dispatch already went through a helper that nulled `window.onerror` — the one unwrapped call site was the whole bug.

Detaching the property is what covers every such handler, so that now happens inside the dispatch helper rather than at each call site where it can be forgotten, and `assertNoGlobalErrorReported` probes with a handler of its own and asserts it is never invoked. That assertion fails under either runner, so this class of harness-dependent failure is now caught locally.

Verifying the fix the same way CI does surfaced a second, latent flake in it: the replacement assertion counted the suspension collector's entries, and that collector spans the module — the app's own boot leaves an incidental `TypeError: Failed to fetch` rejection in it during `beforeEach`, so the count read 2. It now asserts containment of the rejection the test dispatched, which is what it meant and which doesn't depend on unrelated environment noise.

18/18 green against the built dist under testem, exit 0, zero global errors reported.

habdelra:
[Claude Code 🤖] Pushed `9808abbe75`; the description carries the current shape.

packages/host/app/services/client-telemetry.ts
habdelra, line 1: this module seems pretty big and it looks like there are a bunch of functions in module scope in this service that could probably exist fine in a separte file. let's refactor this service and move things out of this file that can safely exist outside

habdelra: [Claude Code 🤖] Agreed — extracted in `32e76f23e9`. The error cluster (~235 lines, the largest module-scope block and the one this branch added) is now `app/lib/client-error-report.ts`, and the service drops from ~1720 to 1431 lines.

I split it along what looked like a real seam rather than purely by size. What a throw *is* — message, stack, throw site, and the identity repeats coalesce under — is pure: the same event always yields the same report. How many events to emit and when is stateful, owned by the ring buffer, the dedup slots, and the window budget. So the new module exposes two entry points, one per channel, returning one `ErrorReport`, and the service's handlers are a line each.

The better payoff was testability. With the pure half importable, the exhaustive cases become a fast unit test with no window listener, no armed instrument, and no realm: `tests/unit/client-error-report-test.ts` now owns seven frame shapes across engines and bundlers, four shapes of oversized message, the frame/header/character budgets, and the grouping key including its digest tail. I moved 9 tests out of the integration suite, which keeps only what a live service can show — listeners wired, repeats coalescing across the buffer, the window budget folding and accounting for the rest, and both gates. 28 unit + 12 integration, green under testem, and re-verified live: a real uncaught throw and a real rejection still beacon correctly, with two identical throws coalescing to one event at `dedup_count: 2`.

One scope question for you. The other module-scope functions in the service — `normalizeEndpoint`, `codeRefURL`, `now`, `byteLength`, `isDocumentHidden`, `supportedEntryTypes`, `nextTick` — could move the same way, but they predate this branch and span three unrelated concerns (endpoint labels, code refs, timing utils), so pulling them here would add churn to an already-large diff. Happy either way: say the word and I'll do it in this PR, otherwise I'll leave them.

jurgenwerk (approved):

habdelra:
[Claude Code 🤖] Pushed `339dfbd628`; the description carries the current shape.

packages/host/app/services/client-telemetry.ts
habdelra: yeah lets move those too

habdelra: [Claude Code 🤖] Will do. Queuing it behind an independent review pass that is mid-flight over `client-telemetry.ts` right now — I moved code under a reviewer earlier in this PR and it invalidated its file reads, so it had to restart. Rather than repeat that, I'll push the move as soon as it reports and follow up here with the SHA.

Plan when I do: `normalizeEndpoint` and `codeRefURL` go together — both turn a url or a code ref into a low-cardinality dashboard label, and both are worth unit-testing directly, which is the real payoff. The environment probes (`now`, `nextTick`, `isDocumentHidden`, `supportedEntryTypes`, `byteLength`) are a second, smaller group: defensive wrappers over browser APIs. That leaves the service holding the event schema, the tunables, the buffer/flush machinery, and the middleware.

I'll leave `createServerRequestTimingMiddleware` where it is unless you'd rather it moved — it's exported and imported by `network.ts`, and it exists to look up the service, so it reads as part of the service's surface rather than a helper stranded in module scope.

habdelra: [Claude Code 🤖] Done in `b0a8c6755a`. `client-telemetry.ts` is 1374 lines, from ~1720 before the first extraction.

Split by what they are for rather than into one file of leftovers. `lib/client-telemetry-labels.ts` holds `normalizeEndpoint` and `codeRefURL` — both exist to keep dashboard cardinality low, since an endpoint or card type carrying a per-instance id would give Loki a series per card. `lib/client-telemetry-env.ts` holds `now`, `isDocumentHidden`, `byteLength`, `supportedEntryTypes`, `nextTick`, and `isRenderContext`: guarded reads of browser APIs the instrument cannot assume exist. I pulled `isRenderContext` along too — it wasn't in the list I posted, but it's the same category and it's the prerender gate several call sites read.

The payoff again is testability: both label functions were reachable only through the service, so they were only tested through it. `tests/unit/client-telemetry-labels-test.ts` now pins what they collapse and what they keep — an instance id becoming `GET card`, two instances of one type sharing a label, a relative module resolving against the instance it came from, a scoped specifier passing through, and the refs that name no export.

`createServerRequestTimingMiddleware` is now the only thing left in module scope, and I left it there deliberately: it's exported, `network.ts` imports it, and its whole job is looking up this service, so it reads as part of the service's surface rather than a helper stranded beside it. Happy to move it if you'd rather.

14 + 42 + 13 tests green under testem.

burieberry:
[Claude Code 🤖]

A few issues in `client-telemetry.ts` / `client-error-report.ts`, verified against this branch:

**1. `resetState()` isn't exception-safe, unlike `teardown()`** (`client-telemetry.ts:400`)

```ts
resetState(): void {
  this.#flush('reset');
  this.#buffer = [];
  this.#matrixUserId = null;
  this.stop();
}
```

`teardown()` wraps its own `#flush('teardown')` call in `try/finally` specifically so a throw during flush can't skip cleanup. `resetState()` doesn't have that guard: if `#flush('reset')` throws (e.g. `JSON.stringify` in `#buildBody` choking on a malformed buffered event), the buffer never clears and `#matrixUserId` never resets. On logout that means the previous user's buffered error text/stack can survive into the next session and get beaconed under the new account's `matrix_user_id`.

**2. `#errorEventsSinceFlush` goes stale outside of successful flushes** (`client-telemetry.ts:806`)

`#pruneErrorDedup()` — the only thing that resyncs `#errorEventsSinceFlush` to the real tracked-error count — is called from inside `#flush()` (line 682), but only after three early-return paths (empty buffer, missing token, origin lookup throw). `#push()`'s own `MAX_BUFFERED_EVENTS` FIFO trim (line 718) splices buffered events out directly without touching `#errorDedup` or the counter. If tracked errors get evicted by either path, `#errorEventsSinceFlush` stays inflated, so the next distinct error incorrectly takes the shared-overflow path and loses its own message/stack.

**3. `drainBufferForTest()` skips the same resync** (`client-telemetry.ts:388`)

```ts
drainBufferForTest(): TelemetryEvent[] {
  let events = this.#buffer;
  this.#buffer = [];
  return events;
}
```

Every other buffer-clearing path either goes through `#flush`'s `#pruneErrorDedup()` or (per #1) should. This one doesn't reset `#errorEventsSinceFlush`, so a test that drains after the overflow threshold is hit and then reports one more error will incorrectly see it fold into the overflow slot — masking a real dedup bug in tests written against this helper.

**4. `isRenderContext()` guard duplicated as a third ad-hoc call site** (`client-telemetry.ts:746`)

The comment above the check already acknowledges this is intentional, repeated here rather than enforced once at listener-attachment time. Worth flagging anyway: a future signal type has to remember to re-add this same guard by hand, or render-context noise leaks into the pipeline the same way this PR is fixing for wedge/errors.

**5. Linear scan on every error occurrence** (`client-telemetry.ts:758`)

`this.#buffer.includes(pending)` in `#reportError` is an O(buffer-length) scan (buffer capped at 400) run synchronously inside the global error handler on every single error. Under a render-loop error storm this adds up; pruning `#errorDedup` at `#push`-eviction time instead of leaving `#reportError` to discover staleness reactively would avoid it.

**6. Third near-duplicate truncation helper** (`client-error-report.ts:163`)

`truncate()` here duplicates two existing, slightly different helpers already in the host app: `error-display.gts`'s `truncate()` and `send-error-to-ai-assistant.gts`'s `truncateForAi()`. The two pre-existing ones already disagree on ellipsis suffix; this adds a third convention rather than reusing or unifying them.

habdelra:
[Claude Code 🤖] All six verified against the branch. Five fixed in `9e1bd97315`; the sixth I'd push back on.

**1 — confirmed and fixed.** Worse than it looks: `#buildBody` runs outside the flush's own `try`, so a `JSON.stringify` throw propagates straight out of `resetState`. That skips the buffer clear, which is the step added one commit earlier specifically to stop one account's error text reaching the next account's id. It now clears in a `finally`, as `teardown` already did.

**2, 3 and 5 — confirmed, and they share one root cause,** so they're fixed as one change. The map was allowed to name events that had already left the buffer, and `#reportError` discovered that reactively via the `includes` scan you flagged. The invariant is now explicit — the map holds exactly the error events still in the buffer — maintained at the three places events leave one: the flush's send, the eviction in `#push`, and the test drain. A hit therefore means "still unsent" by construction, so the scan is gone, the window count is read from the map rather than tracked beside it, and the drain no longer hides the behavior a test written against it means to check. Your point about the pre-login case was the sharpest part: with no token the flush returns early, so nothing ever resynced.

**6 — confirmed, and understated.** The two pre-existing helpers are duplicates *of each other*: same `' …[truncated]'` suffix, same reserve-room-for-the-marker arithmetic, each unexported in a component file. So it was three copies of one function. There's now one `lib/truncate.ts` used by all three, with the marker as a parameter because the audiences genuinely differ — prose a person reads can afford `[truncated]`, a 160-character grouping key cannot. I checked the shared helper against the original implementation across the input space (56 combinations of value and budget, including the boundaries either side of the suffix length) before swapping it in, and the two existing tests that pin the suffix and the length cap still pass.

**4 — I'd leave as is, and I think the suggested direction removes the guarantee.** Enforcing once at listener-attachment time is what arm-time gating already does: `#isAllowed()` refuses to arm when the render-context flag is set. The capture-time check exists precisely because arm-time is not sufficient — a live tab that navigates onto a render-surface route sets the flag while the instrument is legitimately armed, and only a check at capture time keeps that quiet. Your underlying concern is fair, though: a future signal type does have to remember. If that becomes a pattern rather than one channel, the right shape is for the emit path itself to carry the check once, rather than each handler.

42 + 14 + 13 + 9 tests green under testem, including the ErrorDisplay truncation tests.

PR #5642, opened 2026-07-29
Fix host mode stack navbar overflow on mobile

burieberry:
## Summary
- Cap the host-mode breadcrumb pill to the available content width; middle crumbs shrink first (first and last are protected with a wider floor), and trails too long to fit fall back to a horizontal scroller that keeps the current crumb in view.
- Tighten pill padding and icon size at narrow widths.
- Render the current-page crumb as a non-interactive `span[aria-current="page"]` instead of a disabled button, and enlarge the touch target toward the WCAG 2.5.8 minimum.
- Key the crumb list by card id for stable identity across re-renders.
- Resolve breadcrumb clicks by index rather than card id, so a trail that contains the same card twice closes the correct cards (the stack allows duplicate ids).

Fixes CS-12375.

Follow-up: CS-12395 tracks replacing the scroll fallback with the standard breadcrumb collapse pattern (`First › … › Current` with the buried crumbs in an ellipsis menu).

## Test plan
- [x] In host submode at a phone-width viewport, open stacks of 2–3 and 4+ cards: the pill stays within the viewport, and the current crumb is visible when the trail scrolls
- [x] Host acceptance suites (`host-submode-test.gts`, `host-mode-test.gts`) pass
- [x] Screen reader announces the current crumb via `aria-current="page"`

----

Before:
stacknav-before

After: (fallback case for long title that scrolls)
stacknav-after

### DIFF
--- a/packages/host/app/components/host-mode/breadcrumb-item.gts
+++ b/packages/host/app/components/host-mode/breadcrumb-item.gts
@@ -1,3 +1,4 @@
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
 import { on } from '@ember/modifier';
 import { action } from '@ember/object';
 
@@ -19,6 +20,65 @@ interface Signature {
     onClick?: (cardId: string) => void;
   };
 }
+
+interface BreadcrumbItemContentSignature {
+  Args: {
+    icon?: ComponentLike<{ Element: SVGSVGElement }>;
+    hasCard: boolean;
+    isLoading: boolean;
+    label: string;
+  };
+}
+
+const BreadcrumbItemContent: TemplateOnlyComponent<BreadcrumbItemContentSignature> =
+  <template>
+    {{#if @hasCard}}
+      {{#if @icon}}
+        <@icon
+          class='breadcrumb-item-icon'
+          width='18'
+          height='18'
+          aria-hidden='true'
+        />
+      {{/if}}
+      <span class='label'>{{@label}}</span>
+    {{else if @isLoading}}
+      <span class='label muted'>
+        Loading…
+      </span>
+    {{else}}
+      <span class='label'>{{@label}}</span>
+    {{/if}}
+
+    <style scoped>
+      .breadcrumb-item-icon {
+        display: inline-flex;
+        align-items: center;
+        justify-content: center;
+        flex-shrink: 0;
+        color: var(--boxel-highlight);
+      }
+
+      @media (max-width: 30rem) {
+        .breadcrumb-item-icon {
+          width: 0.75rem;
+          height: 0.75rem;
+        }
+      }
+
+      .label {
+        font: 500 var(--boxel-font-sm);
+        color: var(--boxel-light);
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
+
+      .muted {
+        opacity: 0.7;
+        font-weight: 500;
+      }
+    </style>
+  </template>;
 
 export default class HostModeBreadcrumbItem extends Component<Signature> {
   @cached
@@ -43,12 +103,18 @@ export default class HostModeBreadcrumbItem extends Component<Signature> {
     return Boolean(this.args.cardId) && !this.card && !this.cardError;
   }
 
-  private get iconComponent(): ComponentLike | undefined {
+  private get iconComponent():
+    | ComponentLike<{
+        Element: SVGSVGElement;
+      }>
+    | undefined {
     if (!this.card) {
       return undefined;
     }
 
-    return cardTypeIcon(this.card) as ComponentLike | undefined;
+    return cardTypeIcon(this.card) as
+      | ComponentLike<{ Element: SVGSVGElement }>
+      | undefined;
   }
 
   private get label() {
@@ -87,23 +153,12 @@ export default class HostModeBreadcrumbItem extends Component<Signature> {
           title={{this.label}}
           data-test-host-mode-breadcrumb={{@cardId}}
         >
-          {{#if this.card}}
-            {{#if Icon}}
-              <Icon
-                class='breadcrumb-item-icon'
-                width='18'
-                height='18'
-                aria-hidden='true'
-              />
-            {{/if}}
-            <span class='label'>{{this.label}}</span>
-          {{else if this.isLoading}}
-            <span class='label muted'>
-              Loading…
-            </span>
-          {{else}}
-            <span class='label'>{{this.label}}</span>
-          {{/if}}
+          <BreadcrumbItemContent
+            @icon={{Icon}}
+            @hasCard={{if this.card true false}}
+            @isLoading={{this.isLoading}}
+            @label={{this.label}}
+          />
         </span>
       {{else}}
         <button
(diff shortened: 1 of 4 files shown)


packages/host/app/components/host-mode/breadcrumbs.gts
```diff
@@ -95,37 +96,63 @@ export default class HostModeBreadcrumbs extends Component<Signature> {
...
       .list {
         display: inline-flex;
         list-style: none;
         gap: var(--boxel-sp-xs);
         margin: 0;
         padding: 0;
         align-items: center;
+        max-width: 100%;
+        overflow-x: auto;
+        scrollbar-width: none;
```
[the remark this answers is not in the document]
burieberry: [Claude Code 🤖] Fixed in 466c7e1cf0. Added a small `ember-modifier` on the crumb list that sets `scrollLeft = scrollWidth` on mount and whenever the card count changes, so the `aria-current="page"` crumb (always the rightmost one) stays in view instead of being scrolled off behind the hidden scrollbar.

lukemelia (approved):

packages/host/app/components/host-mode/breadcrumbs.gts
```diff
@@ -14,6 +16,19 @@ interface Signature {
...
+// stranded off-screen behind a hidden scrollbar — both on mount (a
+// deep-linked stack can already overflow) and whenever the trail changes.
+let scrollToCurrentCrumb = modifier(
+  (element: HTMLElement, [cardCount]: [number]) => {
+    if (cardCount === 0) {
+      return;
+    }
+    element.scrollLeft = element.scrollWidth;
+  },
+);
```
habdelra, line 31: [Claude Code 🤖] **This pins the current crumb on mount-time overflow, but not on the more common cold-load overflow — the `aria-current="page"` crumb can still end up scrolled off-screen.**

**Mechanism.** The modifier re-runs only when its positional arg `this.cardIds.length` changes; the body reads `element.scrollWidth`/`scrollLeft`, which are plain DOM reads (not tracked), so nothing else re-fires it. On a cold deep-link visit `getCard` resolves each crumb's title asynchronously — at install time every crumb renders `Loading…` (the `breadcrumb item shows loading state before card is available` acceptance test exercises exactly this: the current crumb starts as `Loading…` and only later becomes its real title). The `Loading…` trail is narrow and usually fits, so `scrollLeft = scrollWidth` lands at ~0. When the titles resolve the trail grows and overflows — but `cardIds.length` is unchanged, so the modifier never re-runs and the current crumb stays off the right edge. The off-screen state is only avoided when the overflow already exists at install (warm cache, or a trail long enough to overflow even while loading).

**Way out.** Re-pin when the rendered width changes, not just when the count does — e.g. a `ResizeObserver` on the `<ol>` that sets `scrollLeft = scrollWidth` while the user hasn't manually scrolled, or thread a value that changes as titles resolve into the modifier's args.

**Scope.** Non-blocking follow-up, not a regression — before this change there was no scroll handling at all, so this is strictly better. The scroll fallback is already slated for replacement by the standard collapse pattern, and this can fold into that work; flagging so the current fix isn't assumed to cover the cold deep-link case, which is the one the thread above set out to fix. (Evidence here is reasoned from the modifier's arg dependency plus the loading-state test, not a browser repro.)

---

packages/host/app/components/host-mode/breadcrumbs.gts
```diff
@@ -64,20 +79,21 @@ export default class HostModeBreadcrumbs extends Component<Signature> {
...
       hidden={{not this.hasCards}}
       data-host-mode-breadcrumbs
       data-test-host-mode-breadcrumbs
       ...attributes
     >
       {{#if this.hasCards}}
-        <ol class='list'>
-          {{#each this.cardIds as |cardId index|}}
+        <ol class='list' {{scrollToCurrentCrumb this.cardIds.length}}>
+          {{#each this.cardIds key='@identity' as |cardId index|}}
```
habdelra, line 93: [Claude Code 🤖] **Confirmation + one edge to check (non-blocking).** Keying by `@identity` on a `string[]` keys each row by the card-id string itself, which is the right choice for stable component identity when the trail is spliced — closing a middle card keeps the surviving crumbs' `getCard` resources mounted instead of re-instantiating them by shifted index.

One caveat: `@identity` requires the keys to be unique. `cardIds` is `[primaryCardId, ...stackItemCardIds]` (from `content.gts`), so if the same card id can ever appear twice in one stack, two rows share a key and Glimmer asserts on the duplicate in dev. If a stack can't contain the same card twice, this is a non-issue — worth a quick confirmation.

---

packages/host/app/components/host-mode/breadcrumb-item.gts
```diff
@@ -11,14 +12,74 @@ import { getCard } from '@cardstack/host/resources/card-resource';
...
+          height: 0.75rem;
+        }
+      }
+
+      .label {
+        font: 500 var(--boxel-font-sm);
+        color: var(--boxel-light);
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
```
habdelra, line 74: [Claude Code 🤖] **Adjacent / optional (pre-existing, relocated here).** The `font` shorthand resets `font-family`, so it drops the inherited themed family; `--boxel-font-sm` is itself a `size/line-height family` token, which compounds it. The house convention is to set the axes explicitly (split `--boxel-font-size-*` + `--boxel-line-height-*` tokens both exist and expand to the same values):

```suggestion
      .label {
        font-weight: 500;
        font-size: var(--boxel-font-size-sm);
        line-height: var(--boxel-line-height-sm);
        color: var(--boxel-light);
        overflow: hidden;
        text-overflow: ellipsis;
      }
```

Carried over verbatim from the old `.label` rule, so not introduced by this PR — flagging only because it's being re-typed into the new `BreadcrumbItemContent`. Non-blocking.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with a focus on the a11y contract this PR establishes (current crumb as `span[aria-current="page"]` vs. interactive buttons), the overflow/scroll fallback behavior, and the boxel-ui CSS conventions the styling touches.

**Bottom line: no blocking issues — the change is sound and lands the a11y improvement cleanly. One non-blocking follow-up worth knowing about (the scroll-into-view has a timing gap on cold deep-links), plus two minor notes.** A human approval is already on it; this review is COMMENT only.

**What lands right**
- Replacing the disabled current-page `<button>` with a non-interactive `span[aria-current="page"]` is the correct a11y move — disabled buttons are inconsistently announced and are still a tab stop in some engines, whereas `aria-current="page"` is exactly the semantic for "this is where you are." Scoping the focus outline to `button.breadcrumb-item:focus-visible` correctly follows through, since the span is no longer focusable.
- The icon color change `--boxel-teal` → `--boxel-highlight` follows the semantic-token convention and is visually identical today (`--boxel-highlight: var(--boxel-teal)` in `variables.css:207`). Same story for `--boxel-sp-xxs` → `--boxel-sp-2xs` (the `xxs` token is explicitly deprecated in favor of `2xs` and resolves to the same value), and for promoting the `rgba(255,255,255,…)` literals to the new `--boxel-light-60/70` tokens.
- The overflow construction is well built: `.breadcrumb-item` and `.label` both carry `overflow: hidden`, which zeroes their automatic flex min-size, so middle crumbs genuinely shrink-with-ellipsis before the `.list` falls back to `overflow-x: auto`. First/last get a wider `min-width: 4rem` floor so the anchor crumbs stay legible — matching the description.
- `min-height: 1.5rem` (24px) on the crumb meets the WCAG 2.5.8 minimum target height.

**Recommendations**
1. Scroll-into-view re-pins only when `cardIds.length` changes, so on a cold deep-link the current crumb can still scroll off-screen once titles resolve — see the thread on `scrollToCurrentCrumb` in `breadcrumbs.gts`. Non-blocking; foldable into the collapse-pattern follow-up.
2. Confirm a stack can't contain the same card id twice, since `key='@identity'` asserts on duplicate keys — see the thread on the `{{#each}}` in `breadcrumbs.gts`. Non-blocking.
3. Optional: drop the `font:` shorthand in the relocated `.label` rule — see the thread in `breadcrumb-item.gts`. Pre-existing, non-blocking.

**Adjacent, out of scope**
- No test changes accompany the diff (the 4 changed files are 3 components + `variables.css`). The a11y contract this PR is specifically about is easy to pin in the existing `host-submode-test.gts` breadcrumb suite: assert the current crumb is a `span[aria-current="page"]` and the non-current crumbs are `button`s. Worth adding so a future refactor can't silently regress the span/button split — but genuinely a follow-up, not asked of this PR.

---

packages/host/app/components/host-mode/breadcrumb-item.gts
burieberry: this is fine because the breadcrumbs belong to the host chrome and do not get themed

PR #5643, opened 2026-07-29
ci: Add warning when type-checking is lost

backspace:
#5584 fixed type failures that were masked by `skipLibCheck`. This is about warning us if type checks are lost again. [Here](https://github.com/cardstack/boxel/actions/runs/30542495024/job/90870365110?pr=5643)’s an artificially-provoked example failure.

### DIFF
--- a/packages/boxel-ui/test-app/tests/boxel-ui-type-canary.gts
+++ b/packages/boxel-ui/test-app/tests/boxel-ui-type-canary.gts
@@ -1,0 +1,37 @@
+// Type-checking canary for @cardstack/boxel-ui. Never imported at runtime.
+//
+// Broken declaration re-exports once silently degraded every boxel-ui symbol
+// to `any` under skipLibCheck, so the type check passed while checking
+// nothing. Degraded symbols accept values real ones reject, so each
+// assertion here expects an error that a degraded symbol cannot produce:
+// if a directive below reports as unused, that barrel has stopped being
+// type-checked and the lint fails.
+//
+// (Positive detection à la `IsAny<typeof X>` is NOT sufficient: an import
+// from an unresolvable module is TypeScript's error type, which behaves
+// like `any` in value positions but evades `IsAny` in conditional types.)
+import type { Avatar } from '@cardstack/boxel-ui/components';
+import { BoxelButton } from '@cardstack/boxel-ui/components';
+import type { cn } from '@cardstack/boxel-ui/helpers';
+import type { ArrowLeft } from '@cardstack/boxel-ui/icons';
+import type { setCssVar } from '@cardstack/boxel-ui/modifiers';
+
+// @ts-expect-error a number is not a component class
+export const componentsBarrelIsTyped: typeof Avatar = 42;
+
+// @ts-expect-error a number is not the cn helper function
+export const helpersBarrelIsTyped: typeof cn = 42;
+
+// @ts-expect-error a number is not an icon component
+export const iconsBarrelIsTyped: typeof ArrowLeft = 42;
+
+// @ts-expect-error a number is not a modifier
+export const modifiersBarrelIsTyped: typeof setCssVar = 42;
+
+// If glint ever stops checking boxel-ui component invocations, the expected
+// error below never materializes and the unused expect-error directive
+// itself fails the type check.
+export const invocationCheckingIsLive = <template>
+  {{! @glint-expect-error BoxelButton takes @kind, not @variant }}
+  <BoxelButton @variant='primary'>canary</BoxelButton>
+</template>;
--- a/packages/host/tests/boxel-ui-type-canary.gts
+++ b/packages/host/tests/boxel-ui-type-canary.gts
@@ -1,0 +1,37 @@
+// Type-checking canary for @cardstack/boxel-ui. Never imported at runtime.
+//
+// Broken declaration re-exports once silently degraded every boxel-ui symbol
+// to `any` under skipLibCheck, so the type check passed while checking
+// nothing. Degraded symbols accept values real ones reject, so each
+// assertion here expects an error that a degraded symbol cannot produce:
+// if a directive below reports as unused, that barrel has stopped being
+// type-checked and the lint fails.
+//
+// (Positive detection à la `IsAny<typeof X>` is NOT sufficient: an import
+// from an unresolvable module is TypeScript's error type, which behaves
+// like `any` in value positions but evades `IsAny` in conditional types.)
+import type { Avatar } from '@cardstack/boxel-ui/components';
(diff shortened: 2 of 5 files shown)


packages/boxel-ui/test-app/tests/boxel-ui-type-canary.gts
```diff
@@ -0,0 +1,37 @@
+// Type-checking canary for @cardstack/boxel-ui. Never imported at runtime.
```
habdelra: [Claude Code 🤖]

**The "Never imported at runtime" claim is false in this copy** — it holds only in the host copy. Non-blocking, low severity: the comment misleads rather than breaks anything.

**Mechanism.** This test-app's test entrypoint eagerly imports every module under `tests/`:

```js
// packages/boxel-ui/test-app/tests/index.html
import.meta.glob('./**/*.{js,ts,gjs,gts}', { eager: true });
```

so `boxel-ui-type-canary.gts` is imported and evaluated whenever the test-app suite boots. The host copy genuinely isn't: host's entrypoint globs only `-test` modules (`import.meta.glob('./**/*-test.{js,ts,gjs,gts}')`), and this file isn't a `-test`, so it's never in the set and gets tree-shaken out. That difference is exactly why the byte-identical comment is true in host and false here.

**Why it's harmless but still worth fixing.** The module only defines inert consts and a template-only component that's never rendered, and the sole value import (`BoxelButton`) is side-effect-free, so nothing fails today. But the deliberately-invalid `<BoxelButton @variant=…>` invocation now ships inside the test-app's runtime test bundle, and a future editor trusting "never imported at runtime" could add something with real runtime cost here and be caught out.

**The fix.** Drop or soften the line — it isn't load-bearing to the canary's purpose. Something true in both homes works, e.g. `Runs no assertions at runtime; exists only for the type checker.`

Scope: regression introduced by this PR (a statement that's accurate for host copied verbatim into a package where it isn't); non-blocking.

---

burieberry: [Claude Code 🤖] "Never imported at runtime" isn't true for this copy: `tests/index.html` line 37 does `import.meta.glob('./**/*.{js,ts,gjs,gts}', { eager: true })`, which matches this file, so it's bundled and evaluated at every test-app boot. Harmless today since the module is side-effect-free, but the comment invites a future side-effectful addition that would then run before app setup during every test boot.

Either move this copy to a directory outside the glob (e.g. `type-tests/`, added to the tsconfig `include` so it still type-checks) or reword the header here. The host copy is fine as-is — host's glob only matches `*-test.*`.

burieberry (commented):
[Claude Code 🤖] Reviewed the canary setup — the two-copy design is sound (host lints against `src/*.gts` before Build Boxel UI, test-app against `declarations/*.d.ts` after, so both `exports`-map resolution modes are exercised), and the four sentinel barrels match what consumers actually import. One inline note below.

habdelra (commented):
[Claude Code 🤖]

**Lens.** I reviewed this canary as a safety net: does it actually fail when boxel-ui type-checking silently degrades, does it stay green otherwise, and is it placed where the degradation it guards can occur? I traced the enforcement mechanism, verified the pinned symbols and signatures against boxel-ui source, and read the CI Lint history rather than trusting the description.

**Bottom line: no blocking issues — this is a well-targeted safety net and the mechanism is sound.** One low-severity, non-blocking fix in the inline thread: the "Never imported at runtime" comment is false in the boxel-ui test-app copy.

- **Enforcement is real.** The canary leans on tsc's TS2578 ("Unused '@ts-expect-error' directive") and glint's equivalent for `@glint-expect-error`, both surfaced by `lint:types` (`ember-tsc --noEmit`). If a barrel degrades to `any`, `const x: typeof Avatar = 42` stops erroring, the directive goes unused, and the type check fails. Both files sit in each project's glint `include` (`tests/**/*`), so they're genuinely checked.
- **The pins are valid.** All five imported symbols exist in their barrels (`Avatar`, `BoxelButton` in components; `cn` in helpers; `ArrowLeft` in icons; `setCssVar` in modifiers), and `BoxelButton`'s `Args` is a closed interface (`kind`, no `variant`, no index signature) — so `@variant` is genuinely rejected by glint, making `invocationCheckingIsLive` a real check of live template invocation that complements the value-position assertions (a different failure mode: types resolving but glint not checking templates).
- **Coverage matches the surface.** The four pinned barrels are exactly the type-bearing public barrels host consumes (components/helpers/icons/modifiers); `@cardstack/boxel-ui/styles` is CSS-only with no types to degrade, so its absence is correct rather than a gap.
- **Empirically proven in the gating job.** CI Lint went green → red → green across the three commits: green on the initial canary, red when the demo masked a barrel, green again after the revert. That's both true-positive and true-negative behavior demonstrated in the exact job that gates merges.

**On the duplication (non-blocking).** The two files are byte-identical. That's largely inherent — each package is a separate glint project and must carry its own directives to independently verify its own boxel-ui resolution — so I'm not asking you to dedupe. But nothing cross-references the twins, and the duplication already produced the one defect here: a comment true in host copied verbatim into test-app where it's false. A one-line "keep in sync with the copy in packages/…/tests/…" in each would make the coupling visible to the next editor.

**Recommendations:**
1. Fix/soften "Never imported at runtime" in the test-app copy — see the inline thread. (low, non-blocking)
2. Optionally add a cross-reference note between the two copies. (nit)

**Adjacent, out of scope.** The demo commit masked a barrel only in the host package, so it proved the host copy's true-positive; the test-app copy's failing behavior was never exercised. Not a defect — just noting the demonstration covered one of the two homes.

---

habdelra (approved):

packages/host/tests/boxel-ui-type-canary.gts
```diff
@@ -0,0 +1,37 @@
...
+
+// @ts-expect-error a number is not a modifier
+export const modifiersBarrelIsTyped: typeof setCssVar = 42;
+
+// If glint ever stops checking boxel-ui component invocations, the expected
+// error below never materializes and the unused expect-error directive
+// itself fails the type check.
+export const invocationCheckingIsLive = <template>
+  {{! @glint-expect-error BoxelButton takes @kind, not @variant }}
+  <BoxelButton @variant='primary'>canary</BoxelButton>
```
burieberry, line 36: Awesome, this was driving me crazy

burieberry (approved):

packages/boxel-ui/test-app/tests/boxel-ui-type-canary.gts
backspace: fixed in e9c61948b3

PR #5647, opened 2026-07-30
host: Fix routing use of full URL in scoped chooser

backspace:
It says “1 result” but doesn’t show it:

image

but finding it with a substring does work:

image

The new test failed [here](https://github.com/cardstack/boxel/runs/90918239100) before the fix was added.

With this PR:

s 2026-07-30 at 15 11 52@2x

### DIFF
--- a/packages/host/app/components/search/sheet-results.gts
+++ b/packages/host/app/components/search/sheet-results.gts
@@ -239,8 +239,17 @@ export default class SheetResults extends Component<Signature> {
     this.args.pagination.showMore(sectionId, totalCount);
   }
 
-  @action isSectionCollapsed(sectionId: string): boolean {
-    return this.args.pagination.isCollapsed(sectionId);
+  @action isSectionCollapsed(section: SearchSheetSection): boolean {
+    // A pasted URL is an explicit ask for that one card. A focused
+    // ("show only") realm section — including the one seeded when the
+    // chooser is scoped to a consuming realm — must not hide it: during a
+    // URL paste the realm query sections aren't rendered at all, so
+    // collapsing the URL section would leave the result count with
+    // nothing visible under it.
+    if (section.type === 'url') {
+      return false;
+    }
+    return this.args.pagination.isCollapsed(section.sid);
   }
 
   <template>
@@ -294,7 +303,7 @@ export default class SheetResults extends Component<Signature> {
           @variant={{@variant}}
           @handleSelect={{@handleSelect}}
           @isFocused={{eq @pagination.focusedSection section.sid}}
-          @isCollapsed={{this.isSectionCollapsed section.sid}}
+          @isCollapsed={{this.isSectionCollapsed section}}
           @onFocusSection={{this.onFocusSection}}
           @getDisplayedCount={{this.getDisplayedCount}}
           @onShowMore={{this.onShowMore}}
--- a/packages/host/app/components/search/sheet-results.gts
+++ b/packages/host/app/components/search/sheet-results.gts
@@ -25,6 +25,7 @@ import {
   buildUrlSection,
   type RecentsSection,
   type SearchSheetSection,
+  type UrlSection,
 } from '@cardstack/host/utils/search/sections';
 import type { NewCardArgs } from '@cardstack/host/utils/search/types';
 
@@ -62,6 +63,10 @@ interface Signature {
     resolvedCard: CardDef | undefined;
     isCardResourceLoaded: boolean;
     realms: string[];
+    // True when `realms` is a hard scope (the chooser's realm picker is
+    // locked). A pasted URL that resolves outside the scope is then
+    // suppressed rather than offered.
+    realmsLocked?: boolean;
     baseFilter?: Filter;
     offerToCreate?: { ref: CodeRef; relativeTo: URL | undefined };
     // The recent card ids stripped of any `.json`, for most-recent-first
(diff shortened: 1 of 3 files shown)


packages/host/app/components/search/sheet-results.gts
```diff
@@ -239,8 +239,17 @@ export default class SheetResults extends Component<Signature> {
...
-    return this.args.pagination.isCollapsed(sectionId);
+  @action isSectionCollapsed(section: SearchSheetSection): boolean {
+    // A pasted URL is an explicit ask for that one card. A focused
+    // ("show only") realm section — including the one seeded when the
+    // chooser is scoped to a consuming realm — must not hide it: during a
+    // URL paste the realm query sections aren't rendered at all, so
+    // collapsing the URL section would leave the result count with
+    // nothing visible under it.
+    if (section.type === 'url') {
+      return false;
```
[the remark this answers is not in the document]
backspace: addressed in baaf6d5d9d

packages/host/app/components/search/sheet-results.gts
```diff
@@ -129,16 +134,34 @@ export default class SheetResults extends Component<Signature> {
...
+      (url) => this.network.virtualNetwork.unresolveURL(url),
+    );
+    if (
+      section &&
+      this.args.realmsLocked &&
+      !this.args.realms.includes(section.realmUrl)
+    ) {
+      return undefined;
+    }
+    return section;
```
habdelra, line 158: [Claude Code 🤖] **This suppression is an exact classifier in both directions — worth pinning down why, since nothing fails loudly if it ever drifts.**

`section.realmUrl` here is whatever `buildUrlSection` → `realmUrlForCard(urlForRealmLookup(card), this.args.realms, unresolveURL)` computed against the *same* `realms` array this line then tests. `realmUrlForCard` returns either a verbatim element of `realms` (the `startsWith` loop, or the `unresolveURL` loop that normalizes alias / RRI id forms), or — when nothing matched — a value derived from the card's own origin + parent path. So `!realms.includes(section.realmUrl)` partitions cleanly: in-scope → member → shown, out-of-scope → non-member → suppressed.

The fallback can't produce a false positive either: if its origin+parent-path result equaled a `realms` entry, then `urlForRealmLookup(card)` (the card id) would already `startsWith` that entry and the first loop would have returned it — so the fallback never yields a member the loops missed. Confirmed against the two new tests: `${testRealmURL}Pet/mango` (in-scope) renders, `${baseRealm.url}types/card` (a real base-realm Spec) suppresses.

And view-level suppression is enough to close the cross-realm-selection hole: the only route into `selectFromSearch` is a click on a rendered tile (`@handleSelect`), and the chooser modal wires no Enter/`onKeyDown` submit — with no URL tile there is nothing to select, so Go stays disabled. `allCards` moved in lock-step (`urlSection?.card.id`, not `resolvedCard?.id`), so "Select All" can't smuggle the card in either. Non-blocking confirmation.

---

packages/host/app/components/search/sheet-results.gts
```diff
@@ -239,8 +264,17 @@ export default class SheetResults extends Component<Signature> {
...
+    // ("show only") realm section — including the one seeded when the
+    // chooser is scoped to a consuming realm — must not hide it: during a
+    // URL paste the realm query sections aren't rendered at all, so
+    // collapsing the URL section would leave the result count with
+    // nothing visible under it.
+    if (section.type === 'url') {
+      return false;
+    }
+    return this.args.pagination.isCollapsed(section.sid);
   }
```
habdelra, line 278: [Claude Code 🤖] **Returning `false` for the URL section is safe as a general rule, not only for the locked chooser — good.** A realm ("show only") focus collapses every *other* section, and the seeded `realm:<consumingRealm>` focus is exactly what hid the pasted tile while the summary still read "1 result".

This override is safe beyond that seeded case because a URL paste and the realm query sections are mutually exclusive: `buildQuerySections` returns `null` when `isURL`, so during a paste the only sections are recents + url. The URL section therefore never coexists with a realm section it would legitimately be "shown only" against — it can only ever be collapsed by a *stale* focus (the seeded one, or one left over from a prior query), which is precisely what you want to ignore here. Non-blocking confirmation.

---

habdelra (commented):
[Claude Code 🤖] Reviewed for whether the locked ("hard-scoped") card chooser now correctly handles a *pasted full URL* — both the display bug (a resolved in-scope card counted but hidden) and the scope escape the automated reviewer flagged (a pasted out-of-scope URL becoming selectable).

**Bottom line: correct, complete, and well-covered by the two new tests — no blocking issues.** The fix lands on both halves cleanly, and I couldn't find a remaining path by which a cross-realm card reaches `selectFromSearch`.

**What lands right**
- The scope check is an *exact* classifier because `realmUrlForCard` resolves the card's realm against the same `realms` array the suppression then tests, and its fallback provably can't collide with an in-scope entry (detail in the `urlSection` thread). In-scope shows, out-of-scope suppresses, with no false positives or negatives.
- The escape is closed at the only reachable selection path. The chooser's sole route into `selectFromSearch` is a tile click; with the URL section gone there is no tile, so Go stays disabled — and `allCards` switched to `urlSection?.card.id` in lock-step so "Select All" is covered too. View-level suppression is therefore sufficient here; no `selectFromSearch`-level guard is needed.
- The empty-state split (out-of-scope → "is not in the realms this chooser is limited to" vs. missing → "No card found") is the right call: `${baseRealm.url}types/card` is a real Spec, so "no card found" would have been untrue and sent the user chasing a typo. The test asserts the exact wording, which pins it.
- Never-collapsing the URL section generalizes safely (detail in the `isSectionCollapsed` thread).

**On the raised cross-realm concern:** fully addressed. The suggested alternative (enforce scope inside `selectFromSearch`) isn't required given the tile-click-only selection model — though it would be a reasonable belt-and-suspenders follow-up *if* the chooser ever grows a keyboard "choose the resolved URL" affordance.

**Adjacent, out of scope (pre-existing, non-blocking):** during a URL paste in a chooser with a seeded realm focus, a non-empty Recents section still renders collapsed (its cards hidden via `display:none`) above the URL result, because `isSectionCollapsed` still collapses recents against the stale focus. Harmless and unchanged by this PR — noting it for whoever next touches paste-time section layout.

CI is green (Host + Realm Server suites). Verified the mechanism by tracing the selection paths and the realm classifier rather than re-running locally.

---

habdelra (approved):

PR #5648, opened 2026-07-30
ui: Update addons to strict template mode versions

backspace:
This updates the final non-v2 addons so we can remove `@embroider/compat`.

`ember-freestyle` is still v1 but that’s only for the UI documentation.


packages/boxel-ui/src/components/select/index.gts
```diff
@@ -11,24 +11,92 @@ import { action } from '@ember/object';
...
+    PowerSelectBeforeOptionsSignature<any, any, false>
+  >;
+}
+
+export function toSelectedItemComponent(
+  component: unknown,
+): ComponentLike<PowerSelectSelectedItemSignature<any, any, false>> {
+  return component as ComponentLike<
+    PowerSelectSelectedItemSignature<any, any, false>
+  >;
```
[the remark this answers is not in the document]
backspace: fixed in 3a2db6c

packages/boxel-ui/src/components/multi-select/index.gts
```diff
@@ -2,49 +2,48 @@ import 'ember-power-select/styles';
...
+import BoxelSelectedItem from './selected-item.gts';
+
+// The default subcomponents are generic classes whose inferred ComponentLike
+// shapes glint cannot match against power-select's expected unions; their
+// signatures are structurally compatible (verified by the suite), so pin
+// them to the expected member types once here.
+const DefaultSelectedItem = BoxelSelectedItem as unknown as ComponentLike<
+  PowerSelectSelectedItemSignature<any, any, true>
+>;
+import BoxelMultiSelectDefaultTrigger from './trigger.gts';
```
habdelra: [Claude Code 🤖] **Minor / non-blocking (readability):** the two `Default*` consts are interleaved with the `import` statements — `DefaultAfterOptions` (this hunk) sits between the `after-options` import above and the `import BoxelSelectedItem` below, and `DefaultSelectedItem` sits between that and `import BoxelMultiSelectDefaultTrigger`.

This is correct at runtime — ES module imports are hoisted, so `BoxelSelectedItem` and `BoxelMultiSelectDefaultTrigger` are bound before the consts evaluate, and there's no import cycle back into this module — and it passes `lint:js`. But visually it reads like a mistake, and the next editor has to reason about hoisting to convince themselves it's fine.

Suggest grouping the three sibling imports (`after-options`, `selected-item`, `trigger`) together, then the two `const` casts after them under the shared explanatory comment. Purely cosmetic — no behavior change and not blocking.

---

habdelra (commented):
[Claude Code 🤖] Reviewed as a strict-template-mode dependency upgrade: the lens here is whether the ember-power-select 8→9 / ember-basic-dropdown 8→9 / ember-power-calendar 1→2 API migrations landed on every consumer, and whether dropping `@embroider/compat` leaves any runtime gap.

**Bottom line: solid, no blocking issues.** The migration is complete and internally consistent, and CI backs that up — all 68 real check runs pass, including `lint:types` (ember-tsc/glint) and both the host and realm suites. Because glint is a required gate, the headline contract change here — `BoxelSelect`'s `@onChange` going from `(item) => void` to `(item: ItemT | null) => void` — is *provably* migrated at every call site: a missed handler would fail `strictFunctionTypes` assignability, so the ~20 `| null` widenings and their guards are the compiler's own twin-check made visible, not hand-tracing.

**What lands right (the parts to defend later):**
- **The null↔undefined bridge in `BoxelSelect`.** Power-select expresses "no selection" as `undefined`; app data conventionally uses `null`. Centralizing the translation in `selectedForPowerSelect` (null→undefined inbound) and `handleChange` (undefined→null outbound) means call sites bridge the two conventions once, at the boundary, instead of each re-deriving it. The redeclared `BoxelSelectArgs` (`Omit<PowerSelectArgs, 'selected'|'onChange'>` + null-tolerant re-adds) is the right shape for that.
- **The `to*Component` cast helpers.** Exporting these (rather than scattering `as ComponentLike<…>` casts) is the correct call given the stated constraint that out-of-repo realm packages can't import ember-power-select's own types — it gives them a typed seam without leaking the dependency.
- **Vendoring `ember-basic-dropdown`'s `base.scss` as `ember-basic-dropdown.css`.** This is genuinely necessary, not incidental: v9 ships those structural styles only as SCSS and injects nothing, so without them teleported dropdown content renders in normal flow. Pairing it with `setBasicDropdownConfig({ destination: 'ember-testing' })` in both test harnesses is the matching fix for the "v9 has no boot-time initializer" gap.
- **The picker test refactor.** Extracting `validateSelectAllOption` and asserting it directly is the right move — a constructor throw aborts the glimmer render mid-flight and can't be caught via `setupOnerror`, and I confirmed `testOptions` carries no `select-all` entry so the assertion genuinely fires.

**Codex finding (`toSelectedItemComponent` / `@selected`) — verified resolved.** The fix in the default trigger passes both `@selected={{@select.selected}}` and `@option={{@select.selected}}` at the same value, so a v9-typed consumer (reading `@selected`, e.g. `enum.gts`'s `SelectedItem` now reading `@selected.value`) and a legacy `@option` consumer both render. Traced end-to-end through `select/trigger.gts` → `enum.gts`; consistent. One forward note: `@option` is now permanently load-bearing for out-of-repo realm components, so it's effectively a deprecated-but-supported alias — worth an eventual deprecation path, but that's a follow-up, not this PR.

**Removing the `this.args.select.selected = [...]` mutation (multi-select + picker triggers) — correct, not a regression.** v9 makes `select.selected` read-only (which is *why* the assignment had to go), and `actions.select(newSelected)` is the proper controlled-update path — the parent's `@onChange` fires and flows the new selection back down. No behavior lost.

**Minor / adjacent (all non-blocking, nothing asked of this PR):**
1. Import/const interleaving in `multi-select/index.gts` — see the inline thread. Cosmetic.
2. The picker test now exercises `validateSelectAllOption` in isolation but no longer covers that `Picker`'s *constructor* calls it — the wiring is one line and visually verifiable, so this is an acceptable coverage trade, just noting the seam.
3. `ember-basic-dropdown.css` hardcodes `background-color: #fff` on `.ember-basic-dropdown-content` and is `@import`ed outside the `@layer` cascade (so it sits above layered styles). This faithfully reproduces the addon's shipped default and boxel's own themed dropdown selectors override it with higher specificity, so it's neutral today — but it's a raw hex in a token-convention repo and could show through as a white flash for any future unstyled/dark dropdown surface. Leave it verbatim (fidelity to the vendored source is the right call); flagging only for whoever next touches dropdown theming.

---

packages/boxel-ui/src/components/multi-select/index.gts
backspace: fixed in f45fdc1, thanks

habdelra (approved):

PR #5650, opened 2026-07-30
ci: Change host tests to balance based on duration data

backspace:
This adds a script that uses host test duration artifacts to balance tests among the 20 shards. This means less variation in each shard, and that the whole matrix should finish running earlier.

<table>
<thead>
<tr>
<th><code>main</code></th>
<th>this branch</th>
</tr>
</thead>
<tbody>
<tr>
<td>
s 2026-07-30 at 12 35 40@2x
</td>
<td>
s 2026-07-30 at 12 36 29@2x
</td>
</tr>
</tbody>
</table>

This adds to the CI step on `main` that stores memory baselines to store test timings so they’re balanced continually.


packages/host/scripts/generate-test-module-timings.mjs
```diff
@@ -0,0 +1,273 @@
...
+for (const [moduleName, time] of timeByModule) {
+  // Prefer an exact quoted/backticked occurrence of the title; fall back to
+  // a bare substring match, which catches titles that are a static prefix
+  // of a longer (partially dynamic) module() argument.
+  const quoted = testFiles.filter(({ source }) =>
+    ["'", '"', '`'].some((q) => source.includes(`${q}${moduleName}`)),
+  );
+  const candidates = quoted.length
+    ? quoted
+    : testFiles.filter(({ source }) => source.includes(moduleName));
```
habdelra: [Claude Code 🤖]

**Non-blocking (balance quality, not correctness): the opening-delimiter match mis-attributes any module whose title is a quoted prefix of another module's title.**

**Mechanism.** Tier 1 tests `source.includes(`${q}${moduleName}`)` — the opening quote plus the title, without the closing delimiter. So for `moduleName = "Integration | realm"` (declared in `integration/realm-test.gts` as `module('Integration | realm', …)`), the test is also true for `integration/realm-indexing-test.gts` (``module(`Integration | realm indexing`, …)``) and `realm-querying-test.gts`, because ``` `Integration | realm indexing` ``` contains the prefix `` `Integration | realm ``. `quoted` ends up with 5 entries, the `else` branch never runs, and `candidates.length === 5`, so the 22s that belongs to `realm-test.gts` is split evenly across five files.

**Verified.** I ran this exact script against a synthetic junit containing just these module names:
```
ambiguous: "Integration | realm" appears in 5 files; splitting evenly
./integration/realm-test.gts:                     4.4   (true ≈ 22s)
./integration/realm-indexing-test.gts:          127.8   (123.4 + 4.4 phantom)
./integration/realm-querying-test.gts:           61.1   (56.7  + 4.4 phantom)
./integration/components/realm-config-routing-rule-edit-test.gts: 4.4  (phantom)
./integration/realm-server-subscriber-test.ts:    4.4   (phantom)
```
A sweep of the current suite shows **10 of 343** top-level module titles collide this way — the `Integration | realm*`, `Acceptance | workspace-chooser*`, `Unit | loader*`, `Integration | serialization*`, `Unit | query*`, and `code-ref*` families.

**Why it's non-blocking.** Coverage stays 100% (the time is attributed, just to the wrong siblings), every file still runs on exactly one shard, and the predicted slowest shard is floored at 614.5s by `interact-submode-test.gts` alone — so these ±4–6s errors don't move it today. The cost is that the committed weights are misleading (`realm-test` looks ~5× lighter than it is), and the error would matter more if the giant files are ever split.

**The fix (verified: resolves all 10 collisions, zero titles newly unmatched).** Add an exact-quoted tier ahead of the prefix tier; keep the prefix tier as the fallback that still catches static-prefix dynamic titles:

```suggestion
  // Prefer an exact quoted/backticked occurrence of the full title; then a
  // bare opening-delimiter match (which catches titles that are a static
  // prefix of a longer, partially dynamic module() argument); then a plain
  // substring. The exact tier stops "Integration | realm" from also matching
  // "Integration | realm indexing" and splitting realm-test's time onto it.
  const exact = testFiles.filter(({ source }) =>
    ["'", '"', '`'].some((q) => source.includes(`${q}${moduleName}${q}`)),
  );
  const prefix = exact.length
    ? exact
    : testFiles.filter(({ source }) =>
        ["'", '"', '`'].some((q) => source.includes(`${q}${moduleName}`)),
      );
  const candidates = prefix.length
    ? prefix
    : testFiles.filter(({ source }) => source.includes(moduleName));
```

---

.github/workflows/ci-host.yaml
```diff
@@ -929,7 +934,17 @@ jobs:
...
-      - name: Update memory baseline
+      # The merged junit report feeds the shard-timing update. Best-effort:
+      # if the merge job failed to produce it, the memory baseline still
+      # updates and the timings file is simply left alone this run.
+      - name: Download merged junit report
+        continue-on-error: true
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          path: merged-junit-report
+          name: host-test-report-merged
```
habdelra, line 945: [Claude Code 🤖]

**Confirmation (nothing to change): the degradation wiring correctly preserves the "update baselines even when a shard flaked" property this job is built around.**

Worth spelling out because it hinges on a non-obvious Actions rule. This job now `needs: [host-test, host-merge-reports-and-publish]` (was just `host-test`), but its `if` still contains `!cancelled()`. Because the `if` uses a status-check function, the default "run only if every `needs` succeeded" gate is lifted — so the job runs on exactly the same condition as before (`host-test` == success or failure), even when the merge job itself failed or was skipped. The added `host-merge-reports-and-publish` dependency therefore only sequences the merge before this job (a few minutes of extra latency vs. the old parallel path), it doesn't add a new skip path.

If the merged report is missing anyway, the two guards here — `continue-on-error: true` on this download step, and the `[ -f merged-junit-report/host.xml ]` check before invoking the generator — degrade to "leave shard timings unchanged" while `update-memory-baseline.mjs` still commits. So a merge-job failure can't freeze the memory baseline, which is the whole reason that baseline refreshes on `failure()` too.

One follow-on interaction to be aware of, flagged in the review body rather than here: the generator's coverage-floor `exit 1` runs under the step's implicit `set -e`, so a future attribution regression fails this whole job *before* the commit — stopping the memory-baseline update too, not just the timings.

---

habdelra (commented):
[Claude Code 🤖]

This review focused on three things: whether the custom duration-weighted sharding is deterministic and correct across all 20 browsers, whether swapping `ember exam --split/--partition` for plain `ember test --query` silently breaks the sibling junit and memory-baseline machinery that keys on module names, and whether the new main-only CI wiring degrades safely.

**Bottom line: solid, unusually well-documented change with no blocking issues.** The bin-packing is deterministic and near-optimal, the CI degradation paths are correct, and the command switch is safe for the systems that depend on it. One non-blocking improvement — the module→file attribution splits time across ~10 modules whose titles are quoted prefixes of others (inline on `generate-test-module-timings.mjs`, with a verified one-block fix).

**What lands right (mechanically, not as courtesy):**
- *The packing is near-optimal and correctly floored.* Against the committed weights, the predicted slowest shard is 614.5s vs. an ideal 517s (total 10,340s / 20), and that 614.5s is exactly `interact-submode-test.gts` sitting alone — a single file the packer can't split, not a balance defect. Fastest shard lands at 511s, ~1% under ideal.
- *Determinism is real.* Every browser computes all 20 buckets from the same bundled `test-module-timings.json` and the same `import.meta.glob` map (both baked once in `test-web-assets`), then keeps only its own bucket. The sort key (weight desc, path asc) is total, so input order can't perturb it — `shard-modules-test.ts` pins exactly this (partition-exactness, order-independence, unknown-weight-runs-once, range guard).
- *The exam→`ember test --query` switch is safe for the two module-name consumers.* The junit filename comes from `HOST_TEST_PARTITION` in `testem.js`, not from ember-exam; the memory baseline keys on the QUnit `details.name` in `setup-qunit.js`. Both are independent of the runner, so the merged report and the memory reports are unaffected. Dropping `--preserve-test-name` is consistent — ember-exam's name-mangling went away with the `--split/--partition` flags, so names come out clean either way.

1. Tighten attribution with an exact-quoted tier — see the thread on `generate-test-module-timings.mjs`. Verified to resolve all 10 current collisions with zero titles newly unmatched.
2. *Be aware:* the coverage-floor `exit 1` runs under the baseline-update step's implicit `set -e`, so a future drop below 95% attribution (e.g. many modules gaining fully-dynamic titles) fails the whole "Update Host Baselines" job before the commit — which also stops the *memory* baseline from updating, not just the timings. That's the intended "fail loudly," but the coupling is worth remembering; running the memory commit before the timings regen would decouple them if it ever bites.
3. The drift-gate `pack` / `slowestShardSeconds` in the `.mjs` re-implement `selectShardModules` from the `.ts` helper. The comment correctly bounds the risk (divergence changes only rewrite frequency, never assignment), but the two can drift silently — a small parity test would catch it cheaply.

**Adjacent, out of scope:** the slowest-shard floor is now a single file — `interact-submode-test.gts` (614.5s), with `ai-assistant-test.gts` (518s) and `searchable-search-doc-test.gts` (516s) close behind. The packer has essentially finished its job (fastest 511s ≈ ideal 517s); the only remaining lever to make the matrix wall-clock faster is splitting those three giants into multiple module files. Not this PR.

---

habdelra (approved):

packages/host/scripts/generate-test-module-timings.mjs
backspace: e0ad115f6b has a fix for this, thanks

PR #5652, opened 2026-07-30
Add support for redirection in host routing rules

backspace:
This adds redirection to the routing map:

s 2026-07-31 at 14 36 43@2x

In raw JSON:

```
      "hostRoutingRules": [
        { "path": "/terms" },
        { "path": "/tos", "redirectTo": "/terms" },
        { "path": "/blog", "redirectTo": "https://boxel.ai/", "statusCode": 301 }
      ]
    },
 ```

Here’s the 301:

s 2026-07-31 at 14 40 14@2x

and 302:

s 2026-07-31 at 14 39 27@2x

### DIFF
--- a/packages/base/realm-config.gts
+++ b/packages/base/realm-config.gts
@@ -10,25 +10,34 @@ import {
   realmURL,
 } from './card-api';
 import BooleanField from './boolean';
+import NumberField from './number';
 import StringField from './string';
 import CardInfoTemplates from './default-templates/card-info';
 import {
   cardDefComputedFields,
+  DEFAULT_REDIRECT_STATUS,
   findDuplicateRoutingPaths,
   getField,
   getFieldIcon,
+  REDIRECT_STATUS_CODES,
+  validateRedirectTarget,
   validateRoutingPath,
 } from '@cardstack/runtime-common';
 import {
+  BoxelInput,
   BoxelInputGroup,
+  BoxelSelect,
   FieldContainer,
   Header,
+  RadioInput,
 } from '@cardstack/boxel-ui/components';
 import { eq } from '@cardstack/boxel-ui/helpers';
 import FileSettingsIcon from '@cardstack/boxel-icons/file-settings';
 import LinkIcon from '@cardstack/boxel-icons/link';
+import { fn } from '@ember/helper';
 import { action } from '@ember/object';
 import type Owner from '@ember/owner';
+import { tracked } from '@glimmer/tracking';
 import { startCase } from 'lodash-es';
 import type { FieldsTypeFor } from './card-api';
 
@@ -36,7 +45,12 @@ class RoutingRuleAtom extends Component<typeof RoutingRuleField> {
   <template>
     <span class='routing-rule-atom'>
       <span class='path'>{{if @model.path @model.path '(no path)'}}</span>
-      {{#if @model.instance}}
+      {{#if @model.redirectTo}}
+        <span class='arrow' aria-hidden='true'>→</span>
+        <span class='redirect-target' data-test-redirect-target>
+          {{@model.redirectTo}}
+        </span>
+      {{else if @model.instance}}
         <span class='arrow' aria-hidden='true'>→</span>
         <@fields.instance @format='atom' />
       {{/if}}
@@ -47,7 +61,8 @@ class RoutingRuleAtom extends Component<typeof RoutingRuleField> {
         align-items: center;
         gap: var(--boxel-sp-xxs);
       }
-      .path {
+      .path,
+      .redirect-target {
         font-family: var(--boxel-font-family-mono, monospace);
       }
       .arrow {
@@ -57,7 +72,25 @@ class RoutingRuleAtom extends Component<typeof RoutingRuleField> {
   </template>
 }
 
+let routingRuleKindGroupNumber = 0;
+
 class RoutingRuleEdit extends Component<typeof RoutingRuleField> {
+  // Which target editor is showing. Backed by the data: a rule whose
+  // `redirectTo` is non-null (empty string included — `setKind` seeds
+  // '' so the choice survives a reload) is a redirect rule. The tracked
+  // override exists only so the toggle responds instantly while the
+  // model write settles.
+  @tracked private kindOverride: 'card' | 'redirect' | null = null;
+
+  private kindItems: { id: 'card' | 'redirect'; text: string }[] = [
+    { id: 'card', text: 'Render a card' },
+    { id: 'redirect', text: 'Redirect' },
+  ];
+
+  private kindRadioGroup = `__routing_rule_kind${routingRuleKindGroupNumber++}__`;
+
+  private statusCodeOptions = [...REDIRECT_STATUS_CODES];
+
   constructor(owner: Owner, args: any) {
     super(owner, args);
     // The path input renders an empty input alongside a fixed `/`
@@ -104,6 +137,69 @@ class RoutingRuleEdit extends Component<typeof RoutingRuleField> {
     this.args.model.path = `/${trimmed}`;
   }
 
+  get kind(): 'card' | 'redirect' {
+    return (
+      this.kindOverride ??
+      (this.args.model.redirectTo != null ? 'redirect' : 'card')
+    );
+  }
+
+  get isRedirect(): boolean {
+    return this.kind === 'redirect';
+  }
+
+  // Switching kind clears the other kind's target so a rule is never
+  // ambiguous (the read path prefers `redirectTo` when both are set,
+  // but only a hand-edited realm.json can get into that state).
+  @action
+  setKind(kind: 'card' | 'redirect') {
+    this.kindOverride = kind;
+    if (kind === 'redirect') {
+      this.args.model.instance = undefined;
+      if (this.args.model.redirectTo == null) {
+        this.args.model.redirectTo = '';
+      }
+    } else {
+      this.args.model.redirectTo = undefined;
+      this.args.model.statusCode = undefined;
+    }
+  }
+
+  get redirectToValue(): string {
+    return this.args.model.redirectTo ?? '';
+  }
+
+  @action
+  setRedirectTo(value: string) {
+    this.args.model.redirectTo = value ?? '';
+  }
+
+  get redirectWarning(): string | undefined {
+    return validateRedirectTarget(this.args.model.redirectTo);
+  }
+
+  get selectedStatusCode(): number {
+    return this.args.model.statusCode ?? DEFAULT_REDIRECT_STATUS;
+  }
+
+  @action
+  setStatusCode(code: number) {
+    this.args.model.statusCode = code;
+  }
+
+  @action
+  statusCodeLabel(code: number): string {
+    switch (code) {
+      case 301:
+        return '301 · permanent';
+      case 308:
+        return '308 · permanent, method-preserving';
+      case 302:
+      default:
+        return '302 · temporary';
+    }
+  }
+
   // The chooser is locked to the consuming realm; pass it through
   // explicitly rather than letting LinksToEditor read it from
   // `RealmURLContext`. The context is only provided by the operator-mode
@@ -119,6 +215,21 @@ class RoutingRuleEdit extends Component<typeof RoutingRuleField> {
 
   <template>
     <div class='routing-rule-edit' data-test-routing-rule-edit>
+      <div class='kind-toggle' data-test-routing-rule-kind>
+        <RadioInput
+          @items={{this.kindItems}}
+          @groupDescription='Routing rule target'
+          name='{{this.kindRadioGroup}}'
+          @checkedId={{this.kind}}
+          @spacing='compact'
+          @hideBorder={{true}}
+          as |item|
+        >
+          <item.component @onChange={{fn this.setKind item.data.id}}>
+            {{item.data.text}}
+          </item.component>
+        </RadioInput>
+      </div>
       <div class='row'>
         <div class='path-cell'>
           <BoxelInputGroup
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: packages/host/app/services/host-mode-service.ts [hunk 1]
```diff
@@ -143,13 +154,24 @@ export default class HostModeService extends Service {
...
+
+  // SPA-side counterpart of the server's HTTP redirect for a redirect
+  // routing rule: a full-page navigation to matched paths gets the 3xx
+  // from serve-index, but an in-app transition never leaves the SPA, so
+  // the index route calls this instead. `replace` mirrors an HTTP
+  // redirect, which leaves no history entry for the redirecting URL.
+  // Goes through ember-window-mock's `window` so tests can intercept the
+  // navigation.
+  redirectTo(target: string) {
+    window.location.replace(target);
```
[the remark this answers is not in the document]
backspace: fixed in b898a63e8b

FadhlanR (approved):
[Claude Code 🤖] LGTM — clean layering across the shared vocabulary, map builder, server redirect, and SPA path, and the comments explaining the non-obvious bits (redirect-wins-over-canonicalization, protocol-relative steering, the transition-vs-location.search choice) are excellent.

One gap worth addressing, ideally before merge:

**No test coverage for the SPA-side redirect path.** The server HTTP redirect, the validators, and the editor UI are all well covered, but the in-app redirect — `routes/index.gts` (`resolveRoutedPath` → the `'redirectTo' in routed` branch) → `HostModeService#redirectTo` — has no test exercising it. Nothing in `host-mode-test.gts` / `host-submode-test.gts` navigates in-app to a redirect-ruled path.

This is the subtlest new code in the PR: `redirectTo` deliberately takes `transition.to.queryParams` rather than `window.location.search` to dodge a HistoryLocation staleness bug (per the comment on the method). That kind of subtlety regresses silently. An acceptance test that transitions in-app to a redirect-ruled path and asserts the resulting location would lock it in.

While adding it, worth confirming server↔SPA parity on query carry-over: the server copies the raw `requestURL.search` verbatim, whereas the SPA re-serializes `transition.to.queryParams` via `String(value)` with a `value != null` guard — different data shapes over different code paths, so it's worth verifying they produce the same Location for the same inbound request.

packages/realm-server/handlers/serve-index.ts
```diff
@@ -387,13 +392,41 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
...
+          // second hop for no benefit. A realm-relative target resolves
+          // against the realm's mount pathname, mirroring how the
+          // rule's own `path` is mounted; extra leading slashes are
+          // collapsed so a target can never be read as protocol-
+          // relative. The request's query string carries over unless
+          // the target declares its own.
+          let { redirectTo, statusCode } = matched.rule;
+          let target: URL;
+          if (/^https?:/i.test(redirectTo)) {
+            target = new URL(redirectTo);
```
burieberry: **Redirect loops aren't guarded anywhere.** A rule like `{path: '/tos', redirectTo: '/tos'}` (or a two-rule cycle `/a→/b`, `/b→/a`) passes `validateRedirectTarget` — it's shape-only and never sees the rule's own path — survives `Realm.getHostRoutingMap`, shows no editor warning, and this branch answers `Location: <same URL>` unconditionally → `ERR_TOO_MANY_REDIRECTS`. With `statusCode: 301` the loop is cached by browsers, so fixing the rule doesn't recover already-poisoned clients. The SPA loops the same way via `window.location.replace`.

A same-canonical-path check in `validateRedirectTarget`/`getHostRoutingMap` would catch the direct case cheaply.

🤖 Reviewed by Claude

packages/host/app/services/host-mode-service.ts
```diff
@@ -143,13 +154,39 @@ export default class HostModeService extends Service {
...
+  // navigation.
+  //
+  // `queryParams` is the transition target's query (from
+  // `transition.to.queryParams`) — NOT `window.location.search`, which
+  // with HistoryLocation still shows the URL being navigated away from
+  // while the transition is in flight. Matching serve-index's semantics,
+  // it carries over only when the redirect target declares no query of
+  // its own.
+  redirectTo(target: string, queryParams?: Record<string, unknown>) {
+    let url = new URL(target, this.hostModeOrigin ?? window.location.origin);
```
burieberry: **The redirect base is visitor-controllable.** `hostModeOrigin` returns the raw `?hostModeOrigin` query param whenever it's present (`simulatingHostMode` is `has()`-based), which makes this a top-level navigation sink on visitor input:

- `?hostModeOrigin=` (empty) yields `''`, which is non-nullish, so the `?? window.location.origin` fallback never fires — that fallback is unreachable dead code — and `new URL(target, '')` throws a TypeError inside the index route's model hook.
- `?hostModeOrigin=https://evil.example` is sticky (declared index query param), so an in-app transition to a redirect-rule path resolves the realm-relative target against the attacker's base and `location.replace()` navigates there.

Realm-relative targets only need `window.location.origin` as the base.

🤖 Reviewed by Claude

packages/host/app/services/host-mode-service.ts
```diff
@@ -143,13 +154,39 @@ export default class HostModeService extends Service {
...
+    if (!url.search && queryParams) {
+      for (let [key, value] of Object.entries(queryParams)) {
+        if (value != null) {
+          url.searchParams.set(key, String(value));
+        }
```
burieberry, line 200: **Hardening:** this forwards every transition query param onto the redirect target, including external hosts — `hostModeStack`, `operatorModeState`, `sid`, `clientSecret` are all declared on the index route. Today the sensitive ones only appear on the host-app origin where this branch is inactive, and the server's 3xx does the same verbatim carry-over, so no demonstrated leak — but an allowlist (or stripping host-app-internal params) on both sides would prevent a future secret-bearing param from riding onto a third party. Also note repeated keys diverge from the server: server emits `?tag=a&tag=b`, this rebuilds `?tag=a%2Cb`.

🤖 Reviewed by Claude

packages/base/realm-config.gts
```diff
@@ -57,7 +72,25 @@ class RoutingRuleAtom extends Component<typeof RoutingRuleField> {
...
 
+let routingRuleKindGroupNumber = 0;
+
 class RoutingRuleEdit extends Component<typeof RoutingRuleField> {
+  // Which target editor is showing. Backed by the data: a rule whose
+  // `redirectTo` is non-null (empty string included — `setKind` seeds
+  // '' so the choice survives a reload) is a redirect rule. The tracked
+  // override exists only so the toggle responds instantly while the
+  // model write settles.
+  @tracked private kindOverride: 'card' | 'redirect' | null = null;
```
burieberry: **`kindOverride` is set in `setKind` and never cleared, so it permanently shadows the model.** With the same RealmConfig card open in two stacks: stack A clicks "Redirect" (`kindOverride='redirect'`), stack B switches the rule back to a card rule with an instance — stack A's `kind` getter short-circuits on the stale override forever, keeps rendering the empty redirect editor, and one keystroke there sets `model.redirectTo`, which the read path prefers, silently clobbering stack B's instance choice.

The justifying comment ("while the model write settles") doesn't hold: `setField` → `notifyCardTracking` is synchronous, so contains-field writes are immediately readable. Deleting `kindOverride` and deriving `kind` purely from the model loses nothing.

🤖 Reviewed by Claude

packages/base/realm-config.gts
```diff
@@ -119,6 +207,21 @@ class RoutingRuleEdit extends Component<typeof RoutingRuleField> {
 
   <template>
     <div class='routing-rule-edit' data-test-routing-rule-edit>
+      <div class='kind-toggle' data-test-routing-rule-kind>
+        <RadioInput
+          @items={{this.kindItems}}
+          @groupDescription='Routing rule target'
+          name='{{this.kindRadioGroup}}'
```
burieberry: **The radio group name never reaches `RadioInput`.** `name=` here is an HTML attribute that lands on the `<fieldset>` via `...attributes`; `RadioInput` reads `this.args.name ?? guidFor(this)` (boxel-ui `radio-input/index.gts:51-53`), so grouping only works via the guid fallback and the module-level `routingRuleKindGroupNumber` counter + `kindRadioGroup` field are dead code. Either `@name={{this.kindRadioGroup}}`, or simpler: delete the counter/field/attribute and rely on the guid default, which is what actually happens today.

🤖 Reviewed by Claude

packages/runtime-common/realm.ts
```diff
@@ -6865,6 +6873,29 @@ export class Realm {
...
         let normalizedPath = normalizeRoutingPath(path);
+        let redirectTo = (rule as Record<string, unknown>).redirectTo;
+        if (typeof redirectTo === 'string' && redirectTo.trim()) {
+          // A redirect rule. When a rule carries both `redirectTo` and
+          // `instance` (only possible by hand-editing realm.json — the
+          // editor clears one when the other is chosen), the redirect
+          // wins: it is the more explicit declaration, and silently
+          // dropping the rule would dead-end the path.
+          let target = redirectTo.trim();
+          let warning = validateRedirectTarget(target);
```
burieberry, line 6892: **A rule with a valid `instance` and a malformed `redirectTo` is dropped entirely instead of falling back to serving the card** — the exact dead-end the adjacent comment says this ordering avoids. Hand-edit realm.json to add `redirectTo: 'terms'` (missing leading slash) to a working serve rule `{path: '/pricing', instance: ./pricing-page}`: the redirect branch runs first, `validateRedirectTarget` warns, and `return []` drops the whole rule — `/pricing`, which rendered a card before the edit, now 404s with only a server log. Pre-PR a stray `redirectTo` key was ignored and the instance served. Falling back to the valid instance when `redirectTo` fails validation would honor the comment's own rationale.

🤖 Reviewed by Claude

packages/realm-server/handlers/serve-index.ts
```diff
@@ -387,13 +392,41 @@ export function createServeIndex(deps: ServeIndexDeps): ServeIndexHandlers {
...
+          // matched path — canonicalizing first would cost the client a
+          // second hop for no benefit. A realm-relative target resolves
+          // against the realm's mount pathname, mirroring how the
+          // rule's own `path` is mounted; extra leading slashes are
+          // collapsed so a target can never be read as protocol-
+          // relative. The request's query string carries over unless
+          // the target declares its own.
+          let { redirectTo, statusCode } = matched.rule;
+          let target: URL;
+          if (/^https?:/i.test(redirectTo)) {
```
burieberry: **Redirect-target resolution is triplicated.** The `/^https?:/i` test + `realmPathname` prefix + leading-slash collapse here is duplicated in the `hostScopedMap` builder (~L557-560), with a third variant in the SPA — and the security-relevant slash handling already differs within one screen of code (serve-rule paths use `/^\//`, redirect targets `/^\/+/`). Meanwhile `host-mode-service.ts:42` re-declares the rule union as a shadow `HostScopedRoutingRule` type with `statusCode` widened to `number`, and `index.gts:101` discriminates with a raw `'redirectTo' in routed` instead of the exported `isRedirectRoutingRule` guard this file already uses. Editing one copy silently desyncs the server's `Location` header from the SPA's injected map. One `resolveRedirectTarget(redirectTo, realmPathname)` helper in `host-routing-validation.ts` plus typing the map as `HostRoutingRule[]` collapses all of it.

🤖 Reviewed by Claude

burieberry (commented):
Reviewed the redirect-rule support end to end (server handler, SPA path, realm map building, and the config editor). Seven findings below as inline comments. One more lands in a file this PR doesn't touch, so noting it here:

**`packages/host/app/components/operator-mode/publish-realm-modal.gts` (~L562) — pre-publish dangling-rule check not updated for redirect rules.** Two gaps: (1) a rule carrying both `redirectTo` and a dangling `instance` link — the exact state `realm.ts` documents as "redirect wins" — still has its instance resolved by `checkDanglingRoutingRulesTask`, 404s, and triggers a spurious "degrades to 404" warning even though the published path redirects fine; (2) a pure redirect rule is skipped entirely (`if (!reference) continue;`), so a realm-relative target pointing at nothing publishes silently and the path 302s into a 404 — the failure class this check exists to prevent.

Two other candidates were investigated and ruled out, for the record: `afterModel` does not wipe state after the SPA redirect (`restore()` is a no-op once initialized), and a cached ETag 304 cannot answer before the redirect branch (the ETag key bumps on `last_published_at`, which any rule change necessarily changes via republish).

🤖 Reviewed by Claude (Fable 5)

packages/realm-server/handlers/serve-index.ts
backspace: 6e1d9af180 added a loop guard

packages/host/app/services/host-mode-service.ts
backspace: this is addressed in 08597ca8d4

packages/host/app/services/host-mode-service.ts
backspace: fe77e5a0f4 adds this hardening

packages/base/realm-config.gts
backspace: fixed in f70fd53020

burieberry (approved):

packages/base/realm-config.gts
backspace: fixed in 3271555abf

packages/runtime-common/realm.ts
backspace: 750b4350f8 has a fallback

packages/realm-server/handlers/serve-index.ts
backspace: 2d4d8643dc changes to use a shared helper

backspace:
> One gap worth addressing, ideally before merge:
> 
> **No test coverage for the SPA-side redirect path.**

thanks, 13ac5ddf3f has this

PR #5653, opened 2026-07-30
Try bundling base realm

backspace:


backspace:
this was an experiment, now we’ll be implementing it in pieces

PR #5658, opened 2026-07-31
Add StlDef and ThreeMfDef FileDef subclasses for 3D model files

FadhlanR:
## What

Adds `.stl` and `.3mf` support as base-realm `FileDef` subclasses, built on the shared file-format shells' `model` family.

- **`ThreeDModelDef`** — thin family base under `FileDef`: `static previewComponent = Model3DPreview` (a live WebGL orbit viewer, loaded from a CDN at client render time) and the `model3d` metadata field the isolated shell renders. Inherits all four shared format shells.
- **`Model3dMetadataField`** — the 3D family's metadata field in `file-formats/metadata-fields.gts`. The shells were already wired to render `<@fields.model3d />` and gate the fitted hero on `model3d.triangles`; this fills in the field they expected.
- **`StlDef` / `ThreeMfDef`** — leaves that add only their header-only `extractAttributes`, projecting each format's facts onto `model3d` via `model3dAttributes`. STL uses a bounded ASCII/binary sniffer; 3MF unzips the OPC package (`fflate`) and reads the model prologue + slicer config.
- Registers `.stl`/`.3mf` in the FileDef code-ref map; threads the realm file-size limit through the index-time extractor; adds `fflate`.

The `.stl`/`.3mf` fixtures already in `filedef-fixtures` now resolve to these subclasses instead of the generic `FileDef`.

> Rebased onto current `main` and reworked onto the shared file-format shells that landed while this PR was open (it previously rolled its own four format templates + per-format inspector fields).

## Design notes

- **One `model3d` field (not per-format fields).** The isolated shell renders exactly one 3D slot and no family overrides `isolated`, so both formats project onto a single `Model3dMetadataField` — the shell-native shape.
- **Header-only extraction.** True geometry (transform-correct dimensions, full triangle counts) comes from the live viewer, not the index. A value is present only when it sits cheaply in the header: a binary STL's facet count, a slicer 3MF's configured face/part counts, materials, provenance.
- **Known limitation:** the isolated `3D model` section gates on `meshes || triangles`. An STL reports `meshes: 1` (a solid) so it always shows; a config-less 3MF reports no geometry counts, so its section hides (generic name/type/size rows still show).

## Test plan

- `ember-tsc` clean; prettier clean; template-lint clean.
- Unit tests for the pure parsers (`model-meta-extractor-test.ts`) and an end-to-end file-extract acceptance test (`model-file-extract-test.gts`) asserting the `model3d` projection.
- Pending on a dev stack: index the `filedef-fixtures` corpus and confirm the stl / three-mf FormatPreview pages render with zero error rows.


packages/base/model-file-def.gts
```diff
@@ -0,0 +1,852 @@
...
+  Args: { model: ModelDef };
+}> {
+  <template><ModelIsolatedBody @model={{@model}} /></template>
+}
+
+// Base FileDef subclass for 3D model formats. Owns the shared scene metadata,
+// the thumbnail seam, and the four format templates. Leaves (StlDef,
+// ThreeMfDef) add their format-specific metadata field + extraction and may
+// override `isolated` to append that metadata below the shared body.
+export class ModelDef extends FileDef {
```
lukemelia: Let's name this `3DModelDef` (and rename file accordingly) -- ModelDef sounds too much like some kind of base class.

lukemelia:
What do we lose if we only read the headers of these files and not the whole thing? Parsing 5MB to index should be avoided if possible.

packages/base/stl-model-def.gts
```diff
@@ -0,0 +1,217 @@
...
+    url: string,
+    getStream: () => Promise<ByteStream>,
+    options: {
+      contentHash?: string;
+      contentSize?: number;
+      // Backstop bound on the bytes we're willing to parse at index time,
+      // threaded from the host; defaults to the realm's standard file-size
+      // limit (`DEFAULT_FILE_SIZE_LIMIT_BYTES`), the same ceiling the write
+      // path enforces, so the two can't drift.
+      fileSizeLimitBytes?: number;
```
habdelra: [Claude Code 🤖] **Confirming the automated P2 flag on `fileSizeLimitBytes` (the codex thread on `three-mf-def.gts:286`), and one step further: the seam is never populated by any caller, and this comment's "can't drift" claim is currently false.** Blocking-wise: non-blocking (default deployments are unaffected), but worth resolving since the PR description advertises this as a threaded seam. Regression class: introduced by this PR.

**What I traced.** The only production caller of a leaf `extractAttributes` is `FileDefAttributesExtractor` in `packages/host/app/utils/file-def-attributes-extractor.ts`. Its `tryExtract` calls `klass.extractAttributes(url, stream, { contentHash, contentSize, toolContext })` — no `fileSizeLimitBytes`. The `FileDefExport` type in that same file doesn't declare `fileSizeLimitBytes` either, so TypeScript never flags the omission. So `options.fileSizeLimitBytes` here is always `undefined` and `sizeCap` is always `DEFAULT_FILE_SIZE_LIMIT_BYTES`, regardless of realm configuration.

**Why the comment is wrong.** The write path's ceiling is *not* fixed at the default: `realm.ts` and `server.ts` both read `Number(process.env.FILE_SIZE_LIMIT_BYTES ?? DEFAULT_FILE_SIZE_LIMIT_BYTES)`. A deployment that sets `FILE_SIZE_LIMIT_BYTES` above 5 MB accepts STL/3MF files up to that limit on write, but this extractor still caps the parse at the hard-coded 5 MB — so files between 5 MB and the configured limit are stamped with the 3D type but silently get no `model3d`/metadata. That's exactly the drift the comment says can't happen.

**The fix is cheap because the host already has the value.** `serve-index.ts` injects the server's `fileSizeLimitBytes` into host config, and `environment-service.ts` exposes it as `environmentService.fileSizeLimitBytes`. Thread that through `runFileExtract` → `FileDefAttributesExtractor` → the `tryExtract` options (and add the field to `FileDefExport`), or — if you'd rather not wire it now — drop the `fileSizeLimitBytes` parameter and reword this comment to say the cap is a fixed 5 MB backstop. Either is fine; the in-between state documents behavior that doesn't exist.

The identical comment + read applies verbatim in `three-mf-def.gts`.

---

packages/base/three-mf-meta-extractor.ts
```diff
@@ -0,0 +1,238 @@
...
+export function parseThreeMf(
+  buf: ArrayBuffer,
+): { model3d: Model3dData; threeMfMetadata: ThreeMfMetadata } | undefined {
+  let files: Record<string, Uint8Array>;
+  try {
+    files = unzipSync(new Uint8Array(buf)) as Record<string, Uint8Array>;
+  } catch {
+    // Not a valid ZIP / OPC package.
+    return undefined;
+  }
```
habdelra, line 103: [Claude Code 🤖] **Confirming the automated P1 (bound decompressed 3MF before unzipping).** `unzipSync` here eagerly decompresses *every* entry in the archive into memory, and the only size gate upstream is on the *compressed* bytes (the write limit, default 5 MB). A ~5 MB archive of highly compressible data can expand to hundreds of MB–GBs; since this runs on the indexing/prerender path, an OOM there stalls indexing for the whole worker, not just this file.

**Scope/severity.** Realm files come from authenticated realm members, not anonymous public uploads, so this is hardening rather than an open DoS — non-blocking, follow-up. But it's a real gap and the fix is small.

**Concrete mitigation with the library you already use.** `fflate`'s `unzipSync(data, opts)` takes `filter: (file: UnzipFileInfo) => boolean`, which runs *before* each entry is decompressed and receives `{ name, originalSize, compression, ... }`. Two cheap wins: (1) you only ever read `*.model` and `model_settings.config`, so filter to just those entries — everything else is skipped, not decompressed; (2) reject any entry whose declared `originalSize` exceeds a sane per-entry ceiling. Declared sizes can be forged, so for a hard guarantee you'd stream (`unzip`) with an output-byte counter, but the filter alone removes the obvious bomb and bounds the work to the entries you actually parse.

---

packages/base/three-mf-meta-extractor.ts
```diff
@@ -0,0 +1,238 @@
...
+  for (let vertex of vertices) {
+    for (let axis = 0; axis < 3; axis++) {
+      mins[axis] = Math.min(mins[axis], vertex[axis]);
+      maxs[axis] = Math.max(maxs[axis], vertex[axis]);
+    }
+  }
+  let dimension = (axis: number) =>
+    vertices.length
+      ? Math.round((maxs[axis] - mins[axis]) * 1_000_000) / 1_000_000
+      : undefined;
```
habdelra: [Claude Code 🤖] **Confirming the automated P2 (bounds ignore 3MF transforms).** `dimension()` is computed from the min/max of *every* `<vertex>` aggregated across all objects/model parts in raw resource coordinates — `<build><item transform>` and `<component transform>` are never applied. For a single-object identity-transform model (the common slicer export, and what the tests cover) this is correct, but for a multi-instance or composed build it's wrong: the classic case is two instances of a 10-unit object placed 100 units apart, which still reports `sizeX = 10` instead of ~110. It also unions unrelated objects' local coordinate systems overlaid at the origin.

**What it affects / doesn't.** Only the *indexed* `sizeX/Y/Z` and the silhouette proportions (which read `extents`). The live WebGL viewer is unaffected — `3MFLoader` applies the build/component transforms itself, so the on-screen model is correct; it's just the extracted numbers that are approximate.

Follow-up, not this PR — a correct implementation means walking the build graph with accumulated transforms, which is real work for a metadata nicety. Non-blocking; I'd just avoid treating `sizeX/Y/Z` as authoritative assembled dimensions until then.

---

packages/base/three-mf-meta-extractor.ts
```diff
@@ -0,0 +1,238 @@
...
+      'application/xml',
+    );
+    if (document.getElementsByTagName('parsererror').length) {
+      // Malformed model XML — treat the whole package as unparseable.
+      return undefined;
+    }
+    modelDocuments.push({ path, document, root: document.documentElement });
+  }
+  let primary =
+    modelDocuments.find(({ path }) => path === modelPart) ?? modelDocuments[0];
```
habdelra: [Claude Code 🤖] **Confirming the automated P2 (accepts any `.model` XML).** The only gates to "this is a 3MF" are: it's a ZIP, it has an entry matching `/\.model$/i`, and that entry is well-formed XML (no `<parsererror>`). A ZIP containing `<document/>` in a file named `x.model` passes all three — `parseThreeMf` then returns an all-zero/empty metadata object (not `undefined`), so the file is stamped `ThreeMfDef` and handed to a viewer that can't parse it.

**Fix.** Require the 3MF root before accepting: check `primary.root.localName === 'model'` (and ideally the core namespace `http://schemas.microsoft.com/3dmanufacturing/core/...`), returning `undefined` otherwise so it falls back to base `FileDef` like the other unparseable cases.

**Test gap.** `model-meta-extractor-test.ts` covers malformed XML, missing model part, and non-ZIP, but not "well-formed XML that isn't a 3MF `<model>` root" — that's the negative-space test that would pin this. Worth adding alongside the fix. Minor / non-blocking.

---

packages/base/model-file-def.gts
```diff
@@ -0,0 +1,852 @@
...
+      io = new IntersectionObserver((entries) => {
+        if (entries.some((entry) => entry.isIntersecting)) {
+          io?.disconnect();
+          start();
+        }
+      });
+      io.observe(element);
+    } else {
+      start();
+    }
```
habdelra: [Claude Code 🤖] **Confirming the automated P2 (lazy viewers never released), with one conditional.** The `IntersectionObserver` calls `io.disconnect()` on first intersection (line 371) and boots the viewer; the WebGL context + RAF loop it creates are only torn down in the modifier's destructor — i.e. on component unmount. So the number of live WebGL contexts tracks "tiles ever scrolled past," not "tiles currently visible."

**Whether this bites depends on the container.** If the enclosing collection virtualizes (unmounts off-screen tiles), the destructor runs on scroll-away and this is fine. If it keeps all tiles mounted, contexts grow monotonically and will hit the browser's ~16-context ceiling, after which new viewers fail and later previews break (WebGL context-lost / silhouette fallback). Worth confirming which case the fitted grid is in.

**Fix if needed.** Keep the lazy boot, but don't `disconnect()` permanently — keep the observer and react to *both* `isIntersecting` transitions, disposing/suspending the renderer on exit and re-booting on re-entry. Non-blocking; follow-up unless the target grid is non-virtualized.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I focused on the index-time extraction path — where the parsers actually run and where the feature's cost/correctness live — and on how the new size-cap machinery wires into the existing `FileDefAttributesExtractor`, since that seam isn't visible from the diff alone. I checked out the head branch and traced every extraction call from `runFileExtract` down through both leaves; the template/viewer layer I reviewed more lightly.

**Bottom line.** Clean, well-tested feature — the pure `.ts` parsers, the card-api-free silhouette, the unit + real-route acceptance tests, and the mismatch→base-`FileDef` fallback are all nicely done, and CI is green. **No blocking bug for a default deployment.** The one thing worth fixing before merge-and-forget is documentation-level: the `fileSizeLimitBytes` "threaded seam" is never populated by its only caller, and its "can't drift" comment is false once `FILE_SIZE_LIMIT_BYTES` is raised. The five automated P1/P2 flags are all reachable — I verified each and added mechanism + fix in-thread; none block, but the ZIP-bomb and lazy-viewer ones deserve a follow-up.

**What lands right.**
- Extracting the parsers to pure `.ts` modules (mirroring `png-meta-extractor.ts`) makes them directly unit-testable and keeps card-api out of the test bundle — and the `model-silhouette.ts` split correctly resolved the rolldown `MISSING_EXPORT` cascade.
- The STL single-pass streaming parse with per-facet degeneracy counting is the right shape, and the degenerate-facet test pins exactly the misalignment the old grouping had.
- Over-cap behavior keeps the 3D type + live viewer and only drops the inspector metadata (rather than throwing and demoting to a plain file) — good instinct, well-commented.

**On the open (automated) threads.** All five are real and reachable; verification and recommended fixes are in each thread. Priority order: (1) is the one I'd act on in this PR (cheap, and the comment is currently misleading); (2) and (5) are the hardening ones; (3) and (4) are metadata-accuracy nits.

**Recommendations** (all non-blocking; none require re-architecture):
1. Either thread `environmentService.fileSizeLimitBytes` through to `extractAttributes`, or drop the parameter and reword the comment — thread on `stl-model-def.gts`.
2. Bound 3MF decompression via fflate's `filter`/`originalSize` and restrict to the entries you parse — thread on `three-mf-meta-extractor.ts:58`.
3. Treat STL/3MF `sizeX/Y/Z` as un-transformed resource-space extents, not assembled dimensions (viewer is correct) — thread on `three-mf-meta-extractor.ts:126`.
4. Require a `<model>` root before accepting a 3MF, and add the negative-space test — thread on `three-mf-meta-extractor.ts:77`.
5. Release/suspend fitted WebGL viewers on viewport exit if the grid doesn't virtualize — thread on `model-file-def.gts:373`.

**Adjacent, out of scope.** Binary-STL color detection treats any non-zero attribute-byte-count as "has color," which over-reports for STLs that use those bytes for non-color data — fine for a metadata flag, just noting it. And there are now three small extension-parsing helpers (`getExtension`, `ModelIsolatedBody.extension`, and `file-def-code-ref`'s `extensionOfName`); not worth consolidating across packages here.

This is a COMMENT review — approve / request-changes is the maintainers' call.

---

FadhlanR:
[Claude Code 🤖]

Good push on the header question — we reworked extraction and previews around it (739608c).

**On "avoid parsing 5MB to index":** the answer is asymmetric between the two formats, so we changed both.

- **STL is now header-only.** We read the binary header (facet count, `COLOR=` flag) or the ASCII prologue (solid name) and never scan the facet body. Neither format's header carries a bounding box, so dimensions can't come from a header read at all — they now come from the live client-side viewer, which reads the true bounds off the geometry it already loads (and is transform-correct, unlike the old index-time scan).
- **3MF is now a bounded prologue read.** The useful metadata sits at the top of the model part (root namespaces/unit, `<metadata>`, `<basematerials>`), before the geometry. So fflate's `filter` decompresses only the `.model` + `model_settings.config` entries (skipping embedded thumbnails/textures), and we regex-read the metadata prologue instead of DOM-parsing the vertex/triangle body. One caveat I'll be straight about: we still inflate the model entry's compressed stream (fast, and capped), so 3MF's win is "no geometry DOM-parse + skip unrelated entries," not a literal zero-read. A true zero-geometry-decompress would need entry-level streaming, which we can add if profiling shows the inflate itself matters.

**Previews:** fitted now shows a static cube icon + filename (mirroring the audio FileDef) instead of booting a live WebGL viewer per tile — the shaded-thumbnail seam (`thumbnailUrl`) stays for CS-12401. Embedded/isolated keep the live viewer and surface the client-side dimensions.

Nice side effect: this also closes the automated flags in one go — the ZIP-bomb (bounded decompression + oversized-entry rejection), the WebGL-context exhaustion (see the follow-up below), the transform-incorrect bounds (dimensions are client-side now), and the `<model>`-root validation (3MF now requires the core namespace before it's accepted). And `fileSizeLimitBytes` is now actually threaded from realm config, so the cap tracks the write limit instead of a hard-coded default.

Also renamed `ModelDef` → `ThreeDModelDef` per your other comment — `3DModelDef` can't be the literal class name (it's the `adoptsFrom` export identifier, and JS identifiers can't start with a digit), so I spelled out the leading digit to match the existing `ThreeMfDef` convention.

---

**Follow-up (1dcc629) — correcting the WebGL point above.** Two things: (a) `739608c` alone did *not* fully close the WebGL-context finding — fitted was static, but embedded still booted eagerly and never released its context, so an embedded strip could still leak. (b) The real fix generalizes the gate to **prerender vs live client**, not format:

- Server-side prerender never boots WebGL or the CDN import (`globalThis.__boxelRenderContext` guard) — deterministic, not fail-fallback.
- The live viewer now disposes its WebGL context when a tile scrolls off-screen and re-boots from cached bytes on re-entry, so **embedded strips are bounded too**, not just fitted.
- Fitted stays on the static thumbnail because a grid can exceed the browser's ~16-context cap simultaneously (lazy+dispose only bounds scrolling). A pooled/shared-context renderer for interactive fitted is Option B in CS-12401.

packages/base/model-file-def.gts
```diff
@@ -0,0 +1,852 @@
...
+    // In `lazy` mode (fitted collection tiles) only boot the WebGL engine once
+    // the tile is actually on-screen, so a grid of many models doesn't exhaust
+    // the browser's WebGL context budget. Off-screen tiles keep the silhouette.
+    let io: IntersectionObserver | undefined;
+    if (lazy) {
+      io = new IntersectionObserver((entries) => {
+        if (entries.some((entry) => entry.isIntersecting)) {
+          io?.disconnect();
+          start();
+        }
```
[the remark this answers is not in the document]
FadhlanR: [Claude Code 🤖] Resolved in 1dcc629 — and reframed: the fix isn't format-specific, it's **prerender vs live client**.

- **Live client:** the viewer's modifier now lazy-boots when the tile scrolls on-screen and **disposes the WebGL context** (`renderer.dispose()` + `forceContextLoss()`, canvas removed) when it scrolls off, re-booting from cached bytes on re-entry. So live contexts track the *currently visible* set, not "tiles ever viewed" — the monotonic growth is gone.
- **Prerender/indexing:** the modifier now no-ops entirely when `globalThis.__boxelRenderContext` is set, so server-side rendering never boots WebGL or fetches the CDN engine (deterministic, rather than relying on the CDN import failing).
- **Fitted specifically:** it no longer mounts the viewer at all. A grid can show more tiles than the browser's ~16-context cap at once, which lazy+dispose can't solve (that only bounds scrolling, not simultaneous visibility), so fitted uses the static thumbnail. A pooled/shared-context renderer for interactive fitted is tracked as Option B in CS-12401.

FadhlanR:
[Claude Code 🤖] posted on behalf of @FadhlanR

@lukemelia — rebased this onto `main` and reworked it to align with the FileDef shells + fixture corpus you landed while it was open. What changed:

- **Conforms to the `model` family you pre-wired** instead of inventing a new one — `.stl`/`.3mf` resolve through the existing `model` profile, and `ThreeDModelDef` now supplies just a `previewComponent` + fields, inheriting the four shared shells.
- **Filled in `Model3dMetadataField`** — the field your isolated shell was already wired to render (`<@fields.model3d />`, the `hasModel3d` gate, the `model3d.triangles` hero fact). `StlDef`/`ThreeMfDef` are now thin leaves that project their header-only facts onto that one field, the same way audio/video do with `encoding`.
- **Deleted the bespoke bits** — the four hand-rolled format templates + per-format inspector fields are gone; the live WebGL viewer moved into `Model3DPreview` (embedded/isolated only; fitted stays on the static cube to respect the WebGL context budget).
- **Uses your `filedef-fixtures` corpus** — dropped the old ad-hoc `model-samples/`; the `stl`/`three-mf` FormatPreview entries you seeded now light up as these subclasses.

Two things I'd value your eyes on:
1. I consolidated both formats onto the single `model3d` field rather than keeping per-format fields, since the isolated shell renders exactly one 3D slot and nothing overrides `isolated`.
2. Config-less-3MF edge: the section gates on `meshes || triangles`, so a 3MF with no slicer config reports no counts and its 3D section hides (STL always shows via `meshes: 1`). Open to a better approach if you'd prefer it always render.

CI is green. Screen recording of it rendering across the formats to follow on this comment. 🎥


https://github.com/user-attachments/assets/eebe930a-fdc4-4c3c-a238-5ca5bcbd0109

lukemelia (approved):

PR #5662, opened 2026-07-31
ci: Update index cache for host use

backspace:
As preparation for #5655, which updates host tests in CI to use cached index tables, this tweaks the cache-building job on `main` to:
* modify mtimes based on git commit times so newer files get indexed without having to reindex everything
* have a second cache subset that’s only the realms host tests use (base, skills, openrouter), as excluding bigger ones like catalog and experiments makes the cache download and restoration much faster

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1071,3 +1071,20 @@ jobs:
           name: boxel-index-cache
           path: /tmp/boxel-index-cache.sql.gz
           retention-days: 30
+      # The host-scoped snapshot covers only the realms a host test shard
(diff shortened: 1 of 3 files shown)


FadhlanR (approved):
[Claude Code 🤖] Approving — well-engineered CI change. Verified the load-bearing assumptions: all four cached tables carry `realm_url` (host filter is valid), the import TRUNCATE list matches CACHE_TABLES, and the host dump's COPY format/column order matches the full `pg_dump` so either replays unchanged. The content-hash mtime scheme handles shallow/separately-cloned realms and content drift cleanly.

PR #5668, opened 2026-08-02
fix: make Realm Runner BXL installs self-contained

christse:
## Background and Goal

The Realm Script notebook merge could not publish or deploy because every GitHub Actions job failed during `pnpm install`: `@cardstack/bxl` pointed at a private git repository that the Boxel workflow token cannot read.

This vendors the commit-pinned `runtime-bare` ESM bundle inside the private Realm Runner package, records its source commit, checksum, refresh command, and upstream licenses, and removes the private git dependency from the workspace lockfile. `BXL_API` remains available for local BXL development overrides.

Validated with a clean frozen workspace install, all 62 Realm Runner tests, Realm Runner lint, and a production Boxel CLI build.

### DIFF
--- a/packages/realm-runner/README.md
+++ b/packages/realm-runner/README.md
@@ -198,8 +198,10 @@ BXL is the jq-like transformation layer:
 return await realm.bxl.jq('.items | map(.price) | add', input);
 ```
 
-The package ships with a commit-pinned `@cardstack/bxl` runtime. `BXL_API` can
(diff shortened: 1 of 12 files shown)


christse:
Closing this PR because Realm Script notebooks and activity are experimental and must remain branch-only. The self-contained BXL install fix will be kept on the experiment branch instead.

PR #5673, opened 2026-08-03
Don't insert stray markers when formatting with no selection (CS-11697)

FadhlanR:
## Background and Goal

In the markdown editor, triggering an inline format with **no text selected** (cursor merely placed in the field) inserted a stray empty marker pair — `**`/`*`/`~~`/`` ` `` via `wrapWith`, or `[](url)` via `toggleLink` — that stayed visible in both source and preview. Fixes [CS-11697](https://linear.app/cardstack/issue/CS-11697).

The defect had two layers, and both are fixed:

1. **Command layer** (`packages/host/app/lib/codemirror-context.ts`) — `wrapWith` and `toggleLink` inserted the empty marker pair whenever the selection was collapsed. This fired the same for the toolbar button *and* the `Mod-B`/`Mod-I`/`Mod-\`` keyboard shortcut, so it was wrong regardless of trigger. Both now do nothing (no dispatch) when nothing is selected. The keymap is untouched — the shortcut simply no-ops on a bare cursor instead of producing `****`. One exception is preserved: `toggleLink` still **unlinks** from a bare caret inside an existing `[text](url)` (its syntax-tree unlink branch runs before the collapsed-cursor check).
2. **Toolbar layer** (`packages/base/codemirror-editor.gts`) — the inline-format buttons (Bold, Italic, Strikethrough, Code, Link) now carry `requiresSelection: true` and disable until text is highlighted, so a click can't reach the command and the UI signals inapplicability. An **active** toggle stays clickable without a selection, since untoggling works at a bare caret — concretely, a caret inside a link marks the Link button pressed and keeps it enabled so it can unlink. The link-at-caret signal comes from `findEnclosingLink`, the same syntax-tree check `toggleLink`'s unlink branch uses, so button state can't drift from command behavior. Line-based buttons (headings, lists, blockquote) act on a bare cursor and stay enabled. `hasSelection` was added to `sameToolbarState` so the toolbar re-renders when the selection collapses/expands.

## Tests

- `codemirror-editor-test.gts` — rewrote the `wrapWith` empty-selection test to assert *no change*; added a `toggleLink` no-op test for a caret outside any link and an unlink test for a bare caret inside a link; the selection-based unlink test now calls the real `toggleLink` (it previously simulated the command with a string scan); added a selection-info test pinning `formats.link` at a caret inside vs. outside a link.
- `rich-markdown-field-test.gts` — extended the "controls start disabled" test to cover all five inline-format toggles, and added a test driving the live editor's caret into and out of a link to assert the Link button's `aria-pressed` wiring.

Note: the focused-*with*-selection → enabled transition can't be asserted in headless CI because `view.hasFocus` ANDs `document.hasFocus()` (pre-existing, documented limitation); the disabled side, the active-state wiring, and the command-level behavior are covered.

## Screen Recording

https://github.com/user-attachments/assets/93577fb3-fff7-4c6d-8978-20e9623fda7b

### DIFF
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -7,7 +7,7 @@ import { on } from '@ember/modifier';
 import { scheduleOnce } from '@ember/runloop';
 import { htmlSafe } from '@ember/template';
 import { Tooltip } from '@cardstack/boxel-ui/components';
-import { eq, not } from '@cardstack/boxel-ui/helpers';
+import { eq } from '@cardstack/boxel-ui/helpers';
 
 import {
   baseRRI,
@@ -199,6 +199,13 @@ interface ToolbarItem {
   // binding in the CodeMirror keymap (bold/italic/code); absent items render a
   // label-only tooltip.
   shortcut?: string;
+  // Inline-format toggles (bold/italic/etc.) wrap the current selection, so they
+  // only make sense when text is highlighted. Set for those buttons so they
+  // disable when the selection is collapsed. Line-based buttons omit it.
+  requiresSelection?: boolean;
+  // Computed enablement for this button, folding in focus and (for
+  // selection-requiring buttons) whether text is highlighted.
+  disabled?: boolean;
 }
 
 const EMPTY_FORMATS: SelectionFormats = Object.freeze({
@@ -212,6 +219,9 @@ const EMPTY_FORMATS: SelectionFormats = Object.freeze({
 function sameToolbarState(a: SelectionInfo, b: SelectionInfo): boolean {
   return (
     a.hasFocus === b.hasFocus &&
+    // Selection presence gates the inline-format buttons' enablement, so a
+    // collapse/expand must refresh the toolbar even when nothing else changed.
+    a.hasSelection === b.hasSelection &&
     a.formats.bold === b.formats.bold &&
     a.formats.italic === b.formats.italic &&
     a.formats.code === b.formats.code &&
@@ -467,7 +477,13 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
   get toolbarButtons(): ToolbarItem[] {
     let f = this.toolbarFormats;
     let pressed = (active: boolean) => (active ? 'true' : 'false');
-    return [
+    let enabled = this.toolbarEnabled;
+    let hasSelection = this._selectionInfo?.hasSelection ?? false;
+    // Inline-format toggles additionally require a highlighted selection;
+    // line-based buttons only require focus.
+    let disabledFor = (requiresSelection?: boolean) =>
+      !enabled || (!!requiresSelection && !hasSelection);
+    let items: ToolbarItem[] = [
       {
         testId: 'bold',
         label: 'Bold',
@@ -476,6 +492,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         active: f.bold,
         ariaPressed: pressed(f.bold),
         shortcut: `${modKey}B`,
+        requiresSelection: true,
       },
       {
         testId: 'italic',
@@ -485,6 +502,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         active: f.italic,
         ariaPressed: pressed(f.italic),
         shortcut: `${modKey}I`,
+        requiresSelection: true,
       },
       {
         testId: 'strikethrough',
@@ -493,6 +511,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         action: this._wrapStrikethrough,
         active: f.strikethrough,
         ariaPressed: pressed(f.strikethrough),
+        requiresSelection: true,
       },
       {
         testId: 'code',
@@ -502,6 +521,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         active: f.code,
         ariaPressed: pressed(f.code),
         shortcut: `${modKey}\``,
+        requiresSelection: true,
       },
       {
         testId: 'link',
@@ -510,6 +530,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         action: this._toggleLink,
         active: f.link,
         ariaPressed: pressed(f.link),
+        requiresSelection: true,
       },
       { divider: true },
       {
@@ -550,6 +571,12 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         action: this._toggleBlockquote,
       },
     ];
+    for (let item of items) {
+      if (!item.divider) {
+        item.disabled = disabledFor(item.requiresSelection);
+      }
+    }
+    return items;
   }
 
   /** Prevent mousedown on toolbar/popup buttons from stealing editor focus/selection */
@@ -1212,7 +1239,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
                   is suppressed while the control is disabled. }}
               <Tooltip
                 @placement='top'
-                @disabled={{not this.toolbarEnabled}}
+                @disabled={{btn.disabled}}
                 data-test-toolbar-tooltip={{btn.testId}}
               >
                 <:trigger>
@@ -1222,7 +1249,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
                     type='button'
                     aria-label={{btn.label}}
                     aria-pressed={{btn.ariaPressed}}
-                    disabled={{not this.toolbarEnabled}}
+                    disabled={{btn.disabled}}
                     {{on 'mousedown' this._preventFocusLoss}}
                     {{on 'click' btn.action}}
                   >{{#let btn.icon as |Icon|}}<Icon
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -201,7 +201,9 @@ interface ToolbarItem {
   shortcut?: string;
   // Inline-format toggles (bold/italic/etc.) wrap the current selection, so they
   // only make sense when text is highlighted. Set for those buttons so they
-  // disable when the selection is collapsed. Line-based buttons omit it.
+  // disable when the selection is collapsed — unless the toggle is active
+  // (e.g. the caret sits inside a link), since untoggling works at a bare
+  // caret. Line-based buttons omit it.
   requiresSelection?: boolean;
   // Computed enablement for this button, folding in focus and (for
   // selection-requiring buttons) whether text is highlighted.
@@ -479,10 +481,12 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
     let pressed = (active: boolean) => (active ? 'true' : 'false');
     let enabled = this.toolbarEnabled;
     let hasSelection = this._selectionInfo?.hasSelection ?? false;
-    // Inline-format toggles additionally require a highlighted selection;
-    // line-based buttons only require focus.
-    let disabledFor = (requiresSelection?: boolean) =>
-      !enabled || (!!requiresSelection && !hasSelection);
+    // Inline-format toggles additionally require a highlighted selection,
+    // except when already active — an active toggle can always be untoggled
+    // (unlink works from a bare caret inside the link). Line-based buttons
+    // only require focus.
+    let disabledFor = (item: ToolbarItem) =>
+      !enabled || (!!item.requiresSelection && !hasSelection && !item.active);
     let items: ToolbarItem[] = [
       {
         testId: 'bold',
@@ -573,7 +577,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
     ];
     for (let item of items) {
       if (!item.divider) {
-        item.disabled = disabledFor(item.requiresSelection);
+        item.disabled = disabledFor(item);
       }
     }
     return items;
--- a/packages/host/app/lib/codemirror-context.ts
+++ b/packages/host/app/lib/codemirror-context.ts
@@ -1064,21 +1064,32 @@ function wrapWith(marker: string) {
   };
 }
 
-// Toggle a markdown link around the selection. Uses the syntax tree to detect
-// an enclosing [text](url) — a string scan can match across unrelated brackets
-// and delete text the user never selected.
+// Find the markdown Link syntax node enclosing the given range, if any. Uses
+// the syntax tree — a string scan can match across unrelated brackets and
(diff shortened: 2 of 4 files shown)


packages/base/codemirror-editor.gts
```diff
@@ -510,6 +530,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         action: this._toggleLink,
         active: f.link,
         ariaPressed: pressed(f.link),
+        requiresSelection: true,
```
FadhlanR, line 537: [Claude Code 🤖] **Regression (minor, non-blocking): unlinking with a collapsed cursor is no longer reachable from the UI.**

**The mechanism.** `toggleLink` in `packages/host/app/lib/codemirror-context.ts` resolves the syntax tree and checks for an enclosing `Link` node *before* its collapsed-selection check, and a collapsed cursor inside `[text](url)` satisfies `from >= link.from && to <= link.to` — so previously, placing the caret in a link and clicking the Link button unlinked it. This `requiresSelection: true` disables the button whenever the selection is collapsed, and the toolbar button is `toggleLink`'s only caller (the markdown keymap binds only Mod-B / Mod-I / Mod-`), so the collapsed-cursor entry into the unlink branch is now dead in practice.

**Verified.** Traced the base-branch `toggleLink`: a collapsed caret inside a link takes the unlink branch (it precedes `if (from === to)`) and dispatches the text replacement. Grepped all non-test callers: `toggleLink` is invoked only via `_toggleLink` in this file.

**What still works.** Selecting any range within the link enables the button, and that selection still satisfies the unlink guard — unlink remains reachable, it just now demands a selection. That's why this is minor.

**The way out — two options, either fine:**
1. **Accept it** (select-then-click still unlinks, and it's consistent with the disabled-button story). If so, consider tightening the no-op comment inside `toggleLink` — "Linking applies to selected text only" isn't quite true of the function itself, whose earlier branch still unlinks at a collapsed caret; it's only true of what the UI now lets through.
2. **Restore it**: enable the Link button when the collapsed caret sits inside a link. That needs a link-at-caret signal in `SelectionInfo` — note the update listener in `createEditorState` hardcodes `formats` to all-false for collapsed selections, so `formats.link` can't serve as-is.

packages/host/app/lib/codemirror-context.ts
```diff
@@ -1011,13 +1011,11 @@ function wrapWith(marker: string) {
     let { from, to } = view.state.selection.main;
     let len = marker.length;
 
-    // No selection: insert an empty pair of markers and drop the cursor
-    // between them so the user can type the content (e.g. **|**).
+    // No selection: do nothing. Inserting an empty marker pair here leaves
+    // stray markers (e.g. **) visible in source and preview. Inline formatting
+    // applies to selected text only — the toolbar disables these buttons and
+    // the keyboard shortcut is a no-op until the user highlights something.
     if (from === to) {
```
FadhlanR, line 1018: [Claude Code 🤖] **Confirmation: returning `true` on a collapsed selection is the right call.**

Returning `false` would let the key event fall through CodeMirror's keymap to the browser default, and these chords have browser-level bindings in some contexts (e.g. Ctrl-B toggling a bookmarks sidebar). Consuming the shortcut while editing nothing is the correct no-op, and the rewritten test pins both halves — returns true, document and cursor unchanged.

**When this would stop being safe:** if a lower-priority keymap ever wants to handle Mod-B/Mod-I/Mod-` on a collapsed cursor, this `true` will shadow it silently. Nothing binds them today.

FadhlanR (commented):
[Claude Code 🤖] This review focused on whether the two layers of the fix — the command-level no-op and the toolbar enablement — stay in agreement, and whether any existing collapsed-cursor behavior was load-bearing before being gated off. The diff is small; the interesting questions are all one call site away from it.

**Bottom line: no blocking issues.** One minor behavior regression worth an explicit decision (collapsed-cursor unlink, inline thread on the Link button), everything else verified sound.

### What lands right

- **Fixing at both layers is the correct shape.** The keymap can't be gated by toolbar state, so the command-layer no-op in `wrapWith`/`toggleLink` is the real fix; the disabled buttons are honest UI on top rather than the only line of defense.
- **The `hasSelection` addition to `sameToolbarState` is necessary, not defensive.** Verified by tracing the memo: expanding a selection over *unformatted* text changes neither `hasFocus`, nor any `formats` flag (all remain false), nor `currentRef` — so without the new comparison the update is swallowed, `_selectionInfo` never refreshes, and the inline buttons stay disabled with text visibly highlighted. The new comparison catches exactly that transition.
- **Line-based buttons staying enabled is right** — headings/lists/blockquote act on the cursor's line, so a collapsed cursor is a valid target for them.
- **Test posture is sound.** Both command-level no-ops are pinned (document unchanged *and* cursor position unchanged), and the headless focus limitation is documented rather than papered over with an assertion that would flake.

### Recommendation

1. Decide the collapsed-cursor unlink question — see the inline thread on the Link button's `requiresSelection` in `packages/base/codemirror-editor.gts`. Non-blocking either way, but the code and comment should match whichever way it goes.

### Adjacent, out of scope

- With the caret inside already-bold text, Mod-B now no-ops. Not a regression — the previous behavior inserted a stray empty marker pair there, which is worse — but conventional editors toggle formatting at the caret (unwrap the enclosing marks). If users ask for editor parity, that's the natural follow-up, and it lives entirely in `wrapWith`.
- The Add embed / Edit embed buttons at the left of the toolbar don't participate in the disabled model at all — they render enabled even before the editor has focus, unlike every other control. Pre-existing, noticed while reading the toolbar template; flagging for whoever touches that toolbar next.

packages/base/codemirror-editor.gts
FadhlanR: [Claude Code 🤖] Resolved by restoring the affordance (option 2). What changed:

- `findEnclosingLink` is extracted from `toggleLink`'s unlink branch and shared with the selection-info emitter in `createEditorState`, so the toolbar's Link-button state is computed by the exact same syntax-tree check the command runs — the two can't drift.
- At a collapsed caret the emitter now reports `formats.link` from that helper (the other formats stay false, matching `wrapWith`'s no-op at a caret), and the toolbar's enablement rule gains an escape: an **active** toggle stays clickable without a selection, since untoggling works at a bare caret.
- The stale wording in `toggleLink`'s no-op comment is fixed — that branch is now correctly described as "caret outside any link".

Coverage added: a `toggleLink` unlink-from-bare-caret command test, a selection-info test pinning `formats.link` at a caret inside vs. outside a link, and a component test asserting the Link button's `aria-pressed` flips as the caret enters and leaves a link (drivable headless — only enablement needs OS focus). The pre-existing selection-based unlink test also now calls the real `toggleLink` instead of simulating it with a string scan. Both touched modules pass: codemirror-context 55/55, RichMarkdownField 30/30.

lukemelia (commented):
In most writing environments I could type something, then press Ctrl+B and what I type after that is in bold and then press Ctrl+B again and what I type after that is not bold. Does that pattern work before/after this change?

FadhlanR:
[Claude Code 🤖]

Good question — and the honest answer is that the pattern doesn't fully work today either. Here's what I found tracing both the before and after paths, and why I landed on disabling rather than fixing the toggle in place.

**Before/after, precisely.** That type → `Ctrl+B` → bold → `Ctrl+B` → plain flow is really a WYSIWYG behavior (Word, Docs, Notion), where bold is a *stored attribute* you can toggle at the caret. In a markdown source editor bold is literal `**` characters, so there's no caret-level mode to flip.

Before this PR, `Ctrl+B` at a bare caret inserted an empty `**|**` pair with the cursor in the middle:
- Typing after it landed between the markers → rendered bold, so the *first half* of the pattern worked by accident.
- The second `Ctrl+B` to turn it off did **not** work — the caret sits just inside the closing `**`, so it inserted *another* empty pair (`**world****|**`) instead of exiting bold.
- And if you pressed `Ctrl+B` then clicked away without typing, you left orphan `**` behind — invisible in our live preview but present in the source. That's the bug this ticket was filed on.

After this PR, `Ctrl+B` at a bare caret is a no-op, so none of that happens.

**Why not just fix the toggle** (keep inserting the pair, make the second `Ctrl+B` exit)? It would make the shortcut feel better, but it doesn't fix the actual reported bug — you'd still leave orphan `**` markers any time you hit `Ctrl+B` and navigate away before typing. So it's more code than the no-op while still failing the ticket.

**Why disabling reads cleaner from a UX angle.** Because our markers are *hidden* in live preview, an empty pair gives the user no feedback that they're "in bold mode" — there's literally nothing on screen. So the half-working old behavior was also a bit of a mystery-mode. Requiring a selection makes the model honest and predictable: inline formatting applies to text you've selected, and the disabled button + no-op shortcut say so consistently.

**The real way to get the pattern you described** is proper pending-format state: don't write any markers until the first keystroke, track "bold pending" in editor state, materialize `**…` on that first character, and let a second `Ctrl+B` before typing just clear the flag. That eliminates the orphan markers *and* gives the exact type → bold → type → plain behavior (with the toolbar button staying pressed as the in-mode indicator, since the markers are hidden). It's a real feature though — happy to file it as a follow-up and keep this PR as the minimal bug fix if that works for you.

FadhlanR:
@cmgardella Your input would be really helpful here.

packages/host/app/lib/codemirror-context.ts
```diff
@@ -1258,11 +1264,15 @@ function createEditorState(options: CreateEditorStateOptions): EditorState {
...
+                  // The wrap toggles no-op at a caret, so they read inactive.
+                  // Link is the exception: toggleLink can still unlink an
+                  // enclosing [text](url) from a bare caret, so report it
+                  // active — the toolbar keeps an active toggle clickable.
                   bold: false,
                   italic: false,
                   code: false,
                   strikethrough: false,
-                  link: false,
+                  link: !!findEnclosingLink(update.state, from, to),
```
habdelra, line 1275: [Claude Code 🤖] **Follow-up (non-blocking, pre-existing): the "no drift" invariant this PR introduces covers the caret path but not the selection path.**

**The claim.** `findEnclosingLink`'s doc comment says it's "shared by toggleLink's unlink branch and the selection-info emitter, so the toolbar's Link-button state can't drift from what the command actually supports." That's exactly right for the **no-selection** branch here — `formats.link` at a bare caret is the same syntax-tree walk `toggleLink` runs, so the pressed state and the unlink behavior are guaranteed to agree.

**Where it still drifts.** The sibling branch — `hasSelection ? detectFormats(...)` — computes `formats.link` from a string scan: `detectFormats` returns `link: expanded.includes('](')` over a ±3-char window. That's the same fragile scan this PR just retired inside `toggleLink` ("a string scan can match across unrelated brackets"). So with a *selection*, the Link button's `aria-pressed` can still disagree with what clicking does:

- Doc `see ](x) here`, select `x`. The window contains the literal `](`, so `detectFormats` reports `link: true` and the button reads pressed. But there's no `Link` node (no opening `[`), so `findEnclosingLink` returns null and clicking runs the **wrap** branch — turning the selection into `[x](url)`, the opposite of the "toggle off" the pressed state implies. (`foo[i](bar)` would *not* trip this — that parses as a real link; you need a bare `](` outside any link.)

**Scope.** Pre-existing — `detectFormats` predates this PR — and low-severity: it only affects the visual pressed state, not enablement (with a selection `hasSelection` is true, so `disabledFor` ignores `active`) and not which branch `toggleLink` takes. **Not this PR's bug and not blocking.** Flagging it because this is the PR that establishes the "can't drift" invariant, so the natural follow-up is to source `formats.link` from `findEnclosingLink` in the selection branch too — retiring the string scan for links entirely and making the invariant hold in both directions.

---

habdelra (commented):
[Claude Code 🤖] This review independently re-derives the two-layer fix (command-level no-op + toolbar enablement) and focuses on the one open question left on the PR — the Ctrl+B toggle-at-caret behavior — since that's what's blocking sign-off.

**Bottom line: no blocking issues.** The fix is correct, the enablement rule is sound, and CI is green on the head commit (all 20 host test shards + memory baseline + visual checks passed; the "unstable" mergeable flag is the branch trailing `main`, not a failing check). One non-blocking, pre-existing follow-up on link-detection drift (inline, on the `formats` emitter in `codemirror-context.ts`).

### Independently verified

- **The enablement rule is airtight.** `disabledFor = !enabled || (requiresSelection && !hasSelection && !active)`. Worked the truth table: at a bare caret the four wrap toggles have `active === false` (hardcoded in the emitter), so they resolve to disabled; the Link button gets `active` from `findEnclosingLink`, so a caret inside a link keeps it enabled and clickable — the only path that reaches `toggleLink`'s unlink branch. `active` can only flip enablement when `!hasSelection`, so a stale/false `active` under a selection can't wrongly enable anything.
- **`toggleLink`'s only production caller is the toolbar button.** Grepped every non-test call site: the keymap binds only `Mod-B`/`Mod-I`/`` Mod-` `` (all `wrapWith`), and `_toggleLink` is the sole `toggleLink` caller. That's what made the collapsed-cursor unlink unreachable once the button was gated — and what makes restoring it via the `active` escape sufficient.
- **`hasSelection` in `sameToolbarState` is load-bearing, not defensive.** Expanding a selection over unformatted text changes no `formats` flag, no `hasFocus`, no `currentRef` — without the new comparison the memo swallows the update and the buttons stay disabled with text visibly selected.

### On the open Ctrl+B question (@lukemelia)

The type → Ctrl+B → bold → Ctrl+B → plain pattern does **not** work today, before or after this PR — verified mechanically: the keymap binds `Mod-B` to `wrapWith('**')`, which has no caret-level "stored format" mode; in a markdown source the bold is literal `**`, so there's nothing to toggle at an empty caret. Before this PR, Ctrl+B on a bare caret inserted `**|**` (the first keystroke landed bold by accident, but a second Ctrl+B inserted *another* pair rather than exiting, and clicking away left orphan `**` — the reported bug). After, it's a clean no-op. The author's write-up matches what the code does.

My recommendation: **ship this as the minimal bug fix.** The real WYSIWYG behavior (pending-format state — defer writing markers until the first keystroke, track "bold pending", let a second Ctrl+B clear it) is the only thing that delivers the pattern *and* eliminates the orphan markers, and it's well out of this PR's scope. Worth filing as its own feature.

### Existing threads

- **Collapsed-cursor unlink** — resolved by the current code (option 2). The `active`-keeps-enabled escape plus the shared `findEnclosingLink` signal restore the affordance; confirmed against HEAD. See the inline note above for the one remaining gap (the *selection*-path pressed state still uses the string scan).
- **Returning `true` on a collapsed selection** — agree it's correct. Consuming the chord avoids falling through to browser-level `Mod-B` bindings, and nothing lower in the keymap wants those keys at a caret today.

### Adjacent, out of scope

The toggle-at-caret feature above is the main one and is already well-discussed in the thread, so not repeating the detail. No new adjacent rot noticed beyond what the earlier review already flagged (the left-of-toolbar embed buttons not participating in the disabled model).

---

habdelra (approved):

PR #5674, opened 2026-08-03
Open the embed chooser for the /card slash command

FadhlanR:
## Problem

Typing `/` in the rich markdown editor offers a `/card` autocomplete option. Selecting it opened a custom inline popup ([CS-11698](https://linear.app/cardstack/issue/CS-11698)) that was broken two ways:

1. **Couldn't be dismissed** — the only close path was pressing <kbd>Esc</kbd> while the search input was focused. No click-outside, blur, or Cancel.
2. **Search didn't filter** — the query filtered on `contains: { name: ... }`, but `CardDef` has no queryable `name` field (the searchable title is `cardTitle`), so typing never matched. The result rows and format picker read `card.title` for the same reason, so they rendered blank too.

## Fix

Point the `/card` slash command at the **same markdown embed chooser modal the toolbar's Add-embed button already uses**, instead of the bespoke inline popup. The modal has working `cardTitle` search, card/file tabs, format selection, and standard modal dismissal — so both bugs disappear because the broken code is gone, and there's one card chooser to maintain instead of two.

- `_handleOpenCardSearch` now delegates to `_openEmbedChooser('card')`. The slash completion's `apply` already deletes the typed `/`, so the caret sits where the directive lands.
- Removed the inline card-search popup and format-picker popup: templates, tracked state, keydown/input/select handlers, menu-coordinate logic, the search resource, `_insertCardWithFormat`, and their CSS.
- Removed the now-dead `makeCardRef` / `labelFromUrl` helpers and the `maybeRelativeReference` import. Kept `_preventFocusLoss` (used by the toolbar) and the `getCards` arg (used for embed reference resolution).

Net diff: **+79 / −465**.

## Tests

New acceptance test in `markdown-embed-chooser-test.gts`: drives the `/card` autocomplete (`startCompletion` → `acceptCompletion`), asserts the chooser modal opens on the Cards tab and the old `[data-test-card-search]` popup is gone, then picks a card and verifies the inserted `:card[../Pet/mango]` directive (the typed `/` is consumed).

## Verification

- `ember-tsc` (host `lint:types`) — 0 errors
- `prettier --check` — clean
- `ember-template-lint` — clean

Browser test run for the new acceptance test still pending.

### Before

https://github.com/user-attachments/assets/e428d760-130f-4253-8827-d8c95daf643f

### After

https://github.com/user-attachments/assets/660a4f98-30bf-4197-98bd-4bfb565de110

### DIFF
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -90,7 +90,7 @@ interface CodeMirrorContext {
     content: string;
     onDocChange: (text: string) => void;
     onCardTargetsChange: (targets: CardWidgetTarget[]) => void;
-    onOpenCardSearch: () => void;
+    onOpenEmbedChooser: () => void;
     onSelectionChange?: (info: SelectionInfo) => void;
     livePreview?: boolean;
   }) => any;
@@ -247,13 +247,13 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
     this._isLoaded = true;
   }
 
-  // ── Card search logic ────────────────────────────────────────────────────
+  // ── Slash-command embed chooser ──────────────────────────────────────────
 
   // Typing `/card` reuses the same embed chooser modal as the toolbar's
   // Add-embed button. The slash completion's `apply` already deleted the typed
   // `/`, so the caret sits where the directive should land — `_openEmbedChooser`
   // inserts at the current selection.
-  private _handleOpenCardSearch = () => {
+  private _handleOpenEmbedChooser = () => {
     if (isDestroying(this) || isDestroyed(this)) return;
     this._openEmbedChooser('card');
   };
@@ -869,7 +869,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
         }
       },
       onCardTargetsChange: this._handleTargetChange,
-      onOpenCardSearch: this._handleOpenCardSearch,
+      onOpenEmbedChooser: this._handleOpenEmbedChooser,
       onSelectionChange: this._handleSelectionChange,
     });
 
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -90,7 +90,7 @@ interface CodeMirrorContext {
     content: string;
     onDocChange: (text: string) => void;
     onCardTargetsChange: (targets: CardWidgetTarget[]) => void;
-    onOpenCardSearch: (pos: { from: number; to: number }) => void;
+    onOpenCardSearch: () => void;
     onSelectionChange?: (info: SelectionInfo) => void;
     livePreview?: boolean;
   }) => any;
@@ -1014,7 +1014,6 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
           {{on 'mousedown' this._focusEditorOnPointerDown}}
           {{this.mountEditor this.cm @content @onUpdate this.livePreview}}
         ></div>
-
       </div>
 
       {{#if this.livePreview}}
@@ -1033,9 +1032,9 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
                     <span
                       class='codemirror-card-slot
                         {{if
-                          (eq target.format 'atom')
-                          'codemirror-card-slot--inline'
-                          'codemirror-card-slot--inline-embed'
+                          (eq target.format "atom")
+                          "codemirror-card-slot--inline"
+                          "codemirror-card-slot--inline-embed"
                         }}'
                       style={{if target.style (htmlSafe target.style)}}
                       data-test-codemirror-file-slot-inline={{if
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -11,7 +11,6 @@ import { eq, not } from '@cardstack/boxel-ui/helpers';
 
 import {
   baseRRI,
-  maybeRelativeReference,
   resolveRRIReference,
   rri,
   CardContextName,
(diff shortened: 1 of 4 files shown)


packages/base/codemirror-editor.gts
```diff
@@ -1353,9 +1032,9 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
                     <span
                       class='codemirror-card-slot
                         {{if
-                          (eq target.format 'atom')
-                          'codemirror-card-slot--inline'
-                          'codemirror-card-slot--inline-embed'
+                          (eq target.format "atom")
+                          "codemirror-card-slot--inline"
+                          "codemirror-card-slot--inline-embed"
```
habdelra, line 1037: [Claude Code 🤖] **Cosmetic: unrelated quote-style flip (non-blocking).** These three lines flipped from single to double quotes — `(eq target.format "atom")`, `"codemirror-card-slot--inline"`, `"codemirror-card-slot--inline-embed"` — but every other string literal in this template uses single quotes, including `(eq target.refType 'file')` just four lines below. Both spellings are valid in Glimmer and the rendered class is identical, so this is harmless; it only adds noise to an otherwise tightly-scoped deletion and leaves the block inconsistent with its neighbors. Suggest reverting to single quotes to match:

```hbs
                          (eq target.format 'atom')
                          'codemirror-card-slot--inline'
                          'codemirror-card-slot--inline-embed'
```

Scope: cleanup, non-blocking.

---

packages/base/codemirror-editor.gts
```diff
@@ -282,156 +249,13 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
 
   // ── Card search logic ────────────────────────────────────────────────────
 
-  private _handleOpenCardSearch = (_pos: { from: number; to: number }) => {
+  // Typing `/card` reuses the same embed chooser modal as the toolbar's
+  // Add-embed button. The slash completion's `apply` already deleted the typed
+  // `/`, so the caret sits where the directive should land — `_openEmbedChooser`
+  // inserts at the current selection.
+  private _handleOpenCardSearch = () => {
```
habdelra: [Claude Code 🤖] **Names now trail behind behavior (non-blocking).** After this change there is no card-search logic left in the component — `_handleOpenCardSearch` just opens the shared embed chooser on the card tab, and the callback threaded through `codemirror-context.ts` (`onOpenCardSearch`, and the slash source's `onOpenCardSearch` param) is the same story. The inline comment you added here is clear and carries the intent, but the method name, the callback name, and the `// ── Card search logic ──` section header directly above (line 250) still describe the deleted inline popup. The next reader who greps for "card search" will land on machinery that no longer exists.

Consider renaming to `_handleOpenEmbedChooser` / `onOpenEmbedChooser` and retitling the section (e.g. `// ── Slash-command embed chooser ──`). Purely a readability follow-up — no behavior implication — and fine to defer.

Scope: naming/readability, non-blocking.

---

habdelra (commented):
[Claude Code 🤖] This review focused on whether the delete-and-delegate is behavior-complete: does the caret land where the typed `/` was, is every trace of the removed popup gone with no dangling references, did the two twin callback signatures stay in sync, and does the test actually pin the caret-placement contract.

**Bottom line: no blocking issues.** This is a clean deletion-and-delegation — the `/card` path now reuses the canonical embed chooser modal, the removed popup's state/handlers/CSS/helpers are gone with zero remaining references, and the caret-placement contract is both correct and covered. Two non-blocking cosmetic notes inline.

**What lands right**

- **Caret placement is correct, and I traced it.** The completion's `apply` (`codemirror-context.ts`) dispatches `{ from, to, insert: '' }`, deleting the typed `/…` and collapsing the selection to `from`. Because nothing mutates the document while the modal is open, `_insertBfm` reading `view.state.selection.main.from` inserts exactly there. The mid-line test is the one that actually proves this — the doc-start case can't distinguish "insert at the caret" from "insert at position 0", so adding the `Hello /` case is the right instinct and closes the real gap.
- **Dead-code removal is complete.** Verified 0 remaining references to every removed field/method (`_cardSearchMode`, `_cardSearchText`, `_cardSearchIndex`, `_menuCoords`, `menuStyle`, `_focusSearchInput`, `cardSearchResults`, `isSearchLoading`, `_selectCardResult`, `_insertCardWithFormat`, `_dismissCardSearch`, `_dismissFormatPicker`, `_formatPicker*`, `_searchResource*`), and no import went dead: `makeCardRef` / `labelFromUrl` / `maybeRelativeReference` are fully gone, while `scheduleOnce`, `eq`, `not`, `fn`, `htmlSafe`, `trimJsonExtension`, and the retained `getCards` arg are all still used. `willDestroy` no longer touches any removed field. No stale `data-test-*` selectors remain (the only `[data-test-card-search]` reference left is the new test asserting the old popup is gone).
- **Twin signatures stayed in sync.** The callback narrowed to `() => void` in both homes — `CodeMirrorContext` and `CreateEditorStateOptions` — and both call sites (`this._handleOpenCardSearch`, the integration test's `() => {}`) match, so nothing silently kept the old `(pos) => void` shape.

**Recommendations** (both non-blocking, both cosmetic)

1. Revert the single→double quote flip at the card-slot class expression to match the rest of the template — see inline on `codemirror-editor.gts`.
2. Rename `_handleOpenCardSearch` / `onOpenCardSearch` and the "Card search logic" section header to reflect that this now opens the embed chooser — see inline on `codemirror-editor.gts`.

**Adjacent, out of scope**

- **Paste-a-URL is no longer reachable from `/card`.** The removed inline popup had a working path, independent of the broken search, where pressing Enter on text starting with `http://` / `https://` / `./` embedded that URL directly (its placeholder was "Search cards or paste URL…"). The modal's card tab is search-only (`<MiniCardChooser>`), so that affordance disappears with the popup. This is almost certainly an acceptable trade for consolidating on one chooser — just flagging it in case paste-a-URL was a relied-upon entry point; if so, that's a follow-up on the modal, not this PR.

Not a review finding, for context: the description notes the browser test run was still pending — host tests are re-running on the latest commit and were green (0 failures) on the prior run.

---

habdelra (approved):

lukemelia (approved):

PR #5684, opened 2026-08-04
Fix opening cards in preview mode

jurgenwerk:
Clicking "Open" didn't work:

image

### DIFF
--- a/packages/host/app/components/operator-mode/preview-panel/index.gts
+++ b/packages/host/app/components/operator-mode/preview-panel/index.gts
@@ -32,7 +32,10 @@ import {
   type ResolvedCodeRef,
 } from '@cardstack/runtime-common';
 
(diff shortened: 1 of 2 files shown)


lukemelia (approved):

burieberry (commented):
This is helpful but that teal color is not meant to be used by itself on light background for legibility reasons, even though ai loved to use it that way. Can we make the Open button a primary, secondary, or default boxel ui button?

burieberry:
> This is helpful but that teal color is not meant to be used by itself on light background for legibility reasons, even though ai loved to use it that way. Can we make the Open button a primary, secondary, or default boxel ui button?

I see now that this button wasn't a new addition. In that case, up to you if you want to handle in this pr.

jurgenwerk:
@burieberry I created a new ticket to address that

PR #5704, opened 2026-08-06
fix: wait for a busy room instead of dropping the message

jurgenwerk:

Send the assistant a message while it is mid-turn and the message is silently destroyed. No reply, no error, and nothing in the bot's log.

The room lock is taken per event. When it can't be taken, the handler returned:

```js
if (!roomLock) {
  // Some other instance is already processing a recent event in this room. Ignore it.
  return;
}
```

Ignoring it is safe when a second replica holds the lock — that replica answers the message. With a single instance the contender is the same bot still working through its own tool-call chain, and nothing revisits a room's events once that chain ends, so the turn is simply lost. The return also sits above the line that logs a received message, which is why a dropped message leaves no trace at all.

Observed locally: a message sent at 11:47 while a chain was finishing got no response. It was answered 23 minutes later, when an unrelated "hello?" acquired the lock and pulled the older message in from history. The bot's log has lines for the messages either side of it and none for the lost one.

The lock is now waited for, so the message is handled as soon as the room frees up. The wait is bounded at two minutes — enough to outlast an agentic turn, which routinely runs 40+ seconds — and a holder that dies without releasing times the waiter out and logs it, instead of passing in silence.

Tests cover both paths: a contended room is waited for rather than abandoned, and a room that is never released times the waiter out.

One thing worth review: a message that arrives mid-chain may now be seen twice — once by a later continuation that rebuilds history and includes it, and again by the waiting turn — where previously it was seen zero times. Zero is clearly worse than one, and this trades it for one-or-two. Deduplicating properly needs a way to tell whether the in-flight chain already answered a given event, which is a larger change than this.


jurgenwerk:
[Claude Code 🤖] Closing — the incident this cited turned out to be a host-side stalled send, not a dropped event in the bot. Synapse shows the message only reached the room 23 minutes after it was sent, so the bot never had it. Reopening the underlying question separately if the bare `return` on lock contention is worth hardening on its own merits.

PR #5710, opened 2026-08-06
Fix card embeds in Markdown

backspace:
Various types of card embeds are currently broken:

s 2026-08-10 at 07 37 55@2x

s 2026-08-10 at 07 38 17@2x

This fixes them:

s 2026-08-10 at 07 39 10@2x


packages/runtime-common/query-field-utils.ts
```diff
@@ -221,16 +225,54 @@ export function normalizeQueryDefinition({
...
+      return resolved;
+    }
+    // The resolver knows the realms this process has mappings for, which is not
+    // every realm that exists: a field may name a peer realm directly, and that
+    // has always been honored as written. Distinguish the two by shape — a
+    // realm is addressed by its root and so ends in a slash, while a resource
+    // identifier does not. Keeping an unplaceable root means an explicit
+    // cross-realm target still works; dropping an unplaceable resource means a
+    // reference is never mistaken for a realm to search.
+    return value.endsWith('/') ? value : undefined;
```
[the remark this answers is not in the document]
backspace: fixed in 677c3ba500

packages/runtime-common/virtual-network.ts
```diff
@@ -208,6 +208,45 @@ export class VirtualNetwork {
...
+   * reference into the latter, and the deeper one is the real holder. This is
+   * why a realm can't be recovered by counting path segments — realm roots are
+   * whatever was registered, at whatever depth.
+   */
+  realmForReference(reference: string): string | undefined {
+    let best: string | undefined;
+    let bestLength = -1;
+    for (let [prefix, target] of this.realmMappings) {
+      for (let candidate of [prefix, target]) {
+        if (reference.startsWith(candidate) && candidate.length > bestLength) {
```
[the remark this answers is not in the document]
backspace: fixed in 41562dcfd6

packages/base/card-serialization.ts
```diff
@@ -217,6 +217,49 @@ export function resourceFrom(
...
+// (`resource.id ? toURL(resource.id) : primaryURL`).
+export function rebaseReferencesFor(
+  value: CardDef | FileDef,
+  opts?: SerializeOpts,
+): SerializeOpts | undefined {
+  if (!opts?.maybeRelativeReference || opts.useAbsoluteURL || !value.id) {
+    return opts;
+  }
+  return { ...opts, maybeRelativeReference: relativeReferenceFor(value) };
+}
```
habdelra, line 261: [Claude Code 🤖] **Confirmation — this is the client half of a two-place contract, worth guarding.**

`rebaseReferencesFor` decides whether a link target's included-body references rebind to the target's own id purely by the presence of `value.id`: no id → returns `opts` unchanged, so the target inherits the parent's base. That is the exact rule the served-document path applies in `relativizeResource` (`realm-index-query-engine.ts`): `let resourceURL = resource.id ? virtualNetwork.toURL(resource.id) : primaryURL`. So a target with an id relativizes against itself and a target without one inherits the primary — identically on both sides. Verified by reading both.

The other two guards are correct too: `useAbsoluteURL` short-circuits (absolute refs are never relativized), and `!opts?.maybeRelativeReference` means there is no relativizer to rebind. And the callers keep the relationship-link base separate from the body — `LinksTo`/`LinksToMany` still build `links.self` from the original `opts` (parent base) and only the included body uses the rebased relativizer, matching that a relationship link lives on the primary resource, not on the target.

The reason this deserves a note: it's a silent-drift hazard. If a later change teaches one side to rebase on something other than own-id-vs-primary and not the other, client-serialized and server-served documents would encode included references differently and deserialize to different ids, with nothing failing loudly. Non-blocking.

---

packages/base/card-api.gts
```diff
@@ -1098,9 +1106,15 @@ class Contains<CardT extends FieldDefConstructor> implements Field<CardT, any> {
...
+      // `opts` reaches the nested resource so a composite field serializes the
+      // same way whether it is reached through `contains` or `containsMany` —
+      // most visibly `includeComputeds`, without which a computed declared
+      // inside a FieldDef is filtered out of the nested resource. `visited` is
+      // deliberately a fresh set here, matching the other three field
+      // serialize sites.
       let serialized: JSONAPISingleResourceDocument['data'] & {
         meta: Record<string, any>;
-      } = callSerializeHook(this.card, value, doc);
+      } = callSerializeHook(this.card, value, doc, undefined, opts);
```
habdelra, line 1117: [Claude Code 🤖] **Confirmation — the core fix, and it's the safe kind of change despite touching every card's serialization.**

Forwarding `opts` here makes composite `contains` match the other three field-serialize sites. I checked each in this file: `ContainsMany` primitive (the `values.map` above) and `ContainsMany` composite both already call `callSerializeHook(this.card, value, doc, undefined, opts)`, and `Contains` primitive does too — composite `contains` was the lone site dropping it. So this isn't a new propagation path, it closes the one gap in an existing one.

That matters for blast radius: `opts` carries more than `includeComputeds` (also `includeUnrenderedFields`, `maybeRelativeReference`, `useAbsoluteURL`). Because the sibling `containsMany` composite has threaded all of these through unchanged for as long as it has existed, the newly-reached nested composite now behaves the way an equivalent `containsMany` composite already did — the `includeUnrenderedFields` fallout (a never-authored nested link serializing as `{ links: { self: null } }`) is the same shape that path already produced, which is what the baseline updates reflect.

The user-visible bug this fixes: `RichMarkdownField.cardReferenceUrls` is a computed declared inside the field, so under the old drop it was filtered out of the indexed resource; `linkedCards` interpolates `$this.cardReferenceUrls`, found nothing, and the query aborted — the reference then indexed as an ordinary unresolved link with no error to surface it. The new integration test *can index a card whose contained field has a query-backed link driven by a computed in that same field* pins exactly this chain. Non-blocking.

---

packages/runtime-common/query-field-utils.ts
```diff
@@ -221,16 +240,58 @@ export function normalizeQueryDefinition({
...
+    // as written — a field may target a peer this process has never heard of,
+    // and that worked before. Values that arrived by interpolation are the
+    // instance's own data, so an unplaceable one is dropped rather than
+    // mistaken for a realm to search.
+    //
+    // Provenance decides this, not spelling: a realm href may legitimately be
+    // written without a trailing slash (`buildQuerySearchURL` and
+    // `parseRealmsParam` both normalize that), so shape says nothing about
+    // whether a string names a realm or a resource.
+    return realmsAreInterpolated ? undefined : value;
```
habdelra: [Claude Code 🤖] **Non-blocking follow-up: provenance is tracked for the whole `realms` value, so a literal realm in a mixed list inherits interpolation provenance.**

`realmsAreInterpolated` is computed once over the entire `specifiedRealms` (`isInterpolationToken(specifiedRealms) || (Array.isArray(...) && ...some(isInterpolationToken))`), and `resolveOneRealm` applies that single verdict to every entry. For the two real shapes this is exactly right: a bare `$this.…` string (rich-markdown) and an all-literal explicit list each get a uniform, correct verdict.

The gap is a *mixed* list, e.g. `realms: ['https://peer.example/', '$this.someRef']`. The literal `https://peer.example/` was authored into the query, so the invariant this establishes says it should be honored as written — but because a sibling entry is an interpolation, `realmsAreInterpolated` is true for all entries, so if the resolver can't place `peer.example` it is dropped rather than used verbatim. (And when the interpolation resolves to an array, `interpolateNode` produces a nested array and `resolveOneRealm` throws on the non-string entry — so mixed lists are effectively unsupported today regardless.)

No current caller writes a mixed list, so this blocks nothing. If the invariant should hold for that shape later, provenance would need to be per-entry: a literal string is always "authored", only a value substituted from a `$this.…` token is "interpolated". Flagging so the whole-value approximation is a known boundary rather than a surprise. Follow-up, not this change.

---

habdelra (commented):
[Claude Code 🤖] **Review focus:** the two mechanisms this change rests on — reference relativization across the client/server boundary, and cross-realm realm resolution for query-backed fields — and whether the serialization fix lands without collateral.

**Bottom line: no blocking issues.** The change is well-factored and the invariants it depends on hold on both sides. CI is green (host and realm suites), the two automated-review findings are resolved, and the new integration + unit tests pin the actual failure chains rather than just output shape. The inline comments are mostly confirmations that document the load-bearing contracts so the next editor doesn't break them, plus one non-blocking follow-up and one adjacent note.

**What lands right (mechanistically, not courtesy):**
- The root cause is pinned exactly: composite `contains` was the single field-serialize site dropping `opts`, which filtered computeds out of nested resources and broke the `cardReferenceUrls` → `linkedCards` interpolation chain end to end. The fix is the minimal one that restores parity with the other three sites.
- Relativizing an included resource against its own id (`relativeReferenceFor` + `rebaseReferencesFor`) makes client serialization and the server's `relativizeResource` encode documents identically — a real correctness fix, not just the embed repair.
- The realm resolution is careful about both directions that matter: `realmForReference` places a reference by any spelling its realm is addressable under and folds base back to its virtual URL (where its index rows actually live), and the normalizer separates "realm authored into the query" (honored verbatim) from "realm drawn from instance data" (dropped when unplaceable) by provenance rather than trailing-slash spelling — the right axis, and the fix that resolved the earlier automated finding.
- Per-realm error isolation in the index-engine fan-out: a realm the caller can't read contributes its error and no results instead of sinking the whole field.

**Notes (detail in the inline threads):**
1. `rebaseReferencesFor` — confirmation that the `value.id`-present-vs-absent rule is the exact twin of `relativizeResource`'s `resource.id ? own : primary`; a guard against silent client/server drift. (`card-serialization.ts`)
2. Composite `contains` `opts` forwarding — confirmation the fix is safe because it matches the long-standing `containsMany` composite behavior; the `includeUnrenderedFields` fallout is the same shape that path already produced. (`card-api.gts`)
3. Provenance is tracked per whole `realms` value, so a literal realm in a *mixed* literal+interpolation list inherits interpolation provenance and can be dropped; no caller writes that shape today. Follow-up. (`query-field-utils.ts`)

**Adjacent, out of scope:** the index-engine fan-out issues one `_search` request per targeted realm (`fetchRemoteQueryResults` is called once per realm in the loop). For the markdown case references usually live in one or two realms, so it's immaterial; but if a single field ever spans many remote realms on the same server, the `/_search` endpoint already accepts a multi-realm `realms` param and could batch them. Not asked of this change.

---

habdelra (approved):

packages/base/card-serialization.ts
backspace: noted in 69038f61a6

packages/runtime-common/query-field-utils.ts
backspace: addressed in 96e12eda5a

PR #5719, opened 2026-08-07
Make the profile settings modal usable in a narrow dialog

burieberry:
Before & After (previously unusable due to sidebar):

settings-modal

### DIFF
--- a/packages/host/app/components/modal-container.gts
+++ b/packages/host/app/components/modal-container.gts
@@ -131,6 +131,11 @@ export default class ModalContainer extends Component<Signature> {
           'footer';
         grid-template-rows: auto 1fr auto;
         box-shadow: var(--boxel-deep-box-shadow);
+        /* container-type applies layout containment, which makes .dialog-box
+           the containing block for position: fixed descendants. A nested
+           modal rendered inside :content gets trapped by it (sized and
+           positioned against this box instead of the viewport), so nested
+           modals must portal out to the app root via in-element. */
         container-type: inline-size;
         container-name: dialog-box;
       }
--- a/packages/host/app/components/modal-container.gts
+++ b/packages/host/app/components/modal-container.gts
@@ -131,6 +131,8 @@ export default class ModalContainer extends Component<Signature> {
           'footer';
         grid-template-rows: auto 1fr auto;
         box-shadow: var(--boxel-deep-box-shadow);
+        container-type: inline-size;
+        container-name: dialog-box;
       }
 
       .dialog-box--with-sidebar {
@@ -138,7 +140,7 @@ export default class ModalContainer extends Component<Signature> {
           'sidebar-header header'
           'sidebar content'
           'sidebar footer';
-        grid-template-columns: 300px 1fr;
+        grid-template-columns: auto 1fr;
       }
 
       .dialog-box__sidebar-header {
@@ -149,8 +151,8 @@ export default class ModalContainer extends Component<Signature> {
 
       .dialog-box__sidebar {
         grid-area: sidebar;
+        width: 18.75rem;
         background-color: var(--boxel-100);
-
         border-bottom-left-radius: var(--boxel-border-radius);
       }
 
@@ -239,6 +241,21 @@ export default class ModalContainer extends Component<Signature> {
         border-bottom-left-radius: var(--boxel-border-radius);
         border-bottom-right-radius: var(--boxel-border-radius);
         display: flex;
+      }
+
+      @container dialog-box (width <= 48rem) {
+        .dialog-box--with-sidebar > .dialog-box__sidebar,
+        .dialog-box--with-sidebar > .dialog-box__sidebar-header {
+          display: none;
+        }
+        .dialog-box--with-sidebar > .dialog-box__content {
+          padding-inline: var(--boxel-sp);
+        }
+        .dialog-box--with-sidebar > .dialog-box__footer {
+          height: auto;
+          min-height: var(--stack-card-footer-height);
+          padding-inline: var(--boxel-sp-2xs);
+        }
       }
     </style>
   </template>
--- a/packages/host/app/components/operator-mode/profile/profile-settings-modal.gts
+++ b/packages/host/app/components/operator-mode/profile/profile-settings-modal.gts
@@ -64,7 +64,7 @@ export default class ProfileSettingsModal extends Component<Signature> {
         <ProfileInfo />
       </:sidebar>
       <:content>
-        <form {{on 'submit' this.onSubmit}}>
+        <form class='settings-form' {{on 'submit' this.onSubmit}}>
           {{#unless (bool this.submode)}}
             <FieldContainer @label='Name' @tag='label' class='profile-field'>
               <BoxelInput
@@ -88,7 +88,7 @@ export default class ProfileSettingsModal extends Component<Signature> {
             <FieldContainer
               @label='Current Password'
               @tag='label'
-              class='profile-field'
+              class='profile-field password-settings-field'
             >
               <BoxelInput
                 data-test-current-password-field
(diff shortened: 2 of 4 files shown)


packages/host/app/components/modal-container.gts
```diff
@@ -131,14 +131,16 @@ export default class ModalContainer extends Component<Signature> {
           'footer';
         grid-template-rows: auto 1fr auto;
         box-shadow: var(--boxel-deep-box-shadow);
+        container-type: inline-size;
+        container-name: dialog-box;
```
habdelra, line 135: [Claude Code 🤖]

**Blocking — this re-anchors any nested modal's `position: fixed` overlay to the dialog instead of the viewport.** `ModalContainer` is shared by ~15 modals, and the profile modal renders a *nested* modal inside itself, so this hits a reachable flow.

**The mechanism.** `container-type: inline-size` applies *layout containment* to `.dialog-box`. Per the CSS Containment spec, layout containment makes the element "a containing block for absolutely positioned **and fixed positioned** descendants" — the same gotcha as `transform`/`filter`/`will-change`. Note the distinction from `position`: CardContainer already gives `.dialog-box` `position: relative` (`card-container/index.gts:74`), but `relative` only captures *absolute* descendants, never *fixed* ones. So before this PR a fixed descendant escaped `.dialog-box` and resolved against the viewport; after it, a fixed descendant is trapped in `.dialog-box`.

**The concrete regression.** The email-change flow renders `PasswordModal` ("Confirm Identity") — `profile-email.gts:311` (`{{#if this.showPasswordModal}}`), reachable via Change Email → new email → Save. That modal is a plain DOM descendant of the outer `.dialog-box` (`Modal` renders inline — no `{{in-element}}` portal — and opens via the `open` attribute, not `showModal()`, so its `<dialog>` is *not* in the top layer and is subject to ancestor containment). Its `<dialog>` is `position: fixed; width: 100%; height: calc(100vh - offset)` with the backdrop as a `position:absolute; inset:0` child. Post-PR its containing block is `.dialog-box`, so: `width: 100%` becomes the dialog's width (not the screen), the backdrop covers only the dialog's box (not the page), and `100vh`/`calc((100vh - 21rem)/2)` still measure the viewport but lay out from the dialog's origin — so the "Confirm Identity" modal ends up narrow, mis-centered, and its dim overlay no longer covers the screen. CSS unit tests won't catch this; worth eyeballing on the preview deployment (Email → Change Email → Save).

**The way out.** The robust fix lives in the shared `Modal`: promote its `<dialog>` to the top layer via `showModal()`, or render it through an `{{in-element}}` portal to `document.body`, so nested modals are never captured by a container/containment ancestor. Scoping `container-type` to `.dialog-box--with-sidebar` would *not* help — that class is on this same outer element, and `PasswordModal` lives inside this modal's `:content`. Recommend verifying on the preview first; if it reproduces, the top-layer/portal change is the durable fix.

---

packages/host/app/components/modal-container.gts
```diff
@@ -131,14 +131,16 @@ export default class ModalContainer extends Component<Signature> {
...
       }
 
       .dialog-box--with-sidebar {
         grid-template-areas:
           'sidebar-header header'
           'sidebar content'
           'sidebar footer';
-        grid-template-columns: 300px 1fr;
+        grid-template-columns: auto 1fr;
```
habdelra, line 143: [Claude Code 🤖]

**Confirmation — this `300px` → `auto` change is what makes the sidebar collapse cleanly; keep it.** With `container-type` present and the sidebar `display:none`d in the `<= 48rem` query below, a fixed `300px` first track would have stayed 300px even with no item in it, leaving an empty 300px gutter and pushing content right. As an `auto` track with no grid item it collapses to 0, so content spans the full width — exactly what you want. The fixed sidebar width now lives on `.dialog-box__sidebar { width: 18.75rem }` (= the old 300px), so the visible-sidebar layout is unchanged. Good coupling — noting it so a future editor doesn't "simplify" it back to `300px`. Non-blocking.

---

packages/host/app/components/operator-mode/profile/profile-settings-modal.gts
```diff
@@ -218,9 +219,46 @@ export default class ProfileSettingsModal extends Component<Signature> {
...
+          --settings-modal-gap: var(--boxel-sp);
+        }
+        .password-settings-field {
+          grid-template-columns: 1fr;
+          gap: var(--boxel-sp-xs);
+        }
+        .profile-settings-change-pw-button {
+          margin-right: auto;
+          padding-inline: var(--boxel-sp-xs);
+          font-size: 0.8175rem;
```
habdelra, line 242: [Claude Code 🤖]

**Nit — `0.8175rem` is almost certainly a typo.** `0.8175rem` = `13.08px`, an off-the-grid value; the intended 13px step is `0.8125rem`. Since the type ramp only has tokens at 12px (`--boxel-font-size-xs`) and 14px (`--boxel-font-size-sm`) — neither is 13px — either round to a token (accepting the 1px shift) or drop it to `0.8125rem`. Per the component conventions this repo follows, a scale token is preferred over a raw literal when one fits. Non-blocking.

---

habdelra (commented):
[Claude Code 🤖]

Reviewed through the lens of **blast radius**: the visual improvements are scoped to the profile modal, but the enabling mechanism — `container-type` on `.dialog-box` — lives in the shared `ModalContainer`, so the review focused on what that containment does to *every* modal, not just this one.

**Bottom line: one blocking issue — the new `container-type: inline-size` on the shared `.dialog-box` turns it into a containing block for `position: fixed` descendants, which breaks the nested "Confirm Identity" `PasswordModal` in the email-change flow (it stops covering/centering on the viewport).** Everything else is solid; the responsive design is well thought through and the container-query approach is the right tool here.

**What lands right**
- **Named container queries across component boundaries.** Declaring `container-name: dialog-box` once and querying it from `profile-email`, `profile-subscription`, and `profile-settings-modal` is exactly how cross-component container queries are meant to work — the queries resolve against the runtime ancestor regardless of which scoped stylesheet declares them. Clean.
- **The `300px` → `auto` sidebar track** is what lets the sidebar collapse to zero width when `display:none`d, rather than leaving a fixed gutter. Noted inline so it survives future cleanup.
- **`--boxel-teal` → `--boxel-highlight`** on the payment-link icon color (`profile-subscription.gts`) is the correct direction per the repo's token conventions (semantic role token over the raw palette hue).
- **Spacing refactor** from `.profile-field + .profile-field { margin-top }` to a grid `gap` on `.settings-form` / `.profile-settings-subscription` is a genuine simplification, and it composes correctly since `ProfileEmail`/`ProfileSubscription` render their field containers as direct grid children.

**Recommendations**
1. **[Blocking]** Verify the email-change → "Confirm Identity" modal on the preview deployment; if it's clipped/mis-centered (it should be), fix the shared `Modal` to use the top layer (`showModal()`) or a `body` portal so nested modals escape the new containment — see the thread on `container-type` in `modal-container.gts`.
2. **[Non-blocking]** `0.8175rem` in `profile-settings-modal.gts` looks like a typo for `0.8125rem`; prefer a font-size token — see the inline thread.

**Adjacent, out of scope (not asked of this PR)**
- The stale comment `// Subscription sits below the fold in this 70vh modal` at the top of `profile-settings-modal.gts` — this PR bumped the modal to `80vh` (`:deep(.profile-settings) { height: 80vh }`), so the number is now wrong. A one-word fix if you're already in the file.
- The new `background-color: var(--boxel-100)` blocks (`profile-subscription.gts`, `<= 48rem` query) follow the surrounding code, but the repo's conventions steer component CSS toward semantic role tokens (`--muted`, etc.) over numbered-palette primitives, which don't theme for dark mode. Consistent with existing usage in these files, so purely a heads-up for whenever this modal gets a dark-mode pass.

---

PR #5722, opened 2026-08-07
host: Add shared indexing to some test modules

backspace:
This adds a `setupLocalIndexing` test helper that test modules can use to reuse an index across tests. Claude’s estimate is that this makes these modules 66% to 86% faster.


packages/host/tests/helpers/index.gts
```diff
@@ -941,10 +941,62 @@ class MockLocalIndexer extends Service {
...
+          snapshotName: opts.reuseIndexAcrossTests,
+          captured: false,
+          restored: false,
+        }
+      : undefined;
+  });
+
+  hooks.after(function () {
+    reusableIndex = undefined;
+  });
```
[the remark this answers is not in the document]
backspace: this is fixed in 03b2af49d0

packages/host/tests/helpers/index.gts
```diff
@@ -1345,6 +1420,14 @@ async function setupTestRealm({
...
   await worker.run();
   await realm.start();
+  if (reusableIndex && !reusableIndex.captured) {
+    // First test of a module that reuses its index: this start() just did the
+    // indexing, so keep the result for the rest of the module. Captured before
+    // the test body runs, so what later tests restore is the fixtures as
+    // indexed and nothing a test went on to write.
+    await dbAdapter.exportSnapshot(reusableIndex.snapshotName);
+    reusableIndex.captured = true;
+  }
```
richardhjtan, line 1430: [Claude Code 🤖] **The multi-realm limitation fails silently, and it fails in the shape that costs the most to debug: first test green, every later test reading an index that's missing a realm.** Recommend converting the doc-block warning into a thrown error here.

**Mechanism, step by step.** Take a module that opts in and builds two realms per test.

*Test 1.* `beforeEach` finds `captured === false`, so `reset()` runs and `restored` is set to `false`. Realm A's `setupTestRealm` therefore gets no `skipBootIndex`, indexes normally, reaches this block, exports the snapshot, and sets `captured = true`. Realm B's `setupTestRealm` then runs in the same test — `restored` is still `false`, so B also indexes normally, and this block is skipped because `captured` is now `true`. **The snapshot contains A only.** Test 1 passes, because the live database contains both.

*Test 2.* `beforeEach` finds `captured === true`, imports the snapshot, sets `restored = true`. Now *both* realms take the `skipBootIndex` branch at the `Realm` construction above. A is fine — its rows came back with the snapshot. B mounts and serves with **nothing in the index for it**: `skipBootIndex` is precisely "no from-scratch index, even on a new index" (`packages/runtime-common/realm.ts`).

**Why that's worse than a crash.** B's cards aren't broken, they're absent. A `_search` against B returns zero results; an assertion that a card *doesn't* appear passes; a card-chooser list renders short. Every one of those reads as correct behavior. And the first test still passes, so the module doesn't look newly broken — it looks like tests 2..n developed an unrelated flake. Whoever picks that up has no reason to suspect the index, because the module's own fixtures are right there in the file.

**The check is cheap because the signal is already here.** `captured` being true while a realm is being built inside the *same* test is exactly the "second realm" condition. Tracking the current test's realm count alongside the flag is enough:

```ts
if (reusableIndex) {
  if (!reusableIndex.captured) {
    await dbAdapter.exportSnapshot(reusableIndex.snapshotName);
    reusableIndex.captured = true;
    reusableIndex.realmsThisTest = 1;
  } else if (reusableIndex.restored === false) {
    // Same test that captured the snapshot is building another realm: the
    // snapshot was taken before this one indexed, so later tests would
    // restore an index that is missing it.
    throw new Error(
      `reuseIndexAcrossTests ('${reusableIndex.snapshotName}') does not support a module that builds more than one realm per test`,
    );
  } else {
    reusableIndex.realmsThisTest = (reusableIndex.realmsThisTest ?? 0) + 1;
    if (reusableIndex.realmsThisTest > 1) {
      throw new Error(/* same message */);
    }
  }
}
```

The exact bookkeeping is yours — a per-test counter reset in the `beforeEach` alongside `restored` is probably tidier than branching on `restored` here. The point is that the condition is locally detectable, so the failure can be a message naming the option instead of a search-result count that's quietly short.

**Verified this isn't currently violated:** all seven opted-in modules have exactly one realm-setup call — `ai-assistant-test.gts` and `searchable-search-doc-test.gts` one each, and the five operator-mode modules share the single `setupIntegrationTestRealm` call inside `setupOperatorModeTests`. So this is about the next module to opt in, not about anything in this diff.

Scope: follow-up, non-blocking. Test infrastructure, but the blast radius is "a module silently stops testing what it says it tests", which is why I'd rather it were enforced than documented.

packages/host/tests/helpers/index.gts
```diff
@@ -941,10 +941,82 @@ class MockLocalIndexer extends Service {
...
+    if (reusableIndex?.captured) {
+      // Replaces every table from the snapshot, so this subsumes reset().
+      await dbAdapter.importSnapshot(reusableIndex.snapshotName);
+      reusableIndex.restored = true;
+    } else {
+      await dbAdapter.reset();
+      if (reusableIndex) {
+        reusableIndex.restored = false;
+      }
+    }
```
richardhjtan, line 1019: [Claude Code 🤖] **Confirmation — "this subsumes reset()" is true, and it wasn't obvious enough to take on faith.** Nothing to change. Recording the verification because the two operations select their table lists from different places, and a mismatch would leak state between tests in a way no test would report.

The two table sets:

- `SQLiteAdapter#reset()` iterates `this.tables`, populated by `#loadSchemaMetadata` from `SELECT name FROM pragma_table_list WHERE schema = 'main' AND name != 'sqlite_schema'`.
- `importSnapshot` iterates `SELECT name FROM <snapshot>.sqlite_schema WHERE type = 'table' AND name NOT LIKE 'sqlite_%'`, and per table issues `DELETE FROM main.<table>` followed by a copy — so anything in `main` but *not* in the snapshot is never cleared.

So subsumption needs the snapshot's table set to cover `this.tables`. It does: the snapshot is taken from `main`'s own `sqlite_schema` with the schema unchanged in between, so the sets agree except for names matching `sqlite_%`. The only such table SQLite would create on its own is `sqlite_sequence`, and it exists only with an `AUTOINCREMENT` column — `packages/host/config/schema/*.sql` has none, so the table doesn't exist and the exclusion costs nothing. (If a future migration introduces `AUTOINCREMENT`, this comment stops being true: restored tests would inherit the previous test's id counter instead of resetting it, and a test asserting a concrete id would start failing from the second test onward.)

**What the restore carries is broader than the name suggests**, and it's where the 66–86% comes from: the snapshot is every table in `main`, so `module_transpile_cache`, `modules`, and `prerendered_html` all come back — the transpile and prerender work, not just `boxel_index` rows. It's also transactional (`BEGIN IMMEDIATE` … `COMMIT`, with a rollback on failure), so a failed restore can't leave a half-replaced database for the test to run against.

packages/host/tests/integration/components/operator-mode-test.gts
```diff
@@ -23,7 +23,9 @@ import { setupOperatorModeTests } from './operator-mode/setup';
 import type { CardDef } from '@cardstack/base/card-api';
 
 module('Integration | operator-mode | basics', function (hooks) {
-  let ctx = setupOperatorModeTests(hooks);
+  let ctx = setupOperatorModeTests(hooks, {
+    reuseIndexAcrossTests: 'operatorMode',
```
richardhjtan, line 27: [Claude Code 🤖] **This is the module in the set that sits closest to the "must not care how its index got there" line, and it's worth saying why it still clears it.** No change requested — this is a guard rail for whoever next edits either this module or the reuse helper.

The nested `module('card with an error that has a last known good state')` has a `beforeEach` that rewrites `ExplodingCard/1.json` with `status: 'boom'` so the `cardTitle` compute throws on the next index, and the test then asserts the *previous* clean render is what shows: `.includesText('Stable Example', 'the last known good HTML is rendered')`. Its comment names the dependency outright — "captured by the prior clean indexing pass".

Under `reuseIndexAcrossTests` that prior clean pass no longer happens in this test. The realm mounts with `skipBootIndex`, so the last-known-good HTML has to arrive via the restore instead. It does: `prerendered_html` is a table in `main` (`packages/host/config/schema/*.sql`), so `exportSnapshot` copies it and `importSnapshot` replaces it — the clean render captured during the first test's boot index is what this test's write then supersedes. Same bytes, different provenance.

**What would break it**, so it isn't rediscovered the hard way:

- Narrowing the snapshot to `boxel_index` (a tempting optimization — it's the big table). The last-known-good assertion would fail with nothing to point at the change, because the fixture and the test would both still look correct.
- Reordering so this nested module's test runs first in the module. Then it *is* the capturing test, `restored` is false, the boot index runs live, and the flip happens against a freshly indexed realm — it passes either way, which is the problem: the reuse path stops being exercised while the assertion still reads as covering it.

The more general point: "identical fixtures" was the stated precondition, and this test shows it isn't the whole test — a test can also depend on index *history* rather than index *contents*. The doc block already says as much (`Integration | Store` is the worked counter-example); this module is the case where the dependency exists and the restore happens to satisfy it.

packages/host/tests/helpers/index.gts
backspace: this is addressed in 12806fedcb but on #5725 because that’s where multi-realm support was added

packages/host/tests/helpers/index.gts
backspace: the comment is updated in f899ced035 but on #5725

richardhjtan (approved):

PR #5724, opened 2026-08-07
Publishing action appears to have failed even though publish has succeeded

burieberry:
The error:
err-1

err-2

A published realm could report `X-Boxel-Not-Ready: prerender-html` forever while serving fully rendered HTML: readiness gated on a realm-wide generation watermark that no write is obligated to satisfy, so post-publish polls ran to their full timeout and surfaced as an unattributable `fetch failed`.

### Changes:
- Check each live, non-errored index row against its own generation instead. Readiness now waits for every row to have matching HTML at or beyond that generation, while preserving the established handling of current-generation tombstones and render errors as completed outcomes.
- Improve waitForReady timeout diagnostics with the outstanding stage and transport cause, and add focused PostgreSQL and SQLite coverage for the per-row readiness cases.

Fixes CS-12435.

🤖 Generated with [Claude Code](https://claude.com/claude-code) & Codex

### DIFF
--- a/packages/realm-server/tests/await-realm-index-settled-test.ts
+++ b/packages/realm-server/tests/await-realm-index-settled-test.ts
@@ -6,7 +6,10 @@ import {
   awaitRealmIndexSettled,
   indexingConcurrencyGroup,
 } from '@cardstack/runtime-common/jobs/indexing';
-import { prerenderHtmlConcurrencyGroup } from '@cardstack/runtime-common/jobs/prerender-html';
+import {
+  awaitPublishedHtmlReady,
+  prerenderHtmlConcurrencyGroup,
+} from '@cardstack/runtime-common/jobs/prerender-html';
 import { setupDB } from './helpers/index.ts';
 
 const realmURL = 'http://localhost:4201/test/';
@@ -32,6 +35,60 @@ async function enqueueIndexJob(
     },
   )) as { id: number }[];
   return id;
+}
+
+// A live `boxel_index` row at `generation`, or a tombstone / index-errored one.
+// Sets only the columns the HTML gate reads.
+async function insertIndexRow(
+  dbAdapter: PgAdapter,
+  realm: string,
+  name: string,
+  generation: number,
+  opts?: { isDeleted?: boolean; hasError?: boolean },
+): Promise<void> {
+  await dbAdapter.execute(
+    `INSERT INTO boxel_index
+       (url, file_alias, realm_url, type, generation, is_deleted, has_error, error_doc)
+     VALUES ($1, $2, $3, 'instance', $4, $5, $6, $7)`,
+    {
+      bind: [
+        `${realm}${name}`,
+        `${realm}${name}`,
+        realm,
+        generation,
+        opts?.isDeleted ?? false,
+        opts?.hasError ?? false,
+        opts?.hasError ? JSON.stringify({ message: 'boom' }) : null,
+      ],
+    },
+  );
+}
+
+// The matching `prerendered_html` row, whose `generation` the gate compares
+// against the index row's own. A tombstone (`isDeleted`) carries no HTML,
+// matching what `tombstonePrerenderedHtmlEntries` seeds.
+async function insertHtmlRow(
+  dbAdapter: PgAdapter,
+  realm: string,
+  name: string,
+  generation: number,
+  opts?: { isDeleted?: boolean },
+): Promise<void> {
+  await dbAdapter.execute(
+    `INSERT INTO prerendered_html
+       (url, file_alias, realm_url, type, generation, is_deleted, isolated_html)
+     VALUES ($1, $2, $3, 'instance', $4, $5, $6)`,
+    {
+      bind: [
+        `${realm}${name}`,
+        `${realm}${name}`,
+        realm,
+        generation,
+        opts?.isDeleted ?? false,
+        opts?.isDeleted ? null : '<div>rendered</div>',
+      ],
+    },
+  );
 }
 
 module(basename(import.meta.filename), function (hooks) {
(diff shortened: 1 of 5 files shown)


packages/runtime-common/jobs/prerender-html.ts
```diff
@@ -66,57 +66,68 @@ export function prerenderHtmlConcurrencyGroup(realmURL: string): string {
...
-      'AND generation >=',
-      param(currentGeneration),
-      'LIMIT 1',
+      `  AND (i.is_deleted = false OR i.is_deleted IS NULL)
+          AND i.has_error = false
+          AND i.error_doc IS NULL
+          AND (ph.url IS NULL
+            OR ph.generation < i.generation
+            OR ph.is_deleted = true)
+        LIMIT 1`,
```
habdelra, line 123: [Claude Code 🤖]

**The gate holds on a state the sweep no longer repairs — this re-opens the "readiness never passes, no path to recovery" failure mode this change exists to fix, for tombstone residue.**

`hasCaughtUp` counts a live index row whose `prerendered_html` row is a surviving tombstone (`ph.is_deleted = true`) as *not* caught up, so readiness holds on it. But the reconcile sweep's `findStalePrerenderedHtmlRows` (`prerender-html-reconcile.ts`, unchanged here) has no arm that selects such a row: `ph.url` is non-null, `ph.generation` is not `< i.generation`, and a plain tombstone carries no `visitRequestFailure` `error_doc`, so none of its three arms match. Nothing the scheduled sweep does will ever clear it.

**How the state arises.** `visitForPrerenderedHtml` (`index-runner/prerender-html-visit.ts`) returns without writing when the file is unreadable, so the up-front tombstone survives the swap at the row's own generation. If that URL's `boxel_index` row is still live — file readable at index time, unreadable at the later prerender visit, deletion not yet indexed — you get a live index row paired with a `prerendered_html` tombstone at the current generation. Readiness then returns `X-Boxel-Not-Ready: prerender-html` until a *future index invalidation* re-touches that URL. The scheduled reconcile sweep is not that recovery path — despite the comment on line 111 here ("while the sweep spaces its retries of the same rows") and the one in `readinessCheck` (realm.ts: "A stuck rendering is repaired by the scheduled reconciliation sweep"), both of which describe a sweep behavior that isn't present.

**Fix — make the gate and the sweep agree, one of:**
- Drop the `ph.is_deleted = true` arm so a tombstone-over-live row reads as caught up. This matches how the gate already excludes index-errored rows (no card to render → don't hang on it), and keeps readiness from ever blocking on residue nothing is scheduled to repair. A card mid-deletion briefly reporting "ready" is the same downstream-visible outcome (missing markup, not a hang) as an index-error row.
- Or restore the sweep's tombstone-repair arm so the gate's hold has a recovery path.

Either way, the two comments claiming the sweep retries these rows should match what the sweep does.

**Scope:** regression, non-blocking. Non-blocking only because it needs a genuine index-vs-prerender race and self-heals on the next re-index of that URL — but it recreates this change's own symptom for that window and ships with an incorrect comment, so worth resolving before this lands.

---

packages/realm-server/tests/await-realm-index-settled-test.ts
```diff
@@ -165,4 +222,146 @@ module(basename(import.meta.filename), function (hooks) {
...
+        }),
+        'the unrendered sibling still holds the gate',
+      );
+    });
+
+    // The render visit found the file unreadable and wrote nothing, so the
+    // up-front tombstone survived the swap at the row's own generation. Its
+    // generation is current, but the row serves no markup — counting it as
+    // rendered would report ready for a card without HTML.
+    test('a prerendered_html tombstone at the row generation holds the gate', async function (assert) {
```
habdelra: [Claude Code 🤖]

This pins that the gate *holds* on a tombstone-over-live row, but nothing here proves the state can ever *clear* — contrast `the gate releases when the render lands` below, which exercises recovery for the missing-HTML case. As written the coverage reads as "confirmed correct," while the row it sets up is one the reconcile sweep has no arm to repair (see the note on `hasCaughtUp` in `prerender-html.ts`).

If the gate arm stays, a test that shows what recovers this state would make the recovery contract executable rather than implied. If the arm is dropped, this test should flip to asserting the row reads as caught up.

**Scope:** test coverage, non-blocking — tied to the `prerender-html.ts` finding.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** whether the new per-row readiness gate is *satisfiable and recoverable* — the property the old realm-watermark gate lacked and that this change exists to restore.

**Bottom line:** The core change is right and well-argued. Gating on the same per-row predicate the read path and the reconcile sweep use is satisfiable by construction, strictly stronger than the old `LIMIT 1` (which let one rendered row vouch for a whole unrendered realm), and the watermark removal is correct — an index batch advances `current_generation` unconditionally while a prerender batch never does, so a realm can be fully rendered with the watermark sitting above every rendered row. The per-request budget and the `waitForReady` cause/stage folding directly kill the unattributable `fetch failed`. **One issue re-opens this change's own failure mode in a narrow case:** the gate holds on a `prerendered_html` tombstone over a live row, but the reconcile sweep — left unchanged — has no arm that repairs it, so readiness can block on residue only a future re-index clears. Non-blocking (rare race, self-heals on re-index), but it ships with a comment saying the sweep handles it. Detail inline.

**What lands right:**
- The watermark diagnosis is correct, and the new predicate mirrors the proven sweep on the live-row + generation arms. Because `boxel_index` no longer carries `module` rows (constrained to `instance`/`file`), the `ph.type = i.type` join lines up with `prerendered_html`'s own types, so the anti-join can't spuriously flag a row with no possible HTML counterpart.
- The 60s→10s in-request budget is the real fix for the ALB-idle-timeout reset: a request held to 60s sat exactly on the default idle timeout and reached the client as `fetch failed`. Racing `#startedUp` / `indexing()` against the shared budget is sound — `Promise.race` attaches a rejection reaction to the raced promise, so a late rejection after the timer fires doesn't leak as an unhandled rejection, and behavior on an early rejection is unchanged.
- The `rechecking` coalescing guard is correct and symmetric across both helpers; the ≤1s poll covers any `jobs_finished` notify dropped while a recheck is in flight.

**Recommendations:**
1. Make the gate and the sweep agree on tombstone-over-live rows (drop the gate arm, or restore the sweep's repair arm), and fix the two comments that claim the sweep retries them — see the `hasCaughtUp` thread.
2. Add/adjust a recovery test for the tombstone case — see the test thread.
3. The PR description still describes the readiness-kicks-reconcile behavior, min-age tombstone spacing, and a shared `unrenderedLiveIndexRowWhere` predicate — all of which the current head walks back. Worth updating so a reader or merger doesn't count on a recovery path that isn't shipped.

No blocking issues; #1 is the one I'd resolve before merge.

---

packages/runtime-common/jobs/prerender-html.ts
burieberry: Made changes to this PR and got rid large swaths of changes to keep the PR focus minimal. Tombstone-repair behavior is restored.

packages/realm-server/tests/await-realm-index-settled-test.ts
burieberry: "Arm" has been dropped and row now reads as caught up. Removed `ph.is_deleted = true` from the readiness-blocking predicate.

burieberry:
> [Claude Code 🤖]
> 
> **Lens:** whether the new per-row readiness gate is _satisfiable and recoverable_ — the property the old realm-watermark gate lacked and that this change exists to restore.
> 
> **Bottom line:** The core change is right and well-argued. Gating on the same per-row predicate the read path and the reconcile sweep use is satisfiable by construction, strictly stronger than the old `LIMIT 1` (which let one rendered row vouch for a whole unrendered realm), and the watermark removal is correct — an index batch advances `current_generation` unconditionally while a prerender batch never does, so a realm can be fully rendered with the watermark sitting above every rendered row. The per-request budget and the `waitForReady` cause/stage folding directly kill the unattributable `fetch failed`. **One issue re-opens this change's own failure mode in a narrow case:** the gate holds on a `prerendered_html` tombstone over a live row, but the reconcile sweep — left unchanged — has no arm that repairs it, so readiness can block on residue only a future re-index clears. Non-blocking (rare race, self-heals on re-index), but it ships with a comment saying the sweep handles it. Detail inline.
> 
> **What lands right:**
> 
> * The watermark diagnosis is correct, and the new predicate mirrors the proven sweep on the live-row + generation arms. Because `boxel_index` no longer carries `module` rows (constrained to `instance`/`file`), the `ph.type = i.type` join lines up with `prerendered_html`'s own types, so the anti-join can't spuriously flag a row with no possible HTML counterpart.
> * The 60s→10s in-request budget is the real fix for the ALB-idle-timeout reset: a request held to 60s sat exactly on the default idle timeout and reached the client as `fetch failed`. Racing `#startedUp` / `indexing()` against the shared budget is sound — `Promise.race` attaches a rejection reaction to the raced promise, so a late rejection after the timer fires doesn't leak as an unhandled rejection, and behavior on an early rejection is unchanged.
> * The `rechecking` coalescing guard is correct and symmetric across both helpers; the ≤1s poll covers any `jobs_finished` notify dropped while a recheck is in flight.
> 
> **Recommendations:**
> 
> 1. Make the gate and the sweep agree on tombstone-over-live rows (drop the gate arm, or restore the sweep's repair arm), and fix the two comments that claim the sweep retries them — see the `hasCaughtUp` thread.
> 2. Add/adjust a recovery test for the tombstone case — see the test thread.
> 3. The PR description still describes the readiness-kicks-reconcile behavior, min-age tombstone spacing, and a shared `unrenderedLiveIndexRowWhere` predicate — all of which the current head walks back. Worth updating so a reader or merger doesn't count on a recovery path that isn't shipped.
> 
> No blocking issues; #1 is the one I'd resolve before merge.
> 
> _Generated by [Claude Code](https://claude.ai/code)_


  **- Keep `waitForReady` stage/cause diagnostics:** Already present. It makes failures attributable and is low risk.
  
  **Dropped changes since review:**
  **- Rechecking coalescing guard:** Worth considering as a small performance hardening. The new query can inspect every live row, and jobs_finished is deployment-wide, so preventing overlapping checks avoids unnecessary database load. It is not required to solve CS-12435 and could be a follow-up. If restored, the previous implementation was sound because the one-second poll catches notifications received during an active query.

  **- 10-second whole-request budget:** Do not reintroduce the earlier version. It raced startup and in-process indexing, but merely passed the remaining time into helpers whose initial database query occurred before their timer. It therefore did not actually guarantee a 10-second request. Reintroduce this only as a separate, complete change that races the entire database helper calls—or uses database statement cancellation—and has blocked-query tests.

  **- Tombstone gate and reconciliation arm:** Do not reintroduce. The current branch intentionally preserves origin/main semantics: a same-generation tombstone is a completed outcome. Making it block readiness expands the ticket’s behavior and requires a retry/backoff contract to avoid permanent blocking or repeated rendering.

  **- On-demand reconciliation from readiness:** Do not reintroduce here. Keeping a public, frequently polled readiness endpoint read-only is cleaner. If faster repair is needed, trigger reconciliation once from the publishing workflow instead.

  **Recommendation:** ship the current branch unchanged. If database query pileup is a demonstrated concern, add the rechecking guard in a small follow-up; handle the request deadline as a separate, fully tested change.
  
  _Generated by Codex_

PR #5725, opened 2026-08-07
host: Add multi-realm support to in-test index cache

backspace:
This is an extension of #5722, as some modules start more than one realm.

### DIFF
--- a/packages/host/tests/helpers/index.gts
+++ b/packages/host/tests/helpers/index.gts
@@ -991,7 +991,8 @@ export function setupLocalIndexing(
     // indexes for itself exactly as it did before opting in. Calling it too
     // early is the failure that matters, which is why it belongs immediately
     // after the last realm rather than being inferred from a realm count kept
-    // somewhere else.
+    // somewhere else — and why building a realm after the capture throws (see
+    // setupTestRealm) instead of quietly serving that realm unindexed.
     captureIndexManually?: boolean;
   },
 ) {
@@ -1395,6 +1396,25 @@ async function setupTestRealm({
     definitionLookup = owner.lookup(
       'definition-lookup:main',
     ) as DefinitionLookup;
+  }
+
+  // A realm built in the same test that captured the snapshot cannot be in it —
+  // the capture already happened. Later tests would restore an index missing
+  // this realm and, because they also skip the boot index, serve it with
+  // nothing indexed at all. That reads as "these cards don't exist": a search
+  // returns nothing, an assertion that a card is absent passes, a list renders
+  // short. Nothing looks broken, and the first test still passes, because the
+  // live database did hold both. So fail here, on the first test, naming what
+  // to do about it. `restored` distinguishes the capturing test (false) from
+  // every later one (true, where realms are expected to come from the
+  // snapshot).
+  if (reusableIndex?.captured && !reusableIndex.restored) {
+    throw new Error(
+      `reuseIndexAcrossTests ('${reusableIndex.snapshotName}'): a realm is being built after this module's index snapshot was captured, so the snapshot cannot contain it. ` +
+        (reusableIndex.manual
+          ? 'captureReusableIndex() ran too early — move it after the last realm this module builds.'
+          : 'This module builds more than one realm per test: pass captureIndexManually and call captureReusableIndex() once, after the last realm.'),
+    );
   }
   await insertPermissions(dbAdapter, new URL(realmURL), permissions);
   let worker = new Worker({
--- a/packages/host/tests/helpers/index.gts
+++ b/packages/host/tests/helpers/index.gts
@@ -1025,7 +1025,16 @@ export function setupLocalIndexing(
   hooks.beforeEach(async function () {
     let dbAdapter = await getDbAdapter();
     if (reusableIndex?.captured) {
-      // Replaces every table from the snapshot, so this subsumes reset().
+      // Subsumes reset(), which is a precondition rather than a property:
+      // importSnapshot DELETEs and refills every table it finds in the
+      // *snapshot*, so it covers reset()'s list only because the snapshot was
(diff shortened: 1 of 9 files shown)


packages/host/tests/helpers/index.gts
```diff
@@ -1360,6 +1406,25 @@ async function setupTestRealm({
...
+  // every later one (true, where realms are expected to come from the
+  // snapshot).
+  if (reusableIndex?.captured && !reusableIndex.restored) {
+    throw new Error(
+      `reuseIndexAcrossTests ('${reusableIndex.snapshotName}'): a realm is being built after this module's index snapshot was captured, so the snapshot cannot contain it. ` +
+        (reusableIndex.manual
+          ? 'captureReusableIndex() ran too early — move it after the last realm this module builds.'
+          : 'This module builds more than one realm per test: pass captureIndexManually and call captureReusableIndex() once, after the last realm.'),
+    );
+  }
```
habdelra: [Claude Code 🤖]

**Confirmation, with one residual blind spot worth recording — non-blocking, not introduced here.**

This guard is the right shape and it fires where it's most useful: on the *capturing* test (`restored === false`), the moment a realm is built after the snapshot was taken. I traced both misuse modes and both land here on the first test with the correct message — auto mode capturing after realm #1 then meeting realm #2, and manual mode where `captureReusableIndex()` ran before the last realm.

**The case it can't see.** A realm built for the first time in a *later* test's body — a test that restored, so `restored === true`. The condition is `captured && !restored`, so that build passes the guard; and because `skipBootIndex` keys off the same module-global `restored` flag (`...(reusableIndex?.restored ? { skipBootIndex: true } : {})` in the `new Realm(...)` call below), the new realm mounts with nothing indexed and is served empty — the exact "a search returns nothing / a list renders short" failure this block's comment describes, just reached by a path the guard doesn't cover. The guard tracks *whether* this test restored, not *which* realms the snapshot actually contains, so a realm that only some tests build is invisible to it.

**Why it's fine for this PR.** The `skipBootIndex` behaviour predates this change, so this isn't a regression, and no opted-in module builds a realm in a test body: the `setupInteractSubmodeTests` consumers build all four realms in the shared setup, and the two interact-submode modules that do build per-test realms (`create-file`, `search-sheet-files`) are correctly left un-opted-in. The guard is a strict improvement either way.

Two small things for whoever extends this next:

- The manual-mode message ("`captureReusableIndex()` ran too early — move it after the last realm") assumes a misplaced capture is the only way to reach a post-capture realm. A realm built in a test body is another way in, and there the advice doesn't apply — the fix is "don't build it here / don't opt this module in." Not worth branching the message for today; just noting the assumption.
- If this ever needs to be total, the discriminator would have to be the set of realm URLs the snapshot contains rather than the boolean `restored`.

No change required.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** correctness of the multi-realm capture handoff and the new build-after-capture guard, plus whether the strong mechanistic claims in the comments hold against the snapshot code they describe.

**Bottom line: no blocking issues.** This is a clean, well-reasoned extension of the single-realm reuse path, and the guard added along the way is a real safety improvement. COMMENT-only — approve/request-changes is yours.

- The auto/manual split is mutually exclusive by construction. `setupTestRealm`'s auto-capture is gated on `!reusableIndex.manual` and `captureReusableIndex()` on `reusableIndex.manual`, so a manual module never double-captures. I traced the full state machine across test 1 (reset → build four realms → capture) and test 2+ (import → build four realms with `skipBootIndex` → capture is a no-op); it holds, and the throw fires on the first test in both misuse modes.
- Handing capture timing to the caller rather than inferring it from a realm count is the right call, for the reason the comment gives: QUnit runs `beforeEach` hooks in registration order, `setupLocalIndexing` is registered before the module's realm-building `beforeEach`, so no harness hook runs after the last realm. Putting `captureReusableIndex()` in the shared helper that builds all four realms is where it can't drift out of step with their number.
- Passing `captureIndexManually: true` unconditionally in `setupInteractSubmodeTests` is safe when a consumer omits the name: `reuseIndexAcrossTests: undefined` makes `reusableIndex` undefined in `hooks.before`, so `captureReusableIndex()` early-returns and the module indexes per test exactly as before. Confirmed against the four consumers that don't pass a name via a different setup and the ones that do.

**Doc claim I checked because it's load-bearing.** The expanded `beforeEach` comment says the `sqlite_%` exclusion is free "while nothing in config/schema is AUTOINCREMENT." Confirmed: `packages/postgres/scripts/convert-to-sqlite.ts` — which generates the test DB's schema — never emits `AUTOINCREMENT`, only bare `PRIMARY KEY`, so `sqlite_sequence` never exists and the exclusion drops nothing. The "a migration adding one would leave restored tests inheriting the previous test's id counter" caveat is the correct standing warning.

**Also confirmed harmless:** `insertPermissions` re-running on every restored test across all four realms is fine — `insertPermission` is an `upsert` on `realm_user_permissions_pkey`, not a plain insert, so restored tests don't accumulate duplicate rows.

**One non-blocking observation** on the guard's blind spot — inline on the `captured && !restored` throw in `packages/host/tests/helpers/index.gts`. In short: it catches a realm built after capture in the *capturing* test, but not a realm first built in a *later* restored test's body, which `skipBootIndex` would then serve empty. No current consumer hits it; recorded for the next editor.

**Adjacent, out of scope:** the `Add empty commit` (dfcf2ba5) reads as a CI re-trigger; worth dropping before merge if you're not squashing, otherwise ignore.

Nice work — the comments carry their own mechanism, which is what made this fast to verify.

---

habdelra (approved):

packages/host/tests/helpers/index.gts
backspace: addressed in 5380195

PR #5726, opened 2026-08-08
Fix bug where reading a card also writes it

backspace:
Lazily resolving a `linksTo` assigned the loaded relationship through the field setter, which marked the card dirty and autosaved it, so just reading a card could result in it being rewritten with a version increase and reindex. This fixes that and adds a test.


packages/host/tests/integration/store-test.gts
```diff
@@ -1828,6 +1621,49 @@ module('Integration | Store', function (hooks) {
...
+  // the field, which notifies change subscribers — the same signal a user edit
+  // produces — so without care the mere act of viewing a card can dirty it and
+  // trigger an auto-save. That write is invisible to the user, bumps the
+  // instance's version, and schedules a reindex.
+  test<TestContextWithSave>('loading a card and resolving its linksTo issues no save', async function (assert) {
+    // Write the link unresolved on disk so loading takes the lazy link-load
+    // path (rather than the target already being in the identity map).
+    await testRealm.write(
+      'Person/hassan.json',
+      JSON.stringify({
```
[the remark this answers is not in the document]
backspace: fixed in 17f8f95

packages/base/card-api.gts
```diff
@@ -3710,7 +3710,10 @@ function lazilyLoadLink(
           value[index] = fieldValue;
         }
       } else {
-        (instance as any)[field.name] = fieldValue;
+        // Not `instance[field.name] = fieldValue`: assigning through the field
+        // setter announces a change, which marks the instance dirty and
+        // auto-saves it — a write triggered by reading the card.
+        setResolvedField(instance, field, fieldValue);
```
habdelra: [Claude Code 🤖]

**The singular fix is correct, but it lands on one of four sibling lazy-resolve paths — the other three still write the card back on read.**

`lazilyLoadLink` writes a resolved value onto the instance in four places, and every one except the `else` branch changed here still funnels through `notifySubscribers` → the store's `onInstanceUpdated` (`packages/host/app/services/store.ts:2385`, which sets `hasUnsavedChanges` and calls `doAutoSave`) — the exact read-triggers-write chain this PR removes:

- **Singular success** (the `setResolvedField` call here) — fixed ✅
- **Plural `linksToMany` success** (`value[index] = fieldValue`, the `if (pluralArgs)` branch just above) — the assignment goes through the `WatchedArray` proxy's `set` trap (`packages/base/watched-array.ts:69`), which unconditionally schedules its subscriber for any non-`length` index write; that subscriber calls `notifySubscribers(instance, …)` (`card-api.gts:880`). So reading a `linksToMany` whose slots weren't side-loaded rewrites the card, same bug as the singular case.
- **Singular failure** (`notifySubscribers(instance, field.name, sentinel)` in the catch block) and **plural failure** (`value[index] = sentinel` through the same proxy) — planting a `link-error`/`link-not-found` sentinel also notifies. Viewing a card whose link target was deleted/404s therefore schedules a save on read too. The comment there justifies the notify as letting subscribers "observe … a failed load," but the only change-subscriber that reacts to a link-field notification is auto-save: `listenForCardId` in the recent-cards / spec-panel / playground services all early-return unless `field === 'id'`. So the notify serves no consumer on this path except the one you're suppressing.

**Verified by trace (not execution):** each hop above is unconditional — I read the `WatchedArray` set trap, the subscriber wiring, and `onInstanceUpdated`; `doAutoSave` short-circuits only for the render store, not the live-app read path. The three siblings predate this PR (blame: 2026-07-22), so none is a regression you introduced.

**Way out:** the `linksToMany` getter already entangles card tracking (`entangleWithCardTracking(instance)` at `card-api.gts:1738`), so you can swap the not-loaded slot in the raw backing array via the `rawValues` escape hatch (`rawArrayValues(value)[index] = fieldValue`) and call `notifyCardTracking(instance)` directly — re-render survives, and the proxy `set` trap (with its notify) is bypassed — mirroring what `setResolvedField` does for the singular case. `applySubscribersToInstanceValue` isn't needed for links, since `subscribeToChanges` only recurses into `contains`/`containsMany` (lines 3463 / 3509).

**Scope:** non-blocking — the singular fix is right as written. But the `linksToMany` case is common, and the PR title reads as a complete fix. I'd either extend it to the plural success path here (low-risk, same technique) or file a follow-up covering all three siblings, and narrow the title/description to "a `linksTo`."

---

packages/base/card-api.gts
```diff
@@ -4681,6 +4684,18 @@ function makeDescriptor<
...
+// loaded target of a link the instance already referenced. Change subscribers
+// drive auto-save, so announcing this as a change writes the card back to the
+// server merely because it was read, bumping its version and scheduling a
+// reindex. Glimmer's tracking is still notified, so rendering updates.
+function setResolvedField(instance: BaseDef, field: Field, value: any) {
+  propagateRealmContext(value, instance);
+  value = field.validate(instance, value);
+  getDataBucket(instance).set(field.name, value);
+  notifyCardTracking(instance);
+}
```
habdelra, line 4713: [Claude Code 🤖]

**Confirmation, and a guard rail for the next editor: this is `setField` (just below) minus exactly one line, and dropping that line is safe.** The sole difference is the omitted `notifySubscribers`. That's the right cut:

- `setField` for a singular link does no `applySubscribersToInstanceValue` wiring (that only runs from the `WatchedArray` subscribers for `containsMany`/`linksToMany`), so nothing subscription-related is lost by skipping it.
- `notifyCardTracking` still fires, and the field getter reads `cardTracking.get(instance)` (`field-support.ts:162`), so the resolved link re-renders.
- The only consumer that reacts to a link-field change notification is the store's auto-save — precisely what must not run on a read.

CI is green with the notify removed, which confirms no existing test relied on a lazy `linksTo` resolution announcing a change. Non-blocking; nothing to change here.

---

packages/host/tests/integration/store-test.gts
```diff
@@ -1828,6 +1828,95 @@ module('Integration | Store', function (hooks) {
...
+
+      await storeService.flush();
+      await settled();
+
+      assert.strictEqual(
+        instance.bestFriend?.name,
+        'Ghost',
+        'the lazy load resolved the target and assigned it to the field',
+      );
+      assert.deepEqual(saved, [], 'no save was issued while loading');
```
habdelra, line 1914: [Claude Code 🤖]

**Solid regression test for the path it covers, and the honesty guards are the right call.** Asserting `intercepted > 0` and `bestFriend === undefined` before the flush is what makes this test meaningful rather than shape-confirming: without them a side-loaded `included` would resolve the link synchronously and the test would pass even if the lazy path still auto-saved (the earlier failure mode the bot flagged). With the `fetchJSON` intercept dropping `included`, the lazy path genuinely runs, so this fails before the code change and passes after.

**Scope note, tied to the thread on `lazilyLoadLink`:** this pins only the singular-`linksTo` success path. A near-identical test with a `linksToMany` field (drop `included`, read the field, assert `saved` stays empty after `flush`) would guard the plural sibling that still writes on read; a variant whose link target is missing on disk would guard the failure branches. Without them, both the still-unfixed siblings and any future regression on those paths stay invisible here. Non-blocking.

---

habdelra (commented):
[Claude Code 🤖]

**Review focus:** whether the read/write separation is correct, and whether it's complete across every place `lazilyLoadLink` writes a resolved value back onto the instance.

**Bottom line: the singular `linksTo` fix is correct, minimal, and well-tested — no blocking issues.** `setResolvedField` is a faithful `setField` minus the one `notifySubscribers` call that drives auto-save, and I confirmed no legitimate change-subscriber regresses: the only reactor to a link-field notification is the store's auto-save (`onInstanceUpdated`), and the `listenForCardId` subscribers ignore everything but `id`. The test now genuinely drives the lazy path and would fail before the code change. Green CI corroborates that removing the notify broke nothing.

**What I'd want addressed (non-blocking):** the fix lands on one of four sibling lazy-resolve paths inside `lazilyLoadLink`. The plural `linksToMany` success path and both failure branches still route through `notifySubscribers` → `doAutoSave`, so they retain the identical read-triggers-write behavior — reading a `linksToMany` with un-side-loaded slots, or a card with a deleted/404 link target, still rewrites the card. Mechanism, the trace, and a grounded fix direction (the getter already entangles card tracking, so a raw-backing-array swap + `notifyCardTracking` mirrors the singular fix) are in the inline thread on the changed `else` branch. These siblings predate this PR, so it's a scoping call, not a regression.

**Recommendations:**
1. Extend the fix to the plural `linksToMany` success path, or file a follow-up covering all three siblings — see the thread on `lazilyLoadLink`'s plural/`else` branches in `card-api.gts`.
2. Add mirror tests for the `linksToMany` success case and the broken-link failure case — see the thread on the new test in `store-test.gts`.
3. Consider narrowing the title/description to "a `linksTo`," since as written it reads as a general fix.
4. Confirmation of why `setResolvedField` is safe to drop the notify is on its definition, as a guard rail for the next editor.

Findings were verified by tracing the code paths (`WatchedArray` set trap → subscriber → `onInstanceUpdated` → `doAutoSave`) rather than by running the host suite; each step on the live-app read path is unconditional.

---

habdelra (approved):

packages/base/card-api.gts
backspace: addressed in a60f6a2925 and a0999db16a

packages/host/tests/integration/store-test.gts
backspace: same

PR #5728, opened 2026-08-10
Polish AI Assistant panel and workspace chooser dropdown styling

cmgardella:
## Summary
- Resized the Past Sessions popover to inset 30px from both panel edges, matched its header font to the panel title, restyled its background/border/dividers/hover colors, and fixed a double-height list caused by two nested scroll containers
- The "more options" dropdown now auto-closes on mouse-out via a hover-intent debounce instead of requiring a click elsewhere, and clears lingering keyboard-focus styling on its trigger button
- The active session row keeps its hover background while its own options menu is open
- Restyled the workspace chooser's "View All" sort dropdown: default (closed) state now matches the chooser's own background; open state uses a dark platter with a white 25%-opacity stroke, white text, and a teal checkmark on the active item

## Additional Fixes
- Remove redundant mutation observer from boxel-ui dropdown and select
- Fix bug where selecting "View All" via keyboard on workspace-chooser dropdown opens a random workspace

### DIFF
--- a/packages/host/app/components/ai-assistant/past-session-item.gts
+++ b/packages/host/app/components/ai-assistant/past-session-item.gts
@@ -9,6 +9,7 @@ import { tracked } from '@glimmer/tracking';
 import ExternalLink from '@cardstack/boxel-icons/external-link';
 
 import { format as formatDate, isSameDay, isSameYear } from 'date-fns';
+import { modifier } from 'ember-modifier';
 
 import {
   BoxelDropdown,
@@ -45,6 +46,42 @@ interface Signature {
 
 export default class PastSessionItem extends Component<Signature> {
   @tracked private preventMenuClose = false;
+  private closeDropdownAction: (() => void) | null = null;
+  private closeDropdownTimer: ReturnType<typeof setTimeout> | null = null;
+  private triggerElement: HTMLElement | null = null;
+
+  private registerCloseAction = modifier(
+    (_element: HTMLElement, [closeFn]: [() => void]) => {
+      this.closeDropdownAction = closeFn;
+      return () => {
+        this.closeDropdownAction = null;
+      };
+    },
+  );
+
+  private registerTriggerElement = modifier((element: HTMLElement) => {
+    this.triggerElement = element;
+    return () => {
+      this.triggerElement = null;
+    };
+  });
+
+  @action
+  private scheduleCloseDropdown() {
+    this.cancelCloseDropdown();
+    this.closeDropdownTimer = setTimeout(() => {
+      this.closeDropdownAction?.();
+      this.triggerElement?.blur();
+    }, 200);
+  }
+
+  @action
+  private cancelCloseDropdown() {
+    if (this.closeDropdownTimer) {
+      clearTimeout(this.closeDropdownTimer);
+      this.closeDropdownTimer = null;
+    }
+  }
 
   <template>
     <li
@@ -52,6 +89,8 @@ export default class PastSessionItem extends Component<Signature> {
       data-test-joined-room={{@session.roomId}}
       data-room-id={{@session.roomId}}
       data-is-current-room={{@isCurrentRoom}}
+      {{on 'mouseenter' this.cancelCloseDropdown}}
+      {{on 'mouseleave' this.scheduleCloseDropdown}}
     >
       <button
         class='view-session-button'
@@ -87,7 +126,7 @@ export default class PastSessionItem extends Component<Signature> {
           {{/if}}
         </div>
       </button>
-      <BoxelDropdown>
+      <BoxelDropdown @contentClass='past-session-menu-dropdown'>
         <:trigger as |bindings|>
           <Tooltip @placement='top'>
             <:trigger>
@@ -98,6 +137,7 @@ export default class PastSessionItem extends Component<Signature> {
                 @label='past session options'
                 data-test-past-session-options-button={{@session.roomId}}
                 {{bindings}}
+                {{this.registerTriggerElement}}
               />
             </:trigger>
             <:content>
@@ -108,6 +148,9 @@ export default class PastSessionItem extends Component<Signature> {
         <:content as |dd|>
           <Menu
             class='menu past-session-menu themeless'
+            {{this.registerCloseAction dd.close}}
+            {{on 'mouseenter' this.cancelCloseDropdown}}
+            {{on 'mouseleave' this.scheduleCloseDropdown}}
             @closeMenu={{fn this.handleCloseMenu dd.close}}
             @items={{array
               (menuItem
@@ -139,37 +182,49 @@ export default class PastSessionItem extends Component<Signature> {
       }
 
       .session {
+        position: relative;
         display: flex;
         align-items: center;
         justify-content: space-between;
-        border-top: 1px solid var(--past-sessions-divider-color);
         padding: var(--boxel-sp) var(--boxel-sp-sm);
         margin-right: var(--boxel-sp-xs);
         margin-left: var(--boxel-sp-xs);
         border-radius: var(--boxel-border-radius-xs);
       }
 
-      .session:first-child {
-        border-top: none;
+      .session::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        left: 0;
+        right: 0;
+        height: 1px;
+        background: #4f4b57;
       }
 
-      .session:hover {
+      .session:first-child::before {
+        display: none;
+      }
+
+      .session:hover,
+      .session:has(.menu-button[aria-expanded='true']) {
         background-color: var(--ai-assistant-menu-hover-background);
+        border-radius: calc(var(--boxel-border-radius-xs) + 2px);
         cursor: pointer;
       }
-      .session[data-is-current-room] {
-        border: 1px solid var(--past-sessions-divider-color);
-      }
-      .session:hover + .session:not([data-is-current-room]),
-      .session[data-is-current-room] + .session {
-        border-top-color: transparent;
+      .session:hover::before,
+      .session:has(.menu-button[aria-expanded='true'])::before,
+      .session:hover + .session::before,
+      .session:has(.menu-button[aria-expanded='true']) + .session::before {
+        background: transparent;
       }
       .name {
         font-weight: 600;
       }
       .date {
-        margin-top: var(--boxel-sp-xxs);
+        margin-top: calc(var(--boxel-sp-xxs) - 5px);
         color: var(--boxel-400);
+        font-size: 12px;
       }
       .view-session-button {
         color: var(--boxel-light);
@@ -195,12 +250,17 @@ export default class PastSessionItem extends Component<Signature> {
         visibility: visible;
       }
 
+      :global(.past-session-menu-dropdown) {
+        border: none;
+        box-shadow: none;
+      }
+
       .menu {
         --boxel-menu-item-content-padding: var(--boxel-sp-xxs)
           var(--boxel-sp-sm);
 
-        background: var(--ai-assistant-menu-background);
-        border: 1px solid var(--past-sessions-divider-color);
+        background: #3b394b;
+        border: 1px solid rgba(255, 255, 255, 0.25);
         color: var(--boxel-light);
         padding: var(--boxel-sp-xs);
         box-shadow: var(--boxel-deep-box-shadow);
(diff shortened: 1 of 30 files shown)


packages/host/app/components/ai-assistant/past-session-item.gts
```diff
@@ -195,12 +250,17 @@ export default class PastSessionItem extends Component<Signature> {
...
+      }
+
       .menu {
         --boxel-menu-item-content-padding: var(--boxel-sp-xxs)
           var(--boxel-sp-sm);
 
-        background: var(--ai-assistant-menu-background);
-        border: 1px solid var(--past-sessions-divider-color);
+        background: #3b394b;
+        border: 1px solid rgba(255, 255, 255, 0.25);
```
richardhjtan, line 263: [Claude Code 🤖] **This palette already has tokens, and putting the new values in consumers has split the panel into three different-looking menus.** Blocking in the sense that it produces a visible inconsistency, though the fix is three lines.

**Where the values live today.** `packages/host/app/components/ai-assistant/panel.gts` defines the panel's menu palette:

```css
--ai-assistant-menu-background: #4f4b57;
--past-sessions-divider-color: #75707e;
--ai-assistant-menu-hover-background: #797788;
```

Three components read them: `panel-popover.gts` (`background` + `border`), this file, and `new-session-settings.gts` (`background` + hover).

**What the diff changes, consumer by consumer.** This hunk replaces `var(--ai-assistant-menu-background)` with `#3b394b` and `var(--past-sessions-divider-color)` with `rgba(255, 255, 255, 0.25)`; further down, `var(--ai-assistant-menu-hover-background)` becomes `#272330`. `past-sessions.gts` then overrides the popover from outside with `:deep(.panel-popover) { background: #3b394b; border: 1px solid rgba(255, 255, 255, 0.25); }`.

**What that leaves.** The other two consumers were not updated, so inside one AI Assistant panel:

| Surface | Background | Hover / border |
| --- | --- | --- |
| Past Sessions popover + session menu | `#3b394b` | `#272330` / `rgba(255,255,255,.25)` |
| Rename Session popover (same `AiAssistantPanelPopover`) | `#4f4b57` | `#75707e` border |
| New Session Settings menu | `#4f4b57` | `#797788` |

The hover divergence is the loud one: `#797788` is a mid grey and `#272330` is near-black, so two menus in the same panel now highlight in opposite directions. `rename-session.gts` is the case most likely to be missed in manual verification — it's the same popover component as Past Sessions, so it looks like it should have followed along, and it didn't.

**The fix.** If the new palette is the panel's palette, change it once at the source:

```css
/* panel.gts */
--ai-assistant-menu-background: #3b394b;
--past-sessions-divider-color: rgba(255, 255, 255, 0.25);
--ai-assistant-menu-hover-background: #272330;
```

Then the `var(...)` references in this file go back to being `var(...)`, and the `background`/`border` half of the `:deep(.panel-popover)` override in `past-sessions.gts` deletes itself — the popover already reads both from the tokens. The inset and `max-width` overrides there are genuinely Past-Sessions-specific and stay.

If Past Sessions is *meant* to differ from its siblings, then it needs its own named token rather than literals, so the intent is legible and the next person doesn't "unify" it by accident.

**One value is a straight duplicate.** The new `.session::before { background: #4f4b57; }` divider is byte-identical to `--ai-assistant-menu-background`. Whether that's deliberate or coincidence isn't recoverable from the literal — as `var(--ai-assistant-menu-background)` it would be self-documenting, and as a distinct divider color it wants its own token.

The repo convention is explicit about this (`.claude/skills/gts-component-conventions`, §3): never ship a raw hex in a component; promote a recurring color to a shared token or define a component-scoped custom property. Here the custom properties already exist one component up.

Scope: regression introduced by this PR. Non-blocking mechanically — nothing breaks — but it ships a visible inconsistency, so I'd rather it not land as-is.

packages/host/app/components/operator-mode/workspace-chooser/index.gts
```diff
@@ -567,14 +568,39 @@ export default class WorkspaceChooser extends Component<Signature> {
...
+      :global(
+        .workspace-chooser-sort-dropdown
+          .boxel-select-option-item.ember-power-select-option--selected:hover
+      ) {
+        background-color: #272330;
+      }
+      :global(.workspace-chooser-sort-dropdown .boxel-select-option-checkmark) {
+        --icon-color: #00ffba;
+        color: #00ffba;
+      }
```
richardhjtan, line 603: [Claude Code 🤖] **`#00ffba` is `--boxel-teal`, byte for byte — so this is the highlight token spelled the one way that can't be re-themed.** Non-blocking, one-line fix.

`packages/boxel-ui/src/styles/variables.css`:

```css
--boxel-teal: #00ffba;
--boxel-highlight: var(--boxel-teal);
```

So the intended value is already reachable, and the convention (`.claude/skills/gts-component-conventions`, §6) is to reference `--boxel-highlight` rather than `--boxel-teal` or the literal, because `--boxel-highlight` is the app-wide semantic name for the accent and the one a re-theme changes:

```css
:global(.workspace-chooser-sort-dropdown .boxel-select-option-checkmark) {
  --icon-color: var(--boxel-highlight);
  color: var(--boxel-highlight);
}
```

The same file's other four literals — `#3b394b`, `#272330`, `#fff`, `rgba(255, 255, 255, 0.25)` — are the same four this PR writes into `past-session-item.gts` and `past-sessions.gts`. That's a fourth copy of one palette, in a component in a different feature area, which is the argument for shared tokens rather than per-file literals. `#fff` in particular has `--boxel-light` sitting right there.

**Separate, smaller point on the two blocks above this one.** The first `:global` block drives `BoxelSelect` through its published API — `--boxel-dropdown-background-color`, `--boxel-dropdown-hover-color`, `--boxel-dropdown-highlight-color` and friends are what `packages/boxel-ui/src/components/dropdown/index.gts` reads. Good. The next two reach past that API into ember-power-select's own class names to hit the state combinations the tokens don't express:

```
.boxel-select-option-item.ember-power-select-option--selected.ember-power-select-option--highlighted
.boxel-select-option-item.ember-power-select-option--selected:hover
```

That's three levels into a vendor addon's internals, globally, from an app component. It works, and it's the only route available today — but it's silent-breakage-shaped: an ember-power-select upgrade renames a modifier class and the styling reverts with nothing failing. The durable version is a `--boxel-dropdown-selected-*` pair (or a selected+highlighted token) added alongside the existing ones in `dropdown/index.gts`, which every consumer then gets. Follow-up, not this PR — but worth filing while the need is concrete.

Scope: (1) regression introduced by this PR, trivial fix. (2) follow-up on the boxel-ui token gap. Neither blocks.

packages/host/app/components/ai-assistant/past-session-item.gts
```diff
@@ -45,13 +46,51 @@ interface Signature {
...
+  });
+
+  @action
+  private scheduleCloseDropdown() {
+    this.cancelCloseDropdown();
+    this.closeDropdownTimer = setTimeout(() => {
+      this.closeDropdownAction?.();
+      this.triggerElement?.blur();
+    }, 200);
+  }
```
richardhjtan, line 76: [Claude Code 🤖] **A raw `setTimeout` with no destructor, and it's the one piece of real behavior in an otherwise visual PR.** Non-blocking; two concerns, one of which makes the new behavior untestable.

**No teardown path.** There's no `willDestroy` or `registerDestructor` anywhere in this component, so `closeDropdownTimer` outlives the component. A session row unmounts inside the 200ms window on any of the obvious paths — the popover closes, the room list re-sorts after a rename or delete, `isLoadingMoreAIRooms` resolves and re-renders the list — and the callback still fires.

It happens to be harmless *today*, and it's worth knowing why, because the safety is accidental rather than designed: both things the callback touches are nulled by modifier teardown (`registerCloseAction` clears `closeDropdownAction`, `registerTriggerElement` clears `triggerElement`), so the callback runs as `undefined?.()` twice. That holds only as long as those two modifiers are the only state the callback reads, and only because modifier destructors run before the timer fires. Add one line to the callback that touches `this.args` or a tracked field and it becomes a real destroyed-object access.

**Invisible to `settled()`.** Ember's test helpers track runloop timers and test waiters; a raw `setTimeout` is neither. So `await settled()` returns without waiting for the close, and a test for "the menu auto-closes on mouse-out" would have to sleep on wall-clock time — which is exactly the kind of test that becomes the suite's next flake. Given this PR's test plan is manual browser verification, this behavior currently has no coverage at all, and a Percy snapshot can't see a timer either.

**Both fixed by the same change.** A restartable `ember-concurrency` task makes the timer cancel-on-destroy and settled-visible for free:

```ts
private closeDropdownTask = restartableTask(async () => {
  await timeout(200);
  this.closeDropdownAction?.();
  this.triggerElement?.blur();
});

@action private scheduleCloseDropdown() { this.closeDropdownTask.perform(); }
@action private cancelCloseDropdown() { this.closeDropdownTask.cancelAll(); }
```

`restartable` also subsumes the manual `cancelCloseDropdown()` call at the top of `scheduleCloseDropdown` — a second `perform` cancels the first. If you'd rather not add a task, `@ember/runloop`'s `debounce` plus a `cancel` in a `willDestroy` gets the same two properties with more code.

**Unrelated to the timer, noticed in the same hunk:** `preventMenuClose` is still declared `@tracked` and is now one of several pieces of open/close state on this component alongside `closeDropdownAction`, `closeDropdownTimer`, and `triggerElement`. Worth a look at whether they're all still pulling their weight, or whether the hover-intent path has made one of them dead. Not asking for it here.

Scope: regression introduced by this PR (new code), non-blocking.

packages/host/app/components/ai-assistant/llm-select.gts
```diff
@@ -107,6 +107,15 @@ export default class LLMSelect extends Component<Signature> {
...
       }
 
+      /* LLMSelect renders its own footer-like row inside .llm-list rather
+      than forwarding a :footer block to PillMenu, so PillMenu never marks
+      itself `has-footer` and its default bottom scroll-shadow (meant to be
+      swapped out when a real footer exists) is left dangling over the
+      list content instead of framing the true bottom edge. */
+      :deep(.menu-content::after) {
+        display: none;
+      }
```
richardhjtan, line 117: [Claude Code 🤖] **Confirmation on the diagnosis — it's exactly right — plus a structural alternative that would delete three hunks of this diff instead of adding them.**

**The comment checks out line for line.** `PillMenu` renders `class='pill-menu {{if (has-block "footer") "has-footer"}}'`, and its own stylesheet already contains the suppression this override is duplicating:

```css
.pill-menu.has-footer .menu-content::after {
  display: none;
}
```

Because LLMSelect renders its footer row inside `.llm-list` rather than passing a `:footer` block, `has-block 'footer'` is false, `has-footer` never lands on the element, and that rule never applies — so the bottom scroll-shadow keeps painting over content that has no footer beneath it. Correct diagnosis, and the comment names the mechanism instead of just asserting the effect. Good comment to have.

**The alternative.** Passing the footer row through PillMenu's `:footer` block would make `has-footer` true, which fires the existing rule — and this override disappears. It would also pick up `.menu-footer`'s layout, and that's where it gets interesting, because this PR is currently paying for the split twice:

- `pill-menu/index.gts` gains `display: flex; flex-wrap: wrap; gap: 5px` on `.menu-footer` (the skill-selector fix).
- `room.gts` gains `display: flex; flex-wrap: wrap; gap: 5px` on `.llm-select-footer` (the LLM-select fix).

Same three declarations, same intent, two files — because the two footers are two implementations. That's a decision with two homes, and the first time one learns something the other won't. Routing LLMSelect's footer through the `:footer` block collapses all three hunks: the `.llm-select-footer` layout block, this `::after` override, and the need to keep them in step.

**Why I'm not asking for it in this PR.** The footer currently renders as an `<li>` inside `.llm-list`, and moving it out of the list changes DOM structure inside a menu with its own scroll-shadow and scroll-timeline machinery — a bigger and more visual change than the rest of this diff, and one that wants Percy on it. Landing the override now and filing the structural fix is a reasonable call.

**If it stays as-is**, one thing worth adding to the comment: the same trap waits for the next consumer that renders its own footer row. A note on PillMenu's `::after` rule — that the suppression is keyed on the `:footer` block, so a consumer that hand-rolls a footer must suppress it itself — would put the warning where the next author will be looking.

Scope: confirmation plus a follow-up. Non-blocking.

burieberry:
Making the updates

packages/host/app/components/ai-assistant/llm-select.gts
burieberry: handling this in a follow up PR

PR #5729, opened 2026-08-10
host: Add index reuse to split store tests

backspace:
This is in the vein of #5722 (though with `withCachedRealmSetup`), but the file needed to be broken up.


packages/host/tests/integration/store-test.gts
```diff
@@ -3664,648 +3668,4 @@ module('Integration | Store', function (hooks) {
   // Count full rebuilds by the loader flush each one performs: the coalesced
   // rebuild calls resetLoader exactly once, and nothing else flushes the loader
   // in these tests, so resetLoader-call-count == rebuild-count.
```
habdelra: [Claude Code 🤖]

**Orphaned comment left behind by the split — non-blocking, but a regression this PR introduces.** These three lines documented `countRebuilds()`, which this PR moved to `store-module-rebuild-test.gts` together with the twelve tests. The function is gone from this file, so the comment now dangles at the very end of the `Integration | Store` module body, immediately before the closing `});`, describing a helper that no longer exists here.

The same move also stripped the comment from its subject: `countRebuilds()` in the new file (just above the first rebuild test) now carries no explanatory comment, even though the text is still accurate there — the coalesced rebuild calls `resetLoader` exactly once and nothing else flushes the loader in those tests.

Fix: delete these lines here. If you want to keep the rationale where it's still true, paste them back above `function countRebuilds()` in `store-module-rebuild-test.gts`.

```suggestion
```

---

habdelra (commented):
[Claude Code 🤖]

Reviewed as a test-perf refactor: the lens here is whether the split preserved behavior exactly, whether the dead code and imports were cleaned up on both sides, and whether switching `store-test.gts` to a cached/restored index is safe given that some of its tests write.

**Bottom line: solid, no blocking issues.** One non-blocking nit — an orphaned comment left at the tail of `store-test.gts` (inline thread). Everything else checks out empirically.

**What lands right:**

- **The move is byte-identical.** Diffing the removed block from `store-test.gts@main` against `store-module-rebuild-test.gts` lines 106–749 (the three helpers `countRebuilds`/`renderCard`/`captureTelemetry` plus all twelve tests) reports no differences — nothing was silently reworded or re-scoped in transit.
- **Dead imports pruned correctly, and only those.** `rri`, `RebuildEvent`, and `TelemetryEventInput` are the only imports dropped from `store-test.gts`, and `grep` confirms none of `handleInvalidations` / `isModuleLoaded` / `client-telemetry` / `refreshReferencesForCodeChange` / `saveSource` / `wasModuleFlushedForCodeChange` remain in the file — the loader/telemetry machinery left cleanly. Imports the moved tests shared with the survivors (`waitUntil`, `GlimmerComponent`, `tracked`, `getCard`, `isCardInstance`, `LoaderService`) are all still referenced, so nothing broke on the way out. The three helpers now live only in the new file — no twin copy left behind to drift.
- **Leaving the new file uncached is the right call, and the in-file comment explains why.** The module-rebuild tests assert on the loader's reaction to a source write (a module was flushed, a rebuild fired); a restored snapshot boots with `skipBootIndex`, so no fixture module is ever loaded and those preconditions can't hold. `setupLocalIndexing` with no `withCachedRealmSetup` is exactly what preserves them.
- **Index reuse in `store-test.gts` is safe for the writing tests.** The "each test restores the snapshot rather than continuing from the previous one" claim holds at the DB layer: `withCachedRealmSetup` exports the snapshot during the first test's `beforeEach` (post-index, pre-test-body), and `SQLiteAdapter.importSnapshot` runs `DELETE FROM main.<table>` + copy-from-snapshot for every table, so each subsequent `beforeEach` fully resets index state — a test's write can't leak forward. The one surviving loader-touching test (the boom-person re-index case around line 1000) still passes because it doesn't depend on a boot-loaded module the way the moved tests did; it writes source at test time and waits for live re-indexing. `setupRealmCacheTeardown` scopes snapshot cleanup to this module's prefix, so nothing bleeds across files.
- **CI agrees:** all host shards and the Host Test Results check are green (0 failed), and the new file was auto-discovered (test count rose), so no manifest needed touching.

**Recommendation:**

1. Remove the orphaned `// Count full rebuilds…` comment at the end of `store-test.gts`; optionally re-attach it above `countRebuilds()` in the new file where it's still accurate. Non-blocking — see the inline thread.

---

habdelra (approved):

PR #5731, opened 2026-08-10
feat: Add progress bar for publishing

backspace:
s 2026-08-10 at 11 34 03@2x

s 2026-08-10 at 11 28 59@2x

s 2026-08-10 at 13 57 56@2x


packages/runtime-common/realm-operations.ts
```diff
@@ -343,6 +446,58 @@ export const waitForReady: RealmOperation<WaitForReadyInput, void> = async (
...
+      await new Promise<void>((resolve) => {
+        let timer = setTimeout(resolve, intervalMs);
+        wake = () => {
+          clearTimeout(timer);
+          resolve();
+        };
+      });
+    }
+  })();
+
```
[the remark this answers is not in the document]
backspace: fixed in f5683f8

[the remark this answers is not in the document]
backspace: fixed in f5683f8

packages/realm-server/handlers/handle-publish-progress.ts
```diff
@@ -0,0 +1,245 @@
...
+    return true;
+  }
+  let rows = await query(dbAdapter, [
+    'SELECT 1 FROM prerendered_html WHERE realm_url =',
+    param(realmURL),
+    'AND generation >=',
+    param(genRow.current_generation),
+    'LIMIT 1',
+  ]);
+  return rows.length > 0;
```
habdelra: [Claude Code 🤖] **Non-blocking (follow-up): the `render`→`done` boundary re-implements the readiness gate's generation check, and the two copies can drift silently.**

`publishedHtmlHasCaughtUp` decides the `render`→`done` transition by comparing `prerendered_html.generation` against `realm_generations.current_generation`. As the comment says, this mirrors `awaitPublishedHtmlReady` — the same predicate the `_readiness-check?awaitPrerenderHtml=true` gate uses to decide a publish is actually finished. That's the correct signal, and mirroring it is exactly what keeps the progress bar and the moment readiness passes from disagreeing.

The concern is that it's a *copy*, not a shared call. This endpoint exists so the bar and the "ready" moment can't contradict each other; if the readiness gate's generation semantics ever change (how `current_generation` is bumped, or the `>=` comparison), this predicate has to change in lockstep or the two diverge with nothing failing loudly — the bar reports `done` while readiness still waits, or the reverse.

**Class:** follow-up, non-blocking. Both paths are well tested today, so this is about preventing future drift, not fixing a current bug.

**Way out:** extract the shared "has the current generation's HTML landed?" check into one helper that both `awaitPublishedHtmlReady` and this endpoint call, so the invariant has a single home.

---

packages/realm-server/handlers/handle-publish-progress.ts
```diff
@@ -0,0 +1,245 @@
...
+    dbAdapter,
+    indexingConcurrencyGroup(realmURL),
+  );
+  if (indexJob) {
+    return indexJob.has_worker
+      ? { phase: 'index', ...counts(indexJob) }
+      : // Nothing is working the job, so it has no counts to report and none
+        // are coming until a worker picks it up.
+        { phase: 'queued', filesCompleted: 0, totalFiles: 0 };
+  }
```
habdelra, line 173: [Claude Code 🤖] **Non-blocking (confirmation + one assumption worth naming): this reports on whatever job holds the realm's lane, not the publish's job specifically.**

`readPublishProgress` keys entirely off the concurrency group (`indexing:<realm>` / `prerender_html:<realm>`) and reports whatever unfulfilled job currently holds that lane — no publish or job identity is threaded through. That is safe as wired today: the only caller polls this from inside `waitForReady`, and a *published* realm is normally only indexed as part of publishing, so the job in its lane is the publish's job.

The assumption is implicit, though. If a published realm were re-indexed for an unrelated reason concurrently with a publish, or if a future caller polled this endpoint outside a `waitForReady` window, the reading would attribute that unrelated indexing to publish progress.

**Class:** confirmation — the scoping holds for the current caller, so no change is required in this PR.

**Way out (optional):** a one-line comment on `readPublishProgress` naming the assumption — progress reflects the realm's index/render lane, which during a publish is the publish's own work — so the next person who reuses this endpoint doesn't inherit a subtly wrong reading with no signpost.

---

packages/realm-server/handlers/handle-publish-progress.ts
```diff
@@ -0,0 +1,245 @@
...
+
+      let progress = await readPublishProgress(dbAdapter, publishedRealmURL);
+
+      await setContextResponse(
+        ctxt,
+        new Response(
+          JSON.stringify({
+            data: {
+              type: 'publish-progress',
+              id: publishedRealmURL,
```
habdelra, line 119: [Claude Code 🤖] **Nit (non-blocking):** the JSON:API `data.id` is the full published realm URL. It is the natural key here and the client uses it as one (the host keys its per-target progress map by this value), so this is fine in practice — just flagging that a URL as a resource `id` is a little unconventional. No change needed.

---

packages/host/app/components/operator-mode/publish-realm-modal.gts
```diff
@@ -1837,6 +1893,38 @@ export default class PublishRealmModal extends Component<Signature> {
...
+
+      .progress-target {
+        font-weight: 600;
+      }
+
+      /* The shared bar sizes its track in em, so the font size is the height
+         control — this keeps it a slim rule under the description rather than a
+         full-height labelled bar. */
+      .progress-bar {
+        font-size: 0.375rem;
```
habdelra: [Claude Code 🤖] **Non-blocking (follow-up): sizing the shared `ProgressBar` through `font-size` couples these bars to that component's internals.**

`font-size: 0.375rem` here is being used as a height control — it works only because the shared `ProgressBar` sizes its track in `em`, as the comment above states. That means the em-based track sizing is an undocumented contract this depends on: a future refactor of `ProgressBar` that stopped deriving track height from font-size would silently thicken this bar (and the identical one in `publishing-realm-popover.gts`) with no test catching it.

**Class:** follow-up, non-blocking. It renders correctly today and the comment is honest about the mechanism.

**Way out:** give `ProgressBar` a real height/size affordance (a `@size` arg, or a documented height CSS var) that both call sites use, instead of shrinking the font. The same hack lives in the popover — see that thread; flagging both so whoever adds the affordance updates them together.

---

packages/host/app/components/operator-mode/host-submode/publishing-realm-popover.gts
```diff
@@ -91,14 +131,45 @@ export default class PublishingRealmPopover extends Component<PublishingRealmArg
...
+        color: var(--muted-foreground);
+        font-size: var(--boxel-font-size-xs);
+        line-height: var(--boxel-line-height-xs);
+      }
+
+      /* The shared bar sizes its track in em, so the font size is the height
+         control — this keeps it a slim rule under the URL rather than a
+         full-height labelled bar. */
+      .progress-bar {
+        font-size: 0.375rem;
```
habdelra: [Claude Code 🤖] **Non-blocking (follow-up):** same `font-size`-as-height coupling to the shared `ProgressBar` as in `publish-realm-modal.gts` — see that thread for the mechanism and the suggested `ProgressBar` height affordance. Flagging both call sites so they get updated together when that affordance lands.

---

habdelra (commented):
[Claude Code 🤖] **Review focus: the progress-reporting mechanics — how a reading is produced, how it is transported to the CLI and host, and how the wait behaves when a reading fails.**

**Bottom line: no blocking issues.** This is a well-executed change — a careful security model, strong tests, and the tricky async teardown is correct. Every finding below is low-severity polish or a follow-up.

**What lands right (worth defending):**
- **Reader over a coalesced table, not a per-file event stream.** A reading is produced by reading `jobs` / `job_progress` from Postgres, so any replica answers identically and the change adds no new per-file event traffic. The write side it reads (`IndexingEventSink` → `job_progress`) already coalesces to one write per tick and is pre-existing infrastructure.
- **The security model is careful and tested.** Owner-only via `realm-owner`; an unknown realm and an unauthorized one both return 403, so the endpoint never confirms whether a published URL exists; `cache-control: no-store` so an intermediary can't freeze the bar.
- **The async teardown in `sampleProgress` is the hard part and it is right.** A stalled progress read cannot hold up `waitForReady` (it aborts, and the `stopped` re-check bars a late reading from landing after readiness passes), and a failed sample cannot fail the publish. Each of those failure modes has a dedicated test.
- **Phase-machine coverage is genuinely thorough**, including the two subtle cases: a dead worker's expired reservation must not report its stale counts, and a finished job's lingering `job_progress` row must be ignored by reading the lane by `status` rather than by the presence of a progress row.
- **CLI progress is routed to stderr and is `--quiet`-aware**, so `boxel realm publish --json` stdout stays a clean parseable payload.
- **The host's per-target progress map is cleared in the publish task's `finally`**, so readings don't accumulate across publishes.

1. `render`→`done` re-implements the readiness gate's generation check — extract a shared predicate so the two can't drift. See the `publishedHtmlHasCaughtUp` thread.
2. Progress reflects whatever holds the realm's lane, not the publish's job specifically — safe for the current caller; name the assumption in a comment. See the `readPublishProgress` thread.
3. `ProgressBar` height is set via `font-size` in two components — a real size affordance on `ProgressBar` would be more robust. See the two `.progress-bar` threads.
4. `data.id` as a URL — nit, no change needed. See that thread.

**Adjacent, out of scope:** diffing this branch against `main` locally surfaces a large set of unrelated files (image-metadata extractors, ai-bot changes, a regenerated memory baseline) because the branch's merge base trails `main`. GitHub's own changed-files list for this PR is the authoritative set and does not include them, so it does not affect the review — but bringing the branch up to date with `main` would spare local reviewers and CI from reading noise. Not something I'm asking for in this diff.

---

packages/host/app/components/operator-mode/publish-realm-modal.gts
backspace: in afd974c2ef

packages/host/app/components/operator-mode/host-submode/publishing-realm-popover.gts
backspace: also in afd974c2ef

packages/realm-server/handlers/handle-publish-progress.ts
backspace: 6b1f29a097

packages/realm-server/handlers/handle-publish-progress.ts
backspace: addressed in 00210835

habdelra (approved):

PR #5734, opened 2026-08-11
Give the spec type tag right-hand spacing in fitted view

richardhjtan:
## What this does

Two spacing fixes to the Spec card's type treatment:

**Inside the type pill**: the pill's default inline padding is `--boxel-sp-3xs` (~4.8px), which leaves the type label nearly touching the pill's own right edge. `SpecTag` now sets `--boxel-pill-padding: var(--boxel-sp-4xs) var(--boxel-sp-xs)` so the label gets ~9px each side.

**Around the fitted strip**: in horizontal fitted layouts 250px and wider, the fitted template's padding shrinks to `--boxel-sp-3xs`, crowding the icon against the left edge and the pill against the right. The spec fitted view adds symmetric `--boxel-sp-sm` inline padding and an `--boxel-sp-xs` gap. Narrow badges below 250px keep the tight padding since they have no room to spare; vertical layouts are untouched.

Most visible on the catalog listing detail's Includes tab, where specs render as fitted strips. The single-boundary, taller rows in the screenshots come from a companion change to the listing card in the boxel-catalog repo.

## Before

before

## After
Screenshot 2026-08-13 at 12 55 22 AM

## Testing

- `ember-template-lint` and Prettier pass on `spec.gts`
- Verified live against a local realm server: strip inline padding computes to 12px each side, pill padding to 9px inline, pill 13px from the card edge


lukemelia (approved):

burieberry (commented):
`flex-shrink: 0;` on the pill also works.

burieberry (approved):

richardhjtan:
[Claude Code 🤖] Good call — added in 25b45731c7. The padding override handles the spacing inside the pill, and `flex-shrink: 0` guarantees a long title can never compress the pill itself (previously that was only implicit via the non-wrapping label's min-content floor).

PR #5737, opened 2026-08-11
host: Update `operator-mode…` and `tools-test` with index cache

backspace:


packages/host/tests/acceptance/operator-mode-acceptance-test.gts
```diff
@@ -73,6 +75,7 @@ module('Acceptance | operator mode tests', function (hooks) {
   let testRealm: Realm;
   setupApplicationTest(hooks);
   setupLocalIndexing(hooks);
+  setupRealmCacheTeardown(hooks);
```
habdelra, line 78: [Claude Code 🤖]

**This teardown cleans the top-level module's snapshot, but not the ones the file's four nested modules create — those leak for the rest of the shard.** Non-blocking, and it doesn't affect correctness (the fixtures are identical and CI is green); it's a resource-cleanup gap this file introduces, and the same shape already exists at smaller scale in `code-submode/inspector-test.ts`.

**Mechanism.** The realm-building `beforeEach` (line 94) wraps its setup in `withCachedRealmSetup`, which keys its snapshot on `getCurrentModuleCacheKey()` — the *current* (leaf) module's name. QUnit runs that inherited `beforeEach` for every test in the file, including the four nested modules (`card with an error that has a last known good state`, `2 stacks`, `realm session expiration`, `account popover`), and sets each test's `module` to its own leaf module. So for a test in one of those, the cache key is the nested name, and `withCachedRealmSetup` `exportSnapshot`s a *separate* snapshot under a nested-derived prefix. `setupRealmCacheTeardown(hooks)` here, by contrast, resolved its delete-prefix once at registration from the top-level module name, so the `deleteSnapshotsByPrefix` in its `hooks.after` never matches the nested snapshots. Each `exportSnapshot` ATTACHes an in-memory DB and leaves it attached until an explicit `deleteSnapshot`; the unmatched ones stay attached — holding one of SQLite's capped ATTACH slots (the exact reason `setupLocalIndexing`'s `after` hook deletes its own snapshot) — until the adapter closes at end of shard.

**Verified.** Both the snapshot name and the teardown prefix are built from `simpleHash(moduleName)`:

```
top-level teardown prefix:  snapshot_68dd2136_
nested snapshot names:      snapshot_bcd288b6_...   (card with an error…)
                            snapshot_af6be67_...    (2 stacks)
                            snapshot_b2cc26aa_...   (realm session expiration)
                            snapshot_e77900c8_...   (account popover)
```

None of the nested names start with `snapshot_68dd2136_`, so the top-level teardown skips all four. There's a smaller redundant-work cost too: each nested module runs a full boot index on its first test (its own key has no snapshot yet, and `setupLocalIndexing`'s `reset()` cleared the tables, so `isNewIndex()` is true) rather than restoring the top-level snapshot.

**The fix.** Register `setupRealmCacheTeardown(hooks)` inside each nested module that inherits this setup — the pattern `code-submode/card-playground-test.gts` already uses, where each nested `module(...)` calls both `setupRealmCacheTeardown(hooks)` and `withCachedRealmSetup` on its own `hooks`, so the delete-prefix matches the key. Or, if the bounded leak is acceptable (as it currently is for inspector-test), a one-line note here saying so would save the next reader this trace.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** This is a test-only change with one job — cache the two-realm index across each module's tests instead of rebuilding it per test — so the review is about whether the caching is correct, actually saves the work it claims, and cleans up after itself.

**Bottom line: the mechanism is sound and I found no blocking issues.** There is one non-blocking cleanup gap in `operator-mode-acceptance-test.gts` (see the inline thread), where the file's nested modules leak their cached snapshots.

**What lands right.**
- Wrapping *both* realms in a single `withCachedRealmSetup` block is the correct shape for a multi-realm module: `exportSnapshot` runs only after the callback resolves — i.e. after both realms have finished indexing — so the snapshot can't omit the second realm. That's exactly the failure `captureIndexManually` + `captureReusableIndex()` existed to prevent by hand, and folding the timing into one block removes that footgun. The removed option isn't left orphaned — `interact-submode-setup.gts` is still a user, so the helper stays.
- The speedup is real, and I traced it rather than trusting it: `withCachedRealmSetup` does `importSnapshot` (a full-table replace that refills `realm_generations` among the index tables) before re-running setup, so `realm.start()` → `#startup()` reads `isNewIndex()` as false and never calls `fullIndex()`. Nothing is re-rendered or re-serialized — the boot index is skipped, not repeated. And because setup still runs, the loader / `definition-lookup` bookkeeping that `skipBootIndex` omits is preserved, which is the reason to prefer this over the `reuseIndexAcrossTests` + `skipBootIndex` path here.
- `tools-test.gts` is a flat module, so its single top-level snapshot is exactly what `setupRealmCacheTeardown(hooks)` cleans — no gap there.

**One guard rail worth knowing.** Unlike the `reuseIndexAcrossTests` path (which guards with `reusableIndex.realmURLs` and throws when a test builds a realm the snapshot doesn't contain), `withCachedRealmSetup` has no such check — it restores, sees `isNewIndex()` false, and serves the cached index. So the "Both realms are the same for every test in this module" precondition in the new comment is load-bearing: if a future test in either module builds different fixtures without changing the cache key, it will silently get the cached ones, with nothing failing loudly. Correct as written today; just the condition under which it stops being safe.

**Recommendation.**
1. Register `setupRealmCacheTeardown(hooks)` inside each nested module of `operator-mode-acceptance-test.gts`, or note that the bounded leak is intentional — mechanism and evidence in the inline thread.

---

packages/host/tests/acceptance/operator-mode-acceptance-test.gts
backspace: fixed in d24a6ee

packages/host/tests/acceptance/operator-mode-acceptance-test.gts
```diff
@@ -315,177 +318,181 @@ module('Acceptance | operator mode tests', function (hooks) {
...
-            attributes: {
-              cardTitle: 'Person Card',
-              cardDescription: 'Spec for Person Card',
-              specType: 'card',
-              ref: {
-                module: testRRI('person'),
-                name: 'Person',
+    // Both realms are the same for every test in this module, so the indexed
+    // result is cached and restored rather than rebuilt.
+    await withCachedRealmSetup(async () => {
```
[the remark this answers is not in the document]
backspace: it must be delivering the claimed indexing cache because previously the `searchable search doc` test was taking ≈8.25min and after takes ≈1.25min.

But preserving mtimes in fixtures is a good idea, it’ll now be part of CS-12514.

habdelra (approved):

PR #5738, opened 2026-08-11
host: Add test caching for `host-submode` and `card-copy`

backspace:
This is in the same vein as many host test indexing caching PRs, but uses the older `withCachedRealmSetup` which we may use to entirely replace the recent `setupLocalIndexing` addition.

### DIFF
--- a/packages/host/tests/acceptance/host-submode-test.gts
+++ b/packages/host/tests/acceptance/host-submode-test.gts
@@ -227,13 +227,9 @@ module('Acceptance | host submode', function (hooks) {
 
   module('with a dangling host routing rule', function (hooks) {
     // Each snapshot this module's tests create stays attached until it is
-
     // deleted, and SQLite caps attached databases per connection. The prefix is
-
     // derived from the running module's name, and nested modules have distinct
-
     // names, so every scope that runs tests registers its own teardown.
-
     setupRealmCacheTeardown(hooks);
 
     hooks.beforeEach(async function () {
@@ -309,13 +305,9 @@ module('Acceptance | host submode', function (hooks) {
 
   module('with a realm that is not publishable', function (hooks) {
     // Each snapshot this module's tests create stays attached until it is
-
     // deleted, and SQLite caps attached databases per connection. The prefix is
-
     // derived from the running module's name, and nested modules have distinct
-
     // names, so every scope that runs tests registers its own teardown.
-
     setupRealmCacheTeardown(hooks);
 
     hooks.beforeEach(async function () {
@@ -362,13 +354,9 @@ module('Acceptance | host submode', function (hooks) {
 
   module('with a realm that is publishable', function (hooks) {
     // Each snapshot this module's tests create stays attached until it is
-
     // deleted, and SQLite caps attached databases per connection. The prefix is
-
     // derived from the running module's name, and nested modules have distinct
-
     // names, so every scope that runs tests registers its own teardown.
-
     setupRealmCacheTeardown(hooks);
 
     hooks.beforeEach(async function () {
@@ -782,13 +770,9 @@ module('Acceptance | host submode', function (hooks) {
 
     module('publish and unpublish realm', function (hooks) {
       // Each snapshot this module's tests create stays attached until it is
-
       // deleted, and SQLite caps attached databases per connection. The prefix is
-
       // derived from the running module's name, and nested modules have distinct
-
       // names, so every scope that runs tests registers its own teardown.
-
       setupRealmCacheTeardown(hooks);
 
       let publishDeferred: Deferred<void>;
--- a/packages/host/tests/acceptance/host-submode-test.gts
+++ b/packages/host/tests/acceptance/host-submode-test.gts
@@ -769,10 +769,25 @@ module('Acceptance | host submode', function (hooks) {
     });
 
     module('publish and unpublish realm', function (hooks) {
-      // Each snapshot this module's tests create stays attached until it is
-      // deleted, and SQLite caps attached databases per connection. The prefix is
-      // derived from the running module's name, and nested modules have distinct
-      // names, so every scope that runs tests registers its own teardown.
+      // This module calls no caching helper of its own, so the registration below
+      // looks superfluous. It isn't. The parent's beforeEach runs for these tests
+      // too and caches the realm it builds under a key from
+      // `getCurrentModuleCacheKey()`, which reads the *running test's* module
+      // name — for a test in here that is the composed
+      // "… > with a realm that is publishable > publish and unpublish realm". So
+      // the parent exports a snapshot under this module's name and leaves it
+      // attached.
+      //
+      // Teardown deletes by `snapshot_${simpleHash(moduleName)}_`, and the hash
+      // of the parent's name is not the hash of this one, so the parent's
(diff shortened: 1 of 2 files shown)


packages/host/tests/acceptance/host-submode-test.gts
```diff
@@ -254,9 +255,15 @@ module('Acceptance | host submode', function (hooks) {
...
         },
       };
-      await setupAcceptanceTestRealm({
-        mockMatrixUtils,
-        contents: realmContents,
+      // Each of these nested modules builds the same realm for every one of its
+      // tests, so the indexed result is cached per module and restored instead
+      // of being rebuilt. The seeded realm_metadata row above is part of what
+      // the snapshot captures, so a restored test starts from the same state.
+      await withCachedRealmSetup(async () => {
```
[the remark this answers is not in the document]
backspace: addressed in 2a1ef9c

packages/host/tests/acceptance/host-submode-test.gts
```diff
@@ -709,6 +769,12 @@ module('Acceptance | host submode', function (hooks) {
     });
 
     module('publish and unpublish realm', function (hooks) {
+      // Each snapshot this module's tests create stays attached until it is
+      // deleted, and SQLite caps attached databases per connection. The prefix is
+      // derived from the running module's name, and nested modules have distinct
+      // names, so every scope that runs tests registers its own teardown.
+      setupRealmCacheTeardown(hooks);
```
habdelra, line 791: [Claude Code 🤖] **Confirmation — this registration is required and correct; keep it. Non-blocking.** This is the subtlest teardown in the change, so worth stating plainly enough that it survives a future "cleanup."

`publish and unpublish realm` has no `withCachedRealmSetup` call of its own — its `beforeEach` only wires up the publish/unpublish deferreds. The snapshot this teardown deletes is created by the **parent** `with a realm that is publishable` `beforeEach`, which also runs for these nested tests. `withCachedRealmSetup` keys the snapshot via `getCurrentModuleCacheKey()`, which reads the *running test's* module name; for a nested test that's the composed `… > with a realm that is publishable > publish and unpublish realm`. So the parent's `beforeEach` exports and leaves attached a snapshot under **this child's** prefix.

`setupRealmCacheTeardown` deletes by `snapshot_${simpleHash(moduleName)}_`, and `simpleHash` of the parent name and the child name differ, so the parent's teardown (registered inside `with a realm that is publishable`) never matches the child-prefixed snapshots — this registration is the only thing that removes them. Dropping it would leak one attached snapshot database for the rest of the shard, i.e. exactly the attached-database-limit failure the teardown exists to prevent.

One consequence worth knowing, not a problem to fix: the parent's direct tests and these nested tests build the same realm but cache it under two different keys, so it's indexed once per scope rather than shared. That's inherent to module-name keying and harmless — just not something to "deduplicate" by removing a registration.

---

packages/host/tests/integration/components/card-copy-test.gts
```diff
@@ -167,147 +174,154 @@ module('Integration | card-copy', function (hooks) {
...
-            type: 'card',
-            meta: {
-              adoptsFrom: {
-                module: '@cardstack/base/cards-grid',
-                name: 'CardsGrid',
+    // These three realms are the same for every test in this module, so the
+    // indexed result is cached and restored rather than rebuilt. The fourth
+    // realm one test builds for itself stays outside this: a cache hit
+    // replaces every table, which would discard the realms that test already
+    // has.
```
habdelra: [Claude Code 🤖] **Nit — comment accuracy. Non-blocking.** There's no distinct fourth realm here; the reasoning is right but the noun is off. The test that builds a realm outside this block is `copy button does not appear when destination index belongs to read-only realm`, and it re-provisions `readOnlyRealmURL` — the same URL as the third realm set up above (`setupIntegrationTestRealm({ realmURL: readOnlyRealmURL, … })`) — with broader permissions and an extra `Pet/paper.json`. So it's a re-provision of the read-only realm, not a fourth one.

The load-bearing point still holds: it must stay outside `withCachedRealmSetup` because on a cache hit `importSnapshot` deletes and refills every table, which would wipe the three realms the cached block just restored. Consider rewording to e.g. "the read-only realm one test re-provisions for itself" so a reader counting realm URLs isn't left hunting for a fourth.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with one lens: does adopting `withCachedRealmSetup` here preserve per-test isolation, and is every snapshot the change attaches also torn down — the attached-database-limit failure mode is the one thing this class of change reliably gets wrong. I also checked the new comments as claims, since the diff is mostly explanatory prose.

**Bottom line: no blocking issues.** The caching is applied correctly, the reset/import ordering keeps tests isolated, teardown coverage matches every scope that attaches a snapshot (including the non-obvious nested one), and all 70 checks are green. Two non-blocking notes inline.

**What lands right (verified against the checked-out branch):**

- **Reset/import ordering is safe.** `setupLocalIndexing`'s `beforeEach` runs `dbAdapter.reset()`, and it's registered on the outer hooks before each module reaches `withCachedRealmSetup` (host-submode: outer module hooks vs. nested `beforeEach`; card-copy: `setupLocalIndexing` at the top vs. the realm-building `beforeEach` below it). So on a cache hit `main` is cleared first and then refilled from the snapshot — no rows survive the import into the next test. The seeded `realm_metadata` row is written before `withCachedRealmSetup`, so it's part of what the first (miss) test exports; restored tests start publishable exactly as the comment says, which the green publishable-module tests confirm.

- **Teardown coverage is complete.** Four scopes actually attach a snapshot — the three nested host-submode modules plus `publish and unpublish realm` — and each registers its own teardown; card-copy's single module registers one. The nested `publish and unpublish realm` case is the one that looks wrong at a glance (it registers teardown but never calls `withCachedRealmSetup`); see the inline confirmation for why it's genuinely required — its snapshot is produced by the parent's `beforeEach` under the child's composed-name prefix, which the parent's teardown can't match.

- **`ensureRealmMeta` targets the real gap.** Realm resources were only ever created incidentally by app components rendering during the setup visit; caching changed what rendered, so tests that read the resource before their own visit lost it. Registering it explicitly in `beforeEach`, in all three modules rather than only the one that fails today, is the right defensive call — the failure mode is timing-dependent and cheap to prevent everywhere.

**Open discussion:** the single review thread (missing teardown, flagged by the Codex bot) is resolved correctly. I re-derived the prefixes for all four scopes and they match their snapshots, so the registrations added cover every leak.

**Recommendations (both non-blocking):**
1. Reword the card-copy "fourth realm" comment — it's a re-provision of the read-only realm, not a distinct fourth realm. [inline]
2. Keep the `publish and unpublish realm` teardown as-is; the inline note documents why it's required so it isn't removed later. [inline]

Nothing adjacent worth flagging — the change is scoped tightly to these two files.

---

habdelra (approved):

packages/host/tests/integration/components/card-copy-test.gts
backspace: updated in 59aacad874

packages/host/tests/acceptance/host-submode-test.gts
backspace: as of 651fb83ae8 the comment is more thorough

PR #5739, opened 2026-08-11
Measure the older realm cache against the newer one

backspace:
Claude:

**Experiment — not intended to merge as-is.**

Two mechanisms now cache indexed realms across a module's tests, and nothing has compared them.

| | scope | boot index | tolerates a realm only some tests build |
|---|---|---|---|
| `withCachedRealmSetup` (~120 modules) | per setup call, keyed by running module | restores, then still runs setup, so the index pass finds rows current and does nothing | yes |
| `reuseIndexAcrossTests` (16 modules) | one snapshot per module | `skipBootIndex` — the pass never runs | no |

The difference isn't only speed. Skipping the boot index also skips what it does besides filling tables — evicting a module from the loader, recording that it was flushed — which is why two `store-test` tests had to move into their own module.

This converts `searchable search doc` to the older mechanism, which is the cleanest comparison available: **8.16m** uncached, **1.12m** under `reuseIndexAcrossTests`, same fixtures and assertions either way. Whatever this run reports for that module is what the older mechanism delivers unaided.

**How to read the result**

- **Close to 1.12m** → the newer mechanism is mostly redundant, and its constraints are a cost paid for little. The 16 modules should move across and it should go.
- **Nearer 8.16m** → skipping the index pass is where the win is, and both belong, chosen per module by whether its realms are unconditional and its tests indifferent to how the index came to exist.

The number to read is the `Integration | searchable search doc` module total in the merged junit report, not shard wall time.

### DIFF
--- a/packages/host/tests/integration/searchable-search-doc-test.gts
+++ b/packages/host/tests/integration/searchable-search-doc-test.gts
@@ -13,6 +13,7 @@ import CardStoreWithGarbageCollection from '@cardstack/host/lib/gc-card-store';
 import type StoreService from '@cardstack/host/services/store';
 
 import {
+  withCachedRealmSetup,
   testRealmURL,
   setupCardLogs,
   setupLocalIndexing,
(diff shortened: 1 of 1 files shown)


backspace:
The finding is that the earlier caching was good enough! So new things have used it instead, and we’ll move previous uses of the new one to the old one.

PR #5740, opened 2026-08-11
[FROZEN — DO NOT MERGE] execution runtime architecture - sandboxing prototype

habdelra:
> **This branch is FROZEN, and is not intended to merge.** It is the reference implementation and behavioral oracle for the production sandboxing work, which is built fresh from `main`. Its value is that it keeps working as a comparison target, so it takes **critical fixes only** — nothing that would change what it demonstrates, and no new features.

A complete working prototype of the Boxel execution runtime: authored card code renders and executes without authority over the host application, across three execution contexts, with the card author's API unchanged. A card written against the shipped Card API runs here untouched.

- **Direct** — our own trusted code, executed by the host on the host loader, unchanged from the shipped render path.
- **Capsule** — authored code evaluated in an SES compartment. Its imports of trusted modules resolve to authority-free stand-ins; its template is captured as instructions rather than wired, validated instruction by instruction, and rebuilt host-side into an ordinary Ember component through public APIs. The host's own Glimmer renders the result; only the card's logic is caged.
- **Sandbox** — authored code and its own Glimmer inside an origin-isolated iframe, reachable only over one nonce-established private message channel.

All three tiers render real cards, the same fixture crosses all three through one contract, and the evidence base includes a corpus of real workspace cards.

The branch also carries the normative work that outlives the prototype code: the rendering-protocol spec, in which every statement has an id and CI enforces that statements and conformance tests point at each other, plus the conformance suites, fixtures, and classifier signal vocabularies. Those port forward aggressively. The orchestration code is re-derived rather than copied, guided by this branch's own minimality and parity audits — which name the competing projection builders that collapse to one, the Card API stand-in that shrinks to authored semantics only, the four transports that share one message kernel, and the dead and duplicated surface that never comes across.

## Where to start reading

- [Reviewer's guide](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-execution-runtime-reviewers-guide.md) — the best single narrative of what is here
- [Boxel rendering protocol](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-rendering-protocol.md) — the normative spec, RP-0…RP-21
- [Protocol-first plan](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-rendering-protocol-plan.md) — the delivery philosophy
- [Execution runtime architecture](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-execution-runtime-architecture.md) — the design rationale
- [Capsule ↔ Glimmer/DOM review](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-capsule-glimmer-dom-review.md) — the template-boundary risk analysis and the open upstream questions
- [Minimality review](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-execution-runtime-minimality-review.md) · [Frozen-branch parity audit](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-frozen-branch-parity-audit.md) — what to rebuild versus port
- [Realm-mirror compatibility strategy](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-realm-mirror-compatibility-strategy.md) — the evidence layers and completion gates
- [Sandbox HMR extraction](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-sandbox-hmr-extraction.md) · [Volatile execution plan](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-volatile-execution-plan.md) — the editing and hot-reload design
- [Performance plan](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-execution-performance-plan.md) · [Cold-start baseline](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-execution-runtime-cold-start-baseline.md) · [Wild corpus](https://github.com/cardstack/boxel/blob/codex/boxel-execution-runtime-architecture/docs/boxel-execution-runtime-wild-corpus.md) — measurement and evidence


habdelra:
[Claude Code 🤖] Re-review of the execution runtime architecture. This is a different and far more serious piece of work than the July SES spike (#5663) — a fresh branch with a normative spec, a reviewer's guide, a 12-case acceptance suite, and a CI bijection gate. It directly answers most of my prior findings. Everything load-bearing below was verified against the checked-out code and, where noted, in a live browser against the real runtime rather than trusting the docs. Finding #1 has a working proof-of-concept.

## Bottom line

The architecture is sound and the containment that is implemented largely holds under probing. But the branch's headline claim — *authored code never executes in the trusted host; the tiers contain it* — **does not hold as written**, via an authored `static getComponent` override that renders with full host authority. That is live in every environment (the default Capsule path), and it is the thing to fix first. The three high-severity Sandbox findings are real but latent (the tier is unconfigured in deployed environments). The author's own CI bijection gate is currently red.

## What genuinely improved since #5663

These were my main prior findings and they are addressed:

- **Containment vs entitlement are now separate axes** (RP-21). A misclassification is a smoothness cost, not an authority escalation, because Capsule and Sandbox get identical entitlements. Right model, and it mostly holds in code.
- **The Direct (trusted-host) tier is provenance-gated, not code-gated.** The classifier's return type is literally `'capsule' | 'sandbox'` — it can never emit `'direct'`, so a classifier bug cannot leak authored code into the trusted tier by construction.
- **The iframe origin is fail-closed.** An unconfigured non-localhost deploy throws (`boxel-execution.ts:1183`) instead of silently disabling; the `childOrigin === location.origin` reject fires before `iframe.src` is set (`sandbox-runtime-process.ts:985`); dev uses a genuinely distinct origin (`user.localhost`). My prior origin-collapse finding is closed.
- **The fetch broker now has an exact-URL allowlist, pre-authorization, redirect re-check, and byte caps** on the module and resource lanes.
- **Real hardening added**: prerendered authored CSS is validated against a scope policy before entering the host document (`search-entries.ts:706`); Capsule templates reject top-layer attributes and dynamic inline styles; `prefersFullSandbox` can only strengthen, never request Direct.

## Verified strengths (my own probes)

An adversarial probe suite against the real Capsule evaluator, in-browser — 23/23 assertions green:

- Ambient authority absent: `window`, `localStorage`, `fetch`, `XMLHttpRequest`, `WebSocket`, `Worker`, `indexedDB`, `navigator` all `undefined`; `document` is an inert two-method facade.
- All four `Function`-constructor escape routes blocked, including via an endowed host function; direct `eval` in source rejected (`SES_EVAL_REJECTED`); `import.meta.loader` stripped; intrinsics frozen and the host `Object.prototype` stays clean.

## High-severity findings

### 1. Authored `static getComponent` override executes as a Direct (host-authority) render — live in prod, with a working PoC

Verified end-to-end through the real renderer, not just by inspection. The chain:

- `static getComponent` is an ordinary overridable static on `BaseDef` (`card-api.gts:2631`).
- `DirectBoxelRuntime.getRenderSlot` calls it on the *authored* class: `card.constructor.getComponent(card, field, options)` (`direct-boxel-runtime.ts:264`), and wraps the result as `{ owner: 'direct', component }`.
- The renderer renders that as `<this.directComponent>` under `data-boxel-execution='direct'` (`boxel-execution-renderer.gts:591/605 → 718 → 955`), reached whenever a Capsule card does not author a requested format (the trusted-base fallback at line 590). `fieldPortalsFor(card)` also calls `getRenderSlot(card)` unconditionally on every first render, every tier (`boxel-execution.ts:1051`).

I wrote an integration test with a user-realm card that (a) carries no browser signal, so it classifies **Capsule**, (b) authors no template, so the trusted-base fallback fires, and (c) overrides `static getComponent` to return a component that reads ambient browser authority. Result — 5/5 assertions:

- the authored override's component rendered its own markup, **under `data-boxel-execution="direct"`**;
- inside it, `window` = object, `fetch` = function, `document` = object, `localStorage` = object — real ambient authority;
- it read Host-app DOM **outside its own subtree** (the `#qunit` container it never rendered).

To keep the module Capsule-classified while doing this, the PoC reads globals via the `globalThis.` spelling — which the classifier does not detect (see finding under Coverage). So this composes cleanly: classify Capsule, override `getComponent`, render in the Host document with the full browser authority the Capsule tier exists to deny.

Fix: resolve the trusted ancestor and call *its* static, or assert `card.constructor.getComponent === BaseDef.getComponent` before taking the Direct path for a non-trusted module. The same applies to `buildFieldPortals`' `rootComponent`.

### 2. Sandbox `@cardstack/` prefix + `..` traversal → authenticated cross-realm read — latent (tier off in prod), live in dev

`isTrustedModule` does `moduleIdentifier.startsWith('@cardstack/')` on an *unnormalized* bare specifier (`trusted-modules.ts:36`), while resolution normalizes `..` through the realm import map. So `@cardstack/base/../<other-realm>/private.json` passes the allowlist, resolves out of `/base/`, and the parent fetches it with `authedFetch` (which attaches a token for any realm the user has a session for). The child receives the full body. Fix: reject `..`, or resolve first and check the resolved URL with `isURLWithin`.

### 3. Sandbox media lane is ungated and leaks non-2xx bodies — latent/dev

`respondMedia` is dispatched *before* any allowlist check (`sandbox-fetch-transport.ts:300`) and `fetchBoundedSandboxMedia` (line 71) checks only protocol/content-type/size — no URL allowlist — using the authenticated fetch. The content-type gate is `if (response.ok && !contentType…)` (line 94), so any 401/403/404/500 body passes through to the child regardless of type: a general read/error oracle for arbitrary URLs. No test covers this lane. Fix: gate media against the projected document's declared URLs and drop the `response.ok` short-circuit.

**Framing on #2/#3:** the Sandbox tier is inert in staging/prod today — `BOXEL_SANDBOX_RUNTIME_URL` is set nowhere in the repo or infra, so the tier fails closed there. These ship the moment the tier is enabled, and are exploitable against a developer's own session in local dev now. Note also: the child runs authored source through a bare `eval()` with no SES (`boxel-sandbox-runtime.gts:469`) — the iframe is the only boundary, so every child-side check is a non-control and every parent-side gate must assume a fully malicious child holding the port. There is no CSP on the child route, and `credentialless` is Chromium-only — on Firefox/Safari the child gets persistent storage and can register a service worker on the sandbox origin.

## Medium findings

- **The "data vs rendering" boundary is coherent but the docs understate it.** By design the host still evaluates every authored module, its statics, computeds, getters, and serialize hooks — in the trusted realm, with a credentialed `import.meta.loader` reachable. That is the same as main *except* `resolveBoxelFields` walks every computed eagerly on each materialization (`boxel-projection.ts:309`), which is more host-side authored execution than main does lazily. The reviewer's guide frames host-side work as "deserializing JSON"; it should say plainly that the tiers contain template/DOM execution, not module evaluation.
- **Capsule has no termination boundary.** Verified via probe: card code held the main thread ~180ms with zero host timer ticks. The default tier runs on the main thread; a card can wedge the tab.
- **Dynamic import fails closed but opaquely.** Verified: the transpiler rewrites `import()` → `import.meta.loader.import`, loader is stripped, and the card throws `Cannot read properties of undefined (reading 'import')` at runtime — no policy message to the author.

## Coverage / correctness (lower severity)

- **`getComponent` bypasses the router on several live surfaces** not in the branch's own audit — `fitted-format-gallery.gts:45`, `workspace.gts` (x4), `cards-grid.gts` (x3), markdown/codemirror embeds. Some are trusted Base modules rendering *authored* cards directly. If containment is meant to be an invariant rather than a default, `BaseDef.getComponent` has to be the gate.
- **The classifier misses obvious DOM spellings** — `globalThis.document`, `self.document`, destructured `{ document }`, computed member access — classifying Capsule instead of Sandbox. Normally fail-closed (SES denies the capability, the card just breaks), so it is a smoothness/compat issue — but combined with finding #1 it is load-bearing: the `globalThis.` miss is exactly what lets the PoC keep Capsule classification while reaching host authority. It also contradicts RP-6.4's "not regex over source text" claim (the detector is regex + an AST narrowing pass).
- **Classification is import-order dependent** (a `visited` short-circuit returns a capsule result), so the same graph can classify differently based on statement order. Fail-closed but a real nondeterminism bug.
- **Trusted set is wider than documented** — skills and openrouter realms also reach Direct via the `@cardstack/` alias; and the Direct gate (`isTrustedModule`) and the classifier predicate (`isTrustedImport`, which includes the whole icons origin and catalog) differ. `trusted-modules.ts` — now the single load-bearing trust predicate — has no unit test.

## Process / verification apparatus

- **The RP bijection CI gate is currently RED.** `node scripts/check-rp-bijection.mjs` exits 1 (three test titles cite no RP id), and `ci-lint.yaml:43` runs it.
- **The bijection is citation-matching, not semantic** — a line-regex for `test('… RP-x.y …')`, satisfiable by `assert.ok(true)` or even a commented-out test, and the strongest tests (fetch-transport, capsule-registration) live in files the checker does not scan. But the underlying tests are genuinely rigorous (real production code, adversarial inputs, side-effect assertions). The fair read is "under-verified rigor, not theater." The capability matrix's build-status marks are hand-maintained and drift (e.g. `viewCard` marked deferred but built; RP-9.1 / RP-7.6 marked built but uncovered).
- **The "~75 lines deleted / main is surrounded" claim is ~125 lines** across 27 files (78 excluding the loader refactor). No deletion removes a security control — the claim's character holds, but the number will not survive `git diff --numstat`.
- **The Sandbox tier has no deployable configuration.** `BOXEL_SANDBOX_RUNTIME_URL` is set in no mise task, CI, compose, or infra, so DOM/WebGL cards render as an error in staging/prod today (fail-closed, not silent — an improvement, but the tier is effectively dead there until deploy wiring lands).

## Two things I could not fully resolve (flagging, not asserting)

- **3 of the branch's own `Capsule module registration` tests fail in my worktree** (deterministic, fail in isolation) — all about trusted-facade module resolution (e.g. `ember-provide-consume-context`, which is in the trusted list at `trusted-modules.ts:58`). I copied a possibly-stale `boxel-ui`/`boxel-icons` dist from main and no CI has run on the branch to cross-check, so I cannot cleanly attribute these to the branch vs my environment. Worth confirming a clean run.
- **Write-lane `document.included` is unvalidated** — only `data.id` is identity-checked. Whether an attacker-supplied `included` can mutate a *different* store instance was not traced to a conclusion. If it can, the write identity check is defeated. Worth an explicit answer.

## Suggested priority order

1. Close the `getComponent` hole (#1) — live in prod, defeats the core claim, PoC attached.
2. Fix the Sandbox `@cardstack/` traversal (#2) and gate the media lane (#3) before the tier is enabled.
3. Add a CSP + `frame-ancestors` on the child route; feature-detect `credentialless` or refuse the tier without it.
4. Get the bijection gate green and decide whether to make it semantic; reconcile the capability-matrix drift.
5. Rewrite the "data vs rendering" framing to match what the code actually does.
6. Add a `trusted-modules.ts` unit test — it is now the single load-bearing trust predicate.

## Appendix: PoC for finding #1

Integration test (all data synthetic). A Capsule-classified card overrides `static getComponent`; its component renders under `data-boxel-execution="direct"` with real `window`/`document`/`fetch`/`localStorage` and reaches host DOM outside its subtree. 5/5 assertions pass against the real renderer.

```gts
const escapeSource = `
  import { CardDef, contains, field } from 'https://cardstack.com/base/card-api';
  import StringField from 'https://cardstack.com/base/string';

  class HostEscapeComponent extends GlimmerComponent {
    get hostReach() {
      // 'globalThis.' spelling keeps the classifier from seeing a browser
      // signal, so this module stays on the Capsule tier.
      let g = globalThis;
      let diag = {
        window: typeof g.window, document: typeof g.document,
        fetch: typeof g.fetch, localStorage: typeof g.localStorage,
        sawHostDom: false,
      };
      try { diag.sawHostDom = Boolean(g.document && g.document.querySelector('#qunit')); }
      catch (e) {}
      globalThis.__rpEscapeDiag = diag;
      return diag.sawHostDom ? 'host-dom-reached' : 'contained';
    }
    <template>
      <div data-test-getcomponent-escape data-test-owner={{this.hostReach}}>
        authored getComponent override executed in the Host document
      </div>
    </template>
  }

  export class Escape extends CardDef {
    static displayName = 'Escape';
    // No authored template for any format -> the renderer takes the
    // trusted-base fallback and calls this override to get the Base slot.
    static getComponent(_card, _field, _opts) { return HostEscapeComponent; }
  }
`;
```

The full test (harness setup, assertions) is on my branch as `packages/host/tests/integration/components/rp-getcomponent-escape-test.gts`, alongside `packages/host/tests/unit/lib/capsule-boundary-probe-test.ts` (the 23-assertion Capsule boundary probe). Happy to open these as a PR against this branch if useful.

habdelra:
[Claude Code 🤖] Pushed the two test artifacts referenced in the review above to this branch (commit `548a0a8903`).

**`packages/host/tests/integration/components/rp-getcomponent-escape-test.gts`** — the PoC for finding #1, runnable in the real renderer. It builds a Capsule-classified user card that overrides `static getComponent` and authors no template, then asserts its component renders under `data-boxel-execution="direct"` with real `window`/`document`/`fetch`/`localStorage` and reaches Host-app DOM outside its own subtree. How it helps: it turns finding #1 from a claim into an executable reproduction, and it is written to **fail the moment the hole is closed** — when `getComponent` is gated to the trusted ancestor, the component no longer renders under `direct` and the test goes red. So it doubles as the regression guard for the fix. (It passes today; green here means the escape reproduces, not that the tier is safe — the commit message says so too.)

**`packages/host/tests/unit/lib/capsule-boundary-probe-test.ts`** — a 23-assertion adversarial probe of the Capsule compartment. It locks in the containment invariants that must never regress (no ambient authority, Function-constructor escapes blocked, direct eval censored, `import.meta.loader` stripped, intrinsics frozen), and separately pins two current gaps (opaque dynamic-import failure; main-thread occupancy with no termination) so that closing either one flips the test red. How it helps: it gives the containment claims a standing, attacker-side test that fails loudly on a regression, complementing the RP conformance suite (which tests mostly from the positive/routing side).

Both are lint-clean and green locally. Neither is wired into the RP bijection gate — happy to add RP-id citations to the titles if you want them counted there, or to leave them as standalone characterization tests. Also glad to split the PoC out or mark it pending/skip if you'd rather it not run green in CI while the finding is open — your call on which reads better for the team.

christse:
@habdelra Thanks for the adversarial review and executable PoCs. The full response batch is now pushed, current with `main`, and ready for another pass.

What changed:

- Closed the authored `static getComponent` Direct escape by resolving trusted ancestors at the render boundary; the supplied PoC is retained as an inverted regression test.
- Canonicalized trusted package admission and reject traversal/encoded traversal before resolution.
- Restricted Sandbox media reads to projected resource URLs, reject every non-2xx response, and removed the credentialed fallback.
- Canonically validate read-only `included` relationships on Sandbox writes.
- Refuse Sandbox without a configured distinct origin or `credentialless` support; local serve/preview installs the restrictive CSP, referrer, and nosniff headers.
- Expanded classifier coverage for `globalThis`, `self`, computed members, and destructuring, and made graph classification deterministic.
- Added a named Capsule dynamic-import denial instead of the opaque missing-loader failure.
- Corrected the data-vs-rendering documentation, Capsule CPU limitation, capability matrix, and numerical claims.
- Restored the RP bijection gate and the Capsule registration suite (15/15 in the focused batch).
- Adapted the branch to current `main`'s multi-realm query API and consolidated Boxel UI package layout.

Verification:

- [CI Lint is green across all packages](https://github.com/cardstack/boxel/actions/runs/31538923254), including the RP statement/test bijection.
- The focused reviewer-response batch was 89/89 across trusted modules, Sandbox fetch/media/write, classifier/engine, Capsule boundary, Sandbox process, Capsule registration, and Direct runtime suites.
- Host staging preview: https://codexboxel-execution-runtime-architecture.boxel-host-preview.stack.cards/
- Host production preview: https://codexboxel-execution-runtime-architecture.boxel-host-preview.boxel.ai/

Remaining limits are explicit rather than papered over: Capsule still has no CPU termination boundary; Host-side authored model semantics remain part of Boxel's current data model; and hosted Sandbox stays fail-closed until the preview/production edge can serve the child on a distinct origin with the required response headers. The current host previews are therefore useful for Capsule testing, but do not falsely advertise a production-safe Sandbox deployment.

Please re-review the branch at `b099e65628` when you have a chance. The PR remains a draft and is deliberately marked **DO NOT MERGE** while we use it for preview/corpus evaluation.

christse:
@habdelra The hosted Sandbox deployment follow-up is now pushed at `59ca6384ff` and ready for re-review.

What changed:

- Added a deployable Cloudflare Worker boundary for per-process 128-bit nonce origins, credentialless Host bootstrap/assets, parent-origin allowlisting, restrictive CSP/referrer/permissions/nosniff policy, and explicit API/auth/service-worker denial.
- Wired staging and production Host environments to `boxelusercontent.dev` and `boxelusercontent.com` respectively.
- Preserved the prior preview runtime route while adding the current `/_boxel-sandbox-runtime` route, so the older preview remains usable.
- Fixed the interactive CardPrerender auth context so a cold protected realm source can silently recover its session without weakening dedicated prerender behavior.
- Added focused origin-allocation, auth-context, and Worker boundary tests.

Verification:

- Worker boundary suite: 9/9.
- Staging and production-preview S3 deployment jobs succeeded for this commit.
- Both preview parent origins return 200 through the nonce Worker bootstrap.
- Signed-in staging smoke test: workspace chooser → FORGE Gym Shift Board → Library → isolated card rendered in a nonce `boxelusercontent.dev` iframe → Edit → Finish Editing → isolated iframe rendered again, with no 401 errors.
- Production preview is live and its compiled config points to the DNS-ready `.dev` preview edge.

The PR remains draft and **DO NOT MERGE**. Real production stays configured for `.com`; its Worker is deployed and root health is green, but nonce subdomains still need the proxied wildcard DNS record documented in the branch. Only production PR previews temporarily use the `.dev` edge so both preview builds work now.

Please re-review at `59ca6384ff` when you have a chance.

christse:
Small final follow-up: `b5076969c2` removes the Sandbox child's unnecessary regular Host Store startup, which had produced a CSP-blocked direct `base/card-api` fetch even though rendering succeeded through the broker-backed child store.

Both preview deploy jobs succeeded again at the new head. I repeated the signed-in Library → FORGE card → Edit → Finish Editing → isolated-card flow on the rebuilt staging artifact: nonce iframe present, card rendered before and after the toggle, and zero browser-console errors. Please use `b5076969c2` for re-review.

christse:
Follow-up fix is deployed at head f94b650238. This closes the transient private-realm source 401 that could occur when an interactive Boxel execution source load overlapped background prerender state: interactive source loads now enter a concurrency-safe reauthentication scope, while the dedicated prerender app remains storage-only.

Deployed verification on the production preview URL:
- Preview Host workflow succeeded for both staging and production S3 deployments: https://github.com/cardstack/boxel/actions/runs/31545190594
- Live bundle changed from the failing main-6el6aaSJ.js to main-CTKWSYDM.js
- FORGE Week 32 board rendered in its nonce boxelusercontent.dev iframe on 3 consecutive full reloads
- Edit -> Finish Editing -> isolated return rendered correctly
- No Unable to render panel and no source 401 in browser errors
- Focused authorization middleware tests: 4/4 pass

@habdelra please re-review the latest head when convenient. The PR remains draft and explicitly DO NOT MERGE; the preview is ready for broader corpus testing.

christse:
Google Fonts parity follow-up is pushed and deployed at head 38fe0e693a.

Change:
- Sandbox stylesheet policy now permits only https://fonts.googleapis.com.
- Sandbox font policy now permits only https://fonts.gstatic.com.
- No Google wildcards and no change to script, connect, image, media, worker, or frame lanes.
- Local sandbox headers match hosted behavior.
- The reviewer guide and Worker README explicitly record the accepted Google-only URL/IP/user-agent egress; credentialless and no-referrer remain in force.

Verification:
- Worker boundary tests: 9/9, including exact-origin and no-wildcard assertions.
- Focused Worker ESLint: clean; Host JS/template lint clean. Aggregate Host type lint still reports only the existing unrelated Array.at target errors.
- Wrangler 4.120.1 dry-runs succeeded for both environments.
- Staging Worker deployed: ccdeea9e-9352-4754-b7c3-3481a79bfa5e.
- Production Worker deployed: 066765ab-fc2c-4605-a8fb-e7bd4409d316; root health is green. Production nonce wildcard DNS is still outstanding as documented.
- Live staging header contains the two exact origins.
- FORGE Week 32 preview and real staging both report Barlow Condensed and Manrope available, with matching proportional glyph/layout metrics and no font/CSP errors.
- Fresh staging and production Host preview deployments succeeded: https://github.com/cardstack/boxel/actions/runs/31548037633

@habdelra please include this latest head in the re-review. The PR remains draft and DO NOT MERGE.

habdelra:
Most of the host tests failwith the same error: 
```
  Error: Assertion Failed: You cannot use the same root element (#ember-testing) multiple times in an Ember.Application
                at assert (http://localhost:7357/assets/debug-DPwvJxt5.js:216:20)
                at EventDispatcher.setup (http://localhost:7357/assets/index-DaCZcoCY-CxZu_iBx.js:273:57)
                at ApplicationInstance.setupEventDispatcher (http://localhost:7357/assets/destroyable-BH81RkYa.js:644:14)
                at ApplicationInstance._bootSync (http://localhost:7357/assets/destroyable-BH81RkYa.js:569:35)
                at http://localhost:7357/assets/route-C5UGqfZI.js:1123:17
                at initializePromise (http://localhost:7357/assets/service-BmZUuF2w.js:7828:3)
                at new Promise (http://localhost:7357/assets/service-BmZUuF2w.js:8251:30)
                at ApplicationInstance.boot (http://localhost:7357/assets/route-C5UGqfZI.js:1122:23)
                at http://localhost:7357/assets/dist-DZ-M9wRU.js:180:79
                at invokeCallback (http://localhost:7357/assets/service-BmZUuF2w.js:7812:11)
```
could you have your agent try to address this? it seems critical

christse:
### Direct RP regression coverage

This update deliberately moves the affected Host regression tests through the Direct Render Protocol path instead of continuing to instantiate card components through the legacy shortcut.

- Direct RP is now the canonical trusted execution path exercised by the existing preview, Store live-update, query, loader, indexing, and timestamp regressions.
- The behavioral assertions are retained: these changes update the entry path, not the expected product behavior.
- Capsule and Sandbox retain dedicated conformance, equivalence, lifecycle, and boundary suites.
- Trusted Host/package shims are treated as execution terminals rather than authored realm invalidation edges.
- The test-document environment invariant is checked explicitly so CI cannot silently autoboot a second Ember application on `#ember-testing` again.

Local pre-push evidence: Host lint passed; Base lint passed; Direct preview 10/10; RP conformance 15/15; execution engine 20/20; smoke-runner unit tests 9/9; environment-config verification passed. The accompanying wild-corpus runner contains URLs and observations only—no private workspace files.

habdelra:
looks good! feel free to push on teh CI failures as much as you want. but i think we are probably over the hump in terms of a prototype

christse:
CI follow-up in fc5bf8944a:\n\n- Host regression call sites continue to exercise their original product behavior through the Direct RP adapter. The nested-card editor assertion now keys off the actual isolated-format fixture marker instead of assuming the pre-RP direct-child DOM topology.\n- Direct RP now retains a same-card/same-format presentation across resource acknowledgements, with focused tests for DOM/input identity and edit→isolated format transitions.\n- Authored relationship portals keep present authored values on the execution router while delegating only terminal broken-link slots to trusted Base; coverage includes a live present→broken transition and a real realm-backed 404 inside a Capsule.\n- Software Factory treats repeated opencode transport failures as terminal blocked results instead of mistaking a dead child for idle and retrying until the Playwright timeout.\n\nLocal verification: Host lint; Host Vite test build; RP continuity 12/12; focused relationship tests; formerly failing nested create/edit workflow; Software Factory JS/format lint and new transport-liveness tests. Private workspace files and unrelated dirty files were not committed.

habdelra:
@christse a key aspect that we still need to answer is how does the store work now. i think there is still a central store in this branch, meaning that trusted code and untrusted code can comingle. how can we have card composition in our store and still sandbox untrusted code? like we may just be moving the security gap to a different place now…

habdelra:
[Claude Code 🤖] Closing: this prototype was never meant to merge, and the work derived from it is being backed out in #5994. The branch `codex/boxel-execution-runtime-architecture` stays as the archived reference for the prototype and its docs.

PR #5743, opened 2026-08-11
Add FileDef format regression harness and openly licensed fixture corpus

lukemelia:
## What this adds

A new `packages/experiments-realm/filedef-fixtures/` directory containing a openly licensed fixture corpus and the review surfaces that render it, for auditing FileDef rendering quality across all four card formats:

- **`samples/` — 108 openly licensed fixtures** (36 file types × simple/moderate/complex tiers). The realm indexes each file directly as a FileDef instance — typed by extension where a subclass exists, generic `FileDef` otherwise — so the corpus needs no wrapper cards, and a family's pages light up as its subclass lands. All files sit under the realm's 5 MiB limit.
- **`format-preview.gts` + `FormatPreview/*.json` — the format regression harness.** One instance per file type carries that type's three fixture tiers; the isolated view renders the selection across a tier-comparison row, the 16 canonical fitted envelopes (dimensions mirror `FITTED_FORMATS` in runtime-common, which card modules cannot import), and atom/embedded/isolated panels. `FormatPreview/all-file-types.json` is the cross-type index of complex fixtures.
- **The embedded field guide** (`file-embedding-field-guide.gts` + `embedded-file-field-guide.md`): one reviewable document rendering every leaf type's embedded format through relative `::file` references, with each family's section stating the review contract for a correct rendering.
- **`SOURCES.md`** — provenance, license, and attribution for every fixture (public domain NASA/Wikimedia, Blender Open Movie CC BY 3.0, Apache POI, MIT/OFL/CC0 sources), plus a byte-length/SHA-256 integrity manifest.
- **`QUALITY-AUDIT.md`** — the acceptance gates each fixture satisfies and what each complex sample exercises, so a format-detector fixture can't masquerade as renderer validation.

## Byte stability

`samples/` is registered in `.prettierignore` and the lint-staged verbatim-tree skip: fixture bytes must match the integrity manifest, so formatting tools are deliberately kept away from that tree. The five code fixtures are lightly adapted from their originals (annotation cleanup; one import repointed to `file-formats/file-type-profile` in the base realm) so they module-index cleanly; the manifest records their adapted bytes.

## Test plan

- `pnpm lint:hbs` in experiments-realm passes across the new `.gts` files (including the `.sample.gts` fixtures).
- On a local dev stack, the directory indexes with zero error rows (152 file entries, 38 instances) and prerenders with zero errors (190 `prerendered_html` rows). Spot checks: the AVIF harness page renders real `` elements in atom/embedded/isolated; the ZIP page (no subclass yet) renders all 19 fitted frames plus embedded/isolated through the generic file shells; the field guide prerenders all 36 `::file` slots.
- Integrity: every manifest row re-verified against on-disk bytes after commit.

## Notes for review

- The corpus adds ~45 MB of committed binaries. Placement in the experiments realm means the fixtures also exercise the real indexing/prerender pipeline on deployed environments.
- The in-review 3D-model branch adds `model-samples/` STL/3MF example files to this realm; that overlaps with `samples/` here and is worth reconciling at merge.


packages/experiments-realm/filedef-fixtures/format-preview.gts
```diff
@@ -0,0 +1,736 @@
...
+  },
+  {
+    label: 'Cards',
+    formats: [
+      { key: 'card-c', name: 'Compact Card', w: 400, h: 170 },
+      { key: 'card-f', name: 'Full Card', w: 400, h: 275 },
+      { key: 'card-x', name: 'Expanded Card', w: 400, h: 445 },
+    ],
+  },
+];
```
habdelra, line 56: [Claude Code 🤖] **Confirmation — the 16 envelope specs here match `FITTED_FORMATS` exactly, verified spec-by-spec.** Since this list is hand-copied from `runtime-common/formats.ts` (card modules can't import it), I checked every entry against the source:

- Badges: 150×40 / 150×65 / 150×105 ✓
- Strips: 250×40 / 250×65 / 250×105 / 400×65 / 400×105 ✓
- Tiles: 150×170 / 250×170 / 170×250 (CardsGrid) / 150×275 / 250×275 ✓
- Cards: 400×170 / 400×275 / 400×445 ✓

All 16 dimensions agree, so the fitted matrix renders each file at its true host envelope. **This is the harness's load-bearing correctness property, and nothing ties the two lists together** — if someone edits a spec in `runtime-common/formats.ts`, this copy (and the paired `.size-*` CSS blocks below) drift silently and the "regression harness" would render the wrong boxes without any signal. Non-blocking, but a lightweight unit test in `runtime-common` that asserts this list deep-equals `FITTED_FORMATS` (or a generated constant) would make the mirror enforced rather than trusted. Follow-up, not this PR.

---

packages/experiments-realm/filedef-fixtures/format-preview.gts
```diff
@@ -0,0 +1,736 @@
...
+  static prefersWideFormat = true;
+
+  @field title = contains(StringField);
+  @field previewFiles = linksToMany(() => FileDef);
+
+  @field cardTitle = contains(StringField, {
+    computeVia: function (this: FileDefFormatPreview) {
+      return this.title ?? 'FileDef Format Preview';
+    },
+  });
```
habdelra, line 657: [Claude Code 🤖] **Optional simplification (non-blocking): the custom `title` field + `cardTitle` override are redundant here.** The base `CardDef` already defines `cardTitle` as a computed that reads `cardInfo.name` (`packages/base/card-api.gts`, ~L3269: `this.cardInfo.name?.trim()?.length ? this.cardInfo.name : \`Untitled ${displayName}\``). Every `FormatPreview/*.json` instance sets `cardInfo.name`, and I verified across all 37 that `attributes.title` is byte-identical to `attributes.cardInfo.name` (e.g. `avif.json` → both `"avif · Format Preview"`). So this override recomputes the exact value the base already yields, and each instance carries the same string twice — two fields that can drift apart under a later edit, with `cardTitle` silently preferring `title`.

Dropping `@field title`, the `cardTitle` override, and the duplicated `title` attribute in the 37 FormatPreview instances would let `{{@model.cardTitle}}` resolve through the base and remove the duplication.

Note this does **not** apply to `file-embedding-field-guide.gts`: there the instance sets `title: "FileDef embeds, family by family"` while `cardInfo.name: "File Embedding Field Guide"`, so the override is doing real work — decoupling the rendered `<h1>` from the catalog name. That one is intentional; only the FormatPreview copy is redundant.

---

lint-staged.config.mjs
```diff
@@ -23,7 +23,12 @@ const PNPM_OWNED_FILES = new Set(['pnpm-lock.yaml', 'pnpm-workspace.yaml']);
 // it and drifts the copy from its source. Regenerating (`pnpm build:skills`)
 // then reports phantom diffs. Match on a path segment so absolute staged paths
 // work.
-const VERBATIM_COPIED_TREES = ['/packages/boxel-cli/plugin/skills/'];
+const VERBATIM_COPIED_TREES = [
+  '/packages/boxel-cli/plugin/skills/',
+  // Licensed FileDef sample files whose bytes must match the integrity
+  // manifest in the sibling SOURCES.md.
+  '/packages/experiments-realm/filedef-fixtures/samples/',
+];
```
habdelra, line 31: [Claude Code 🤖] **Confirmation — the verbatim skip is correct and doesn't hide anything CI enforces.** `isVerbatimCopy` matches on `posix.includes('/packages/experiments-realm/filedef-fixtures/samples/')`, so the `.gts`/`.ts` code fixtures under `samples/` bypass eslint and ember-template-lint at commit time. I checked what that means against CI: `ci-lint.yaml` runs `pnpm run lint` in `packages/experiments-realm`, whose only `lint:*` script is `lint:hbs` (`ember-template-lint .`) — there is no eslint step for this package. So:

- The samples are still swept by `ember-template-lint .` in CI (the skip only affects the pre-commit hook), and the PR's test plan confirms that passes.
- They are never eslint/prettier-checked anywhere — not by the hook, not by CI — so the commit-time skip removes no gate that would otherwise catch a drift-from-source reformat.

That's exactly the intended outcome (keep formatters off the byte-stable tree without losing a real check), so no change requested. Worth being aware the same holds for the *harness* `.gts` files outside `samples/` (`format-preview.gts`, `file-embedding-field-guide.gts`): experiments-realm has no eslint, so those are template-lint-only too — a pre-existing property of the package, not this PR.

---

packages/experiments-realm/filedef-fixtures/SOURCES.md
```diff
@@ -0,0 +1,160 @@
...
+| [Khronos glTF Sample Assets](https://github.com/KhronosGroup/glTF-Sample-Assets) | Model-specific: Box public domain; Duck CC BY 3.0 | Real glTF and GLB scene assets.                                                                |
+| [Mutopia Project](https://www.mutopiaproject.org/)                               | Public-domain editions for selected scores        | Bach and Beethoven MIDI files, including a 15-track General MIDI arrangement.                  |
+
+## Boundary fixtures
+
+[Sindre Sorhus/file-type](https://github.com/sindresorhus/file-type/tree/main/fixture) (MIT) and [Mathias Bynens/small](https://github.com/mathiasbynens/small) (CC0) supply minimum-valid, detector, truncation, and empty-file fixtures. They are not used as evidence that a complex renderer is complete.
+
+## Integrity manifest
+
+Fixture bytes are committed to the repository and served by the realm directly from disk. The `samples/` tree is exempt from repository formatting tools so these bytes stay stable; any intentional change to a fixture must update its row here. A mismatch between this manifest and the file on disk means the fixture no longer matches its recorded provenance.
```
habdelra, line 49: [Claude Code 🤖] **Follow-up (non-blocking): the integrity manifest is accurate today but nothing enforces it going forward.** I spot-checked several rows against the committed bytes and they match exactly — e.g. `avif-simple.avif` (871 bytes, `c627d127…`), `csv-simple.csv` (84, `4aa41ce8…`), `gts-complex.sample.gts` (1123, `b4ec7a2d…`), `json-simple.data.json` (1, `5feceb66…`) — so the manifest is correct as committed.

The gap is that this table is documentation only. `.prettierignore` + the lint-staged skip keep formatters *away* from `samples/`, but I searched `scripts/` and `.github/workflows/` and found no check that recomputes these SHAs against disk. So the sentence just above — "A mismatch between this manifest and the file on disk means the fixture no longer matches its recorded provenance" — is upheld only by manual discipline; a future edit to a fixture that forgets to update its row drifts silently and the manifest quietly stops meaning what it says.

A tiny guard would make it load-bearing: a test in `experiments-realm` (or a CI step) that hashes each `samples/*` file and diffs against this table. Worth filing; out of scope for this PR.

---

packages/experiments-realm/filedef-fixtures/SOURCES.md
```diff
@@ -0,0 +1,160 @@
+# Fixture sources and licenses
+
+Every supported FileDef leaf type has simple, moderate, and complex fixtures in this directory. The [integrity manifest](#integrity-manifest) at the end of this document records each file's byte length and SHA-256; the tables below record source, license, and attribution.
+
+The three levels have different jobs:
+
+- **Simple** is a minimum-valid or boundary fixture. Tiny files are intentional here.
```
habdelra, line 7: [Claude Code 🤖] **Minor doc/fixture inconsistency (non-blocking): the "simple = minimum-valid, tiny files intentional" rule doesn't hold for the photographic families.** From the manifest, the "simple" fixture is the *largest* of its three tiers for JPEG and WebP:

- `jpeg-simple.jpg` = **3,066,842 bytes** vs `jpeg-moderate` 218,565 and `jpeg-complex` 1,237,372 — the simple tier is ~2.5× the complex one and the biggest JPEG in the set.
- `webp-simple.webp` = **705,444 bytes** vs `webp-moderate` 518,966 and `webp-complex` 701,620 — simple is again the largest.

That reads as either simple/complex being swapped for these families, or (more likely) the tier taxonomy for real-photo families being about visual/structural variety rather than the size ladder this bullet and QUALITY-AUDIT §"Intentional simplicity of the simple tier" describe. Both docs currently state the size-ladder framing without a carve-out, so a reader auditing the corpus hits a contradiction. Either swap the JPEG/WebP simple⇄complex assignments or add one sentence noting the photo families use three distinct real images and don't follow the minimum-valid→representative size progression. Cosmetic; not blocking.

---

packages/experiments-realm/filedef-fixtures/SOURCES.md
```diff
@@ -0,0 +1,160 @@
...
+| samples/pptx-complex.pptx       | 68822   | 19fde9b87e33dd1a95fdbba0cf6abc2278bf03874f4665c7f8b88b6afe4a2571 |
+| samples/pptx-moderate.pptx      | 47219   | 44a7d14a137f59d4328659f38b8531477ccf09fb0559f4d2495ae8d8a308e7db |
+| samples/pptx-simple.pptx        | 30226   | daad647369762d66974948aa06600b45bf029b1020486d1b5455ba4cca5d7502 |
+| samples/stl-complex.stl         | 1286884 | 1952b004b0b3f58578df6021b92c4ff9d3d2d8f006f095711e6859edf73a5844 |
+| samples/stl-moderate.stl        | 107884  | 0012b62d4bb485f902a972192eef29d6ce4b97a9deb2de67a02d30a5b738f461 |
+| samples/stl-simple.stl          | 80861   | 5c0d95ca55352ccf5cca12197a5f9fa17eb2e905d1bb3a45c8ba51c9a22699a4 |
+| samples/svg-complex.svg         | 1379307 | 2d835fb5ca7086db175b89850e6bca609feb335f5214988b1467b09b35e4d984 |
+| samples/svg-moderate.svg        | 1294    | 7f023e8d1e767370306ec9437d2de5d5af97491d6208dd7799a3808a7c10a36a |
+| samples/svg-simple.svg          | 822     | 3bf8cceead820aec50d4ee825a3fd02c5a1cd6665cc9cf4cbf3d9c8861a204bb |
+| samples/text-complex.txt        | 19586   | 40b7538276e1ecf86d1224b93bd147a2819a4dd4c67f8650c825e7da757c8447 |
```
habdelra, line 128: [Claude Code 🤖] **Minor (non-blocking): `text-complex.txt` is byte-identical to `markdown-complex.md`.** Note this row's length/SHA (19586, `40b75382…`) exactly matches the `markdown-complex.md` row above it — I confirmed with `cmp` that the two files are identical bytes. So the "complex plain-text" fixture is literally the complex markdown file under a `.txt` extension.

That's defensible (Markdown *is* valid plain text, and it exercises the text renderer on real content), but it means the text-complex tier adds no coverage distinct from markdown-complex — the same content flows through whichever FileDef subclass the extension selects. If the intent was to exercise plain-text rendering on genuinely non-Markdown prose (no headings/tables/fences), a distinct `.txt` sample would test more; if the reuse is deliberate, a one-line note here would save the next auditor the double-take when two rows share a hash. Your call — not blocking.

---

habdelra (commented):
[Claude Code 🤖] Reviewed as a fixtures-and-harness change: the lens was (a) do the harness cards render what they claim, (b) is the corpus's integrity/provenance story true and durable, and (c) do the shared-infra changes (`.prettierignore`, `lint-staged.config.mjs`) do the right thing without hiding a real gate. I checked the head branch out in a worktree, traced the indexing path for the code/JSON fixtures, and verified the manifest and licensing claims empirically rather than reading them.

**Bottom line: no blocking issues — this is a clean, well-documented corpus and the harness is correct. Everything below is non-blocking discussion.** The scope is confined to the experiments realm and two byte-stability config edits; no production code path changes.

- **The fitted matrix is faithful.** All 16 envelope specs in `FORMAT_GROUPS` match `runtime-common/formats.ts` `FITTED_FORMATS` dimension-for-dimension, so the harness renders each file at its true host envelope (see the inline confirmation on `format-preview.gts`).
- **The indexing model is sound for every fixture kind.** I traced `index-runner/visit-file.ts`: a `.json` that isn't a card resource (`isCardResource` false — e.g. `json-*.data.json`, which are plain `package.json`-shaped bodies) falls through to a file-only extract with no error row, and every file (including `.ts`/`.gts` modules) gets a FileDef file render. The `(get @fields.previewFiles idx)` indexing the isolated view relies on is a supported pattern — plural fields are typed `BoxComponent[]` in `FieldsTypeFor`. The code fixtures' imports resolve (`file-api` re-exports `FileDef as default`; `typescript-moderate.ts`'s three imports exist in base `file-formats/file-type-profile`).
- **Integrity and provenance hold up.** Spot-checked manifest rows match on-disk bytes exactly; all 36 `::file[./samples/…]` references in the field guide resolve to real files; SOURCES.md attributes every family to a stable, openly licensed source.

1. Weigh the permanent cost of committing ~46 MB of binaries to git history — see "Adjacent" below; this is the one item genuinely worth a maintainer decision before merge.
2. Add a check that hashes `samples/*` against the SOURCES.md manifest so the integrity table is enforced, not just documented — inline on SOURCES.md L49.
3. Resolve the "simple tier" size contradiction for the JPEG/WebP photo families (swap simple⇄complex, or add a carve-out sentence) — inline on SOURCES.md L7.
4. Consider tying `FORMAT_GROUPS` to `FITTED_FORMATS` with a unit test so the hand-copied envelope list can't drift silently — inline on `format-preview.gts` L15.
5. Optional: drop the redundant `title` field + `cardTitle` override in `FileDefFormatPreview` and the duplicated `title` attribute across the 37 FormatPreview instances (the field-guide card's override is intentional and should stay) — inline on `format-preview.gts` L650.
6. Optional: note or de-duplicate that `text-complex.txt` is byte-identical to `markdown-complex.md` — inline on SOURCES.md L128.

**Adjacent, out of scope:**
- **~46 MB of binary fixtures, verified with `du`, are permanent in git history.** The PR acknowledges this. It's a deliberate trade (the fixtures exercise the real indexing/prerender pipeline on deployed environments, which is genuinely valuable), but every clone pays the weight forever, so it deserves an explicit maintainer sign-off rather than defaulting in. If the team wants the coverage without the permanent weight, git-lfs or out-of-repo asset storage are the usual levers — a follow-up decision, not a change requested here.
- **The `.ts`/`.gts` fixtures are live modules, not inert data.** `typescript-moderate.ts` imports `contentTypeForFile` / `extensionOfFile` / `profileForFile` from base `file-formats/file-type-profile`, and the `.sample.gts` fixtures define real `CardDef`/`FieldDef` subclasses in the experiments realm. That's fine and intended, but it means a future rename in that base module surfaces as an experiments-realm index error from a file that looks like a static sample — worth keeping in mind for whoever refactors `file-type-profile`.
- The PR notes an in-review 3D-model branch adds `model-samples/` overlapping this `samples/` tree; reconcile at merge as it already flags.

---

packages/experiments-realm/filedef-fixtures/format-preview.gts
lukemelia: [Claude Code 🤖] Agreed the mirror is trusted rather than enforced. Filed [CS-12493](https://linear.app/cardstack/issue/CS-12493/filedef-fixture-harness-guardrails-fitted-formats-parity-integrity) to add a parity test (deep-equal against `FITTED_FORMATS`, or generate the constant) — it also covers the integrity-manifest check from your `SOURCES.md` comment.

packages/experiments-realm/filedef-fixtures/SOURCES.md
lukemelia: [Claude Code 🤖] Agreed — the manifest is documentation-only today. Tracked in [CS-12493](https://linear.app/cardstack/issue/CS-12493/filedef-fixture-harness-guardrails-fitted-formats-parity-integrity) together with the `FITTED_FORMATS` parity test: a small check that hashes each `samples/*` file and diffs against the table, so the "mismatch means broken provenance" sentence becomes load-bearing.

packages/experiments-realm/filedef-fixtures/format-preview.gts
lukemelia: [Claude Code 🤖] Your analysis checks out — base `cardTitle` already computes from `cardInfo.name`, every FormatPreview instance sets it, and the guide card is the only place the override does real work. Dropping `@field title`, the override, and the duplicated attribute across the 37 instances is a mechanical change; leaving it to Luke to say whether to fold it into this PR or keep the diff frozen as-extracted from the handoff.

packages/experiments-realm/filedef-fixtures/SOURCES.md
lukemelia: [Claude Code 🤖] Right — for the photographic families the tiers are three distinct real photographs (varying orientation, color, metadata; see the "Real-world complex samples" table), not a size ladder, and the tier-definition bullet states the size framing without that carve-out. The one-sentence doc fix is the right resolution rather than swapping assignments; flagging for Luke since the fixture tree is intentionally byte-frozen and doc edits are the only safe touch.

packages/experiments-realm/filedef-fixtures/SOURCES.md
lukemelia: [Claude Code 🤖] Good catch on the shared hash. Markdown-as-valid-plain-text does exercise the text renderer on real content, but you are right it adds no coverage distinct from `markdown-complex` — a genuinely non-Markdown prose sample (or a one-line note in the manifest acknowledging the reuse) would each resolve the double-take. Leaving the pick to Luke.

FadhlanR (approved):
Approving. The corpus + preview harness is a clean, additive way to audit FileDef rendering across formats, and placing it in the experiments realm so it runs through the real index/prerender pipeline is the right call.

Reconciliation note: my 3D-model PR (#5658) currently adds its STL/3MF examples under `experiments-realm/model-samples/`, which overlaps with the `filedef-fixtures/samples/` fixtures here. Since I'll be merging after this lands, I'll own the reconciliation — I'll move #5658's 3D model samples into this `filedef-fixtures/samples/` tree, drop the separate `model-samples/` directory, and confirm the STL/3MF pages render through the typed `StlDef`/`ThreeMfDef` viewers. This PR needs no changes for that.

PR #5745, opened 2026-08-11
Render image FileDefs through the shared file-format shells

lukemelia:
Closes [CS-12237](https://linear.app/cardstack/issue/CS-12237/media-formats-image-jpeg-png-gif-webp-avif-svg)

## What changed

`ImageDef` previously carried four hand-rolled format templates that bypassed the shared file-format shells, so images got none of the shells' file-state handling, fitted-envelope adaptation, or metadata inspection, and the shells' `previewKind: photo/gif/svg` branches were unreachable. Following the FileDef handoff architecture:

- **`ImageDef` now supplies a `previewComponent` and inherits `FileDef`'s shell-based formats.** The new `file-formats/image-preview.gts` renders one native `` via the `FileImage` primitive; the four hand-rolled `image-def-*` templates are deleted. Net effect on every card's dependency graph (guarded by the realm-indexing deps test): four template modules out, `image-preview` + `file-resources` in.
- **Fit rules per format.** Fitted cells fill edge to edge (`cover`), letterboxing only proportions a center-crop would destroy (aspect ≥ 2.5 or ≤ 0.4) — mirroring the fitted shell's thumbnail-rail rule. Embedded/isolated never crop and never upscale a raster (`scale-down`). SVGs always letterbox at their own proportions (`contain`) since a vector is as likely a diagram as a picture. The fitted format now uses a real lazy-loaded `` instead of a CSS `background-image`.
- **Extracted metadata now surfaces.** The isolated inspector gains a **Color profile** pane beside EXIF, and the view model's facts row picks up color space and bit depth. This also fixes the view model reading `colorProfile` from inside `exif` — it is a sibling field on `RasterImageDef` — and strengthens the isolated shell's `hasExif` so an always-instantiated empty `contains` field no longer renders an empty EXIF pane.
- **Image extensions join the content-type fallback table** (`jpg/jpeg/png/gif/webp/avif/svg`), and `ImageDef` pins `fileFamily = 'image'`, so an image served without a useful MIME type (or a hand-linked thumbnail with no content type at all) still routes to the image presentation.

## Testing

- Updated the image-def acceptance suites to the shell markup (`[data-test-image-preview]`), and added assertions that the isolated shell hosts the preview, that SVG previews letterbox, and a new test that EXIF (camera name) and color profile (sRGB label, bit depth) render in the isolated inspector.
- Rewrote the `FileDef format templates` test that pinned the old ImageDef-bypasses-the-shell behavior to assert the new shared-shell behavior.
- Locally green: image-def acceptance suites, `FileDef format templates` (45), `realm indexing` (140), `card-basics` (431), `Unit | file-formats` (57), file-attachment integration. The pre-existing "authenticated images display in browser" tests fail on the local dev test page on baseline main as well (test-realm service worker doesn't serve image bytes headlessly); CI is the arbiter there.
- Percy will show visual diffs for image cards — expected with this change.

Thumbnail generation + responsive `srcset` renditions are split to [CS-12492](https://linear.app/cardstack/issue/CS-12492/image-thumbnails-generation-pipeline-responsive-renditions).


packages/base/file-formats/image-preview.gts
```diff
@@ -0,0 +1,84 @@
+// The image family's renderer, projected into the four format shells by
+// `FilePreviewStage`. The browser is the decoder, so this is one native
+// `<img>` via the `FileImage` primitive; the component's whole job is choosing
+// how the pixels meet the frame the shell hands it.
+import GlimmerComponent from '@glimmer/component';
+
+import { FileImage } from './file-resources';
```
habdelra: [Claude Code 🤖]

**Importing `FileImage` from `file-resources` pulls the whole media-primitives module into card-api's universal dependency graph.** `card-api.gts` now statically imports `image-preview` (its line 133), which imports `./file-resources` here. `file-resources.gts` is the ~530-line module that *also* defines `FileAudio`, `FileVideo`, `FileObject`, the `loadProtectedMediaBlob` fetch/blob transport, the seek-intent playback modifier, and the `FontFace` loader.

Verified: on `main`, `file-formats/file-resources` is absent from the realm-indexing deps list; this PR adds it (the deps-test diff reflects exactly that). So every card in every realm now carries `file-resources` as a module dependency purely to reach one native `` renderer.

That crosses the dependency-isolation discipline this subsystem documents for itself. `image-file-def.gts` and `metadata-fields.gts` both explain that image/metadata fields deliberately live one level *below* `card-api` so that "only realms that actually hold image files pay." Routing the audio/video/font transport machinery into the base graph to obtain an `` is the same cost that split was designed to avoid.

You already called out the graph change in the description, so this is a conscious decision rather than an oversight — but consider extracting `FileImage` (plus the small `fileResourceURL`/`stringValue` helpers it needs) into its own lean module that both `file-resources` re-exports and `image-preview` imports. card-api's graph then gains only the img primitive, while the media families keep the full `file-resources`.

Class: follow-up / design refinement. Non-blocking.

---

packages/base/file-formats/image-preview.gts
```diff
@@ -0,0 +1,84 @@
...
+  // the fitted shell applies to its thumbnail rail. The reading formats never
+  // crop and never upscale a raster; blowing a small image up to the stage
+  // would trade real pixels for blur.
+  get fit() {
+    if (this.isSvg) {
+      return 'contain';
+    }
+    if (this.args.mode === 'fitted') {
+      let ratio = this.args.model?.aspectRatio;
+      return ratio != null && (ratio >= 2.5 || ratio <= 0.4)
```
habdelra: [Claude Code 🤖]

**The "extreme aspect ratio → letterbox" rule is now encoded in two places that must stay in agreement.** Here `get fit()` returns `contain` for `ratio >= 2.5 || ratio <= 0.4` (and always for SVG). `file-shell-fitted.gts`'s `get thumbContain()` repeats the identical thresholds (`r >= 2.5 || r <= 0.4`, plus the `previewKind === 'svg'` case).

These aren't redundant-but-harmless: the fitted **stage** renders the image through this component, whereas the fitted shell's wide list-view **thumbnail rail** (`.fit-thumb`) renders its own `` and reaches for `thumbContain` instead. The two describe the same picture in two layouts, so they have to letterbox it the same way. The first time someone tunes `2.5`/`0.4` in one file and not the other, a panorama will `cover` in the tile stage while it `contain`s in the list-row thumbnail (or vice-versa).

Suggest hoisting the threshold (and the svg rule) into one shared helper both the shell and this component call.

Class: follow-up / drift-by-duplication. Non-blocking.

---

packages/base/file-formats/file-shell-isolated.gts
```diff
@@ -190,7 +190,27 @@ export class FileIsolatedShell extends GlimmerComponent<FileIsolatedShellSignatu
   // A `contains` FieldDef is always an instance, so presence has to mean "has
   // at least one meaningful value" rather than "is not null".
   get hasExif() {
```
habdelra, line 194: [Claude Code 🤖]

**Strengthening `hasExif` is the right fix, but the predicate tests a subset of the fields the EXIF pane can render — so a thinly-tagged photo now hides data it used to show.** Previously `hasExif` was `Boolean(model.exif)`, which is always true for the always-instantiated `contains` field, so the pane rendered even when empty — the real bug you're closing.

The new predicate checks `capture.cameraName | lensModel | iso | exposureTime.display | capturedAt` and `location.coordinates | place`. But `CameraCaptureField` (in `metadata-fields.gts`) also renders `focalLength`, `aperture`, `flash`, `orientation`, and `software`, and `GeoLocationField` renders `altitude`. An image whose EXIF carries only, say, aperture + focal length + flash — no make/model, ISO, exposure time, or capture time — now evaluates `hasExif` false and the entire pane is suppressed, even though `<@fields.exif />` would have shown those rows. Uncommon (most cameras write make/model), but it is a content regression against the old always-render behavior for that edge.

Consider widening the check to include those fields, or deriving presence from whether the rendered field would be non-empty.

For contrast, `hasColorProfile` right below is **complete** — `colorSpace.code | bitDepth | channels | iccProfile | hasAlpha != null` covers every value field on `ColorProfileField`, and the deliberate `hasAlpha != null` correctly keeps a genuine `false` ("no alpha channel") visible rather than laundering it into "unknown." That one's a good template for what `hasExif` could become.

Class: minor, now load-bearing (this PR rewrote the getter). Non-blocking.

---

habdelra (commented):
[Claude Code 🤖]

**Review lens:** whether `ImageDef` genuinely lands on the shared FileDef shell path (not just visually), whether the metadata surfacing is wired to real fields, and what the refactor costs the base dependency graph. I checked out the head branch, traced the routing end to end, and verified claims against the code rather than the description.

**Bottom line: functionally sound, well-tested, and the design is the right one — no blocking issues.** Everything below is non-blocking: one architectural refinement and a couple of follow-ups. CI is green except `percy/-cardstack-host` ("2 visual changes need review"), which is expected per your description.

### What lands right (verified, not courtesy)
- **The family-renderer indirection is clean.** `FileDef` declares the four format statics (`embedded`/`fitted`/`isolated`/`atom`); `ImageDef` drops its four overrides and sets `previewComponent = ImagePreview`; `filePreviewComponentFor` reads `previewComponent` off the *instance's* constructor, so a `linksTo(FileDef)` populated with a `JpgDef` still resolves the image renderer. The shells then project it through `FilePreviewStage`. Traced the whole path — it holds.
- **The `fileFamily = 'image'` pin does what the description claims.** For a MIME-less / octet-stream image, `profileForFile` would fall back to `generic`, but `fileProfileSource` overrides `family` to `image`, so `fileViewModel` sets `imageUrl = url` and the presentation still routes to the image renderer. The new extension-fallback rows (`jpg/png/gif/webp/avif/svg`) cover the MIME-less-but-extensioned case in parallel.
- **The `colorProfile` fix is a real bug fix, not a cosmetic move.** Confirmed in `image-file-def.gts`: `colorProfile` is a sibling `@field` on `RasterImageDef`, not nested under `exif`. The old view model read `file.exif?.colorProfile`, which was always `undefined` — so this pane could never have populated before. The new EXIF + color-profile acceptance test pins camera name, `sRGB`, and `8-bit`, and CI is green on it.
- **The tests re-pin behavior rather than shape.** The `FileDef format templates` test flipped from asserting "ImageDef bypasses the shell" to asserting the shell wraps a native ``; the realm-indexing deps test swaps the four `image-def-*` modules for `image-preview` + `file-resources`. Both would fail if the routing regressed.

### Recommendations (all non-blocking; detail in the inline threads)
1. **Keep `file-resources` out of card-api's universal graph.** Importing `FileImage` drags the whole media-primitives module (audio/video/font transport) into every card's deps for one ``. Consider splitting `FileImage` into its own lean module. — see the thread on the `FileImage` import in `image-preview.gts`.
2. **Deduplicate the letterbox thresholds.** `2.5`/`0.4` (and the svg rule) live in both `ImagePreview.get fit()` and `FileFittedShell.get thumbContain()`, which must agree for the tile stage and the list-view thumb rail to match. — see the thread on `get fit()`.
3. **`hasExif` checks a subset of the EXIF pane's fields**, so an image tagged only with aperture/focal-length/flash now hides its pane; `hasColorProfile` is complete and is a good template. — see the thread on `get hasExif()` in `file-shell-isolated.gts`.
4. **Confirm the atom change is intended.** By inheriting `FileAtomShell`, image atoms now render icon + name + extension — the old `ImageDefAtomTemplate`'s inline 20px `` thumbnail is gone. This follows the shell design ("atom mounts no domain renderer"), but it's a user-visible change not mentioned in "What changed," so it's worth an explicit confirm and a check that Percy covers the atom snapshots.

### Adjacent, out of scope
- **`padding: 0.375rem` on the contained/scale-down `.image-preview`** is applied to an `` with `width/height: 100%` and default `box-sizing: content-box`, inside `.stage { overflow: hidden }`. Without `box-sizing: border-box`, the padded box grows past the frame and is clipped, so the intended breathing room may not actually render. This mirrors the pre-existing `.fit-thumb.contain .thumb-img { padding: 6px }` in the fitted shell, so it's consistent with the house pattern and **not** a regression from this PR — flagging for whoever owns the shell CSS to verify visually (I didn't render it live).
- **The MIME-less hand-linked thumbnail** routes to the image presentation in the tile/embedded/isolated stages (`family = 'image'` → `imageUrl` set), but the fitted shell's wide list-view thumbnail rail (`thumbUrl`) gates on `previewKind ∈ {photo,gif,video,svg}`, and such a file keeps `previewKind = 'generic'` — so in the list layout it shows the type icon rather than the picture. Pre-existing shell behavior; the "still routes to the image presentation" claim holds everywhere except that one rail.

---

packages/base/file-formats/image-preview.gts
lukemelia: [Claude Code 🤖] Done in 722143e — `FileImage` (with `fileResourceURL`/`stringValue` and the `FileResourceLike` contract) now lives in its own lean `file-formats/file-image.gts`; `file-resources` imports and re-exports it so the public barrel surface is unchanged. card-api's universal graph gains only the img primitive, and the deps-guard lists now record `file-image` instead of `file-resources`.

packages/base/file-formats/image-preview.gts
lukemelia: [Claude Code 🤖] Done in 722143e — the thresholds and the svg rule are hoisted into `letterboxImage` in `file-presentation.gts`, and both `ImagePreview.fit` and the fitted shell's `thumbContain` call it, so the stage and the thumbnail rail can't drift apart.

packages/base/file-formats/file-shell-isolated.gts
lukemelia: [Claude Code 🤖] Done in 722143e — `hasExif` now probes every field the pane's templates render: all ten `CameraCaptureField` rows (via the computed/display projections, e.g. `focalLength.display`, `flash.label`) and all four `GeoLocationField` rows, so an image tagged only with aperture + focal length + flash keeps its pane. Thanks for the `hasColorProfile` framing — same shape now.

habdelra (approved):

PR #5748, opened 2026-08-11
Code (TypeScript & GTS) FileDef formats on the shared shells

lukemelia:
Migrates the TypeScript and GTS FileDef families onto the shared FileDef format shells (atom/embedded/fitted/isolated), so each family supplies only a `static previewComponent` while identity/facts/budgets/state live in the shared shell layer. Sibling of #5746 (Markdown & Plain Text); built on the shared shells already on `main`.

## Changes
- **`ts-file-def.gts`** — drop the bespoke atom/embedded/fitted/isolated templates; add a `CodePreview` renderer that syntax-highlights via the existing `highlightTs` (full source for embedded/isolated, the already-budgeted head snippet for fitted). Pin the code profile axes (`code` family/preview-kind, `text` adapter) so a `.ts`/`.gts` served without a content type still routes to the code renderer, and extract a `lineCount` the shells surface as the "N lines" hero fact.
- **`gts-file-def.gts`** — inherits everything; only relabels the kind to "Glimmer TS". `highlightTs` already marks up GTS template tags, so GTS reuses the same renderer.
- **`file-shell-isolated.gts`** — a full-width, auto-height code document stage for the `code` preview-kind, added as separate selectors so it doesn't collide with #5746's markdown/doc rules.
- **Tests** — realm-server indexing test moves to the shared `data-test-file-*` hooks and asserts the code preview mounts + highlights; a `lineCount` assertion on the ts extraction test; and a new integration test covering render across all four shells for TS and GTS.

## Verification
- `ember-tsc` typecheck, template-lint, eslint (test files): clean
- ts/gts acceptance (extraction / indexing / file-meta / `lineCount`): 6 tests, 33 assertions
- ts/gts preview render integration (mount into all four shells + TS/GTS highlighting): 7 tests, 20 assertions

## Notes
- Two design calls: independent PR off `main` (not stacked on #5746); GTS uses the `code` syntax-highlight preview rather than a schema summary (none exists today).
- The pre-commit hook reports an eslint "Parsing error" on `ts-file-def.gts` / `file-shell-isolated.gts`; this is a pre-existing `ember-eslint-parser`-on-base-`.gts` artifact already present on `main` (the original `ts-file-def.gts` fails identically) and is not run by CI (base's `lint` script runs no eslint). glint is the type-check of record and passes.

Linear: https://linear.app/cardstack/issue/CS-12235


packages/base/ts-file-def.gts
```diff
@@ -495,6 +291,9 @@ export class TsFileDef extends FileDef {
       title: fallbackTitle,
       excerpt: truncateExcerpt(source.replace(/\s+/g, ' ').trim()),
       content: source,
+      // A trailing newline shouldn't inflate the count, and empty content is
+      // zero lines rather than one.
+      lineCount: source ? source.replace(/\n$/, '').split('\n').length : 0,
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Good catch — fixed in fb82038. `lineCount` now normalizes `\r\n`/`\r` to `\n` before trimming the trailing newline, so a CR-only file no longer collapses to a single line and the count matches the fitted projection. Existing `lineCount` assertion (LF content = 5 lines) still passes.

packages/host/tests/integration/components/ts-file-def-preview-test.gts
```diff
@@ -0,0 +1,121 @@
...
+      content: GTS_SOURCE,
+    });
+    await renderCard(loader, file, 'isolated');
+    assert.dom('[data-test-file-isolated]').exists();
+    assert.dom('[data-test-ts-preview]').exists('GTS mounts the code preview');
+    // `highlightTs` marks `<template>` as a keyword, so a GTS component reads
+    // correctly without a second pass.
+    assert
+      .dom('[data-test-ts-preview] .ts-keyword')
+      .exists('GTS keywords and <template> are highlighted');
```
habdelra: [Claude Code 🤖] **Non-blocking — test-coverage.** This assertion is the only thing standing behind the test's name ("highlights its `<template>` tags"), but it can't actually catch a regression in that behavior. `GTS_SOURCE` contains `import`, `export`, `default`, `class`, and `extends`, all of which `highlightTs` marks as `.ts-keyword` independently of the `<template>` handling. So a `.ts-keyword` element exists whether or not the template tags are highlighted — delete `TEMPLATE_TAG_RE` from `ts-highlight.ts` and this test stays green.

**The behavior is real**, just not what this asserts: `TEMPLATE_TAG_RE = /<\/?template>/g` → `ts-keyword` (`ts-highlight.ts:15,74`). `<template>`/`</template>` are the one thing that distinguishes a `.gts` from a `.ts` here, so they're what's worth pinning. Since the tag renders as `<span class="ts-keyword">&lt;template&gt;</span>`, the highlighted span's `textContent` is `<template>` — assert on that specific token:

```js
assert.ok(
  [
    ...(this.element.querySelectorAll(
      '[data-test-ts-preview] .ts-keyword',
    ) as NodeListOf<HTMLElement>),
  ].some((el) => el.textContent === '<template>'),
  '<template> tag is marked as a keyword',
);
```

That fails if `TEMPLATE_TAG_RE` ever stops firing; the current `.exists()` check does not.

---

packages/base/ts-file-def.gts
```diff
@@ -495,6 +291,15 @@ export class TsFileDef extends FileDef {
...
       excerpt: truncateExcerpt(source.replace(/\s+/g, ' ').trim()),
       content: source,
+      // Normalize CRLF/CR to LF first so the count is the same across newline
+      // styles (and matches the fitted projection); a trailing newline
+      // shouldn't inflate the count, and empty content is zero lines.
+      lineCount: source
+        ? source
+            .replace(/\r\n?/g, '\n')
+            .replace(/\n$/, '')
+            .split('\n').length
```
habdelra, line 301: [Claude Code 🤖] **Non-blocking — test-coverage.** Correct as written, but none of the three behaviors packed into this expression is exercised by a test: CRLF→LF, lone-CR→LF, and the single trailing-newline strip. The newline-normalization was added specifically in response to the review thread on this line, yet it landed without a test that reaches it.

The one assertion that touches `lineCount` (`ts-file-def-test.gts`, `lineCount === 5`) uses a `utils.ts` fixture with no `\r` and no trailing newline, so it passes equally on the un-normalized `source.split('\n').length` path — its "without a trailing-newline inflation" message names a branch the fixture never reaches, and the CR case that motivated the fix is untested entirely.

A few direct cases would pin all three and guard against a future refactor quietly reverting to a bare `split('\n')`:

- `"a\r\nb\r\n"` → 2  (CRLF + trailing)
- `"a\rb"` → 2  (classic-Mac CR — the exact case the thread flagged)
- `"a\nb\n"` → 2  (LF + trailing)
- `""` → 0

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I reviewed this as a shell-adoption migration: whether TS/GTS land cleanly on the shared FileDef shell contract, whether the class-pinned profile axes are actually load-bearing, and whether the tests pin the behavior unique to the code family. I traced the projection (`fileViewModel`), the stage/shell wiring, and `highlightTs` against the checked-out head.

**Bottom line: no blocking issues — a clean, well-reasoned migration.** The two inline comments are non-blocking test-coverage notes; the code itself is correct.

- **The class-pinned axes are necessary, not decorative.** `.ts`/`.gts` are absent from `CONTENT_TYPE_BY_EXTENSION` in `file-type-profile.ts`, so a `.ts`/`.gts` served without (or with an uninformative) content type falls through `profileForFile` to `GENERIC` — no code preview, no "N lines" hero fact. Pinning `fileFamily`/`previewKind`/`previewAdapter`/`previewSource`/`fileKind` as statics makes `fileViewModel` (`file-view-model.ts:330-344`) read them via `fileProfileSource(model)` = the instance constructor, so identity survives a missing content type. Override precedence confirmed: `previewKind = profileSource?.previewKind ?? profile.previewKind`.
- **`previewKind='code'` deliberately overrides the registry.** The `text/typescript+glimmer` entry declares `previewKind:'schema', previewAdapter:'schema'`; because class statics win in `fileViewModel`, a real `GtsFileDef` renders as `code`, matching the PR's "no schema summary exists yet" call. This is the right lever — adding `.gts` to the extension map would instead have inherited the registry's `schema` and missed the intent.
- **GTS reuse is real.** `highlightTs` matches `</?template>` via `TEMPLATE_TAG_RE` → `ts-keyword` (`ts-highlight.ts:15,74`), so a `.gts`'s template tags highlight through the same pass.
- **No unbounded-growth regression.** The shared embedded shell bounds `.emb-body` to 220px (`data-embed-shape`, not in this diff), so `CodePreview`'s `overflow:auto` scrolls rather than expanding the card; the old bespoke embedded clipped at 200px, so this is a wash. The isolated `height:auto` code stage is the intended "read top-to-bottom" change, and `previewFrameStyle` is video-only (`file-shell-isolated.gts:315-322`), so it doesn't fight the new rule.
- **`lineCount` is coherent:** normalized the same way as the fitted snippet (CRLF/CR→LF), surfaced by `heroFactFor` for `previewKind ∈ {code,json,doc}` (`file-view-model.ts:285`), and no orphaned `data-test-ts-{isolated,embedded,fitted,atom}` references remain anywhere in the tree. The pattern also matches the existing `ImageFileDef` precedent (`static previewComponent = ImagePreview`).

**On the red realm-server shard (non-blocking, not this PR).** "Realm Server Tests (3, 6)" reports one failure (`+1` vs baseline `fb820380`). The changed `indexing-test.ts` assertions were present identically in that baseline commit and in head — they landed in the first shell-adoption commit, and head differs from the baseline only by the `lineCount` newline-normalization, which doesn't touch realm-server rendering. A deterministic break in those assertions would fail in both commits and show a `±0` delta, not `+1`, so the new failure is a different, non-deterministic test; meanwhile the host integration test exercising the same `GtsFileDef` isolated render is green. Recommend a re-run to confirm — if it reproduces, capture the failing test name.

**Recommendations (both non-blocking):**
1. The GTS `<template>` test asserts only that *some* `.ts-keyword` exists, which `import`/`class` already satisfy — it doesn't pin the one behavior unique to GTS. Detail in the inline comment on the integration test.
2. The `lineCount` CRLF/CR/trailing-newline normalization (the fix from the earlier review thread) has no test exercising those branches. Detail in the inline comment on `ts-file-def.gts`.

**Adjacent, out of scope.** The `text/typescript+glimmer` registry entry's `schema`/`schema` axes are now shadowed for every real `GtsFileDef` (the class pins `code`/`text`); they'd take effect only for a bare `FileDef` pointed at a `.gts`, which the indexer never produces. Worth pruning or wiring up when GTS schema summaries actually land — not this PR.

---

habdelra (approved):

packages/host/tests/integration/components/ts-file-def-preview-test.gts
lukemelia: [Claude Code 🤖] Fixed in 519d384. The test now asserts a `.ts-keyword` span whose `textContent` is exactly `<template>`, so it fails if `TEMPLATE_TAG_RE` ever stops firing — `import`/`class` no longer satisfy it. Thanks for the precise repro (delete `TEMPLATE_TAG_RE` → stays green); that was exactly the gap.

packages/base/ts-file-def.gts
lukemelia: [Claude Code 🤖] Fixed in 519d384. Added direct `lineCount` cases that call `extractAttributes` with in-memory bytes and reach each branch: `"a\r\nb\r\n"`→2 (CRLF+trailing), `"a\rb"`→2 (the classic-Mac CR the earlier thread flagged), `"a\nb\n"`→2 (LF+trailing), `""`→0. So the normalization is now covered rather than only the LF fixture that reached none of it.

PR #5751, opened 2026-08-12
Fix loader cycle race that fails host CI shards

lukemelia:
Closes [CS-12494](https://linear.app/cardstack/issue/CS-12494/loader-cycle-race-expected-to-be-registered-completing-deps-but-was)

## Problem

Host CI shards 5/6 fail — on main as well as on open PRs, surviving `rerun --failed` — with a QUnit global error between tests:

```
Global error: Uncaught Error: expected https://…/base/default-templates/isolated-and-edit
to be 'registered-completing-deps' but was 'registered'
```

The frequency jumped after the FileDef family PRs reshaped the base realm module graph, and a red main run silently blocks staging deploys, so this stopped being ignorable flake.

## Mechanism

`Loader.advanceToState` records a `completing-dep` for a module that is mid-completion on the *recording task's own recursion stack* (cycle handling — `isolated-and-edit` sits in the `card-api ↔ default-templates` import cycle). That leaves a window where the dep is still `'registered'`: the recording task must resume from whatever `await` suspended it (e.g. an unrelated module fetch) before the dep advances. A **concurrent import root** that reaches the recorded module during that window found the not-yet-advanced dep and treated it as a broken invariant — the throw above.

## Fix

The `registered-completing-deps → registered-with-deps` transition now advances such a dep itself (to `registered-completing-deps`, with the current stack) and re-enters the state machine, instead of asserting some other task already did it. Module state transitions are monotonic and re-entrant — the suspended root finds the work done when it resumes and returns early — so doing the pending work from whichever root gets there first is safe and guarantees forward progress.

## Testing

New `Unit | loader concurrent cycle completion` test pins the interleaving deterministically with a hand-controlled fetch: root 1 walks `a → b` (which cycles back to `a`, recording the completing-dep) then suspends awaiting a gated `slow` module; root 2 then imports `b` directly. Against the unpatched loader this reproduces the exact CI error; with the fix both roots evaluate.

Also green locally: `Unit | loader fetch-failure caching` (6), `Unit | Loader transient 5xx retry` (28), `Unit | loader | missing boxel icon` (3), host type-check, prettier/eslint. The `Unit | loader` and `loader prefetch` modules don't run on the local dev test page (they detach the harness frame — pre-existing local limitation); CI covers them.

### DIFF
--- a/packages/host/tests/unit/loader-concurrent-cycle-test.ts
+++ b/packages/host/tests/unit/loader-concurrent-cycle-test.ts
@@ -1,0 +1,62 @@
+import { module, test } from 'qunit';
+
+import { Deferred, Loader } from '@cardstack/runtime-common';
+
+// The loader records a `completing-dep` for a module that is mid-completion on
+// the recording task's own recursion stack — cycle handling. That leaves a
+// window where the dep is still 'registered': the recording task must resume
+// before the dep advances. A *concurrent* import root that reaches the
+// recorded module during that window used to treat the not-yet-advanced dep as
+// a broken invariant and throw
+// `expected <url> to be 'registered-completing-deps' but was 'registered'`;
+// it must instead complete the dep itself, since state transitions are
+// monotonic and re-entrant.
+module('Unit | loader concurrent cycle completion', function () {
+  test('a second import root completes a cycle participant that a suspended root left mid-completion', async function (assert) {
+    let releaseSlow = new Deferred<void>();
+    let slowRequested = new Deferred<void>();
+    let sources: Record<string, string> = {
+      '/a': `import './b'; import './slow'; export const a = 'a';`,
+      '/b': `import './a'; export const b = 'b';`,
+      '/slow': `export const slow = 'slow';`,
+    };
+    let fetchImpl: typeof globalThis.fetch = async (input) => {
+      let url =
+        typeof input === 'string'
+          ? input
+          : input instanceof URL
+            ? input.href
+            : input.url;
+      let path = new URL(url).pathname;
+      if (path === '/slow') {
+        slowRequested.fulfill();
+        await releaseSlow.promise;
+      }
+      let source = sources[path];
+      if (!source) {
+        return new Response('not found', { status: 404 });
+      }
+      return new Response(source, {
+        status: 200,
+        headers: { 'content-type': 'text/javascript' },
+      });
+    };
+    let loader = new Loader(fetchImpl);
+
+    // Root 1 walks a → b (which cycles back to a, so b records a
+    // completing-dep on a) → slow, and suspends awaiting slow's bytes with
+    // `a` still in state 'registered'.
+    let root1 = loader.import<{ a: string }>('http://race.example/a');
+    await slowRequested.promise;
+
+    // Root 2 enters at `b` while root 1 is suspended: b's completing-dep `a`
+    // has not been advanced yet, which is exactly the interleaving that used
+    // to throw.
+    let root2 = loader.import<{ b: string }>('http://race.example/b');
+    releaseSlow.fulfill();
+
+    let [modA, modB] = await Promise.all([root1, root2]);
+    assert.strictEqual(modA.a, 'a', 'the suspended root still evaluates');
+    assert.strictEqual(modB.b, 'b', 'the concurrent root evaluates the cycle');
+  });
+});
--- a/packages/host/tests/unit/loader-concurrent-cycle-test.ts
+++ b/packages/host/tests/unit/loader-concurrent-cycle-test.ts
@@ -6,11 +6,10 @@ import { Deferred, Loader } from '@cardstack/runtime-common';
 // the recording task's own recursion stack — cycle handling. That leaves a
 // window where the dep is still 'registered': the recording task must resume
 // before the dep advances. A *concurrent* import root that reaches the
-// recorded module during that window used to treat the not-yet-advanced dep as
-// a broken invariant and throw
-// `expected <url> to be 'registered-completing-deps' but was 'registered'`;
-// it must instead complete the dep itself, since state transitions are
-// monotonic and re-entrant.
+// recorded module during that window completes the dep itself rather than
+// treating the not-yet-advanced dep as a broken invariant, since state
+// transitions are monotonic and re-entrant. Regression guard against
+// `expected <url> to be 'registered-completing-deps' but was 'registered'`.
 module('Unit | loader concurrent cycle completion', function () {
   test('a second import root completes a cycle participant that a suspended root left mid-completion', async function (assert) {
     let releaseSlow = new Deferred<void>();
@@ -50,11 +49,13 @@ module('Unit | loader concurrent cycle completion', function () {
     await slowRequested.promise;
 
     // Root 2 enters at `b` while root 1 is suspended: b's completing-dep `a`
-    // has not been advanced yet, which is exactly the interleaving that used
-    // to throw.
+    // has not been advanced yet, which is exactly the interleaving under test.
     let root2 = loader.import<{ b: string }>('http://race.example/b');
     releaseSlow.fulfill();
 
+    // Each module's own export is the strongest stable assertion here: what a
+    // module sees *through* the cycle depends on which end the cycle is
+    // entered from, which is exactly what the second root varies.
     let [modA, modB] = await Promise.all([root1, root2]);
     assert.strictEqual(modA.a, 'a', 'the suspended root still evaluates');
     assert.strictEqual(modB.b, 'b', 'the concurrent root evaluates the cycle');
--- a/packages/host/tests/unit/loader-concurrent-cycle-test.ts
+++ b/packages/host/tests/unit/loader-concurrent-cycle-test.ts
@@ -14,6 +14,13 @@ module('Unit | loader concurrent cycle completion', function () {
   test('a second import root completes a cycle participant that a suspended root left mid-completion', async function (assert) {
     let releaseSlow = new Deferred<void>();
     let slowRequested = new Deferred<void>();
+    // Every axis of this graph is load-bearing: the two-node cycle, `./b`
+    // preceding `./slow` (so root 1 suspends with `a` still 'registered'),
+    // and root 2 entering at `b` specifically. A longer chain closes the
+    // window on its own — the suspended root resumes and advances the dep
+    // before the second root arrives — and the interleaving pins nothing.
+    // Anyone restructuring this test must re-verify it fails with the
+    // loader's completing-dep recovery reverted to a throw.
     let sources: Record<string, string> = {
       '/a': `import './b'; import './slow'; export const a = 'a';`,
       '/b': `import './a'; export const b = 'b';`,
--- a/packages/runtime-common/loader.ts
+++ b/packages/runtime-common/loader.ts
@@ -763,10 +763,30 @@ export class Loader {
             switch (depModule?.state) {
               case undefined:
               case 'fetching':
-              case 'registered':
-                throw new Error(
-                  `expected ${entry.moduleURL.href} to be 'registered-completing-deps' but was '${depModule?.state}'`,
+              case 'registered': {
+                // A `completing-dep` is recorded while the task completing it
+                // is still on its own recursion stack. A concurrent import
+                // root can reach this transition while that task is suspended
+                // at an await, so finding the dep not-yet-completed is a real
+                // interleaving, not a broken invariant. State transitions are
+                // monotonic and re-entrant, so do the pending work here and
+                // re-enter the state machine rather than asserting some other
+                // task already did it.
+                await this.advanceToState(
+                  entry.moduleURL,
+                  'registered-completing-deps',
+                  {
+                    ...stack,
+                    ...{
+                      'registered-completing-deps': [
+                        ...stack['registered-completing-deps'],
+                        resolvedURL.href,
+                      ],
+                    },
+                  },
                 );
+                break outer_switch;
+              }
               case 'registered-completing-deps': {
                 if (
                   !stack['registered-with-deps'].includes(entry.moduleURL.href)


packages/runtime-common/loader.ts
```diff
@@ -763,10 +763,30 @@ export class Loader {
...
+                    ...{
+                      'registered-completing-deps': [
+                        ...stack['registered-completing-deps'],
+                        resolvedURL.href,
+                      ],
+                    },
+                  },
                 );
+                break outer_switch;
+              }
```
habdelra, line 789: [Claude Code 🤖] **Confirmation — this is the right fix, and the reasoning in the comment is the load-bearing part.** Verified end-to-end rather than read; details below so the next editor knows what guards this.

**The cycle is real and it is a runtime cycle, not a type-only one.** `card-api.gts` imports `default-templates/isolated-and-edit`, which imports `default-templates/card-info`, which does `import { ImageDef } from '../card-api'` — a *value* import, so it survives type elision and the loader really does have to complete a cycle here. A reader checking `isolated-and-edit` alone would see only `import type { … } from '../card-api'` and wrongly conclude there's no cycle; the edge runs through `card-info`.

**Why doing another task's pending work is safe.** The claim rests on an invariant that isn't obvious from this hunk: `advanceToState` is *atomic per pass*. Both the `registered` and `registered-completing-deps` cases either (a) hit a dep needing work, `await`, and `break outer_switch` without writing, or (b) run the whole dep loop synchronously and then `setModule`. Because every pass re-reads `module` at the top of the `for (;;)`, a task that suspends can never write a dependency list computed from stale state. So two tasks concurrently advancing the same module cannot tear each other's writes — the later one recomputes from current state or returns early. That is exactly what makes "do the pending work here" sound rather than merely convenient, and this hunk preserves the idiom instead of breaking it.

**Verified the repro and the fix.** Driving the interleaving directly against this file with a gated fetch:

```
unpatched: RESULT: threw -> expected http://race.example/a to be
           'registered-completing-deps' but was 'registered'
patched:   RESULT: both roots evaluated { a: 'a', b: 'b' }
```

Same error string as the shard failure, so the repro pins the actual mechanism and not a lookalike.

**Verified termination**, which is the main risk of replacing a throw with a re-entrant loop — if `advanceToState(dep, 'registered-completing-deps')` could return with the dep still non-evaluatable, this becomes a live-lock (an import that never settles) instead of a throw. It can't: every terminal case of the state machine leaves the dep at `registered-completing-deps` or above (`broken` / `evaluated` / `preparing` all satisfy `isEvaluatable`), and an unobtainable dep propagates the fetch exception out rather than returning. Exercised both shapes with a 5s settle deadline:

```
dep-404-inside-cycle : settled -> rejected(unable to fetch http://race.example/gone: not found) | rejected(…)
cycle-participant-404: settled -> rejected(unable to fetch http://race.example/b: not found)    | rejected(…)
```

Both settle, and the surfaced error is now the underlying fetch failure rather than the opaque state assertion — a side benefit worth keeping in mind, since that assertion text was actively misleading about the cause.

**One note on the stack push**, so it doesn't read as a live guard to whoever edits next: appending `resolvedURL.href` here mirrors the sibling branch, but the module being appended is in state `registered-completing-deps`, which is already `isEvaluatable`. The on-stack check in the `registered` case is only consulted inside `if (!isEvaluatable(depModule))`, so while states stay monotonic this entry is never actually read. It is harmless and symmetric, and recursion depth stays bounded by the `registered` case's own guard, which does grow the stack. Not a change request — just worth knowing it's defensive symmetry rather than the thing preventing runaway recursion.

Non-blocking; nothing to change here.

packages/host/tests/unit/loader-concurrent-cycle-test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// The loader records a `completing-dep` for a module that is mid-completion on
+// the recording task's own recursion stack — cycle handling. That leaves a
+// window where the dep is still 'registered': the recording task must resume
+// before the dep advances. A *concurrent* import root that reaches the
+// recorded module during that window used to treat the not-yet-advanced dep as
+// a broken invariant and throw
+// `expected <url> to be 'registered-completing-deps' but was 'registered'`;
+// it must instead complete the dep itself, since state transitions are
+// monotonic and re-entrant.
```
habdelra: [Claude Code 🤖] **Evergreen: "used to" is temporal rot — this comment should state the contract, not the change.** Non-blocking but worth fixing before merge, since it's the comment a future reader leans on to understand what the test pins.

Line 9 (`…during that window used to treat the not-yet-advanced dep as a broken invariant and throw`) and lines 53-54 (`…which is exactly the interleaving that used to throw`) both anchor on a "before" a reader years from now has no reference point for. The comment reads as if something just changed, even long after it did. Everything else in the block is exactly right — it explains the mechanism, which is what makes this test maintainable.

Suggested rewrite of the header block, same technical content with the temporal anchor removed:

```suggestion
// The loader records a `completing-dep` for a module that is mid-completion on
// the recording task's own recursion stack — cycle handling. That leaves a
// window where the dep is still 'registered': the recording task must resume
// before the dep advances. A *concurrent* import root that reaches the
// recorded module during that window completes the dep itself rather than
// treating the not-yet-advanced dep as a broken invariant, since state
// transitions are monotonic and re-entrant. Regression guard against
// `expected <url> to be 'registered-completing-deps' but was 'registered'`.
```

For lines 52-54, `…which is exactly the interleaving under test` carries the same meaning without the "used to".

packages/host/tests/unit/loader-concurrent-cycle-test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    // Root 2 enters at `b` while root 1 is suspended: b's completing-dep `a`
+    // has not been advanced yet, which is exactly the interleaving that used
+    // to throw.
+    let root2 = loader.import<{ b: string }>('http://race.example/b');
+    releaseSlow.fulfill();
+
+    let [modA, modB] = await Promise.all([root1, root2]);
+    assert.strictEqual(modA.a, 'a', 'the suspended root still evaluates');
+    assert.strictEqual(modB.b, 'b', 'the concurrent root evaluates the cycle');
```
habdelra, line 68: [Claude Code 🤖] **Confirmation, plus a trap worth documenting here: these assertions are deliberately weak, and strengthening them the obvious way would make the test order-dependent and flaky.**

Asserting each module's *own* export looks like it under-tests the fix — the instinct is to also assert what each module sees *through* the cycle, e.g. have `/a` re-export a binding from `/b` and assert it resolved. I tried exactly that against this branch and it fails:

```
single root         : bSeenByA = "b"
gated, single root  : bSeenByA = "b"
gated + second root : bSeenByA = undefined
```

That `undefined` is **not** a defect in this fix. It's ordinary cycle-entry-point semantics: whichever module the cycle is entered at runs its factory while the other is still `preparing`, so the earlier one reads a not-yet-populated export. Control case, no race and no gating at all — just importing `/b` first:

```
cycle entered at b, no race: bSeenByA = undefined
```

Same result. The AMD-transpiled analogue of what real ESM expresses as a TDZ error on a `const` in a cycle. Since the second root here is precisely what changes which end the cycle gets entered from, any cross-cycle binding assertion would encode one arbitrary entry order and break the moment scheduling shifted — a new flake in a test written to kill one.

So "both roots evaluate without throwing" is the strongest assertion that is actually stable for this interleaving, and the choice is right. Suggest capturing that in a line of comment so the next person doesn't "improve" the test into a flake:

```js
// Each module's own export is the strongest stable assertion here: what a
// module sees *through* the cycle depends on which end the cycle is entered
// from, which is exactly what the second root varies.
```

Non-blocking.

packages/runtime-common/loader.ts
habdelra, line 1: [Claude Code 🤖] **Adjacent, pre-existing, explicitly not this PR's job — but it's the nearest sibling of the bug being fixed, and it fails silently instead of loudly.** Filing it here because whoever next opens this state machine should know.

**The hole.** In the `registered` case of `advanceToState`, a dep is classified by three branches. When `!isEvaluatable(depModule)` and the dep *is* on `stack['registered-completing-deps']`, the code pushes a `completing-dep` only `else if (isRegistered(depModule))`. If the dep's map entry is missing or still `fetching`, neither branch fires and **nothing is pushed for that entry at all** — it silently vanishes from `maybeReadyDeps`.

**Why that is worse than a throw.** `evaluate` builds the factory arguments positionally — `module.dependencies.map(...)` then `module.implementation(...dependencies)`. A dropped entry shifts every argument after it, so the module's factory binds the *wrong modules* with no error at the point of the mistake. The failure surfaces later and somewhere else.

**Reachability — verified, not inferred.** It needs the modules map to be discarded mid-traversal, which the loader does on realm-mapping changes: `addRealmMapping` / `removeRealmMapping` call `notifyMappingChange`, and this file's `onMappingChange` subscription does `this.modules.clear()`. `removeRealmMapping`'s own doc comment notes it is used by tests scoping a temporary prefix per test — i.e. this fires routinely in the host suite, concurrently with in-flight imports.

My first attempt at a repro (two-module cycle, mapping change during the suspension) recovered cleanly with correct bindings, because the suspended frame re-reads and re-fetches on resume. The window is narrower than that: it needs a dep that is on the recursion stack *and* whose entry is gone. A three-module cycle `a → b → c → a` where `c` also pulls a gated module gets there — `c` suspends with both `a` and `b` on the stack, the mapping change lands, and when `c` resumes it re-classifies `a` as on-stack-but-absent and drops it:

```
rejected -> Cannot read properties of undefined (reading 'marker')
```

`c`'s factory received `[exports, slow, marker]` where it expected `[exports, a, slow, marker]`, so the last parameter bound `undefined`.

**Not introduced or widened here.** I ran that same probe against this branch and against its parent commit and got identical output, so this predates the change. One connection worth noting: the new branch adds another site that appends to `stack['registered-completing-deps']`, which is the precondition this hole keys on — so if it's ever fixed, the fix wants to cover both sites rather than just the older one.

**The way out**, whenever someone takes it: treat "on the stack but no entry" as work to do rather than as a no-op — recurse the way the sibling branch now does — or, at minimum, make the drop impossible to do silently by pushing the entry unconditionally and letting `evaluate`'s existing `isEvaluatable` check raise. The one thing to avoid is leaving a path where the dependency list can come out shorter than `dependencyList`, since nothing downstream can detect that.

Follow-up, not blocking.

habdelra (commented):
[Claude Code 🤖] Reviewed this as a concurrency fix rather than a flake patch: the questions that matter are whether the described interleaving is the one actually failing, whether replacing an assertion with a re-entrant loop can spin instead of settling, and whether the new test pins the mechanism or just its symptom. Verified each by driving the loader directly with a gated fetch, against this branch and its parent.

**Bottom line: the mechanism is correct, the fix is the right shape, and I found no blocking issues.** The diagnosis holds up against the real base-realm graph, the repro produces the exact shard error on the parent commit and passes here, and the new branch settles rather than live-locking when a dep can never become evaluatable. Two non-blocking comments and one adjacent pre-existing bug.

**What lands right.** The fix preserves an invariant that is easy to break in this file and never stated in it: `advanceToState` is atomic per pass — every case either awaits and `break outer_switch`es without writing, or completes its dep loop synchronously and then writes, and every pass re-reads state at the top of the `for (;;)`. That is what makes "whichever root gets there first does the work" safe rather than a source of torn dependency lists, and this hunk follows the idiom instead of bolting a lock or a retry onto it. Worth defending if anyone proposes serializing the traversal instead.

The diagnosis also survives the check I most expected it to fail. `isolated-and-edit` imports `card-api` with `import type`, which is elided — so on its own it isn't in a runtime cycle at all. The real edge runs through `card-info`, which does `import { ImageDef } from '../card-api'` as a value import. The cycle is genuine.

Recommendations, in priority order:

1. Drop the "used to" phrasing from the test file's header comment and from the comment above root 2 — it's the one thing here that will read as stale to a future reader. Suggested rewrite in the thread on that block.
2. Consider a line of comment recording *why* the assertions are deliberately weak. I tried the obvious strengthening — asserting what each module sees through the cycle — and it fails on this branch for a reason unrelated to the fix: cycle-entry order, reproducible with no race at all. Encoding one entry order would turn this test into a new flake. Evidence in the thread on the assertions.
3. Nothing to do on the loader hunk; the thread there records what guards the fix so the next editor doesn't have to re-derive it.

**Adjacent, out of scope.** The sibling branch of the one being fixed can silently *drop* a dependency rather than throw about it — when a dep is on the recursion stack but its map entry has been discarded, neither classification branch fires and the entry disappears from the list. Because factory arguments are positional, that shifts every later argument and binds the wrong modules with no error at the point of the mistake. It's reachable through the mapping-change path that clears the modules map, I reproduced it with a three-module cycle, and it behaves identically on this branch and its parent — so it predates this work and isn't asked of this PR. Detail in the file-level comment on the loader, including the one way this change touches it: the new code adds a second site that populates the stack the hole keys on, so a future fix wants to cover both.

One caveat on the evidence for the flake itself: all 20 host shards are green here, which is supporting but not conclusive on its own for an intermittent failure. The deterministic repro is the stronger signal, and it is unambiguous.

packages/runtime-common/loader.ts
lukemelia: [Claude Code 🤖] Thanks for the end-to-end verification — especially pinning the runtime cycle edge through `card-info`'s value import and the termination check on the 404 shapes; the per-pass atomicity framing is exactly the invariant the fix leans on. Agreed the stack push in the new arm is defensive symmetry rather than a live guard.

packages/host/tests/unit/loader-concurrent-cycle-test.ts
lukemelia: [Claude Code 🤖] Fixed in 262dca3 — the header now states the contract with the regression-guard framing you suggested, and the inline comment reads "the interleaving under test".

packages/host/tests/unit/loader-concurrent-cycle-test.ts
lukemelia: [Claude Code 🤖] Added in 262dca3 — your suggested comment now sits on the assertions, so the cycle-entry-order trap is recorded where the next person would otherwise "strengthen" the test into a flake. The `bSeenByA = undefined` control case was a genuinely useful probe.

packages/runtime-common/loader.ts
lukemelia: [Claude Code 🤖] Filed as [CS-12496](https://linear.app/cardstack/issue/CS-12496/loader-silently-drops-a-dep-entry-when-a-cycle-participants-map-entry) with your repro shape, the mis-binding consequence, and the note that this PR adds a second `stack['registered-completing-deps']` append site a fix should cover. The "dependencies must never come out shorter than dependencyList" invariant is captured as the acceptance bar.

packages/host/tests/unit/loader-concurrent-cycle-test.ts
```diff
@@ -0,0 +1,63 @@
...
+// before the dep advances. A *concurrent* import root that reaches the
+// recorded module during that window completes the dep itself rather than
+// treating the not-yet-advanced dep as a broken invariant, since state
+// transitions are monotonic and re-entrant. Regression guard against
+// `expected <url> to be 'registered-completing-deps' but was 'registered'`.
+module('Unit | loader concurrent cycle completion', function () {
+  test('a second import root completes a cycle participant that a suspended root left mid-completion', async function (assert) {
+    let releaseSlow = new Deferred<void>();
+    let slowRequested = new Deferred<void>();
+    let sources: Record<string, string> = {
```
habdelra, line 24: [Claude Code 🤖] **The shape of this graph is load-bearing, and that isn't obvious from reading it — worth a guard note so it isn't "simplified" into a test that cannot fail.**

I tried to generalize this to the three-node shape the base realm actually has — `card-api -> isolated-and-edit -> card-info -> card-api` — as `/a -> /b -> /c -> /a` with `/a` also pulling the gated module, and the second root entering at `/c`, the module that closes the cycle. That variant passes against the **unpatched** loader, so it reproduces nothing and says nothing either way about this fix. I'm reporting it only as a caution, not as a finding.

The reason is timing rather than topology: `loader.import` yields before the walk reaches the `registered-with-deps` transition, and with a longer chain the suspended root resumes and advances the dep past `registered` before the second root gets there, so the window closes on its own. The two-node construction plus entering the second root at `/b` specifically is what holds the window open.

So the graph, the dep ordering in `/a` (`'./b'` before `'./slow'`, so the walk is suspended with `a` still `registered`), and which module root 2 enters at are all part of what makes this observable. Anyone restructuring this test should re-verify it still fails with the `loader.ts` change reverted — otherwise it silently stops being a regression guard while still passing.

habdelra (commented):
[Claude Code 🤖] **Verification report rather than a second opinion on the mechanism** — the fix, its termination argument, the evergreen comment fix, and the adjacent silent dependency-drop hole are already covered in this thread, so I deliberately did not re-cover them. I found no new defects. What I could add is empirical coverage of the two things the description defers.

**1. The new test under the harness that reports the failure at all.** The description notes the `Unit | loader` modules don't run on the local dev test page and leaves them to CI. The dev server has no testem client, so it never reports a global error — `ember test --path dist` is the only local runner that does, which makes it the one worth confirming this under. Green there:

```
ok 1 Chrome 151.0 - [27 ms] - Unit | loader concurrent cycle completion:
     a second import root completes a cycle participant that a suspended root left mid-completion
# tests 48  # pass 48  # fail 0     (whole `Unit | loader` filter, new module included)
```

**2. The actual module from the failing shard.** `Acceptance | code-submode | card playground` is **40/40 with no global error emitted** on this branch under the same runner. That is the module worth running because it is the *entire* content of the shard-5 failure: the junit artifact for `Host Tests (5, 20)` on that run contains 34 test cases, all of them card-playground, and the global error as a trailing 35th entry after the last one. That corroborates the description's "between tests" framing from the artifact side — the named test is simply whoever was active when the rejection landed, and nothing in that shard mutates realm mappings, so the churn came from the module graph itself.

Both runs were against `71c2b0689d`; `262dca300a` changes only test comments, so `loader.ts` is byte-identical to what I exercised.

One inline note on the test's construction, which is a caution for future editors rather than a change request.

habdelra (commented):
[Claude Code 🤖] **Re-review of `262dca3`, plus triage of the two red shards — those are not related to this change.**

## The red shards

`Host Tests (9, 20)` and `Host Tests (11, 20)`. Four independent lines all point away from this PR:

**Blast radius.** The `loader.ts` change is a single hunk that replaces an unconditional `throw` with work, in the `case undefined / 'fetching' / 'registered'` arm of the `registered-with-deps` transition. An execution that never reaches that arm is byte-identical to before; an execution that did reach it previously failed the run outright with the state error. There is no path by which this converts a passing test into a failing one — only a throwing one into a progressing one.

**What actually failed.** Neither failure is in module loading:

- Shard 9 — `Browser failed to connect within 240s. testem.js not loaded?`. Its junit holds a single synthetic `error` case and no test results at all, so the harness never came up and nothing ran.
- Shard 11 — 160 cases, 1 failure: `Expected an active AI assistant room before adding skill "…/Skill/card-editing"`, thrown from mock-matrix's `addSkillToAiAssistant` at `tools-test`. That is a room-setup precondition in the AI-assistant helper.

**Cross-commit control, which is the decisive one.** Both shards were **green on `71c2b068`** and red on `262dca3`. The only difference between those two commits is comment text in a test file, so neither failure can be a behavioral consequence of anything here.

**Local check.** The shard-11 test passes against this branch's build under `ember test --path dist` (13.4s):

```
ok 1 Chrome 151.0 - [13390 ms] - Acceptance | Tools tests: ShowCard command added from
     a skill, is not automatically executed when agentId does not match
```

For completeness, a branch of my own off the same base passed all 20 shards, so the base isn't carrying these. Read as two independent flakes — a Chrome/harness startup failure and an AI-assistant room-readiness race — and a rerun of just those two shards should come back green.

## The changes

The temporal anchors are gone from both sites, and the file greps clean for the usual markers (`used to`, `previously`, `no longer`, `originally`, …). The header now states the contract and the inline note reads as the interleaving under test.

Worth recording, since it is easy to mistake for an oversight later: after this change the string `…to be 'registered-completing-deps' but was 'registered'` has **no throw site left in `loader.ts`**, so the header comment is the only place it appears in the tree. That is the right home for it — the CI history someone will grep years from now is full of that string, and this is the test that pins it.

One placement nit, non-blocking: the cycle-entry-order note sits on the `Promise.all` line rather than the assertions it describes. Two lines lower puts it directly where someone about to "strengthen" an assertion would read it.

packages/host/tests/unit/loader-concurrent-cycle-test.ts
lukemelia: [Claude Code 🤖] Captured in 026d755ad2 — the sources block now documents that the two-node cycle, the `./b`-before-`./slow` ordering, and root 2's entry at `b` are each load-bearing, that longer chains close the window on their own, and that any restructuring must be re-verified against the reverted loader. Your negative result on the three-node shape is exactly the trap that note guards.

lukemelia:
[Claude Code 🤖] Thanks for both verification passes. The 01:31 report closes the two gaps the description deferred — the new module green under `ember test --path dist` (the one local runner that surfaces global errors) and `card playground` 40/40 with no trailing global error on this branch — and the shard triage matches what I found from the logs (shard 9 harness-never-connected, shard 11 an AI-assistant room precondition), with the cross-commit control settling it. CI is re-running in full at the graph-shape-guard commit; no open review items remain on my side.

PR #5752, opened 2026-08-12
Polish profile popover, pricing modal spacing, and dropdown hover styling

cmgardella:
## Summary
- Restyled the profile popover: lighter divider color, teal "Upgrade Plan" CTA matching "Sign Out"'s style/font, custom hover colors for the black and teal buttons, removed the credit icon next to "Not available on Free plan", flat `#EFEDED` daily-grant note with no border, and precise 20px spacing around the "Buy more credits" CTA
- Added 50px of space above the pricing modal's headline while keeping scroll-snap stable via a matching `scroll-padding-top`
- Added a 100%-opacity white stroke on hover for the search, AI assistant, and profile avatar icon buttons, matching the rest of the app's icon-button hover treatment
- Made the workspace chooser's "View All" sort dropdown and the hosted-site-url dropdown's hover state span edge-to-edge like the standard boxel dropdown/menu pattern, instead of being inset from the platter

Stacked on #5728 so this diff only shows the new changes.

## Test plan
- [ ] Open the pricing modal (avatar menu → Upgrade Plan) and confirm the headline has breathing room at top and scroll-snap still works
- [ ] Open the profile popover and confirm Upgrade Plan/Sign Out/Buy more credits styling and spacing
- [ ] Hover the search, AI assistant, and avatar icon buttons and confirm the white stroke goes to full opacity
- [ ] Open the workspace chooser "View All" dropdown and a workspace's hosted-site-url dropdown and confirm hover fills edge-to-edge

### DIFF
--- a/packages/host/app/components/ai-assistant/past-session-item.gts
+++ b/packages/host/app/components/ai-assistant/past-session-item.gts
@@ -9,6 +9,7 @@ import { tracked } from '@glimmer/tracking';
 import ExternalLink from '@cardstack/boxel-icons/external-link';
 
 import { format as formatDate, isSameDay, isSameYear } from 'date-fns';
+import { modifier } from 'ember-modifier';
 
 import {
   BoxelDropdown,
@@ -45,6 +46,42 @@ interface Signature {
 
 export default class PastSessionItem extends Component<Signature> {
   @tracked private preventMenuClose = false;
+  private closeDropdownAction: (() => void) | null = null;
+  private closeDropdownTimer: ReturnType<typeof setTimeout> | null = null;
+  private triggerElement: HTMLElement | null = null;
+
+  private registerCloseAction = modifier(
+    (_element: HTMLElement, [closeFn]: [() => void]) => {
+      this.closeDropdownAction = closeFn;
+      return () => {
+        this.closeDropdownAction = null;
+      };
+    },
+  );
+
+  private registerTriggerElement = modifier((element: HTMLElement) => {
+    this.triggerElement = element;
+    return () => {
+      this.triggerElement = null;
+    };
+  });
+
+  @action
+  private scheduleCloseDropdown() {
+    this.cancelCloseDropdown();
+    this.closeDropdownTimer = setTimeout(() => {
+      this.closeDropdownAction?.();
+      this.triggerElement?.blur();
+    }, 200);
+  }
+
+  @action
+  private cancelCloseDropdown() {
+    if (this.closeDropdownTimer) {
+      clearTimeout(this.closeDropdownTimer);
+      this.closeDropdownTimer = null;
+    }
+  }
 
   <template>
     <li
@@ -52,6 +89,8 @@ export default class PastSessionItem extends Component<Signature> {
       data-test-joined-room={{@session.roomId}}
       data-room-id={{@session.roomId}}
       data-is-current-room={{@isCurrentRoom}}
+      {{on 'mouseenter' this.cancelCloseDropdown}}
+      {{on 'mouseleave' this.scheduleCloseDropdown}}
     >
       <button
         class='view-session-button'
@@ -87,7 +126,7 @@ export default class PastSessionItem extends Component<Signature> {
           {{/if}}
         </div>
       </button>
-      <BoxelDropdown>
+      <BoxelDropdown @contentClass='past-session-menu-dropdown'>
         <:trigger as |bindings|>
           <Tooltip @placement='top'>
             <:trigger>
@@ -98,6 +137,7 @@ export default class PastSessionItem extends Component<Signature> {
                 @label='past session options'
                 data-test-past-session-options-button={{@session.roomId}}
                 {{bindings}}
+                {{this.registerTriggerElement}}
               />
             </:trigger>
             <:content>
@@ -108,6 +148,9 @@ export default class PastSessionItem extends Component<Signature> {
         <:content as |dd|>
           <Menu
             class='menu past-session-menu'
+            {{this.registerCloseAction dd.close}}
+            {{on 'mouseenter' this.cancelCloseDropdown}}
+            {{on 'mouseleave' this.scheduleCloseDropdown}}
             @closeMenu={{fn this.handleCloseMenu dd.close}}
             @items={{array
               (menuItem
@@ -139,37 +182,49 @@ export default class PastSessionItem extends Component<Signature> {
       }
 
       .session {
+        position: relative;
         display: flex;
         align-items: center;
         justify-content: space-between;
-        border-top: 1px solid var(--past-sessions-divider-color);
         padding: var(--boxel-sp) var(--boxel-sp-sm);
         margin-right: var(--boxel-sp-xs);
         margin-left: var(--boxel-sp-xs);
         border-radius: var(--boxel-border-radius-xs);
       }
 
-      .session:first-child {
-        border-top: none;
+      .session::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        left: 0;
+        right: 0;
+        height: 1px;
+        background: #4f4b57;
       }
 
-      .session:hover {
+      .session:first-child::before {
+        display: none;
+      }
+
+      .session:hover,
+      .session:has(.menu-button[aria-expanded='true']) {
         background-color: var(--ai-assistant-menu-hover-background);
+        border-radius: calc(var(--boxel-border-radius-xs) + 2px);
         cursor: pointer;
       }
-      .session[data-is-current-room] {
-        border: 1px solid var(--past-sessions-divider-color);
-      }
-      .session:hover + .session:not([data-is-current-room]),
-      .session[data-is-current-room] + .session {
-        border-top-color: transparent;
+      .session:hover::before,
+      .session:has(.menu-button[aria-expanded='true'])::before,
+      .session:hover + .session::before,
+      .session:has(.menu-button[aria-expanded='true']) + .session::before {
+        background: transparent;
       }
       .name {
         font-weight: 600;
       }
       .date {
-        margin-top: var(--boxel-sp-xxs);
+        margin-top: calc(var(--boxel-sp-xxs) - 5px);
         color: var(--boxel-400);
+        font-size: 12px;
       }
       .view-session-button {
         color: var(--boxel-light);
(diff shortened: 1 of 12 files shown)


packages/host/app/components/ai-assistant/past-session-item.gts
```diff
@@ -195,12 +250,17 @@ export default class PastSessionItem extends Component<Signature> {
...
+      }
+
       .menu {
         --boxel-menu-item-content-padding: var(--boxel-sp-xxs)
           var(--boxel-sp-sm);
 
-        background: var(--ai-assistant-menu-background);
-        border: 1px solid var(--past-sessions-divider-color);
+        background: #3b394b;
+        border: 1px solid rgba(255, 255, 255, 0.25);
```
richardhjtan, line 263: [Claude Code 🤖] These are named tokens, and swapping a token out for its literal costs the theming that token exists to provide. `packages/boxel-ui/src/styles/variables.css` already defines every value the diff hardcodes:

| literal in the diff | token | note |
| --- | --- | --- |
| `#3b394b` | `--boxel-650` | also in `past-sessions.gts`, `workspace-chooser/index.gts` |
| `#272330` | `--boxel-700` | replaces `var(--ai-assistant-menu-hover-background)` here |
| `#4f4b57` | `--boxel-575` | replaces `var(--past-sessions-divider-color)` |
| `#afafb7` | `--boxel-400` | `profile-info-popover.gts`, replaces `var(--boxel-dark)` |
| `#00ffba` | `--boxel-teal` / `--boxel-highlight` | `workspace-chooser/index.gts` checkmark |
| `rgba(255, 255, 255, 1)` | — | this is `#fff`; `submode-layout.gts` and `ai-assistant/button.gts` too |

Four more have no token and are a hairline off one that exists, which is the tell that they were sampled from a mockup rather than looked up: `#00d99e` against `--boxel-dark-teal: #00da9f` (which is what `--boxel-highlight-hover` resolves to), `#efeded` against `--boxel-50: #f5f5f5`, plus `#00c38e`, `#404040`, `#595959`.

`.claude/skills/gts-component-conventions/SKILL.md` §3 asks for two things here, and the second matters more than the first: no raw literal in a component, and reach for the **semantic role** token rather than the numbered palette, because `--boxel-400`…`--boxel-700` are primitives that do not flip between light and dark while `--muted`, `--muted-foreground`, `--foreground` and `--boxel-highlight-hover` do. So `--boxel-highlight-hover` for the teal hover, not a new `#00d99e`; and where a hover surface recurs across these four components, one component-scoped custom property rather than the same literal in four files.

Blocking, and a regression — the tokens being replaced were correct before this diff.

packages/host/app/components/operator-mode/workspace-chooser/index.gts
```diff
@@ -567,14 +568,46 @@ export default class WorkspaceChooser extends Component<Signature> {
...
+      :global(
+        .workspace-chooser-sort-dropdown
+          .boxel-select-option-item.ember-power-select-option--selected.ember-power-select-option--highlighted
+      ) {
+        background-color: #3b394b;
+      }
+      :global(
+        .workspace-chooser-sort-dropdown
+          .boxel-select-option-item.ember-power-select-option--selected:hover
+      ) {
```
richardhjtan, line 604: [Claude Code 🤖] `ember-power-select-option--selected` and `--highlighted` are ember-power-select's own class names, and nothing in this repo pins them — a version bump that renames them takes the selected-row hover with it, silently and only in this one dropdown.

`@dropdownClass` plus the `--boxel-dropdown-*` custom properties above is the supported seam, and it is already doing most of the work here. What is left reaching past it is only the selected-and-highlighted state. Ask: can `BoxelSelect` expose that state as an override knob (a `--boxel-dropdown-selected-hover-color`, say) so the rule lives next to the other five and the third-party name stays inside the component that owns the dependency? If it can't, a comment naming ember-power-select as the source of those two classes at least tells the next reader why the selector looks like this.

Non-blocking; a decision about where the seam goes.

packages/host/app/components/ai-assistant/past-session-item.gts
```diff
@@ -45,13 +46,51 @@ interface Signature {
...
+  });
+
+  @action
+  private scheduleCloseDropdown() {
+    this.cancelCloseDropdown();
+    this.closeDropdownTimer = setTimeout(() => {
+      this.closeDropdownAction?.();
+      this.triggerElement?.blur();
+    }, 200);
+  }
```
richardhjtan, line 76: [Claude Code 🤖] Two things on this path.

**The blur is unconditional.** `mouseleave` on the row schedules a close that also calls `blur()` on the trigger, so a keyboard user who opened the menu with Enter and then moves the mouse — anywhere, for any reason — loses focus to `<body>` and has to tab back in. Gate the blur on the interaction that needs it (the pointer path) or drop it; the close itself is fine.

**The timer outlives the component.** Nothing clears `closeDropdownTimer` on destroy. The callback is harmless once the modifiers have nulled `closeDropdownAction` and `triggerElement`, so this is not a user-visible bug — but a pending timer per row past teardown is exactly the shape the host suite's memory work chases, and the fix is three lines in the constructor:

```ts
registerDestructor(this, () => this.cancelCloseDropdown());
```

Non-blocking. The blur is the half worth a decision.

packages/host/app/components/ai-assistant/past-session-item.gts
```diff
@@ -139,37 +182,49 @@ export default class PastSessionItem extends Component<Signature> {
...
+      .session:hover + .session::before,
+      .session:has(.menu-button[aria-expanded='true']) + .session::before {
+        background: transparent;
       }
       .name {
         font-weight: 600;
       }
       .date {
-        margin-top: var(--boxel-sp-xxs);
+        margin-top: calc(var(--boxel-sp-xxs) - 5px);
```
richardhjtan, line 225: [Claude Code 🤖] `calc(var(--boxel-sp-xxs) - 5px)` subtracts a magic offset from a token, so the result moves whenever the token is retuned — in a direction nobody chose. If the intended margin is a fixed value, write that value; if it's a smaller step on the spacing scale, use that token.

Same ask for `font-size: 12px` on the next line and the other raw pixels the diff adds — `left/right: 30px` in `past-sessions.gts`, `gap: 5px` in `matrix/room.gts` and `pill-menu/index.gts`, `margin-bottom: 4px` in `profile-info-popover.gts`. §2 of the `.gts` conventions asks for `rem` or an existing `--boxel-sp-*` / `--boxel-font-size-*` token; the root is 16px, so 12px is `0.75rem` and there is a `--boxel-font-size-xs`.

Non-blocking, except that the `calc()` is the one that will actually surprise someone.

packages/host/app/components/operator-mode/choose-subscription-plan-modal.gts
```diff
@@ -41,11 +41,39 @@ export default class ChooseSubscriptionPlanModal extends Component<Signature> {
 
   <template>
     <style scoped>
+      :global(.choose-subscription-plan-modal .dialog-box__content) {
+        scroll-snap-type: y mandatory;
+        scroll-padding-top: 50px;
```
richardhjtan, line 46: [Claude Code 🤖] This `50px` and the `padding-top: 50px` on `.boxel-pricing-container` have to stay equal or scroll-snap lands the headline under the absolutely-positioned header — but nothing here says so, and they are in different rules 20 lines apart. One custom property used by both makes the constraint the code rather than a coincidence:

```css
.boxel-pricing-container { --pricing-header-offset: 3.125rem; padding-top: var(--pricing-header-offset); }
:global(.choose-subscription-plan-modal .dialog-box__content) { scroll-padding-top: var(--pricing-header-offset); }
```

Also in this block: `@title=''` on `ModalContainer` empties the header rather than removing it, which is why the header then needs `position: absolute` + `pointer-events: none` and the close button needs `pointer-events: auto` handed back. If `ModalContainer` can take no title at all, that whole four-rule workaround goes away. Worth checking before this lands, since an empty-string title is the kind of thing a later change to `ModalContainer` breaks without knowing anyone depended on it.

Non-blocking.

richardhjtan (commented):
[Claude Code 🤖] Went after this as a styling change: the colors and units against the repo's `.gts` CSS conventions, the `:global()` reach into other packages' internals, the new hover-close behavior's keyboard and teardown paths — plus where this PR can actually land.

**Blocking, and mechanical: the base branch has already merged.** This targets `ai-assistant-panel-design-polish`, and the PR that owned that branch merged two weeks ago. Retarget to `main` first — `main` has since taken 13, 12, 10 and 7 commits on `workspace-chooser/index.gts`, `matrix/room.gts`, `workspace-chooser/workspace.gts` and `submode-layout.gts`, and a merge into `main` conflicts in four files: `ai-assistant/past-session-item.gts`, `matrix/room.gts`, `workspace-chooser/index.gts`, `pill-menu/index.gts`. Everything below is worth doing on the rebased diff, not this one.

Recommendations:

1. Every hardcoded color in the diff is either an existing named token or a hairline variant of one — and four of them replace a token that was already there. See the thread on `.menu` in `past-session-item.gts`; it carries the full mapping and applies to `past-sessions.gts`, `workspace-chooser/index.gts`, `profile-info-popover.gts` and `submode-layout.gts` too.
2. The sort dropdown is styled through third-party class names nothing in this repo pins. See the thread on `workspace-chooser/index.gts`.
3. The new hover-close blurs the trigger and leaves a timer running past teardown. See the thread on `scheduleCloseDropdown`.
4. Raw `px` where the conventions ask for `rem` or a token, including one `calc()` that mixes a token with a magic offset. See the threads on `past-session-item.gts` and the pricing modal.

**The red check is stale, not a signal.** `Host Memory Baseline` failed on a run started three weeks ago, before the current head. It needs a re-run on the rebased branch before it means anything.

**Adjacent, out of scope.** `llm-select.gts`'s new `:deep(.menu-content::after) { display: none }` describes, in its own comment, a `PillMenu` bug: `LLMSelect` builds a footer row inside `.llm-list` instead of passing a `:footer` block, so `PillMenu` never sets `has-footer` and leaves its bottom scroll-shadow floating over the list. Suppressing the shadow from the consumer is the right size of fix for a polish PR; the footer block belongs in `PillMenu`'s own API and not here.

PR #5755, opened 2026-08-12
Show per-request and per-session AI token usage behind showTokens

jurgenwerk:
This helps us see how much AI assistant responses are costing us. 

It was of great help when I was debugging caching problems after landing https://github.com/cardstack/boxel/pull/5676.  

Turned on by default in development. In a deployed environment, can use `showTokens=true` query param to show it.  

image


packages/host/app/lib/matrix-classes/message-builder.ts
```diff
@@ -227,6 +228,13 @@ export default class MessageBuilder {
     message.continuationOf = isCardMessageEvent(this.event)
       ? (this.event.content[APP_BOXEL_CONTINUATION_OF_CONTENT_KEY] ?? null)
       : null;
+    // The token counts arrive on a late edit — the one that completes the
+    // answer, or a follow-up edit when the counts trailed it. Earlier edits
+    // don't carry them, so only ever set, never clear.
+    let usage = (this.event.content as CardMessageContent)?.data?.usage;
+    if (usage) {
+      message.setUsage(usage);
+    }
```
habdelra, line 237: [Claude Code 🤖]

**The late-usage edit this block depends on is dropped by the host in a live session — so in exactly the timing that `finalize()`'s follow-up edit exists to handle, the counts silently never appear until a reload.** (Regression introduced by this change; blocking for the live late-update path, though see the severity note below.)

**The mechanism.** `setUsage` here only runs if `loadRoomMessage` actually reaches `updateMessage`. But `loadRoomMessage` gates every build/update behind `if (!message?.isStreamingOfEventFinished)` (`resources/room.ts` ~L765), and `isStreamingOfEventFinished` is just `_isStreamingFinished === true` (`message.ts` L201-203). There are two ways the usage reaches the room, and only one survives that guard:

- **Usage rides the finishing edit** (the common timing: the usage chunk arrives before the throttled final edit, so `finalize()`'s flush sends a single edit carrying both `isStreamingFinished: true` and `usage`). The placeholder message isn't finished yet, the guard passes, `updateMessage` runs, and this line fires. ✅
- **Usage trails** — the path `responder.finalize()` adds (responder.ts ~L406-417): the final edit already went out without usage, so a *second* `m.replace` edit is sent to carry it. On the host, `getEffectiveEventId` resolves that second edit back to the same parent event id (`room.ts` L1020-1023), so it maps to the *same* cached message — which the first edit already set `_isStreamingFinished = true`. The guard now fails, `updateMessage` is never called, and this `setUsage` never runs. The counts are dropped. ❌

**Why it looks like it works.** A page reload masks it: on a cold load matrix-js-sdk hands `loadRoomMessage` the aggregated (latest-edit) content and the message isn't in cache, so `coreMessageArgs` (L105) reads `data.usage` at construction — the counts appear after refresh even when the live update was lost. The ai-bot tests here only assert the *send* side (that two room events are emitted); nothing asserts the host applies the second one, which is why this passes CI.

**The way out (host side).** Let a late usage-only edit through even when the message is finished. The finished-guard exists to stop stale previews from *un-finishing* a message (see `applyResponseStreamPreview`'s comment), and a finished→finished edit can't do that. A targeted carve-out in `loadRoomMessage`, just before the existing guard, falling through to the continuation linking below:

```ts
// A late usage-only edit legitimately lands after the message is finished;
// the guard below skips reprocessing, so apply it directly.
if (message?.isStreamingOfEventFinished) {
  let usage = (event.content as CardMessageContent)?.data?.usage;
  if (usage && !message.usage) {
    message.setUsage(usage);
  }
}
if (!message?.isStreamingOfEventFinished) {
  // ...existing build/update...
}
```

For a split answer the follow-up re-sends the last fragment, so `setUsage` lands on that fragment's message and the head still chases it via the `usage` getter — consistent. The send-side alternative (hold the finishing edit until usage is attached) delays the visible answer and is unreliable when a provider never reports usage, so the host-side fix is the better one.

**Severity note.** This is diagnostic-only and a reload recovers it, so you may reasonably judge how much it blocks — but as written, the entire follow-up-edit machinery (and its two tests) has no effect on the live UI, which seems worth closing before merge.

---

packages/host/app/components/matrix/room.gts
```diff
@@ -1105,6 +1135,52 @@ export default class Room extends Component<Signature> {
...
+    let sawCached = false;
+    let costUsd = 0;
+    let sawCost = false;
+    let turnsWithUsage = 0;
+    for (let message of this.messages) {
+      let usage = message.usage;
+      if (!usage) {
+        continue;
+      }
+      turnsWithUsage++;
```
habdelra, line 1166: [Claude Code 🤖]

Two small, non-blocking notes on this getter:

**Evaluated twice per render.** The template references it in both `{{#if this.conversationTokenUsage}}` and the interpolation right after, so this O(n) scan over `this.messages` runs twice on each render while `showTokens` is on. The `messages` getter in this same file is `@cached`; matching that here (or wrapping the two template uses in one `{{#let this.conversationTokenUsage as |total|}}`) collapses it to a single pass. Minor since it's debug-gated, but free.

**Turn gate diverges from the per-message line.** This counts a turn for any message with a truthy `usage` object (`if (!usage) continue`), whereas the per-message line in `ai-assistant/message/index.gts` (`tokenUsage`) hides unless `promptTokens`/`completionTokens` is present. In practice ai-bot always writes both counts, so the two never disagree today — but if a `usage` object ever carried only `cachedTokens`/`costUsd`, the session total would count a phantom turn and sum the missing counts as 0 while the per-message line showed nothing. Gating this loop on `usage.promptTokens != null || usage.completionTokens != null` keeps the two views consistent. Non-blocking.

---

packages/host/app/components/ai-assistant/message/index.gts
```diff
@@ -600,13 +612,41 @@ export default class AiAssistantMessage extends Component<Signature> {
...
   </template>
 
   private get hasBotMessage() {
     return this.args.messageHTMLParts?.length || this.args.reasoningContent;
   }
 
+  // The provider's token counts for the turn that produced this message.
+  // Shown only behind the `showTokens` query param: it is a diagnostic for
+  // measuring prompt cost, not something every reader needs to see.
+  private get tokenUsage() {
```
habdelra, line 638: [Claude Code 🤖]

Same double-eval as the session-total getter: `this.tokenUsage` is referenced in both `{{#if this.tokenUsage}}` (L541) and the interpolation (L546), so `formatTokenUsage` runs twice per render when `showTokens` is on. Marking this `@cached` (or wrapping the block in `{{#let this.tokenUsage as |usage|}}`) removes the redundant call. Non-blocking, debug-gated — noted for consistency with the room.gts thread.

---

habdelra (commented):
[Claude Code 🤖]

This review followed the token counts end to end — ai-bot stream → Matrix event → host `Message` → UI — because the feature's correctness turns on whether the late-arriving usage actually lands on the rendered message, plus the streaming-finish invariant this change edits.

**Bottom line: the send side is careful and well-tested, but there is one correctness gap worth closing before merge — the trailing-usage follow-up edit that `finalize()` sends is dropped by the host in a live session, so in exactly the timing that path exists for, the counts never appear until a page reload.** Everything else is small and non-blocking. Submitted as a comment, not request-changes — the approve/block call is yours, especially as this is diagnostic-only and a reload recovers it.

- The sticky `isStreamingFinished` fix in `onChunk` and its two regression tests are correct. The trap is real: a trailing usage chunk carries no `finish_reason: 'stop'`, so recomputing the flag per-chunk would flip a finished stream back to streaming. Latching it on `responseState.isStreamingFinished` is the right shape, and both provider chunk shapes (empty `choices`, and a choice with `finish_reason: null`) are covered.
- Enabling `usage: { include: true }` doesn't only feed this feature — it also repairs the inline-cost billing fast path (`main.ts` L655 reads `chunk.usage.cost`; without the request option that was always falling back to the slower generation-cost API in `credit-tracking.ts`). The main.ts comment describes this accurately.
- Split-answer usage rides the continuation chain with **no** double-count: `roomResource.messages` filters out `continuationOf` parts and the head chases the chain in the `usage` getter, exactly like `tools`/`body`. The session-total loop therefore counts each answer once.
- The numbered-palette CSS (`--boxel-300/400/600`) is a deliberate, correct exception to the semantic-role-token convention here — the assistant panel is a fixed dark surface and the surrounding meta/timestamp styles already use the same primitives; the inline comments say so.
- Query-param wiring is sound: `showTokens` is in `HOST_APP_QUERY_PARAMS`, the index controller sets `queryParams = HOST_APP_QUERY_PARAMS`, and the env-based default doubles as the serialization baseline as documented.

**Recommendations:**
1. **[blocking for the live path]** Host drops the follow-up usage edit once the message is finished — the added `setUsage` in `message-builder.updateMessage` is never reached for it. Full mechanism + a targeted host-side fix in the `message-builder.ts` thread.
2. **[non-blocking]** `@cached`/`{{#let}}` the two usage getters — each is evaluated twice per render (`{{#if}}` + interpolation); the session-total one is O(n). See the `room.gts` and `message/index.gts` threads.
3. **[non-blocking]** The session-total turn gate counts any truthy `usage` object, diverging from the per-message line's count-present gate. Harmless today, but the two views can drift. See the `room.gts` thread.
4. **[non-blocking]** No host coverage for the UI, `formatTokenUsage`, the `turnsWithUsage < 2` gate, or the late-usage live update. `formatTokenUsage` is a pure function that's trivially unit-testable, and a host test for the late update is what would have caught #1 — the current tests assert only the ai-bot send side.

---

packages/host/app/lib/matrix-classes/message-builder.ts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Fixed in 62e9aa2350 with the host-side carve-out in `loadRoomMessage`, plus an integration test that replays the exact sequence (finishing edit without usage, then a trailing usage edit) — it fails without the carve-out.

packages/host/app/components/matrix/room.gts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Both done in 62e9aa2350: `@cached` on the getter, and the loop now uses the same has-token-counts gate as the per-message line (covered by a test).

packages/host/app/components/ai-assistant/message/index.gts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Done in 62e9aa2350 — `@cached`.

packages/host/app/components/matrix/room.gts
```diff
@@ -1105,6 +1135,52 @@ export default class Room extends Component<Signature> {
     return this.args.roomResource.messages;
   }
 
+  // The session's token bill so far: every turn's counts summed. Input is
+  // summed rather than read off the last turn because each turn re-sends the
+  // whole conversation — the sum is what the provider actually billed. Shown
+  // only behind the `showTokens` query param.
+  private get conversationTokenUsage() {
```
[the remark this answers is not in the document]
jurgenwerk: _(Written by Claude on Matic's behalf.)_ Done in 62e9aa2350 — `@cached`.

packages/host/app/components/matrix/room.gts
```diff
@@ -1105,6 +1135,52 @@ export default class Room extends Component<Signature> {
...
+    let sawCost = false;
+    let turnsWithUsage = 0;
+    for (let message of this.messages) {
+      let usage = message.usage;
+      if (!usage) {
+        continue;
+      }
+      turnsWithUsage++;
+      promptTokens += usage.promptTokens ?? 0;
+      completionTokens += usage.completionTokens ?? 0;
```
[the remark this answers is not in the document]
jurgenwerk: _(Written by Claude on Matic's behalf.)_ Fixed in 62e9aa2350 — the loop now skips usage objects without token counts, same gate as the per-message line.

packages/host/app/components/ai-assistant/message/index.gts
```diff
@@ -534,6 +537,15 @@ export default class AiAssistantMessage extends Component<Signature> {
...
         {{/if}}
+
+        {{#if this.tokenUsage}}
+          <div
+            class='token-usage'
+            data-test-token-usage={{@usage.promptTokens}}
+          >
+            {{this.tokenUsage}}
+          </div>
+        {{/if}}
```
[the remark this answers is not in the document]
jurgenwerk: _(Written by Claude on Matic's behalf.)_ Done in 62e9aa2350 — `@cached` on the getter.

packages/host/app/components/matrix/room.gts
```diff
@@ -1107,6 +1137,58 @@ export default class Room extends Component<Signature> {
...
+    // line right below it; a sum only earns its place once there is
+    // something to add up.
+    if (turnsWithUsage < 2) {
+      return undefined;
+    }
+    return formatTokenUsage({
+      promptTokens,
+      completionTokens,
+      ...(sawCached ? { cachedTokens } : {}),
+      ...(sawCost ? { costUsd } : {}),
```
richardhjtan: [Claude Code 🤖]

**`sawCost` means "at least one turn reported cost", so a mixed session renders a partial sum with a `$` in front of it and nothing saying it's partial.** Non-blocking and diagnostic-only — but it is the number most likely to be screenshotted into a cost discussion, and it is the same view-divergence class as the turn gate this commit just tightened, one level down.

**The mechanism.** A turn enters the accumulator on the token-count gate alone. Cost and cached tokens are then added conditionally:

```js
if (usage.costUsd != null) {
  costUsd += usage.costUsd;
  sawCost = true;
}
```

So a turn with counts but no `costUsd` still increments `turnsWithUsage` and contributes its tokens, while contributing nothing to `costUsd` — and `sawCost`, already `true` from another turn, keeps the cost clause in the output. `formatTokenUsage` then renders `$${usage.costUsd.toFixed(4)}` with no qualifier. The reader sees a four-decimal dollar figure beside a complete token total and reasonably concludes it covers the same turns. It doesn't.

**Reachability.** `costUsd` traces to `chunk.usage.cost`, which the provider supplies under the `usage: { include: true }` option. Any turn whose provider or model omits `cost` while still reporting token counts produces the mixed state — as does any session spanning a model switch, since the room keeps its history. `cachedTokens` has the identical shape via `prompt_tokens_details`, and a mixed cached figure is more likely still, since cache reporting varies more by provider than cost does.

**The fix, matching the gate pattern already here** — count the turns that reported each optional figure, and only present it when every counted turn did:

```js
    let turnsWithCost = 0;
    let turnsWithCached = 0;
    // ...
      if (usage.cachedTokens != null) {
        cachedTokens += usage.cachedTokens;
        turnsWithCached++;
      }
      if (usage.costUsd != null) {
        costUsd += usage.costUsd;
        turnsWithCost++;
      }
    // ...
    return formatTokenUsage({
      promptTokens,
      completionTokens,
      // An optional figure is only a session total when every counted turn
      // reported it; a partial sum would read as the whole bill.
      ...(turnsWithCached === turnsWithUsage ? { cachedTokens } : {}),
      ...(turnsWithCost === turnsWithUsage ? { costUsd } : {}),
    });
```

That drops the figure rather than qualifying it, which is the conservative choice and needs no formatter change. If losing it entirely is too blunt — a session where one old turn lacks cost would hide cost for the whole session — the alternative is to keep the sum and mark it, which does mean teaching `formatTokenUsage` a "partial" flag. I'd start with dropping it; a diagnostic that goes quiet is better than one that quietly understates.

---

packages/host/app/components/ai-assistant/message/index.gts
```diff
@@ -534,6 +538,15 @@ export default class AiAssistantMessage extends Component<Signature> {
...
```
richardhjtan, line 546: [Claude Code 🤖]

**The test hook is bound to a value the widened gate explicitly allows to be missing, so it disappears from the DOM in exactly that case.** Non-blocking, but it makes the hook unreliable for the states this feature now supports — and I think it already cost the new test something.

**The mechanism.** `data-test-token-usage={{@usage.promptTokens}}`: Glimmer omits an attribute whose value is `null` or `undefined` rather than rendering it empty. Meanwhile `tokenUsage` — and the session-total gate this commit aligned it with — admits a usage object carrying *either* count:

```js
usage.promptTokens == null && usage.completionTokens == null
```

So a usage object with `completionTokens` and no `promptTokens` renders this `<div>` with the text `? in · 340 out` (that's what `formatTokenUsage`'s `?` fallback is for), while `[data-test-token-usage]` is **not present in the DOM at all**. Any test or debugging query keyed on that attribute reports the usage line as absent while it is on screen.

**The corroboration.** The new test's negative assertion reaches for the styling class instead of the hook:

```js
assert.dom('[data-test-message-idx="0"] .token-usage').doesNotExist();
```

while every positive assertion in the same file uses `[data-test-token-usage]`. Selecting on `.token-usage` works today but pins the assertion to a CSS class name, so renaming the class turns the precondition into a vacuous pass — it would then "not exist" for the wrong reason, and the test would still be green.

**The fix.** Make the hook depend on the element existing rather than on a data value, which lets both the positive and negative assertions use it:

```suggestion
        {{#if this.tokenUsage}}
          <div class='token-usage' data-test-token-usage>
            {{this.tokenUsage}}
          </div>
        {{/if}}
```

Then the test's negative assertion becomes `assert.dom('[data-test-message-idx="0"] [data-test-token-usage]').doesNotExist()`, and the whole file keys on one selector. If the prompt-token count is wanted as an attribute value for some assertion, it belongs in a second attribute that is allowed to be absent, not in the presence hook.

---

richardhjtan (commented):
[Claude Code 🤖]

**Lens:** the threads on this change are all answered and the fixes are pushed, so this pass verified the *fix* rather than the original code — reading the commit itself instead of trusting its description, and re-running the twin checks on what it touched. Then it looked at the numbers the feature actually puts on screen, since that is the part no thread has covered.

**Bottom line: the late-usage carve-out is correct, and I verified the two mechanisms that could plausibly have made it inert — it is not. Two new non-blocking findings, both about the numbers rather than the plumbing: the session total presents a partial cost sum as a total, and the per-message test hook disappears in exactly the case the new gate was widened to admit.**

**Disposition of the pushed fix — verified, not taken on trust.** Two ways the carve-out could have silently done nothing, both checked:

- **`content.data` is a JSON string on the wire, so `.data?.usage` looks like it should read `undefined`.** `sendMatrixEvent` in `runtime-common/ai/matrix-utils.ts` does `content.data = JSON.stringify(content.data)` before sending. It resolves correctly because `Room.addEvent` parses it back at ingestion (`typeof event.content.data === 'string'` → `JSON.parse`), which runs before `loadRoomMessage` ever sees the event. So the read is against an object, and it matches how `message-builder.ts` already reads `.data?.usage`. Placement is also right: the block sits *after* `getAggregatedReplacement`, so on a live edit it reads the edit's own content and on a reload it reads the aggregated latest.
- **`message.usage` chases the continuation chain forward** (`continuedInMessage?.usage ?? this._usage`), so `!message.usage` could in principle be satisfied by a *different* part of a split answer and block the write. It can't: `getEffectiveEventId` resolves the trailing `m.replace` to its parent, which is the part the publisher attached usage to — and the publisher attaches it only to the last part ("Only the last part of a split answer carries the usage"), which by definition has no `continuedInMessage`. So the getter degenerates to `_usage` for exactly the message the guard is applied to.
- **Reactivity holds under the new `@cached`.** `_usage` is `@tracked` and `setUsage` writes it, so `@cached` on both getters invalidates on the late write rather than pinning a stale value — which is the one way adding `@cached` could have broken the fix it shipped alongside.
- **The regression test replays the real shape**, not a convenient one: a finishing edit with no usage, then a second `m.relates_to: { rel_type: 'm.replace' }` edit carrying `data.usage`, asserted through the rendered DOM. That is the sequence the carve-out exists for.

The turn-gate and `@cached` items are likewise correctly applied. All threads resolved from my side.

**New findings.**

1. **The session total's cost and cached-token figures are partial sums presented as totals.** Inline thread on the accumulator. Same class of view-divergence as the turn gate just fixed, one level down.
2. **`data-test-token-usage` is bound to a value that the widened gate allows to be absent**, so the hook vanishes from the DOM in that case — which appears to be why the new test's negative assertion had to fall back to a styling class. Inline thread.

**Still open from the earlier coverage note.** The new integration tests cover the three behaviors that were broken or ambiguous, which is the right priority — but `formatTokenUsage` itself is still untested, and it is a pure function with four distinct branches (the `?` fallback for a nullish count, the cached clause, the cost clause, and `toFixed(4)` rounding). It is the one piece of this feature whose output is a string a human reads and compares against a provider bill, and a unit test for it costs a few lines. Worth closing the last part of that thread.

---

packages/host/app/components/matrix/room.gts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Fixed in bfcfd43e16 with the per-figure turn counts — an optional figure now appears only when every counted turn reported it, covered by a mixed-session test.

packages/host/app/components/ai-assistant/message/index.gts
jurgenwerk: _(Written by Claude on Matic's behalf.)_

Done in bfcfd43e16 — `data-test-token-usage` is now a bare presence attribute and the test's negative assertion keys on it instead of the styling class.

lukemelia (approved):

PR #5757, opened 2026-08-12
Port @cardstack/bxl into packages/bxl, retaining its git history

habdelra:
BXL (Boxel Expression Language) is the platform's spreadsheet-formula language: readable Excel-style syntax compiled to canonical jq and evaluated by a sandboxed TypeScript jq engine — no I/O, budget-limited, with execution profiles for `computeVia` derivation, query predicates, authorization, and mutation planning. It bundles a curated Formula.js subset (300+ Excel functions, heavy families as lazy chunks), validator.js helpers, and our own readable-syntax compiler/linter/formatter. Until now it lived in its own repository, and realms consumed it as a generated per-realm bundle — no single source of truth, no CI coverage against the real card runtime, and platform code couldn't call it at all.

This PR moves the whole library into the monorepo as its own workspace package, `packages/bxl`, keeping the `@cardstack/bxl` package identity.

## History-preserving import

The original repo's history is rewritten into `packages/bxl` with `git filter-repo --to-subdirectory-filter` and merged with `--allow-unrelated-histories`, so the original commits are reachable here:

```
$ git log --follow --format='%h %ad %s' --date=short -- packages/bxl/src/index.ts | tail -1
fb1524f81e 2026-04-22 Port BXL from staging tree to three-origin-folder layout
```

Standalone-repo scaffolding did not come along: its CI workflows, lockfile, editor dotfiles, eslint/typedoc config, the esbuild build and per-realm bundle scripts, the browser perf harness and demo site, and planning/handoff docs. What moved is the library itself (jq engine, Excel formulas, readable-syntax compiler, expression factory, authorization, mutation, types), its tests, examples, docs, and the license/attribution files (`NOTICE.md`, `LICENSES/` — we vendor jq-tools and Formula.js, so attribution must survive).

## Conforming to monorepo conventions

Modeled on runtime-common, the reference isomorphic package:

- **Raw erasable TypeScript, no build step.** The `exports` map points at `src/*.ts`; Node 24 runs the sources directly and the host bundles them. All relative import specifiers are `.ts` (Node's type stripping resolves specifiers literally), the two `enum`s became `as const` objects paired with union types, and constructor parameter properties became explicit class fields.
- **Deps via the pnpm catalog** (`bessel`, `jstat`, `validator`, plus dev deps), with new catalog entries in `pnpm-workspace.yaml`.
- **Repo-standard linting.** `lint` / `lint:js` / `lint:types` scripts; the package lints under the root ESLint config (prettier, `import type`, the erasable-syntax selectors) with `tsc --noEmit` for types. The standalone tsconfig only checked `src/`, so tests and scripts carried latent type errors — the new tsconfig covers everything and those are fixed.
- **Package-local test runner under native type stripping.** `pnpm test` spawns each suite as a plain Node process — no loader hook. All 63 suites (~1,300 cases) pass, matching a baseline run of the original repo before the port.

## Evergreen docs pass

The imported prose carried standalone-repo history: pointers to deleted planning docs, instructions for the retired build/bundle-sync flow, doc indexes listing files that never existed. Every reference now resolves and describes the package as it is — `docs/README.md`, the test READMEs, and the main README were rewritten against what the directories actually contain, and comments that narrated old fixes now state the rule being upheld.

## CI

`ci-lint` gains a BXL lint step. `ci.yaml` gains a change-check filter and a `BXL Tests` job that runs the suite — plain Node against plain objects, no services stack, no browser.

## Verification

- `pnpm lint` and `pnpm test` (63/63 suites) green in `packages/bxl`
- `git log --follow` on ported files reaches the original commits
- every package script (`example:*`, `bench:authorization`, `fixtures:authorization:verify`, `test:authorization:*`) runs clean


packages/bxl/package.json
```diff
@@ -0,0 +1,84 @@
...
+    "bench:authorization": "node scripts/bench-authorization.ts",
+    "example:authorization": "node examples/authorization/run.ts",
+    "example:mutation": "node tests/unit/bxl-mutation-fixtures-cli.ts",
+    "example:mutation:realm": "node tests/unit/bxl-mutation-realm-fixtures-cli.ts",
+    "fixtures:authorization:verify": "node scripts/verify-authorization-fixtures.ts",
+    "lint": "concurrently \"pnpm:lint:*(!fix)\" --names \"lint:\"",
+    "lint:fix": "concurrently \"pnpm:lint:*:fix\" --names \"fix:\"",
+    "lint:js": "eslint . --report-unused-disable-directives --cache",
+    "lint:js:fix": "eslint . --report-unused-disable-directives --fix",
+    "lint:types": "tsc --noEmit",
```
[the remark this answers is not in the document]
habdelra: agreed, glint is a superset of tsc, and it's better to remain consistent across all the pacakges in this mono repo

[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Agreed — fixed in c11aaf9471: `lint:types` now runs `ember-tsc --noEmit` (with `@glint/ember-tsc` as a dev dependency), matching every other package in the workspace.

packages/bxl/package.json
```diff
@@ -0,0 +1,84 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/cardstack/boxel/issues"
+  },
+  "type": "module",
+  "engines": {
+    "node": ">=24"
+  },
+  "exports": {
+    ".": "./src/index.ts",
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Correct that a plain-Node npm consumer cannot run raw `.ts` out of `node_modules` — Node refuses to type-strip there. That constraint does not bite inside this repo: the host bundles the sources, and the realm-server consumes the package through the pnpm workspace link, whose realpath resolves outside `node_modules`, so type stripping applies. This is the same shape as `@cardstack/runtime-common`, the repo's reference isomorphic package.

No workflow publishes this package from the monorepo; the `publishConfig`/`files` stanzas only preserve the package's npm identity. Wiring up publishing is a separate piece of work, and its pack step will need to emit JavaScript entry points (plus declarations) exactly as you describe — the raw-`.ts` exports map is the in-repo consumption contract, not the wire format for npm.

backspace (approved):

.github/workflows/ci.yaml
```diff
@@ -219,6 +226,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+    concurrency:
+      group: bxl-test-${{ github.head_ref || github.run_id }}
+      cancel-in-progress: true
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: ./.github/actions/init
+      - name: BXL test suite
+        run: pnpm test
+        working-directory: packages/bxl
```
richardhjtan, line 242: [Claude Code 🤖]

**This job runs `pnpm test`, which does not reach the OpenFGA authorization conformance corpus — so the authorization kernel lands in the monorepo with the same absence of CI coverage the port exists to fix.** Follow-up rather than a blocker, but it is squarely against this change's own stated goal, and the fix is one step.

**The mechanism.** `scripts/run-tests.mjs` hardcodes what it scans:

```js
const allRoots = ['tests/unit', 'tests/smoke', 'tests/boxel'];
```

`tests/authorization` is not in that list, and its three `.ts` files are all support modules — `openfga-conformance-core.ts`, `openfga-conformance.ts`, `openfga-fixtures.ts`. Nothing in them is a self-executing suite, so even adding the root would not run them. The corpus is driven instead by `scripts/run-authorization-conformance.ts`, wired to the separate `test:authorization:conformance` script, which nothing in either workflow invokes.

**Verified by running the discovery.** The runner finds exactly 41 + 13 + 9 = 63 suites and never touches `tests/authorization/`, whose fixture tree includes a commit-pinned vendored OpenFGA corpus (`tests/authorization/fixtures/openfga/<commit>/`). So the conformance assertions — Check, ListObjects, ListUsers — run only when someone runs that script by hand.

That matters more than a normal coverage gap because authorization is one of the four execution profiles the description names, the corpus is pinned to an upstream commit (so it can drift when the pin is bumped), and the description's own framing is that the library previously had "no CI coverage against the real card runtime."

**The way out is cheap, because the script is already CI-shaped** — `run-authorization-conformance.ts` sets `process.exitCode = 1` on any failure, importer failure, or unsupported case, and catches and reports thrown errors the same way. So it only needs invoking:

```yaml
      - name: BXL test suite
        run: pnpm test
        working-directory: packages/bxl
      - name: BXL authorization conformance
        if: ${{ !cancelled() }}
        run: pnpm run test:authorization:conformance
        working-directory: packages/bxl
```

The alternative — folding it into the `test` script so `pnpm test` means "everything" — is tidier for local use and removes the chance of the next suite-with-its-own-runner going unnoticed. Either works; the second is the one that stops this recurring.

---

packages/bxl/scripts/run-tests.mjs
```diff
@@ -0,0 +1,51 @@
...
+
+const filterArg = process.argv[2];
+const allRoots = ['tests/unit', 'tests/smoke', 'tests/boxel'];
+const roots = filterArg ? [filterArg.replace(/\/$/, '')] : allRoots;
+
+const suites = roots.flatMap((root) =>
+  readdirSync(join(packageRoot, root))
+    .filter((f) => f.endsWith('.ts'))
+    .map((f) => join(root, f)),
+);
```
richardhjtan, line 30: [Claude Code 🤖]

**Two properties of this discovery worth making deliberate, because both fail as a green tick rather than as an error.** Non-blocking; the runner is correct today and I verified its exit-code behavior is sound (it exited 7 with 7 failing suites, so failures do propagate).

**1. `readdirSync` is not recursive, so a suite in a subdirectory is silently skipped.** Today this is benign — the only nested `.ts` under the three roots is `tests/boxel/fixtures/hospital.ts`, a fixture rather than a suite, so skipping it is correct. But the reason it's correct is a coincidence of the current layout, not a rule the code states. A future `tests/unit/mutation/round-trip.ts` would never execute, and the summary would still read `N / N suites passed` — the count is derived from what was discovered, so it can never reveal what wasn't.

Node's `readdirSync(dir, { recursive: true })` covers it, with fixtures excluded by path rather than by depth:

```js
const suites = roots.flatMap((root) =>
  readdirSync(join(packageRoot, root), { recursive: true })
    .map(String)
    .filter((f) => f.endsWith('.ts') && !f.split(/[\\/]/).includes('fixtures'))
    .map((f) => join(root, f)),
);
```

That also means a new root directory gets picked up only if added to `allRoots` — which is the gap the authorization corpus falls into, covered in the thread on the CI job.

**2. A suite that asserts nothing passes.** The pass/fail contract here is exit status alone, so a suite whose assertions are never reached — file truncated, an early `return` in a guard, a top-level `if` that stops matching — exits 0 and prints `OK` with an empty summary column. Every suite already ends with a self-describing count line (`150 cases passed`, `22/22 cases passed`, `7/7 passed`), so that line is available as evidence the body actually ran:

```js
  if (result.status === 0 && summary) {
    console.log(`OK   ${label} ${summary}`);
  } else if (result.status === 0) {
    failures++;
    console.log(`FAIL ${label} (exited 0 but printed no summary)`);
  } else {
```

Cheap, and it converts the one failure mode this runner cannot currently see into a normal red.

---

richardhjtan (commented):
[Claude Code 🤖]

**Lens:** a history-preserving port of a working library is not usefully reviewed line by line — the library already has its own history and its own test suite. What is genuinely new here, and what a reviewer can actually get wrong, is the monorepo wiring: whether the no-build-step consumption contract really holds, and whether the CI the port exists to gain actually covers the library. So this pass verified the description's mechanical claims by executing them, and traced what the new CI jobs do and don't run.

**Bottom line: the wiring is sound and every mechanical claim in the description checks out under execution — no blocking issues. One real gap: the OpenFGA authorization conformance corpus is not reachable from the new CI job, so the one execution profile with a vendored fixture corpus stays exactly as uncovered as it was outside the monorepo.** That is the PR's own stated motivation, which is why I'd act on it rather than file it.

**Verified by running it, not by reading it.**

- **The no-build-step contract holds.** I ran suites from all three roots under plain Node 24 in a clean checkout with **no `node_modules` at all** — `tests/unit/bxl-150-cli.ts` (150 cases), `tests/boxel/compiler-readable.ts` (22/22), `tests/smoke/boxel-runtime-async.ts` — and all three passed. Raw `.ts` with `.ts` specifiers executes directly, as claimed; nothing needs a loader hook.
- **Every erasable-syntax claim is true across the whole package, not just the files that motivated it.** Zero `enum` declarations remain; zero constructor parameter properties; zero `namespace` declarations or decorators. And every relative specifier carries an extension — the single relative import not ending in `.ts` is `tests/unit/eslint-no-bxl-escape-gotcha.ts` importing `'../../eslint-rules/no-bxl-escape-gotcha.js'`, which resolves to a genuine `.js` file. So the specifier-literal requirement is met with no exceptions.
- **"63 suites" is exact, not rounded** — the runner discovers 41 + 13 + 9 across its three roots, which is precisely 63.
- **The runner propagates failures, so the CI job can actually go red.** With the external deps absent, 7 suites failed on `ERR_MODULE_NOT_FOUND` and the process exited **7** — one per failing suite. Worth stating because a hand-rolled runner that swallows exit codes would make the new job decorative, and this one doesn't. (The 7 failures are entirely `bessel` / `jstat` / `validator` / `esbuild` not being installed in my checkout; 56/63 passing with zero dependencies makes the full-green claim credible.)
- **The exports map is complete and consistent with the reference package.** All nine explicit targets resolve to files that exist, and all sit inside the `files` allowlist. The shape — raw `.ts` targets plus a `"./*": "./src/*.ts"` wildcard — is the same shape `runtime-common` uses (`"./*": "./*.ts"`), so the description's "modeled on runtime-common" is accurate rather than aspirational.
- **No orphan devDependencies.** `esbuild` is used by `tests/smoke/bundle-first-filter-chain.ts`; `yaml` and `@openfga/syntax-transformer` by `tests/authorization/support/openfga-conformance-core.ts`. Each new catalog entry has a live consumer.
- **The change-filter and lint wiring are right.** The `bxl` filter output is declared, consumed by the job's `if`, and the `*shared` inclusion is deliberate and explained. The lint step matches its siblings' `if: ${{ !cancelled() }}` shape.

**Recommendations.**

1. **Add the authorization conformance suite to CI.** It already exits non-zero on failure, so this is one step. Inline thread on the new job.
2. **Suite discovery silently skips anything it doesn't expect.** Non-blocking, but the failure mode is a green tick. Inline thread on the runner.

**On the existing threads.** Both Codex findings are settled and I'm not reopening either — but I did verify the premise the second dismissal rests on, since it was load-bearing for waving off a P1: the only publish workflow in the repo is `boxel-cli-publish.yml`, and nothing anywhere publishes this package, so the raw-`.ts` exports map is indeed an in-repo consumption contract rather than a broken npm artifact today. The `lint:types` fix to `ember-tsc --noEmit` is in place as described.

**Adjacent, out of scope.** No job in `ci.yaml` except one sets `timeout-minutes`, so the new job's absence of one matches convention — noting only so it doesn't read as an oversight. Separately, `main` currently has no `required_status_checks` configured and no repository rulesets, so neither new job gates a merge regardless of how it's written; that is a repo-settings matter, not this PR's.

---

.github/workflows/ci.yaml
habdelra: [Claude Code 🤖] The corpus does run under `pnpm test` — the premise that it is only reachable by hand misses one hop. `tests/unit/authorization-fixtures-cli.ts` is one of the 63 discovered suites, and it spawns both `scripts/verify-authorization-fixtures.ts` and `scripts/run-authorization-conformance.ts` as subprocesses, then asserts on their accounting: `discovered=1227 passed=1227`, `failed=0`, `importer_failures=0`, and `unsupported=0` must all appear or the suite fails. So the CI job executes every Check/ListObjects/ListUsers assertion in the pinned corpus on every run, and a drifted pin surfaces as a red suite, not a silent skip.

Verified by running it: `node tests/unit/authorization-fixtures-cli.ts` → "Authorization fixtures: pinned hashes and executable 1,227-case zero-skip accounting verified", and the discovery scan lists it among the 63.

Your discovery-robustness point still stands on its own merits — addressed in 4a4fa4ecba (see the other thread).

packages/bxl/scripts/run-tests.mjs
habdelra: [Claude Code 🤖] Both taken — fixed in 4a4fa4ecba. Discovery now uses `readdirSync(..., { recursive: true })` with fixtures excluded by path segment rather than by depth, and a suite that exits 0 without printing its final summary line is reported as `FAIL (exited 0 but printed no summary)` and counted.

Both were trip-tested before pushing: a throwaway `tests/unit/nested-trip/trip.ts` was discovered and ran, and the same file rewritten to exit cleanly without output was flagged as a failure. The full suite still reports 63/63 — recursion picks up no stray modules beyond the fixtures it excludes.

PR #5758, opened 2026-08-12
Data (JSON & CSV) FileDef formats on the shared shells

lukemelia:
Migrates the JSON and CSV FileDef families onto the shared FileDef format shells (atom/embedded/fitted/isolated), so each family supplies only a `static previewComponent` while identity/facts/budgets/state live in the shared shell layer. Sibling of #5746 (text/markdown) and #5748 (code); built on the shared shells already on `main`. Completes the "priority families" milestone (text, code, data, markdown).

## Changes
- **`csv-file-def.gts`** — drop the bespoke templates; a `CsvPreview` renders the parsed rows as a real table (embedded budgets to the first 20 rows with a "… N more rows" note, isolated shows all), and a fitted cell shows a rows × columns count instead of a cropped table. Pin the data profile axes. Extraction (columns/columnCount/rowCount) unchanged.
- **`json-file-def.gts`** — replace the flat highlighted pretty-print with a collapsible **structured tree** (recursive native `<details>`/`<summary>`, so it stays interactive in prerendered HTML; type-colored values; root and direct children start expanded; a 2000-node budget, 200 in an embedded pane), with a raw-text fallback for invalid JSON, and a fitted summary that branches on the root: objects count keys, arrays count items, a scalar root shows its type, and an unparsed document is labeled "Invalid JSON". Now extracts `rootType`/`keyCount`/`lineCount`. Pin the data profile axes.
- **`file-shell-isolated.gts`** — a full-width, auto-height data document stage for the `json`/`csv` preview-kinds; the full-width column rules are shared with the other document kinds in one selector list, while each kind's stage keeps its own floor (CSV theme-aware, JSON dark to match its inspector panel).
- **Tests** — `rootType`/`keyCount` assertions on the JSON extraction test, and a new integration test covering the tree/table render across shells, the auto-open depths, the embedded node budget, the fitted count summaries (including scalar-root and invalid-JSON cells), the invalid-JSON raw-text fallback, and the kind labels.

## Verification
- `ember-tsc` typecheck, template-lint, eslint (test files): clean
- JSON acceptance (extraction / indexing / `rootType` / `keyCount`): 3 tests, 20 assertions
- CSV acceptance (extraction / indexing): 3 tests, 30 assertions
- Integration render (tree/table across shells, auto-open depths, embedded node budget, fitted summaries, invalid-JSON fallbacks, "JSON" / "CSV data" kind labels): 11 tests, 39 assertions

## Notes
- JSON's tree uses a dark data-inspector surface with the VS Code palette (contrast-safe, matches the code preview); CSV keeps the theme-aware light table. JSON's shell hero fact stays "N lines"; the key count lives in the fitted preview summary.
- The pre-commit hook reports an eslint "Parsing error" on the base `.gts` files; this is a pre-existing `ember-eslint-parser`-on-base-`.gts` artifact already on `main` and is not run by CI (base's `lint` script runs no eslint). glint is the type-check of record and passes.

Linear: https://linear.app/cardstack/issue/CS-12236


packages/base/json-file-def.gts
```diff
@@ -66,399 +86,349 @@ function truncateExcerpt(text: string): string {
...
-    // `prettyPrintJson()` formats plain text, `highlightJson()` escapes that text,
-    // and only then adds the span wrappers we control for styling. Running the
-    // result back through DOMPurify adds DOMParser churn during prerender/indexing
-    // without providing extra protection.
-    return htmlSafe(highlightJson(prettyPrintJson(content)));
   }
+  let children = tag('div', childrenHtml, 'class="jt-children"');
+
+  let attrs =
+    depth < TREE_AUTO_OPEN_DEPTH ? 'class="jt-node" open' : 'class="jt-node"';
```
habdelra: [Claude Code 🤖]

**The tree opens one level less than the constant documents.** `TREE_AUTO_OPEN_DEPTH = 1` is commented as "Root and its direct children render expanded," but the `open` attribute is gated on `depth < TREE_AUTO_OPEN_DEPTH`, so only the root (depth 0) opens; a direct child that is itself an object/array (depth 1) fails `1 < 1` and starts collapsed.

Traced against this PR's own acceptance fixture (`config.json` = `{ name, version, dependencies: { lodash } }`): the root renders `<details open>`, `name`/`version` are leaf rows, but `dependencies` renders `<details>` with no `open` — collapsed. So a reader lands on a preview whose nested objects are all shut one level in, contrary to the stated intent.

The integration test (`renders a structured tree`) only asserts the tree contains the text `dependencies`, which is the collapsed `<summary>` label, so it passes either way and doesn't pin this.

Pick the intent and make code and comment agree:
- to expand the root **and** its direct children, compare inclusively — `depth <= TREE_AUTO_OPEN_DEPTH` (or keep `<` and set the constant to `2`);
- to expand only the root, reword the constant's comment to say so.

Non-blocking, but it's a real behavior/comment mismatch in new code. (Copilot flagged the same line; this confirms it with the concrete render trace.)

---

packages/base/json-file-def.gts
```diff
@@ -66,399 +86,349 @@ function truncateExcerpt(text: string): string {
...
+      return this.keyCount === 1 ? 'item' : 'items';
     }
-    // This component produces the entire HTML payload itself:
-    // `prettyPrintJson()` formats plain text, `highlightJson()` escapes that text,
-    // and only then adds the span wrappers we control for styling. Running the
-    // result back through DOMPurify adds DOMParser churn during prerender/indexing
-    // without providing extra protection.
-    return htmlSafe(highlightJson(prettyPrintJson(content)));
+    return this.keyCount === 1 ? 'key' : 'keys';
   }
```
habdelra, line 292: [Claude Code 🤖]

**A scalar-root JSON reads as "0 keys" in the fitted summary.** `countLabel` returns `key`/`keys` for every non-array root, and the template renders `{{this.keyCount}}` beside it. `entryCount` (the source of `keyCount`) returns 0 for a scalar root, so a valid top-level string / number / boolean / null document — e.g. a file whose entire content is `42` — renders "0 · keys · number" in a fitted cell.

Verified: `extractAttributes` sets `rootType` to the scalar kind and leaves `keyCount` at 0, so the label is both wrong (it isn't an object) and vacuous (0).

It's a narrow case — a `.json` whose root is neither object nor array is unusual — so I'd treat it as a follow-up rather than a blocker. If you want to close it now, branch `countLabel` on `rootType`: `object` → key(s), `array` → item(s), anything else → a `value` label (showing the leaf rather than `0`). Objects and arrays, the overwhelmingly common roots, are already correct. (Same issue Copilot raised on `countLabel` and the `keyCount` template binding — folding both here.)

---

packages/base/csv-file-def.gts
```diff
@@ -16,9 +17,13 @@ import {
   type ByteStream,
   type SerializedFile,
 } from './file-api';
+import type { FilePreviewSignature } from './file-formats/file-preview-stage';
 import { fencedCodeBlock } from './markdown-helpers';
 
 const EXCERPT_MAX_LENGTH = 500;
+// An embedded cell shows the head of the table under a fade; the isolated view
+// shows the whole thing.
+const EMBEDDED_MAX_ROWS = 20;
```
habdelra: [Claude Code 🤖]

**"under a fade" no longer describes the code.** This comment — plus the identical phrasing on `get tableHtml`, and the PR description's "embedded budgets to the first 20 rows under a fade" — describes the *previous* embedded CSV, which used a `mask-image` gradient. The refactor replaced that with an explicit affordance: `csvToHtml(content, 20)` renders the first 20 rows and appends a `… N more rows` note, and the shell's 220px `.emb-body` scrolls (`.data-preview { overflow: auto }`).

Verified there is no mask left in the render path: a search for `mask-image` / `-webkit-mask` across `file-formats/` and the data defs returns nothing, and the only `linear-gradient` in the shells is the loading-skeleton shimmer. So the embedded preview clips/scrolls with a "more rows" note — not a fade.

The explicit note is arguably the better affordance, so the cheapest fix is to align the words to the code: drop "under a fade" from both comments and the PR description (e.g. "budgets to the first 20 rows with a *… N more rows* note"). If a fade *is* wanted, `data-mode='embedded'` is already set on the wrapper and unused in the CSS — that's the hook to reintroduce a mask on. Either way the comment should describe what the code does. Non-blocking.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** does the JSON/CSV migration plug into the shared-shell contract the same way the already-merged text/code/markdown families do, and are the new index-time facts and the tree/table rendering correct? That's where a "migrate onto the shared shells" change either quietly diverges from its siblings or ships a render bug.

**Bottom line: no blocking issues.** The migration is faithful to the shared-shell contract. I verified the pinned axes (`fileFamily`/`fileKind`/`previewKind`/`previewAdapter`/`previewSource`) against `fileViewModel` + `file-type-profile` (read as class statics with a registry fallback) and against the markdown/text siblings, which use the identical shape; `previewComponent` is picked up by the default `file-def-*` templates; the removed bespoke atom/embedded/fitted/isolated leave no orphaned references or stale `data-test-*` selectors; and the isolated-shell `json`/`csv` rules mirror the `code` full-width-document pattern already on `main`. The three findings below are all non-blocking — two comment/label mismatches and one narrow edge case.

**What lands right (worth keeping):**
- The fitted summary is read from index-time facts (`rowCount`/`columnCount`, `rootType`/`keyCount`) so a collection cell never parses the document — exactly the projection-budget intent, and it's honored.
- The JSON tree is XSS-safe by construction: every key and value passes through `escapeHtml` before `htmlSafe`, so the "no second sanitizer pass" claim holds; the invalid-JSON `{{this.rawText}}` fallback is auto-escaped by Glimmer.
- `lineCount` earns its place: `heroFactFor` surfaces "N lines" for `previewKind: 'json'`, so the field has a live consumer rather than being dead extraction.

**Numbered recommendations (all non-blocking):**
1. Tree auto-open depth doesn't match `TREE_AUTO_OPEN_DEPTH`'s comment — inline thread on `json-file-def.gts`.
2. Scalar-root JSON renders "0 keys" in fitted — inline thread on `countLabel`.
3. "under a fade" comments + PR description no longer match the code (no mask remains) — inline thread on `csv-file-def.gts`.

**Confirmation / guard rail:** `JsonPreview`/`CsvPreview` read `this.args.model?.source?.content` — the *raw* FileDef content — rather than the fitted-budgeted `viewModel.contentPreview`. That's safe today only because the fitted branch renders counts, never `content`. If a future edit makes the fitted branch render `content` (or the tree/table), it bypasses the collection-cell budget the projection promises. A one-line note on the `content` getter that fitted must stay count-only would lock that in.

**Adjacent, out of scope (for whoever touches it next, not this PR):**
- `tag` / `spanWrap` / `escapeHtml` and the five escape-regex constants are now duplicated verbatim in `json-file-def.gts` and `csv-file-def.gts` (a third `highlightJson` lives in `metadata-panel.gts`). A shared `html-escape` helper would keep them from drifting apart.
- The JSON preview uses bare greys (`#9d9d9d`, `#808080`) where the merged code sibling uses `var(--boxel-400, #808080)` for the same dark-inspector role; matching it keeps the inspector palette in one vocabulary. (The VS Code value hexes match the code sibling, so those are fine as-is.)

---

habdelra (approved):

richardhjtan (commented):
[Claude Code 🤖]

**Lens:** the shared-shell contract and the pinned axes are already covered by the existing review, so this pass took a different angle — it treated JSON and CSV as two families landing in one change and asked where they were merged that they should have been split, and split where they should have been merged. That is where a two-family migration goes wrong, and it turned up one visual defect plus a structural duplication the isolated shell is accumulating.

**Bottom line: no blocking issues, and I agree with the existing approval. One real cosmetic defect — JSON's isolated stage paints a light background behind a deliberately dark preview, which shows as a light band under short JSON documents — plus a layout decision now spread across three duplicate rule blocks, and a budgeting asymmetry between the two families this change introduces.**

**What lands right (verified, not courtesy).**

- **The escaping claim on `treeHtml` is true on every path, which matters because the comment asserts it and the output goes through `htmlSafe`.** I walked all six branches `kindOf` can produce: keys go through `escapeHtml` in `keyLabelHtml`; string, number, boolean and null values go through `leafValueHtml`, where the string branch escapes the quoted form and the numeric/boolean branch escapes `String(value)`; the `{n}` / `[n]` meta labels interpolate only a `length`; and every `class=` and tag name in `tag`/`spanWrap` is a literal. `escapeHtml` also replaces `&` before the other four, which is the order that avoids double-encoding. `JSON.parse` cannot yield a `typeof` outside those six, so there is no unescaped path. The invalid-JSON fallback renders through `{{this.rawText}}` in a `<pre>`, which Glimmer escapes on its own.
- **Reading the fitted summary from index-time facts rather than parsing is the right call and is actually honored** — `rootType`/`keyCount` for JSON and `rowCount`/`columnCount` for CSV all read `this.source.*`, so a collection cell never calls `JSON.parse` or `parseCsv`. `columnCount` falling back to `this.columns.length` when the declared count is absent is a sensible belt-and-braces.
- **Native `<details>` for the tree is the right primitive** given prerendered HTML has no JS: collapse state survives with no hydration, which a JS-driven tree would not.

**Recommendations.**

1. **JSON's isolated stage background should be dark, not `--card`.** The one finding I'd call a defect rather than a cleanup. Inline thread on the new `.iso-stage` rule.
2. **The full-width-document layout decision now lives in three separate rule blocks with three near-identical comments.** This change adds the third. Inline thread on the new `.iso-cols` rule.
3. **CSV budgets its embedded render; JSON does not.** The two families diverge on this within one change. Inline thread on `treeHtml`.
4. **The comment above the new fields attributes the wrong hero fact to the shells.** Inline thread; the description has it right, the comment does not.

**On the existing threads.** I re-derived the `TREE_AUTO_OPEN_DEPTH` off-by-one and the `countLabel` scalar-root case independently and agree with both as stated — no need to restate them here. One extension worth folding into the `countLabel` thread rather than opening a new one: invalid JSON reaches the same vacuous summary by a shorter and much more likely route than a scalar root, since `extractAttributes` catches the parse failure and stores `rootType: ''` with `keyCount: 0`. A truncated or hand-edited `.json` therefore renders "0 · keys" in a fitted cell with the type line suppressed by `{{#if this.rootType}}`. Branching `countLabel` on `rootType` closes both cases at once provided `''` is treated as its own branch. Posting that in the existing thread.

---

packages/base/file-formats/file-shell-isolated.gts
```diff
@@ -747,6 +759,15 @@ export class FileIsolatedShell extends GlimmerComponent<FileIsolatedShellSignatu
...
       }
+      /* A data table or tree reads the same way, but clips at its rounded
+         border instead of overflowing. */
+      .iso[data-preview-kind='json'] .iso-stage,
+      .iso[data-preview-kind='csv'] .iso-stage {
+        height: auto;
+        min-height: 240px;
+        overflow: hidden;
+        background: var(--card);
+      }
```
richardhjtan: [Claude Code 🤖]

**`json` and `csv` are merged here, but they have opposite preview backgrounds — so this rule is right for CSV and wrong for JSON.** Regression introduced by this change; non-blocking (cosmetic only), but it undercuts the dark-inspector design decision the JSON preview makes deliberately.

**The mechanism.** The two previews commit to different surfaces on purpose:

- `json-file-def.gts` — `.data-preview { background: var(--boxel-dark, #1e1e1e); color: var(--boxel-light, #d4d4d4) }`, with a comment stating the intent: *"A data inspector stays a dark panel in either theme."*
- `csv-file-def.gts` — `.data-preview { background: var(--card); color: var(--foreground) }`, the theme-aware light table.

This rule gives both stages `background: var(--card)`, which matches CSV and contradicts JSON.

**Why the mismatch becomes visible rather than being covered up.** The rule also sets `height: auto; min-height: 240px`. With the stage's computed height `auto`, the `height: 100%` on the descendant `.stage` and `.data-preview` resolves against a content-dependent height and therefore behaves as `auto` — so the dark panel is only as tall as the tree. For any JSON document whose tree renders shorter than 240px, `min-height` holds the stage open past the panel and the leftover strip paints the stage's own `--card`: a light band directly beneath the dark tree, inside the rounded border.

That is not an exotic case — this branch's own acceptance fixture (`{ name, version, dependencies: { lodash } }`) renders well under 240px, and small config files are the common case for `.json`.

**The `code` kind already solved exactly this, and is the precedent to copy.** It has the same dark preview and the same `height: auto; min-height: 240px; overflow: hidden` shape, and it sets `background: var(--boxel-dark, #1e1e1e)` on the stage precisely so the floor matches the panel. JSON is in the identical position and should follow it, which means splitting this selector:

```css
      /* A data table reads as a full-width document, clipped at its rounded
         border rather than overflowing. */
      .iso[data-preview-kind='csv'] .iso-stage {
        height: auto;
        min-height: 240px;
        overflow: hidden;
        background: var(--card);
      }
      /* The JSON inspector is a dark panel in either theme, so its floor is
         dark too — a short document must not leave a light band beneath it. */
      .iso[data-preview-kind='json'] .iso-stage {
        height: auto;
        min-height: 240px;
        overflow: hidden;
        background: var(--boxel-dark, #1e1e1e);
      }
```

This one is derived from the cascade rather than observed, since I reviewed from the source rather than the running app — the isolated JSON view of a small file in the preview deployment should show it immediately either way.

---

packages/base/file-formats/file-shell-isolated.gts
```diff
@@ -695,6 +695,18 @@ export class FileIsolatedShell extends GlimmerComponent<FileIsolatedShellSignatu
...
+         it. */
+      .iso[data-preview-kind='json'] .iso-cols,
+      .iso[data-preview-kind='csv'] .iso-cols {
+        grid-template-columns: minmax(0, 1fr);
+        gap: 24px;
+      }
+      .iso[data-preview-kind='json'] .inspector,
+      .iso[data-preview-kind='csv'] .inspector {
+        width: 100%;
+      }
```
richardhjtan: [Claude Code 🤖]

**One layout decision, now spelled out in three separate rule blocks with three near-identical comments.** Drift-by-duplication; non-blocking, and cheaper to fold now than after a fifth family arrives.

**All the sites, enumerated.** Every one of these sets exactly `grid-template-columns: minmax(0, 1fr); gap: 24px` on `.iso-cols` plus `width: 100%` on `.inspector` — byte-identical declarations:

- `html`, `markdown`, `doc` — the original block, commented *"HTML and long-form text own a full-width document stage…"*
- `json`, `csv` — this block, commented *"A data table or tree reads as a full-width document too…"*
- `code` — the block immediately below, commented *"Source code reads as a full-width document too…"*

So seven preview kinds share one decision across three rules, and two of the three comments are reworded copies of the first. The cost is not the bytes: it is that the next family to want full width has a coin-flip choice of which block to extend, and that changing the decision — the `24px`, or adding an `align-items` — now means finding three places and noticing the third.

**The single home** is one selector list per property group, with the comment stated once:

```css
      /* A document is read top to bottom, so its metadata follows below in
         reading order rather than competing for width beside it. Long-form
         text, source code, and data tables and trees all read this way. */
      .iso[data-preview-kind='html'] .iso-cols,
      .iso[data-preview-kind='markdown'] .iso-cols,
      .iso[data-preview-kind='doc'] .iso-cols,
      .iso[data-preview-kind='code'] .iso-cols,
      .iso[data-preview-kind='json'] .iso-cols,
      .iso[data-preview-kind='csv'] .iso-cols {
        grid-template-columns: minmax(0, 1fr);
        gap: 24px;
      }
      .iso[data-preview-kind='html'] .inspector,
      .iso[data-preview-kind='markdown'] .inspector,
      .iso[data-preview-kind='doc'] .inspector,
      .iso[data-preview-kind='code'] .inspector,
      .iso[data-preview-kind='json'] .inspector,
      .iso[data-preview-kind='csv'] .inspector {
        width: 100%;
      }
```

Worth contrasting with the `.iso-stage` rules below, where the per-kind split is *justified* — those blocks genuinely differ in `overflow` and `background`, and the comments explain the differences rather than repeating one idea. That is the distinction to preserve: split where the declarations differ, merge where they don't.

Follow-up rather than a blocker — nothing misbehaves today.

---

packages/base/json-file-def.gts
```diff
@@ -66,399 +86,349 @@ function truncateExcerpt(text: string): string {
...
+  // The tree markup is produced here: every key and value is escaped before it
+  // is wrapped, so the resulting HTML is our own and needs no second sanitizer
+  // pass during prerender/indexing.
+  get treeHtml() {
+    let { ok, value } = this.parsed;
+    if (!ok) {
+      return htmlSafe('');
+    }
+    return htmlSafe(jsonToTreeHtml(value as JsonValue));
+  }
```
richardhjtan, line 247: [Claude Code 🤖]

**CSV budgets its embedded render by mode; JSON doesn't. The two families in this change diverge on the one budgeting decision they share.** Non-blocking, and a follow-up is a fine home for it — but worth deciding deliberately rather than by omission, since the sibling sitting in the same diff answers it the other way.

**The two shapes, side by side.** `CsvPreview.tableHtml`:

```js
let maxRows = this.args.mode === 'embedded' ? EMBEDDED_MAX_ROWS : undefined;
```

`JsonPreview.treeHtml` takes no mode into account at all — embedded and isolated both call `jsonToTreeHtml`, which renders until the global `TREE_MAX_NODES` cap of 2000.

**What that costs in the embedded shell.** `file-shell-embedded.gts` gives the stage `.emb-body { height: 220px }` for the document shape. So the embedded JSON pane is a fixed 220px scroller that can contain up to 2000 rendered nodes, where the CSV pane beside it contains at most 20 rows plus its "more rows" note. A `package-lock.json` or any nested config reaches the cap comfortably.

Collapsing does not recover it: `renderTreeNode` emits every descendant's markup regardless of the `open` attribute, and a closed `<details>` still holds its children in the DOM — it only hides them. So the node count is paid in full on render whatever the auto-open depth ends up being.

**The way out**, matching the sibling's shape — a mode-aware cap threaded into the existing `TreeCtx` budget, which already carries the truncation flag and note:

```js
const TREE_MAX_NODES = 2000;
const EMBEDDED_MAX_NODES = 200;

get treeHtml() {
  let { ok, value } = this.parsed;
  if (!ok) {
    return htmlSafe('');
  }
  let max =
    this.args.mode === 'embedded' ? EMBEDDED_MAX_NODES : TREE_MAX_NODES;
  return htmlSafe(jsonToTreeHtml(value as JsonValue, max));
}
```

The existing `… tree truncated at N nodes` note then does the same job for embedded that CSV's `… N more rows` note does, so the affordance stays consistent across the two families too.

---

packages/base/json-file-def.gts
```diff
@@ -489,14 +459,31 @@ export class JsonFileDef extends FileDef {
...
-
-  static isolated: BaseDefComponent = Isolated;
-  static embedded: BaseDefComponent = Embedded;
-  static fitted: BaseDefComponent = Fitted;
-  static atom: BaseDefComponent = Atom;
+  // Surfaced by the shells as the hero fact ("N keys") and by the fitted
+  // preview, computed once at index time so a cell never parses the document.
+  @field rootType = contains(StringField);
+  @field keyCount = contains(NumberField);
+  @field lineCount = contains(NumberField);
```
richardhjtan: [Claude Code 🤖]

**The parenthetical names the wrong fact: the shells surface "N lines", not "N keys".** Non-blocking comment-accuracy fix — but it points a reader at the wrong field when they go looking for where the hero fact comes from, and the description already states the correct behavior, so the comment is the only thing carrying the error.

**Verified in the code that draws it.** The hero fact is chosen in `file-formats/file-view-model.ts`, and the branch a JSON file takes is:

```js
if (['code', 'json', 'doc'].includes(kind) && model.lineCount != null) {
  return `${model.lineCount.toLocaleString()} lines`;
}
```

There is no branch anywhere in that function that reads `keyCount`. So of the three fields this comment sits above, `lineCount` is the one the shells present as the hero fact; `rootType` and `keyCount` are consumed only by `JsonPreview`'s fitted summary in this file. The description gets this exactly right — *"JSON's shell hero fact stays 'N lines'; the key count lives in the fitted preview summary"* — so this is just the comment lagging.

**The fix**, keeping the useful half (the index-time rationale) and splitting the two audiences:

```suggestion
  // Structural facts computed once at index time so a fitted cell never parses
  // the document. `rootType` and `keyCount` drive the fitted preview's summary;
  // `lineCount` is what the shells present as the hero fact.
  @field rootType = contains(StringField);
  @field keyCount = contains(NumberField);
  @field lineCount = contains(NumberField);
```

---

packages/base/json-file-def.gts
richardhjtan: [Claude Code 🤖]

**Extending this thread rather than opening a second one: invalid JSON reaches the same vacuous summary by a much likelier route than a scalar root, and the fix proposed here closes it only if `''` is treated as its own branch.**

`extractAttributes` catches the parse failure and stores the empty-root pair explicitly:

```js
try {
  let parsed = JSON.parse(text) as JsonValue;
  rootType = kindOf(parsed);
  keyCount = entryCount(parsed);
} catch {
  rootType = '';
  keyCount = 0;
}
```

So for a truncated or hand-edited `.json`, `rootType` is `''` and `keyCount` is `0`. In the fitted cell `countLabel` falls through its `array` check to the default and returns `keys`, the template renders `{{this.keyCount}}` as `0`, and `{{#if this.rootType}}` suppresses the type line — giving "0 · KEYS" with nothing to qualify it. Same wrong-and-vacuous output as the scalar case, but arriving via a corrupt file rather than an unusual-but-valid one, which makes it the more likely of the two to be seen.

Worth noting the non-fitted formats already handle this state well — `isValid` is false, so embedded and isolated fall back to the raw `<pre>`. It is only the fitted summary that has no representation for "couldn't parse".

So when branching `countLabel` on `rootType`, `''` wants to be distinct from both the scalar and the object cases — something like `object` → key(s), `array` → item(s), `''` → an "invalid" or "unparsed" label with the count suppressed, anything else → a `value` label. That covers corrupt files and scalar roots in the same change, and the `{{#if this.rootType}}` guard already in the template means the type line needs no further work.

---

packages/base/json-file-def.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b`. The `open` gate is now inclusive (`depth <= TREE_AUTO_OPEN_DEPTH`), so the root and its direct children render expanded and deeper levels start collapsed, matching the constant's comment. The integration test now pins both sides: it asserts an `[open]` node nested inside the root's `[open]` node, and a depth-2 container without `open` (the fixture gained a nested `engines` object for exactly that).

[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` — the gate is inclusive now (`depth <= TREE_AUTO_OPEN_DEPTH`), so the root and its direct children open as documented, and the integration test pins the expanded/collapsed depths.

packages/base/json-file-def.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b`, folding in the invalid-JSON branch from the follow-up below. The fitted summary now branches on `rootType`: `object` counts keys, `array` counts items, a scalar root shows its type as the fact (`value` · `number` — the hero slot carries the type rather than a vacuous `0`), and an empty `rootType` (the unparsed marker `extractAttributes` stores) renders an "Invalid JSON" label with the count suppressed. The template reads a derived `countValue` instead of `keyCount` directly. New integration tests cover the scalar root and the fitted invalid case.

packages/base/json-file-def.gts
```diff
@@ -66,399 +86,349 @@ function truncateExcerpt(text: string): string {
...
-      return htmlSafe('');
+  get keyCount(): number {
+    return Number(this.source?.keyCount ?? 0);
+  }
+
+  get countLabel(): string {
+    let type = this.rootType;
+    if (type === 'array') {
+      return this.keyCount === 1 ? 'item' : 'items';
     }
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` — `countLabel` now branches on `rootType` (object → keys, array → items, scalar → a `value` label showing the type, unparsed → "Invalid JSON").

packages/base/json-file-def.gts
```diff
@@ -66,399 +86,349 @@ function truncateExcerpt(text: string): string {
...
-      <div class='json-embedded__content'>
-        <pre class='json-embedded__pre'>{{this.highlightedContent}}</pre>
+    {{#if this.isFitted}}
+      <div class='data-preview data-preview--fitted' data-test-json-preview>
+        {{#if this.hasContent}}
+          <dl class='data-summary'>
+            <div class='data-summary__metric'>
+              <dt>{{this.countLabel}}</dt>
+              <dd>{{this.keyCount}}</dd>
+            </div>
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` — the template renders a derived `countValue` (the entry count for containers, the type for scalar roots) instead of `keyCount` directly.

packages/base/csv-file-def.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` by aligning the words to the code: both comments now say the embedded table budgets to the head of the file with a "… N more rows" note, and the PR description is updated to match. The explicit note stays as the affordance; no mask was reintroduced.

packages/base/csv-file-def.gts
```diff
@@ -166,349 +171,182 @@ function csvToHtml(content: string, maxRows?: number): string {
...
+        color: var(--foreground);
+        text-align: left;
       }
-
-      .csv-embedded__title {
-        color: var(--boxel-900);
-        font-weight: 600;
+      .data-preview--table {
+        padding: var(--boxel-sp);
       }
```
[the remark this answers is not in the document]
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` — the comments and PR description now describe the "… N more rows" note the code actually renders; the fade wording is gone.

packages/base/file-formats/file-shell-isolated.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b`, taking the split exactly as proposed: CSV keeps `background: var(--card)` and JSON's stage floor is `var(--boxel-dark, #1e1e1e)`, so a short document sits on a dark floor matching its inspector panel — same shape as the `code` kind's stage rule.

packages/base/file-formats/file-shell-isolated.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` — the three blocks are folded into one selector list per property group (`.iso-cols` and `.inspector`), with the reading-order rationale stated once. The per-kind `.iso-stage` rules stay split since those genuinely differ in `overflow` and `background`.

packages/base/json-file-def.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b`, matching the sibling's shape: `treeHtml` threads a mode-aware cap into the existing `TreeCtx` budget (`EMBEDDED_MAX_NODES = 200`, isolated keeps `TREE_MAX_NODES = 2000`), and the truncation note reports the active cap. A new integration test renders a 300-node document and asserts the embedded pane truncates at 200 while isolated renders it in full.

packages/base/json-file-def.gts
lukemelia: [Claude Code 🤖] Fixed in `30209c261b` with the suggested wording: the comment now says `rootType`/`keyCount` drive the fitted preview's summary and `lineCount` is what the shells present as the hero fact.

PR #5761, opened 2026-08-12
Graceful generic fallback format for files with no dedicated FileDef

lukemelia:
## What

A file whose extension has no dedicated `FileDef` subclass (a plain binary, an unknown type) renders through the base `FileDef` and the four shared format shells. The shells already carry its identity — icon, name, extension, kind, size — but the preview *stage* fell back to a bare "No preview" glyph in every format.

This makes that fallback graceful in the reading formats:

- **Embedded / isolated**: the pane now presents what the realm knows — the file's name, kind, and size.
- **Embedded** additionally gets a **Download** affordance. It's the only reading format whose shell offers no download of its own; the isolated shell already exposes Download + Copy link in its header, so its pane defers to that rather than showing a second button.
- **Fitted** keeps its bare glyph — the budgeted cell already has the metadata strip beside it, so the pane shouldn't repeat the name and kind.

Net result across the four formats for a generic/unknown file:

| Format | Icon | Filename | Type/kind | Size | Download |
|---|---|---|---|---|---|
| atom | ✓ | ✓ | — | — | — |
| fitted | ✓ | ✓ | ✓ | ✓ | — |
| embedded | ✓ | ✓ | ✓ | ✓ | ✓ |
| isolated | ✓ | ✓ | ✓ | ✓ | ✓ (header) |

## Tests

- New `generic-file-def-preview-test.gts` pins the fallback across all four formats for an `application/octet-stream` binary and an unregistered extension.
- Updated the existing format-templates test to assert the graceful embedded pane and the minimal fitted pane.

Note: the full host test stack wasn't available locally to run these to green; relying on CI.

### DIFF
--- a/packages/base/file-formats/file-preview-stage.gts
+++ b/packages/base/file-formats/file-preview-stage.gts
@@ -13,7 +13,7 @@ import { eq } from '@cardstack/boxel-ui/helpers';
 // dependency graph, so a glyph taken from it costs nothing extra.
 import { Warning } from '@cardstack/boxel-ui/icons';
 
-import { fileIconFor, humanSize } from './file-presentation';
+import { downloadNameFor, fileIconFor, humanSize } from './file-presentation';
 import type { FileFormat, FileViewModel } from './file-view-model';
 
 import type { ComponentLike } from '@glint/template';
@@ -96,10 +96,6 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
     return humanSize(this.model?.contentSize);
   }
 
-  get downloadName() {
-    return this.model?.name || undefined;
-  }
-
   // The reading formats have room for a labeled fallback card; the budgeted
   // fitted cell stays a bare glyph so it never competes with the metadata strip
   // the fitted shell draws beside it.
@@ -152,11 +148,7 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
         {{#if @preview}}
           <@preview @model={{@model}} @mode={{@mode}} @fields={{@fields}} />
         {{else}}
-          <div
-            class='generic-pane'
-            data-mode={{@mode}}
-            data-test-file-no-preview
-          >
+          <div class='generic-pane' data-test-file-no-preview>
             <this.icon
               class='gp-icon'
               width={{this.genericIconSize}}
@@ -166,7 +158,7 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
             {{#if this.showGenericDetail}}
               <div
                 class='gp-name'
-                title={{@model.name}}
+                title={{this.displayName}}
                 data-test-file-generic-name
               >{{this.displayName}}</div>
               <div class='gp-facts'>
@@ -181,7 +173,7 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
                 <a
                   class='gp-download'
                   href={{@model.url}}
-                  download={{this.downloadName}}
+                  download={{downloadNameFor @model}}
                   data-test-file-generic-download
                 >Download</a>
               {{/if}}
--- a/packages/base/file-formats/file-preview-stage.gts
+++ b/packages/base/file-formats/file-preview-stage.gts
@@ -13,7 +13,7 @@ import { eq } from '@cardstack/boxel-ui/helpers';
 // dependency graph, so a glyph taken from it costs nothing extra.
 import { Warning } from '@cardstack/boxel-ui/icons';
 
-import { fileIconFor } from './file-presentation';
+import { fileIconFor, humanSize } from './file-presentation';
 import type { FileFormat, FileViewModel } from './file-view-model';
 
 import type { ComponentLike } from '@glint/template';
@@ -84,6 +84,40 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
     return fileIconFor(this.model);
   }
 
+  // The generic pane stands in for a file whose family has no renderer — most
+  // often a plain binary. Rather than an empty "No preview" void, the reading
+  // formats present what the realm does know: name, kind, and size, plus a way
+  // to get the bytes.
+  get displayName() {
+    return this.model?.name || this.model?.baseName || 'Untitled file';
+  }
+
+  get size() {
+    return humanSize(this.model?.contentSize);
+  }
+
+  get downloadName() {
+    return this.model?.name || undefined;
+  }
+
+  // The reading formats have room for a labeled fallback card; the budgeted
+  // fitted cell stays a bare glyph so it never competes with the metadata strip
+  // the fitted shell draws beside it.
+  get showGenericDetail() {
+    return this.args.mode === 'embedded' || this.args.mode === 'isolated';
+  }
+
+  // Embedded is the only reading format whose shell offers no download of its
+  // own, so the fallback pane carries the affordance there. The isolated shell
+  // already exposes Download and Copy link in its header.
+  get showGenericDownload() {
+    return this.args.mode === 'embedded' && Boolean(this.model?.url);
+  }
+
+  get genericIconSize() {
+    return this.showGenericDetail ? '44' : '30';
+  }
+
   get srcTag() {
     return (this.model?.previewSource ?? '').toUpperCase();
   }
@@ -118,9 +152,42 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
         {{#if @preview}}
           <@preview @model={{@model}} @mode={{@mode}} @fields={{@fields}} />
         {{else}}
-          <div class='generic-pane' data-test-file-no-preview>
-            <this.icon width='30' height='30' aria-hidden='true' />
-            <span class='pane-label'>No preview</span>
+          <div
+            class='generic-pane'
+            data-mode={{@mode}}
+            data-test-file-no-preview
+          >
+            <this.icon
+              class='gp-icon'
+              width={{this.genericIconSize}}
+              height={{this.genericIconSize}}
+              aria-hidden='true'
+            />
+            {{#if this.showGenericDetail}}
+              <div
+                class='gp-name'
+                title={{@model.name}}
+                data-test-file-generic-name
+              >{{this.displayName}}</div>
+              <div class='gp-facts'>
+                {{#if @model.kind}}<span
+                    data-test-file-generic-kind
+                  >{{@model.kind}}</span>{{/if}}
+                {{#if this.size}}<span
+                    data-test-file-generic-size
+                  >{{this.size}}</span>{{/if}}
+              </div>
+              {{#if this.showGenericDownload}}
+                <a
+                  class='gp-download'
+                  href={{@model.url}}
+                  download={{this.downloadName}}
+                  data-test-file-generic-download
+                >Download</a>
+              {{/if}}
+            {{else}}
+              <span class='pane-label'>No preview</span>
+            {{/if}}
           </div>
         {{/if}}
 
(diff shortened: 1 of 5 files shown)


packages/base/file-formats/file-preview-stage.gts
```diff
@@ -118,9 +152,42 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
         {{#if @preview}}
           <@preview @model={{@model}} @mode={{@mode}} @fields={{@fields}} />
         {{else}}
-          <div class='generic-pane' data-test-file-no-preview>
-            <this.icon width='30' height='30' aria-hidden='true' />
-            <span class='pane-label'>No preview</span>
+          <div
+            class='generic-pane'
+            data-mode={{@mode}}
```
richardhjtan: [Claude Code 🤖]

**This `data-mode` is dead, and it duplicates an attribute the parent already carries.** Regression introduced here, non-blocking — but it reads as a styling hook, so it will cost the next editor a search.

**Evidence.** `generic-pane` appears exactly twice in this repo, both in this file: this element, and the `.generic-pane, .state-pane { … }` rule in the `<style scoped>` block below. No selector anywhere qualifies it with `[data-mode]`, and because the block is scoped, no ancestor stylesheet can reach in to do so either. The new tests select on `[data-test-file-no-preview]` only, so nothing reads it as a test hook.

Meanwhile `.stage` — this element's direct parent, six lines up — already has `data-mode={{@mode}}`. So the mode is already in the DOM at a level from which `.stage[data-mode='embedded'] .generic-pane` would work today.

**The way out**, whichever matches the intent:

- if the per-mode difference is meant to be CSS, drop this attribute and hang the rules off the parent's existing one;
- if it isn't, drop it — the two mode-dependent decisions here (`showGenericDetail`, `genericIconSize`) are already made in JS.

```suggestion
          <div class='generic-pane' data-test-file-no-preview>
```

(Worth noting the same shape shows up in the sibling data-formats change, where a `data-mode` on the CSV preview wrapper is likewise set and unread. Consistent habit rather than a one-off, so it may be worth deciding once whether these wrappers get a mode hook by convention.)

---

packages/base/file-formats/file-preview-stage.gts
```diff
@@ -84,6 +84,40 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
...
+    return this.model?.name || this.model?.baseName || 'Untitled file';
+  }
+
+  get size() {
+    return humanSize(this.model?.contentSize);
+  }
+
+  get downloadName() {
+    return this.model?.name || undefined;
+  }
```
richardhjtan: [Claude Code 🤖]

**This getter is now the second verbatim copy of the same decision.** Drift-by-duplication; non-blocking, and cheap to close now while both copies are still identical.

**Both sites, enumerated.** `file-shell-isolated.gts` has, character for character:

```js
get downloadName() {
  return this.args.model?.name || undefined;
}
```

and pairs it with an anchor of the same shape as the new one here — `href={{@model.url}}`, `download={{this.downloadName}}`, differing only in CSS class and test attribute. So the "what filename does a download get, and what do we do when there isn't one" decision now lives in two components, and the `|| undefined` detail — which is load-bearing, since `download=""` and an absent `download` are not the same to a browser — has to stay in sync by hand.

**The way out.** `file-presentation.gts` is already the home for exactly this layer: `humanSize` and `fileIconFor` are both imported from it into this file, `fileIconFor` by both this stage and the isolated shell. A `downloadNameFor(model)` beside them, imported at both sites, puts the decision in one place at the cost of one import.

If the duplication grows past the getter — a second consumer wanting the anchor's `href`/`download`/label together — the shared thing to reach for is a small `FileDownloadLink` component in the same directory, leaving each shell to supply only styling. Not needed at two sites; worth naming as the boundary if a third appears.

Follow-up rather than a blocker: nothing is broken today, precisely because the copies still agree.

---

packages/base/file-formats/file-preview-stage.gts
```diff
@@ -118,9 +152,42 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
...
+            <this.icon
+              class='gp-icon'
+              width={{this.genericIconSize}}
+              height={{this.genericIconSize}}
+              aria-hidden='true'
+            />
+            {{#if this.showGenericDetail}}
+              <div
+                class='gp-name'
+                title={{@model.name}}
```
richardhjtan: [Claude Code 🤖]

**Agreed with the bot that these should be one value — and the reachable case is narrower and more specific than "name missing".** Non-blocking. Recording the reachability because it decides whether this is a bug fix or a consistency fix, and it is the latter.

**What the model guarantees.** `FileViewModel.name` is a required `string`, not optional, and `buildFileViewModel` derives it as:

```js
let url = String(file.url ?? file.sourceUrl ?? file.id ?? '');
let name = String(file.name ?? url.split('/').pop() ?? '');
```

So `name` is never `undefined` — the divergence needs `name === ''`, and reaching that needs the file to have no `name` **and** a URL whose last path segment is empty. A file with no URL at all cannot get here: with `contentSize` unset and neither `url` nor `content`, `fileState` derives to `'empty'`, and `showReal` excludes `'empty'`, so the empty pane renders instead of this one. (That is also why the no-URL fallback test dropped from this branch had nothing to assert against.)

What's left is a file with a trailing-slash URL and no name — so `title=""` alongside a rendered `Untitled file`. Unusual enough that I would not call it a live defect, and `title=""` degrades to "no tooltip" rather than to anything broken.

**Still worth changing**, because the tooltip's whole job is to reveal the full text when `.gp-name`'s `text-overflow: ellipsis` truncates it — which only works if it *is* the rendered text. Binding both to the same getter makes that true by construction rather than by coincidence:

```suggestion
                title={{this.displayName}}
```

---

richardhjtan (commented):
[Claude Code 🤖]

**Lens:** the value of a fallback pane is entirely in its edge cases — it exists for files the system knows least about — so this review chased what the pane renders when each fact it presents is missing, and whether the per-format gating claimed in the description matches what the shells actually pass.

**Bottom line: no blocking issues. The gating is exactly right and the missing-fact behavior is sound — I could not find a state where the pane renders something wrong or empty.** Three non-blocking items: one dead attribute this change introduces, one getter that is now duplicated verbatim across two files, and a precise reachability answer for the bot's tooltip note.

**What lands right (verified, not courtesy).**

- **The per-format gating is correct at every site.** Every shell passes a literal mode matching its own name — `@mode='embedded'`, `@mode='fitted'` (twice, for the preview and waveform models), `@mode='isolated'` — so `showGenericDetail` and `showGenericDownload` can never be misled by a mismatched mode. `file-shell-atom.gts` never mounts the stage at all, which is what makes the atom row of the description's table correct: that row is the atom shell's own identity line, not this pane.
- **"Embedded is the only reading format whose shell offers no download of its own" is exactly true, not approximately.** `download=` occurs at precisely two places in `packages/base/file-formats/` — the isolated shell's header and this new pane. The embedded shell has no anchor of any kind. The reasoning for putting the affordance here and not in isolated holds.
- **Missing facts are omitted rather than rendered as zeros.** `humanSize` returns `''` for `null`/`undefined`, so `{{#if this.size}}` correctly drops the size fact instead of printing a placeholder. The one truthy-but-vacuous case, `contentSize === 0` returning `'0 bytes'`, cannot reach this pane: `buildFileViewModel` maps `contentSize === 0` to `fileState: 'empty'`, and `showReal` excludes `'empty'`, so a zero-byte file gets the dedicated empty pane instead.
- **The fitted decision is self-verifying.** The reasoning for keeping fitted a bare glyph is that the fitted shell already draws name, kind and size beside it — and the new test asserts exactly those three in the fitted cell before asserting the pane stays minimal. The test pins the premise, not just the conclusion.

**Recommendations.**

1. **`data-mode` on the pane is a hook nothing reads, and the parent already has one.** Regression, non-blocking. Inline thread.
2. **`downloadName` is now defined identically in two files.** Drift-by-duplication; `file-presentation.gts` is the established home. Inline thread.
3. **Use `this.displayName` for the `title` too.** The bot's note is directionally right; the reachable case is narrower than it says, and I've worked out exactly what it is. Inline thread.

**Adjacent, out of scope.** The new anchor copies the isolated header's `href={{@model.url}}` + `download=` shape, which means it inherits whatever that pattern's behavior is for a realm file on a different origin than the host — the HTML `download` attribute is same-origin-only, and a cross-origin URL makes the browser navigate rather than save. That is pre-existing and shared with the header link, so it is not this PR's to fix, but it is now a second call site depending on it, which makes it worth confirming once for both. Not asked of this change.

---

packages/base/file-formats/file-preview-stage.gts
lukemelia: [Claude Code 🤖] Done in 9dbd439 — dropped the `data-mode` from the pane. The mode-dependent decisions here are all made in JS, and the parent `.stage` already carries `data-mode` if a CSS hook is ever wanted, so no rules were rehomed.

packages/base/file-formats/file-preview-stage.gts
lukemelia: [Claude Code 🤖] Done in 9dbd439 — extracted `downloadNameFor(model)` into `file-presentation.gts` (beside `humanSize`/`fileIconFor`), removed the duplicated getter from both this stage and `file-shell-isolated.gts`, and both anchors now call the helper directly (`download={{downloadNameFor @model}}`). The load-bearing `|| undefined` lives in one place now.

packages/base/file-formats/file-preview-stage.gts
lukemelia: [Claude Code 🤖] Done in 9dbd439 — bound the tooltip to `{{this.displayName}}`, the same value the element renders, so it reveals the displayed text when `.gp-name` ellipsizes. Agreed it was a consistency fix rather than a live defect.

backspace (approved):

packages/base/file-formats/file-preview-stage.gts
```diff
@@ -119,8 +149,37 @@ export class FilePreviewStage extends GlimmerComponent<StageSignature> {
...
+                data-test-file-generic-name
+              >{{this.displayName}}</div>
+              <div class='gp-facts'>
+                {{#if @model.kind}}<span
+                    data-test-file-generic-kind
+                  >{{@model.kind}}</span>{{/if}}
+                {{#if this.size}}<span
+                    data-test-file-generic-size
+                  >{{this.size}}</span>{{/if}}
+              </div>
```
habdelra, line 171: [Claude Code 🤖]

**The embedded pane repeats the name and size its own shell header already shows.** Non-blocking design note.

In embedded mode this block renders name (`gp-name`) + kind + size (`gp-facts`), while `file-shell-embedded.gts`'s `.emb-head` already draws the icon, the name (`.emb-name`), the size (`.emb-meta`), and the extension pill directly above the stage. So for a generic file the embedded format shows the filename in both places and the size verbatim in both — `256.0 KB` in the header's `.emb-meta` and again here in `gp-facts`. The pane's genuinely new content is `kind`, which the header omits, and the Download affordance.

This is the same repetition the fitted pane is kept minimal to avoid: the reasoning there is that the fitted shell already has the metadata strip beside it, so the pane shouldn't repeat name and kind. The embedded header *is* that strip for embedded, yet the pane repeats it.

I don't think it's wrong — a 220px `.emb-body` filled with a lone glyph would read as empty, and a labeled identity card is a better fill. But the asymmetry with the fitted rationale is worth a deliberate call rather than leaving it implicit: either keep it (the body needs a meaningful fill) or trim the embedded pane to the facts the header lacks — `kind` + Download — since name and size are already one row up.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** a re-review of the three changes made since the last pass — the shared `downloadNameFor`, the tooltip binding, and the dropped `data-mode` — verified as changes against the current tree rather than as described, plus a fresh look at how the graceful embedded pane sits next to the shell that frames it.

**Bottom line: no blocking issues. All three prior findings are correctly resolved — I traced each against the head tree, not the commit message — and the download/size/kind behavior is sound. One non-blocking design observation: the embedded pane repeats the name and size the embedded header already shows, which is the exact repetition the fitted pane is kept minimal to avoid.**

**The three fixes, verified.**

- **`downloadNameFor` is now the single home for the download-name rule.** `download=` appears at exactly two sites in `packages/base/file-formats/` — the isolated header (`file-shell-isolated.gts`) and the generic pane (`file-preview-stage.gts`) — and both now read `download={{downloadNameFor @model}}`. The getter is gone from both components, and the load-bearing `|| undefined` (an absent `download` attribute vs. `download=""`) lives once in `file-presentation.gts`, beside `humanSize`/`fileIconFor`. Resolves the duplication.
- **The tooltip and the rendered name are one value.** `title={{this.displayName}}` now matches the `{{this.displayName}}` the `.gp-name` div renders, so the tooltip reveals the displayed text when `text-overflow: ellipsis` truncates it. Resolves the divergence.
- **The dead `data-mode` is gone.** The pane is now `<div class='generic-pane' data-test-file-no-preview>`; the parent `.stage` still carries `data-mode={{@mode}}`, so a future CSS hook has a home without the unread attribute. Resolves it.

**What else I checked (holds).**

- **Gating is correct at every mount site.** `@mode` is a literal matching each shell's name — embedded, fitted (twice: preview + waveform), isolated — and atom never mounts the stage, so `showGenericDetail`/`showGenericDownload` cannot be misled by a mismatched mode.
- **Missing facts are omitted, not zeroed.** `humanSize` returns `''` for `null`/`undefined`, so `{{#if this.size}}` drops the size fact; a zero-byte file routes to `fileState: 'empty'` (excluded from `showReal`) and never reaches this pane.
- **The download button matches the isolated header's primary button** — same `--fd-slate`/`--fd-paper` tokens, radius, and weight (padding differs by 2px), so the two affordances read as one control.

**Recommendation (non-blocking).**

1. The embedded pane repeats the name and size the embedded header already draws — see the inline note on the detail block. `kind` and the Download button are the pane's genuinely new content; name and size are shown twice. Reads as an intentional fill of the 220px body rather than a lone glyph, but it is the same repetition the fitted pane is deliberately kept minimal to avoid, so it is worth making a decision on record.

**Adjacent, out of scope.** The pane's `href={{@model.url}}` + `download` inherits the HTML `download` attribute's same-origin-only behavior — a realm file served from a different origin than the host navigates rather than saves. Pre-existing and shared with the isolated header; this change just adds a second call site depending on it. Not this PR's to fix.

---

habdelra (approved):

PR #5762, opened 2026-08-12
Render ZIP archives through the shared FileDef shells

lukemelia:
Adds a `ZipDef` FileDef subclass so `.zip` files get high-quality atom / fitted / embedded / isolated formats, composed from the shared shells (CS-12229) rather than bespoke templates. Part of the High-Quality FileDef Formats project (CS-12244).

## What it does

- **Reads the archive without inflating it.** `packages/base/zip-archive.ts` is a dependency-free reader that walks a ZIP's central directory from a **bounded tail window** — a ZIP's table of contents lives at the end of the file — so a large archive is listed without decompressing an entry or holding the whole file in memory. Handles ZIP64, drops explicit directory markers, decodes DOS timestamps, and reports honest truncation when a directory outruns the window.
- **Draws the contents.** `ArchivePreview` (the family `previewComponent`) renders an indented folder tree in the embedded/isolated formats — folders recovered from entry paths, row count budgeted — and a compact "N entries · total size" summary with a few leading names in the budgeted fitted cell.
- **Keeps every fact a real field.** Entries are a `containsMany(ArchiveEntryField)` with per-entry path/size/timestamp, so the isolated inspector lists them and the default editor can correct them. `extractAttributes` populates the listing, totals, and a truncation flag.
- **Routes `.zip` → `ZipDef`** in `file-def-code-ref.ts`.

## Files

- `packages/base/zip-archive.ts` — the central-directory reader (new)
- `packages/base/zip-file-def.gts` — `ZipDef`, `ArchiveEntryField`, `ArchivePreview` (new)
- `packages/runtime-common/file-def-code-ref.ts` — `.zip` routing
- `packages/host/tests/unit/zip-archive-test.ts` — parser unit tests (new)
- `packages/host/tests/integration/components/zip-file-def-test.gts` — rendering tests across the four shells (new)

## Test plan

- Parser: unit tests assemble real stored ZIPs (local headers + central directory + EOCD) and assert paths, per-entry and total sizes, directory-marker dropping, DOS-date decoding, non-ZIP → `undefined`, chunked-stream reads, and tail bounding. Verified green running the logic standalone (11/11).
- Rendering: integration tests mirror the shipped `ts-file-def-preview-test` pattern — renderer mounts into all four shells, folder tree in the reading formats, compact summary in fitted, entries listed in the isolated inspector, empty-archive and non-`.zip`-rejection paths.
- Typecheck (`ember-tsc`) and template-lint pass clean.

> Note: the rendering integration tests require the base realm + prerender servers; they'll run under CI's full stack. The parser tests are self-contained.


packages/base/zip-archive.ts
```diff
@@ -0,0 +1,380 @@
...
+  let low = view.getUint32(offset, true);
+  let high = view.getUint32(offset + 4, true);
+  return high * 0x1_0000_0000 + low;
+}
+
+function decodePath(bytes: Uint8Array, _isUtf8: boolean): string {
+  // A non-fatal UTF-8 decode handles the common (and flagged-UTF-8) case and
+  // degrades legacy CP437 names to replacement characters rather than throwing.
+  return new TextDecoder('utf-8').decode(bytes);
+}
```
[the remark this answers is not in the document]
lukemelia: Good catch — hoisted a single reused `PATH_DECODER` at module scope so the central directory no longer allocates a `TextDecoder` per entry. Fixed in 206a6cb5c8.

packages/base/zip-file-def.gts
```diff
@@ -0,0 +1,531 @@
...
+      );
+    }
+
+    let base = await super.extractAttributes(url, getStream, options);
+    let listing = await extractZipListing(await getStream());
+    if (!listing) {
+      // A `.zip` that carries no readable central directory (empty or corrupt)
+      // still gets its identity; it simply lists nothing.
+      return base;
+    }
```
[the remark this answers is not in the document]
lukemelia: Agreed. `extractAttributes` now throws `FileContentMismatchError` when `extractZipListing` returns undefined (no readable central directory), so a mis-extensioned or corrupt `.zip` falls back to the plain FileDef instead of being mislabeled as an empty archive — matching the PNG/MP4 extractors. A valid *empty* archive still carries an EOCD and reads as an empty listing, so it's unaffected; added a test for a `.zip` whose bytes are not a ZIP. Fixed in 206a6cb5c8.

packages/host/tests/unit/zip-archive-test.ts
habdelra, line 1: [Claude Code 🤖] **Two prominent parser claims land in untested code — both work, but nothing pins them.** Non-blocking, but this is the highest-value follow-up in the PR: the two most intricate branches of the reader are exactly the ones a future refactor is most likely to break silently, and neither has a test.

**ZIP64.** `readZip64Directory` and `readZip64ExtraSizes` in `zip-archive.ts` are ~60 lines of raw offset arithmetic — the EOCD64 locator at `eocd − 20`, the record's 64-bit fields at `+32`/`+48`, and the per-entry `0x0001` extra block walked as `(id, size, payload)` with the uncompressed size preceding the compressed one, each present only when its 32-bit field was the sentinel. A single wrong offset here reads garbage sizes, and no test would catch it.

**Truncation.** The `directoryStartInTail < 0 → scanForCentralDirectory` fallback and the `truncated: true` flag are the whole point of the bounded-tail design, but every fixture here is small enough that the directory always sits in the window, so `truncated` is only ever asserted `false` (the two `assert.false(...truncated)` / `notOk(truncatedListing)` cases).

**Both verified working.** I ported the pure functions and exercised them directly: a ZIP64 archive with a 5 GB (`0xffffffff`-sentinel) entry read back `uncompressedSize === 5000000000` and the right total; a windowed tail (`parseZipListing(fullZip.subarray(fullZip.length - N), fullZip.length)` with `N` cutting into the central directory) returned a partial listing with `truncated === true`. So these are coverage gaps, not defects.

**Cheap to pin — no realm/prerender stack needed** (these are the self-contained parser tests). Truncation needs no >4 MB fixture at all: pass a subarray of a normal `buildZip(...)` output as `tail` and the full length as `totalLength`, and assert `truncated === true` with a shorter `entries` list. ZIP64 needs `buildZip` extended to stamp the EOCD sentinels + an EOCD64 record + locator and a per-entry `0x0001` extra; happy to hand over the builder I used.

---

packages/base/zip-file-def.gts
```diff
@@ -0,0 +1,537 @@
...
+  // The last path segment, for a listing that shows names rather than full
+  // paths. A trailing slash (a directory marker) resolves to its own last
+  // segment rather than an empty string.
+  @field name = contains(StringField, {
+    computeVia: function (this: ArchiveEntryField) {
+      let path = this.path ?? '';
+      let segments = path.split('/').filter(Boolean);
+      return segments[segments.length - 1] ?? path;
+    },
+  });
```
habdelra, line 73: [Claude Code 🤖] **This `name` field is an orphan, and the last-segment logic it holds is duplicated twice more.** Non-blocking cleanup.

Nothing reads `ArchiveEntryField.name`. The two reading-format listings both recompute the basename inline instead:

- `ArchivePreview.fittedNames` — `String(path).split('/').filter(Boolean)` then `segments[segments.length - 1] ?? String(path)`
- `ArchivePreview.tree` — `path.split('/').filter(Boolean)` then `segments.pop() ?? path`

Both are the same "last path segment" decision this field already encodes, so it lives in three places. The isolated inspector renders each entry through `ArchiveEntryField.embedded`, which shows `@model.path` (the full path), and the `atom` format likewise — so the field's stated purpose ("a listing that shows names rather than full paths") isn't realized anywhere.

Two ways out: drop the field, or make it the single home — but note the duplication can't be fully collapsed onto it, because `fittedNames` and `tree` operate on projected `string` paths / plain objects (`model.source.archiveContents`), not `ArchiveEntryField` instances, so they can't call the field's getter. The practical fix is a small module-scope `basename(path)` helper that all three call. Your call whether that's worth it for a two-line function; flagging it because a copy that drifts here would show three different names for the same entry across formats.

---

habdelra (commented):
[Claude Code 🤖] This review focused on the ZIP central-directory reader's correctness across its full input range — ZIP64, tail-window truncation, and EOCD disambiguation — and on how the archive family composes into the shared FileDef shells, since the parser is the load-bearing new machinery and the rendering is mostly composition over existing infrastructure.

**Bottom line: no blocking issues.** The parser is correct on every path I could exercise. I ported the pure functions from `zip-archive.ts` and ran them directly: a ZIP64 archive with a 5 GB entry (32-bit sizes sentineled to `0xffffffff`, real sizes in the `0x0001` extra) read back exact 64-bit sizes and totals; a comment-bearing ZIP64 archive found the right EOCD; and a deliberately windowed tail produced a partial listing with `truncated === true`. The findings below are test-coverage gaps and minor cleanups, none of them gating.

**What lands right**
- The bounded-tail contract is genuinely enforced end to end: `readTailBytes` drops whole leading chunks while ≥ `n` bytes remain, `parseZipListing` maps absolute directory offsets back into the window via `tailStart`, and nothing inflates an entry — sizes and paths come straight from the central-directory header. The "honest truncation rather than unbounded read" promise holds.
- `extractAttributes` matches the sibling extractor idiom exactly — `super.extractAttributes(...)` then a fresh `getStream()` for the format-specific read (same shape as the MP4 extractor), which is safe because `getStream` is a fresh-stream factory. The resolved review thread's fix is correct: a valid *empty* archive keeps its EOCD and reads as an empty listing, while non-ZIP bytes now throw `FileContentMismatchError` and fall back to the base FileDef, consistent with the PNG/MP4 extractors.
- The archive family was already scaffolded across the shared infrastructure — the `application/zip` → `archive` profile, `FITTED_ARCHIVE_ENTRY_BUDGET`, the `archiveEntries`/`archiveEntryCount` projection, the `heroFactFor` archive case, and the isolated shell's "Archive contents" group. `ZipDef` slots into all of it by declaring `archiveContents` as a real `containsMany` field and pinning the profile axes as statics; it touches none of the shells. Clean composition.

**Recommendations**
1. Add ZIP64 and truncation unit tests — the two most fragile parser branches are currently unpinned. Detail and a cheap recipe in the file-level comment on `zip-archive-test.ts`. (Non-blocking, high value.)
2. `ArchiveEntryField.name` is unused and its last-segment logic is duplicated in `fittedNames` and `tree` — consolidate onto a small `basename()` helper or drop the field. See the inline comment. (Non-blocking.)

**Minor / adjacent — not asked of this PR**
- `decodePath` ignores its `_isUtf8` argument and always decodes UTF-8, so the `flags & UTF8_NAME_FLAG` read at the call site is dead. This is a deliberate, documented tradeoff (no CP437 table; legacy non-UTF-8 names degrade to replacement characters), but the parameter and the flag read could be dropped for clarity since they never influence behavior.
- In the tree footer, when the `TREE_ROW_BUDGET` overflow message ("+N more … not shown") fires *and* the listing was also parser-truncated, only the "+N more" line renders (the `{{else if this.isTruncated}}` branch is unreachable in that case), and its count is a lower bound because it's derived from the already-truncated `archiveContents`. Cosmetic.
- `findEndOfCentralDirectory` scans the entire retained tail backward when no length-consistent EOCD exists — i.e. for a non-ZIP file that carries a `.zip` extension — up to ~4M `getUint32` reads before returning `undefined`. Bounded and one-time per bad file, so fine; noting it for whoever profiles the malformed-input path.

---

habdelra (approved):

packages/host/tests/unit/zip-archive-test.ts
lukemelia: Done in 4350199b98 — pinned both branches. Added a ZIP64 test: a single entry whose 32-bit sizes are the 0xffffffff sentinel, its real 5 GB size read through the EOCD64 locator → record walk and the 0x0001 extra block; it reads back `uncompressedSize === 5_000_000_000`. And a truncation test that passes a subarray starting inside the central directory as the tail (with the full length), asserting `truncated === true` and a partial entry list — exactly your suggested approach, no >4 MB fixture. Both fixtures are built in-code and I cross-checked them against the parser directly (9/9).

packages/base/zip-file-def.gts
lukemelia: Fixed in 4350199b98 — hoisted a module-scope `basename(path)` helper and pointed all three sites at it (`ArchiveEntryField.name`'s computeVia, `fittedNames`, and `tree`), so the last-path-segment rule lives in one place and can't drift to three different names for one entry. Kept the `name` field (now a one-liner over the shared helper) rather than dropping it.

PR #5763, opened 2026-08-13
Fix flaky Percy diff related to resizable panel widths

backspace:
While Percy diffs are definitely reduced since the font fixes, I’ve seen this pattern occasionally:

image

This fixes it by:
* copying persisted panel widths instead of reusing, as that leads to bleed-through
* changing the test to drag by a fraction vs absolute amount

### DIFF
--- a/packages/host/app/components/operator-mode/code-submode.gts
+++ b/packages/host/app/components/operator-mode/code-submode.gts
@@ -207,11 +207,16 @@ export default class CodeSubmode extends Component<Signature> {
         rightPanel: !this.codePath ? 0 : persistedDefaultPanelWidths.rightPanel,
       }) <= 100
         ? persistedDefaultPanelWidths
-        : defaultPanelWidths;
+        : // Copy rather than alias. The layout callbacks below mutate whichever
+          // object lands here, and `defaultPanelWidths` is module-level state
+          // shared by every instance that ever falls back to it — mutating it
+          // would make one instance's final layout the next one's starting
+          // point, with nothing but a page load to clear it.
+          { ...defaultPanelWidths };
     this.defaultPanelHeights =
       persistedDefaultPanelHeights && sum(persistedDefaultPanelHeights) <= 100
         ? persistedDefaultPanelHeights
-        : defaultPanelHeights;
+        : { ...defaultPanelHeights };
 
     registerDestructor(this, () => {
       this.operatorModeStateService.unsubscribeFromOpenFileStateChanges(this);
--- a/packages/host/tests/acceptance/code-submode-test.ts
+++ b/packages/host/tests/acceptance/code-submode-test.ts
@@ -1620,6 +1620,27 @@ module('Acceptance | code submode tests', function (_hooks) {
       let previewResizeHandle = handles.at(-1);
       assert.ok(previewResizeHandle, 'preview panel resize handle exists');
 
+      // The panels either side of the handle, in group order.
+      let editorPanel = previewResizeHandle!
+        .previousElementSibling as HTMLElement;
+      let previewPanel = previewResizeHandle!.nextElementSibling as HTMLElement;
+      assert
+        .dom(previewPanel)
+        .hasAttribute(
+          'data-boxel-panel-id',
+          /.+/,
+          'the handle sits between two resizable panels',
+        );
+
+      // Drag by a fraction of the preview panel's own width rather than a
+      // fixed pixel count. A drag wider than the panel is clamped where the
+      // panel runs out of room, so only part of it lands — and the reverse
+      // drag, which is not clamped, then gives back more than the first one
+      // took, leaving the group in a layout neither drag asked for.
+      let initialEditorWidth = editorPanel.getBoundingClientRect().width;
+      let initialPreviewWidth = previewPanel.getBoundingClientRect().width;
+      let dragDistance = Math.round(initialPreviewWidth * 0.75);
+
       let shrinkRect = previewResizeHandle!.getBoundingClientRect();
       await triggerEvent(previewResizeHandle!, 'pointerdown', {
         button: 0,
@@ -1630,12 +1651,12 @@ module('Acceptance | code submode tests', function (_hooks) {
       await triggerEvent(previewResizeHandle!, 'pointermove', {
         pointerId: 1,
         buttons: 1,
-        clientX: shrinkRect.x + shrinkRect.width / 2 + 320,
+        clientX: shrinkRect.x + shrinkRect.width / 2 + dragDistance,
         clientY: shrinkRect.y + shrinkRect.height / 2,
       });
       await triggerEvent(previewResizeHandle!, 'pointerup', {
         pointerId: 1,
-        clientX: shrinkRect.x + shrinkRect.width / 2 + 320,
+        clientX: shrinkRect.x + shrinkRect.width / 2 + dragDistance,
         clientY: shrinkRect.y + shrinkRect.height / 2,
       });
 
@@ -1680,12 +1701,12 @@ module('Acceptance | code submode tests', function (_hooks) {
       await triggerEvent(previewResizeHandle!, 'pointermove', {
         pointerId: 2,
         buttons: 1,
-        clientX: widenRect.x + widenRect.width / 2 - 320,
+        clientX: widenRect.x + widenRect.width / 2 - dragDistance,
         clientY: widenRect.y + widenRect.height / 2,
       });
       await triggerEvent(previewResizeHandle!, 'pointerup', {
         pointerId: 2,
-        clientX: widenRect.x + widenRect.width / 2 - 320,
+        clientX: widenRect.x + widenRect.width / 2 - dragDistance,
         clientY: widenRect.y + widenRect.height / 2,
       });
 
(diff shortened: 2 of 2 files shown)


FadhlanR (approved):

packages/host/app/components/operator-mode/code-submode.gts
```diff
@@ -207,11 +207,16 @@ export default class CodeSubmode extends Component<Signature> {
         rightPanel: !this.codePath ? 0 : persistedDefaultPanelWidths.rightPanel,
       }) <= 100
         ? persistedDefaultPanelWidths
-        : defaultPanelWidths;
+        : // Copy rather than alias. The layout callbacks below mutate whichever
+          // object lands here, and `defaultPanelWidths` is module-level state
+          // shared by every instance that ever falls back to it — mutating it
+          // would make one instance's final layout the next one's starting
+          // point, with nothing but a page load to clear it.
+          { ...defaultPanelWidths };
```
richardhjtan, line 215: [Claude Code 🤖]

**Confirmation — the copy is load-bearing, and the comment describes the mechanism accurately.** Non-blocking; nothing to change here. Recording the verification because the reason this matters is not visible from this line.

**Why the alias was a real write, not a theoretical one.** The two layout callbacks lower in this file mutate the object in place rather than replacing it:

- `updatePanelWidths` assigns `this.defaultPanelWidths.leftPanel`, `.codeEditorPanel`, `.rightPanel`, `.emptyCodeModePanel`
- `updatePanelHeights` assigns `this.defaultPanelHeights.filePanel`, `.recentPanel`

So the field is the mutation target, and before the spread the fallback field *was* the module `const` declared at the top of this file. Every instance that fell back shared one object, and the first drag rewrote the module's idea of "default" for the rest of the page's life. The spread makes each instance's starting point independent, which is what `defaultPanelWidths` being a `const` was already implying.

**The other branch is already safe** — `persistedDefaultPanelWidths` is a `JSON.parse` result, a fresh object per construction, so it needs no copy.

**When this stops being safe.** The invariant is "the object in this field is owned by this instance." A future nested-object field in `PanelWidths` would make the shallow spread insufficient; and the guard disappears if anyone later replaces the in-place assignments with a whole-object assign and then "simplifies" the spread away as redundant. The comment carries enough of the why to stop that, which is the main thing.

---

richardhjtan (commented):
[Claude Code 🤖]

**Lens:** this change claims a specific mechanism — a shared mutable default bleeding one instance's layout into the next — so the review traced every path by which a code-mode panel layout can survive past the instance that produced it, and checked whether the two commits together actually close all of them.

**Bottom line: the fix is correct and the test is genuinely stronger, no blocking issues. But the copy closes the narrower of two bleed-through channels — the persisted-layout channel is still wide open, and there is a test in the suite leaking through it today.** Details in recommendation 1; everything else is confirmation.

**What lands right.**

- **The copy is necessary, not defensive.** `updatePanelWidths` / `updatePanelHeights` assign into `this.defaultPanelWidths.*` and `this.defaultPanelHeights.*` in place, so whatever object lands in those fields is the mutation target. Both fallbacks were module-level `const` objects, which makes the aliasing a real cross-instance write. Verified in the inline thread.
- **No twin to fix.** The only other panel-width state in operator mode, `aiPanelWidths` in `submode-layout.gts`, is a `new TrackedObject({...})` class-field initializer — evaluated per instance, so it never aliased shared state. The persisted branches here are `JSON.parse` results, also fresh per instance. This diff covers every site that had the problem.
- **The proportional drag is the right shape and the reasoning behind it checks out** against the actual constraint values, which live in a different file than the comment. Verified in the inline thread.
- **Asserting the widths rather than the chooser mode is the substantive half of the test change.** The chooser reports `full` for any container above its natural width, so the old assertions passed across a wide band of layouts; the pair of width checks is what actually pins "the widen drag gave back what the shrink drag took".

**Recommendations.**

1. **The persisted-layout channel still bleeds between tests, and it is the one that survives.** Non-blocking follow-up, but it is the larger half of the problem this PR names. `updatePanelWidths` writes the layout to `localStorage` under `CodeModePanelWidths` on every layout change, and nothing in the host test setup clears it — `clearLocalStorage` in `app/utils/local-storage-keys.ts` does cover both panel keys, but its only caller is the logout path in `matrix-service.ts`. Testem runs the whole acceptance suite in one Chrome page, one `localStorage`. So after the first code-mode resize anywhere in a run, every later `CodeSubmode` takes the *persisted* branch of this constructor — the branch where the new copy is irrelevant — and starts from that layout.

   Not hypothetical: `tests/acceptance/code-submode/inspector-test.ts` sets `CodeModePanelHeights` to `{ filePanel: 50, recentPanel: 50 }` and never removes it, and that module takes Percy snapshots at five points after it. Every subsequent code-mode test in the run inherits a 50/50 file/recent split instead of the computed default (roughly 76/24 at the CI window height), which is exactly the kind of shifted-panel diff this PR is chasing.

   The structural fix is a `beforeEach` in `setupApplicationTest` that resets both panel keys, mirroring `seedAiAssistantClosed` — which exists for precisely this cross-test `localStorage` persistence problem and is the established precedent for it. That would also mean individual tests no longer have to be well-behaved about restoring layout for the suite to stay deterministic.

2. **Assert both siblings are panels, not just one** — one line, keeps the test's own premise from going quiet. Inline thread on the sibling lookup.

**Adjacent, out of scope.** The module-level defaults are computed once at module evaluation from `document.documentElement.clientWidth` / `clientHeight` and never recomputed. That is deterministic under CI's pinned `--window-size=1440,900`, so it is not a Percy flake source — but it does mean local runs and CI compute different fallback percentages, which is worth knowing when a layout assertion passes locally and fails in CI. Not asked of this PR.

---

packages/host/tests/acceptance/code-submode-test.ts
```diff
@@ -1620,6 +1620,27 @@ module('Acceptance | code submode tests', function (_hooks) {
...
+        );
+
+      // Drag by a fraction of the preview panel's own width rather than a
+      // fixed pixel count. A drag wider than the panel is clamped where the
+      // panel runs out of room, so only part of it lands — and the reverse
+      // drag, which is not clamped, then gives back more than the first one
+      // took, leaving the group in a layout neither drag asked for.
+      let initialEditorWidth = editorPanel.getBoundingClientRect().width;
+      let initialPreviewWidth = previewPanel.getBoundingClientRect().width;
+      let dragDistance = Math.round(initialPreviewWidth * 0.75);
```
richardhjtan, line 1649: [Claude Code 🤖]

**Confirmation — the asymmetry this comment describes is real, and 75% is inside the constraints in both directions.** Non-blocking. Verifying it here because the comment asserts a mechanism whose deciding values live in two other files, so nothing near this line can be used to check it.

**Why a fixed 320px was asymmetric.** In `boxel-ui`'s `adjust-layout-by-delta.ts`, the shrinking side of a drag is clamped through `resizePanel`, which floors a panel at `minSize` (or `0` when `collapsible`). The two panels either side of this handle have very different constraints, both set in `code-submode.gts`:

- the preview panel: `@collapsible={{false}}` and **no** `@minSize`, so `minSize` defaults to `0`
- the code editor panel: `@minSize={{20}}`

Shrink drag (positive delta): the preview is the shrinking side and floors at `0`. A 320px request against a preview narrower than 320px applies only as much as the panel had, and the rest is dropped. Widen drag (negative delta): the editor is now the shrinking side, and it has 20% of headroom to give, so the full 320px applies. The reverse drag hands back more than the first one took — the group lands somewhere neither drag asked for, which is the drift the snapshot was catching.

**Why the proportional form is symmetric.** `0.75 * initialPreviewWidth` leaves the preview at 25% of its starting width, comfortably above its `0` floor, so the shrink is unclamped and `deltaApplied` equals the full request. The widen then asks the editor for the same amount, returning it to its starting width — which stays above `minSize: 20` from the roughly-43% starting split. Both drags apply in full, so the round trip is exact, and the `<= 1` tolerance covers the `toPrecision` rounding in the layout math.

Worth noting the symmetry also depends on `widenRect` being re-read after the shrink, which it is — so `-dragDistance` is measured from the handle's new position rather than its original one.

---

packages/host/tests/acceptance/code-submode-test.ts
```diff
@@ -1620,6 +1620,27 @@ module('Acceptance | code submode tests', function (_hooks) {
...
+      // The panels either side of the handle, in group order.
+      let editorPanel = previewResizeHandle!
+        .previousElementSibling as HTMLElement;
+      let previewPanel = previewResizeHandle!.nextElementSibling as HTMLElement;
+      assert
+        .dom(previewPanel)
+        .hasAttribute(
+          'data-boxel-panel-id',
+          /.+/,
+          'the handle sits between two resizable panels',
```
richardhjtan: [Claude Code 🤖]

**The assertion checks one of the two siblings this test depends on.** Non-blocking nit, one line to close.

`editorPanel` and `previewPanel` are both taken positionally from the handle, but only `previewPanel` is asserted to carry `data-boxel-panel-id`. `editorPanel` is used just as load-bearingly — the last assertion in the test compares its width against `initialEditorWidth`. If the template ever gains a wrapper or a non-panel element before the handle, `previousElementSibling` silently becomes something else whose width is probably stable, so that final assertion would keep passing while no longer testing the editor panel at all.

Since the assertion for the premise is already being written, covering both siblings costs one line:

```js
assert
  .dom(editorPanel)
  .hasAttribute(
    'data-boxel-panel-id',
    /.+/,
    'the handle sits between two resizable panels',
  );
```

(Then the existing `previewPanel` assertion's message can narrow to naming the preview side.)

---

richardhjtan (approved):

backspace:
> Testem runs the whole acceptance suite in one Chrome page, one `localStorage`.

this isn’t true, `setupWindowMock` means it’s cleared between tests

PR #5764, opened 2026-08-13
Clip adorn card overlays behind the rich-markdown compose toolbar

FadhlanR:
## Problem

In the rich-markdown **Compose** editor, `:card[…]` references render as live embedded card previews. When you hover or select one, the operator-mode adorn treatment (green box-shadow outline + type-label tab) appears. Scrolling a decorated card up behind the sticky formatting toolbar made that green outline paint **on top of** the toolbar instead of disappearing underneath it.

## Root cause

The adorn treatment is the operator-mode overlay layer — an absolutely-positioned **later sibling** of the card renderer, so it paints above the card's entire subtree. The sticky toolbar's `z-index: 10` is sealed inside two stacking contexts it can't escape (`.boxel-card-container`'s `z-index: 0`, and `.default-card-template`'s `container-type` layout containment). So no z-index change can fix it: the overlay is one flat layer above the whole card, while the toolbar is a sub-region buried inside it.

The stack-item header already avoids this — but only because it sits in a grid row **outside** the `overflow: auto` scroll container that clips the overlays. This change reproduces that same overflow-occlusion, applied dynamically at the toolbar's bottom edge.

## Fix

- `operator-mode/overlays.gts`: the overlay's existing per-frame floating-ui `offset` middleware now sets `clip-path` on the overlay, cutting off the strip that has slid behind a marked sticky header. Two pure, unit-tested helpers: `stickyClipHeaderFor()` (find the header for a card's clip container) and `stickyHeaderClipPath()` (build the inset). Negative side/bottom insets preserve the box-shadow ring on the still-visible portion; the clip clears when the card is in the open. It's a no-op for overlays outside a marked container, so other overlay consumers (spec-preview, playground, preview-panel) are unaffected.
- `codemirror-editor.gts`: marks the compose editor scope with `data-overlay-clip-container` and its sticky toolbar with `data-overlay-clip-header` — the boundary contract, avoiding CSS-class coupling.

Cost is negligible: an overlay is rendered only for a hovered/selected card (0–1 during a scroll), and the layout read already happens each frame.

## Interactivity is preserved

The clip is purely visual. The overlay is `pointer-events: none`; hover and click-to-open are wired to listeners on the underlying **card element**, not the overlay. The only in-overlay controls (type-label menu, select chip) are clipped only where they sit behind the (opaque, unreachable) toolbar. Embedded preview cards remain fully interactive.

## Testing

- **Unit** (`tests/unit/overlays-clip-test.ts`): `stickyHeaderClipPath` (no clip below/flush with the header, correct top inset when the card slides under it) and `stickyClipHeaderFor` (finds the header, returns null outside a container, scopes to the nearest container when editors are nested).
- `ember-tsc` clean on the changed host file.
- **Pending**: manual visual check in the running app (outline cut cleanly at the toolbar edge, box-shadow ring intact on the visible sides, smooth on scroll, controls still work when clear). Scope is compose-only; the preview sub-mode bar is out of scope.

## Risk

The one behavioral dependency is `clip-path: inset()` with negative offsets (to keep the box-shadow ring on the un-clipped sides) — well-supported in the target Chromium. Fallback if it ever regresses: `inset(<top>px 0 0 0)`, accepting a slightly thinner side outline.

### DIFF
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -929,7 +929,7 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
           boundary the operator-mode overlay layer clips against: adorn overlays
           drawn over embedded compose-preview cards are cut off where a card
           scrolls behind the sticky toolbar, so the outline no longer paints on
-          top of it (CS-11699). Consumed by the `offset` middleware in
+          top of it. Consumed by the `offset` middleware in
           operator-mode/overlays.gts. }}
       <div
         class='codemirror-editor'
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -926,11 +926,12 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
   <template>
     {{#if this.cm}}
       {{! `data-overlay-clip-container` / `data-overlay-clip-header` mark the
-          boundary the operator-mode overlay layer clips against: adorn overlays
-          drawn over embedded compose-preview cards are cut off where a card
-          scrolls behind the sticky toolbar, so the outline no longer paints on
-          top of it. Consumed by the `offset` middleware in
-          operator-mode/overlays.gts. }}
+          scroll scope and its sticky header so an overlay layer can clip a
+          decorated card's overlay behind the header: adorn overlays drawn over
+          embedded compose-preview cards are cut off where a card scrolls behind
+          the sticky toolbar, so the outline no longer paints on top of it. This
+          is a data-attribute contract — the consumer finds these markers by
+          attribute, not by importing this module. }}
       <div
         class='codemirror-editor'
         data-overlay-clip-container
--- a/packages/base/codemirror-editor.gts
+++ b/packages/base/codemirror-editor.gts
@@ -925,10 +925,25 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
 
   <template>
     {{#if this.cm}}
-      <div class='codemirror-editor' data-test-codemirror-editor ...attributes>
+      {{! `data-overlay-clip-container` / `data-overlay-clip-header` mark the
+          boundary the operator-mode overlay layer clips against: adorn overlays
+          drawn over embedded compose-preview cards are cut off where a card
+          scrolls behind the sticky toolbar, so the outline no longer paints on
+          top of it (CS-11699). Consumed by the `offset` middleware in
+          operator-mode/overlays.gts. }}
+      <div
+        class='codemirror-editor'
+        data-overlay-clip-container
+        data-test-codemirror-editor
+        ...attributes
+      >
         {{! ── Docked toolbar ── }}
         {{! template-lint-disable no-pointer-down-event-binding }}
-        <div class='codemirror-toolbar' data-test-markdown-toolbar>
+        <div
+          class='codemirror-toolbar'
+          data-overlay-clip-header
+          data-test-markdown-toolbar
+        >
           {{yield to='leadingControls'}}
           {{#if (has-block 'leadingControls')}}
             <span class='toolbar-divider'></span>
--- a/packages/host/app/components/operator-mode/overlays.gts
+++ b/packages/host/app/components/operator-mode/overlays.gts
@@ -21,7 +21,7 @@ import type { CardDefOrId } from './stack-item';
 
 import type { RenderedCardForOverlayActions } from '../../resources/element-tracker';
 import type { CardDef, Format, ViewCardFn } from '@cardstack/base/card-api';
-import type { MiddlewareState } from '@floating-ui/dom';
+import type { MiddlewareState, ReferenceElement } from '@floating-ui/dom';
 
 interface OverlaySignature {
   Args: {
@@ -74,6 +74,29 @@ export function stickyClipHeaderFor(reference: Element): HTMLElement | null {
       .closest('[data-overlay-clip-container]')
       ?.querySelector<HTMLElement>('[data-overlay-clip-header]') ?? null
   );
+}
+
+// Reads the live rects of `reference` and its clip container's sticky header
+// and assigns the resulting clip to `floating.style.clipPath` — the glue that
+// turns the pure helpers above into the running fix. Both elements must be laid
+// out in the document. Clears to '' when the reference is not an HTMLElement,
+// sits in no marked container, or has scrolled clear of the header. Exported so
+// a test can drive it without standing up the floating-ui middleware.
+export function applyStickyHeaderClip(
+  reference: ReferenceElement,
+  floating: HTMLElement,
+): void {
+  let clipPath = '';
+  if (reference instanceof HTMLElement) {
+    let header = stickyClipHeaderFor(reference);
+    if (header) {
+      clipPath = stickyHeaderClipPath(
+        reference.getBoundingClientRect().top,
+        header.getBoundingClientRect().bottom,
+      );
+    }
+  }
+  floating.style.clipPath = clipPath;
 }
 
 export default class Overlays extends Component<OverlaySignature> {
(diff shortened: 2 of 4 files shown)


packages/base/codemirror-editor.gts
```diff
@@ -1130,10 +1130,25 @@ export default class CodeMirrorEditor extends GlimmerComponent<CodeMirrorEditorS
...
+          drawn over embedded compose-preview cards are cut off where a card
+          scrolls behind the sticky toolbar, so the outline no longer paints on
+          top of it. Consumed by the `offset` middleware in
+          operator-mode/overlays.gts. }}
+      <div
+        class='codemirror-editor'
+        data-overlay-clip-container
+        data-test-codemirror-editor
+        ...attributes
+      >
```
habdelra, line 940: [Claude Code 🤖]

**Cross-package contract — non-blocking, design note.** These `data-overlay-clip-container` / `data-overlay-clip-header` markers live in `packages/base`, but their only consumer is the operator-mode overlay `offset` middleware in `packages/host` (`operator-mode/overlays.gts`). That's a sound design as a *data-attribute* contract — `base` doesn't import `host`, and decoupling from a CSS class is the right call — but the comment here pins the contract to a host file by path. Nothing compiles against that reference, so it silently rots if the host file is renamed or moved. Consider phrasing the contract semantically (e.g. "marks the scroll scope and its sticky header so an operator-mode overlay layer can clip decorated cards behind the header") rather than naming the host file.

**Scope wording — non-blocking.** The markers are emitted for *every* `CodeMirrorEditor` instance, including `source` mode — `rich-markdown.gts` renders the editor for both compose and source and only toggles `@livePreview={{eq this._mode 'compose'}}`. It's harmless (no card previews register overlays in source mode, so `stickyClipHeaderFor` may find a header but no overlay ever anchors inside a source editor), but the PR description's "marks the compose editor scope" slightly overstates it. The effective scope is "wherever an operator-mode overlay lands over a card inside a marked container," which today is compose-only.

---

packages/host/app/components/operator-mode/overlays.gts
```diff
@@ -45,6 +45,37 @@ interface OverlaySignature {
...
+// keep the adorn box-shadow ring (≤0.25rem spread) visible on the un-clipped
+// edges; '' fully restores the overlay (and its top-overhanging type-label) when
+// nothing occludes it. Exported for unit testing.
+export function stickyHeaderClipPath(
+  refTop: number,
+  headerBottom: number,
+): string {
+  let clipTop = Math.max(0, headerBottom - refTop);
+  return clipTop > 0 ? `inset(${clipTop}px -0.5rem -0.5rem -0.5rem)` : '';
+}
```
habdelra, line 78: [Claude Code 🤖]

**Confirmation — the negative insets are correctly sized.** Verified against `adorn/adorn-context.gts`: the widest adorn ring is `box-shadow: 0 0 0 0.25rem var(--adorn-accent)` (`.adorn-stroke.selected:hover`), and `clip-path` clips box-shadow, so a plain `inset(${clipTop}px 0 0 0)` would shave the ring off the visible edges. The `-0.5rem` side/bottom insets extend the clip region 0.5rem past the box, preserving the full 0.25rem ring with margin. The "≤0.25rem spread" note is accurate — worth keeping it in sync if that box-shadow ever grows past 0.5rem.

**One narrow visual edge for the pending manual check — non-blocking.** The `-0.5rem` bottom inset is sized for the ring, but a type-label that flips *below* the card also lives below the box: in `modifiers/position-adorn-label.ts`, `side === 'bottom'` sets `anchorTopY = cardRect.bottom`, so a flipped label extends ~one label-height below the box and an active clip would cut it to 0.5rem. In practice this shouldn't bite — `AdornContext`'s boundary resolves to the whole stack-item content area (well above the compose toolbar), so `spaceAbove` stays large and the label stays on `top` (where clipping it *is* the goal) whenever a card is behind the toolbar; the flip-below branch is only reachable when a card sits right at the boundary top. Just worth a glance if the editor ever sits near the top of the stack item.

---

packages/host/app/components/operator-mode/overlays.gts
```diff
@@ -135,6 +166,34 @@ export default class Overlays extends Component<OverlaySignature> {
...
+      if (reference instanceof HTMLElement) {
+        let header = stickyClipHeaderFor(reference);
+        if (header) {
+          clipPath = stickyHeaderClipPath(
+            refRect.top,
+            header.getBoundingClientRect().bottom,
+          );
+        }
+      }
+      floating.style.clipPath = clipPath;
```
habdelra: [Claude Code 🤖]

**Test coverage — follow-up, non-blocking.** `stickyHeaderClipPath` and `stickyClipHeaderFor` are well covered by the new unit tests, but the wiring here — reading `refRect.top` and `header.getBoundingClientRect().bottom`, then assigning `floating.style.clipPath` — is the part that actually produces the fix, and nothing exercises it. I understand why an exact-pixel integration assertion is impractical: the `#ember-testing` scale is the very reason `clipTop` isn't scale-divided, so the clip line is only approximate under the runner. But a coarse assertion is still achievable and would lock the contract against regression: render a decorated card whose top sits above a marked header's bottom and assert `floating.style.clipPath` is non-empty, then scroll it clear and assert it returns to `''`. `rich-markdown-field-test.gts` / `codemirror-embed-toolbar-test.gts` already stand up the compose editor, so the fixture cost is low. Given the PR's own "manual visual check pending," filing this as a follow-up rather than blocking is reasonable.

---

habdelra (commented):
[Claude Code 🤖]

**Lens:** whether the clip geometry is correct in production vs. the scaled test runner, whether the change stays inert for the other overlay consumers, and how the new `base`↔`host` marker contract holds up.

**Bottom line: the mechanism is sound and correct for production — no blocking issues.** The adorn overlay is a positioned sibling that paints above the card's stacking-context-sealed subtree, so z-index genuinely can't arbitrate it against a sub-region toolbar; physically clipping the strip behind the sticky header is the right fix, and it's implemented cleanly and cheaply (0–1 overlays during a scroll, layout read already happens each frame). Everything below is confirmation plus non-blocking follow-ups.

**What lands right** (verified against the checked-out head):
- **Geometry.** In production the floating element is unscaled — its height is set to `refRect.height` and its top aligned to `refRect.top` — so an inset top of `headerBottom - refTop` lands exactly at the toolbar's bottom edge. The scale caveat is real but confined to `#ember-testing`, and the inline comment documents it precisely.
- **Box-shadow ring preservation.** Confirmed against `adorn-context.gts`: the widest ring is `0 0 0 0.25rem`, and the `-0.5rem` side/bottom insets extend the clip 0.5rem past the box, so the ring survives on the un-clipped edges. Detail in the thread on `stickyHeaderClipPath`.
- **Inert for other consumers.** `stickyClipHeaderFor` returns null outside a marked container, so `clipPath` is set to `''` (cleared) each frame for spec-preview / playground / preview-panel. No other code or CSS sets `clip-path` on the overlay element — `adorn-label`'s own `clip-path` is on the child label (a different element) and composes correctly under the parent clip.
- **No test-selector regression.** `data-test-codemirror-editor` and `data-test-markdown-toolbar` are preserved; the new `data-overlay-clip-*` attributes are purely additive.

1. Base-side comment pins the contract to a host file by path — see the thread on `codemirror-editor.gts`.
2. Markers are emitted for every `CodeMirrorEditor`, not just compose; the description's scope wording slightly overstates it — same thread.
3. The pure helpers are tested but the middleware wiring that produces the fix isn't — see the thread on the clip assignment in `overlays.gts`.
4. A narrow flip-below type-label edge to glance at during the pending manual check — see the thread on `stickyHeaderClipPath`.

**CI note:** the red `Host Tests (5, 20)` shard is an infrastructure flake — it died during stack setup (no testem log produced, followed by orphan-process teardown) before running tests, while the aggregate `Host Test Results` check is green with 0 failures at this commit. Not attributable to this change.

---

habdelra (approved):

PR #5770, opened 2026-08-13
Update assistant with search/replace marker tolerance

backspace:
In a recent assistant chat I was repeatedly seeing things meant to be diffs shown not as diffs:

s 2026-08-13 at 10 34 13@2x

The search/replace divider there is `╠═══════════════════╣`. That’s 19 inner characters when it’s expected to be 15. This is surely the known difficulty LLMs have with counting.

This adds: 
* flexibility when trying to find search/replace markers
* a warning when a patch is malformed

Claude had a grander idea about bypassing this problem altogether (“moving patch authority out of the prose into structured event content”), but that’s something to discuss.

on the preview deployment, it’s fixed, applyable:

s 2026-08-13 at 11 28 25@2x


packages/runtime-common/search-replace-markers.ts
```diff
@@ -0,0 +1,110 @@
...
+// REPLACE markers, whose runs are anchored either side of a word, do not drift.
+// Match the runs by shape rather than by length — a miscounted rule still
+// describes an unambiguous patch, and treating it as prose silently discards a
+// diff the user asked for.
+//
+// The constants remain the canonical form to *emit* (skill instructions, tool
+// descriptions); these patterns are the tolerant form to *recognize*.
+
+export const SEARCH_MARKER_PATTERN = /╔═+ SEARCH ═+╗/;
+export const SEPARATOR_MARKER_PATTERN = /╠═+╣/;
```
habdelra: [Claude Code 🤖] **Confirmation, with one boundary worth naming — non-blocking.**

Recognizing by shape rather than by exact width is the right call, and routing every consumer through one primitive is the strongest part of this change. I verified there are no raw-marker consumers left in production code: a grep for the string constants across non-test source finds only their definitions in `constants.ts` and the patterns here — every one of the five former exact-string sites (`matrix-service.containsSearchReplaceMarker`, `apply-search-replace-block.parseCodeBlock`, `matrix-utils.extractCodePatchBlocks`, `prompt.elideCodeBlocks`, and the host `isCompleteSearchReplaceBlock`) now goes through `findSearchReplaceBlock` / `containsSearchReplaceMarker`. So client apply, server correctness-check, and display can no longer drift apart on what counts as a marker.

**The boundary:** `SEPARATOR_MARKER_PATTERN = /╠═+╣/` recognizes *any* `╠…═…╣` run, so a separator-shaped sequence inside search or replace **content** — not just a trailing one — is now taken as the divider. I confirmed this empirically: for `SEARCH … {a ╠═╣ b} … SEP … REPLACE`, `findSearchReplaceBlock` picks the inner `╠═╣` as `separator`, truncating the search half. The old exact matcher had the same failure class but a narrower mouth (only the literal 15-═ run). The block comment's "real code/JSON never legitimately ends with it" justifies the *trailing*-separator strip in `stripTrailingSeparatorMarker`, but doesn't cover a divider-shaped run appearing mid-content. Risk is low (box-drawing characters in edited code are rare) and I'm not asking for a code change — just flagging the widened surface so whoever debugs a mangled patch someday knows this is where content-vs-marker ambiguity lives.

---

packages/host/app/lib/formatted-message/utils.ts
```diff
@@ -102,15 +103,17 @@ export function extractCodeData(
...
   return {
     language: language ?? '',
     code: codeToDisplay,
     fileUrl: fileUrl ?? null,
     isNewFile,
-    searchReplaceBlock: isCompleteSearchReplaceBlock(contentWithoutFirstLine)
-      ? contentWithoutFirstLine
-      : null,
+    searchReplaceBlock,
+    malformedPatch: isBeginningOfSearchReplaceBlock && !searchReplaceBlock,
```
habdelra, line 116: [Claude Code 🤖] **Two notes on the `malformedPatch` computation — non-blocking.**

**1. The warning's trigger is looser than its wording.** `malformedPatch` is `isBeginningOfSearchReplaceBlock && !searchReplaceBlock`, and `isBeginningOfSearchReplaceBlock` (line 79) is decided by `startsSearchMarker(lines[1])` → `/^╔═/`. That is much looser than the full `SEARCH_MARKER_PATTERN` (`/╔═+ SEARCH ═+╗/`): any `<pre>` whose second line merely *begins* `╔═` — box-drawing art, a table, a decorative rule — that doesn't parse as a complete block will, once streaming ends, render the footer "the search/replace markers in this block are malformed, so it cannot be applied automatically." For genuinely truncated patches that message is exactly right; for non-patch content that happens to open `╔═` it asserts an intent the author never had. The looseness is deliberate (a drifted SEARCH marker must still be caught), so this is a wording/precision trade-off to be aware of, not a bug — but if false positives show up in practice, tightening `startsSearchMarker` toward the real marker shape (or requiring the header line to look like a file path) is the lever.

**2. This computation is the one new behavior with no direct test.** `code-patching-test.ts` covers the recognition primitives well (`parseSearchReplace`, `isCompleteSearchReplaceBlock`, `containsSearchReplaceMarker`), but nothing asserts `extractCodeData` itself sets `malformedPatch` — true for a `╔═`-headed block that never completes, false for a well-formed one, false for ordinary code. A short unit test over `extractCodeData` would pin the field that drives the whole new UI path. Follow-up, not blocking.

---

packages/host/app/components/ai-assistant/message/aibot-message.gts
```diff
@@ -252,6 +254,20 @@ class HtmlGroupCodeBlock extends Component<HtmlGroupCodeBlockSignature> {
       : null;
   }
 
+  // A block that opened a SEARCH marker but never resolved into a complete
+  // search/replace block is unapplyable, and without this reads as an ordinary
+  // code block with stray box-drawing characters in it. Wait for streaming to
+  // finish first, since a patch mid-stream is incomplete for a moment.
+  private get showMalformedPatchError() {
+    return !this.args.isStreaming && this.args.codeData.malformedPatch;
+  }
```
habdelra, line 263: [Claude Code 🤖] **The streaming gate is the subtle half of this feature, and it's the untested half — non-blocking follow-up.**

`showMalformedPatchError = !isStreaming && malformedPatch` is what keeps the warning from flashing on every patch mid-stream (where `malformedPatch` is legitimately true before the SEPARATOR/REPLACE arrive). I traced the two facts it leans on and both hold:

- **Split patches reassemble before this is evaluated.** A continuation-split block would otherwise leave the first `<pre>` holding only the SEARCH half → `malformedPatch` true forever. But `Message.body` is `[this._body, this.continuedBody].filter(Boolean).join('')`, so `htmlParts` → `extractCodeData` sees the joined body, and `isStreamingFinished` follows the last continuation. So the gate stays closed until the whole thing has landed. Good.
- The gate correctly renders in the `{{else}}` of `{{#if (bool @codeData.searchReplaceBlock)}}` — i.e. only when there is no complete block to diff — which is consistent with `malformedPatch ⇒ searchReplaceBlock == null`.

What's missing is a test that pins this contract. The integration file `formatted-aibot-message-test.gts` is the natural home: a `╔═`-headed block that never completes should show `[data-test-malformed-patch-error]` **after** streaming finishes and **not** while `@isStreaming` is true. Without it, a future change that flips the gate (or stops joining continuations) regresses silently — the failure mode is a warning that either never appears or flickers on every valid patch. Worth adding as a follow-up.

---

habdelra (commented):
[Claude Code 🤖] Reviewed through one lens: does making marker recognition tolerant land consistently across every place that recognizes a patch, and does the new "malformed" signal fire on exactly the right blocks.

**Bottom line: no blocking issues.** This is a clean, well-motivated change. The core move — recognize markers by *shape* behind a single primitive, keep the constants as the canonical form to *emit* — is correct, and the consolidation removes a real drift hazard. Everything below is confirmation or a follow-up; nothing needs to change before merge.

- **The consolidation is the headline, and it's genuine.** Five sites each re-implemented the exact-string test; a drifted separator failed all five the same way. I grepped non-test source for the string constants and found only their definitions — every recognizer now routes through `findSearchReplaceBlock` / `containsSearchReplaceMarker`, so client apply, server correctness-check, and display can't disagree about what a marker is.
- **The refactors are behaviorally equivalent for well-formed input and strictly more robust for malformed input.** I reproduced the patterns and confirmed: `extractCodePatchBlocks` matches the old non-greedy `SEARCH.*?SEP.*?REPLACE` regex block-for-block; the `for(;;)` loop can't spin (each block's `end` strictly exceeds `from`); and `elideCodeBlocks` is actually *safer* than the old `while (includes && includes && includes)` form, which could take an out-of-order `indexOf(..., -1)` path — the new in-order `findSearchReplaceBlock` simply returns null and stops.
- **`containsSearchReplaceMarker` matching any single marker is the right check for the split-event queueing path**, since a continuation-split block leaves each half holding only part of the block. The new test pins both halves.

1. `search-replace-markers.ts` — the tolerant separator recognizes a divider-shaped run *inside* content, not just a trailing one; naming that boundary in the thread (widened, but low-risk). See the comment on `SEPARATOR_MARKER_PATTERN`.
2. `utils.ts` — the malformed-warning trigger (`startsSearchMarker` → `/^╔═/`) is looser than the message's wording implies, so non-patch `╔═`-headed blocks can draw the warning; and `extractCodeData`'s `malformedPatch` computation has no direct test. See the comment on line 116.
3. `aibot-message.gts` — the `!isStreaming` gate is the subtle part of the feature and is untested; an integration test in `formatted-aibot-message-test.gts` asserting the footer appears after streaming and not during would lock it in. See the comment on `showMalformedPatchError`.

**Adjacent, out of scope:** the emit-side constants in `constants.ts` and the format documented in the `apply-search-replace-block.ts` / `search-replace-block-parsing.ts` header comments are now the *only* place the canonical width lives — worth a one-line pointer from the constants to `search-replace-markers.ts` someday so a future editor who tightens the emitted marker doesn't forget the recognizer is deliberately looser. Not this PR.

---

habdelra (approved):

packages/runtime-common/search-replace-markers.ts
backspace: [Claude Code 🤖] Confirmed the ambiguity and closed it in 34f60714ad, so this went a bit further than "no code change needed".

You're right that the mouth widened, and the reasoning about *why* the separator is the vulnerable one is the useful part: SEARCH and REPLACE carry a word no real content contains, but a separator is only a run of ═ between two brackets — a shape that can plausibly occur in the code being patched. The old exact matcher had the same failure class, just a narrower mouth.

The fix leans on something the format already guarantees: the model writes each marker on a line of its own. So the separator is now anchored to one:

```js
export const SEPARATOR_MARKER_PATTERN = /^[ \t]*╠═+╣[ \t]*\r?$/m;
```

SEARCH and REPLACE stay unanchored — their embedded word already disambiguates them, and leaving them loose keeps the diff small.

Two details worth recording. The `\r?` is load-bearing: under `/m`, JS `$` matches before `\n` but not before `\r`, so CRLF content would otherwise stop parsing entirely. And anchoring changes what `separator.index`/`.end` point at (they now bracket the leading/trailing whitespace), so I checked every consumer of that arithmetic — `parseSearchReplace` trims and strips a leading newline either side, and `apply-search-replace-block.parseCodeBlock` does the same, so the slices are unchanged.

Verified against your exact example in both halves — `let box = "{a ╠═╣ b}"` now stays in the content and the real divider is still found — plus indented separators, CRLF, and the 19-═ drift regressions. Unit suite 17/17, integration 14/14 locally.

packages/host/app/lib/formatted-message/utils.ts
backspace: [Claude Code 🤖] Took the second point and deliberately left the first.

**On the missing test (point 2):** added in 34f60714ad — a `Unit | code patching | malformed patch detection` module driving `extractCodeData` directly and asserting the field four ways: true for a `╔═`-headed block that never completes, false for a well-formed one, false for a block whose separator drifted (since that still parses), and false for ordinary code with no marker. You were right that this was the one new behaviour with nothing holding it, and it's the field the entire new UI path keys off.

**On the loose trigger (point 1):** leaving `startsSearchMarker` as `/^╔═/`, and I want to be explicit that this is a decision rather than an oversight. Tightening it toward the real marker shape — say requiring the word `SEARCH` — would cost streaming behaviour: the header line is only read as a file URL once line 1 looks like a marker opening, so demanding the word delays that until it arrives mid-stream. There's an existing test, `it will text code clocks as they are sent, without hiding the first line url as it streams`, that covers exactly this window. Trading a real streaming regression for a hypothetical false positive isn't worth it.

Your framing of the trade-off is right though, and the lever you named is the one I'd reach for if false positives ever show up in practice. The new tests at least pin the current behaviour, so anyone tightening it later will see what they're trading against.

packages/host/app/components/ai-assistant/message/aibot-message.gts
backspace: [Claude Code 🤖] Added in 14030b92b0, in the file you suggested.

Three tests, covering both directions of the gate rather than just the happy one, since as you say the failure modes are symmetric — a warning that never appears and one that flickers on every valid patch are equally bad and neither shows up in a single-direction test:

- silent while a `╔═`-headed block is still streaming (`@isStreaming` true)
- shown once streaming ends with the block still unfinished
- absent entirely for a patch that completed

The middle one drives the transition within a single component instance — flipping `isStreaming` from true to false on a tracked test component — rather than rendering two separate scenarios, so it pins the transition itself and not just the two end states.

Thanks for tracing the continuation-joining half. That was the fact I was least sure of when I wrote the gate: if `htmlParts` ever saw a split block's first `<pre>` in isolation, `malformedPatch` would be true forever and the gate would be load-bearing in a way it isn't designed to be. Having `Message.body` join `_body` and `continuedBody` before `extractCodeData` runs is what makes it safe, and that's now a documented reason rather than an assumption I happened to get right.

Integration suite 14/14 locally against test services.

PR #5771, opened 2026-08-13
host: Move standalone test modules to shared realm cache

backspace:
This is a step toward removing `reuseIndexAcrossTests`, as `withCachedRealmSetup` already existed and produced similar time savings.


packages/host/tests/integration/components/operator-mode/setup.gts
```diff
@@ -62,7 +57,12 @@ export function setupOperatorModeTests(
     operatorModeStateService = getService('operator-mode-state-service');
   });
 
-  setupLocalIndexing(hooks, opts);
+  setupLocalIndexing(hooks);
+  // Registers for the scope this helper is called from. A consumer with nested
+  // modules registers its own too: the delete-prefix is fixed at registration
+  // from the module name, and a nested test's snapshot is keyed by its own
+  // composed name, which the outer prefix cannot match.
+  setupRealmCacheTeardown(hooks);
```
habdelra, line 65: [Claude Code 🤖] **The comment is right, and that is the problem: "a consumer with nested modules registers its own too" is a contract nothing enforces, and 23 nested scopes already in the suite violate it.**

**The mechanism.** `setupRealmCacheTeardown` resolves the module name at *registration* time and freezes `snapshot_${simpleHash(name)}_` as its delete prefix (`snapshotPrefixForModule` in `packages/host/tests/helpers/index.gts`). `withCachedRealmSetup` resolves the module name at *call* time — `getCurrentModuleCacheKey()` reads `QUnit.config.current?.module?.name` — and builds the snapshot name from it via `snapshotNameForCacheKey`. When those two names differ, which is exactly the nested-module case, the `after` hook's prefix never matches the snapshot that was created and the snapshot is never deleted.

**What a miss costs.** `exportSnapshot` opens a separate in-memory database, `ATTACH`es it, and copies every table out of `main`; only `deleteSnapshot` `DETACH`es and closes it, and the adapter is a page singleton whose `close()` never runs mid-shard. So a missed registration is one attached database holding a full copy of that module's indexed fixtures, retained for the rest of the shard — the cost already spelled out in the comment on `setupLocalIndexing`'s `after` hook.

**Enumerated against the tree.** 131 test files reference `withCachedRealmSetup`. Brace-matching their `module(...)` blocks gives 40 nested scopes; 23 of them have no `setupRealmCacheTeardown` of their own *and* reach a `withCachedRealmSetup` call through an ancestor scope's `beforeEach`:

```
5  acceptance/code-submode/create-file-test.gts
4  integration/resources/search-test.ts
4  integration/tools/show-card-test.gts
3  acceptance/theme-card-test.gts
2  integration/tools/copy-card-test.gts
1  each: code-submode-test.ts, code-submode/{editor,file-tree,inspector}-test, code-submode/spec-test.gts
```

Several are declared `function ()` or `() => {}` with no `hooks` parameter, so they cannot register without the signature change made here in two places.

Replaying `selectShardModules` over `tests/test-module-timings.json` at the CI `shardTotal: 20` puts at most 5 of these in one shard (shard 5, all from `create-file-test.gts`). SQLite's compile-time attach ceiling is 10 unless the build raises it — I did not confirm what `@sqlite.org/sqlite-wasm` ships — so nothing is failing today, which is why the suite is green. The margin is a handful of slots, and nothing reports when it narrows.

**The way out — let the registration declare the scope.** `setupRealmCacheTeardown` already resolves the right key in the right scope at definition time. Have it also publish that key for the duration of its scope, and have `withCachedRealmSetup` use the innermost published key instead of the running test's module name:

```ts
const scopeKeys: string[] = [];

export function setupRealmCacheTeardown(hooks, moduleCacheKey?) {
  let key = moduleCacheKey ?? getCurrentModuleCacheKey();
  // …existing dedupe…
  hooks.before(() => { scopeKeys.push(key); });
  hooks.after(async () => {
    scopeKeys.pop();
    await (await getDbAdapter()).deleteSnapshotsByPrefix(snapshotPrefixForModule(key));
  });
}

// in withCachedRealmSetup:
let moduleCacheKey = scopeKeys.at(-1) ?? getCurrentModuleCacheKey();
```

QUnit runs every `before` ahead of every `beforeEach` and nests them parent-before → child-before → child-after → parent-after, so the innermost entry is always the nearest registered scope. That gives:

- files registering only at the top level (this one, and the 10 above) — one snapshot per file, reclaimed by the single registration that exists;
- files registering per nested scope (`card-playground-test.gts`, `host-submode-test.gts`, `field-playground-test.gts`, which put the `withCachedRealmSetup` call and the registration in the same scope) — unchanged, since each scope publishes its own key;
- the three nested registrations added here — no longer needed.

**If re-keying is more than this change should carry:** make omission loud instead. `realmCacheTeardownRegistrations` is a `WeakMap<hooks, Set<prefix>>`, so it cannot be queried by key; making it a module-level `Set<string>` of registered prefixes would let `withCachedRealmSetup` throw when the prefix for the key it is about to use is unregistered. That converts all 23 into a failure with a fix in the message rather than silent retention.

**Scope:** pre-existing — the helper and its adopters predate this change — but now load-bearing: seven more modules move onto the mechanism here and three instances of the footgun are hand-patched, which makes this the natural place to fix it, and it gets more expensive once the remaining nine adopters convert. Non-blocking; nothing fails today.

---

packages/host/tests/integration/components/operator-mode-test.gts
```diff
@@ -831,7 +834,11 @@ module('Integration | operator-mode | basics', function (hooks) {
     );
   });
 
-  module('expand to full width', function () {
+  module('expand to full width', function (hooks) {
+    // The helper's realm-building beforeEach runs for these tests too, and caches
+    // under this module's name, which the outer teardown's prefix cannot match.
+    setupRealmCacheTeardown(hooks);
```
habdelra, line 840: [Claude Code 🤖] **Per-scope keying makes this file index its fixtures three times where `reuseIndexAcrossTests` indexed them once — and this scope pays a full build to serve a single test.**

**Why the count changes.** `reuseIndexAcrossTests` keeps `reusableIndex` in module-scoped state in `packages/host/tests/helpers/index.gts`, set by `setupLocalIndexing`'s `hooks.before` registered on the outer module. `before` runs once for the whole module tree, so the snapshot is captured on the first test and restored for every later test *including the nested ones* — one from-scratch build per file. `withCachedRealmSetup` keys on the module name of the running test, so each nested scope is a separate key whose first test is a cache miss.

**The numbers for this file.** 18 tests: 15 in the outer scope, 2 in `card with an error that has a last known good state`, 1 here. Builds go from 1 to 3, plus three `exportSnapshot` calls — each opens a fresh in-memory database and `INSERT … SELECT`s every row of every table in `main`. This scope is a pure loss: it builds the fixture set from scratch, exports a snapshot, never restores it (there is no second test), and deletes it in `after`.

`recents section` in the card-chooser module does amortize — 8 tests behind one build — but it too is a build the option did not charge.

**The fix.** The scope-stack change described on `setupRealmCacheTeardown` in `operator-mode/setup.gts` removes the extra keys and these three registrations together, and is where I would put the effort. Locally, `expand to full width` is a pure grouping with no hooks of its own; folding its one test into the outer scope also restores the single build, at the cost of the name.

Either way, worth a measurement before merging: the stated win is wall-clock, and this file spends part of it back.

**Aside on the comment itself:** the same two lines now appear verbatim in three places. If the scope-stack change lands they all go; if not, one sentence in the helper with a pointer from each site would keep them from drifting apart.

**Scope:** regression introduced here, non-blocking.

---

packages/host/tests/integration/searchable-search-doc-test.gts
```diff
@@ -55,12 +57,8 @@ module('Integration | searchable search doc', function (hooks) {
...
   });
 
-  // Every test in this module builds the same fixtures and only reads the
-  // search docs they produce, so the realm is indexed once and restored for
-  // each subsequent test.
-  setupLocalIndexing(hooks, {
-    reuseIndexAcrossTests: 'searchableSearchDoc',
-  });
+  setupLocalIndexing(hooks);
+  setupRealmCacheTeardown(hooks);
```
habdelra, line 61: [Claude Code 🤖] **This is the shape the mechanism is designed for, and it is correct — recording why, plus one caveat that got dropped on the way.**

43 tests, no nested modules, one realm, one `withCachedRealmSetup` call site. Every test resolves `getCurrentModuleCacheKey()` to the same module name, so it is one from-scratch build for 43 tests, and this single registration's prefix matches the only snapshot the module ever creates. It stops being true the moment a nested `module(...)` is introduced here — see the thread on `setupRealmCacheTeardown` in `operator-mode/setup.gts`.

**The caveat that got dropped.** The comment that moved down to the call site lost "and only reads the search docs they produce". That clause was the safety argument, not decoration: identical fixtures make the cache *possible*, no test mutating them is what makes it *sound*. Worth carrying back into the new comment.

**Small extra cost, worth knowing about.** With the option gone, `reusableIndex` is undefined for this module, so `setupLocalIndexing`'s `beforeEach` now takes the `else` branch and runs `dbAdapter.reset()` before every test; `withCachedRealmSetup` then runs `importSnapshot`, which `DELETE`s and refills every table in the snapshot. Each cached test now wipes the tables twice where the option's restore path wiped them once. Cheap next to an index build, but the comment on that `beforeEach` in `helpers/index.gts` — "Subsumes reset(): a precondition rather than a property" — is now true only of the option's path, and will read as stale once the option is gone.

**Scope:** confirmation, plus one non-blocking comment fix.

---

habdelra (commented):
[Claude Code 🤖] This change exists to buy wall-clock, so I reviewed it against the two questions that decide whether it does: do the converted modules still index their fixtures exactly once, and is every snapshot they create reclaimed?

**Bottom line: no correctness problems in the seven converted modules — `searchable search doc` is a clean win and the five operator-mode consumers are converted in the one right place. Two things are worth settling before merge, neither blocking. `operator-mode | basics` now indexes its fixtures three times where the option it replaces indexed them once, because the cache key follows the running test's module rather than the scope that owns the setup. And the per-scope teardown registration hand-applied three times here is a contract 23 nested scopes already in the suite quietly violate.**

## What lands right

**Wrapping the callback beats inferring a capture point.** `reuseIndexAcrossTests` captures right after the first `realm.start()`, which forced a whole sub-mechanism for modules that build more than one realm: `captureIndexManually`, `captureReusableIndex()`, a `realmURLs` set recorded at capture, and two distinct error messages telling you which of the two ways in you hit. `withCachedRealmSetup` captures when the callback returns, so a multi-realm setup is just a bigger callback. That is a real reduction in moving parts, not a relocation.

**The silent-empty-index failure cannot happen on the new mechanism.** The loud guards in `setupTestRealm` exist because under `reuseIndexAcrossTests` a realm missing from the snapshot mounts with `skipBootIndex` and serves nothing indexed — searches come back empty and absence assertions pass. Under `withCachedRealmSetup` a realm missing from the snapshot has no `realm_generations` row, so `Realm.#startup`'s `isNewIndex()` is true and it indexes normally (`packages/runtime-common/realm.ts`, `packages/runtime-common/index-writer.ts`). Those guards become unnecessary rather than merely unreplaced — worth defending if anyone asks why they were not ported.

**Converting inside `setupOperatorModeTests` rather than at five call sites** is right: the five modules build identical fixtures through one helper, so the helper is the only place the decision belongs.

## Recommendations

1. **Make the teardown scope-aware instead of hand-registered.** Sketch and the 23-scope enumeration are in the thread on `setupRealmCacheTeardown` in `operator-mode/setup.gts`. It also removes recommendation 2.
2. **Weigh the extra from-scratch builds in `operator-mode | basics`.** Numbers in the thread on `expand to full width`.
3. **Give `withCachedRealmSetup` the doc block that `reuseIndexAcrossTests` carries, before the option is deleted.** `reuseIndexAcrossTests` documents about forty lines of hard-won conditions — that a module whose tests are *about* indexing cannot use it, the worked `Integration | Store` example where 71 of 73 tests are indifferent and two assert on module-rebuild bookkeeping, why identical fixtures are necessary but not sufficient. `withCachedRealmSetup` has no doc comment at all, and the same class of caveat applies to it: once a `realm_generations` row exists, `#startup` takes neither the `isNewIndex` nor the `fullIndexOnStartup` branch, so nothing the boot index does beyond filling those tables has happened.

   There is a second caveat that belongs only to the new mechanism. `TestRealmAdapter`'s constructor stamps every fixture with `unixTime(Date.now())` (`packages/host/tests/helpers/adapter.ts`), so a restored test runs a freshly built adapter reporting the current second against an index still holding the first test's stamps. Nothing compares them at boot, so it is latent — but `discover-invalidations.ts` treats `lastModified !== indexEntry.lastModified` as an invalidation, so any test that provokes a from-scratch pass invalidates the entire fixture set and then races the reindex. That is precisely what the next module to trip it should be able to read rather than rediscover. Deleting the option should move its documentation, not just its code.

## Adjacent, out of scope

- **`integration/tools/copy-and-edit-test.gts` has a live cache-key collision.** It calls `withCachedRealmSetup` twice inside one `hooks.beforeEach` in one module scope, both without an `additionalKey`, so both resolve to the same snapshot name. The second call finds `hasSnapshot` true, imports the snapshot the first call just exported — `DELETE`ing and refilling every table in `main` — and returns without exporting, so `otherRealmURL` is never cached and re-indexes on every test. It works today only because nothing writes between the two calls; anything that did would be silently discarded. Fix is one `withCachedRealmSetup` around both realms, the shape `operator-mode/setup.gts` uses.
- **The `additionalKey` overload of `withCachedRealmSetup` has zero call sites in the suite.** It is the only way to tell two cache sites in one scope apart, so the collision above is what its absence looks like. Either use it there or drop the overload.
- **CI on this head.** The merged host report covers 117 tests in 6m25s against 4,237 in about two hours on the previous head — the shape of a partial re-run rather than a full suite, and the Percy status is reporting missing snapshots to match. Worth a clean full run before merge so the green means what it looks like.

---

packages/host/tests/integration/components/operator-mode/setup.gts
backspace: [Claude Code 🤖] Verified the mechanism and agreed on all of it — the enumeration matches what I hit twice by hand (host-submode needed four registrations, operator-mode-acceptance another four), which is exactly the symptom of a contract nothing enforces.

I'm not taking the scope-stack change in this PR, for one reason beyond scope: it converts an omission from "leaks a snapshot" into "shares the ancestor's snapshot", and nested scopes that build *different* fixtures do exist — host-submode's three nested modules each build their own realm with different contents. Sharing a key there means restoring the wrong fixtures, and the only thing that saves it is the mtime invalidation rebuilding the index. So the scope-stack change is safe *because* invalidation isn't suppressed, which couples it to a second decision: a separate review comment proposed suppressing that invalidation by pinning fixture mtimes, and doing both would be silently wrong. That pairing deserves its own PR and its own CI verdict rather than riding along here.

Recorded in CS-12514 with the mechanism, the 23-scope count, the caveat above, and your note that it would remove the hand-added registrations from #5738, #5737 and this PR. The loud-omission variant is captured too — worth noting it can't land alone, since it would fail 23 existing scopes on day one.

One correction to the framing: pre-existing, yes, but I made it worse in a specific way — three hand-patched registrations here plus eight more across #5738 and #5793 are all instances of the same footgun, so the case for fixing the keying is stronger than "natural place".

packages/host/tests/integration/components/operator-mode-test.gts
backspace: [Claude Code 🤖] You asked for a measurement before merging, so here it is — merged junit, this PR against main, per scope:

| scope | main | this PR |
| --- | --- | --- |
| `operator-mode \| basics` (18 tests) | 1.13m | 0.99m |
| ↳ `card with an error…` (2 tests) | 0.10m | 0.31m |
| ↳ `expand to full width` (1 test) | 0.03m | 0.22m |
| **file** | **1.26m** | **1.51m** |

So the claim holds exactly, including which scope is the pure loss: `expand to full width` spends 0.19m building fixtures and exporting a snapshot it never restores, for one test.

Two things it doesn't hold for, from the same run: `operator-mode | card chooser` goes 2.74m → 2.25m, its 8-test `recents section` amortising as you predicted, and `searchable search doc` 1.43m → 1.34m. Net across the three, this PR is about 0.33m ahead — so it's positive overall, with one file paying.

I'd rather not fold `expand to full width` into the outer scope. It buys back 0.19m, but the module name is what the memory-baseline gate and the timings file key on, so dropping it costs a baseline entry and a timings row for a saving the scope-stack change would deliver anyway across every affected file. Both the regression table and that reasoning are in CS-12514.

On the aside: agreed, and the two other instances are in `resources/search` and `store-test`, which this PR doesn't touch — so de-duplicating them here would widen the diff for text that disappears if the scope-stack change lands. Captured in the issue alongside it.

packages/host/tests/integration/searchable-search-doc-test.gts
backspace: [Claude Code 🤖] Right on both counts, and the dropped clause is fixed in 36249b37b0.

You're correct that it was the argument rather than decoration, so the comment now says both halves and why each carries weight: identical fixtures make the cache possible, and no test mutating them is what makes it sound. I'd rewritten it while moving it to the call site and kept only the half that explains *why it works*, losing the half that explains *why it's allowed*.

Also agreed that this file's correctness is conditional on staying flat — one call site, one key, one snapshot, and the single registration's prefix matches it. That stops being true the moment someone adds a nested `module(...)` here, which is the thread on `setupRealmCacheTeardown`.

lukemelia (approved):

PR #5773, opened 2026-08-13
Add the font FileDef family with a live specimen and metadata

lukemelia:
High-quality atom / fitted / embedded / isolated rendering for the font FileDef families — WOFF2, WOFF, TTF, OTF (CS-12242).

## What this adds
Web fonts (WOFF2/WOFF) and desktop fonts (TTF/OTF) share one renderer and one metadata shape, because once the container is unwrapped they are the same sfnt tables. The four shared shells mount a single live FontFace specimen for every subclass, and only the labeled kind differs per extension.

- **`font-meta-extractor.ts`** — reads name / OS-2 / head / maxp / fvar facts and outline technology (TrueType vs PostScript-CFF) from bare sfnt and from WOFF (per-table zlib via `DecompressionStream`), and the outline flavor from a WOFF2 header. A non-font throws `FileContentMismatchError` so the extract falls back to a plain FileDef.
- **`FontMetadataField`** — carries family / subfamily / full / PostScript names, foundry/vendor, weight, width, glyph count, em grid, outline type, and variation axes. The isolated inspector already mounts it via `<@fields.fontMetadata />`.
- **`FontSpecimen`** — renders the real linked face into each format: a showcase glyph pair (fitted), a name plus pangram (embedded), and a waterfall with a character map (isolated).
- **`Woff2Def` / `WoffDef` / `TtfDef` / `OtfDef`** — route `.woff2` / `.woff` / `.ttf` / `.otf` to the family.

## Metadata coverage
TTF, OTF, and WOFF index with full metadata. WOFF2 tables are Brotli-compressed and no browser exposes a Brotli decoder, so a WOFF2 indexes with the outline flavor only — but its **live specimen renders fully**, since the browser decodes WOFF2 natively.

## Tests
- Unit test builds synthetic sfnt / WOFF / WOFF2 fixtures and covers name/OS-2/head/maxp/fvar reads, the CFF-vs-TrueType split, WOFF zlib inflation, WOFF2 header-only reads, the non-font mismatch, and out-of-bounds degradation.
- Acceptance test drives extract → index → isolated render against a synthetic TTF and asserts the live specimen and stored metadata.
- The existing `experiments-realm` `FileDefFormatPreview` cards for the four extensions now resolve to these classes automatically.

### DIFF
--- a/packages/base/file-formats/metadata-fields.gts
+++ b/packages/base/file-formats/metadata-fields.gts
@@ -24,6 +24,7 @@ import RulerIcon from '@cardstack/boxel-icons/ruler';
 import TagIcon from '@cardstack/boxel-icons/tag';
 import KeyboardMusicIcon from '@cardstack/boxel-icons/keyboard-music';
 import TagsIcon from '@cardstack/boxel-icons/tags';
+import TypographyIcon from '@cardstack/boxel-icons/typography';
 
 import FileCodeIcon from '@cardstack/boxel-icons/file-code';
 import FileInfoIcon from '@cardstack/boxel-icons/file-info';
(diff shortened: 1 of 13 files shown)


packages/base/font-meta-extractor.ts
```diff
@@ -0,0 +1,510 @@
...
+}
+
+// WOFF header is 44 bytes; the table directory is 20-byte records. A table whose
+// compressed length equals its original length is stored raw; otherwise it is
+// zlib-compressed and inflated on demand.
+function woffTableSource(bytes: Uint8Array): TableSource {
+  let view = new DataView(bytes.buffer, bytes.byteOffset, bytes.byteLength);
+  let flavor = view.getUint32(4);
+  let numTables = view.getUint16(12);
+  let directory = 44;
```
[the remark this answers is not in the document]
lukemelia: Fixed in 4226ff8d3a. `woffTableSource` now checks for the full 44-byte header up front and throws `FileContentMismatchError` ("WOFF header is truncated") when the stream is shorter, so a truncated `wOFF` degrades to the plain FileDef instead of throwing a RangeError out of the extract. Added a unit test with a 20-byte `wOFF` stream.

packages/base/font-meta-extractor.ts
```diff
@@ -0,0 +1,510 @@
...
+    case WOFF_SIGNATURE:
+      return woffTableSource(bytes);
+    case WOFF2_SIGNATURE:
+      return woff2TableSource(bytes);
+    case SFNT_TRUETYPE:
+    case SFNT_TRUE:
+    case SFNT_TYP1:
+    case SFNT_OTTO:
+    case SFNT_TTCF:
+      return sfntTableSource(bytes);
```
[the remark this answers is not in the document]
lukemelia: Fixed in 4226ff8d3a. `ttcf` now routes to a dedicated header-only `collectionTableSource` that exposes the flavor and no tables, rather than `sfntTableSource` — so the reader recognizes the collection as a font (matching the container comment: "reports as a font but does not walk into") without misreading the TTC version/face-count as `numTables` and a table record. Added a unit test asserting a `ttcf` header reads as a TrueType font with no name/glyph tables.

packages/base/font-meta-extractor.ts
```diff
@@ -0,0 +1,534 @@
...
+    let stream = new Blob([bytes as BlobPart])
+      .stream()
+      .pipeThrough(new DecompressionStream('deflate'));
+    return new Uint8Array(await new Response(stream).arrayBuffer());
+  } catch {
+    return undefined;
+  }
+}
+
+function sfntTableSource(bytes: Uint8Array, headerOffset = 0): TableSource {
```
habdelra: [Claude Code 🤖] **Non-blocking (cleanup / follow-up).** `headerOffset` is dead flexibility. The only caller is `tableSourceFor` at the bottom of this file, which invokes `sfntTableSource(bytes)` with no second argument, so `headerOffset` is always `0` and the `headerOffset + 4` / `headerOffset + 12` arithmetic never moves off the fixed sfnt header. It reads as a leftover from walking into a TTC face at a computed offset — the path `collectionTableSource` now handles instead by exposing no tables. Dropping the parameter and inlining the literal offsets (`getUint32(0)`, `getUint16(4)`, `directory = 12`) keeps the reader honest about what it actually supports. Not worth a revision on its own; fold it in if you touch this function again.

---

packages/base/font-meta-extractor.ts
```diff
@@ -0,0 +1,534 @@
...
+  };
+}
+
+// A TrueType Collection packs several faces behind a `ttcf` header, which is not
+// an sfnt table directory — its bytes after the tag are a version and a face
+// count, not `numTables` and table records. As the container comment says, this
+// pass reports a collection as a font (the flavor names it) but does not walk
+// into a face, so it exposes no tables rather than misreading the TTC header as
+// one.
+function collectionTableSource(bytes: Uint8Array): TableSource {
```
habdelra, line 224: [Claude Code 🤖] **Non-blocking (minor correctness, niche).** `collectionTableSource` sets `flavor` to the container signature (`getUint32(0)` = `ttcf`), so `outlineTypeFor` takes the `source.flavor === SFNT_TTCF` branch and reports `TrueType` for *every* collection. A TrueType Collection can also carry CFF/PostScript faces (an "OpenType Collection"), which would then be mislabeled.

The blast radius is small and I don't think it blocks: `.ttc`/`.otc` aren't in the extension→codeRef map, so this only fires when a file named `.ttf`/`.otf`/`.woff`/`.woff2` actually contains a `ttcf` signature (a misnamed file), and it only affects the displayed `outlineType` string — no crash, no bad index. It's worth noting because it's an asymmetry with WOFF2: a WOFF2's outline type *is* derived correctly, because its header's `flavor` field is the wrapped sfnt version, whereas the `ttcf` signature carries no outline flavor of its own.

If you want the label honest without violating the "report as a font but don't walk into a face" rule, the `TTCHeader` stores `offsetTable[0]` (the offset to the first face's sfnt header) at byte 12 — the same field the unit fixture writes as "offset to the first face". Reading the 4-byte sfnt version at that offset would name the flavor without touching any table directory. Otherwise this is fine to leave with the current comment; flagging it so the next reader knows the label here is a signature default, not a read of the faces.

---

packages/base/file-formats/metadata-fields.gts
```diff
@@ -910,6 +911,134 @@ export class MidiMetadataField extends FieldDef {
...
+  @field weightClass = contains(NumberField);
+  // Label derived from usWidthClass, set only when the face isn't normal width.
+  @field widthName = contains(StringField);
+  @field glyphCount = contains(NumberField);
+  @field unitsPerEm = contains(NumberField);
+  // 'TrueType' or 'PostScript/CFF'.
+  @field outlineType = contains(StringField);
+  @field isVariable = contains(BooleanField);
+  @field isItalic = contains(BooleanField);
+  @field isBold = contains(BooleanField);
```
habdelra, line 948: [Claude Code 🤖] **Non-blocking (clarity / confirmation).** `isItalic` and `isBold` are the only two extracted facts with no reader anywhere. The extractor derives them carefully — OS/2 `fsSelection` (italic bit 0, bold bit 5), falling back to `head` `macStyle` (bold bit 0, italic bit 1) when OS/2 is absent, with a dedicated unit test — and `pruneUndefined` deliberately preserves a genuine `false`. But nothing consumes them: this field's embedded template renders family/style/weight/width/glyphs/outline/axes/postscript/vendor/version and skips these two; `file-view-model.ts`'s hero facts surface style via `subfamilyName · weightClass`; and the isolated shell's `hasFontMetadata` checks only `familyName || fullName || postscriptName || vendorId`.

That's a reasonable design if they're meant to be query-only boolean facets (e.g. a "show me the italics" filter) — worth a one-line comment saying so, since the omission otherwise reads as an oversight. If they were meant to display, they're currently silently dropped. Either resolution is fine; I'm confirming intent so the next reader doesn't take the gap for a bug.

---

habdelra (commented):
[Claude Code 🤖]

**Lens.** I focused on the byte-level parser — the code that runs untrusted bytes through fixed-offset `DataView` reads during the index pass — plus how the four shells and the pre-existing view-model / isolated-shell consumers line up with the new `FontMetadataField` shape.

**Bottom line: no blocking issues.** The extractor degrades safely on malformed input, the container abstraction is clean, the stored metadata shape matches what the existing consumers already read, and the tests pin the contracts that matter. Both earlier bot findings (truncated WOFF header; TTC routed through the sfnt path) are fixed and I verified each against the checked-out branch. Three non-blocking notes inline.

**What lands right.**
- **The `TableSource` seam is the right abstraction.** The name/OS-2/head/maxp/fvar readers never learn which container the bytes came from, and the WOFF2 "flavor-only" degradation falls out of `getTable → undefined` rather than a special case in every reader. Adding a fifth container later would touch one function, not five.
- **Every fixed-offset read is length-guarded before the `DataView` access** — `OS/2 < 64`, `head < 46`, `maxp < 6`, plus name-record and table-directory bounds checks — and the extract-vs-mismatch split is principled: a non-font throws `FileContentMismatchError` so the framework falls back to a plain FileDef, while a real font with unreadable tables returns a partial result. The unit suite exercises both sides, including the out-of-bounds table offset and the truncated-WOFF path.
- **`DecompressionStream('deflate')` is the correct decoder for WOFF's per-table zlib** (RFC 1950, not raw RFC 1951), and the fixture builder deflates symmetrically, so the inflation test is a real round-trip rather than a shape assertion.
- **Reusing the shared `FontMetadataField` shape and the existing hero-fact wiring** (rather than a per-extension column spray) means `familyName` / `weightClass` / `foundry` / `glyphCount` read identically whether the bytes arrived as TTF, OTF, or WOFF; I checked the consumer field names against the extractor output and they line up.

**Recommendations (all non-blocking).**
1. `sfntTableSource`'s `headerOffset` parameter is dead — inline on `font-meta-extractor.ts`.
2. A `ttcf` collection is always labeled `TrueType`, even for a CFF collection — inline on `collectionTableSource`.
3. `isBold` / `isItalic` are extracted and persisted but never surfaced — confirm query-only intent — inline on `metadata-fields.gts`.

**Adjacent, out of scope.** The `.metadata-rows / .row / dt / dd` `<style>` block is now repeated across nine FieldDefs in `metadata-fields.gts`; `FontMetadataField` follows that established pattern, so this isn't this PR's to change — noting it as the spot a shared style block would pay off whenever someone next touches the file.

**CI.** Host and realm suites are green (0 failures per the status comment, and every Host Tests shard shows `success`). The only red mark is the Percy commit status reporting 3 new snapshots awaiting the usual human approval — expected for the new specimen UI, not a code defect.

---

habdelra (approved):

packages/base/font-meta-extractor.ts
lukemelia: Done in 3a1630c75a — dropped the `headerOffset` parameter and inlined the fixed sfnt offsets (`getUint32(0)`, `getUint16(4)`, `directory = 12`). You're right it was a leftover from the walk-into-a-face path that `collectionTableSource` now owns.

packages/base/font-meta-extractor.ts
lukemelia: Good call — fixed in 3a1630c75a. `collectionTableSource` now reads the first face's sfnt version from the offset the TTC header stores at byte 12 and uses that as the flavor, so `outlineTypeFor` names an OpenType (CFF) Collection as PostScript/CFF instead of defaulting every collection to TrueType — without touching any table directory. Falls back to the `ttcf` signature when the offset is out of range. Added an OpenType-Collection test asserting the CFF label.

packages/base/file-formats/metadata-fields.gts
lukemelia: Confirmed intent in 3a1630c75a: `isItalic`/`isBold` are query-only style facets — persisted so a search can filter for italic/bold faces, but deliberately not rendered since the specimen and inspector already convey style via `subfamilyName · weightClass`. Added a comment on the fields saying so, so the omission doesn't read as an oversight.

PR #5777, opened 2026-08-13
Media formats: render the Audio and MIDI families through the shared shells

lukemelia:
Closes CS-12238.

## What

The sampled-audio family (**MP3, WAV, M4A, Ogg/Opus, FLAC**) and the **MIDI** family predated the CS-12229 shared format shells and still carried their own standalone atom/fitted/embedded/isolated templates — a bare icon and name, with none of the shells' identity/facts/budgets/state handling and no use of the waveform, encoding, tags, and sequence metadata the extract pass already reads. This adopts both families into the shared-shell architecture, matching the Image, HTML, PDF, and Office families.

The whole data layer already existed and shipped — the extractors, the `MediaEncodingField`/`MediaTagsField`/`WaveformMetadataField`/`MidiMetadataField` metadata, and the view-model projection (`waveformBars`, `mediaUrl`, `durationSeconds`, waveform/midi hero facts). Even the shells were built anticipating these families (the embedded shell's `audio` and `sequence` shapes, the fitted shell's live waveform rail, the isolated inspector's encoding/waveform/tags groups). **Only the two renderers and their wiring were missing.**

## Audio — `AudioPreview`
- One `previewComponent`; the four formats come from `FileDef`'s shared shells.
- Reading formats mount a native `<audio controls>` player over a **waveform** drawn from the extracted amplitude envelope.
- A **fitted** cell shows the waveform + running time with **no** live transport (a grid of players is not a preview).
- A file whose envelope couldn't be decoded (over-budget FLAC, a stream that won't re-read) still renders: the family glyph and duration carry the cell.

## MIDI — `MidiPreview` (rendered distinctly)
A Standard MIDI File is symbolic performance data with no sound until synthesized, and no browser plays one from `<audio>`. So it draws **no player and no waveform**. Instead it presents the sequence itself:
- a **pitch band** — the piano's span with the piece's range shaded across it;
- a fact summary (tracks, notes, length, tempo, key, meter);
- the **General MIDI voices** the programs name (`program 40` → "Violin").

## Also
- Fixes the isolated inspector to read audio tags from the family's actual `tags` field (the shell/view-model previously referenced a `mediaTags` name no family declares), and adds a **MIDI sequence** group there.
- Gives a SMPTE-timed MIDI file a track/note hero fact when it has no seconds to clock.
- Removes the four now-unused `default-templates/audio-def-*` templates.

Unlike the Office and PDF families, the fitted rendering here **is** the waveform (real extracted data), not a rasterized poster — so this needs no thumbnail-contract (CS-12231) follow-up.

## Tests
- Updates the 5 existing audio acceptance tests (mp3/wav/ogg/m4a/flac) to assert the new shared-shell preview: the player mounts and the extracted duration renders.
- Adds a MIDI acceptance test (extract / render / index / mismatch fallback) built on an in-code minimal Standard MIDI File — no binary fixtures committed.

## Validation
- `ember-tsc --noEmit` (host `lint:types`): **0 errors**.
- Base template-lint: clean.
- Remaining eslint findings (`<template>` parse + `@field` decorators) are the pre-existing repo-wide baseline shared by every base `.gts`.
- I did **not** boot the browser to run the QUnit/acceptance suites locally — CI on this PR runs them.


packages/base/file-formats/audio-preview.gts
```diff
@@ -0,0 +1,308 @@
...
+        {{#if this.mediaUrl}}
+          <audio
+            class='audio-player'
+            src={{this.mediaUrl}}
+            controls
+            preload='metadata'
+            data-test-audio-player
+          >
+            <track kind='captions' />
+          </audio>
```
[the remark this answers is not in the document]
lukemelia: Fixed in e89165af71. Removed the empty `<track>` from the player — an arbitrary audio file ships no captions, so it only misled/warned. Since the a11y `require-media-caption` rule genuinely doesn't apply here (there are no captions to provide), it's disabled for that one element with a rationale comment.

packages/base/file-formats/file-shell-isolated.gts
```diff
@@ -256,6 +256,11 @@ export class FileIsolatedShell extends GlimmerComponent<FileIsolatedShellSignatu
     return Boolean(t?.trackTitle || t?.artist);
   }
 
+  get hasMidi() {
+    let m = this.args.model?.midi;
+    return Boolean(m?.trackCount || m?.noteCount || m?.durationSeconds);
+  }
```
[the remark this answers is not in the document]
lukemelia: Good catch — fixed in e89165af71. `hasMidi` now shows the MIDI sequence group whenever any extracted sequence fact is present (tempo map, key/meter, programs, pitch range, format, ppq, percussion, channels), not just track/note counts or duration — so a SMPTE-timed conductor track with 0 notes no longer hides its metadata.

packages/base/file-formats/midi-preview.gts
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  // The piece's pitch range as a fraction of the piano, or undefined when either
+  // endpoint isn't a note name we recognize.
+  get rangeBand(): { left: number; width: number } | undefined {
+    let range = this.midi?.pitchRange;
+    if (!range) {
+      return undefined;
+    }
+    let parts = String(range).split(/\s*[–—-]\s*/);
```
habdelra: [Claude Code 🤖]

**The pitch band silently won't draw when the piece's lowest note is below MIDI 12.** `rangeBand` splits `pitchRange` on `/\s*[–—-]\s*/`, and that character class includes the ASCII hyphen `-`. The extractor (`midi-meta-extractor.ts`, `noteName`) writes negative octaves as `C-1` — MIDI note 60 is `C4`, so pitches 0–11 land in octave `-1` — and joins the two endpoints with an en-dash `–` (U+2013, verified). So for a range like `C-1–G9` the split fires on both the hyphen inside `C-1` and the en-dash, yielding three parts → `parts.length !== 2` → `rangeBand` returns `undefined` → the `.pitch-range` div never renders. The `Range` chip still shows the text, so it fails halfway: text present, band gone.

Verified against the extractor's actual output format:

```
pitches 21 108 -> "A0–C8"    split=["A0","C8"]     bandDrawn=true
pitches 60 72  -> "C4–C5"    split=["C4","C5"]     bandDrawn=true
pitches 0 127  -> "C-1–G9"   split=["C","1","G9"]  bandDrawn=false
pitches 10 120 -> "A#-1–C9"  split=["A#","1","C9"] bandDrawn=false
```

The extractor always joins with the en-dash, so the fix is to match only the en/em dash and leave the hyphen to the note names:

```suggestion
    let parts = String(range).split(/\s*[–—]\s*/);
```

Class: regression (new code), **non-blocking** — sub-MIDI-12 pitches are effectively sub-audible, so ordinary audio-range music won't hit it, but experimental / synth / data-driven sequences can, and the fix is free.

---

packages/base/file-formats/file-view-model.ts
```diff
@@ -526,8 +535,11 @@ export function fileViewModel(
     // authority, with the EXIF tag already folded in at extract time.
     colorProfile: file.colorProfile,
     encoding: file.encoding,
-    mediaTags: file.mediaTags,
+    // `AudioDef` declares this field as `tags`; the older `mediaTags` name is
+    // kept as a fallback so a future family may use either.
+    mediaTags: file.tags ?? file.mediaTags,
```
habdelra: [Claude Code 🤖]

**Minor — the `mediaTags` fallback is only half-wired, so the "either name" the comment promises can't actually render.** This projects `mediaTags: file.tags ?? file.mediaTags` and the comment says the older `mediaTags` name is "kept as a fallback so a future family may use either." But the isolated inspector renders that group with a hardcoded field name — in `file-shell-isolated.gts`, `{{#if this.hasTags}} … <@fields.tags /> {{/if}}`, where `hasTags` reads the *projection* (`this.args.model?.mediaTags`) while the render reaches for the *raw field* `@fields.tags`. So a hypothetical family that declared its field as `mediaTags` would pass `hasTags` (the projection picks it up through the `?? file.mediaTags` branch) yet render an empty group, because `@fields.tags` wouldn't exist on it.

Today only `AudioDef` declares this field and it declares it as `tags` (grep confirms no family declares `mediaTags`), so this isn't a live bug — the `?? file.mediaTags` branch is currently dead. Two clean options: drop the dead fallback (the honest state: one family, one name), or, if you want to keep the door open, have the shell pick the field name dynamically rather than hardcoding `tags` in the template. Flagging so the comment and the code don't drift into a promise the shell can't keep.

Class: confirmation / follow-up, **non-blocking**.

---

packages/base/file-formats/midi-preview.gts
```diff
@@ -0,0 +1,435 @@
...
+        border-radius: 5px;
+        border: 1px solid var(--border);
+        overflow: hidden;
+        /* An octave lattice: the twelve-semitone repeat reads as a keyboard. */
+        background-color: var(--fd-stage, var(--muted, #eceef1));
+        background-image: repeating-linear-gradient(
+          90deg,
+          transparent 0 calc(100% / 88 * 7 - 1px),
+          var(--border) calc(100% / 88 * 7 - 1px) calc(100% / 88 * 7)
+        );
```
habdelra, line 353: [Claude Code 🤖]

**Cosmetic — the "octave lattice" comment doesn't match the stride the gradient actually draws.** The comment reads "An octave lattice: the twelve-semitone repeat reads as a keyboard," but the `repeating-linear-gradient` repeats every `100% / 88 * 7` — a 7-semitone (perfect-fifth) stride, not a 12-semitone octave. Across the 88-key span that paints ~12–13 vertical lines rather than the ~7 octave boundaries the comment implies. Nothing breaks — it's decorative and the `A0 / middle C / C8` axis carries the real orientation — but the divisor and the comment disagree. Either make it `100% / 88 * 12` to be the octave lattice the comment claims, or reword the comment to describe decorative striping.

Class: evergreen/cosmetic, **non-blocking**.

---

packages/host/tests/acceptance/audio-def/midi-audio-def-test.gts
```diff
@@ -0,0 +1,324 @@
...
+    assert.strictEqual(status, 'ready', 'render completed');
+
+    let preview = document.querySelector(
+      '[data-prerender] [data-test-midi-preview]',
+    );
+    assert.ok(preview, 'midi preview renders in the shared shell');
+    let player = document.querySelector(
+      '[data-prerender] [data-test-audio-player]',
+    );
+    assert.notOk(player, 'a symbolic MIDI file mounts no audio player');
```
habdelra, line 282: [Claude Code 🤖]

**Test coverage — the headline "program 40 → Violin" mapping isn't pinned by any assertion.** The fixture writes a program-change to GM voice 40 and the PR calls voice-naming out as a defining feature of `MidiPreview`, but this test asserts only that `[data-test-midi-preview]` and `[data-test-midi-duration]` render — not that the voice list actually resolves to "Violin". The `.voices-list` items carry no `data-test` hook, so there's nothing to query for it today.

Since the whole point of the feature is that `program 40` becomes a human name (and it's a 0-based index into a 128-entry table — exactly the kind of thing that silently goes off-by-one), a small addition would guard it: a `data-test-midi-voices` hook on the `.voices` block plus an assertion that its text contains `Violin`. That turns the extract→index→render→name chain the fixture already exercises into something a regression would trip.

Class: test-coverage follow-up, **non-blocking**.

---

habdelra (commented):
[Claude Code 🤖]

This review focused on two things: whether the two new renderers stay faithful to the data the extract pass actually produces, and whether unhooking audio/MIDI from their standalone templates onto the shared shells leaves anything stranded.

**Bottom line: a clean, well-scoped migration with no blocking issues.** The data layer already shipped, and this lands the renderers onto it without orphaning templates or leaving dangling references. Four non-blocking notes below; the most concrete is a pitch band that silently won't draw for sub-octave note ranges.

**What lands right**
- The audio/MIDI family split is honest. `MidiDef` deliberately stays off `AudioDef` so a sequence never advertises sample-rate / bit-depth / waveform fields it can't fill, and the General-MIDI mapping is correct: I verified `GM_INSTRUMENTS.length === 128`, `[0] === 'Acoustic Grand Piano'`, `[40] === 'Violin'`, `[127] === 'Gunshot'` — a 0-based index that matches the raw `STATUS_PROGRAM_CHANGE` byte the extractor pushes, with the percussion channel excluded upstream.
- Removing the four `audio-def-*` templates is clean. Grep finds no remaining imports of them and no stray `.audio-isolated__*` / `audio-def-*` selectors anywhere in `packages/`. The old isolated template was icon + name + duration + a native player (with the very empty `<track kind='captions'>` this PR later drops); the shell path preserves the player and the extracted duration and adds the inspector, facts, budgets, and state handling.
- Replacing the per-template `formatDuration` with the shared `formatClock` de-duplicates the clock formatting; `formatClock(2) === '0:02'`, matching the fixtures.
- `hasMidi` reading `format != null` (so format `0` counts) rather than a truthy check is the right instinct for the SMPTE-timed conductor-track case, and the hero-fact fallback to tracks/notes when there are no seconds to clock is consistent with it.
- Both earlier Copilot findings (empty caption `<track>`, `hasMidi` too narrow) are genuinely resolved, not just acknowledged.

**Open discussion:** both existing review threads are resolved and outdated; nothing left open there.

1. `midi-preview.gts` — the pitch-band split class swallows the hyphen in `C-1`, so the band silently won't render for ranges whose low note is below MIDI 12. Regression, cheap fix (split on the en/em dash only). Non-blocking.
2. `file-view-model.ts` — the `mediaTags: file.tags ?? file.mediaTags` fallback is half-wired against the shell's hardcoded `<@fields.tags />`; the "either name" the comment promises can't actually render. Currently dead code, not a live bug. Non-blocking.
3. `midi-preview.gts` — the `.pitch-band` "octave lattice" comment says twelve-semitone but the gradient strides every 7. Cosmetic.
4. `midi-audio-def-test.gts` — the "program 40 → Violin" voice mapping (the feature's headline) isn't asserted; add a `data-test` hook + a `Violin` assertion so an off-by-one can't pass silently. Follow-up.

Nothing here blocks; approve/request-changes is the human reviewer's call.

---

habdelra (approved):

packages/base/file-formats/midi-preview.gts
lukemelia: [Claude Code 🤖]

Fixed in 9aed94bcdf. The split now matches only the en/em dash the extractor joins with, leaving the ASCII hyphen to negative-octave note names — so a `C-1–G9` range splits into exactly two parts and the band renders. A comment on the split explains why the hyphen is excluded.

packages/base/file-formats/file-view-model.ts
lukemelia: [Claude Code 🤖]

Fixed in 9aed94bcdf. Dropped the dead `?? file.mediaTags` branch — the projection now reads only the `tags` field, and the comment now states the real constraint: the isolated shell renders the group via `<@fields.tags />`, so the projection must come from that same field. If a future family wants a different field name, the shell lookup is the thing to generalize then.

packages/base/file-formats/midi-preview.gts
lukemelia: [Claude Code 🤖]

Fixed in 9aed94bcdf. The gradient now strides every 12 semitones, so it draws the octave lattice the comment describes (~7 boundaries across the 88-key span).

packages/host/tests/acceptance/audio-def/midi-audio-def-test.gts
lukemelia: [Claude Code 🤖]

Added in 9aed94bcdf. The voices block now carries a `data-test-midi-voices` hook, and the isolated-preview test asserts the list renders and contains `Violin` — pinning the full extract → index → render → GM-voice-name chain for program 40 so an off-by-one in the 128-entry table can no longer pass silently.

PR #5780, opened 2026-08-13
Widen and correct the Synapse readiness gate in CI

lukemelia:
`waitForSynapse` (in `packages/matrix/scripts/register-matrix-users.ts`) polled a `HEAD` on the homeserver root with a hard **24 × 5s = 120-second** budget and no backoff. On a cold or loaded CI runner the image pull + database migrations can take longer, and when they did the gate threw — taking out the **entire test shard before a single test ran**.

- Polls **`/_matrix/client/versions`** instead — it returns 200 only once Synapse has finished migrating, a truer readiness signal than a root `HEAD` that can answer non-2xx mid-boot.
- Widens the budget to **60 × 5s (~5 min)** so a slow start is tolerated rather than fatal.

Test-harness timing only; no product code. `ember-tsc --noEmit` on `packages/matrix`: 0 errors.

Split out of #5779 (paired with the AI-assistant room-helper fix, now its own PR).

### DIFF
--- a/packages/matrix/scripts/register-matrix-users.ts
+++ b/packages/matrix/scripts/register-matrix-users.ts
@@ -67,20 +67,32 @@ const EXTRA_USERS: ExtraUser[] = [
 ];
 
 async function waitForSynapse(): Promise<void> {
-  const url = getSynapseURL();
-  const maxAttempts = 24;
+  const base = getSynapseURL().replace(/\/$/, '');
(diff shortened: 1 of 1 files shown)


backspace (approved):
I hope this helps with the failed-before-starting shards, if not maybe there’s a way to get visibility into what’s actually failing

habdelra (approved):

PR #5783, opened 2026-08-13
Single-source the fitted-format envelopes; pin boxel-ui's copy

lukemelia:
The filedef-fixtures `format-preview` harness rendered its fitted envelopes from a hand-copied list of the 16 specs plus 16 paired `.size-*` CSS blocks, and `boxel-ui` carries its own independent copy of the same data. Any edit to `FITTED_FORMATS` in runtime-common would silently drift both — the harness would render the wrong envelope boxes with no signal, which is worse than no harness because it gives false visual confidence.

Two fixes, matched to what each copy allows:

- **Harness: single-sourced.** Card code can import `@cardstack/runtime-common` — the host shims it as a platform module (`packages/host/app/lib/externals.ts`), and several experiments-realm cards already do. The harness now imports `FITTED_FORMATS` and sizes each envelope frame with an inline style derived from the spec (the same pattern as the host's fitted-format gallery), deleting the copied list and all 16 `.size-*` CSS blocks. Nothing is left to drift.
- **boxel-ui: parity-tested.** `boxel-ui/src/utils/fitted-formats.ts` stays a copy, because a standalone design-system package must not depend on runtime-common. `packages/realm-server/tests/fitted-formats-parity-test.ts` lifts that array literal out of the source text (the suite can't import a v2-addon source file) and asserts it deep-equals runtime-common's `FITTED_FORMATS`, so an edit to either list fails loudly.

Also rewords the SOURCES.md integrity-manifest note: the per-fixture hashes are a reference for manual audits (`shasum -a 256`), not something CI recomputes — the wording previously implied enforcement that does not exist.

## Test plan

- `TEST_FILES=./fitted-formats-parity-test pnpm test` in `packages/realm-server` — passing.
- `pnpm run lint` in `packages/experiments-realm` — passing.
- Verified via `content-tag` directly that the reworked `.gts` transpiles with all three `<template>` tags consumed, identical to the prior version.

### DIFF
--- a/packages/realm-server/tests/fitted-formats-parity-test.ts
+++ b/packages/realm-server/tests/fitted-formats-parity-test.ts
@@ -4,22 +4,11 @@ import { readFileSync } from 'node:fs';
 import { basename, join } from 'node:path';
 import { FITTED_FORMATS } from '@cardstack/runtime-common';
 
-// Two places in the repo carry a hand-maintained copy of the fitted envelope
-// specs, because neither can import runtime-common's FITTED_FORMATS: the
-// filedef-fixtures format-preview harness (a card module, loaded through the
-// realm) and boxel-ui (which must not depend on runtime-common). These tests
-// pin both copies — and the harness's paired .size-* CSS blocks — to the
-// canonical list so an edit to any of them fails loudly instead of letting
-// the copies drift.
+// boxel-ui carries a hand-maintained copy of the fitted envelope specs,
+// because as a standalone design-system package it must not depend on
+// runtime-common. This test pins that copy to the canonical FITTED_FORMATS
+// so an edit to either list fails loudly instead of letting the two drift.
 
-const HARNESS_PATH = join(
-  import.meta.dirname,
-  '..',
-  '..',
-  'experiments-realm',
-  'filedef-fixtures',
-  'format-preview.gts',
-);
 const BOXEL_UI_PATH = join(
   import.meta.dirname,
   '..',
@@ -30,10 +19,10 @@ const BOXEL_UI_PATH = join(
   'fitted-formats.ts',
 );
 
-// The copies live in modules this suite cannot import (a .gts card module and
-// a v2-addon source file), so we lift the array literal out of the source
-// text and evaluate it. The literals are plain data — no identifiers — so the
-// evaluation cannot reach anything else.
+// The copy lives in a v2-addon source file this suite cannot import, so we
+// lift the array literal out of the source text and evaluate it. The literal
+// is plain data — no identifiers — so the evaluation cannot reach anything
+// else.
 function extractArrayLiteral(filePath: string, declaration: string): unknown {
   let source = readFileSync(filePath, 'utf8');
   let match = source.match(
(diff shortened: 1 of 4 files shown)


packages/realm-server/tests/fitted-formats-parity-test.ts
```diff
@@ -0,0 +1,49 @@
...
+  'boxel-ui',
+  'src',
+  'utils',
+  'fitted-formats.ts',
+);
+
+// The copy lives in a v2-addon source file this suite cannot import, so we
+// lift the array literal out of the source text and evaluate it. The literal
+// is plain data — no identifiers — so the evaluation cannot reach anything
+// else.
```
habdelra, line 25: [Claude Code 🤖] Confirmation, non-blocking — the extraction approach is the right call, and I verified it empirically: lifting the literal out of the source text and evaluating it returns 16 specs that deep-equal `runtime-common`'s, and mutating a single width flips the comparison, so a real drift fails loudly rather than passing silently. A regex miss or a reformat that breaks the `^\]` anchor throws (either "could not find…" or a `new Function` syntax error), so there's no silent-pass failure mode.

One nuance on the *why*, for whoever reads this next: "this suite cannot import a v2-addon source file" understates the actual benefit. `boxel-ui/src/utils/fitted-formats.ts` currently has zero imports — it's pure data — so a direct `import` of just that file would in fact resolve. The stronger reason to lift the literal is the one the comment below states: it reads *only* the array, so it stays immune to whatever the boxel-ui module graph pulls in now or later. That's the durable justification; the "cannot import" framing is the weaker half.

---

packages/realm-server/tests/fitted-formats-parity-test.ts
```diff
@@ -0,0 +1,49 @@
+import QUnit from 'qunit';
+const { module, test } = QUnit;
+import { readFileSync } from 'node:fs';
+import { basename, join } from 'node:path';
+import { FITTED_FORMATS } from '@cardstack/runtime-common';
+
+// boxel-ui carries a hand-maintained copy of the fitted envelope specs,
+// because as a standalone design-system package it must not depend on
+// runtime-common. This test pins that copy to the canonical FITTED_FORMATS
+// so an edit to either list fails loudly instead of letting the two drift.
```
habdelra, line 10: [Claude Code 🤖] Adjacent, explicitly out of scope for this PR — just flagging what this guardrail does and doesn't cover so the next editor knows the boundary.

This test pins the `boxel-ui` data copy to `runtime-common`'s, which closes the drift path the PR set out to close. But the same 16 specs are encoded in two other places that this test does not reach:

- The actual fitted-rendering breakpoints — the `@container fitted-card (...)` queries in `boxel-ui/src/components/fitted-card/index.gts` and `basic-fitted/index.gts` (e.g. `/* Double Wide Strip 400x65 */`). Those breakpoints are arguably the real source of truth for how a fitted card renders; the data table has to agree with them, but nothing enforces it.
- The skill reference doc `packages/boxel-cli/plugin/skills/boxel/references/fitted-formats.md`, which lists the same specs as prose.

Neither is this PR's job and I wouldn't expand scope here — noting it only so the "nothing is left to drift" framing is understood as scoped to the data-array copies, not the CSS breakpoints or the docs.

---

habdelra (commented):
[Claude Code 🤖] Reviewed for one thing above all: does single-sourcing the harness change what it renders, and does the parity test actually catch drift rather than just look like it does. Checked out the head branch and verified both empirically.

**Bottom line: no blocking issues, no bugs found. This is a faithful, behavior-preserving refactor plus a real guardrail.** The two inline notes are both non-blocking (one confirmation, one out-of-scope flag).

**What lands right (verified, not assumed):**
- *No visual regression.* I diffed all 16 entries of the deleted `FORMAT_GROUPS` against `runtime-common`'s `FITTED_FORMATS` — every width, height, label, group name, and ordering is identical, so the matrix renders exactly the boxes it did before. The two copies were in sync at this moment; the change removes the path by which they could silently stop being.
- *The card-runtime plumbing is sound.* `frameSize` is a line-for-line twin of the host gallery's `setContainerSize`, and plain-function-as-helper (`{{frameSize spec}}`) is already how sibling experiments-realm cards work (`categoryStyle`, `setBackgroundImage`, both returning `htmlSafe`). Importing a runtime *value* from `@cardstack/runtime-common` is done by 25 experiments-realm cards already and is exposed to card code by the whole-namespace shim in `host/app/lib/externals.ts`, so it resolves both at render time and during prerender/indexing.
- *The parity test is not vacuous.* Replicating `extractArrayLiteral` against the real source lifts 16 specs that deep-equal `runtime-common`'s; a one-field mutation flips the result; a regex miss or reformat throws. There's no silent-pass failure mode. `import.meta.dirname/filename` is fine — 176 realm-server tests already use it.
- *The design rationale holds.* `boxel-ui` can't depend on `runtime-common` without inverting the layering, and `runtime-common` deliberately treats `@cardstack/boxel-ui` as an external/shimmed module (no source-level import of it), so re-exporting from one into the other isn't the clean win it might look like. Two copies + a parity bridge is the right shape.

**SOURCES.md** — the reworded integrity-manifest paragraph is accurate (nothing recomputes the hashes; it's a manual-audit reference) and reads cleanly.

The event here is COMMENT, not an approval — that call is yours; the words above are the verdict.

---

PR #5785, opened 2026-08-14
Add StlDef and ThreeMfDef 3D-model FileDefs on the shared file-format shells

FadhlanR:
## What

Adds `.stl` and `.3mf` support as base-realm `FileDef` subclasses, built on the shared file-format shells' `model` family.

- **`ThreeDModelDef`** — thin family base under `FileDef`: supplies `static previewComponent = Model3DPreview` (a live WebGL orbit viewer, loaded from a CDN at client render time) and the `model3d` metadata field the isolated shell already renders. Inherits all four shared format shells.
- **`Model3dMetadataField`** — the 3D family's metadata field in `file-formats/metadata-fields.gts`. The shells were already wired to render `<@fields.model3d />` and gate the fitted hero on `model3d.triangles`; this fills in the field they expected.
- **`StlDef` / `ThreeMfDef`** — leaves that add only their header-only `extractAttributes`, projecting each format's facts onto `model3d` via `model3dAttributes`. STL uses a bounded ASCII/binary sniffer; 3MF unzips the OPC package (`fflate`) and reads the model prologue + slicer config.
- Registers `.stl`/`.3mf` in the FileDef code-ref map; threads the realm file-size limit through the index-time extractor; adds `fflate`.

The `.stl`/`.3mf` fixtures already in `filedef-fixtures` now resolve to these subclasses instead of the generic `FileDef`.

Supersedes #5658 — that PR predated the shared file-format shells and rolled its own four format templates + per-format inspector fields; this rebuilds the same feature on the shells the rest of the FileDef families now use.

## Design notes

- **One `model3d` field (not per-format fields).** The isolated shell renders exactly one 3D slot and no family overrides `isolated`, so both formats project onto a single `Model3dMetadataField` — the shell-native shape.
- **Header-only extraction.** True geometry (transform-correct dimensions, full triangle counts) comes from the live viewer, not the index. A value is present only when it sits cheaply in the header: a binary STL's facet count, a slicer 3MF's configured face/part counts, materials, provenance.
- **Known limitation:** the isolated `3D model` section gates on `meshes || triangles`. An STL reports `meshes: 1` (a solid) so it always shows; a config-less 3MF reports no geometry counts, so its section hides (generic name/type/size rows still show).

## Test plan

- `ember-tsc` clean; prettier clean; template-lint clean.
- Unit tests for the pure parsers (`model-meta-extractor-test.ts`) and an end-to-end file-extract acceptance test (`model-file-extract-test.gts`) asserting the `model3d` projection.
- Pending on a dev stack: index the `filedef-fixtures` corpus and confirm the stl / three-mf FormatPreview pages render with zero error rows.

### DIFF
--- a/packages/base/file-formats/metadata-fields.gts
+++ b/packages/base/file-formats/metadata-fields.gts
@@ -27,6 +27,7 @@ import TagsIcon from '@cardstack/boxel-icons/tags';
 
 import FileCodeIcon from '@cardstack/boxel-icons/file-code';
 import FileInfoIcon from '@cardstack/boxel-icons/file-info';
+import CubeIcon from '@cardstack/boxel-icons/cube';
 
 import {
   Component,
@@ -534,7 +535,9 @@ export class MediaEncodingField extends FieldDef {
             >{{@model.container}}</dd></div>
         {{/if}}
         {{#if @model.videoCodec}}
-          <div class='row'><dt>Video</dt><dd class='mono'>{{@model.videoCodec}}</dd></div>
+          <div class='row'><dt>Video</dt><dd
+              class='mono'
+            >{{@model.videoCodec}}</dd></div>
         {{/if}}
         {{#if @model.audioCodec}}
           <div class='row'><dt>{{if @model.videoCodec 'Audio' 'Codec'}}</dt><dd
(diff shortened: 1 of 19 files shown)


FadhlanR:
Closing — folded this work directly into #5658 by force-pushing the shells-aligned rebase onto that branch, so the original PR updates in place instead of a new one. All discussion + the screen recording live on #5658.

PR #5791, opened 2026-08-15
Cover every function the BXL registry exposes, and anchor the date functions to UTC

habdelra:
BXL exposes a large function surface — jq's builtins, a curated Excel/formulajs
subset with four lazily-chunked families, the validator.js helpers, and the
authorization builtins — and the project asks that every one of them be
exercised by tests. Until now a good deal of that surface had never been called
by anything.

The original framing was to invoke every function through a `computeVia` on a
real card. Proving BXL works on cards and proving each function computes
correctly are different jobs, though, and only the first one needs a card. The
host integration suites already cover the first: all three syntax modes driving
`computeVia`, `linksTo` and query-backed `linksToMany` traversal,
`{ as: FieldDef }` materialization, null tolerance, Excel error sentinels,
cyclic graphs, and correct search docs across incremental reindex. Running 800
Excel functions through card fields on top of that would cost a lot of test time
to re-prove the plumbing, once per function.

So the exhaustive matrix lives in the package's own node suite against plain
JSON, where a case costs microseconds, and the card side gets the one thing it
alone can prove.

## The coverage suite

`packages/bxl/tests/unit/function-coverage-cli.ts` reads the list of functions
to cover out of the **resolved builtin registry** — every `NAME/arity` entry
across all libraries — and fails if any of them has no case. There is no
hand-maintained function list, so adding a builtin fails the suite until someone
covers it. The gate also rejects a case covering a name the registry no longer
exposes, which turns a future rename into a test failure rather than silent
coverage loss.

It currently reports:

```
BXL function coverage: 814 of 815 exposed functions invoked across 859 cases,
1 unreachable, 20 known defect(s)
```

and runs in under a second, so it costs nothing to keep in the normal suite.

Coverage is credited by **observed invocation**, not by what a case declares.
Every library's filters are wrapped in recorders before the registry resolves
them, so a case that claims `ROUND/2` but never reaches it fails, as does one
that only names it inside a branch that never runs. Both dispatch paths are
wrapped — the streaming filter and the bare filter that the compiled-scalar fast
path calls directly — so instrumenting does not push evaluation off the route
production takes.

Expected values were derived from each function's documented semantics *before*
running anything, rather than recorded from current behavior; a matrix that
photographs the implementation proves nothing. Where a closed form was available
it was used in preference to a table value — the standard normal CDF at 0 is
exactly ½, Beta(1,1) is uniform, `IPMT + PPMT` equals `PMT`, `EFFECT` and
`NOMINAL` invert each other — so the expectation is independently checkable.

One exposed name is unreachable and documented as such: `modulemeta/0` is a
reserved tokenizer keyword, so naming it is a parse error before dispatch, and
the filter itself only raises "not implemented".

## Time zones

A date case is evaluated under five zones — UTC, a negative whole-hour offset, a
half-hour offset, and both extremes of the offset range — and must give the same
answer in all of them. CI's zone is not a developer's zone, so a date case that
holds in one is a flaky test waiting to happen.

Building that guard turned up that the Excel date functions genuinely were
zone-dependent, which is a product bug and not a test-authoring problem:
`EOMONTH(DATE(2026,2,5),1)` returned 46112 under UTC and 46111 under
`America/Los_Angeles`. An Excel serial names a calendar day, so this matters
well beyond CI — a card's computeds are evaluated server-side during indexing,
in UTC, and again in the viewer's browser, in their zone, so the value baked
into the search doc and the prerendered HTML could disagree with what the viewer
sees, and two viewers in different zones could see two different cards from one
instance.

`TODAY`, `WEEKDAY`, `ISOWEEKNUM`, `EDATE` and `EOMONTH` built their epoch with a
local-time `Date` and read local accessors; a date string naming no zone was
resolved against the host. The local form is also self-inconsistent, because the
serial epoch materializes at a zone's 1899 offset while the result is read at
its modern one — which is how `EOMONTH(DATE(2026,1,31),-1)` could land a full
month out at +14. They are now constructed with `Date.UTC` and read with
`getUTC*`, and a zone-less date string has its civil fields re-anchored to UTC
while strings carrying an explicit offset still resolve as instants. Verified
zone-invariant and Excel-correct across nine zones.

Two upstream behaviors that had been lost came back with it: `EDATE` clamps
day-of-month to the target month's last day (a January 31st plus one month gave
March 3rd), and `TODAY` reads the same UTC clock `NOW` reads, so
`FLOOR(NOW()) = TODAY()` now holds everywhere.

The authoring-side consequence — that "today" and "now" are irreducibly local
however well the rest behaves, so an indexed computed reading one is a trap — is
recorded on the BXL authoring-skill issue rather than here.

## On cards

The host suites drove only the financial chunk through a `computeVia`, via
`PMT`, so nothing showed that the statistical, Bessel, engineering and
validator.js chunks reach a card at all. Each is a separate dynamic import and
`computeVia` is synchronous, unable to await one mid-compute, so a family
missing from the bundle surfaces as an unknown-function error rather than a
wrong number. A `LazyFormulaProbe` card in the tracking fixture now carries one
computed per remaining family and asserts its indexed search doc. It sits apart
from the insurance cards because those families have no place in that domain.

## What it found

Writing a case per function turned up 21 functions that return the wrong answer
— among them `capture` always returning `{}` because named groups are discarded,
`TIMEVALUE` dropping AM/PM so every afternoon time comes back twelve hours
early, and `WEIBULL_DIST` transposing shape and scale so that realistic
parameters saturate to 0 or 1. Separately, numeric literals with an exponent
(`1e3`) fail to tokenize at all.

One of the 21 is fixed here rather than filed: **`sort_by` was ordering by the
whole item instead of by the key expression**, because `_sort_by_impl` compared
the `{item, ref}` envelope and `"item"` sorts ahead of `"ref"`, leaving the key
to break ties only. So `sort_by(.v)` ordered elements by their entire JSON body
— and `docs/syntax-reference.md` ships `.orders | sort_by(.placedAt) | reverse`
as a worked example that did not do what it says. It is the most likely of these
to bite an author today and the fix is one line, matching what the sibling
`_group_by_impl` already did. Verified against jq 1.7 on ordering by a field, by
a derived key, by a negated key, on tie stability, and on multiple keys.

For the remaining 20, rather than assert the wrong answers to get a green suite,
each case keeps the **correct** assertion and carries a `knownDefect` note plus
a `produces` record of what the function returns today. The suite then requires
all three: the program reaches the function, misses the correct answer, and
misses it in the documented way. So a fix promotes the case, and a defect that
changes shape — a different wrong value, or a throw where there was a value —
reports that its note has gone stale rather than staying silently green.

The full inventory, with mechanism and file for each, is on its own issue. None
of it is introduced here; it was all already shipping, just untested.

## Reviewing

The family tables under `tests/unit/fixtures/function-coverage/` are the bulk of
the diff and are data — the interesting code is `runner.ts` (recording,
zone-looping, known-defect inversion), `case.ts` (the contract), and the gate at
the bottom of `function-coverage-cli.ts`.

`tests/unit/validation-helpers-cli.ts` still asserts values that the coverage
table also covers. That is deliberate rather than leftover duplication: it drives
them through `runNativeJqAsync`, which auto-loads the validation chunk from the
program source, whereas the coverage suite preloads every family up front. The
two prove different things about the same functions.


packages/bxl/src/formulajs/dateSerial.ts
```diff
@@ -2,6 +2,12 @@ import { parseExcelNumber } from './common.ts';
 import { EXCEL_ERROR, throwExcelError } from './errors.ts';
 
 const MS_PER_DAY = 24 * 60 * 60 * 1000;
+// A trailing `Z`, a GMT/UTC token, or a numeric offset means the string fixes
+// its own zone; anything else leaves the zone to the host. The offset has to
+// follow a time of day, because a bare trailing `-2026` — as in the
+// `DD-Mon-YYYY` form `30-Apr-2026` — is a year, not a -20:26 offset.
+const HAS_EXPLICIT_ZONE =
+  /[zZ]\s*$|\b(?:GMT|UTC)\b|\d:\d{2}(?::\d{2})?(?:\.\d+)?\s*[+-]\d{2}:?\d{2}\s*$/;
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed and fixed in c412aadbb9.

Reproduced exactly as described: `DATEVALUE("April 30, 2026 23:30 EST")` gave 46143 under `TZ=UTC` and 46142 under `TZ=America/Los_Angeles`.

The deeper problem is that the check was a blacklist. Enumerating every way a zone can be written — `Z`, `+05:30`, `GMT`, `EST`, `Europe/Paris` — is a losing game, and each one missed silently reintroduces exactly the host dependence this change exists to remove. So the test is now a whitelist of the zone-*less* civil forms (`4/30/2026`, `2026/04/30`, `30-Apr-2026`, `April 30, 2026`, each with an optional time of day); anything else keeps the instant the parser produced. That inverts the failure mode: an unfamiliar format now falls back to the parser's answer, which is already correct whenever the string carries a zone.

Verified across 9 zones spanning −11 to +14: 15 forms, zoned and zone-less, all stable and matching Excel. Both your case and its zone-less twin are now coverage cases, so the distinction can't regress.

packages/bxl/tests/unit/fixtures/function-coverage/runner.ts
```diff
@@ -0,0 +1,225 @@
...
+  // fixed, and the case should go back to being an ordinary assertion.
+  if (testCase.knownDefect) {
+    if (!reached.has(testCase.covers)) {
+      return `the program never invoked ${testCase.covers}`;
+    }
+    return checkExpectation(testCase, outputs, thrown)
+      ? undefined
+      : `this case documents a known defect but now produces the expected ` +
+          `result. If ${testCase.covers} was fixed, drop its knownDefect note ` +
+          `so the case asserts normally.`;
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Good catch — fixed in 01c2f5170b.

You're right that this was much weaker than it looked: `checkExpectation` returning any truthy string counted as "still broken", so a function could start returning a different wrong value, or throw where it had returned one, and the case would stay green while its `knownDefect` note quietly became fiction.

A `knownDefect` case now also carries `produces`, recording what the function returns today, and the runner requires all three conditions: the program reaches the covered function, misses the correct answer, and misses it *in the documented way*. A defect that changes shape now reports that its note has gone stale rather than passing. `produces` is required whenever `knownDefect` is set, enforced at runtime alongside the existing "exactly one of expected/outputs/throws/check" check.

Verified all four transitions: documented shape passes; a different wrong value reports "the defect changed shape"; a throw where there was a value reports the same; and a genuine fix still reports "drop its knownDefect and produces so the case asserts normally."

packages/bxl/src/formulajs/dateSerial.ts
```diff
@@ -93,12 +99,29 @@ export function parseExcelDate(value: unknown): Date {
...
+        return HAS_EXPLICIT_ZONE.test(value)
+          ? parsed
+          : utcDate(
+              parsed.getFullYear(),
+              parsed.getMonth(),
+              parsed.getDate(),
+              parsed.getHours(),
+              parsed.getMinutes(),
+              parsed.getSeconds(),
+            );
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Confirmed and fixed in 04b7ee581b — `utcDate` now takes milliseconds and the re-anchoring passes `parsed.getMilliseconds()`.

This was a regression I introduced, not pre-existing: returning the parsed `Date` directly had preserved them, and rebuilding the civil fields silently truncated. Agreed on the consequence too — `daysBetween` rounds fractional-day differences, so a dropped sub-second could move an `XNPV`/`XIRR` day count.

packages/bxl/src/formulajs/dateSerial.ts
```diff
@@ -2,6 +2,15 @@ import { parseExcelNumber } from './common.ts';
 import { EXCEL_ERROR, throwExcelError } from './errors.ts';
 
 const MS_PER_DAY = 24 * 60 * 60 * 1000;
+// A date string denotes a calendar day only when it names no zone at all, and
+// only then may its civil fields be re-anchored to UTC. Spotting every way a
+// zone can be written is a losing game — `Z`, `+05:30`, `GMT`, `EST`,
+// `Europe/Paris` — and missing one silently reintroduces host dependence, so
+// recognize the zone-less civil forms instead and leave anything else as the
+// instant the parser produced. Covers `4/30/2026`, `2026/04/30`,
```
habdelra, line 10: [Claude Code 🤖] **This regex recognizes only a fraction of the zone tokens V8 actually honors, and every one it misses becomes host-dependent — the exact bug the rest of this change removes.** Blocking; this is a regression introduced here, not a pre-existing gap.

**What the runtime accepts.** V8's legacy date parser resolves a fixed offset for twelve tokens, and it is case-insensitive about all of them. Measured on the branch's Node:

```
token  V8 accepts  resolves to               regex says explicit
GMT    true        2026-04-30T12:00:00.000Z  true
UTC    true        2026-04-30T12:00:00.000Z  true
gmt    true        2026-04-30T12:00:00.000Z  false
utc    true        2026-04-30T12:00:00.000Z  false
UT     true        2026-04-30T12:00:00.000Z  false
PST    true        2026-04-30T20:00:00.000Z  false
pst    true        2026-04-30T20:00:00.000Z  false
EST/EDT/CST/CDT/MST/MDT/PDT — all accepted, all missed
```

`\b(?:GMT|UTC)\b` is case-sensitive and the alternation has no entry for `UT` or the eight US abbreviations, so all of those fall into the `else` branch of `parseExcelDate` and get their civil fields read back through the host zone and re-anchored to UTC.

**What that costs.** A string that names its own instant is re-read as a calendar day in whatever zone the process happens to be in, so the serial moves. Through the shipped functions, on this branch:

```
                                          UTC   | LA    | Kolkata | Kiritimati | Niue
DAY("30 Apr 2026 12:00:00 PST")           30    | 30    | 1       | 1          | 30
DATEVALUE("30 Apr 2026 12:00:00 PST")     46142 | 46142 | 46143   | 46143      | 46142
DATEVALUE("Thu, 30 Apr 2026 12:00:00 UT") 46142 | 46142 | 46142   | 46143      | 46142
```

**It is a regression.** I ran the same three expressions with `origin/main`'s `dateSerial.ts` swapped in and everything else on this branch. On `main` all three are stable at `30` / `46142` / `46142` in all five zones, because the old code always returned the parsed instant and the UTC-reading callers took it from there. The re-anchoring is what makes them move.

**The fix.** Cover V8's full accepted set and make the whole pattern case-insensitive:

```suggestion
const HAS_EXPLICIT_ZONE =
  /[zZ]\s*$|\b(?:UT|UTC|GMT|[ECMP][SD]T)\b|\d:\d{2}(?::\d{2})?(?:\.\d+)?\s*[+-]\d{2}:?\d{2}\s*$/i;
```

I ran both patterns over a fourteen-string battery under all five `TIMEZONES`. The proposed one makes `gmt`, `UT`, `PST` and `edt` stable while leaving every string the current one already handles unchanged — `30-Apr-2026`, `April 30, 2026`, `04/30/2026` and `2026-04-30T12:00:00` still re-anchor as calendar days, and `…Z`, `…+05:30`, `…-07:00` and the `Date.prototype.toString()` form still resolve as instants. `\bUT\b` needs the word boundaries it already has; it does not fire inside `August`.

**And the reason the suite is green.** No coverage case feeds a zone-abbreviation string to a date function — `formula-date.ts` covers the ISO form, the `DD-Mon-YYYY` form, the prose form and the explicit numeric offset, but nothing in the abbreviation family. Worth one case per shape here (`DATEVALUE("30 Apr 2026 12:00:00 PST")` and one lowercase `gmt`), since those are the strings this regex exists to classify.

packages/bxl/tests/unit/fixtures/function-coverage/formula-date.ts
```diff
@@ -0,0 +1,277 @@
...
+  // whenever they run. TODAY is a whole serial at or after 2026-01-01; NOW
+  // carries the same day plus a time fraction, so flooring it lands on TODAY.
+  {
+    covers: 'TODAY/0',
+    source: 'TODAY()',
+    check: (outputs) => {
+      strictEqual(outputs.length, 1);
+      const [today] = outputs as number[];
+      ok(Number.isInteger(today), `expected a whole serial, got ${today}`);
+      ok(
```
habdelra: [Claude Code 🤖] **`TODAY/0` is the one date fix in this change with no test holding it — the pre-branch implementation passes this suite.** Blocking, in the sense that the fix ships unpinned; the production change itself is right.

**The experiment.** I replaced the new `TODAY/0` body with `main`'s exact one:

```js
const now = new Date();
const epoch = new Date(1899, 11, 30);
yield Math.floor((now.getTime() - epoch.getTime()) / 86400000);
```

`node tests/unit/function-coverage-cli.ts` still exited 0. For contrast I did the same to `EOMONTH/2` and it failed immediately, on exactly the case you would want:

```
EOMONTH/2: EOMONTH(DATE(2026, 2, 5), 1) == DATE(2026, 3, 31)
```

So the fixed-value date cases have real teeth — this one does not.

**Why the zone loop does not save it.** `runCoverageCase` sets `process.env.TZ` and re-applies *the case's own assertion* once per zone; it never compares the outputs across zones. For a case pinned to a literal that is the same thing. For a `check` it is not: this one only asks for an integer at or after 46023, which the broken form satisfies in every zone, so five zones buy nothing.

**And `NOW/0` only half-covers it, intermittently.** `FLOOR(NOW()) = TODAY()` does relate the two, but the local-time form diverges from the UTC one only inside a daily window, because the divergence comes from the zone's *1899* offset rather than its modern one:

```
zone                 epoch skew   UTC hours where the two forms disagree
UTC                          0    none
America/Los_Angeles  +28800000    0–7
Asia/Kolkata         -19270000    19–23
Pacific/Kiritimati   +37760000    0–10
Pacific/Niue         +40780000    0–11
```

My green run happened at about 15:00 UTC, which is outside all five windows. Run in the small hours and it would go red. That is a flaky signal, not a gate — and it would stay flaky-shaped even if you added a cross-zone output comparison, since at a given instant all five zones agree on the broken value for 7 of 24 hours.

**The fix.** Compute the reference inside the check from `Date.UTC`, which is zone-independent by construction and deterministic at any hour:

```js
check: (outputs) => {
  strictEqual(outputs.length, 1);
  const now = new Date();
  const expected = Math.floor(
    (Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate()) -
      Date.UTC(1899, 11, 30)) / 86400000,
  );
  strictEqual(outputs[0], expected, 'TODAY reads the UTC calendar day');
},
```

That fails on `main`'s implementation in four of the five zones at any hour, and never fails on the new one.

packages/bxl/tests/unit/fixtures/function-coverage/runner.ts
```diff
@@ -0,0 +1,235 @@
...
+// A jq-source builtin is an AST, not a function, so it records when the
+// evaluator reads the definition it is about to run.
+function recordJqDef<T extends object>(name: string, def: T): T {
+  return new Proxy(def, {
+    get(target, property, receiver) {
+      invoked.add(name);
+      return Reflect.get(target, property, receiver);
+    },
+  });
+}
```
habdelra, line 61: [Claude Code 🤖] **Confirmation, and the evidence for it — the recorder is both precise and transparent.** This is the load-bearing piece of the suite, so it is worth writing down what holds and what would break it.

**It does not credit a function the program never ran.** I fed the runner adversarial cases whose programs *name* the covered function on a path that never executes:

```
sort_by/1        if false then sort_by(.a) else 1 end      no credit
ascii_downcase/0 def f(g): 1; f(ascii_downcase)            no credit
now/0            [limit(0; now)]                           no credit (reached empty/0, limit/2)
ABS/1            [limit(0; ABS(-1))]                       no credit
ABS/1            [empty | ABS(-1)]                         no credit
map/1            [empty | map(.)]                          no credit
BIN2DEC/1        try BIN2DEC("nope") catch "caught"        credited
```

The last one is the behavior you want: a function that raises still counts as invoked. The `DefAst` `get` trap is the subtle one and it comes out right, because `annotateBuiltinFilters` in `src/bxl/bridge/native.ts` only does a map lookup (`registry.jq[node.name]`) and an assignment onto the AST node — no property read on the def — while `isNativeFilter` is a bare `typeof`, which a `get` trap does not intercept. The first read of `.args` / `.body` in `evaluate.ts`'s `case 'filter'` happens after the evaluator has committed to running the def, and inside `for (const item of input)`, so an empty input stream records nothing.

**It does not move evaluation off the production route.** The comment above `recordNative` claims the re-wrap keeps the `bareNativeFilter` handle; measured, that is exact. Of the 702 native filters in the fully-loaded registry, 700 carry a bare handle before the recorder and 700 after — none lost, none gained. The only two on the `Proxy` path are `path/1` and `builtins/0`, and `builtins/0` is the one `recordResolvedBuiltinsFilter` re-wraps post-resolution. So nothing falls off `compiledScalar.ts`'s fast path, which is the failure mode the comment is guarding against.

**And it changes no answer.** I dumped all 856 cases' raw outputs from two fresh processes — one with `installInvocationRecorder()`, one without — under a fixed `TZ=UTC`. The two dumps are identical except for `now/0`, `NOW/0`, `RAND/0` and `RANDBETWEEN/2`, i.e. exactly the four genuinely nondeterministic cases.

**What would stop this being true.** Two things, worth knowing before editing here: a new post-resolution rewrite in `resolveRegistry` (today `builtins/0` is the only one, and it has its own handler above) would drop its recorder the same way, and any future code that reads a property off a registry `DefAst` for a reason other than invoking it — an arity pre-check, a diagnostic dump, a `structuredClone` — would start crediting jq builtins that never ran, because the trap is on `get` with no discrimination.

packages/bxl/tests/unit/function-coverage-cli.ts
```diff
@@ -0,0 +1,126 @@
...
+  installInvocationRecorder,
+  runCoverageCase,
+} from './fixtures/function-coverage/runner.ts';
+
+// The lazy families register themselves on load, so every chunk has to be in
+// before the registry can report the full function surface.
+await loadAllFormulaExtensions();
+
+const libraries = installInvocationRecorder();
+const exposed = new Set(resolveBuiltinRegistry(libraries).publicNames);
```
habdelra, line 35: [Claude Code 🤖] **The gate enumerates `publicNames`, which is narrower than "every function BXL exposes" in three specific ways.** Non-blocking, but the first two are worth a guard in this file since the whole point is that nobody has to remember; the third is the one a future contributor could actually trip.

**1. `_`-prefixed builtins are excluded from the enumeration but are callable.** `publicBuiltinNames` in `src/jqtools/evaluate/filters/registry.ts` filters `name.startsWith('_')`, and 15 names sit behind that filter. They are reachable from a program:

```
[1,2,3,4] | _nwise(2)   =>  [[1,2],[3,4]]
[1,2,3,4] | _flatten(1) =>  [[1,2,3,4]]
```

The sibling exclusion, `env/0`, is genuinely unreachable — `env` throws `env is not available in the public BXL sandbox` at dispatch — so that one is sound. The underscore family is not; it is just undocumented surface. Reasonable to declare out of scope, but the suite header says "every function BXL exposes must be invoked", and these are exposed to any expression that names them.

**2. Six names have a native implementation that can never dispatch, and the gate reports them covered anyway.** `resolveRegistry` keeps `jq` and `native` in separate maps, and both lookup paths — `annotateBuiltinFilters` in `src/bxl/bridge/native.ts` and `Environment.getVar` in `evaluate.ts` — check `jq` first. So where a name exists in both, the native side is dead:

```
FALSE/0, INDEX/2, INDEX/3, TRUE/0, bsearch/1, unique/0
```

I replaced each of those six native entries with a generator that throws, then evaluated `unique`, `bsearch(2)`, `TRUE()`, `FALSE()`, `INDEX(["a","b","c"], 2)` and `INDEX([[1,2],[3,4]], 2, 1)`. All six returned normally — the throw was never reached. The coverage case credits the name off the jq path, so the enumeration cannot see that a second implementation exists and is unreachable. `case.ts`'s `libraries` field handles cross-*library* shadowing (and `core-jq.ts` uses it for jq's own `INDEX/1`), but jq-over-native shadowing inside one library has no escape hatch at all.

Cheap guard, in the same spirit as `UNREACHABLE_BUILTINS`: assert that the resolved `jq` and `native` key sets are disjoint, with the six current overlaps listed explicitly and a reason each. Then a seventh overlap is a failure instead of silent dead code.

**3. A lazy family that is not wired into `loadAllFormulaExtensions()` is invisible to the gate.** The `await` above is what makes the enumeration complete, and what it awaits is hand-maintained — four `ensure*Loaded()` calls plus a hand-listed `ALL_LAZY_LIBRARIES` in `src/bxl/bridge/lazy-formulas.ts`. A fifth family added without touching both lists never enters `BXL_REGISTRY`, so `exposed` never lists its names and this suite passes with the whole family uncovered. That is the one route by which someone really can "add a builtin and still pass".

The check that closes it is also useful on its own merits: every lazy family already publishes a manifest of the names that trigger its chunk load (`FORMULA_STATISTICAL_FILTERS`, `FORMULA_BESSEL_FILTERS`, `FORMULA_ENGINEERING_FILTERS`, `FORMULA_FINANCIAL_FILTERS`, `VALIDATION_FILTERS`). Asserting each manifest set equals the set of non-underscore names its library owns would catch a family missing from the loader *and* catch manifest drift, which today would mean a card expression naming a function that never triggers its chunk. I checked both directions across all five manifests — 317 names, zero drift — so it lands green and stays a guard rather than a chore.

packages/bxl/tests/unit/fixtures/function-coverage/formula-logic.ts
```diff
@@ -0,0 +1,169 @@
...
+  // Branching.
+  //
+  // A wholly scalar IF/IFS program is answered by the compiled-scalar fast
+  // path, which carries its own copy of both and never reaches the builtin.
+  // The blank guard on each condition keeps these programs on the builtin
+  // route, which is also where a non-scalar formula lands in production.
+  { covers: 'TRUE/0', source: 'TRUE()', expected: true },
+  { covers: 'FALSE/0', source: 'FALSE()', expected: false },
+  {
+    covers: 'IF/3',
```
habdelra, line 15: [Claude Code 🤖] **This header is right about the fast path, and the consequence is bigger than it reads: `IF` and `IFS` have two implementations, and the cases cover the one a scalar card expression skips.** Pre-existing, not introduced here — but this PR is the thing that now claims the surface is covered, so it is worth either covering both or saying which one is covered.

**The twin.** `src/jqtools/evaluate/compiledScalar.ts` special-cases the names directly:

```js
case 'filter':
  if (node.name === 'IF/2' || node.name === 'IF/3') return compileExcelIf(node.args);
  if (node.name.startsWith('IFS/')) return compileExcelIfs(node.args);
  return compileBareNativeFilter(node);
```

`compileExcelIf` and `compileExcelIfs` are hand-written re-implementations, not calls into the registry. Every other name on that path (`compileBareNativeFilter`) borrows the registry's own bare filter, so `IF`/`IFS` are the only true twins.

**Proof they are what runs.** I deleted every `IF*` / `IFS*` entry from the registry and re-evaluated:

```
IF(1 > 0, "yes", "no")             => "yes"   (fast path answered it)
IFS(1 > 2, "a", 1 > 0, "b")        => "b"     (fast path answered it)
IF(ISBLANK(null), "blank", "set")  => 'IF/3' is not defined
```

So the blank guard this table adds is precisely what moves the program onto the registry copy. The covered implementation is the one production takes only when something else in the expression already forced it off the scalar path.

**They can and do disagree.** `compileExcelIfs` accepts any even argument count from 2 up; the registry stops at `IFS/4`:

```
registry: IF/2, IF/3, IFS/4, IFS/6, IFS/8, IFS/10, IFS/12, IFS/14, IFS/16

IFS(1 > 2, "a")            => #N/A                      (fast path, Excel-correct)
IFS(ISBLANK(1), "a")       => 'IFS/2' is not defined    (registry path)
```

`IFS/2` therefore exists only on the fast path, which also means it is not in `publicNames` and this gate never asks for it. On the arities both sides implement I could not find a divergence — `IF(0, …)`, `IF("", …)` and the omitted-else default all agree — so the exposure is narrow, but it is real.

**Suggestion.** Keep these nine guarded cases and add unguarded scalar twins of them (`IF(1 > 0, "y", "n")`, `IFS(1 > 2, "a", 1 > 0, "b")`, and so on up the arities). They cost nothing, they pin `compileExcelIf`/`compileExcelIfs`, and they would fail the day the two copies drift. The `covers` bookkeeping is unaffected — a duplicate `covers` key is already fine, `covered` is a `Set`.

packages/bxl/tests/unit/fixtures/function-coverage/formula-engineering.ts
```diff
@@ -0,0 +1,251 @@
...
+  { covers: 'DEC2BIN/1', source: 'DEC2BIN(-100)', expected: '1110011100' },
+  { covers: 'DEC2BIN/2', source: 'DEC2BIN(9, 4)', expected: '1001' },
+  { covers: 'DEC2HEX/1', source: 'DEC2HEX(100)', expected: '64' },
+  // Excel renders hex digits uppercase — DEC2HEX(-54) is documented as
+  // 'FFFFFFFFCA'.
+  {
+    covers: 'DEC2HEX/1',
+    source: 'DEC2HEX(255)',
+    expected: 'FF',
+    knownDefect: 'lowercase hex digits, as for BASE/2',
```
habdelra, line 41: [Claude Code 🤖] **This is the one of the 21 that the repo has already written down as a divergence, which is what `case.ts` says disqualifies `knownDefect`.** Non-blocking, but it needs a decision either way, because right now two artifacts in the same package disagree about whether lowercase hex is intended.

`docs/syntax-reference.md` says, in the Excel-paste section:

> **Paste test:** … Known paste gap: `MODE` is not implemented. `AND` / `OR` / `XOR` use array-style arguments, and `DEC2HEX` returns lowercase.

And `case.ts` says:

> Reach for this only for a divergence from the function's specification. A divergence this codebase deliberately documents is not a defect: assert the documented behavior and note it in a comment.

I think `knownDefect` is defensible here — that doc line files the behavior as a *gap*, in the same breath as "`MODE` is not implemented", which reads as an acknowledged shortfall rather than a chosen semantic. But then the doc sentence is the thing that should move: it currently reads as settled behavior a paste author should plan around, and it names only `DEC2HEX` while the defect note (correctly) covers `BIN2HEX`, `OCT2HEX` and `BASE` too. I verified the family claim — `BIN2HEX("11111111")` and `OCT2HEX("377")` both give `ff` — and the round-trip claim in the note as well: `HEX2DEC("ff")` and `DECIMAL("ff", 16)` are both 255, so uppercasing the output side breaks nothing.

Two clean resolutions, either fine:

- Keep `knownDefect`, and reword the doc line to say the hex-emitting family currently returns lowercase where Excel returns uppercase, naming all four functions — so the doc points at the same open item instead of blessing it.
- Or treat it as settled, flip both cases to assert lowercase with a comment pointing at the doc line, and drop them from the defect list.

The one thing not to leave is the current state, where the doc says "this is how it works" and the suite says "this is wrong and must stay wrong".

Everything else in this table checks out, for the record — I re-derived the whole file independently (base conversions and their two's-complement widths, the 48-bit `BITAND` ceiling, all fifteen `IM*` values at `z = i` against their hyperbolic closed forms, `CONVERT`, `DELTA`/`GESTEP`, `ERF`/`ERFC`), and found no photographed values. One small nit while you are here: `ERF(0, 1)` does not discriminate `ERF/2` from `ERF/1`, because `erf(0)` is 0 and the expected value is the same `0.8427007929`. `ERF(1, 2)` would.

packages/bxl/tests/unit/fixtures/function-coverage/core-jq.ts
```diff
@@ -0,0 +1,738 @@
...
+  {
+    covers: 'scalars/0',
+    source: '[(1, "a", [2], {"b":1}, null, true) | scalars]',
+    expected: [1, 'a', null, true],
+  },
+  {
+    covers: 'scalars_or_empty/0',
+    source: '[(1, [2]) | scalars_or_empty]',
+    expected: [1],
+  },
```
habdelra, line 47: [Claude Code 🤖] **This case cannot tell `scalars_or_empty` from `scalars`, and the difference between them is exactly where BXL diverges.** Non-blocking; a two-token change to the input turns it into a real discriminator.

`[(1, [2]) | scalars_or_empty]` is byte-for-byte the answer plain `scalars` gives. The "or empty" half of the name is what admits `[]` and `{}` — jq 1.6's definition is

```
def scalars_or_empty: select(. == null or . == true or . == false
  or type == "number" or type == "string"
  or ((type == "array" or type == "object") and length == 0));
```

and `builtinNativeFilters.ts` gates on `isScalar()`, which is null/boolean/number/string only. Run against an input that contains the discriminating values:

```
BXL  [(1,[2],[],{},{"a":1},null,"s") | scalars_or_empty]  =>  [1,null,"s"]
BXL  [(1,[2],[],{},{"a":1},null,"s") | scalars]           =>  [1,null,"s"]
jq   (1.6 definition, re-run under jq 1.7.1)              =>  [1,[],{},null,"s"]
```

So in BXL `scalars_or_empty` *is* `scalars`. Widening the input to `(1, [2], [], {}, {"a": 1})` either promotes this case to something that pins the semantics, or surfaces a divergence that deserves a `knownDefect` alongside the other 21. `src/jqtools/UPSTREAM-DIFFS.md` calls it "yield-or-empty — Trivial" and `docs/syntax-reference.md` lists it under Misc with no semantics, so nothing currently documents it as a chosen narrowing.

**The rest of this table verified against real jq.** For context on how much weight I put on the finding above: I ran every value-asserting case in `core-jq.ts` and `core-math.ts` through `/usr/bin/jq` 1.7.1 with the same input and the same output normalization — 229 cases with an oracle, 203 reproduce exactly. Every one of the 26 that did not is either a jq 1.8 addition verified against `builtin.jq` from `master` (`trim`, `ltrim`, `rtrim`, `trimstr`, `add/1`, `skip/2`, `toboolean`), a builtin removed in jq 1.7 verified against its 1.6 definition (`leaf_paths`, `recurse_down`), a jq-tools binary math form that upstream never registers, a build-configuration builtin, or a divergence `docs/syntax-reference.md` already documents (`gamma/0` canonicalized to true Γ, `lgamma_r/0` returning the magnitude alone). Arity discrimination is solid throughout — I re-ran every `NAME/k` case with the trailing argument dropped and none produced the same answer at lower arity. No photographed values anywhere in either table.

packages/bxl/tests/unit/fixtures/function-coverage/core-jq.ts
```diff
@@ -0,0 +1,738 @@
...
+    covers: 'capture/2',
+    source: '"XYZ" | capture("(?<low>xyz)"; "i")',
+    expected: { low: 'XYZ' },
+    knownDefect: 'named captures are discarded, as for capture/1',
+    produces: { expected: {} },
+  },
+  {
+    covers: 'scan/1',
+    source: '"abcabc" | [scan("a.")]',
+    expected: ['ab', 'ab'],
```
habdelra, line 463: [Claude Code 🤖] **The `capture/1` defect note claims a cascade into `sub`/`gsub`, but none of these four cases uses a named capture, so the cascade is prose only.** Non-blocking; it means fixing `transformRegExpMatch` would silently fix behavior nothing asserts, which is the opposite of what the `knownDefect` mechanism is for.

All four replacements here are literal strings, so they pass today and would pass whatever `transformRegExpMatch` does with `groups`. The divergent input is a replacement that interpolates a named group:

```
BXL  "abc123" | sub("(?<n>[0-9]+)"; "<\(.n)>")   =>  "abc<null>"
jq   same program                                =>  "abc<123>"
```

Two things follow. First, the note's wording is slightly off — the captures are not *empty*, they interpolate the literal string `null`, which is worse for a card author because it silently produces plausible-looking text rather than a blank. Second, a case for this belongs in the set: `sub/2` (or a fifth case) with a named group and a `knownDefect` note pointing at the same `transformRegExpMatch` mechanism, so that the fix has to promote two cases rather than one.

The same shape applies to `isinfinite/0`: its note describes a cascade into `isfinite/0`, but the `isfinite/0` case in `core-math.ts` is `infinite | isfinite`, which agrees with jq today. The divergent input is `nan | isfinite` — `true` in jq, `false` in BXL (I confirmed both). A case asserting `nan | isfinite` → `true` with a `knownDefect` would keep the second half of that defect honest too.

packages/bxl/tests/unit/fixtures/function-coverage/core-math.ts
```diff
@@ -0,0 +1,348 @@
...
+  },
+  // Roots and powers
+  {
+    covers: 'sqrt/0',
+    source: 'sqrt',
+    input: 2,
+    expected: 1.4142136,
+    tolerance: 1e-7,
+  },
+  { covers: 'cbrt/0', source: 'cbrt', input: 27, expected: 3 },
```
habdelra, line 152: [Claude Code 🤖] **The only value in this table that real jq does not produce, asserted without a tolerance and without a note saying so.** Non-blocking nit.

```
BXL  27 | cbrt  =>  3
jq   27 | cbrt  =>  3.0000000000000004
```

BXL's answer is the mathematically right one — `Math.cbrt` is exact at a perfect cube where glibc's `cbrt` is not — so nothing needs to change in the code. But this table's contract is "jq's own answers", and every other transcendental in it (`log2(8)`, `log10(1000)`, `exp2(3)`, `exp10(2)`) matches jq bit for bit, so a reader reasonably takes the bare `3` as jq's. One line of comment — that the exactness is BXL's and jq returns a value one ulp up — keeps the table honest, the way the neighbouring `lgamma_r` comment already does for its divergence.

Same shape one entry down at `gamma/0`: the comment says "not the historical POSIX log-Γ alias", but jq 1.7.1 on Linux returns log-Γ *today*, not historically — `syntax-reference.md` already explains the canonicalization, and pointing the fixture at it would land the same way the `lgamma_r` comment does.

packages/bxl/src/bxl/bridge/formula-contrib-native.ts
```diff
@@ -2265,49 +2284,50 @@ const bareNativeFilters: Record<string, BareNativeFilter> = {
...
-    const date = new Date(epoch.getTime() + serial * 86400000);
-    date.setMonth(date.getMonth() + m + 1, 0); // last day of target month
-    yield Math.floor((date.getTime() - epoch.getTime()) / 86400000);
+    const date = serialToUtcDate(serial);
+    // Day 0 of the following month is the last day of the target month.
+    date.setUTCMonth(date.getUTCMonth() + m + 1, 0);
+    yield utcDateToSerial(date);
   },
 
   // ═══════════════════════════════════════════════════════════════
```
habdelra, line 2333: [Claude Code 🤖] **Confirmation: both of these are correct and zone-invariant, including the cases the single coverage case each does not reach.** Nothing to change.

I ran both against an independent reference (`Date.UTC` month normalization plus a day-0 last-day-of-month computation, written from Excel's rule rather than from this code), under all five `TIMEZONES`:

**`EDATE`, 14 cases, 0 mismatches, all zone-stable.** Forward and backward, across leap boundaries and from a Feb-29 origin:

```
EDATE(2026-01-31, +1)   2026-02-28      EDATE(2024-01-31, +1)   2024-02-29
EDATE(2026-03-31, -1)   2026-02-28      EDATE(2026-05-31, -3)   2026-02-28
EDATE(2026-01-31, +13)  2027-02-28      EDATE(2026-03-31, -14)  2025-01-31
EDATE(2023-12-31, +2)   2024-02-29      EDATE(2024-02-29, +12)  2025-02-28
EDATE(2024-02-29, -12)  2023-02-28      EDATE(2028-01-29, +1)   2028-02-29
EDATE(2026-01-31,  0)   2026-01-31      EDATE(2026-07-15, -25)  2024-06-15
```

The clamp is structurally right rather than incidentally right: `Date.UTC(y, m0 + m, 1)` normalizes month overflow and underflow in both directions before the last-day lookup, so nothing depends on `m` being small or positive.

**`EOMONTH`, 8 cases, 0 mismatches, all zone-stable**, including `-1` across a year boundary, `+1` across one, and `-14`. The two-argument `setUTCMonth(month, 0)` form sidesteps the usual day-overflow trap, since month and date land atomically — `EOMONTH(2026-01-31, 0)` correctly stays on 2026-01-31 rather than sliding through February.

**And the motivating bugs are real**, which I checked by running `origin/main`'s version of this file with everything else on the branch:

```
                                UTC   | LA    | Kolkata | Kiritimati | Niue
EOMONTH(DATE(2026,2,5),1)       46112 | 46111 | 46112   | 46111      | 46112
EOMONTH(DATE(2026,1,31),-1)     46022 | 46022 | 46022   | 46052      | 46022
EDATE(DATE(2026,1,31),1)        46084 | 46084 | 46084   | 46081      | 46084
WEEKDAY(DATE(2026,4,30))            5 |     5 |     5   |     6      |     5
```

The 30-day `EOMONTH` swing at +14 is the one the description calls a full month out, and it reproduces exactly. The `EDATE` 46084 is 2026-03-03, the overflow the clamp removes.

One consistency note for whoever edits here next, not a request: `serialToUtcDate` maps serial *N* to `1899-12-30 + N` with no adjustment for Excel's phantom 1900-02-29, while `parseExcelDate` → `serialToExcelDate` in `dateSerial.ts` does carry that adjustment. The two therefore disagree for serials below 61, which is why `WEEKDAY`/`ISOWEEKNUM`/`EDATE`/`EOMONTH` and `DAY`/`MONTH`/`YEAR`/`WEEKNUM` are not interchangeable down there. That predates this change — the old epoch arithmetic had the same property — but naming the helper makes it read more authoritative than it is, so a one-line comment on `serialToUtcDate` saying it is the raw 1899-12-30 mapping would help.

packages/bxl/src/formulajs/UPSTREAM-DIFFS.md
```diff
@@ -71,6 +71,34 @@ preamble.
...
+
+The local-time form is not merely zone-sensitive but self-inconsistent, because
+the serial epoch materializes at a zone's 1899 offset while results are read at
+its modern one — the two differ by 8 minutes in Kolkata and by nearly a day at
++14, which is how a month-end could land 30 days off.
+
+Two other divergences ride along, both restoring upstream behavior BXL had lost:
+`EDATE` clamps day-of-month to the target month's last day (upstream clamps;
+BXL let the day overflow, so a January 31st plus one month gave March 3rd), and
+`TODAY` reads the same UTC clock `NOW` reads, so `FLOOR(NOW()) = TODAY()` holds
```
habdelra, line 99: [Claude Code 🤖] Two things on this new section — one wording, one gap. Neither blocks.

**Wording.** "both restoring upstream behavior BXL had lost" narrates the package's own history in a document that otherwise describes the current state of the divergence. A reader six months from now wants to know what BXL does versus what upstream does, not which of the two got there second. "`EDATE` clamps day-of-month to the target month's last day, as upstream does" carries the same information without the arrow of time.

The two numeric claims in the paragraph above both check out, incidentally: the 1899 versus modern offset gap is 8 minutes 50 seconds for Kolkata, and Kiritimati's is 24 hours 29 minutes — so "8 minutes" and "nearly a day" are accurate, and the 30-day `EOMONTH` swing they explain reproduces on `main`.

**Gap.** This file is now the home for the date family's divergences, and four more of them live in the same family without an entry — all four surfaced by this PR's own `knownDefect` list, all four pre-existing:

- `WEEKDAY` branches on return types 1, 2 and 3 only; 11 through 17 fall through to the Sunday default.
- `WEEKNUM` branches on return type 2 only; 11 through 17 and the ISO 21 all fall through.
- `ISOWEEKNUM` has no year-boundary branch. The note in `formula-date.ts` names the early-January direction; it fails at the other end too — `ISOWEEKNUM(DATE(2024,12,30))` returns 53 where ISO 8601 gives week 1 of 2025 (2024-12-30 is the Monday that opens ISO 2025-W01).
- `TIMEVALUE` drops the meridiem suffix.

Since the section you are adding already covers "how BXL's dates differ", listing these as known shortfalls rather than chosen semantics would put the whole date story in one place — and would mean the `knownDefect` notes and the doc agree, which is the thing that goes wrong in the `DEC2HEX` case (see the comment on `formula-engineering.ts`).

Separately, and outside this PR: `src/jqtools/UPSTREAM-DIFFS.md` currently states `atan2/2(x; y)` uses "**Excel argument order** (`x` first), NOT vanilla-jq (`y; x`)". The implementation takes `(y, x)`, its own module header says "jq/POSIX order", `docs/syntax-reference.md` says `atan2(y; x)` is jq/POSIX order, and empirically `atan2(1; 0)` is `1.5707963267948966` and `atan2(0; 1)` is `0` — identical to jq. So the coverage case is right and that paragraph is the thing that is wrong. Worth a follow-up for whoever next edits that file.

packages/bxl/tests/unit/fixtures/function-coverage/case.ts
```diff
@@ -0,0 +1,107 @@
...
+  zones?: string[];
+}
+
+/** Marks a table of jq builtins, which are written in canonical jq. */
+export function jqCases(cases: CoverageCase[]): CoverageCase[] {
+  return cases.map((entry) => ({ readableSyntax: false, ...entry }));
+}
+
+/**
+ * The zone spread date cases run under: UTC, a negative whole-hour offset, a
```
habdelra, line 92: [Claude Code 🤖] Three notes on the contract itself. All non-blocking.

**1. The `knownDefect` inversion accepts *any* failure, not the specific wrong answer.** `runInAmbientZone` returns success whenever `checkExpectation` returns a message, so a case stays green if the function starts throwing, starts returning `undefined`, or starts returning a *differently* wrong value. Combined with the invocation check that runs first, the guarantee is only "it dispatches and does not produce the right answer" — a function that breaks completely still reads as a documented defect. Pinning today's observed output alongside the correct one, and requiring both (`this is wrong, and it is wrong in this specific way`), would make a defect case notice drift rather than only notice repair. Optional — it doubles the maintenance on 21 cases — but worth a sentence in the doc comment either way, so a reader knows what the inversion does and does not promise.

**2. `expected` cannot distinguish an empty stream from a single `null`, nor a single array output from a stream of its elements.** The normalization in `runner.ts` collapses `values.length === 0` to `null`, so a case asserting `expected: null` passes both ways; and `expected: [0.75, 4]` passes whether the program emitted one pair or two scalars. `frexp/0` and `modf/0` are the cases where that property is exactly what the comment claims ("yields one `[mantissa, exponent]` pair"), and `keys`, `explode`, `gmtime`, `strptime` have the same shape. `outputs: [...]` pins it where it matters — worth mentioning on the `expected` doc line so authors reach for it.

**3. `TIMEZONES` says "the two extremes of the offset range"; `Pacific/Niue` is −11.** The bottom of the range is −12 (`Etc/GMT+12`). Not worth changing the list — nothing plausible breaks only at −12 once everything is UTC-anchored, and Niue is a real place where Etc/GMT+12 is not — but the comment overclaims slightly. "a wide negative offset" would be accurate.

For the record on this list working as intended: I ran all 856 cases under all five zones and compared their raw outputs. Only `RAND/0` and `RANDBETWEEN/2` vary, and those are the RNG. Every date-touching case carries `zones` — there is no date-dependent case missing it, which was the thing most worth checking here. The one place the spread has no teeth is a `check`-based case, which re-runs its own assertion per zone rather than comparing across them; see the comment on `TODAY/0` in `formula-date.ts`.

packages/host/tests/helpers/cards/bxl-tracking.ts
```diff
@@ -60,6 +64,38 @@ export const bxlTrackingCardSource = `
...
+    });
+    // formula-engineering, co-bundled with financial as formula-extras
+    @field decodedBinary = contains(NumberField, {
+      computeVia: expression(fx\`BIN2DEC(BinaryCode)\`),
+    });
+    // validation, the validator.js helpers
+    @field contactEmailValid = contains(BooleanField, {
+      computeVia: expression(fx\`isEmail(ContactEmail)\`),
+    });
+  }
```
habdelra, line 97: [Claude Code 🤖] **Confirmation on the probe card, plus one thing worth double-checking before merge.** I did not run the host suite — it needs the stack — so this is a read of the fixture and its consumers rather than a test result.

**The four values are right and each is exact rather than rounded-to-taste.** `NORM_S_DIST(0, TRUE)` is exactly ½ by symmetry, so the `ROUND(… * 10000) / 10000` returns `0.5` with nothing lost. `BESSELJ(1, 0)` is J₀(1) = 0.7651976866…, which rounds to `0.7652`. `BIN2DEC("101010")` is 42 with no sign ambiguity — six digits, well short of the ten that would make it two's-complement. `isEmail` on a `.example` address is unambiguously true. Good choices: each one has a single defensible answer, so a chunk that loaded but dispatched to something else would show up as a wrong number rather than a near-miss.

**The claim the test makes is the right one, and it is the one the node suite genuinely cannot make.** Each family is a separate dynamic import and `computeVia` is synchronous, so the failure mode is an unknown-function error at compute time, not a numeric drift — which is exactly what asserting the indexed search doc catches. The node coverage suite front-loads `loadAllFormulaExtensions()` and so proves nothing about the bundle.

**Worth confirming:** this fixture is shared by three host suites — the platform-module one this test lives in, plus the expression and cyclic-graph suites — so the new card lands in all three realms. I read through them for anything count-sensitive and found nothing realm-wide (the one `assert.strictEqual(panels.length, 2)` in the expression suite is a field on a locally constructed `Quote`, and the cyclic-graph queries filter on the `Claim` type ref, which a `LazyFormulaProbe` cannot match). So I expect all three to stay green, but since I could not run them, a green host shard on this branch is the thing that closes it.

One small thing: `contactEmailValid` is the only computed here that does not go through a numeric round-trip, and `assert.true` would also pass on a truthy non-boolean. `assert.strictEqual(searchDoc.contactEmailValid, true, 'validation')` would pin that the validator chunk returned an actual boolean, which is what the `BooleanField` contract wants.

habdelra (commented):
[Claude Code 🤖] I reviewed this along the two axes where a coverage suite can lie to you: whether the invocation recorder credits things it should not, and whether the expectations are derived or merely photographed — plus the UTC change, which is the only part that alters shipping behavior. Everything below was run against the branch rather than read off it.

**Bottom line: the suite is sound and the tables are genuinely derived, but there are two blocking items in the date work — a zone-detection regression in `HAS_EXPLICIT_ZONE` that makes zone-abbreviation date strings host-dependent, and a `TODAY/0` fix that no test pins (the pre-branch implementation passes this suite).** Everything else is non-blocking: three enumeration holes in the gate, an `IF`/`IFS` twin the cases steer away from, one documented-divergence conflict, and a handful of nits.

### What lands right

**The recorder is precise and transparent, and I could not break either property.** Dead branches, filter arguments, `limit(0; …)` and empty input streams all correctly earn no credit, while a function that throws still counts as invoked. The `DefAst` `get` trap — the piece most likely to over-credit — comes out clean because `annotateBuiltinFilters` only does a map lookup and an assignment, and `isNativeFilter` is a `typeof` check that no trap intercepts. On transparency the comment's claim is exact: 700 of 702 native filters carry a bare handle before the recorder and 700 after, none lost, so nothing falls off the compiled-scalar fast path. And dumping all 856 cases' raw outputs from two fresh processes, with and without the recorder, gives identical results except for the four genuinely nondeterministic cases. Detail in the thread on `runner.ts`.

**The expectations are derived, not recorded.** I re-derived independently and found zero mismatches: every closed form in `formula-financial` (the hand-amortized two-period loans, `IPMT + PPMT = PMT`, `EFFECT`/`NOMINAL` inverting, the T-bill and 30/360 day counts); roughly thirty constants in `formula-statistical` against their closed forms (chi-square with 2 df as an exponential, F(1,2) as a squared Cauchy, the t(1) Cauchy identities, `T_TEST` reducing to 1 − 3/√13); the whole `formula-engineering` table including all fifteen `IM*` values at z = i; the five Bessel references; `formula-text` against Microsoft's own documented examples; `formula-stats` (including the published `SKEW`/`KURT` examples); every calendar fact in `formula-date`; and `formula-lookup`. Separately, all 229 value-asserting cases in `core-jq` and `core-math` went through real `jq` 1.7.1 as an oracle — 203 reproduce exactly and every one of the other 26 is a jq-version difference or a divergence the docs already record. Arity discrimination holds: re-running every `NAME/k` case with the trailing argument dropped changed the answer every time.

**Zone-proofing is complete for the cases that exist.** Running all 856 under all five zones and comparing outputs, only `RAND/0` and `RANDBETWEEN/2` vary. There is no date-dependent case missing `zones`.

**`EDATE` and `EOMONTH` are correct and zone-invariant**, verified across 22 cases against an independent reference — leap years, backward offsets, year boundaries, Feb-29 origins. And the motivating bugs reproduce exactly on `main`, including the 30-day `EOMONTH` swing at +14.

**All 21 known defects are genuine**, each verified against an independent oracle (real jq for the jq family, the published rule for the Excel ones, `jstat`'s actual signature for `WEIBULL_DIST`), each still failing for the reason its note gives, and each note's blamed mechanism confirmed in the source. Only one has a documentation conflict, and it is a narrow one.

### Recommendations

Blocking:

1. **Widen `HAS_EXPLICIT_ZONE` to V8's full accepted token set, case-insensitively.** `UT`, lowercase `gmt`/`utc`, and the eight US abbreviations are all honored by the parser and all missed by the regex, so those strings get re-anchored through the host zone. `DAY("30 Apr 2026 12:00:00 PST")` is 30 under UTC and 1 under Kolkata on this branch, and stable in all five zones on `main` — a regression. Suggested pattern, verified against a fourteen-string battery, in the thread on `dateSerial.ts`. Add a case per shape while you are there; the absence of one is why the suite is green.
2. **Give `TODAY/0` a deterministic assertion.** Reverting it to `main`'s exact body leaves the suite green; reverting `EOMONTH/2` fails immediately. The zone loop re-applies each case's own assertion per zone rather than comparing across them, so a loose `check` gets no zone guarantee. Computing the expected serial from `Date.UTC` inside the check fixes it at any hour. Thread on `formula-date.ts`.

Non-blocking, worth doing here:

3. **Close two of the gate's enumeration holes with assertions.** Six names have an unreachable native implementation shadowed by a jq definition, which the gate credits as covered — assert the two key sets are disjoint, listing today's six with reasons. And assert each lazy family's manifest matches the names its library owns, which also closes the "new family missing from `loadAllFormulaExtensions()`" route. Both land green today (I checked 317 manifest names in both directions). Thread on `function-coverage-cli.ts`.
4. **Add unguarded scalar twins of the `IF`/`IFS` cases.** `compiledScalar.ts` carries its own `IF`/`IFS` implementations, and the blank guard in `formula-logic.ts` is what moves the program onto the registry copy — so the covered implementation is the one a plain scalar card expression skips. The two already diverge at `IFS/2`. Thread on `formula-logic.ts`.
5. **Reconcile the `DEC2HEX` lowercase note with `docs/syntax-reference.md`**, which already records it as a known gap — either reword the doc line to name all four affected functions as open, or flip the cases. Thread on `formula-engineering.ts`.
6. **Assert the two cascades the defect notes describe** — a named-capture `sub`/`gsub` replacement (BXL gives `"abc<null>"` where jq gives `"abc<123>"`) and `nan | isfinite`. Both are prose today, so fixing them would promote nothing. Thread on `core-jq.ts`.
7. **Widen the `scalars_or_empty` input** so it can tell that function from `scalars` — in BXL it currently is `scalars`, and this case cannot see that. Thread on `core-jq.ts`.
8. Smaller ones, each in its thread: the evergreen wording and the four undocumented date shortfalls in `UPSTREAM-DIFFS.md`; the `knownDefect` inversion accepting any failure and `expected`'s inability to distinguish an empty stream from `null`, both in `case.ts`; the `cbrt` comment in `core-math.ts`; `ERF(0, 1)` not discriminating its arity; and `assert.strictEqual(…, true)` for the validator probe.

I did not run the host suite. The new probe card lands in three shared realms; I read the other two suites for count-sensitive assertions and found none, but a green host shard is what closes that.

### Adjacent, out of scope

- `src/jqtools/UPSTREAM-DIFFS.md` says `atan2/2` uses Excel argument order. The implementation, its own module header, `docs/syntax-reference.md` and the runtime all say jq/POSIX `(y; x)` — `atan2(1; 0)` is π/2 and `atan2(0; 1)` is 0, identical to jq. The coverage case is right; that paragraph is stale.
- `docs/syntax-reference.md` documents `SUBSTITUTE("aaa", "a", "b", 2)` → `"aba"`. The code returns `"aab"`, which is the `SUBSTITUTE/4` defect this PR catalogues — so the doc example is currently a specification the code does not meet, sitting in a table a reader takes as descriptive.
- `parseExcelDate`'s re-anchoring drops milliseconds, since `utcDate` takes no ms argument. Only reachable for a zoneless string carrying fractional seconds, and only observable through `daysBetween`'s `Math.ceil`, so it is a curiosity rather than a bug — but it is a small behavior delta from `main` that nothing records.
- Fifteen `_`-prefixed builtins sit outside `publicNames` yet are callable from any expression (`_nwise(2)` over `[1,2,3,4]` gives `[[1,2],[3,4]]`). Declaring them out of scope is fine; saying so in the suite header would stop the next reader from assuming otherwise.

packages/bxl/src/formulajs/dateSerial.ts
habdelra: [Claude Code 🤖] This one is already fixed — the worktree this was reviewed in predates c412aadbb9, which replaced `HAS_EXPLICIT_ZONE` with `ZONELESS_CIVIL_DATE`.

The diagnosis was right, and the same class of bug was reported independently by Codex earlier on this thread (`"April 30, 2026 23:30 EST"` gave 46143 under UTC and 46142 in `America/Los_Angeles`). What that prompted was not a longer token list but inverting the test: enumerating every way a zone can be written is a losing game — `Z`, `+05:30`, `GMT`, `EST`, `UT`, lowercase `gmt`, `Europe/Paris` — and each one missed silently reintroduces host dependence. So the check is now a whitelist of the zone-*less* civil forms, and anything else keeps the instant the parser produced. That inverts the failure mode: an unfamiliar format falls back to the parser's answer, which is already correct whenever the string carries a zone.

All four shapes you flagged are stable on current HEAD, across UTC / America/Los_Angeles / Asia/Kolkata / Pacific/Kiritimati / Pacific/Niue:

```
STABLE DAY("30 Apr 2026 12:00:00 PST")   30(5z)
STABLE DAY("30 Apr 2026 12:00:00 EST")   30(5z)
STABLE DAY("30 Apr 2026 12:00:00 gmt")   30(5z)
STABLE DAY("30 Apr 2026 12:00:00 UT")    30(5z)
STABLE DATEVALUE("30-Apr-2026")          46142(5z)
STABLE DATEVALUE("2026-04-30T23:30:00-07:00")  46143(5z)
```

Your point that no case covered this path was fair at the time; `DATEVALUE` now has cases for a named zone, the same text without one, and an explicit numeric offset.

packages/bxl/tests/unit/fixtures/function-coverage/formula-date.ts
habdelra: [Claude Code 🤖] Confirmed, and fixed in 2618aedb73 — though not deterministically, and the measurement is worth recording.

I reproduced it exactly: swapping `TODAY/0` back to the pre-branch local-epoch body left the suite green. The old assertion only wanted a whole serial at or after 2026-01-01, which the broken form satisfies. It now derives the expected serial from `Date.UTC` and compares.

Where I have to correct the suggestion: that is **not** deterministic. A local-epoch `TODAY` agrees with a UTC `TODAY` whenever no day boundary falls between them, so no assertion evaluated at a single instant can tell them apart during that window. I measured it by sweeping the whole UTC day against the five zones the case runs under — both comparing to the UTC-derived serial and requiring cross-zone equality catch the broken form in **16 of 24 hours**, and neither catches it in the other 8. Your green run at ~15:00 UTC and mine at 15:48 both landed in that blind window, which is why the first attempt at this fix also passed with the bug in place.

So the case goes from catching it 0/24 to 16/24, and the comment says so rather than implying the hole is closed. Reaching 24/24 needs an injectable clock, which is a testability change to the engine and out of scope here. What limits the residual risk is that the epoch arithmetic `TODAY` shares with `EDATE`, `EOMONTH` and `WEEKDAY` is pinned exactly by those cases — you confirmed an `EOMONTH` revert fails instantly — so what rides on this case alone is a regression to `TODAY`'s own body, which two-thirds of the day now catches.

packages/bxl/tests/unit/fixtures/function-coverage/runner.ts
habdelra: [Claude Code 🤖] Thanks — that transparency check (all 700 bare handles preserved, byte-identical output with and without the recorder) is exactly the evidence I wanted and did not have. Recorded.

packages/bxl/tests/unit/function-coverage-cli.ts
habdelra: [Claude Code 🤖] Agreed on all three, and none is fixed here — filed as CS-12532 so they are not lost.

The shadowed-native one is the interesting one: six names whose native implementation can never dispatch because a jq def of the same key wins, so the gate credits them off the jq path. That is not a coverage lie exactly — the function a program reaches *is* covered — but it does mean six native implementations are dead code that the suite reports as exercised, which is worth knowing either way. The hand-maintained `loadAllFormulaExtensions()` is the one I would close first, since a lazy family missing from it is invisible to a gate whose whole premise is that the registry is the source of truth.

packages/bxl/tests/unit/fixtures/function-coverage/formula-logic.ts
habdelra: [Claude Code 🤖] Agreed, and this is the finding I would most want a future maintainer to know, so it is now the second item in CS-12531 (a BXL engine maintainer skill): registry-enumerated coverage structurally cannot see the fast-path copy, so someone fixing `IFS` in the registry can leave the version that actually runs untouched and watch the suite go green. `IFS/2` existing only on the fast path makes the divergence concrete rather than hypothetical.

Not fixing it here — reaching the compiled-scalar copy needs its own approach, not another case.

packages/bxl/tests/unit/fixtures/function-coverage/formula-engineering.ts
habdelra: [Claude Code 🤖] Fair challenge, and I am keeping `knownDefect` — but the contract wording is what should change, so I have noted it on CS-12531.

The distinction I intended is between behavior the repo documents as *intended* and behavior it documents as a *shortfall*. `docs/syntax-reference.md` lists this under "Known paste gap", which records that we fall short of Excel, not that lowercase hex is the contract. A gap someone wrote down is still a gap. Had it been in `UPSTREAM-DIFFS.md` as a deliberate divergence, `knownDefect` would be wrong.

Your second point stands unqualified though: the doc names only `DEC2HEX` while the defect covers `BIN2HEX`, `OCT2HEX` and `BASE` too, and CS-12530 records the wider scope.

packages/bxl/tests/unit/fixtures/function-coverage/core-jq.ts
habdelra: [Claude Code 🤖] Confirmed — `[(1, [2], "a", null) | scalars_or_empty]` and the same program with `scalars` both give `[1,"a",null]`, so nothing in reach of this case separates them. Logged on CS-12532 with the other gate and case-quality items rather than guessed at here, since what the intended difference *is* needs settling before a case can pin it.

packages/bxl/tests/unit/fixtures/function-coverage/core-jq.ts
habdelra: [Claude Code 🤖] Right, and fixed in 4850c6bffe — the cascade is now asserted rather than described. `"abc" | sub("(?<x>b)"; "[" + .x + "]")` gives `a[]c` where jq gives `a[b]c`, and the `gsub` twin likewise, both carrying their own `knownDefect` and `produces`. I did the same for the other prose-only cascade you flagged, `nan | isfinite`.

This was the right catch for the right reason: a prose claim is exactly how a fix lands untested, and the `sub`/`gsub` half is the more user-visible of the two, since a replacement referencing a named group silently substitutes nothing into the output string.

packages/bxl/tests/unit/fixtures/function-coverage/core-math.ts
habdelra: [Claude Code 🤖] Confirmed the values — BXL gives exactly 3, real jq 1.7 gives 3.0000000000000004 — and I am leaving the assertion at 3.

The table asserts what each function *should* produce, and the cube root of 27 is 3; jq's answer is its `Math.cbrt` rounding, not a contract worth reproducing. What is missing is the note saying so, which is a fair nit — logged with the other case-quality items on CS-12532.

packages/bxl/src/bxl/bridge/formula-contrib-native.ts
habdelra: [Claude Code 🤖] Appreciated — an independent re-derivation of the calendar facts across 22 cases is worth more than my own, since I wrote them.

packages/bxl/src/formulajs/UPSTREAM-DIFFS.md
habdelra: [Claude Code 🤖] Taking the wording fix and logging the gap.

On the gap: the section documents the divergence for the functions this change actually moved to UTC, and the ones you name as still local are pre-existing rather than introduced here — but you are right that a reader of that section will reasonably infer the whole surface is UTC now, which is worse than saying nothing. Logged on CS-12532 with the other suite-and-docs items.

packages/bxl/tests/unit/fixtures/function-coverage/case.ts
habdelra: [Claude Code 🤖] On your first point — the `knownDefect` inversion accepting any failure — that is fixed already, in 01c2f5170b, which your worktree predates. Copilot raised the same hole independently. A `knownDefect` case now also carries `produces`, recording what the function returns today, and the runner requires all three conditions: it reaches the function, misses the correct answer, and misses it *in the documented way*. A defect that changes shape, including one that starts crashing, now reports that its note has gone stale.

The other two — `expected` not distinguishing an empty stream from a single `null`, and the `TIMEZONES` comment calling Niue an extreme when it is −11 — are both fair and both on CS-12532. The first is a contract-documentation matter rather than a mechanism change, since `outputs: []` already expresses the difference.

packages/host/tests/helpers/cards/bxl-tracking.ts
habdelra: [Claude Code 🤖] That is the right thing to have checked, and it is closed: I ran the host suite after every source change on this branch, most recently after the `sort_by` and date fixes. 18 tests, 65 assertions, 0 failed, including the three suites that share the tracking fixture — `bxl expressions on real cards`, `bxl platform module`, and `bxl cyclic card graphs`. So the probe card landing in those shared realms breaks nothing, which matches your read that none of their assertions are count-sensitive.

habdelra:
[Claude Code 🤖] Both blocking items are resolved as of 4850c6bffe. Taking them in turn, since one of them was already fixed and the other needs a correction to the suggested remedy.

> **1. Widen `HAS_EXPLICIT_ZONE` to V8's full accepted token set, case-insensitively.** `UT`, lowercase `gmt`/`utc`, and the eight US abbreviations are all honored by the parser and all missed by the regex, so those strings get re-anchored through the host zone. `DAY("30 Apr 2026 12:00:00 PST")` is 30 under UTC and 1 under Kolkata on this branch, and stable in all five zones on `main` — a regression.

This was already fixed in c412aadbb9, six commits before the tip you reviewed. The diagnosis was right and the bug was real — Codex reported the same class of failure independently on this PR — but the remedy went the other way rather than widening the token list.

Enumerating zone spellings is unwinnable: `Z`, `+05:30`, `GMT`, `EST`, `UT`, lowercase `gmt`, `Europe/Paris`, and whatever V8 accepts next. Every miss silently reintroduces host dependence, which is the failure mode this change exists to remove. So the test is now a whitelist of the zone-*less* civil forms, and anything it does not recognize keeps the instant the parser produced. That inverts the failure mode: an unfamiliar format falls back to the parser's answer, which is already correct whenever the string carries a zone. All four shapes you named are stable on the tip, across UTC / America/Los_Angeles / Asia/Kolkata / Pacific/Kiritimati / Pacific/Niue, and `DATEVALUE` now has cases for a named zone, the same text without one, and an explicit numeric offset.

> **2. Give `TODAY/0` a deterministic assertion.** Reverting it to `main`'s exact body leaves the suite green; reverting `EOMONTH/2` fails immediately. … Computing the expected serial from `Date.UTC` inside the check fixes it at any hour.

The finding is correct and I had missed it — I reproduced it exactly, and the old assertion only wanted a whole serial at or after 2026-01-01, which the broken form satisfies. Fixed in 2618aedb73.

But the last clause is not right, and it matters: **that does not fix it at any hour.** A local-epoch `TODAY` agrees with a UTC one whenever no day boundary falls between them, so no assertion taken at a single instant can separate them during that window. Sweeping the whole UTC day against the five zones the case runs under, comparing against the `Date.UTC`-derived serial catches the broken form in **16 of 24 hours** — and requiring cross-zone equality instead catches exactly the same 16, because in the other 8 all five zones agree with UTC. Your green run at ~15:00 UTC and my first attempt at this fix at 15:48 both landed in that blind window, which is why the first attempt also passed with the bug in place.

So the case goes from 0/24 to 16/24, and its comment says so rather than implying the hole is closed. Getting to 24/24 needs an injectable clock, which is a testability change to the engine and not something I want to smuggle in here. What bounds the residual risk is your own observation: the epoch arithmetic `TODAY` shares with `EDATE`, `EOMONTH` and `WEEKDAY` is pinned exactly by those cases, so what rides on this one is a regression to `TODAY`'s own body.

## The non-blocking items

Done here: **(6)** both cascades are now asserted rather than described (4850c6bffe) — `sub`/`gsub` with a named capture, and `nan | isfinite`. That was the best of the non-blocking catches, because a prose-only claim is exactly how a fix lands untested, and the `sub`/`gsub` half is user-visible: a replacement referencing a named group substitutes nothing into the output string.

Also already fixed before your review: the millisecond truncation you filed as adjacent-and-curious (04b7ee581b, prompted by Copilot), and the `knownDefect` inversion accepting any failure (01c2f5170b) — a case now also records what the function `produces` today, so a defect that changes shape reports that its note has gone stale instead of staying green.

Tracked rather than done: **(3)** and **(7)** and the `case.ts`/`cbrt`/`ERF` nits from **(8)** are on CS-12532, along with the `UPSTREAM-DIFFS` gap. **(4)**, the `IF`/`IFS` twin, is on CS-12531 — it is the sharpest example of the thing that skill exists for, since registry-enumerated coverage structurally cannot see the fast-path copy, and reaching it needs its own approach rather than another case. The stale `atan2` paragraph and the `SUBSTITUTE` doc example that documents behavior the code does not meet are on CS-12530.

Declined, with reasoning in its thread: **(5)**. `docs/syntax-reference.md` files the lowercase hex under "Known paste gap", which records that we fall short of Excel rather than that lowercase is the contract — a gap someone wrote down is still a gap. Had it been in `UPSTREAM-DIFFS.md` as a deliberate divergence, `knownDefect` would be wrong. Your second point stands though, and CS-12530 carries it: the doc names only `DEC2HEX` while the defect covers `BIN2HEX`, `OCT2HEX` and `BASE` too.

> I did not run the host suite.

Closed: run after every source change on this branch, most recently after the `sort_by` and date fixes — 18 tests, 65 assertions, 0 failed, including all three suites that share the tracking fixture. That matches your read that none of their assertions are count-sensitive.

Thanks for the transparency and derivation work in particular. The 700-of-702 bare-handle count and the with/without-recorder output diff are evidence I wanted and could not supply myself, and an independent re-derivation of tables I wrote is worth more than my own second pass over them.

habdelra:
[Claude Code 🤖] A second independent review pass found a regression in the date work that the first pass missed, fixed in 60bc989a7c and 17a51b892f.

**The zone-less shape whitelist was too narrow.** It ended its optional time at seconds and required the string to begin with the date, so an AM/PM suffix or a leading weekday name fell through to the parser's host-resolved instant — the exact failure the whitelist exists to prevent, and a regression against `main`. A card reading `YEAR` off a human-entered `"Dec 31, 2026 11:30 PM"` indexed as 2026 and rendered as 2027 in a Los Angeles browser. It now accepts a leading weekday and a meridiem; verified over 24 forms across 9 zones spanning −11 to +14, in both directions — zone-less forms are calendar days, and anything naming a zone (numeric, `Z`, `GMT`, `UT`, or a US abbreviation in either case) still keeps its instant. Both previously-missing shapes now have cases, and narrowing the pattern back fails them.

**Four cases pinned BXL's answer rather than the function's contract**, and one was blessed by a doc entry resting on a false premise. `lgamma_r` returns only the log magnitude, dropping the sign it exists to report — `UPSTREAM-DIFFS.md` said jq has no way to return it, but jq returns it as element 1 of a pair, so that sentence is corrected here. `inputs` raises where jq yields an empty stream. `scalars_or_empty` is a byte-identical copy of `scalars`, and its case could not see that because the input held no empty collection. And `match` discards capture names — the root that the `capture`, `sub` and `gsub` defects all grow from, which nothing asserted directly, so patching `capture` alone would have promoted three cases and left `match` wrong. All four now assert the correct answer under `knownDefect`, and the behaviors are catalogued on the defect issue.

Suite is 66/66 at 865 cases and 27 known defects; host suite 18 tests / 65 assertions green. The remaining non-blocking findings — chiefly that the suite is strong against a function being replaced wholesale but weak against a single argument being ignored, with 99 names never asserted in both boolean polarities — are measured and recorded on the follow-up issue rather than fixed here.

lukemelia (approved):

PR #5794, opened 2026-08-17
Update boxel-icons to latest lucide and tabler icon sets

lukemelia:
## What this does

Bumps the icon source packages in `packages/boxel-icons` and regenerates the committed icon components via `pnpm rebuild:all`:

- `lucide-static` 0.447.0 → 1.31.0
- `@lucide/lab` 0.1.2 → 0.2.0
- `@tabler/icons` 3.19.0 → 3.46.0

Net effect on the icon set: **457 added, 43 removed, 623 updated** (6,841 icons total).

## Removed icons

The 43 removals are almost entirely lucide's brand icons (github, twitter, facebook, slack, figma, chrome, …), which lucide dropped from its set. Tabler's `brand-*` equivalents remain available for all of them.

- The only in-repo usages of a removed icon were two `linkedin` imports in `packages/experiments-realm`; they use tabler's `brand-linkedin` (visually equivalent) instead.
- Deployed icon URLs stay safe: `deploy:s3` uses `aws s3 sync` without `--delete`, so previously published icon modules remain served at boxel-icons.boxel.ai even after removal from the source set.

## Identifier collisions in the meta module

Lucide ships some icons under both an old and a new name (e.g. `arrow-down-0-1` and `arrow-down-01`) with identical artwork. Those distinct file names PascalCase to the same identifier (`ArrowDown01`), which broke `boxel-icons-meta.ts` compilation. `bin/rebuild-all.mjs` now generates every icon file (each remains importable by path) but lists only the first of each colliding pair in the meta module, keeping the existing `ArrowDown01`-style exports bound to the same files as before.

## Test plan

- `pnpm rebuild:all` then `pnpm build` (rollup + type declarations) pass in `packages/boxel-icons`.
- Grepped all packages for import-style and URL-style references to each of the 43 removed icons; `linkedin` was the only hit.
- `boxel-icons-meta` consumers (`icons-grid.gts`, host loader test) use the icon list generically, not named imports of removed icons.


backspace (approved):
The Percy diffs (apart from missing ones 😬) actually correct for once

PR #5797, opened 2026-08-17
Serve realm files with Content-Length and HTTP byte-range support

lukemelia:
## What this does

Raw file responses streamed from disk carried no `Content-Length` and ignored the `Range` request header entirely — every request got a `200` with the full body and no `Accept-Ranges`. Browsers therefore treat realm-hosted media as an unseekable stream of unknown size: Safari's native player refuses playback without byte-range support, and Chromium reports `duration: Infinity` for formats whose duration derives from byte size (PCM WAV, plain MP3 without a Xing header). Container formats that embed their duration (M4A's `moov` atom) mask the gap, which is why some media appeared to work.

`serveLocalFile` now:

- emits `Content-Length` whenever the byte size is knowable without reading the body — the adapter's stat for streamed files, `byteLength` for in-memory bytes;
- advertises `Accept-Ranges: bytes` and answers a single satisfiable `bytes` range with a `206` + `Content-Range`, streamed bounded from disk (`fs.createReadStream(path, { start, end })`) rather than sliced in memory;
- answers an unsatisfiable range with a `416` carrying `Content-Range: bytes */<size>`;
- honors `If-Range`: a validator that doesn't match the current ETag means the client's offsets refer to a different representation, so the full body is served;
- keeps `If-None-Match` precedence — a `304` wins over any `Range`.

`FileRef` gains optional `size` and `createRangeStream` members, populated by the node adapter from the stat it already performs. The range paths never touch `FileRef.content`, which is a lazy getter that opens a full-file stream on first access — a `206`/`416` must not strand an unconsumed file handle. Multi-range and malformed `Range` headers are answered with the full `200` representation, which RFC 9110 permits; browsers only issue single ranges.

## Scope

Range support alone doesn't make protected-realm media play: the auth service worker rewrites media requests to `mode: 'cors'`, and `Range` isn't in the realm server's CORS allow-list, so those preflights still fail. That CORS change is a separate follow-up; this PR is the serving-path half and is fully exercisable on public realms.

## Test plan

- `packages/realm-server/tests/http-range-test.ts` — unit coverage of the range-header resolution (bounded/open-ended/suffix ranges, clamping, unsatisfiable cases, multi-range and malformed headers).
- `packages/realm-server/tests/range-request-test.ts` — integration coverage over a live realm server: `Content-Length`/`Accept-Ranges` on full GETs, exact byte slices on `206`s, `416`, `If-Range` both ways, and `If-None-Match` precedence.
- Existing file-serving suites pass locally through the card-source GET coverage (later modules in that suite hit a local test-pg disk limit unrelated to this change; CI runs them all).

### DIFF
--- a/packages/runtime-common/realm.ts
+++ b/packages/runtime-common/realm.ts
@@ -1,4 +1,5 @@
 import { Deferred } from './deferred.ts';
+import { resolveRangeHeader } from './http-range.ts';
 import {
   awaitRealmIndexSettled,
   indexingConcurrencyGroup,
@@ -307,6 +308,19 @@ export interface FileRef {
   path: LocalPath;
   content: ReadableStream<Uint8Array> | Readable | Uint8Array | string;
   lastModified: number;
+  // Total byte size of `content`, when the adapter knows it without reading
+  // the bytes (e.g. from the stat it already performed). Lets streamed file
+  // responses carry a Content-Length, which browsers need before they will
+  // treat media as seekable / of known duration.
+  size?: number;
+  // Open a bounded byte range of the file, [start, end] inclusive, without
+  // materializing the rest. Present when the adapter can do this cheaply
+  // (e.g. a bounded fs read stream); together with `size` it enables HTTP
+  // Range (206) serving. Both offsets are within `size`.
+  createRangeStream?: (
+    start: number,
+    end: number,
+  ) => ReadableStream<Uint8Array> | Readable;
 
   [key: symbol]: object;
 }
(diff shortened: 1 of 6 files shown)


packages/runtime-common/realm.ts
```diff
@@ -4021,6 +4035,86 @@ export class Realm {
...
+    // Byte size when knowable without reading the bytes. `ref.size` is
+    // consulted before `ref.content` because adapters expose `content` as a
+    // lazy getter that opens a real stream on first touch — a ranged or 416
+    // response must never pay for (and then strand) a full-file stream.
+    // String bodies are left to the HTTP layer, which measures and sets
+    // Content-Length for them itself.
+    let sliceableBytes =
+      ref.size == null && ref.content instanceof Uint8Array
+        ? ref.content
+        : undefined;
```
lukemelia, line 4048: [Claude Code 🤖] **Confirmation plus a latent trap worth one line of documentation. Non-blocking.**

This `sliceableBytes` branch — an in-memory `Uint8Array` body sliced with `subarray` — is the one range/Content-Length path the node realm server never reaches today, and it hides a Content-Length correctness trap if that ever changes.

**Why it's unreachable now.** `NodeAdapter.openFile` always returns `content` as a lazy full-file `Readable` and sets `size`, so `ref.size != null` short-circuits the `ref.size == null && ...` guard and `sliceableBytes` stays `undefined`; `materializeFileRef` only yields a `Uint8Array` when the original content already was one, which the node adapter never produces (it decodes streams to text). So server-side, disk bodies always take the `createRangeStream` / `nodeStream` path — exactly what the integration tests exercise.

**The trap.** When a body *is* an in-memory `Uint8Array` (or the full-body `ReadableStream` branch below), `setContextResponse` in `packages/realm-server/middleware/index.ts` runs it through `webStreamToText`, which decodes with a UTF-8 `TextDecoder`. For non-UTF-8 bytes that is lossy (invalid sequences collapse to U+FFFD), so Koa re-encodes a body whose length differs from the byte-exact `Content-Length` set here from `byteLength` — a truncated or over-long response with a mismatched header. It's latent only because the node file path never lands here.

**Ask:** nothing for this PR. A one-line note that this branch assumes the body reaches the client as raw bytes would pay off if anyone later adds a "return small files as an in-memory `Uint8Array`" optimization to the node adapter — that would need to keep such bodies off the `webStreamToText` path (or drop the byte-exact length). Also flagging for coverage: no test exercises this branch; the suite only covers the `createRangeStream` slice.

packages/runtime-common/realm.ts
```diff
@@ -4021,6 +4035,86 @@ export class Realm {
...
+    let totalSize = ref.size ?? sliceableBytes?.byteLength;
+    if (totalSize != null) {
+      headers['content-length'] = String(totalSize);
+    }
+    let rangeCapable =
+      totalSize != null &&
+      (sliceableBytes != null || typeof ref.createRangeStream === 'function');
```
lukemelia, line 4055: [Claude Code 🤖] **Confirmation — the new Content-Length is safe here; documenting the mechanism so the next editor need not re-derive it. Non-blocking, nothing to change.**

Emitting `Content-Length` on the streamed full-body `200` (previously absent) is the load-bearing change, and it reaches the client intact:

- **No re-encoding layer.** The realm server runs no compression middleware (no `koa-compress`/gzip in the chain), so nothing strips or invalidates the length. `setContextResponse` sets headers via `ctxt.set(...)` *before* assigning `ctxt.body = nodeStream`; since that is the first body assignment (`original == null`), Koa's stream-body setter leaves `Content-Length` in place, so the value set here is what ships.
- **Value matches the bytes.** `ref.size` is the same `stat` the node adapter used to open the file, and the full-body path streams that same file — length equals bytes. The `full GET` integration test pins it (10-byte upload → `content-length: 10`, body round-trips).
- **No transpiled-vs-source mismatch.** Modules never reach `serveLocalFile`: `loadModuleFromDisk` routes executable extensions through `#transpileWithLayers`, and only the raw `non-module` branch calls this method. And because `cloneFileRefWithContent` drops `size`, a materialized (cached) source ref carries none, so this line is skipped for it and Koa measures the string itself. The only responses that get a `size`-derived `Content-Length` are raw disk files whose bytes equal `size`.

lukemelia (commented):
[Claude Code 🤖] **Review lens:** correctness of the HTTP serving contract this adds — that the emitted `Content-Length` always equals the bytes shipped, that the range math and conditional-request precedence follow RFC 9110, and that all of it survives the Koa/node-stream layer intact.

**Bottom line: no blocking issues.** The change is well-constructed and, as far as I can trace and test, correct. The two inline notes are a guard-rail confirmation and one latent (currently-unreachable) trap — both non-blocking.

- **304 beats Range.** `serveLocalFile` returns the `if-none-match === etag` 304 before any range handling runs, so a conditional request never degrades to a 206. Confirmed by the `If-None-Match takes precedence over Range` integration test.
- **The range resolver is sound.** `resolveRangeHeader` treats only a well-formed single `bytes=` range as sliceable and folds multi-range / non-bytes / inverted / offset-less headers to `full`, with inclusive offsets that line up with `fs.createReadStream({ start, end })` (also inclusive) and with `Uint8Array.subarray(start, end + 1)`. The unit suite covers bounded/open-ended/suffix/clamped/unsatisfiable/malformed, including the empty-representation and zero-length-suffix corners.
- **Content-Length is accurate and survives the pipeline** — see the inline note on `totalSize`.

**On the two things I checked hardest and cleared:**
- **If-Range only matches the ETag, never an HTTP-date.** A date-form `If-Range` (or any non-matching validator) falls through to a full `200`, which is always a safe answer, and since the response always carries an ETag, browsers send the ETag form. Correct as written.
- **Source-endpoint asymmetry, harmless.** `materializeFileRef` → `cloneFileRefWithContent` drops `size`/`createRangeStream`, so cached-source responses advertise no `Accept-Ranges`, while the `bypassCache` source path (which serves the raw handle) does. Only raw disk files get range support; source text neither needs nor gets it. Worth knowing, nothing to change.

**Adjacent, out of scope (correctly deferred):** the PR body notes protected-realm media still won't play until `Range` is added to the realm server's CORS allow-list, since the auth service worker rewrites these to `mode: 'cors'`. Agreed that's a separate change; this serving-path half is fully exercisable on public realms and stands on its own.

backspace (approved):

PR #5798, opened 2026-08-17
Allow Range through realm-server CORS for authenticated media requests

lukemelia:
## What this does

Native `<audio>`/`<video>` elements cannot attach `Authorization`, so on protected realms the host's auth service worker re-issues media requests as `mode: 'cors'` with the token injected. That rewrite turns the media element's `Range` header into an author header needing preflight approval — and the realm server's `Access-Control-Allow-Headers` didn't include `Range`, so the preflight failed and the native player errored before any bytes flowed.

- `Range` and `If-Range` join the CORS allow list in the realm server's `@koa/cors` config.
- `Content-Range`, `Accept-Ranges`, and `Content-Length` join **both** expose lists: the `@koa/cors` `exposeHeaders` and the `Access-Control-Expose-Headers` that `createResponse` stamps on every realm handler response — the latter replaces rather than merges with the former, so an addition to only one of them is silently dropped on real responses. Exposure matters to more than app JS: the service worker hands the CORS-filtered `Response` from its own `fetch()` straight to the media element via `respondWith`, so an unexposed `Content-Range`/`Accept-Ranges` is invisible to the player's loading stack itself.

## Test plan

- `packages/realm-server/tests/range-request-test.ts` — alongside the byte-range serving coverage (same realm fixture): a preflight carrying `Access-Control-Request-Headers: range, if-range, authorization` is approved with all three in the allow list, and a cross-origin `206` carries `Access-Control-Allow-Origin` (without which no expose list is consulted) and exposes `Content-Range`/`Accept-Ranges`/`Content-Length`.
- The realm-server suite loads an explicit module list in `tests/index.ts`; `http-range-test` and `range-request-test` are registered there so the range and CORS coverage actually runs in CI.
- Browser-level verification that protected-realm media plays end-to-end (service worker → preflight → authed `206`s → native playback) is deliberately left to follow-up work with the FileDef player integration coverage.


packages/realm-server/server.ts
```diff
@@ -1109,8 +1109,15 @@ export class RealmServer {
...
         cors({
           origin: '*',
+          // Range/If-Range are here for native media playback: <audio>/<video>
+          // elements cannot attach Authorization, so the host's auth service
+          // worker re-issues their requests as mode:'cors' with the token
+          // injected. That rewrite turns the media element's Range header into
+          // an author header needing preflight approval — without Range in
+          // this list the preflight fails and the player errors before any
+          // bytes flow.
           allowHeaders:
```
lukemelia, line 1119: [Claude Code 🤖] **Confirmation — `Range`/`If-Range` genuinely need to be here, and here's the subtlety that should stop a future maintainer from "cleaning up" `Range` as redundant. Non-blocking.**

The comment's framing ("the Range header becomes an author header needing preflight approval") is right in effect, but a maintainer who half-remembers the Fetch spec might object that `Range` is CORS-safelisted and try to drop it. The precise picture, which is why both names must stay:

- A **single range with a defined start** (`bytes=2-5`, `bytes=90-`) is a CORS-safelisted request header, so *on its own* it neither forces a preflight nor appears in `Access-Control-Request-Headers`.
- A **suffix range** (`bytes=-N`, null start) is **never** safelisted, and **`If-Range` is never** safelisted. Native players issue both, and the service worker's injected `Authorization` forces the preflight regardless — so once preflighting, a suffix `Range` and any `If-Range` land in `Access-Control-Request-Headers` and must be in this allow list or the preflight fails.
- Browser implementations of `Range` safelisting have also not been uniform, so listing `Range` explicitly is the robust choice even for the simple-range case.

Net: keep both. The mechanism is sound; this note is just the guard-rail so the list isn't trimmed later on a "Range is safelisted anyway" reading.

packages/runtime-common/create-response.ts
```diff
@@ -28,8 +28,11 @@ export function createResponse({
       // which is why `Retry-After` appears below: readiness pairs it with
       // `X-Boxel-Not-Ready` on a 503, and a cross-origin caller that could read
       // the stage but not the retry hint would only have half the answer.
+      // Content-Range/Accept-Ranges let a cross-origin caller reason about a
+      // 206 byte-range response (Content-Length is safelisted, but is listed
+      // for symmetry with the range pair).
       'Access-Control-Expose-Headers':
-        'X-Boxel-Realm-Url,X-Boxel-Realm-Public-Readable,X-Boxel-Realm-Archived,X-Boxel-Canonical-Path,X-Boxel-Not-Ready,Authorization,Cache-Control,ETag,Retry-After',
+        'X-Boxel-Realm-Url,X-Boxel-Realm-Public-Readable,X-Boxel-Realm-Archived,X-Boxel-Canonical-Path,X-Boxel-Not-Ready,Authorization,Cache-Control,ETag,Retry-After,Content-Range,Accept-Ranges,Content-Length',
```
lukemelia: [Claude Code 🤖] **Confirmation — this is the list that reaches the wire on a media `206`; documenting the ordering that makes that true. Non-blocking, nothing to change.**

Verified against `@koa/cors@4.0.0`: for a non-OPTIONS response the middleware sets `Access-Control-Expose-Headers` from its own `exposeHeaders` *before* `await next()` (`index.js:95-105`), then the handler runs and `setContextResponse` copies this `createResponse` header onto the Koa context via `ctxt.set(...)`, replacing the middleware's value. So the `206` from `serveLocalFile` (which routes through `createResponse`) exposes exactly this list — the `content-range` / `accept-ranges` / `content-length` a cross-origin native player reads. The `a ranged response exposes its range headers` test pins this list specifically.

`Content-Length` is a CORS-safelisted response header, so it's exposed to JS with or without being named here — listing it for symmetry (as the comment says) is harmless and reads clearly.

lukemelia (commented):
[Claude Code 🤖] **Review lens:** does the CORS surface actually let an authenticated, cross-origin, ranged media request through — the preflight approves the request headers the service worker sends, and the `206` exposes the headers a native player needs — and is every home of that contract updated, not just one.

**Bottom line: no blocking issues.** The change is correct and the mechanism reasoning in both the PR body and the code comments checks out against the middleware source. Two inline confirmations below document *why* it's right so the next editor doesn't undo it; both non-blocking.

**What lands right (verified against `@koa/cors@4.0.0` source, not just asserted):**
- **The "replace, not merge" claim is real.** For an actual (non-OPTIONS) response, `@koa/cors` sets `Access-Control-Expose-Headers` from its `exposeHeaders` *before* `await next()` (`node_modules/.pnpm/@koa+cors@4.0.0/.../index.js:95-105`); the handler then runs, and `setContextResponse` (`packages/realm-server/middleware/index.ts`) copies `createResponse`'s headers onto the Koa context with `ctxt.set(...)`, overwriting it. So on any realm-handler response, `createResponse`'s list is final — which is exactly the list the media `206` carries. Updating `createResponse` is therefore the load-bearing half; updating the `@koa/cors` `exposeHeaders` covers responses that never go through `createResponse`.
- **The preflight list is final.** For OPTIONS, `@koa/cors` sets `Access-Control-Allow-Headers` from the configured `allowHeaders` and returns `204` without invoking the handler (`index.js:155-163`), so nothing downstream can drop `Range`/`If-Range`.
- **Twin check came back clean.** The only other CORS header sites — `proxy-forward.ts` (SSE / `proxyAsset` for the auth-SW script), the host dev-CORS middleware, `matrix-backend-authentication.ts` — none serve realm file bytes, so none need the range headers. The realm server's `cors()` + `createResponse` are the two real homes, and both are updated.

**Test coverage:** the two tests pin the two *winning* lists — the preflight's `Allow-Headers` and the `206`'s `Expose-Headers` (which is `createResponse`'s). Worth knowing that the `@koa/cors` `exposeHeaders` addition on `server.ts` is *not* exercised by either test, because every range response is produced by `createResponse` and so never surfaces the middleware's list; it's belt-and-suspenders for a hypothetical non-`createResponse` range response. Fine to keep, just uncovered.

**Adjacent, out of scope (correctly deferred):** end-to-end browser verification (service worker → preflight → authed `206`s → native playback) is left to the FileDef-player follow-up, which is the right place for it — this PR's unit/integration level is the right altitude for the CORS contract itself.

packages/runtime-common/create-response.ts
habdelra: [Claude Code 🤖] **Confirmation, plus a comment-precision suggestion — the list is right, but the stated reason is weaker than the real one, and the weak version is what invites a future trim. Non-blocking.**

**What the comment says vs. what actually consumes these headers.** "let a cross-origin caller reason about a 206 byte-range response" reads as if the beneficiary is application JS inspecting a `Response`. It isn't — or rather, that's the incidental case. The load-bearing consumer is the media element's own loading stack.

Trace it through `packages/host/public/auth-service-worker.js`: `buildAuthedRequest` re-issues the media request with `mode: 'cors'`, and the fetch handler does `event.respondWith(fetch(buildAuthedRequest(request, token)))`. The `Response` that `fetch` resolves is a **CORS-filtered** response — its header list is pruned to the CORS-safelisted names plus whatever `Access-Control-Expose-Headers` names. `respondWith` then hands *that filtered response* to the player. So a `Content-Range` or `Accept-Ranges` missing from this list isn't merely unreadable by app JS; it is absent from the response the player itself inspects to decide whether the resource is seekable. That is why omitting them breaks playback rather than just degrading introspection.

**Verified.** Against `@koa/cors@4.0.0` (the catalog version), with a Koa app mirroring this shape — `cors()` middleware, then a handler that builds a `Response` and copies it onto the context with the per-header `ctx.set()` loop that `setContextResponse` uses:

```
206 (createResponse-shaped handler)
  status              : 206
  allow-origin        : *
  expose-headers      : X-Boxel-Realm-Url,ETag,Retry-After,Content-Range,Accept-Ranges,Content-Length
  -> Location present in expose list? false

202 (plain-Response handler, no expose list of its own)
  expose-headers      : ETag, Location, Retry-After, Content-Range, Accept-Ranges, Content-Length
  -> Location present in expose list? true
```

The first case is the replace-not-merge claim from the description, reproduced: the handler's own list wins outright, `Location` (present only in the middleware's list) is gone. `@koa/cors` sets `Access-Control-Expose-Headers` at `index.js:95-96` and reaches `await next()` at `index.js:105`/`108`, so the middleware's value is always in place before the handler runs and always overwritten afterward for any header the `Response` carries. Updating this list was mandatory, not defensive — every `206` and the `416` in `Realm`'s file-serving path is built with `createResponse`, so this is the list that reaches the wire on a media range response.

**The fix — reword so the list reads as load-bearing:**

```suggestion
      // Content-Range/Accept-Ranges must be exposed because the host's auth
      // service worker hands the CORS-filtered Response from its own fetch()
      // straight to the media element via respondWith. A header absent from
      // this list is pruned from that filtered Response, so it is invisible
      // to the player's loading stack, not just to app JS. Content-Length is
      // CORS-safelisted and survives regardless; it is named alongside the
      // pair so the range trio reads as one unit.
      'Access-Control-Expose-Headers':
        'X-Boxel-Realm-Url,X-Boxel-Realm-Public-Readable,X-Boxel-Realm-Archived,X-Boxel-Canonical-Path,X-Boxel-Not-Ready,Authorization,Cache-Control,ETag,Retry-After,Content-Range,Accept-Ranges,Content-Length',
```

**Scope:** confirmation of the change plus a wording improvement. Non-blocking — the header list itself needs no edit.

---

packages/realm-server/server.ts
```diff
@@ -1121,7 +1128,11 @@ export class RealmServer {
           // protocol invisible to JS. Location/Retry-After are likewise
           // non-safelisted; expose them so a cross-origin client can read
           // the async-publish status monitor target off the 202 response.
-          exposeHeaders: 'ETag, Location, Retry-After',
+          // Content-Range/Accept-Ranges are what a cross-origin caller needs
+          // to reason about a 206 byte-range response (Content-Length is
+          // safelisted, but is listed for symmetry with the range pair).
+          exposeHeaders:
+            'ETag, Location, Retry-After, Content-Range, Accept-Ranges, Content-Length',
```
habdelra, line 1135: [Claude Code 🤖] **Confirmation — `Location` is in this list but not in `createResponse`'s, which looks like precisely the silent drop that `create-response.ts`'s comment warns about. It isn't. Documenting why, so the asymmetry isn't "fixed" and the invariant isn't dismissed as wrong. Non-blocking, nothing to change.**

**Why it looks broken.** The comment two lines up says these are exposed "so a cross-origin client can read the async-publish status monitor target off the 202 response," while `create-response.ts` states the rule that "anything koa means to expose has to be repeated here or it is silently dropped." Read together, `Location` appearing only here reads like a live bug: the 202's `Location` should be getting stripped.

**Why it isn't.** The rule is narrower than "the handler's headers replace koa's wholesale." `setContextResponse` (`packages/realm-server/middleware/index.ts`) iterates only the headers **present on the handler's `Response`** and calls `ctxt.set(header, value)` for each — a per-header overwrite, not a wipe. So which list survives depends on whether the response carries an expose list of its own:

- The async-publish 202 in `packages/realm-server/handlers/handle-publish-realm.ts` is built with a plain `new Response(...)`, not `createResponse`. It carries no `Access-Control-Expose-Headers`, so the middleware's value survives untouched and `Location` **is** exposed. The comment here is accurate.
- The `createResponse` call sites that do stamp `Location` are all 302s — the canonical-path and fallback-handle redirects in `Realm`. Those don't need it exposed either way: under the default `redirect: 'follow'` the browser consumes `Location` and JS never sees the intermediate response at all, and under `redirect: 'manual'` a cross-origin response is an `opaqueredirect` whose header list is empty regardless of what any expose list says.

**Verified**, with a Koa app mirroring this shape (`@koa/cors@4.0.0`, per-header `ctx.set()` copy):

```
206 (createResponse-shaped handler — carries its own expose list)
  expose-headers      : X-Boxel-Realm-Url,ETag,Retry-After,Content-Range,Accept-Ranges,Content-Length
  -> Location present in expose list? false

202 (plain-Response handler — no expose list of its own)
  expose-headers      : ETag, Location, Retry-After, Content-Range, Accept-Ranges, Content-Length
  -> Location present in expose list? true
```

**The rule worth carrying forward,** which is what makes adding the range trio to *both* lists correct rather than redundant: a response built through `createResponse` gets `createResponse`'s list and nothing else; a response built with a plain `new Response` keeps the middleware's; and anything that never reaches a handler at all — the preflight, koa-level 404s and errors — also gets the middleware's. Adding to one list only would have covered one of those three populations.

**Scope:** confirmation. No change requested. Adding `Location` to `createResponse`'s list would be harmless but would buy nothing, since no response that needs it goes through there.

---

packages/realm-server/tests/cors-range-test.ts
habdelra, line 1: [Claude Code 🤖] **This module stands up a realm fixture that an existing test module already stands up with a byte-identical configuration, to add two assertions. They belong in that module. Non-blocking, but cheap to act on.**

**The duplication is exact.** `packages/realm-server/tests/range-request-test.ts` opens with the same `realmURL`, the same `fixture: 'simple'`, and the same `permissions` map. Diffing the two setup blocks, the only delta is this module additionally capturing `args.request` into `serverRequest`:

```
2a3
>     let serverRequest: SuperTest<Test>;
11a13
>         serverRequest = args.request;
```

That one line is all this file needs from a separate module — and `range-request-test.ts`'s existing `onRealmSetup` could capture it just as easily.

**What the existing module already provides that this one re-derives, more weakly:**

- `uploadSample(path)` — asserts the upload returned `204`. The POST here asserts nothing, so if writing the sample ever broke, this test fails at the `206` assertion on the subsequent GET and reads as a CORS regression rather than a broken fixture.
- `binaryParser` / `getSample` — a supertest parser that collects raw bytes instead of letting a text/JSON parser get picked from the content type.
- Comprehensive coverage of the range machinery itself: `Content-Range` values, exact body slices, `416`, suffix and open-ended ranges, last-byte clamping, multi-range fallback, `If-Range` match and stale, and `If-None-Match` precedence over `Range`. Which is why the two tests here are right to assert only the CORS surface — that part isn't duplicated.

**No collision blocks the move:** `range-request-test.ts` names a distinct file per test (`full.png`, `bounded.png`, `suffix.png`, …) and never uses `sample.png`.

**Cost of keeping them split.** `setupPermissionedRealmCached` derives its template-database cache key from the fixture, realm URL, permissions and prerenderer, so with configurations this identical the expensive index build *is* shared across the two modules — the waste is not the index build. What isn't shared is the per-module `setupPermissionedRealm` realm boot and teardown, paid once more per run against a realm-server suite that already runs ~16 minutes.

**Scope:** follow-up / test organization. No behavioral risk either way, and nothing here blocks the change — the CORS assertions are correct where they sit.

---

packages/realm-server/tests/cors-range-test.ts
```diff
@@ -0,0 +1,91 @@
...
+          .set('Content-Type', 'application/octet-stream')
+          .send(Buffer.from(new Uint8Array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])));
+
+        let response = await request
+          .get('/sample.png')
+          .set('Accept', 'image/*')
+          .set('Origin', 'https://app.example')
+          .set('Range', 'bytes=2-5');
+
+        assert.strictEqual(response.status, 206, 'HTTP 206 status');
```
habdelra: [Claude Code 🤖] **This test asserts the expose list without asserting the one precondition that makes an expose list mean anything — so the exact regression it exists to catch could land and it would still pass. Non-blocking, two lines to close.**

**The gap.** `Access-Control-Expose-Headers` is inert on a response that is not a CORS response: with no `Access-Control-Allow-Origin`, the browser rejects the response before any expose list is consulted, and every named header is unreadable anyway. This test names `Origin` on the request and then verifies the expose list, but never checks that `Access-Control-Allow-Origin` came back — so it pins half the contract and reads as pinning all of it.

**Why that half is fragile rather than obviously safe.** `ACAO` on this path comes from the `@koa/cors` middleware, and it survives only because `createResponse` happens not to stamp an `Access-Control-Allow-Origin` of its own. `setContextResponse` overwrites per header, so the day `createResponse` grows an `ACAO` — or any handler in front of it sets a specific origin — the middleware's `*` is replaced and cross-origin reads of `Content-Range` break, while this test keeps passing because the expose list it inspects is untouched. That is precisely the failure mode the file was written to guard.

**Verified**, with a Koa app mirroring this shape (`@koa/cors@4.0.0`, `cors()` then a handler whose `Response` headers are copied on with a per-header `ctx.set()` loop):

```
206 (createResponse-shaped handler)
  status              : 206
  allow-origin        : *
  expose-headers      : X-Boxel-Realm-Url,ETag,Retry-After,Content-Range,Accept-Ranges,Content-Length
```

`allow-origin: *` survives today — worth pinning while it does.

**Also unasserted:** the POST's status. `range-request-test.ts`'s `uploadSample` asserts `204` on the write; without that, a broken fixture write surfaces as a failed `206` assertion and looks like a CORS regression.

**The fix:**

```suggestion
      test('a ranged response exposes its range headers to cross-origin JS', async function (assert) {
        let upload = await request
          .post('/sample.png')
          .set('Content-Type', 'application/octet-stream')
          .send(Buffer.from(new Uint8Array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])));
        assert.strictEqual(upload.status, 204, 'sample file uploaded');

        let response = await request
          .get('/sample.png')
          .set('Accept', 'image/*')
          .set('Origin', 'https://app.example')
          .set('Range', 'bytes=2-5');

        assert.strictEqual(response.status, 206, 'HTTP 206 status');
        assert.strictEqual(
          response.headers['access-control-allow-origin'],
          '*',
          'the response is a CORS response, without which no expose list is consulted',
        );
```

**Scope:** test strength. Non-blocking — the production change it covers is correct; this is about the test failing when it should.

---

habdelra (commented):
[Claude Code 🤖] **Review lens:** the change itself is unchanged since the earlier pass — byte-for-byte — so this pass re-verified the CORS contract from scratch against the moved base (which now carries the byte-range file serving this depends on), and went after what the earlier pass didn't: whether the *reason* recorded in the comments matches the mechanism that actually consumes these headers, and whether the new tests fail when they should.

**Bottom line: no blocking issues.** The header lists are correct in both homes and the replace-not-merge reasoning in the description reproduces exactly. Four non-blocking comments: two confirmations that document why the code is right (including one asymmetry that looks like a bug and isn't), one comment-wording suggestion, and two test observations.

**On the rebase.** Comparing each tip against its own parent, the diff is identical to what was reviewed before — same three files, same 108/3. Only the base moved, onto a main where byte-range serving is merged. Re-checked the contract against that base rather than assuming it carried over: every `206` and the `416` in `Realm`'s file-serving path is built with `createResponse`, so `create-response.ts` remains the list that reaches the wire on a media range response.

**What lands right (verified, not inherited).** I re-derived the middleware behavior with a standalone Koa app on `@koa/cors@4.0.0` — `cors()` middleware, then a handler that builds a `Response` and copies it onto the context with the same per-header `ctx.set()` loop `setContextResponse` uses:

- **Replace-not-merge is real, and it is per-header.** `@koa/cors` sets `Access-Control-Expose-Headers` at `index.js:95-96` and only reaches `await next()` at `index.js:105`/`108`, so the middleware's value is always in place before the handler runs. `setContextResponse` then iterates only the headers present on the handler's `Response` and `ctxt.set()`s each — so a response carrying its own expose list replaces the middleware's outright, and a response without one keeps it. Updating `createResponse` was mandatory; updating the middleware's list covers the responses that never touch a `createResponse` handler.
- **The preflight list is final.** For OPTIONS, `@koa/cors` returns `204` with `Access-Control-Allow-Headers` echoing the configured list verbatim and never invokes the handler — nothing downstream can drop `Range`/`If-Range`. Confirmed in the repro alongside `Access-Control-Max-Age: 86400`.
- **Twin check came back clean, independently.** The only other sites that set CORS headers are `proxy-forward.ts` (`setupSSEHeaders`, `text/event-stream` only), `matrix-backend-authentication.ts` (exposes `Authorization` alone), and the host dev-CORS middleware. None serves realm file bytes. There is no static-file middleware in the realm server that could emit a `206` outside `createResponse`. The `--cors=…,Range,…` list on the icon static server is a separate surface serving built icon assets and needs nothing from this change.
- **No deploy-time lag from the 24 h preflight cache**, which is a fair thing to wonder about given `maxAge: 86400`. Per the Fetch spec the CORS-preflight cache is keyed per header *name*, and a request carrying a name with no cache entry forces a fresh preflight — so a newly allowed `Range` can't be answered from a pre-existing entry. Spec-derived, not measured here.

**CI.** Realm-server tests are green (2,174 passing, 0 failing) including the two new ones. One host shard is red, and it emitted no junit at all — it died before test output rather than on an assertion, so it reads as infrastructure. Nothing in this diff can reach host tests: both changes are strictly additive CORS *permissions*, an allow list and an expose list, and neither can turn a request that previously succeeded into a failure. Worth a re-run rather than an investigation.

**Recommendations, all non-blocking:**

1. Reword the `Content-Range`/`Accept-Ranges` rationale in `create-response.ts` — the real consumer is the media element's own loading stack, not app JS, because the service worker returns the CORS-*filtered* `Response` to it via `respondWith`. Stronger reason, and the weak version is what invites a future trim. See the thread on `create-response.ts`.
2. Assert `Access-Control-Allow-Origin` in the `206` test — an expose list is inert without it, so today the test pins half its contract. Also assert the upload's `204`. See the thread on the ranged-response test.
3. Fold the two CORS tests into `range-request-test.ts`, whose setup block is byte-identical and which already has `uploadSample`/`binaryParser` and no `sample.png`. See the file-level thread on `cors-range-test.ts`.
4. Nothing to change for the `Location` asymmetry between the two expose lists — it reads like the silent drop the comments warn about but is safe, and the thread on `server.ts` records why so it isn't "fixed" and the invariant isn't dismissed.

**Adjacent, out of scope.** `packages/host/public/auth-service-worker.js` still documents itself as being for `` and CSS `background-image`, while the header comments added here correctly describe it as the path native `<audio>`/`<video>` depend on. Whoever next touches that file should widen its opening comment to match. Also agreed on deferring browser-level end-to-end verification of protected-realm playback to the player integration work — this PR's level is the right altitude for the CORS contract itself.

---

habdelra (approved):

packages/runtime-common/create-response.ts
lukemelia: [Claude Code 🤖] Applied in d718a693b8 — the comment now states the load-bearing consumer: the auth service worker hands the CORS-filtered Response from its own fetch() to the media element via respondWith, so a header absent from the expose list is pruned from the Response the player's loading stack inspects, not just hidden from app JS. Verified the mechanism against `packages/host/public/auth-service-worker.js` (`mode: 'cors'` upgrade in buildAuthedRequest, `event.respondWith(fetch(...))` in the fetch handler) before adopting the wording.

packages/realm-server/server.ts
lukemelia: [Claude Code 🤖] Agreed, nothing to change. The distinction worth keeping: setContextResponse overwrites per header present on the handler's Response, so a plain `new Response` (the async-publish 202) keeps the middleware's expose list with `Location`, while a `createResponse` response carries its own list — and the only `createResponse` call sites stamping `Location` are 302s the browser consumes before JS ever sees them. Left `Location` out of createResponse's list since no response that needs it goes through there.

packages/realm-server/tests/cors-range-test.ts
lukemelia: [Claude Code 🤖] Done in d718a693b8 — the two CORS tests now live in `range-request-test.ts` on the shared fixture, the separate module is deleted, and the exposure test uploads through `uploadSample` so a broken fixture write fails as a 204 assertion instead of masquerading as a CORS regression.

Acting on this surfaced something bigger: the realm-server suite loads an explicit module list in `tests/index.ts`, and neither `range-request-test` nor `http-range-test` was ever registered there — so none of the byte-range or CORS coverage was actually running in CI. Both modules are registered now, and all 19 tests pass locally via `TEST_FILES='./http-range-test,./range-request-test'`.

packages/realm-server/tests/cors-range-test.ts
lukemelia: [Claude Code 🤖] Fixed in d718a693b8. The test (now in `range-request-test.ts`) asserts `access-control-allow-origin: *` on the 206 before inspecting the expose list — pinning that the middleware's ACAO survives `setContextResponse`'s per-header overwrite, which is the precondition that makes the expose list mean anything. The upload goes through `uploadSample`, which asserts the 204.

PR #5800, opened 2026-08-17
Match type filters through re-export spellings via canonical code refs

lukemelia:
## What this does

A `filter.type`/`on` code ref that names a type through a re-exporting module — the canonical example being base `FileDef` addressed via its documented `file-api` module, which re-exports it from `card-api` — silently matched nothing. Index rows stamp `types` with the canonical (defining-module) key from `identifyCard`, so only the canonical spelling ever joined the membership keys: filtering on base `FileDef` returned zero rows while concrete subtype filters matched, making the base-type filter a silent dead end.

Two halves, kept in agreement:

- **Query engine**: each type condition's ref now resolves through the definition lookup, and the compiled `types` membership predicates union the canonical ref's spelling-tolerant keys (RRI / real-URL / virtual-alias) with the as-given ref's. Because the definition lookup is async, the type condition becomes a deferred expression node resolved in pass 1, alongside the existing deferred kinds. A ref whose definition doesn't resolve keeps only its spelling-based keys and matches nothing, exactly as before.
- **Host search resource**: the same rewrite is applied (via the loader + `identifyCard`) before client-side matching — the matcher compares refs against `identifyCard` of an instance's class, so an uncanonicalized re-export spelling would strip the server's correct results during live reconciliation. Until the rewrite settles for the active filter, or when a ref doesn't resolve, the search stays a server-only passthrough so the two evaluations can't disagree.

The shared walker (`runtime-common/query-canonicalization.ts`) rewrites `type`/`on` refs through `any`/`every`/`not` without mutating the input tree, and single-flights duplicate ref spellings by memoizing the in-flight promise — sibling nodes are walked concurrently, so a value-memo would double-resolve.

## Test plan

- `packages/realm-server/tests/search-entries-engine-test.ts`: a pure base-FileDef filter matches every file row across subtypes, and the `file-api` re-export spelling matches the same rows as the canonical `card-api` spelling (36/36 pass locally).
- `packages/host/tests/unit/query-canonicalization-test.ts`: rewrite of top-level and nested refs, input-tree immutability, unresolvable refs marking the result incomplete, and duplicate refs resolving once (8/8 pass locally via the dev test page).
- Typechecks pass across runtime-common, realm-server, and host.

### DIFF
--- a/packages/host/app/resources/search.ts
+++ b/packages/host/app/resources/search.ts
@@ -27,6 +27,9 @@ import {
   isFileDefInstance,
   isFileDefCodeRef,
   isClientEvaluable,
+  canonicalizeFilterRefs,
+  identifyCard,
+  isResolvedCodeRef,
   matchInstanceAgainstFilter,
   makeInstanceComparator,
   logger as runtimeLogger,
@@ -44,6 +47,7 @@ import type { Filter, Query } from '@cardstack/runtime-common/query';
 
 import { searchErrorEntry } from '../lib/search-error-entry';
 
+import type LoaderService from '../services/loader-service';
 import type NetworkService from '../services/network';
 import type RealmServerService from '../services/realm-server';
 import type StoreService from '../services/store';
@@ -162,6 +166,7 @@ export class SearchResource<
   @service declare private network: NetworkService;
   @service declare private realmServer: RealmServerService;
   @service declare private store: StoreService;
+  @service declare private loaderService: LoaderService;
   #storeServiceOverride: StoreService | undefined;
   @tracked private realmsToSearch: RealmIdentifier[] = [];
   // Resist the urge to expose this property publicly as that may entice
@@ -183,6 +188,19 @@ export class SearchResource<
   // once it lands.
   @tracked private matchAPI: CardAPIForMatching | undefined;
   #matchAPILoading = false;
+  // The active filter with its `type`/`on` refs rewritten to canonical
+  // (defining-module) form, mirroring the rewrite the query engine applies
+  // before compiling `types` membership predicates. The matcher compares refs
+  // against `identifyCard` of an instance's class — the canonical ref — so a
+  // filter carrying a re-export spelling (e.g. file-api's FileDef) must be
+  // canonicalized here too or the reconciler would strip the server's correct
+  // results. `source` ties the rewrite to the exact filter object it was
+  // computed from; `incomplete` means some ref didn't resolve, in which case
+  // the search stays a server-only passthrough rather than risking a client
+  // evaluation the server disagrees with.
+  @tracked private canonicalizedFilter:
+    | { source: Filter; filter: Filter; incomplete: boolean }
+    | undefined;
   // The query currently driving results, tracked so the client filtering step
   // re-derives when the filter/sort changes (the server result set also
   // changes, but reading this keeps the derivation self-contained).
@@ -234,6 +252,53 @@ export class SearchResource<
       })
       .finally(() => {
         this.#matchAPILoading = false;
+        waiter.endAsync(token);
+      });
+  }
+
+  // Canonicalizes the live filter's `type`/`on` refs via the loader (import
+  // the named module, identify the export's defining-module ref) — the same
+  // rewrite the query engine performs through its definition lookup. Until
+  // this settles for the current filter, `isClientFilterEligible` keeps the
+  // search a server-only passthrough.
+  private loadCanonicalizedFilter(filter: Filter | undefined): void {
+    if (!filter) {
+      this.canonicalizedFilter = undefined;
+      return;
+    }
+    if (this.canonicalizedFilter?.source === filter) {
+      return;
+    }
+    let token = waiter.beginAsync();
+    canonicalizeFilterRefs(filter, async (ref) => {
+      try {
+        let module = await this.loaderService.loader.import<
+          Record<string, unknown>
+        >(ref.module);
+        let canonical = identifyCard(
+          module[ref.name] as Parameters<typeof identifyCard>[0],
+        );
+        return canonical && isResolvedCodeRef(canonical)
+          ? canonical
+          : undefined;
+      } catch (error) {
+        this.#log.warn(
+          `could not canonicalize filter ref ${ref.module}/${ref.name}: ${error}`,
+        );
+        return undefined;
+      }
+    })
+      .then(({ filter: canonical, incomplete }) => {
+        if (isDestroyed(this) || isDestroying(this)) {
+          return;
+        }
+        this.canonicalizedFilter = {
+          source: filter,
+          filter: canonical,
+          incomplete,
+        };
+      })
+      .finally(() => {
         waiter.endAsync(token);
       });
   }
@@ -356,6 +421,7 @@ export class SearchResource<
       // matcher dependencies eagerly so the first eligible result set can be
       // reconciled without waiting on a Store mutation to trigger it.
       this.loadMatchAPI();
+      this.loadCanonicalizedFilter(query.filter);
     }
     this.activeQuery = query;
     this.#doWhileRefreshing = doWhileRefreshing;
@@ -527,7 +593,9 @@ export class SearchResource<
       return serverInstances;
     }
     let api = this.matchAPI!;
-    let filter = this.activeQuery?.filter;
+    // Eligibility (below) guarantees the canonicalized form is present and
+    // complete whenever the active query carries a filter.
+    let filter = this.canonicalizedFilter?.filter ?? this.activeQuery?.filter;
 
     // Reading the mutation-version signal here establishes the dependency that
     // re-derives on in-place field edits/saves (adds and deletes already flow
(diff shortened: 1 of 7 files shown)


packages/host/app/resources/search.ts
```diff
@@ -238,6 +256,53 @@ export class SearchResource<
...
+    })
+      .then(({ filter: canonical, incomplete }) => {
+        if (isDestroyed(this) || isDestroying(this)) {
+          return;
+        }
+        this.canonicalizedFilter = {
+          source: filter,
+          filter: canonical,
+          incomplete,
+        };
```
lukemelia, line 308: [Claude Code 🤖] **Out-of-order canonicalization leaves a stale `source` and stickily disables client reconciliation** (regression introduced by this PR; non-blocking — it degrades to a safe server-only passthrough, but silently and until the next query change).

**The mechanism.** `loadCanonicalizedFilter` runs on every live `modify()` (`search.ts` :424, before `this.activeQuery = query` at :426). Its early-return guard (`this.canonicalizedFilter?.source === filter`, :269) compares only against the *settled* value, so while a first canonicalization for filter A is still in flight, a second `modify()` carrying filter B (a new object) passes the guard and starts a *second* concurrent `canonicalizeFilterRefs`. This `.then` then writes `canonicalizedFilter = { source: filter, … }` unconditionally, so whichever promise settles **last** wins.

The two resolutions are `loader.import(ref.module)` calls: A triggers the cold import (possibly a network fetch); B, arriving later, resolves the same module from the loader cache almost immediately. So B settles first and sets `source: B`, then A settles and clobbers it with `source: A`. Now `this.activeQuery.filter === B` but `this.canonicalizedFilter.source === A`.

**The consequence.** `isClientFilterEligible` requires `canonicalized.source === filter` (:712–717), so a stale `source` makes it return `false` and `displayedInstances` returns the server set untouched (:589–593). Results stay correct, but live client-side reconciliation (candidate add + local `no-match` removal) stays off until the *next* `modify()` whose canonicalization happens to win the race — a live-refresh realm event calls `this.search.perform(this.#previousQuery)` without re-running `modify()`, so nothing re-triggers it in between. Sticky.

**The fix.** Drop a stale result in the `.then`. It fires synchronously after `this.activeQuery = query` is set within the same `modify()`, so `activeQuery.filter` is already the latest by the time any resolution lands:

```suggestion
      .then(({ filter: canonical, incomplete }) => {
        if (isDestroyed(this) || isDestroying(this)) {
          return;
        }
        // A newer filter may have become active while this resolved; drop the
        // stale result so a late resolution can't clobber the current one.
        if (this.activeQuery?.filter !== filter) {
          return;
        }
        this.canonicalizedFilter = {
          source: filter,
          filter: canonical,
          incomplete,
        };
      })
```

**Related.** Because the guard and memo key on filter *object identity*, an unstable `() => query` thunk that rebuilds a deeply-equal filter on each recompute re-triggers canonicalization (and a transient passthrough window) on every `modify()`, even though the `querySignature` deep-equal check at :537 already skips the redundant *search*. Keying eligibility off `querySignature` rather than identity would also close that gap — but the staleness guard above is the load-bearing fix.

Non-blocking.

packages/runtime-common/index-query-engine.ts
```diff
@@ -1118,22 +1120,62 @@ export class IndexQueryEngine {
...
+            this.#virtualNetwork,
+          )) {
+            if (!keys.includes(key)) {
+              keys.push(key);
+            }
+          }
+        }
+      } catch {
+        // fall through to the spelling-based keys
+      }
```
lukemelia: [Claude Code 🤖] **`typeKeysFor` swallows every `lookupDefinition` error, so a transient definition-lookup failure silently under-matches** (follow-up / non-blocking; the not-found case matches prior behavior).

This `catch {}` treats all failures the same — fall back to spelling-only keys. For a genuinely nonexistent type that's correct and matches the pre-existing behavior (a bare `{ type: X }` filter never validated existence). But `lookupDefinition` can also throw *transiently* — a prerender timeout or module-cache miss under load, a DB hiccup — and on a **valid** re-export ref (`file-api`'s `FileDef`) that means the canonical (`card-api`) key never joins `keys`, so the query silently returns *fewer* rows (the canonical-stamped rows drop out) with no error surfaced. The engine raises `FilterRefersToNonexistentTypeError` for field-path lookups elsewhere; a type condition now carries the same async dependency but degrades silently instead.

Not this PR's blocker: the failure is transient and self-heals on retry, and the canonical-key expansion is what's new here rather than the whole type filter. Worth considering whether to distinguish "definition not found" (fall through, match nothing — fine) from an unexpected lookup error (surface it, as the field-path walk does). Non-blocking.

lukemelia (commented):
[Claude Code 🤖] **Review focus: the two-sided canonical-ref contract** — that the server's `types`-membership compile and the host's client-side matcher agree on what a re-export spelling matches, and that the new deferred `type-condition` node composes exactly as the inline `any()` it replaces.

**Bottom line: the design is correct — I found no correctness defect in the compiled SQL or in the server/matcher agreement. Two non-blocking findings: a stale-source race in the host's async canonicalization, and a swallow-all error path in `typeKeysFor`. Details inline.** No blocking issues.

**What lands right.**

- The mechanism checks out end-to-end. Index rows stamp `types` via `identifyCard` (the defining-module ref — `makeDefinition` in `host/app/routes/module.ts` sets `codeRef = identifyCard(cardOrFieldDef)`). The matcher's `instanceIsType` walks the instance's prototype chain comparing `identifyCard(klass)` through `codeRefEquals`, which tolerates URL-*spelling* variance (RRI / real-URL / virtual-alias) but **not** module-*identity* variance. So a raw `file-api`/`FileDef` filter makes `instanceIsType` return false and the reconciler strips the server's correct rows. Canonicalizing the filter's refs to the `card-api` defining-module ref on **both** sides is exactly the right fix, and doing it upstream in `search.ts` rather than inside the matcher keeps that single call site as the only consumer needing it (verified: `search.ts` is the only non-test caller of `matchInstanceAgainstFilter` / `isClientEvaluable`).
- The deferred `type-condition` node is structurally equivalent to the old inline path. `filterCondition` still wraps `this.typeCondition(ref)` (now `[typeConditionNode]`) through `every`/`any`, whose `addExplicitParens` parenthesizes the single-element array; pass 1 splices `handleTypeCondition`'s `any(keys.map(typesContains))` — the same token shape the old code emitted — inside those parens. All three type-key sites (`handleTypeCondition`, `hasFileType`, `hasInstanceType`) route through `typeKeysFor`, so no twin is left on the old inline path.
- `typeKeysFor` mutating `internalKeysFor`'s return with `.push` is safe — `internalKeysFor` returns a fresh `.map(...)` array per call.
- The canonicalization walker's single-flight memo is correct: `canonicalRef` runs `memo.has` / `memo.set` synchronously before its first `await`, so concurrent sibling walks can't both miss (the `duplicate refs resolve once` test pins it), and the per-node `{ ...node }` shallow clone keeps the input tree immutable (also pinned).

**Recommendations.**

1. Guard the host `.then` against a stale filter so an out-of-order `loader.import` resolution can't stickily disable client reconciliation — see the `loadCanonicalizedFilter` thread. (non-blocking)
2. Consider distinguishing "definition not found" from a transient lookup failure in `typeKeysFor`, so a valid re-export ref doesn't silently under-match under load — see the `typeKeysFor` thread. (non-blocking)

**Adjacent, out of scope.** The realm test-results status comment reports realm-server tests as ±0 against the base even though this PR adds two tests to `search-entries-engine-test.ts` — most likely a status-comment/base-comparison artifact rather than the tests not running, but worth a glance to confirm they executed in CI.

packages/host/app/resources/search.ts
lukemelia: [Claude Code 🤖] Fixed in `2a58eb1006`. The `.then` now drops its result when `this.activeQuery?.filter !== filter`, so a late resolution (a cold `loader.import` settling after a cache-warm sibling) can no longer overwrite `canonicalizedFilter` with a `source` that points at a filter that is no longer active. The guard is sound because the `.then` fires only after `this.activeQuery = query` has run synchronously within the same `modify()`, so the comparison always sees the current filter.

The related identity-keying note (an unstable `() => query` thunk re-triggering canonicalization) is left as-is — the staleness guard makes the transient passthrough self-correct, and moving eligibility onto `querySignature` is a larger change than this finding warrants.

packages/runtime-common/index-query-engine.ts
lukemelia: [Claude Code 🤖] Fixed in `2a58eb1006`. The blanket `catch {}` is now `catch (error)` that re-throws unless `isFilterRefersToNonexistentTypeError(error)`. A ref resolving to no known type still falls through to the spelling-based keys — matching how the engine already treats a nonexistent type as an empty result (the top-level `_search` catch), so the intentional spelling-key fallback and the transient-module-error behavior are unchanged. What changes is that a genuinely unexpected error (a bug, not a missing type) now propagates instead of being silently swallowed and narrowing the match.

Note the engine deliberately treats a `FilterRefersToNonexistentType` — which wraps transient module/prerender errors too — as match-nothing everywhere (field-path lookups included), so this keeps `typeKeysFor` consistent with that convention rather than special-casing transient failures here.

backspace (approved):

PR #5802, opened 2026-08-17
Give every FileDef subtype module a default export; name the field in thunk errors

lukemelia:
## What this does

The FileDef subtype family had two export conventions: the shell-based modules export their def class both named and as the default, while 17 older leaf modules (the text, image, and audio families — `markdown-file-def`, `png-image-def`, `mp3-audio-def`, …) were named-only. A default import of a named-only module silently evaluates to `undefined`, and `linksTo(TsFileDef)` then fails far from the import at schema time with an error blaming a cyclic dependency — naming neither the field, the module, nor the actual cause. Authors had to probe `Object.keys(ns)` to find out which import form a given module wanted.

- All 17 named-only subtype modules now also `export default` their primary class, matching the convention the rest of the family already follows, so both import forms work everywhere.
- `cardThunk` receives the field name and owning prototype from every field initializer (`contains`/`containsMany`/`linksTo`/`linksToMany`); the parameter is required, so a future call site can't silently degrade the message. An undefined field card class throws an error naming the exact declaration — `The card class for field 'src' on 'MyCard' was undefined` — and describes both real causes: an import that doesn't match the module's export shape, and a genuine cycle needing the thunk form. (The offending module itself isn't knowable at that point — the import has already evaluated to `undefined` by the time the value reaches the field machinery.)
- The thunk form gets the same guard: a thunk that resolves to `undefined` throws the same field-and-owner-named error at the first `field.card` read, instead of an anonymous `TypeError: Cannot use 'in' operator…` later in `getFieldDefinitions`. This matters because cause (2) of the message advises switching to the thunk form — that path now fails just as loudly and specifically.
- `FILEDEF_CODE_REF_BY_EXTENSION` is exported from runtime-common (as `Readonly` — every consumer is a read) so the export-shape guard test can walk the registry and automatically cover subtypes added later.

## Test plan

- `packages/host/tests/unit/filedef-export-shape-test.ts` — walks every registered subtype extension plus the six family base modules a subtype author imports from (`image-file-def`, `audio-file-def`, `video-file-def`, `font-file-def`, `three-d-model-def`, `file-api`), loads each module, and asserts the named export exists, a default export exists, and (for single-class modules) that the two are the same class. What this pins is "a default import never yields `undefined`"; for the two modules carrying a second named-only class (`image-file-def`, `zip-file-def`) only the registered class is pinned to the default. 117 assertions, passing locally against a full dev stack; the local base realm also completed a from-scratch index over the modified modules.
- `packages/host/tests/unit/card-thunk-error-test.ts` — drives the `field` decorator exactly as Babel's transform does through all four field types, asserting the error names the field, the owner, and both causes; covers the eager form (throws at decoration) and the thunk form (throws at first `field.card` read).
- Typechecks pass across runtime-common and host; template-lint clean on the touched `.gts` files.

### DIFF
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -4794,23 +4794,37 @@ function cardThunk<CardT extends BaseDefConstructor>(
   // value came from isn't knowable here — by the time the value reaches us
   // the import has already evaluated to undefined — so the message names the
   // two ways that happens instead.
-  fieldContext?: { fieldName: string; ownerPrototype: BaseDef },
+  fieldContext: { fieldName: string; ownerPrototype: BaseDef },
 ): () => CardT {
-  if (!cardOrThunk) {
-    let fieldDescription = fieldContext
-      ? `field '${fieldContext.fieldName}' on '${
-          fieldContext.ownerPrototype.constructor?.name ?? 'unknown card'
-        }'`
-      : 'a field';
-    throw new Error(
-      `The card class for ${fieldDescription} was ${cardOrThunk}. Two common causes:
+  // `||` rather than `??`: an anonymous class's `.name` is `''`, not nullish.
+  let fieldDescription = `field '${fieldContext.fieldName}' on '${
+    fieldContext.ownerPrototype.constructor?.name || 'unknown card'
+  }'`;
+  let causes = `Two common causes:
       (1) the import doesn't match the module's export shape — e.g. \`import X from '…'\` where the module only has a named export; use \`import { X } from '…'\`;
-      (2) a cyclic dependency between cards — use the thunk form in all cards in the cycle, e.g. '@field friend = linksTo(() => Person)'.`,
+      (2) a cyclic dependency between cards — use the thunk form in all cards in the cycle, e.g. '@field friend = linksTo(() => Person)'.`;
+  if (!cardOrThunk) {
+    throw new Error(
+      `The card class for ${fieldDescription} was ${cardOrThunk}. ${causes}`,
     );
   }
-  return (
-    'baseDef' in cardOrThunk ? () => cardOrThunk : cardOrThunk
-  ) as () => CardT;
+  if ('baseDef' in cardOrThunk) {
+    return () => cardOrThunk as CardT;
+  }
+  // The thunk form fails the same two ways the eager form does, just later —
+  // at the first `field.card` read instead of at decoration. The accessors
+  // re-invoke the thunk on every read (nothing is memoized), so this adds a
+  // truthiness check per read, not an extra call.
+  let thunk = cardOrThunk as () => CardT;
+  return () => {
+    let card = thunk();
+    if (!card) {
+      throw new Error(
+        `The card class thunk for ${fieldDescription} returned ${card}. ${causes}`,
+      );
+    }
+    return card;
+  };
 }
 
 export type SignatureFor<CardT extends BaseDefConstructor> = {
--- a/packages/base/card-api.gts
+++ b/packages/base/card-api.gts
@@ -2397,10 +2397,10 @@ export function containsMany<FieldT extends FieldDefConstructor>(
   options?: Options,
 ): BaseInstanceType<FieldT>[] {
   return {
-    setupField(fieldName: string, _ownerPrototype: BaseDef) {
+    setupField(fieldName: string, ownerPrototype: BaseDef) {
       let { computeVia, searchable } = options ?? {};
       let instance = new ContainsMany({
-        cardThunk: cardThunk(field),
+        cardThunk: cardThunk(field, { fieldName, ownerPrototype }),
         computeVia,
         name: fieldName,
         searchable,
@@ -2417,10 +2417,10 @@ export function contains<FieldT extends FieldDefConstructor>(
   options?: Options,
 ): BaseInstanceType<FieldT> {
   return {
-    setupField(fieldName: string, _ownerPrototype: BaseDef) {
+    setupField(fieldName: string, ownerPrototype: BaseDef) {
       let { computeVia, searchable } = options ?? {};
       let instance = new Contains({
-        cardThunk: cardThunk(field),
+        cardThunk: cardThunk(field, { fieldName, ownerPrototype }),
         computeVia,
         name: fieldName,
         searchable,
@@ -2439,7 +2439,10 @@ export function linksTo<CardT extends LinkableDefConstructor>(
   return {
     setupField(fieldName: string, ownerPrototype: BaseDef) {
       let { computeVia, searchable, query } = options ?? {};
-      let fieldCardThunk = cardThunk(cardOrThunk);
+      let fieldCardThunk = cardThunk(cardOrThunk, {
+        fieldName,
+        ownerPrototype,
+      });
       if (query) {
         validateRelationshipQuery(ownerPrototype, fieldName, query);
       }
@@ -2465,7 +2468,10 @@ export function linksToMany<CardT extends LinkableDefConstructor>(
   return {
     setupField(fieldName: string, ownerPrototype: BaseDef) {
       let { computeVia, searchable, query } = options ?? {};
-      let fieldCardThunk = cardThunk(cardOrThunk);
+      let fieldCardThunk = cardThunk(cardOrThunk, {
+        fieldName,
+        ownerPrototype,
+      });
       if (query) {
         validateRelationshipQuery(ownerPrototype, fieldName, query);
       }
@@ -4783,12 +4789,23 @@ function notifySubscribers(
 
 function cardThunk<CardT extends BaseDefConstructor>(
   cardOrThunk: CardT | (() => CardT),
+  // Where the field lives, so the thrown error can name the exact field
+  // instead of leaving the author to bisect their schema. The module the bad
+  // value came from isn't knowable here — by the time the value reaches us
+  // the import has already evaluated to undefined — so the message names the
+  // two ways that happens instead.
+  fieldContext?: { fieldName: string; ownerPrototype: BaseDef },
 ): () => CardT {
   if (!cardOrThunk) {
+    let fieldDescription = fieldContext
+      ? `field '${fieldContext.fieldName}' on '${
+          fieldContext.ownerPrototype.constructor?.name ?? 'unknown card'
+        }'`
+      : 'a field';
     throw new Error(
-      `cardOrThunk was ${cardOrThunk}. There might be a cyclic dependency in one of your fields.
-      Use '() => CardName' format for the fields with the cycle in all related cards.
-      e.g.: '@field friend = linksTo(() => Person)'`,
+      `The card class for ${fieldDescription} was ${cardOrThunk}. Two common causes:
+      (1) the import doesn't match the module's export shape — e.g. \`import X from '…'\` where the module only has a named export; use \`import { X } from '…'\`;
+      (2) a cyclic dependency between cards — use the thunk form in all cards in the cycle, e.g. '@field friend = linksTo(() => Person)'.`,
     );
   }
   return (
--- a/packages/base/markdown-file-def.gts
+++ b/packages/base/markdown-file-def.gts
@@ -558,3 +558,5 @@ export class MarkdownDef extends FileDef {
     return attributes;
   }
 }
+
+export default MarkdownDef;
--- a/packages/host/tests/unit/card-thunk-error-test.ts
+++ b/packages/host/tests/unit/card-thunk-error-test.ts
@@ -1,0 +1,36 @@
+import { getService } from '@universal-ember/test-support';
+import { module, test } from 'qunit';
+
+import { setupRenderingTest } from '../helpers/setup';
+
+// When a field's card class evaluates to undefined — most often a default
+// import of a module that only has named exports, or a genuine cycle — the
+// thrown error names the field and its owning card, so the author lands on
+// the exact declaration instead of bisecting the schema. These call the
+// `field` decorator the same way Babel's decorator transform does.
+module('Unit | card field thunk errors', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('an undefined field card class names the field and its owner', async function (assert) {
+    let loader = getService('loader-service').loader;
+    let api = (await loader.import('@cardstack/base/card-api')) as any;
+    class Broken extends api.CardDef {}
+
+    let cases: [string, string, (value: any) => unknown][] = [
+      ['linksTo', 'src', api.linksTo],
+      ['linksToMany', 'links', api.linksToMany],
+      ['contains', 'meta', api.contains],
+      ['containsMany', 'items', api.containsMany],
+    ];
+    for (let [label, fieldName, fieldFn] of cases) {
+      assert.throws(
+        () =>
+          api.field(Broken.prototype, fieldName, {
+            initializer: () => fieldFn(undefined),
+          }),
+        new RegExp(`field '${fieldName}' on 'Broken'`),
+        `${label} names the field and owner`,
+      );
+    }
+  });
+});
--- a/packages/host/tests/unit/card-thunk-error-test.ts
+++ b/packages/host/tests/unit/card-thunk-error-test.ts
@@ -5,31 +5,67 @@ import { setupRenderingTest } from '../helpers/setup';
 
 // When a field's card class evaluates to undefined — most often a default
 // import of a module that only has named exports, or a genuine cycle — the
-// thrown error names the field and its owning card, so the author lands on
-// the exact declaration instead of bisecting the schema. These call the
-// `field` decorator the same way Babel's decorator transform does.
+// thrown error names the field and its owning card and spells out both
+// causes, so the author lands on the exact declaration instead of bisecting
+// the schema. The eager form throws at decoration time; the thunk form
+// throws at the first `field.card` read. These call the `field` decorator
+// the same way Babel's decorator transform does.
 module('Unit | card field thunk errors', function (hooks) {
   setupRenderingTest(hooks);
 
-  test('an undefined field card class names the field and its owner', async function (assert) {
+  let fieldKinds = (api: any): [string, string, (value: any) => unknown][] => [
+    ['linksTo', 'src', api.linksTo],
+    ['linksToMany', 'links', api.linksToMany],
+    ['contains', 'meta', api.contains],
+    ['containsMany', 'items', api.containsMany],
+  ];
+
+  test('an undefined field card class names the field, the owner, and both causes', async function (assert) {
     let loader = getService('loader-service').loader;
     let api = (await loader.import('@cardstack/base/card-api')) as any;
     class Broken extends api.CardDef {}
 
-    let cases: [string, string, (value: any) => unknown][] = [
-      ['linksTo', 'src', api.linksTo],
-      ['linksToMany', 'links', api.linksToMany],
-      ['contains', 'meta', api.contains],
-      ['containsMany', 'items', api.containsMany],
-    ];
-    for (let [label, fieldName, fieldFn] of cases) {
+    for (let [label, fieldName, fieldFn] of fieldKinds(api)) {
       assert.throws(
         () =>
           api.field(Broken.prototype, fieldName, {
             initializer: () => fieldFn(undefined),
           }),
-        new RegExp(`field '${fieldName}' on 'Broken'`),
-        `${label} names the field and owner`,
+        (err: Error) =>
+          new RegExp(`field '${fieldName}' on 'Broken'`).test(err.message) &&
+          /export shape/.test(err.message) &&
+          /cyclic dependency/.test(err.message),
+        `${label} names the field, the owner, and both causes`,
+      );
+    }
+  });
+
+  test('a thunk resolving to undefined names the field at first read', async function (assert) {
+    let loader = getService('loader-service').loader;
+    let api = (await loader.import('@cardstack/base/card-api')) as any;
+    let { isField } = (await loader.import('@cardstack/runtime-common')) as any;
+    class BrokenThunk extends api.CardDef {}
+
+    for (let [label, fieldName, fieldFn] of fieldKinds(api)) {
+      // The thunk defers evaluation, so decoration itself succeeds. The
+      // decorator returns the property descriptor whose getter carries the
+      // Field object; `field.card` is the accessor every consumer reads
+      // through.
+      let descriptor = api.field(BrokenThunk.prototype, fieldName, {
+        initializer: () => fieldFn(() => undefined),
+      });
+      let field = (descriptor?.get as any)?.[isField];
+      assert.ok(field, `${label} field is registered`);
+      // …and the named error surfaces the first time the class is needed.
+      assert.throws(
+        () => field.card,
+        (err: Error) =>
+          new RegExp(`field '${fieldName}' on 'BrokenThunk'`).test(
+            err.message,
+          ) &&
+          /export shape/.test(err.message) &&
+          /cyclic dependency/.test(err.message),
+        `${label} thunk form names the field and owner at first read`,
       );
     }
   });
--- a/packages/host/tests/unit/filedef-export-shape-test.ts
+++ b/packages/host/tests/unit/filedef-export-shape-test.ts
@@ -1,28 +1,39 @@
 import { getService } from '@universal-ember/test-support';
 import { module, test } from 'qunit';
 
-import { FILEDEF_CODE_REF_BY_EXTENSION } from '@cardstack/runtime-common';
+import {
+  FILEDEF_CODE_REF_BY_EXTENSION,
+  baseRealm,
+} from '@cardstack/runtime-common';
 
 import { setupRenderingTest } from '../helpers/setup';
 
-// Every FileDef subtype module exports its def class both named and as the
-// default, so `import X from '…'` and `import { X } from '…'` are equally
-// valid. A named-only module makes a default import silently evaluate to
-// undefined, which then fails far away at schema time; a consistent shape
-// removes the trap. Walking the extension registry keeps this guard covering
-// every registered subtype, including ones added later.
+// Every FileDef module exports its def class both named and as the default,
+// so `import X from '…'` and `import { X } from '…'` are equally valid. A
+// named-only module makes a default import silently evaluate to undefined,
+// which then fails far away at schema time; a consistent shape removes the
+// trap. Walking the extension registry keeps this guard covering every
+// registered subtype, including ones added later.
+//
+// What this pins is "a default import never yields undefined" — not "a
+// default import yields the class you named". A module carrying a second,
+// named-only class is the sharper trap: `import RasterImageDef from
+// './image-file-def'` evaluates to ImageDef (the default), a perfectly valid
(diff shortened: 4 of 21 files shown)


packages/host/tests/unit/filedef-export-shape-test.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (seen.has(key)) {
+        continue;
+      }
+      seen.add(key);
+      let ns = (await loader.import(ref.module)) as Record<string, unknown>;
+      assert.ok(ns[ref.name], `${ref.module} has named export ${ref.name}`);
+      assert.ok(ns['default'], `${ref.module} has a default export`);
+      // A module registered under exactly one class defaults to that class.
+      // Multi-leaf modules (e.g. gltf-model-def's GltfDef/GlbDef) default to
+      // one of their leaves; existence is all that's required there.
```
lukemelia: [Claude Code 🤖] **Non-blocking nit (test-comment accuracy / evergreen).** This comment names `gltf-model-def`'s `GltfDef`/`GlbDef` as a live example of a multi-leaf module, but no such module (or those class names) exists anywhere in `packages/base` — I grepped the tree and the registry. More to the point, *no* module in `FILEDEF_CODE_REF_BY_EXTENSION` maps to two distinct class names today: all 33 registered modules resolve to exactly one `name`, so `namesInModule.size` is always `1` and the `else` path (skipping the `strictEqual`) is currently dead.

That's not a bug — keeping the `size === 1` guard is reasonable future-proofing for the day a module does register two leaf classes. The only issue is the comment: it presents a fabricated module as if it ships, so a future reader who greps for `gltf-model-def` finds nothing and is left unsure whether the code or the comment is stale.

Suggest describing the mechanism generically instead of citing a nonexistent module, e.g.:

```suggestion
      // A module registered under exactly one class defaults to that class.
      // A module that ever registers two leaf classes would default to one of
      // them; there, existence of *a* default is all that's required.
```

lukemelia (commented):
[Claude Code 🤖] **Lens:** I reviewed this for correctness of the new error path, completeness of the export-shape fix against the whole subtype registry, and whether adding a default export to card modules is safe given how the base realm indexes them.

**Bottom line: this is a clean, well-motivated change with no blocking issues.** The fix is complete, the mechanism is correct, and the tests pin what they claim. One non-blocking nit on a test comment (inline).

**What lands right — and why it's more than cosmetic.**
- *The motivation is real, not hypothetical.* The boxel-cli skill docs already instruct card authors (and the assistant) to reach these leaf classes via a **default** import — `packages/boxel-cli/plugin/skills/boxel-file-def/references/import-paths.md` and `.../boxel/references/quick-reference.md` both show `import PngDef from '…/png-image-def'`, `import MarkdownDef from '…/markdown-file-def'`, etc. Against the named-only modules those imports evaluated to `undefined` and blew up far from the import with a cyclic-dependency red herring. This PR makes the documented usage actually work.
- *The fix is complete against the registry.* I walked every module referenced by `FILEDEF_CODE_REF_BY_EXTENSION` (33 distinct modules) and each now has exactly one `export default`. The 17 touched here plus the shell/other modules (pdf, docx, pptx, xlsx, html, midi, video, 3d, zip, fonts) all export both forms, so the new guard test passes rather than tripping on an untouched module.
- *No double-indexing risk.* Adding `export default X` where `X` is already a named export isn't a new pattern — the shell modules (`pdf-file-def`, `docx-file-def`, `image-file-def`, …) have shipped that exact shape and index cleanly, so the 17 leaf modules adopting it carry no new indexing behavior.
- *The error-path change is correct.* All four field initializers (`contains`/`containsMany`/`linksTo`/`linksToMany`) now thread `{ fieldName, ownerPrototype }` into `cardThunk`; the four remaining `cardThunk()` call sites are stored-thunk invocations, not creations, so nothing is missed. `ownerPrototype.constructor?.name` resolves to the declaring card's name (`Broken` in the test, via `Broken.prototype.constructor.name`). The `card-thunk-error-test.ts` drives the `field` decorator exactly as Babel does and asserts each of the four types names the field and owner — it would fail if any initializer dropped the context.

**Assertion count checks out:** 33 unique modules × (named + default + strictEqual) = 99, matching the test plan.

**One non-blocking recommendation** (see inline thread on `filedef-export-shape-test.ts`): the multi-leaf branch is currently unreachable and its comment cites a module that doesn't exist. Reword the comment; the defensive branch itself is fine as future-proofing.

Non-blocking, and the COMMENT event here is deliberate — approval is the human's call.

packages/base/card-api.gts
```diff
@@ -4783,12 +4789,23 @@ function notifySubscribers(
...
     throw new Error(
-      `cardOrThunk was ${cardOrThunk}. There might be a cyclic dependency in one of your fields.
-      Use '() => CardName' format for the fields with the cycle in all related cards.
-      e.g.: '@field friend = linksTo(() => Person)'`,
+      `The card class for ${fieldDescription} was ${cardOrThunk}. Two common causes:
+      (1) the import doesn't match the module's export shape — e.g. \`import X from '…'\` where the module only has a named export; use \`import { X } from '…'\`;
+      (2) a cyclic dependency between cards — use the thunk form in all cards in the cycle, e.g. '@field friend = linksTo(() => Person)'.`,
     );
   }
   return (
```
habdelra: [Claude Code 🤖] **The new message covers the eager form only — the thunk form still fails un-named, and cause (2) actively routes authors into it.** Non-blocking, but it's a handful of lines and it's what makes the PR's claim ("names the exact declaration") true for both spellings.

**What the guard covers.** The `if (!cardOrThunk)` check fires at decoration time, which is the `linksTo(TsFileDef)` / `contains(MarkdownDef)` shape — the argument itself is `undefined`. That's the case the description describes, and it's now well named.

**What it doesn't.** When the argument *is* a thunk it's returned unwrapped by the line below, and every field kind reads it through an unguarded accessor — `get card() { return this.cardThunk(); }` on `Contains`, `ContainsMany`, `LinksTo` and `LinksToMany` (four separate copies in this file, none of them checked). So `linksTo(() => TsFileDef)` where `TsFileDef` evaluated to `undefined` leaves `field.card === undefined`, and the first consumer to touch it throws with no field name attached. The nearest one on the indexing path is `getFieldDefinitions` in `packages/runtime-common/definitions.ts`:

```ts
let isPrimitive = primitive in field.card;
```

which raises `TypeError: Cannot use 'in' operator to search for 'primitive' in undefined` — the same "fails far from the import, names neither the field nor the module" symptom this change set out to remove.

**Why that matters specifically here.** Cause (2) in the new message tells the author to switch to the thunk form. An author who hit this through cause (1) *and* has a cycle follows that advice, converts to `linksTo(() => X)`, and lands in the un-named `TypeError` instead of the named error — strictly worse guidance than the message implies.

**The fix.** `fieldContext` is already in the closure, so guarding the wrapper covers both forms. Hoist `fieldDescription` out of the `if` and wrap the thunk branch:

```ts
  let fieldDescription = fieldContext
    ? `field '${fieldContext.fieldName}' on '${
        fieldContext.ownerPrototype.constructor?.name || 'unknown card'
      }'`
    : 'a field';
  if (!cardOrThunk) {
    throw new Error(`The card class for ${fieldDescription} was ${cardOrThunk}. …`);
  }
  if ('baseDef' in cardOrThunk) {
    return () => cardOrThunk as CardT;
  }
  let thunk = cardOrThunk as () => CardT;
  return () => {
    let card = thunk();
    if (!card) {
      throw new Error(
        `The card class for ${fieldDescription} resolved to ${card} when the field was first read. …same two causes…`,
      );
    }
    return card;
  };
```

Runtime cost is one truthiness check per `field.card` read. The accessors above memoize nothing — each read already re-invokes the thunk — so this adds no call, just the test.

**Scope:** pre-existing gap, now load-bearing (the new message's advice depends on it). Non-blocking; recommended in this PR.


---

packages/base/card-api.gts
```diff
@@ -4783,12 +4789,23 @@ function notifySubscribers(
...
+  // the import has already evaluated to undefined — so the message names the
+  // two ways that happens instead.
+  fieldContext?: { fieldName: string; ownerPrototype: BaseDef },
 ): () => CardT {
   if (!cardOrThunk) {
+    let fieldDescription = fieldContext
+      ? `field '${fieldContext.fieldName}' on '${
+          fieldContext.ownerPrototype.constructor?.name ?? 'unknown card'
+        }'`
+      : 'a field';
```
habdelra: [Claude Code 🤖] **Two small things about the optional parameter and its fallback.** Both non-blocking.

**1. The `'a field'` fallback is unreachable, so nothing can test it.** `cardThunk` is module-private (no `export`), and it has exactly four call sites — the `setupField` bodies of `containsMany`, `contains`, `linksTo` and `linksToMany` — all of which now pass `{ fieldName, ownerPrototype }`. I grepped the file: those four plus this definition are every mention. So `fieldContext?` can never be undefined at runtime and the `: 'a field'` branch is dead. Making the parameter required (`fieldContext: { fieldName: string; ownerPrototype: BaseDef }`) drops a branch and documents the invariant in the signature — TypeScript then catches a future fifth call site that forgets it, instead of silently degrading the message.

**2. `??` doesn't catch the case it looks like it catches.** `constructor?.name` on a prototype only yields `undefined` if the prototype chain was built without a `constructor`, which doesn't happen for a `class`. What it *does* yield for an anonymous class is the empty string, and `'' ?? x` is `''` — so the message would read ``field 'src' on ''``. `||` covers both the nullish and the empty-name case at no cost. This is prophylactic rather than observed: I grepped for `class extends CardDef` / `class extends FieldDef` (and the `api.`-prefixed forms) across `packages/` and there are no anonymous card classes today.

Both fold naturally into the restructure suggested in the comment on the `return` a few lines below — worth doing them together rather than as separate edits.

**Scope:** simplification / nit. Neither blocks.


---

packages/host/tests/unit/filedef-export-shape-test.ts
```diff
@@ -0,0 +1,53 @@
...
+        continue;
+      }
+      let key = `${ref.module}#${ref.name}`;
+      if (seen.has(key)) {
+        continue;
+      }
+      seen.add(key);
+      let ns = (await loader.import(ref.module)) as Record<string, unknown>;
+      assert.ok(ns[ref.name], `${ref.module} has named export ${ref.name}`);
+      assert.ok(ns['default'], `${ref.module} has a default export`);
```
habdelra, line 46: [Claude Code 🤖] **The registry-driven walk is the right instinct, but the registry stops short of the modules where a default import is most likely to be written — and for two modules in the family the invariant this test states isn't the one that protects the author.** Non-blocking, follow-up.

**Coverage stops at leaf subtypes.** `FILEDEF_CODE_REF_BY_EXTENSION` maps extensions to leaves, so the family's base modules appear in no entry and are never checked: `image-file-def`, `audio-file-def`, `video-file-def`, `font-file-def`, `three-d-model-def`, and `file-api`. Those are exactly the modules a subtype author imports from. All six do have defaults today (`file-api` via `FileDef as default`, the rest via a trailing `export default`), and one of them is already load-bearing in the shipped tree — `packages/base/svg-image-def.gts` line 3 is `import ImageDef from './image-file-def';`, a default import that silently becomes `undefined` the day someone drops that line. Nothing currently pins it. Adding those six module ids as an explicit list beside the registry loop covers them.

**The multi-class case is a sharper trap than "existence is all that's required" suggests.** Two modules in the family export more than one def class, and in both the default is only one of them:

| module | default export | also exported, named-only |
| --- | --- | --- |
| `packages/base/image-file-def.gts` | `ImageDef` | `RasterImageDef` |
| `packages/base/zip-file-def.gts` | `ZipDef` | `ArchiveEntryField` |

`import RasterImageDef from './image-file-def'` does not evaluate to `undefined` — it evaluates to `ImageDef`. That's a perfectly valid `LinkableDefConstructor`, so it type-checks, the new `cardThunk` error can never fire, and the field is silently wired to the wrong class. Worth stating in this comment that what the PR establishes is "a default import never yields `undefined`", not "a default import yields the class you named" — the second only holds for single-class modules, and the two above are where the failure mode changes from a loud `undefined` into a quiet wrong class.

These two are also the concrete examples that belong in the comment three lines below, in place of the `gltf-model-def` module named in the open thread there — they're real, and they make the caveat actionable instead of hypothetical.

**Minor, same walk:** the skip keys on the literal string `'.mismatch'`. Skipping on the entry's shape instead — `if (!ref.module.includes('://')) continue;` — covers any future synthetic entry without a second edit, since every real entry is built by `baseModule()` in full-URL form while that one is a bare relative specifier.

**Scope:** test coverage / follow-up. Nothing here blocks; the guard as written already pins the 33 registered modules (99 assertions, matching the description).


---

packages/host/tests/unit/card-thunk-error-test.ts
```diff
@@ -0,0 +1,36 @@
...
+    ];
+    for (let [label, fieldName, fieldFn] of cases) {
+      assert.throws(
+        () =>
+          api.field(Broken.prototype, fieldName, {
+            initializer: () => fieldFn(undefined),
+          }),
+        new RegExp(`field '${fieldName}' on 'Broken'`),
+        `${label} names the field and owner`,
+      );
```
habdelra, line 69: [Claude Code 🤖] **Confirmation plus one coverage gap.** Non-blocking.

Driving the decorator the way Babel does (`api.field(Broken.prototype, name, { initializer })`) is the right level to test this at — it exercises the real `setupField` → `cardThunk` path rather than a hand-built descriptor, and `Broken.prototype.constructor.name` is what the message interpolates, so the assertion is pinning the actual production expression. I traced all four initializers: each evaluates `cardThunk(...)` as the first thing in its `new Contains/ContainsMany/LinksTo/LinksToMany({...})` argument, so the throw happens before any side effect and reusing one `Broken` class across the four cases can't leak state. The regexes contain no metacharacters, so they match literally.

**What isn't pinned: the guidance itself.** These four assertions cover the naming half of the message. The half that motivates the change for a reader who default-imported a named-only module — cause (1), "the import doesn't match the module's export shape … use `import { X }`" — is asserted nowhere, so a future edit could drop it entirely and stay green. Cheap to close: add a second assertion per case, or match on both halves at once, e.g.

```ts
assert.throws(
  () => api.field(Broken.prototype, fieldName, { initializer: () => fieldFn(undefined) }),
  (err: Error) =>
    new RegExp(`field '${fieldName}' on 'Broken'`).test(err.message) &&
    /export shape/.test(err.message) &&
    /cyclic dependency/.test(err.message),
  `${label} names the field, the owner, and both causes`,
);
```

**And the negative space:** the thunk form. `fieldFn(() => undefined)` through the same four cases is the natural companion test, and it currently would not throw at all — see the comment on `cardThunk`'s `return` in `packages/base/card-api.gts`. If that guard lands, this loop is the place the case belongs; if it doesn't, a comment here saying the thunk form is deliberately out of scope would keep the next reader from assuming this file covers it.

**Scope:** test coverage. Does not block.


---

packages/base/markdown-file-def.gts
```diff
@@ -558,3 +558,5 @@ export class MarkdownDef extends FileDef {
     return attributes;
   }
 }
+
+export default MarkdownDef;
```
habdelra, line 562: [Claude Code 🤖] **Confirmation — this is safe, and the reason is worth writing down because it depends on statement order.** No change requested; this comment is the guard rail for whoever edits these modules next. (It applies identically to the other 16 modules in this diff; I'm anchoring it once here.)

**Why adding a default export can't move the class's code ref.** A second export of the same class is a real hazard in this codebase, because a class's persisted `adoptsFrom` comes from a reverse map keyed on export name:

- `Loader.setIdentities` (`packages/runtime-common/loader.ts`) walks `Object.keys(module)` and registers the **first** key whose value is a function — `if (!this.identities.has(exportedEntity))` — so identity is first-key-wins, and key order decides.
- The AMD transpiler (`packages/runtime-common/amd-transpile/index.ts`) collects export setters into one `exportStatements` array in source-traversal order and appends them all after the module body. `export class MarkdownDef` emits `Object.defineProperty(_exports, "MarkdownDef", { enumerable: true, configurable: true, get … })` (`defineLocalGetter` in `amd-transpile/emit-helpers.ts`), and the trailing `export default MarkdownDef` emits `_exports.default = MarkdownDef;`.
- The named key is therefore defined first and is `enumerable: true`, so `Object.keys(module)` yields `['MarkdownDef', 'default']` and `identifyCard` keeps returning `{ module: …/markdown-file-def, name: 'MarkdownDef' }`.

So `meta.adoptsFrom` on existing instances and the field refs inside indexed definitions are untouched — which is also why the modules that already shipped this exact shape (`pdf-file-def`, `docx-file-def`, `mp4-video-def`, `zip-file-def`, the four font defs, …) index under their named refs today.

**The condition that would break it:** putting the default first — `export default class MarkdownDef {}` with a separate `export { MarkdownDef }` later. Then `_exports.default` is defined first, `identifyCard` starts returning `name: 'default'`, and what gets written into `adoptsFrom` changes. Every module in the family currently orders it the safe way; keeping `export default X;` as the trailing line is load-bearing, not stylistic.

**One knock-on worth knowing about.** `packages/host/app/routes/module.ts` builds one definition per exported name (`for (let [name, maybeBaseDef] of Object.entries(module))`), and that model is what the module prerender page serializes for the indexer to persist (`captureModule` in `packages/realm-server/prerender/utils.ts` reads it out of `[data-prerender-module]`). So each of these 17 modules now contributes a second definition entry keyed `…/default` alongside the named one. It carries the same `fields`/`fieldDefs` payload, and its `definition.codeRef` is still the *named* ref because `makeDefinition` takes it from `identifyCard` rather than from the key. `getTypes` is memoized per class, so there's no extra type-resolution work at index time; the cost is one duplicated JSON blob per module. Consistent with the already-defaulted modules, and arguably the point — the `default` ref now resolves through definition lookup too, so a default-import `linksTo` works end to end and not just in the browser.

**Scope:** confirmation. Nothing to change.


---

packages/runtime-common/file-def-code-ref.ts
```diff
@@ -15,7 +15,7 @@ function baseModule(name: string): RealmResourceIdentifier {
   return `${baseRealm.url}${name}` as RealmResourceIdentifier;
 }
 
-const FILEDEF_CODE_REF_BY_EXTENSION: Record<string, ResolvedCodeRef> = {
+export const FILEDEF_CODE_REF_BY_EXTENSION: Record<string, ResolvedCodeRef> = {
```
habdelra: [Claude Code 🤖] **Nit, non-blocking:** this is now genuinely public — `packages/runtime-common/index.ts` has `export * from './file-def-code-ref.ts'`, which is what lets the new test import it from `@cardstack/runtime-common` — but the type still permits writes, so any consumer can add or repoint an extension mapping at runtime.

Every use in the tree is read-only: the four sites in this file (`isFileDefCodeRef`, the two `in` checks, `resolveFileDefCodeRef`) and the new test's walk. So tightening it costs nothing:

```suggestion
export const FILEDEF_CODE_REF_BY_EXTENSION: Readonly<
  Record<string, ResolvedCodeRef>
> = {
```

Worth doing while the export is new, since it's harder to add once something outside the package is mutating it. (The `TODO: Replace with realm metadata configuration` below suggests this map is destined to become data anyway — a read-only type is the right shape for the interim.)


---

habdelra (commented):
[Claude Code 🤖] I reviewed this through one lens above all: adding a second export of a class is not cosmetic in this codebase, because a class's persisted `adoptsFrom` comes from a reverse map keyed on **export name** and populated first-key-wins. So the first question was whether 17 new default exports can move any class's code ref, and the second was whether the new error actually covers the failure modes it enumerates.

**Bottom line: no blocking issues. The default-export additions are safe, for a reason that depends on statement order and is worth recording; the error-message change is a real improvement that covers one of the two spellings it advises.** Two gaps worth closing — the thunk form still fails un-named, and the new guard test's coverage stops short of the modules a default import is most likely to be written against.

**What lands right**

- The convention choice is the conservative one. `export default X;` *trailing* the named `export class X` means the AMD transpiler defines the named key first and `Loader.setIdentities` registers the named ref, so `identifyCard` output — and therefore `meta.adoptsFrom` on existing instances and the field refs in indexed definitions — is unchanged. Verified by reading `defineLocalGetter` (`enumerable: true`), the `exportStatements` ordering in `amd-transpile/index.ts`, and the `!this.identities.has(...)` guard in `loader.ts`. The 16 modules that already shipped this shape are the empirical confirmation. Detail and the ordering hazard are in the comment on `packages/base/markdown-file-def.gts`.
- Driving the guard test off the extension registry rather than a hardcoded list is right: new subtypes are covered the day they're registered. I confirmed all 33 registered modules resolve to exactly one class name and all 33 have default exports, which is where the 99 assertions come from.
- The error test calls the decorator the way Babel's transform does, so it pins the real `setupField` → `cardThunk` path rather than a hand-built descriptor.

**Recommendations, most to least valuable**

1. Guard the thunk branch too. `linksTo(() => X)` with `X === undefined` still yields `field.card === undefined` and dies later in `getFieldDefinitions` as `TypeError: Cannot use 'in' operator to search for 'primitive' in undefined` — no field name. Cause (2) of the new message tells authors to switch to exactly that form. `fieldContext` is already in the closure. See the comment on `cardThunk`'s `return` in `packages/base/card-api.gts`.
2. Extend the export-shape walk to the family's base modules (`image-file-def`, `audio-file-def`, `video-file-def`, `font-file-def`, `three-d-model-def`, `file-api`) — none is in the registry, so none is checked, and `svg-image-def.gts` already default-imports `image-file-def` today. See the comment on the walk in `filedef-export-shape-test.ts`.
3. Say in that same comment that the guarantee is "a default import never yields `undefined`", not "a default import yields the class you named". `image-file-def` (`ImageDef` default, `RasterImageDef` named-only) and `zip-file-def` (`ZipDef` default, `ArchiveEntryField` named-only) are the two modules where a default import silently produces a valid *wrong* class that the new error can never catch — and they're the real examples that belong where `gltf-model-def` currently sits.
4. Make `fieldContext` required: all four call sites pass it and `cardThunk` isn't exported, so the `'a field'` fallback is unreachable. While there, `constructor?.name ?? 'unknown card'` → `||`, since an anonymous class's `.name` is `''`, not nullish.
5. Assert the export-shape guidance in `card-thunk-error-test.ts`, not just the field naming — cause (1) is the sentence this change exists for and nothing pins it.
6. `Readonly<Record<string, ResolvedCodeRef>>` on the newly-public registry.

**On the open thread**

The thread about the `gltf-model-def` example in `filedef-export-shape-test.ts` is still open and no commit has landed since it was raised; recommendation 3 above is the concrete replacement text it was asking for. Nothing else in the discussion is settled or relitigated here.

**On the "unstable" mergeability state**

Not a code signal, as far as I can tell. The first full run on `296f100f` was 20/20 host shards green plus realm-server green. A later run on the same commit shows shards 1, 4, 11 and 20 failed and 16 and 19 cancelled after ~6 hours — but the failing shards uploaded no test output at all (`No files were found with the provided path: junit/host-testem.log`), i.e. they died before testem produced results, and the merged "Host Test Results" check for the same commit still reports 2 836 tests / 0 failed. That reads as harness/boot failure in the re-run rather than anything in this diff. Worth a re-run to get the status clean before merge rather than a code change.

**Adjacent, out of scope**

- `'.mismatch'` is a test-only entry sitting in the production `FILEDEF_CODE_REF_BY_EXTENSION` (module `./filedef-mismatch`, which doesn't exist), which is why the new test needs a skip at all. Pre-existing; whoever does the `TODO: Replace with realm metadata configuration` work is the natural person to move it into the fixture that needs it.
- `RasterImageDef` living in `image-file-def.gts` as a named-only second class is the structural reason recommendation 3 exists. Splitting it into its own module would make the family's "one def class per module, exported both ways" rule total instead of near-total. Not this PR.


---

habdelra (approved):

packages/host/tests/unit/filedef-export-shape-test.ts
lukemelia: [Claude Code 🤖] Fixed in 741f8c2810. The in-loop comment now describes the mechanism generically (a module registered under exactly one class is pinned to default that class; a module registering two leaf classes would only be required to have *a* default), and the module comment cites the two real multi-class modules — `image-file-def` (`RasterImageDef` named-only) and `zip-file-def` (`ArchiveEntryField` named-only) — in place of the fabricated example.

packages/base/card-api.gts
lukemelia: [Claude Code 🤖] Fixed in 741f8c2810. `cardThunk` wraps the thunk form: a thunk resolving to `undefined` throws the same field-and-owner-named error, with both causes, at the first `field.card` read — so following the message's cause (2) advice lands on an equally specific error rather than the anonymous `TypeError` in `getFieldDefinitions`. The accessors re-invoke the thunk on every read, so the wrapper adds a truthiness check per read, not an extra call. `card-thunk-error-test.ts` covers the thunk form through all four field kinds.

packages/base/card-api.gts
lukemelia: [Claude Code 🤖] Fixed in 741f8c2810. `fieldContext` is required — all four call sites pass it, and TypeScript now catches a future call site that forgets — and the owner-name fallback uses `||`, so an anonymous class's empty-string `.name` reads as `unknown card` instead of an empty quote.

packages/host/tests/unit/filedef-export-shape-test.ts
lukemelia: [Claude Code 🤖] Fixed in 741f8c2810. A second test walks the six family base modules (`image-file-def`, `audio-file-def`, `video-file-def`, `font-file-def`, `three-d-model-def`, `file-api`) with the same three assertions, pinning the default import `svg-image-def` already relies on. The module comment states the guarantee precisely — "a default import never yields `undefined`", not "a default import yields the class you named" — and names `image-file-def`/`RasterImageDef` and `zip-file-def`/`ArchiveEntryField` as the modules where a default import of the named-only class silently yields a valid wrong class. The synthetic-entry skip keys on shape (`!ref.module.includes('://')`) rather than on the literal key.

packages/host/tests/unit/card-thunk-error-test.ts
lukemelia: [Claude Code 🤖] Fixed in 741f8c2810. Each eager-form assertion now matches the field naming and both halves of the guidance (export shape + cyclic dependency), and a second test drives the thunk form through all four field kinds: decoration succeeds, and the first `field.card` read throws the named error. The `Field` object is read off the descriptor the decorator returns (via the `isField` symbol), since `field.card` is the accessor every consumer reads through.

packages/base/markdown-file-def.gts
lukemelia: [Claude Code 🤖] Acknowledged — nothing to change. The trailing `export default X;` ordering is the convention throughout the family, and this thread stands as the written rationale for why it is load-bearing.

packages/runtime-common/file-def-code-ref.ts
lukemelia: [Claude Code 🤖] Fixed in 741f8c2810. The registry is `Readonly<Record<string, ResolvedCodeRef>>`; every consumer is a read.

PR #5804, opened 2026-08-18
Disable pnpm's implicit install-on-run in favor of explicit installs

lukemelia:
## What

1. Sets `verifyDepsBeforeRun: false` in `pnpm-workspace.yaml`. pnpm 11 defaults this setting to `install`, which makes every `pnpm run` / `pnpm exec` check node_modules freshness first and silently run a workspace-wide install when the check fails. With `false`, scripts run against node_modules as-is and installs happen only when explicitly requested — the pnpm 10 behavior this workspace ran on previously.
2. Adds `pnpm-workspace.yaml` to every workflow path filter that already lists `pnpm-lock.yaml` (and to ci-host's index-cache invalidation case). The file carries dependency-resolution settings (catalog, overrides, patchedDependencies, allowBuilds, verifyDepsBeforeRun), but a settings-only change that doesn't touch the lockfile previously ran zero suites.

## Why

The freshness check never passes in CI — not even immediately after the init action's `pnpm install --frozen-lockfile`, and not after one of the implicit installs itself completes — so every pnpm invocation in a CI job pays a redundant 4–16s install. Green host-shard logs show the sequence plainly: the explicit install finishes, then three more implicit installs run during setup, then several *concurrent* ones fire when `run-p` boots the test services.

That concurrency is the fatal case. `start:matrix`, `start:smtp`, `start:host-dist`, and the rest are each `pnpm run` invocations, so each spawns its own workspace-wide install into the same node_modules. They stomp each other's bin links (the `Failed to create bin … ENOENT` warnings visible even in passing shards) and occasionally wedge each other. When `start:matrix`'s install is the one that hangs, `start-matrix.sh` never reaches `pnpm assert-synapse-running`, the Synapse container is never created, and the shard fails with `Failed to reach Synapse at http://localhost:8008/_matrix/client/versions after 60 attempts (~300s)` — the recurring "Synapse startup" flake. Failing-shard logs show that install still running at teardown, killed by SIGTERM five minutes after it started, with `Error response from daemon: No such container: boxel-synapse-ci` in between.

The same mechanism slows local dev: every `pnpm <script>` pays the check, and `pnpm start:all` fans out the same concurrent implicit installs.

## Test plan

- `pnpm config get verify-deps-before-run` reports `false` with this change (previously unset, defaulting to `install`).
- With the path filters updated, CI on this PR exercises the fix directly: host-shard logs should show no `Done in Ns using pnpm` implicit-install output between the init action's install and the test run, and no `Failed to create bin … ENOENT` warnings during service boot.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -257,6 +257,15 @@ catalog:
   yaml: ^2.5.1
   yargs: ^17.5.1
 
+# pnpm 11 defaults this to `install`, making every `pnpm run` first check
+# node_modules freshness and silently run a full install when the check
+# fails. The check never passes in CI, so each of the ~9 services `run-p`
+# boots kicks off its own concurrent workspace-wide install into the same
+# node_modules; when they collide badly enough, start:matrix never reaches
+# `docker run` and the shard dies with "Failed to reach Synapse". Disabling
+# restores pnpm 10 behavior: scripts run against node_modules as-is.
+verifyDepsBeforeRun: false
+
 minimumReleaseAge: 1440
 minimumReleaseAgeExclude:
   - "@glint/ember-tsc"
--- a/.github/workflows/ci-host.yaml
+++ b/.github/workflows/ci-host.yaml
@@ -16,6 +16,7 @@ on:
       - ".github/workflows/ci-host.yaml"
       - "package.json"
       - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
       # CI boot path: see comment on the equivalent block in ci.yaml.
       - ".mise.toml"
       - ".github/actions/**"
@@ -149,7 +150,7 @@ jobs:
           use_cache=true
           while IFS= read -r file; do
             case "$file" in
-              packages/realm-server/*|packages/runtime-common/*|packages/postgres/*|packages/boxel-ui/*|package.json|pnpm-lock.yaml)
+              packages/realm-server/*|packages/runtime-common/*|packages/postgres/*|packages/boxel-ui/*|package.json|pnpm-lock.yaml|pnpm-workspace.yaml)
                 use_cache=false
                 echo "Index-behavior change detected: $file"
                 break
--- a/.github/workflows/ci-software-factory.yaml
+++ b/.github/workflows/ci-software-factory.yaml
@@ -20,6 +20,7 @@ on:
       - "packages/software-factory/**"
       - "package.json"
       - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
       # CI boot path: see comment on the equivalent block in ci.yaml.
       - ".mise.toml"
       - ".github/actions/**"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,6 +54,7 @@ jobs:
               - 'packages/runtime-common/**'
               - 'package.json'
               - 'pnpm-lock.yaml'
+              - 'pnpm-workspace.yaml'
               # CI boot path. Each gated suite runs `./.github/actions/init`
               # (which calls `jdx/mise-action`, sourcing `.mise.toml` for tool
               # versions) and then `mise run test-services:*` (which sources
--- a/.github/workflows/pr-boxel-ui.yml
+++ b/.github/workflows/pr-boxel-ui.yml
@@ -8,6 +8,7 @@ on:
       - ".github/workflows/pr-boxel-ui.yml"
       - "package.json"
       - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
 
 permissions:
   contents: read
--- a/.github/workflows/preview-host.yml
+++ b/.github/workflows/preview-host.yml
@@ -10,6 +10,7 @@ on:
       - ".github/workflows/preview-host.yml"
       - "package.json"
       - "pnpm-lock.yaml"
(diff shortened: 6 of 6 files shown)


richardhjtan (commented):
[Claude Code 🤖] I reviewed this as a configuration change with two independent halves: does the setting actually take effect (a mistyped key in `pnpm-workspace.yaml` fails silently), and is the path-filter change complete (a half-applied filter is worse than none, because it looks handled). I verified both against the checked-out branch rather than reading them.

**Bottom line: both halves are correct and complete, and the reasoning in the description holds up. No blocking issues. One follow-up — the change fixes a symptom whose cause is still unexplained, and the artifact that would explain it is one `cat` away in CI.**

**The setting takes effect.** `pnpm config get verify-deps-before-run` reports `undefined` on `main` and `false` on this branch, under the pinned pnpm 11.0.9. The key is top-level rather than nested under the `catalog:` map it follows — the mistake this file's shape invites.

**The path-filter change is complete.** I enumerated every reference to `pnpm-lock.yaml` under `.github/` and checked each against the diff. Six sites carry it as a change-detection input — the `paths:` filters in `ci-host.yaml`, `ci-software-factory.yaml`, `pr-boxel-ui.yml`, `preview-host.yml`, the `dorny/paths-filter` block in `ci.yaml`, and the index-cache invalidation `case` in `ci-host.yaml` — and all six gained `pnpm-workspace.yaml`. None was missed.

**The seventh reference is correctly left alone**, which is worth recording because it looks like a miss. `.github/actions/init/action.yml` keys the `actions/cache` step on `hashFiles('**/pnpm-lock.yaml')`, and that cache's `path` is `pnpm store path` — the content-addressed package store, not `node_modules`. Settings like `verifyDepsBeforeRun` don't change which tarballs are needed, so adding the file there would only cost cache hits; and the resolution-affecting settings that *do* change them (`catalog`, `overrides`, `patchedDependencies`) cannot change without moving the lockfile, which already moves the key. Leaving it out is right in both directions.

**Also worth noting the index-cache case is the right call, not just a consistent one.** `pnpm-workspace.yaml` carries the catalog, so a version bump there changes what the indexer actually runs — which is the same reason `packages/runtime-common/*` is in that list. A settings-only edit will now invalidate the index cache unnecessarily, but erring that way costs a cache miss rather than a wrong green.

**Follow-up:** capture why the freshness check fails, before the comment in `pnpm-workspace.yaml` becomes the last word on it. Detail and the one-command diagnostic are in the thread on that block.

---

pnpm-workspace.yaml
```diff
@@ -257,6 +257,15 @@ catalog:
...
   yargs: ^17.5.1
 
+# pnpm 11 defaults this to `install`, making every `pnpm run` first check
+# node_modules freshness and silently run a full install when the check
+# fails. The check never passes in CI, so each of the ~9 services `run-p`
+# boots kicks off its own concurrent workspace-wide install into the same
+# node_modules; when they collide badly enough, start:matrix never reaches
+# `docker run` and the shard dies with "Failed to reach Synapse". Disabling
+# restores pnpm 10 behavior: scripts run against node_modules as-is.
+verifyDepsBeforeRun: false
```
richardhjtan, line 267: [Claude Code 🤖] **Confirmation, plus the one question this change leaves open.**

**The setting takes effect** — verified rather than assumed, since a mistyped or wrongly-placed key in `pnpm-workspace.yaml` fails silently:

```
pnpm --version                            → 11.0.9
main:        pnpm config get verify-deps-before-run → undefined
this branch: pnpm config get verify-deps-before-run → false
```

Placement is right too: the block is top-level, not nested under the `catalog:` map it follows, which is exactly the mistake this file's shape invites.

**`false` is the correct value here, and it's worth saying why in the comment.** pnpm also offers `warn`, `error`, and `prompt`, and `warn` is the obvious-looking compromise — keep the staleness signal, skip the install. It isn't one, given the premise of this change: if the check never passes in CI, `warn` emits a warning on every `pnpm` invocation in every job, and locally on every `pnpm <script>`. That's noise with no action attached. One sentence saying the middle values were considered and rejected for that reason would stop the next person from "improving" this to `warn` and then reverting it a week later.

**The open question: why does the check fail?** The description establishes that it does — including immediately after a clean `pnpm install --frozen-lockfile` — and that is enough to justify the change. But nothing here says *why*, and that matters for two reasons: if the divergence is real rather than a pnpm bug, this hides it; and if it is a pnpm bug, it is worth reporting upstream rather than carrying a workspace-local workaround indefinitely.

**One command answers it.** The check's inputs are recorded in `node_modules/.pnpm-workspace-state-v1.json`. On this workspace it holds:

```
lastValidatedTimestamp, projects (per-project manifests, keyed by absolute path),
pnpmfiles, settings, filteredInstall
```

Dumping that file in a CI job immediately after the init action's install, and again before the first `pnpm run`, shows which input moved. Two of the fields are already interesting here without any speculation about the cause:

- **`pnpmfiles`** lists this repo's `.pnpmfile.cjs`. A pnpmfile is a resolution input, so it is one of the things the check compares.
- **`filteredInstall`** is `false`, which matches CI — `.github/actions/init/action.yml` runs a single unfiltered `pnpm install --frozen-lockfile`, and it is the only `pnpm install` in `.github/` — so a filtered/unfiltered mismatch is already ruled out.

**Scope.** Follow-up, not a blocker. The change stands on its measured effect; this is about not losing the thread on the cause, since the comment above will be read as the final word on it.

PR #5819, opened 2026-08-19
feat: Bring the BXL authoring skill into the plugin and pin its claims

habdelra:
The BXL engine gives card authors no guard rails: it tolerates missing values, catches spreadsheet error values rather than throwing, and compiles a readable spreadsheet dialect down to jq, so several of the ways to get an expression wrong land a plausible value in the field instead of an error. The authoring skill for that surface lives in `cardstack/boxel-skills`, which is the single source that reaches both agent harnesses — the boxel-cli plugin via `build:skills`, and the in-app AI assistant via the skills-realm sync. This PR is the half the monorepo owns: bringing the skill into the plugin, and proving that what it teaches is what the engine does.

## The skills pin

`BOXEL_SKILLS_VERSION` moves to `v0.1.1`, the first tag carrying `bxl-authoring`, and `pnpm build:skills` regenerates the plugin copies from it (18 skills). That is what puts the skill in front of authors, and what gives the drift guard below a file to read — without it the suite refuses to run rather than passing vacuously against a missing file. The regenerated copies also carry the other upstream skill edits made since `v0.1.0`: linked catalog entries, SEARCH/REPLACE as the only file-editing path, two cardinal rules for instance-authoring traps, and corrected `@cardstack/base` import references.

## The drift guard

`packages/bxl/tests/boxel/authoring-skill-claims.ts` — 32 cases against the copy `build:skills` brings in. Each asserts both halves of a claim: the snippet still appears in the skill, and the behavior it describes still holds against the engine. Whitespace is normalized before matching, so rewrapping prose or realigning a table is not a failure — only changing what an example says is.

Substring matching alone is too weak for that, so the pinning is shaped against the ways a skill actually drifts:

- Multi-line snippets tie each code line to the comment above it, so swapping a WRONG and a RIGHT example fails.
- The refused and allowed lists are read as their backticked tokens and compared as sets against the names the cases cover — an entry added without a case, dropped, or moved between the lists fails. Exact tokens, so a short name like `env` can't be satisfied by a longer word containing it. The refusal set is read from the engine's own ban tables rather than a hand-copied list, so a name the engine starts or stops refusing shows up here.
- Table rows are pinned including their reason column, because a stated mechanism is a claim too.
- The cited-path case names the paths it requires rather than counting them, and resolves the bare `*.md` filenames as well.
- Every allowed form is evaluated for its value rather than merely constructed — construction proves nothing, since the derive profile only screens the names it bans and an unknown name constructs happily, which an unknown-name control asserts. It loads the lazy formula families first, as the host does before serving the module, or `NPV` and `isEmail` would be checked against a narrower library than a card gets.

Two things the guard cannot reach, stated in its header rather than implied away: prose, and anything needing a live card runtime. Query-backed inverse staleness and the `{ id }` clip across a cycle are pinned by the host integration suites the skill names, and the path case checks those pointers resolve; the stale-paint interaction with Glimmer's render flush is not pinned anywhere — it follows from runloop ordering, and what is pinned is the `memoize: false` mitigation the skill gives.

The `bxl` CI change filter covers the skill's directory and the host paths the guard cites, so editing either runs this suite on the commit that changes it rather than on the next one.

## Engine-doc corrections

Two `packages/bxl/docs/syntax-modes.md` examples handed an iterating path straight to `SUM` — `SUM(Patients[].Billing.RoomCharge)` calls `SUM` once per patient and yields one output per patient rather than a total — and one used a quoted multi-word label, which throws without a schema. Both corrected, with the collect rule and its two boundaries (comma lists collect themselves; scalar parameters must not be wrapped, or `ROUND([1.234], 2)` blanks the field) added to the cross-cutting behaviors list. The doc's import line now names the platform module `@cardstack/bxl`, keeping the relative-bundle form for realms that carry their own.

## Test plan

- `pnpm test` in `packages/bxl`: 69/69 suites against the copy `build:skills` writes, the guard among them at 32/32.
- `pnpm exec vitest run tests/scripts/build-skills.test.ts` in `packages/boxel-cli`: 15/15.
- Guard verified by mutation: moving `reduce` between the lists, swapping the WRONG/RIGHT aggregate examples, renaming a cited doc, falsifying a reason cell in either table, adding an unpinned entry to a list, and listing an error value the engine never raises each fail their case with a pointer to the line to update.


packages/bxl/docs/syntax-modes.md
```diff
@@ -16,21 +16,26 @@ locked down by a case in [`../tests/boxel/`](../tests/boxel/).
...
 
-In a Boxel realm, import from the uploaded realm bundle using the
-relative path to that bundle:
+In a Boxel realm, import the platform module by its bare specifier —
+the host serves it to card code, and the package root is the
+card-facing entry:
 
 ```ts
-import { expression, fx, jq } from '../bxl';
+import { expression, fx, jq } from '@cardstack/bxl';
```
burieberry, line 24: [Claude Code 🤖] This fix (bare `@cardstack/bxl` import, collected aggregate arguments) also applies to the "Using BXL inside Boxel" section of `packages/bxl/README.md`, which this PR leaves untouched. That README still teaches `import { expression, fx, jq } from '../bxl'` as the only import form and the `computeVia: expression(fx`SUM("Line Item".Amount)`)` example — the exact form this PR's new "a quoted label with no schema fails loudly" test pins as throwing `Cannot index string with string` on first read. An author following the package's front-page README hits an indexing error on the instance. Suggest applying the same edits to README.md's section.

.github/workflows/ci.yaml
```diff
@@ -137,6 +137,10 @@ jobs:
               # shared CI-boot paths trigger it.
               - *shared
               - 'packages/bxl/**'
+              # The card-authoring skill's claims are pinned by a BXL suite
+              # that reads the skill file, so an edit to the skill has to run
+              # that suite too.
+              - 'packages/boxel-cli/plugin/skills/bxl-authoring/**'
```
burieberry, line 144: [Claude Code 🤖] The new drift-guard suite asserts that four `packages/host` paths cited by the skill exist (`bxl-tracking.ts` and three `bxl-*-test.gts` integration tests), but this filter only gains the skill directory — no host path. A host-only PR that renames one of those files won't match the `bxl` filter, so `bxl-test` is skipped and the PR merges green; then on the push to main, `bxl-test` runs unconditionally (`|| github.ref == 'refs/heads/main'`) and the `existsSync` assertion fails — red main from a green PR, bisecting to a commit whose CI never ran the failing suite. Suggest either adding the four cited host paths to this filter, or dropping the cross-package existence assertions from the suite.

packages/bxl/tests/boxel/authoring-skill-claims.ts
```diff
@@ -0,0 +1,903 @@
...
+// The host folds every lazy formula family into the default library set before
+// serving `@cardstack/bxl` to card code, so a card reaches `NPV` and `isEmail`
+// as readily as `ROUND`. Do the same here or those cases would assert against a
+// narrower library than the skill's audience has.
+await loadAllFormulaExtensions();
+
+const REPO_ROOT = join(import.meta.dirname, '..', '..', '..', '..');
+const SKILL_PATH = join(
+  REPO_ROOT,
+  'packages/boxel-cli/plugin/skills/bxl-authoring/SKILL.md',
```
burieberry, line 54: [Claude Code 🤖] The hand-authored skill this suite reads lives in `packages/boxel-cli/plugin/skills/`, a directory `scripts/build-skills.ts` repopulates from the pinned `cardstack/boxel-skills` tag — `copyEntries()` does `rmSync`+`cpSync` for every upstream entry by name with no protection for entries it never wrote, and the `boxel-cli-publish` workflow auto-commits the regenerated directory to main with `[skip ci]`. If a future `BOXEL_SKILLS_VERSION` bump pulls an upstream tag that has gained a `bxl-authoring` skill, the monorepo-authored SKILL.md gets silently replaced on a commit no CI runs on, and this suite then fails on the next bxl-touching PR, blaming an author who touched nothing related. Suggest a reserved-names guard in `copyEntries` (refuse to overwrite entries absent from the manifest), or homing the skill under `packages/bxl/` and copying it into the plugin at build time.

burieberry (commented):
[Claude Code 🤖] Automated review — 3 findings, each verified against the branch. Details inline.

packages/boxel-cli/plugin/skills/bxl-authoring/SKILL.md
```diff
@@ -0,0 +1,341 @@
...
+  ),
+});
+```
+
+Without `as`, a structured value reaches the serializer as an anonymous object
+and fails to identify.
+
+## Reviewing a card's BXL
+
+1. Every aggregate's argument is wrapped in `[…]`.
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Correct, and thanks — the checklist item over-reached into a real failure. Fixed in `763d7795`; the skill itself has since moved to cardstack/boxel-skills#116, where the current text lives.

Probing the compiler showed the rule is narrower than the body stated too, not just the checklist:

```
SUM(Claims[].Paid)      -> SUM(.claims[].paid)        [10, 5]   <- the trap
SUM([Claims[].Paid])    -> SUM([.claims[].paid])      15
SUM(Amounts)            -> SUM(.amounts)              15        <- array field, no wrap
SUM(Paid, Reserve)      -> SUM([.paid, .reserve])     5         <- compiler collects
ROUND([1.234], 2)       -> ROUND([1.234]; 2)          null      <- wrapped scalar
NPV([0.1], CashFlows)   -> NPV([0.1]; .cashFlows)     null
```

A variadic aggregate gets its comma list collected into a jq array, while a fixed-arity function gets `;`-separated arguments — so the trap is specific to a single-argument aggregate over an iterating path, and wrapping a scalar parameter is its own bug. The body now says "collect the argument that iterates, and only that one" with those four boundaries, and the checklist matches it. The drift-guard suite pins each of the six rows above.

packages/bxl/tests/boxel/authoring-skill-claims.ts
```diff
@@ -0,0 +1,565 @@
...
+check('a schema makes implicit iteration collect on its own', () => {
+  shows('`SUM("Line Item"."Line Total")` compiles to');
+  shows('`SUM([.lineItems[].lineTotal])`');
+  const schema = {
+    fields: [
+      {
+        key: 'lineItems',
+        label: 'Line Item',
+        kind: 'array' as const,
+        item: { fields: [{ key: 'lineTotal', label: 'Line Total' }] },
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Right, and the gap was wider than the four names — fixed in `53fc9beb`, then closed properly in `c83b657`.

`53fc9beb` compares both lists as *sets* against the names the cases cover, so an entry added without a case, dropped, or moved between the lists fails. That still left a closed loop, though: the comparison was skill-vs-suite, so a ban neither enumerated stayed invisible to both. `c83b657` anchors it to the engine — it walks `BXL_DERIVE_DENIED_CALLS` and requires every denied call to be either named on the page or waived in the suite with the category the page describes it by, each waiver checked against `categoryForBxlFunction` so it cannot outlive its reason.

That immediately paid for itself: the profile's volatile set is six calls, not four. `BXL_VOLATILE_CALLS` folds in `VOLATILE_VALIDATION_FUNCTIONS`, so `isAfter` and `isBefore` are refused — while the page blessed "validator helpers (`isEmail`)" two lines below and its dates section never mentioned them. `isBefore(StartDate, EndDate)` threw at module load with nothing to warn about it. Both are now named, with the deterministic alternative.

`@Env`, `$old`, `stderr`, `halt` and bare `=` gained cases too. Verified the new check bites by injecting a ban into the engine table:

```
the derive profile bans these and the skill neither names nor waives them: SIMULATE_NEW_BAN
```

packages/bxl/tests/boxel/authoring-skill-claims.ts
```diff
@@ -0,0 +1,565 @@
...
+  strictEqual(
+    expression(fx`IFERROR(VLOOKUP(Sku, Rows, 2, FALSE), "unlisted")`).call({
+      sku: 'x',
+      rows: [],
+    }),
+    'unlisted',
+  );
+  shows('fx`IFNA(NA(), "none")`');
+  strictEqual(expression(fx`IFNA(NA(), "none")`).call(noAmounts), 'none');
+  shows('An `AVERAGE` over an empty collection raises');
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in `53fc9beb`. Each case evaluates once now, inside a try that reports the throw with the case's name, then asserts the value — so a throw and a wrong value are told apart by where the failure lands rather than by running the source twice.

packages/bxl/tests/boxel/authoring-skill-claims.ts
```diff
@@ -0,0 +1,565 @@
...
+  strictEqual(expression(fx`PaidAmount`).bxl.compiledSource, '.paidAmount');
+});
+
+check('IF is the Excel function, if/then/end is the jq construct', () => {
+  shows(
+    '`IF(cond, t, f)` is the Excel function, `if cond then … end` is the jq',
+  );
+  strictEqual(
+    expression(fx`IF(.status == "Open", 1, 0)`).call({ status: 'Open' }),
+    1,
```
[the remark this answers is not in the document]
habdelra: [Claude Code 🤖] Fixed in `53fc9beb`. Both lists are now read as their backticked spans and compared as exact tokens, so a short name like `env` cannot be satisfied by a longer word containing it or by appearing inside a neighbouring code span.

I checked whether a collision existed before changing it — none did today, so the concern was latent rather than live. Worth doing anyway: every name in both lists is already an exact backticked span, so exact matching was free, and it let the check become a set comparison that also catches an entry added to the page without a case.

`c83b657` then anchored the refused side to `BXL_DERIVE_DENIED_CALLS` rather than to this suite's own list, which is what caught two genuinely missing bans (`isAfter` / `isBefore`).

packages/bxl/docs/syntax-modes.md
habdelra: [Claude Code 🤖] Confirmed and fixed in `48d3f73`. You were right down to the failure mode — construction passes the profile check and the throw waits for the first read:

```
SUM("Line Item".Amount)  → constructs OK, compiles to SUM("Line Item".amount)
                         → first read THROWS: Cannot index string with string
```

With no schema the quoted label stays a string literal, so jq indexes the string `"Line Item"` with `"amount"`. The README section now names the platform module and collects the iterating argument (`SUM([LineItems[].Amount])` → 15), with a comment saying why.

I checked the rest of that README while I was in there: the other `SUM("Line Item"."Line Total")` occurrences all pass a schema, where implicit iteration collects, so they are correct as written.

.github/workflows/ci.yaml
habdelra: [Claude Code 🤖] Good catch — red main from a green PR is exactly the shape, and it's fixed in `48d3f73`. The filter gains the two host globs:

```yaml
- 'packages/host/tests/integration/bxl-*'
- 'packages/host/tests/helpers/cards/bxl-*'
```

Took the "add the cited paths" branch rather than dropping the assertions, since keeping the skill's pointers honest is most of what that case buys. Verified the globs with picomatch rather than eyeballing them — all four cited paths match, and an unrelated host test (`realm-test.gts`) does not, so `bxl-test` doesn't start firing on every host change.

Residual gap worth naming: a rename *out* of the `bxl-` prefix still slips past the filter and lands on main. Narrowing that further means widening to `packages/host/tests/**`, which trades a rare miss for running the suite on every host test edit — cheap, since it's plain-node and self-contained, but I left it narrow. Happy to widen if you'd rather close it completely.

packages/bxl/tests/boxel/authoring-skill-claims.ts
habdelra: [Claude Code 🤖] The specific exposure is gone as of `f485ee8a` — the skill moved to cardstack/boxel-skills#116, so it is manifest-owned upstream rather than hand-authored here, and this suite reads the copy `build:skills` brings in. It moved for a bigger reason than this thread: the monorepo plugin tree only reaches boxel-cli agents, and the in-app assistant loads skills from the skills realm, so a card-authoring skill homed here would have missed most of its audience.

The hazard you describe is real and outlives this PR, though. Verified both halves:

- `copyEntries` does `rmSync(dest, { recursive: true, force: true })` + `cpSync` per upstream entry name, with no check against the prior manifest. The stale-sweep is safe for hand-authored entries — its header comment says so — but the *copy* is not.
- `boxel-cli-publish` stages `packages/boxel-cli/plugin/skills` and commits with `[skip ci]`.

So if an upstream tag ever ships a skill whose directory name collides with one of the nine monorepo-authored ones (`boxel-api`, `boxel-command`, `indexing-errors`, `file-ops`, `profile`, `realm-history`, `realm-sync`, `search`, `boxel-file-structure`), that skill is silently replaced on a commit no CI runs against.

Left out of this PR deliberately: the fix belongs in `scripts/build-skills.ts` (a reserved-names guard refusing to overwrite an entry absent from the prior manifest), which would pull `packages/boxel-cli/**` back into a diff that no longer touches it and re-trigger the conventional-commit title requirement. Happy to file it as its own issue and fix it there.

lukemelia (approved):

PR #5831, opened 2026-08-20
feat: create a personal realm for accounts minted during profile add

lukemelia:
A Google sign-in during `boxel profile add` whose verified email matches no existing account gets a brand-new Matrix account, minted by Synapse's OIDC mapping provider with no Boxel page in the loop — Synapse redirects straight from the IdP back to the CLI's loopback listener. The web signup bootstrap that gives every new account its personal realm therefore never runs for these accounts, and they come out of `profile add` with no realms at all.

## What this does

- Adds `ensurePersonalRealm` (`packages/boxel-cli/src/lib/personal-realm.ts`): if an account's realm list in Matrix account data is empty, it creates the `personal` realm on the profile's realm server — named `<display name>'s Workspace` from the account's Matrix display name (which Synapse sets from the Google profile), with the same icon/background defaults the web signup uses — and records the realm in the account's realm list.
- The browser path of `profile add` runs this bootstrap right after saving the profile. Accounts that already have realms are left alone, so it's a no-op for existing accounts and for accounts that just registered through the cli-auth page (where the web bootstrap runs before the hand-off).
- If the realm exists on the server but is missing from the account's realm list, it is re-linked rather than failing on the endpoint collision.
- A bootstrap failure warns and points at the web app instead of failing the profile add — the profile itself is usable without it.

The terminal password path and the non-interactive `-u`/`-p` path are unchanged: they can only sign in to accounts that already exist, so no account minting happens there.

## Tests

- Unit tests for `ensurePersonalRealm` covering: accounts with realms untouched, creation named from the display name, localpart fallback, re-linking an existing realm, and error propagation without touching the realm list.
- A matrix e2e test driving a brand-new Google identity through the real cli-auth flow (mock OIDC → Synapse → CLI loopback) and asserting the account ends up with its personal realm in account data, and that a second run is a no-op.
- `pnpm test:unit-exclude-smoke` in `packages/boxel-cli`: 496 passed.

### DIFF
--- a/packages/boxel-cli/src/lib/personal-realm.ts
+++ b/packages/boxel-cli/src/lib/personal-realm.ts
@@ -7,7 +7,7 @@ import { ensureTrailingSlash } from '@cardstack/runtime-common/paths';
 import {
   addRealmToMatrixAccountData,
   getRealmServerToken,
-  getUserRealmsFromMatrixAccountData,
+  requireUserRealmsFromMatrixAccountData,
   type MatrixAuth,
 } from './auth.ts';
 
@@ -35,12 +35,16 @@ export type EnsurePersonalRealmResult =
 //
 // "Has none" is judged from the account's realm list in Matrix account data —
 // the same list the web app assembles a session from — so accounts that
-// already have realms, however they got them, are left alone.
+// already have realms, however they got them, are left alone. The read is
+// strict (confirmed answer or throw): this gate triggers an automatic
+// mutation, so a failed read must surface as "couldn't determine" — which the
+// caller turns into a warning — rather than read as "has no realms" and mint
+// a realm for an account that already has them.
 export async function ensurePersonalRealm(
   auth: MatrixAuth,
   realmServerUrl: string,
 ): Promise<EnsurePersonalRealmResult> {
-  let realms = await getUserRealmsFromMatrixAccountData(auth);
+  let realms = await requireUserRealmsFromMatrixAccountData(auth);
   if (realms.length > 0) {
     return { outcome: 'has-realms' };
   }
--- a/packages/boxel-cli/src/lib/personal-realm.ts
+++ b/packages/boxel-cli/src/lib/personal-realm.ts
@@ -1,0 +1,140 @@
+import {
+  iconURLFor,
+  getRandomBackgroundURL,
+} from '@cardstack/runtime-common/realm-display-defaults';
+import { ensureTrailingSlash } from '@cardstack/runtime-common/paths';
+
+import {
+  addRealmToMatrixAccountData,
+  getRealmServerToken,
+  getUserRealmsFromMatrixAccountData,
+  type MatrixAuth,
+} from './auth.ts';
+
+// The web app's signup flow ends by creating this realm for the new account
+// (see MatrixService.initializeNewUser in the host); these values mirror it so
+// an account bootstrapped from the CLI is indistinguishable from one
+// bootstrapped from the web.
+const PERSONAL_REALM_ENDPOINT = 'personal';
+
+export type EnsurePersonalRealmResult =
+  // The account already has at least one realm; nothing was done.
+  | { outcome: 'has-realms' }
+  // A personal realm was created and recorded in the account's realm list.
+  | { outcome: 'created'; realmUrl: string }
+  // The personal realm already existed on the server but was missing from the
+  // account's realm list; it was re-linked rather than created.
+  | { outcome: 'linked'; realmUrl: string };
+
+// Give an account its personal realm if it has none — the account-level
+// bootstrap the web app runs at signup, for accounts that never went through
+// that flow. Synapse mints such accounts itself when a Google sign-in's
+// verified email matches nothing (see boxel_oidc_mapping_provider.py), so a
+// brand-new user can finish `boxel profile add` without any Boxel code having
+// run in their browser.
+//
+// "Has none" is judged from the account's realm list in Matrix account data —
+// the same list the web app assembles a session from — so accounts that
+// already have realms, however they got them, are left alone.
+export async function ensurePersonalRealm(
+  auth: MatrixAuth,
+  realmServerUrl: string,
+): Promise<EnsurePersonalRealmResult> {
+  let realms = await getUserRealmsFromMatrixAccountData(auth);
+  if (realms.length > 0) {
+    return { outcome: 'has-realms' };
+  }
+
+  let displayName =
+    (await fetchMatrixDisplayName(auth)) ?? localpart(auth.userId);
+  let name = `${displayName}'s Workspace`;
+
+  let serverToken = await getRealmServerToken(auth, realmServerUrl);
+  let response = await fetch(
+    `${realmServerUrl.replace(/\/$/, '')}/_create-realm`,
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/vnd.api+json',
+        Authorization: serverToken,
+      },
+      body: JSON.stringify({
+        data: {
+          type: 'realm',
+          attributes: {
+            endpoint: PERSONAL_REALM_ENDPOINT,
+            name,
+            iconURL: iconURLFor(displayName),
+            backgroundURL: getRandomBackgroundURL(),
+          },
+        },
+      }),
+    },
+  );
+
+  if (!response.ok) {
+    let errorBody = await response.text();
+    // The realm exists but the account's realm list doesn't mention it (the
+    // list read above was empty). Re-link it instead of failing: the point of
+    // this bootstrap is an account whose realm list names a realm it can use.
+    if (errorBody.includes('already exists')) {
+      let realmUrl = errorBody.match(/'(https?:\/\/[^']+)'/)?.[1];
+      if (realmUrl) {
+        realmUrl = ensureTrailingSlash(realmUrl);
+        await addRealmToMatrixAccountData(auth, realmUrl);
+        return { outcome: 'linked', realmUrl };
+      }
+    }
+    throw new Error(
+      `Could not create the personal realm: realm server returned ${response.status}: ${errorBody}`,
+    );
+  }
+
+  let result = (await response.json()) as { data?: { id?: unknown } };
+  let rawRealmUrl = result?.data?.id;
+  if (typeof rawRealmUrl !== 'string' || rawRealmUrl.trim() === '') {
+    throw new Error(
+      'Could not create the personal realm: realm server response did not include a realm URL (data.id)',
+    );
+  }
+  let realmUrl = ensureTrailingSlash(rawRealmUrl);
+  await addRealmToMatrixAccountData(auth, realmUrl);
+  return { outcome: 'created', realmUrl };
+}
+
+function localpart(userId: string): string {
+  return userId.replace(/^@/, '').split(':')[0];
+}
+
+// The account's Matrix display name, which for a Google-provisioned account
+// Synapse set from the Google profile's name — the closest thing such an
+// account has to the display name the web signup form would have collected.
+async function fetchMatrixDisplayName(
+  auth: MatrixAuth,
+): Promise<string | undefined> {
+  let response: Response;
+  try {
+    response = await fetch(
+      new URL(
+        `_matrix/client/v3/profile/${encodeURIComponent(auth.userId)}/displayname`,
+        auth.matrixUrl,
+      ).href,
+      { headers: { Authorization: `Bearer ${auth.accessToken}` } },
+    );
+  } catch {
+    return undefined;
+  }
+  if (!response.ok) {
+    return undefined;
+  }
+  try {
+    let { displayname } = (await response.json()) as {
+      displayname?: unknown;
+    };
+    return typeof displayname === 'string' && displayname.trim()
+      ? displayname.trim()
+      : undefined;
+  } catch {
+    return undefined;
+  }
+}
--- a/packages/boxel-cli/tests/lib/personal-realm.test.ts
+++ b/packages/boxel-cli/tests/lib/personal-realm.test.ts
@@ -1,0 +1,190 @@
+import { describe, it, expect, afterEach, vi } from 'vitest';
+
+import { ensurePersonalRealm } from '../../src/lib/personal-realm.ts';
+import type { MatrixAuth } from '../../src/lib/auth.ts';
+
+const MATRIX_URL = 'https://matrix.example.com';
+const REALM_SERVER_URL = 'https://realms.example.com/';
+const USER_ID = '@newuser:example.com';
+
+const AUTH: MatrixAuth = {
+  accessToken: 'matrix-token',
+  deviceId: 'DEVICE',
+  userId: USER_ID,
+  matrixUrl: MATRIX_URL,
+};
+
+const ACCOUNT_DATA_PATH = `/_matrix/client/v3/user/${encodeURIComponent(
+  USER_ID,
+)}/account_data/app.boxel.realms`;
+const DISPLAYNAME_PATH = `/_matrix/client/v3/profile/${encodeURIComponent(
+  USER_ID,
+)}/displayname`;
+const OPENID_PATH = `/_matrix/client/v3/user/${encodeURIComponent(
+  USER_ID,
+)}/openid/request_token`;
+
+function jsonResponse(body: unknown, init?: ResponseInit): Response {
+  return new Response(JSON.stringify(body), {
+    status: 200,
+    headers: { 'Content-Type': 'application/json' },
+    ...init,
+  });
+}
+
+interface StubOptions {
+  // What the initial realm-list read returns; a 404 is "never set".
+  realmsResponse?: Response;
+  // What GET .../displayname returns.
+  displaynameResponse?: Response;
+  // What POST /_create-realm returns.
+  createRealmResponse?: Response;
+}
+
+// Routes every request ensurePersonalRealm can make, recording each so tests
+// can assert on what was (and wasn't) called and with which bodies.
+function stubFetchRoutes(options: StubOptions = {}) {
+  const calls: { method: string; url: string; body?: string }[] = [];
+  const fetchStub = vi.fn(async (input: any, init?: any) => {
+    const url = new URL(typeof input === 'string' ? input : input.url);
+    const method = init?.method ?? 'GET';
+    const call = { method, url: url.href, body: init?.body as string };
+    calls.push(call);
+
+    if (method === 'GET' && url.pathname === ACCOUNT_DATA_PATH) {
+      return (
+        options.realmsResponse ??
+        jsonResponse({ errcode: 'M_NOT_FOUND' }, { status: 404 })
+      );
+    }
+    if (method === 'PUT' && url.pathname === ACCOUNT_DATA_PATH) {
+      return jsonResponse({});
+    }
+    if (method === 'GET' && url.pathname === DISPLAYNAME_PATH) {
+      return (
+        options.displaynameResponse ?? jsonResponse({ displayname: 'New User' })
+      );
+    }
+    if (method === 'POST' && url.pathname === OPENID_PATH) {
+      return jsonResponse({ access_token: 'openid-token' });
+    }
+    if (method === 'POST' && url.pathname === '/_server-session') {
+      return jsonResponse(
+        {},
+        { headers: { Authorization: 'Bearer server-jwt' } },
+      );
+    }
+    if (method === 'POST' && url.pathname === '/_create-realm') {
+      return (
+        options.createRealmResponse ??
+        jsonResponse(
+          { data: { id: 'https://realms.example.com/newuser/personal' } },
+          { status: 202 },
+        )
+      );
+    }
+    throw new Error(`unexpected fetch: ${method} ${url.href}`);
+  });
+  vi.stubGlobal('fetch', fetchStub);
+  return { calls };
+}
+
+function callsTo(
+  calls: { method: string; url: string; body?: string }[],
+  method: string,
+  pathname: string,
+) {
+  return calls.filter(
+    (c) => c.method === method && new URL(c.url).pathname === pathname,
+  );
+}
+
+describe('ensurePersonalRealm', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it('leaves an account that already has realms alone', async () => {
+    const { calls } = stubFetchRoutes({
+      realmsResponse: jsonResponse({
+        realms: ['https://realms.example.com/newuser/existing/'],
+      }),
+    });
+
+    const result = await ensurePersonalRealm(AUTH, REALM_SERVER_URL);
+
+    expect(result).toEqual({ outcome: 'has-realms' });
+    // Nothing but the realm-list read: no realm-server traffic at all.
+    expect(calls).toHaveLength(1);
+  });
+
+  it('creates a personal realm named after the Matrix display name and records it in the realm list', async () => {
+    const { calls } = stubFetchRoutes();
+
+    const result = await ensurePersonalRealm(AUTH, REALM_SERVER_URL);
+
+    expect(result).toEqual({
+      outcome: 'created',
+      realmUrl: 'https://realms.example.com/newuser/personal/',
+    });
+
+    const [create] = callsTo(calls, 'POST', '/_create-realm');
+    const attributes = JSON.parse(create.body!).data.attributes;
+    expect(attributes.endpoint).toBe('personal');
+    expect(attributes.name).toBe("New User's Workspace");
+    expect(attributes.iconURL).toBeTruthy();
+    expect(attributes.backgroundURL).toBeTruthy();
+
+    const [put] = callsTo(calls, 'PUT', ACCOUNT_DATA_PATH);
+    expect(JSON.parse(put.body!)).toEqual({
+      realms: ['https://realms.example.com/newuser/personal/'],
+    });
+  });
+
+  it('falls back to the Matrix ID localpart when the account has no display name', async () => {
+    const { calls } = stubFetchRoutes({
+      displaynameResponse: jsonResponse(
+        { errcode: 'M_NOT_FOUND' },
+        { status: 404 },
+      ),
+    });
+
+    await ensurePersonalRealm(AUTH, REALM_SERVER_URL);
+
+    const [create] = callsTo(calls, 'POST', '/_create-realm');
+    expect(JSON.parse(create.body!).data.attributes.name).toBe(
+      "newuser's Workspace",
+    );
+  });
+
+  it('re-links a personal realm that exists on the server but is missing from the realm list', async () => {
+    const { calls } = stubFetchRoutes({
+      createRealmResponse: new Response(
+        `realm 'https://realms.example.com/newuser/personal/' already exists on this server`,
+        { status: 400 },
+      ),
+    });
+
+    const result = await ensurePersonalRealm(AUTH, REALM_SERVER_URL);
+
+    expect(result).toEqual({
+      outcome: 'linked',
+      realmUrl: 'https://realms.example.com/newuser/personal/',
+    });
+    const [put] = callsTo(calls, 'PUT', ACCOUNT_DATA_PATH);
+    expect(JSON.parse(put.body!)).toEqual({
+      realms: ['https://realms.example.com/newuser/personal/'],
+    });
+  });
+
+  it('throws on any other realm-server failure without touching the realm list', async () => {
+    const { calls } = stubFetchRoutes({
+      createRealmResponse: new Response('boom', { status: 500 }),
+    });
+
+    await expect(ensurePersonalRealm(AUTH, REALM_SERVER_URL)).rejects.toThrow(
+      /returned 500/,
+    );
+    expect(callsTo(calls, 'PUT', ACCOUNT_DATA_PATH)).toHaveLength(0);
+  });
+});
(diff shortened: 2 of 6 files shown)


packages/boxel-cli/src/lib/personal-realm.ts
```diff
@@ -0,0 +1,140 @@
...
+// the same list the web app assembles a session from — so accounts that
+// already have realms, however they got them, are left alone.
+export async function ensurePersonalRealm(
+  auth: MatrixAuth,
+  realmServerUrl: string,
+): Promise<EnsurePersonalRealmResult> {
+  let realms = await getUserRealmsFromMatrixAccountData(auth);
+  if (realms.length > 0) {
+    return { outcome: 'has-realms' };
+  }
```
lukemelia: [Claude Code 🤖] **The zero-realms gate treats a failed realm-list read as "the account has no realms", so a degraded Matrix account-data endpoint can mint a realm for — or in the worst case rewrite the realm list of — an account that already has realms.**

**The mechanism.** `getUserRealmsFromMatrixAccountData` (`packages/boxel-cli/src/lib/auth.ts`) returns `[]` in three failure cases, not just the confirmed-absent one: a thrown fetch (network error), *any* non-ok status other than 401/403 — a 500 or a proxy 502 falls into the `!response.ok` branch whose comment only contemplates 404 — and a JSON parse failure. Its pre-existing callers (`getUserRealms`, `addToUserRealms` in `profile-manager.ts`) use it for listing and user-initiated adds, where empty-on-failure is conservative. This function is the first place that uses it as the gate for an *automatic mutation*, and there empty-on-failure inverts: the failure mode is action, not inaction.

**The failure path.** Take an account that already has realms, and a transient 502 on the account-data GET:
- `realms.length === 0` reads as "has none", so the bootstrap proceeds to `_create-realm` with endpoint `personal`.
- If that account's `personal` endpoint happens to be free (its personal realm was deleted, or it only ever had other realms), a brand-new realm is created for an established account.
- If `personal` exists, the collision path calls `addRealmToMatrixAccountData`, which *re-reads the list with the same best-effort semantics*. The two reads hit the same endpoint seconds apart, so their failures are correlated — and if the second read also comes back empty while the PUT succeeds, the account's realm list is replaced with `{ realms: [<personal-url>] }`, silently dropping every other realm from it.

**The way out.** Gate on a confirmed answer: either a strict variant of the read used here (and by `addRealmToMatrixAccountData` when called from this path) that throws on anything other than 2xx or a 404, or a return shape that distinguishes `{ confirmed: true, realms }` from a failed read. The caller in `profile.ts` already converts a throw into a warning plus a pointer at the web app, which is exactly the right degrade for "couldn't determine".

**Scope.** The helper's semantics are pre-existing, but using them as a creation gate is introduced here — recommend fixing in this PR. This is the one finding I'd treat as blocking-leaning; the fix is small.

packages/boxel-cli/tests/lib/personal-realm.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    );
+  });
+
+  it('re-links a personal realm that exists on the server but is missing from the realm list', async () => {
+    const { calls } = stubFetchRoutes({
+      createRealmResponse: new Response(
+        `realm 'https://realms.example.com/newuser/personal/' already exists on this server`,
+        { status: 400 },
+      ),
+    });
```
lukemelia, line 173: [Claude Code 🤖] **This stub pins a wire shape the server never sends — the real collision response is a JSON:API envelope, not plain text — so the test validates the parser against the wrong body.** Please make the stub mirror the real envelope.

**What the server actually sends.** `handleCreateRealmRequest` (`packages/realm-server/handlers/create-realm.ts`) routes the 400 through `sendResponseForBadRequest` → `sendResponseForError` (`packages/realm-server/middleware/index.ts`), which serializes it as `{"errors":["realm '<url>' already exists on this server"]}` with a JSON:API content type.

**Verified.** I ran the parser against that exact envelope: `errorBody.includes('already exists')` matches and the single-quote regex captures the URL, so the production code is correct today — only the fixture is off. But because the parser is substring/regex over raw text, the two shapes don't fail the same way under change: e.g. a parse rewritten to expect plain text, or an envelope change that re-quotes the message, would keep this test green while breaking against the real server.

**The fix** is one line:

```ts
createRealmResponse: new Response(
  JSON.stringify({
    errors: [
      `realm 'https://realms.example.com/newuser/personal/' already exists on this server`,
    ],
  }),
  { status: 400 },
),
```

**Scope.** Regression class (test authored in this change), non-blocking. The end-to-end half of pinning this contract is covered in the comment on the new test in `cli-sso.spec.ts`.

packages/matrix/tests/cli-sso.spec.ts
```diff
@@ -106,6 +110,84 @@ test.describe('boxel-cli browser authorization', () => {
...
+      expect(result).toEqual({
+        outcome: 'created',
+        realmUrl: `${serverIndexUrl}/${username}/personal/`,
+      });
+
+      // Running it again — a second `boxel profile add` for the same account —
+      // must leave the account alone rather than erroring or double-linking.
+      expect(await ensurePersonalRealm(auth, `${serverIndexUrl}/`)).toEqual({
+        outcome: 'has-realms',
+      });
```
lukemelia, line 170: [Claude Code 🤖] **The `linked` outcome — the one path in `ensurePersonalRealm` that depends on parsing the realm server's error prose — is exercised only against a hand-written stub; nothing executable ties the CLI's parser to the string the server actually emits.** This test already has everything needed to pin that contract; please add a third leg.

**Why it matters.** The CLI recognizes the collision by `errorBody.includes('already exists')` plus a single-quoted-URL regex (`packages/boxel-cli/src/lib/personal-realm.ts`), and the producing string lives in `createRealm` (`packages/realm-server/handlers/create-realm.ts`). Neither side knows the other depends on the exact wording, and the unit test's stub merely restates the CLI's own expectation. If the server message is ever reworded, re-link silently degrades to the generic throw — which `profile add` turns into a warning — and the account whose realm list was lost stays realm-less: the exact state this bootstrap exists to prevent. No test would notice.

**The way out.** After the existing assertions, simulate the lost-realm-list state and run the bootstrap again — `updateAccountData` (`packages/matrix/support/synapse/index.ts`) can clear the list directly:

```ts
// The realm exists on the server but the account's realm list has lost it —
// the bootstrap must re-link via the server's collision response rather than
// failing on it. This is the one leg that pins the CLI's parsing of that
// response against what the realm server actually emits.
await updateAccountData(
  auth.userId,
  auth.accessToken,
  APP_BOXEL_REALMS_EVENT_TYPE,
  JSON.stringify({ realms: [] }),
);
expect(await ensurePersonalRealm(auth, `${serverIndexUrl}/`)).toEqual({
  outcome: 'linked',
  realmUrl: `${serverIndexUrl}/${username}/personal/`,
});
```

(plus re-asserting the account data afterwards). A marker comment on the server-side message noting that the CLI parses it would help too, but the executable check is the real guard.

**Scope.** Test-coverage gap for a cross-package contract introduced by this change; non-blocking, recommended.

lukemelia (commented):
[Claude Code 🤖] This bootstrap writes durable account state from a CLI based on reads and error prose from three different services (Matrix account data, the Matrix profile API, the realm server), so the review focused on the failure modes of those cross-service contracts, plus fidelity with the web signup twin.

**Bottom line: no blocking issues on the happy path, and the web-twin mirroring is faithful — but one finding is worth fixing in this PR: the zero-realms gate trusts a best-effort read that reports "no realms" on a failed fetch, which can mint a realm for (or in a correlated-failure case, rewrite the realm list of) an account that already has realms.** The other two findings are test-fidelity and contract-pinning improvements.

Recommendations:

1. Only proceed with creation on a *confirmed*-empty realm list — see the comment on the realm-list read in `personal-realm.ts`. Worth fixing in this PR.
2. Stub the collision response in the unit test with the server's real JSON:API envelope — see the comment in `personal-realm.test.ts`. Non-blocking.
3. Pin the re-link contract end-to-end with a third leg in the new e2e test — see the comment in `cli-sso.spec.ts`. Non-blocking, recommended.

Adjacent, out of scope: a first-time Google sign-in in the web app itself has the same gap this PR closes for the CLI — the web SSO login path starts a session without ever running the new-user initialization, so a user whose first-ever contact is web SSO also lands realm-less. Different surface, deserves its own fix; `ensurePersonalRealm`'s confirmed-zero-realms check is the pattern to mirror. Separately, `addRealmToMatrixAccountData`'s read-modify-write of the realm list is unguarded against concurrent writers for all its callers — pre-existing, noted only because this PR adds another writer.

packages/boxel-cli/src/lib/personal-realm.ts
lukemelia: [Claude Code 🤖] Fixed in 1698612dca. `requireUserRealmsFromMatrixAccountData` is the strict variant: a confirmed answer or a throw — a 404 confirms an empty list, while network errors, any other non-2xx status, and unparseable bodies all throw. The bootstrap gate uses it, and `addRealmToMatrixAccountData` builds its read-modify-write on the strict read as well, so the PUT body can never be assembled from a failed read's empty guess. A throw surfaces through `profile add`'s existing degrade — a warning pointing at the web app. Unit tests pin the gate: an unreadable (502) or unparseable realm list throws without any `_create-realm` call or realm-list write.

packages/boxel-cli/tests/lib/personal-realm.test.ts
lukemelia: [Claude Code 🤖] Fixed in 1698612dca — the stub sends the JSON:API error envelope (`{"errors":[…]}`) that `sendResponseForError` in the realm server middleware produces, with a comment on the fixture noting where the shape comes from.

packages/matrix/tests/cli-sso.spec.ts
lukemelia: [Claude Code 🤖] Added in 1698612dca. The test clears the account's realm list via `updateAccountData` and runs the bootstrap again, asserting the `linked` outcome with the realm URL; the closing account-data assertion then confirms the list names the personal realm once more. The collision message in `create-realm.ts` also carries a marker comment noting that boxel-cli parses the "already exists" phrase and the single-quoted URL, pointing at this spec as the executable guard.

backspace (approved):
I tried this out locally, it worked as expected

packages/boxel-cli/src/lib/personal-realm.ts
habdelra: [Claude Code 🤖] **This resolves the finding.** Verified against the head commit: `readUserRealmsFromMatrixAccountData` now takes the 404 branch *before* the `!response.ok` branch, so a confirmed-absent event returns `[]` while a 502, a thrown fetch, and an unparseable body all throw under `'strict'`. Both halves of the failure path are closed — the gate in `ensurePersonalRealm` and the read-modify-write inside `addRealmToMatrixAccountData` — so the PUT body can no longer be assembled from a guess. The two new unit tests cover exactly the two throw paths that mattered (502 and non-JSON), asserting no `_create-realm` and no PUT.

Two consequences of widening `addRealmToMatrixAccountData` to the strict read, both downstream of this thread rather than new problems with the fix:

**`removeRealmFromMatrixAccountData` is the remaining read-modify-write on the best-effort read.** The new doc comment on `getUserRealmsFromMatrixAccountData` states "Never gate a mutation on this", and that function's other caller in the same file gates a PUT on it: `removeRealmFromMatrixAccountData` reads the list, filters, and writes only when the length changed. It fails safe today — a failed read yields `existing = []`, `next = []`, the lengths match, so it returns `false` without writing, and `removeRealm` in `commands/realm/remove.ts` has already returned `notInList` before reaching it (it reads the list itself first via `pm.getUserRealms()` and short-circuits before the server-side `DELETE /_delete-realm`). So no data loss and no premature delete — but the user is told `Realm is not in app.boxel.realms. Nothing to remove.` when the truth is "couldn't read your realm list", and the file now documents an invariant it doesn't keep. Switching that read to `requireUserRealmsFromMatrixAccountData` makes the failure legible and makes the comment true; `removeRealm` already converts a throw into `error: 'Failed to load realm list: …'`. Class: pre-existing, now load-bearing (this PR documents the rule). Non-blocking.

**`boxel realm create` now swallows a new failure class silently.** `createRealm` in `commands/realm/create.ts` wraps `pm.addToUserRealms(realmUrl)` in `try { … } catch { /* Non-critical */ }`. Before this change a failed read still produced a PUT (the corrupting one); now it throws into that empty catch, so `realm create` can print `Realm created: <url>` while the realm never lands in the account's realm list, with no output at all. Worth a `console.warn` in that catch naming the realm URL and pointing at re-linking. Class: follow-up, non-blocking.

---

packages/boxel-cli/tests/lib/personal-realm.test.ts
habdelra: [Claude Code 🤖] **This resolves the finding.** Verified the fixture against the producer: `sendResponseForError` in `packages/realm-server/middleware/index.ts` serializes as `JSON.stringify({ errors: [message] })` — a flat array of strings, not error objects — with `content-type: application/vnd.api+json`, and `sendResponseForBadRequest` is the 400 wrapper `handleCreateRealmRequest` routes `errorWithStatus(400, …)` through. The stub now matches that shape exactly, including the single-quoted URL surviving JSON encoding, so the substring-plus-regex parse is being exercised against the bytes the server actually emits.

---

packages/matrix/tests/cli-sso.spec.ts
habdelra: [Claude Code 🤖] **This resolves the finding.** Traced the added leg: `updateAccountData` PUTs `{ realms: [] }`, so the strict read returns a confirmed-empty list (200, not 404), the bootstrap proceeds to `_create-realm`, the registry row from the first leg makes `existingRows.length > 0`, and the 400 body carries `realm 'https://localhost:4205/<username>/personal/' already exists on this server`. That is now the one place the CLI's `includes('already exists')` + single-quoted-URL regex meets real server output, and the closing account-data assertion confirms the list was rebuilt rather than emptied. The `has-realms` leg in between also pins the second-run no-op.

One caveat on the marker comment added alongside it: it names `personal-realm.ts` as *the* consumer, and there are three. Detail in a new comment on `handlers/create-realm.ts`.

---

packages/realm-server/handlers/create-realm.ts
```diff
@@ -130,6 +130,11 @@ export async function createRealm(
     param(url),
   ])) as { url: string }[];
   if (existingRows.length > 0) {
+    // boxel-cli's ensurePersonalRealm recognizes this collision by the
+    // phrase "already exists" and extracts the single-quoted URL to re-link
+    // the realm; keep both if rewording (packages/boxel-cli/src/lib/
+    // personal-realm.ts, pinned end-to-end by packages/matrix/tests/
+    // cli-sso.spec.ts).
```
habdelra, line 137: [Claude Code 🤖] **This marker names one of three dependents, and the one it omits in the host needs a *longer* phrase than the two it points at — so a reword guided by this comment can pass both CLI call sites and silently break workspace duplication.** Please either name all three or, better, give the message a single home.

**The three consumers, enumerated.** `grep -rn "already exists"` across `packages/boxel-cli/src`, `packages/host/app`, and `packages/runtime-common` turns up three live parses of this exact throw:

- `packages/boxel-cli/src/commands/realm/create.ts` — `errorBody.includes('already exists')`, then `extractRealmUrlFromError` regexes the single-quoted URL. Pre-existing; this is what `boxel realm create` uses to report "already exists" instead of failing.
- `packages/boxel-cli/src/lib/personal-realm.ts` — the same two steps, re-implemented inline. Added by this change; the one the comment names.
- `packages/host/app/services/workspace-duplication.ts` — `if (!message.includes('already exists on this server')) { throw error; }`, inside the endpoint-numbering loop that walks `<endpoint>-copy`, `-copy-2`, … until one is free. `message` is the string `RealmServerService.createRealm` builds, which embeds this response body verbatim, so the match lands on the same words.

**Why the asymmetry bites.** The two CLI sites match `already exists`; the host site matches `already exists on this server`. Reword to, say, `realm '<url>' already exists`, and both CLI paths keep working while `workspace-duplication` stops recognizing the collision: the loop rethrows on the first attempt, so duplicating a workspace fails outright instead of picking the next free endpoint. Nothing in this file would hint at that, because the comment says the dependents are in boxel-cli.

**The way out**, in preference order:

1. Give the phrase one home. Both `@cardstack/runtime-common/realm-display-defaults` and `@cardstack/runtime-common/paths` are already imported by boxel-cli *and* the host, so a sibling module exporting `realmAlreadyExistsMessage(url)` plus `isRealmAlreadyExistsError(body)` is reachable from all four sites (producer + three consumers). Then a reword is a one-line change in the shared module and the consumers cannot drift — the executable version of what this comment is trying to do by hand.
2. If that is more than this PR should carry, extend the comment to name all three files, and say which phrase each one needs — the host's dependency on `on this server` is the part a rewriter must not miss.

A follow-up worth filing either way: `_create-realm` could answer a collision with a machine-readable shape (a 409, or a JSON:API error carrying a `code` and the existing realm URL as a field) so none of the consumers parse prose at all. That's out of scope here.

**Scope.** The marker comment is new in this change, so the incomplete-documentation half is a regression; the underlying prose-parsing fragility is pre-existing and this PR adds the third dependent. Non-blocking.

---

packages/boxel-cli/src/lib/personal-realm.ts
```diff
@@ -0,0 +1,144 @@
...
+      if (realmUrl) {
+        realmUrl = ensureTrailingSlash(realmUrl);
+        await addRealmToMatrixAccountData(auth, realmUrl);
+        return { outcome: 'linked', realmUrl };
+      }
+    }
+    throw new Error(
+      `Could not create the personal realm: realm server returned ${response.status}: ${errorBody}`,
+    );
+  }
```
habdelra, line 95: [Claude Code 🤖] **This collision handler is a second copy, inside the same package, of the parse that `commands/realm/create.ts` already performs — and the copy drops two guards the original has.** Please import the existing helper rather than re-deriving it.

**The original.** `createRealm` in `packages/boxel-cli/src/commands/realm/create.ts` does the same four steps against the same endpoint: `errorBody.includes('already exists')` → `extractRealmUrlFromError(errorBody, realmServerUrl, realmName)`, whose body is `errorBody.match(/'(https?:\/\/[^']+)'/)` + `ensureTrailingSlash` — character-for-character the regex on line 85 — → link into the realm list → otherwise `throw new Error(\`Realm server returned ${response.status}: ${errorBody}\`)`. Same assembly for the success path too: `data.id`, the `typeof … !== 'string' || trim() === ''` check, `ensureTrailingSlash`.

**What the copy loses.** Two guards exist in `create.ts` and not here:

- **No status check.** `create.ts` reaches its parse the same way, so this is a shared weakness rather than a new one, but it matters more here because this path *mutates the realm list on a match*: `handleCreateRealmRequest`'s 500 branch responds with `` `${e.message}: at ${e.stack}` ``, so any future 500 whose message or stack contains the words "already exists" plus a single-quoted URL would be read as a collision and linked. Gating on `response.status === 400` costs one condition and confines the parse to the branch that actually produces this message.
- **No non-serialized-body guard.** `create.ts` normalizes a body containing `[object Object]` into a readable message; here such a body lands raw in the thrown string, which then reaches the user through `profile add`'s warning line.

**Why not just call `createRealm`.** Reusing the whole function isn't available: it resolves its credentials through `getProfileManager().getActiveProfile()`, and `addProfileWithAuth` only promotes the new profile to active when there was no active profile (`if (!this.config.activeProfile)`), so a user adding a second profile would have the realm created against whichever account was already active. Taking `auth` explicitly is right. That makes the extraction the narrow one: export `extractRealmUrlFromError` (and ideally the `already exists` predicate — see the comment on `handlers/create-realm.ts`) and have both sites call it, so the two copies cannot answer differently about what the server said.

**Scope.** Regression class — the duplicate is introduced here. Non-blocking, but this is the finding I'd most want in this PR, since it is what keeps the cross-package contract to a single reader.

---

packages/boxel-cli/src/lib/personal-realm.ts
```diff
@@ -0,0 +1,144 @@
...
+  getRealmServerToken,
+  requireUserRealmsFromMatrixAccountData,
+  type MatrixAuth,
+} from './auth.ts';
+
+// The web app's signup flow ends by creating this realm for the new account
+// (see MatrixService.initializeNewUser in the host); these values mirror it so
+// an account bootstrapped from the CLI is indistinguishable from one
+// bootstrapped from the web.
+const PERSONAL_REALM_ENDPOINT = 'personal';
```
habdelra, line 18: [Claude Code 🤖] **The comment claims parity with the web bootstrap ("indistinguishable"), but two of the four attributes are duplicated literals with nothing tying them to the web's copy — so the claim is an invariant no code or test enforces.** Please move the two shared decisions into the module both sides already import.

**What is shared and what is copied.** The web's bootstrap is `MatrixService.initializeNewUser` → `createPersonalRealmForUser` (`packages/host/app/services/matrix-service.ts`), which posts:

```ts
endpoint: 'personal',
name: `${displayName}'s Workspace`,
iconURL: iconURLFor(displayName),
backgroundURL: getRandomBackgroundURL(),
```

`iconURL` and `backgroundURL` genuinely are shared — both sides call the same `@cardstack/runtime-common/realm-display-defaults` (the host through its `app/lib/utils` re-export). The other two are literals in two places: `grep -rn "endpoint: 'personal'\|'personal'"` across `packages/host/app`, `packages/realm-server`, `packages/boxel-cli/src`, and `packages/runtime-common` returns exactly `matrix-service.ts:793` and `PERSONAL_REALM_ENDPOINT` here; `grep -rn "s Workspace"` over the same tree returns exactly `matrix-service.ts:794` and line 54 here (the other hits are boxel-ui usage stubs).

**Why the tests don't catch drift.** Each convention is pinned twice, independently, never compared: the web's name by `packages/matrix/tests/registration-with-token.spec.ts` (`[data-test-workspace="Test User's Workspace"]`, and `realms: ['https://localhost:4205/<username>/personal/']`), the CLI's by the new unit test (`expect(attributes.name).toBe("New User's Workspace")`, `endpoint` `'personal'`) and the new e2e leg. Change the web's template to, say, `${displayName}'s Space` and `registration-with-token.spec.ts` goes red, you update it, and every CLI assertion still passes — the CLI keeps minting the old convention, and an account bootstrapped from `profile add` is now distinguishable from a web one in exactly the way this comment promises it isn't. The endpoint is the more consequential half of the two: it's the URL path segment, so a drift there means two realms per account rather than one differently-named realm.

**The way out.** Add to `@cardstack/runtime-common/realm-display-defaults` (already a dependency of both packages, and already the home for the other two attributes):

```ts
export const PERSONAL_REALM_ENDPOINT = 'personal';
export function personalRealmName(displayName: string): string {
  return `${displayName}'s Workspace`;
}
```

then have `matrix-service.ts` and this module both call them. That converts the comment's claim into something the compiler keeps true, and lets the comment shrink to naming where the defaults live.

**Scope.** Regression class in the sense that this PR creates the second copy; the fix is small and belongs here. Non-blocking.

---

packages/matrix/tests/cli-sso.spec.ts
```diff
@@ -106,6 +110,100 @@ test.describe('boxel-cli browser authorization', () => {
...
+    // A brand-new account, its localpart derived from the email's local part.
+    expect(auth.userId).toBe(`@${username}:localhost`);
+
+    // The realm server speaks HTTPS with the mkcert leaf, which this runner
+    // process has no CA path to (NODE_EXTRA_CA_CERTS is only guaranteed in
+    // mise-run children — see isolated-realm-server.ts, which relaxes its
+    // spawned services the same way). Loopback only, so relax validation just
+    // around the bootstrap calls.
+    const priorTlsSetting = process.env.NODE_TLS_REJECT_UNAUTHORIZED;
+    process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
habdelra, line 158: [Claude Code 🤖] **Was the relaxation actually needed under CI, or only in a shell without the mise env?** The answer decides where it belongs, and the in-test form has a failure mode worth avoiding either way.

**What I could establish.** `NODE_EXTRA_CA_CERTS` is exported by `mise-tasks/lib/env-vars.sh` (sourced via `.mise.toml`'s `_.source`), guarded on `mkcert` being present, so it reaches `mise run` children and an activated shell. In CI the Playwright runner is started by `pnpm test:group` → `packages/matrix/scripts/test.sh`, which does not source it and does not set it, and `.github/actions/init` doesn't publish it to `$GITHUB_ENV`. So whether the runner trusts the mkcert leaf depends on whether `jdx/mise-action` exported the env for the job — which I couldn't settle from the tree.

**Why it isn't answered by precedent.** No other spec makes a runner-side call to `https://localhost:4205`. The one that looks like it — `host-mode.spec.ts`'s `fetch('https://localhost:4205/_publish-realm', …)` — runs inside `page.evaluate`, i.e. in Chromium, which launches with `--ignore-certificate-errors --allow-insecure-localhost`. `publish-realm.spec.ts` uses `request.get(…)`, which honors `ignoreHTTPSErrors: true` from `playwright.config.ts`. Neither applies to `ensurePersonalRealm`, which calls global `fetch`, so the fixture can't be injected — that part of the reasoning holds.

**The hazard in the in-test form.** `process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0'` is process-wide, and the `finally` that restores it only runs if the block settles. A Playwright timeout abandons the test body, so the restore doesn't run; the blast radius is bounded only by Playwright replacing the worker after a failure. It also hides the thing worth knowing: if the runner's CA trust regresses, every other spec fails loudly and this one keeps passing.

**The way out**, whichever answer applies:

- **Needed in CI** → set it once for the runner in `packages/matrix/scripts/test.sh`, alongside the `START_SERVER_AND_TEST_INSECURE=1` escape hatch already there for the same class of problem (or in `tests/global.setup.ts`, whose env assignments are inherited by workers). One place, no test mutating a global, and every runner-side fetch behaves identically.
- **Only outside mise** → drop it from the test and let the local case be fixed by running the suite through mise, which is how every other service in this harness gets its trust.

**Scope.** Test-hygiene finding on code added in this change; non-blocking, and it doesn't affect what the test proves.

---

habdelra (commented):
[Claude Code 🤖] Re-review of the head commit. Two lenses: whether each of the three fixes closes the mechanism its thread named (rather than the symptom), and — since all three fixes are about a contract shared across packages — whether the decisions this bootstrap now depends on have a single home.

**Bottom line: all three threads are resolved, verified against the code rather than the replies, and there are no blocking issues. Every new finding is the same shape: a contract this change now relies on is stated in a comment instead of enforced by code.**

Dispositions (detail in each thread):

- Strict realm-list read — resolved. The 404 branch precedes `!response.ok`, and both the gate and the read-modify-write inside `addRealmToMatrixAccountData` use the strict variant. Two consequences downstream of that widening are noted in the thread: `removeRealmFromMatrixAccountData` is now the only read-modify-write still on the best-effort read (fails safe, but misreports "not in your list" on an unreadable list, and contradicts the invariant the new comment states), and `boxel realm create`'s `catch { /* Non-critical */ }` now swallows "couldn't read" with no output at all.
- Unit-test fixture — resolved. Checked against `sendResponseForError`: `{ errors: [message] }`, flat strings, JSON:API content type. The stub matches.
- End-to-end re-link leg — resolved. The added leg is the only place the CLI's parse meets real server output, and it does reach the collision branch (confirmed-empty list → `_create-realm` → existing registry row → 400).

New findings:

1. The collision marker in `handlers/create-realm.ts` names one of three dependents, and the omitted one (`workspace-duplication.ts` in the host) matches a longer phrase than the two it names — so a reword guided by this comment breaks workspace duplication while both CLI paths keep working. Name all three, or give the message a single home in runtime-common. → thread on `handlers/create-realm.ts`.
2. `ensurePersonalRealm`'s collision handler re-implements the parse that `commands/realm/create.ts` already has (`extractRealmUrlFromError` is the same regex), minus that file's status/body guards. Import it. Includes why calling `createRealm` wholesale isn't available — it resolves through the *active* profile, which a second `profile add` isn't. → thread on `personal-realm.ts`.
3. `endpoint: 'personal'` and `` `${displayName}'s Workspace` `` are duplicated literals shared with `MatrixService.initializeNewUser`, pinned by two independent tests that never compare, so the comment's "indistinguishable from one bootstrapped from the web" is unenforced. Both sides already import `realm-display-defaults`; move the two constants there. → thread on `personal-realm.ts`.
4. The `NODE_TLS_REJECT_UNAUTHORIZED` window in the new e2e test — a question about whether CI needs it, plus where it should live if so. → thread on `cli-sso.spec.ts`.

Verified clean, for what it's worth as scope boundaries: `_create-realm` and `_server-session` have no subscription gate (`ensureSessionRoom` auto-provisions the user row and DM), so a freshly minted Google account can complete the bootstrap in production, not just in the test harness; the register path of `/cli-auth` awaits `initializeNewUser` before `onComplete`, so there is no race between the web bootstrap and this one, and the PR description's no-op claim holds; not caching the new realm's token is fine, since `getRealmTokenForUrl` mints lazily on first use; and the two other 400s in `createRealm` say "already mounted at the origin", so the CLI's substring can't misfire on them.

Adjacent, out of scope:

- `ensurePersonalRealm` returns on the 202 without waiting for readiness, so "✓ Created your personal workspace" can precede the realm being servable. The web bootstrap behaves identically, and `boxel realm create` has `waitForReady` if a follow-up ever wants parity in the other direction.
- The bootstrap's success line prints before the caller's "✓ Profile created", so the output reads workspace-then-profile.
- Nothing pins the wiring in `addProfileViaBrowser` (that the bootstrap runs on the browser path, and that a failure warns instead of failing the add) — `addProfileViaBrowser` is module-private, so testing it would mean extracting the announce/degrade block first. Only worth doing if that degrade behavior gets more complicated than it is today.

---

PR #5836, opened 2026-08-20
Fill the isolated image preview to its stage frame

lukemelia:
## Problem

In the isolated view of an image FileDef, a small image rendered as its top edge only, sitting at the bottom of a mostly-empty stage frame.

The `` carries its intrinsic pixel dimensions as `width`/`height` HTML attributes. In `image-preview.gts` the renderer relied on `height: 100%`, but the image is a grid child of the preview stage (`display: grid; place-items: center`), whose implicit row is `auto`. In that layout the percentage height never resolves, so the browser derived the box height from `width: 100%` × the aspect ratio implied by those attributes — a box taller than the 380px stage. `object-fit: scale-down` then drew the picture at its natural size centered inside that oversized box, which placed it below the visible frame; `overflow: hidden` clipped everything but the top edge.

Reproduced in a browser: for a 154×160 image in a 740px-wide stage, the `` box measured **740×769** — matching the reported symptom exactly.

## Fix

`.image-preview` now fills the frame the way the stage's own `.generated-thumbnail` sibling already does: an absolute box (`position: absolute; inset: 0`) against the relatively-positioned stage resolves both axes to the frame, and `object-fit` fits the pixels within a correctly-sized box. `box-sizing: border-box` keeps the `contain` mode's padding inside the frame.

After the fix the same image measures **740×380** (the stage box) across all three fit modes — scale-down (small image centered on the paper matte), cover (fills + crops), and contain (letterboxed).

## Test

Adds a regression test to `file-def-format-templates-test.gts` that renders an isolated `ImageDef` and asserts the rendered `` height fills the stage rather than overflowing it.

### DIFF
--- a/packages/base/file-formats/image-preview.gts
+++ b/packages/base/file-formats/image-preview.gts
@@ -60,10 +60,20 @@ export class ImagePreview extends GlimmerComponent<FilePreviewSignature> {
       data-test-image-preview
     />
     <style scoped>
+      /* Fill the stage frame like the stage's own thumbnail sibling does: an
+         absolute box against the relatively-positioned stage resolves both axes
+         to the frame. A plain height:100% grid child does not — the stage's auto
+         row leaves the percentage unresolved, so the browser falls back to the
+         intrinsic aspect ratio from the img's width/height attributes and
+         overflows a frame taller than it. object-fit then does the fitting
+         within a correctly sized box. */
       .image-preview {
-        display: block;
+        position: absolute;
+        inset: 0;
         width: 100%;
         height: 100%;
+        box-sizing: border-box;
+        display: block;
       }
       .image-preview[data-image-fit='cover'] {
         object-fit: cover;
--- a/packages/host/tests/integration/components/file-def-format-templates-test.gts
+++ b/packages/host/tests/integration/components/file-def-format-templates-test.gts
@@ -1,4 +1,5 @@
 import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import { find } from '@ember/test-helpers';
 import type { RenderingTestContext } from '@ember/test-helpers';
 
 import GlimmerComponent from '@glimmer/component';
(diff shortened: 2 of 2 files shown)


lukemelia (commented):
[Claude Code 🤖]

This review focused on two things: whether the CSS fix actually resolves the overflow by the mechanism described, and whether the added test is a genuine regression guard rather than one that would pass with or without the fix.

**Bottom line: no blocking issues. The fix is correct and the test genuinely guards it — both verified in Chrome against the exact stage structure.**

**Fix verified.** I reproduced the isolated stage layout standalone (`position: relative; height: 380px` frame → `.stage { height:100%; display:grid; place-items:center }` → the `` grid child with `width=154 height=160`):

- Old CSS (`display:block; width:100%; height:100%`): the `` box measured **740×768.8** — `height:100%` never resolves against the auto grid row, so the box is sized from `width:100%` × the `160/154` attribute aspect ratio (`740 × 160/154 = 768.8`), overflowing the 380px frame. This holds whether the image loads or not (broken src → `naturalWidth 0` → same 768.8), which matches the reported symptom.
- New CSS (`position:absolute; inset:0; width:100%; height:100%; box-sizing:border-box`): the box measured **740×380** — it fills the frame and `object-fit` does the fitting inside a correctly-sized box.

**Test is a real guard.** Because the fixture's `http://example.com` src does not decode to an image, the `` keeps the `154/160` attribute ratio, so under the old CSS the box is `stageWidth × 160/154` (≈768.8 at 740px) — `Math.abs(imgHeight - stageHeight)` is ~389, well past the 1px tolerance, so the assertion fails without the fix and passes with it. The guard does not depend on network or image decoding.

**Blast radius is safe in all three shells.** `ImagePreview` renders in fitted, embedded, and isolated, so this CSS change applies to all of them, not only the isolated case in the description. The absolute box can't collapse the frame because every stage host is `position: relative` with a definite height — isolated `380px`, embedded fixed-px or `aspect-ratio`-derived, fitted the `minmax(0,1fr)` grid row — and `.generated-thumbnail` in the same stage already uses this identical `position:absolute; inset:0` pattern, so the invariant this fix leans on is already load-bearing here.

habdelra (approved):

PR #5837, opened 2026-08-20
Make the pinned-card Open action a filled button

lukemelia:
## What

On the workspace card's Pinned section, each pinned card's "Open" control was a small borderless text link, easy to miss next to the tile. It is now a compact filled-accent button using the same treatment as the workspace card's New button (accent background, dark ink, 8px radius, subtle inset highlight), with the matching darker-teal hover.

## Test plan

- CSS-only change to `.door-open` in the workspace card; no markup or behavior changes.
- glint and ember-template-lint are clean on the changed file; no tests reference the styling.

### DIFF
--- a/packages/base/workspace.gts
+++ b/packages/base/workspace.gts
@@ -1942,13 +1942,20 @@ class Isolated extends Component<typeof Workspace> {
       }
       .door-open {
         flex: 0 0 auto; /* Keep the action inside its tile while the label ellipsizes */
-        border: none;
-        background: none;
-        padding: 0;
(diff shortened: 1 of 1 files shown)


habdelra (approved):

packages/base/workspace.gts
```diff
@@ -1942,13 +1942,20 @@ class Isolated extends Component<typeof Workspace> {
       }
       .door-open {
```
burieberry, line 1943: why not use a primary button? you can use one of the smaller sizes with rectangular arg

PR #5840, opened 2026-08-20
Screenshot serving internals: MediaCache streaming, realm-read auth, cache headers

lukemelia:
Stacked on #5838 (the MediaCache store); this PR adds the serving side that streams from it.

## What this adds

**The `_screenshot/{instanceLocalPath}` realm route.** Dispatch is by path prefix inside `internalHandle` rather than the router table: the router keys routes on the `Accept` header, and the requests this route exists to serve — `` loads, og:image fetches — send `image/*`-shaped Accept values that match no supported mime type. The branch sits after `checkPermission`, so the route inherits realm-read auth exactly like any realm resource: a world-readable realm serves unauthenticated (crawlers, og:image), a private realm 401s without a JWT and 403s without the read grant. The dispatch is GET-only — `checkPermission` exempts HEAD from auth realm-wide, so admitting HEAD would hand unauthenticated callers an existence/size/content-hash oracle over a private realm's captures.

Beyond realm read, the parent instance must be live in the index — per-instance ACLs get a place to land, and captures of a deleted instance stop serving as soon as its tombstone appears, ahead of GC reclaiming them. The gate is a narrow existence probe (`hasLiveInstance`: `SELECT 1` with `getInstance`'s exact predicate, effective-error channel included), never an instance hydration — it runs on every request, 304 revalidations included, and `getInstance` selects every wide column in the index.

**The `_screenshot/` subtree is reserved.** Since the GET dispatch claims the whole prefix, a realm file stored under it could never be read back. Creation writes (PUT/PATCH/POST, and `/_atomic` operations targeting the subtree) are refused up front so the collision surfaces at write time; DELETE stays admitted as the recovery path.

**The serving core** (`packages/runtime-common/media-cache-serving.ts`, browser-safe):

- Streams the ledger entry's object via `nodeStream` — the one body shape the realm-server's Koa bridge streams verbatim (every other shape is drained through text, which corrupts image bytes). An adapter's node `Readable` passes through; a bare async iterable is wrapped by a node/browser conditional import.
- `Content-Type` comes from the ledger (the adapters store no metadata); `ETag` is the entry's object key — the content hash — so revalidation is exact and a re-capture rotates the validator while the URL stays put.
- `Cache-Control` is `public`/`private` by realm world-readability (same derivation as `serveLocalFile`), `max-age=60, stale-while-revalidate=3600`; `If-None-Match` matches (via the RFC-9110 helper: `*`, lists, `W/`) answer as bodyless 304s. The SWR window is a deliberate trade: an image grid renders from cache while revalidating in the background, at the cost of a re-capture taking up to max-age + SWR to reach every viewer — acceptable for derived, eventually-consistent captures whose URL never lies about identity.
- A hit — 200 or 304 — bumps the entry's `last_accessed_at`, which is what keeps an in-use on-demand capture out of the GC's age-out lane. The bump is lane-scoped and throttled: only the GC's on-demand arm reads the stamp (declared captures age out on generation, so their serves skip the write entirely), and on-demand bumps write at most hourly per entry — never an unthrottled UPDATE ahead of every response body. Best-effort either way: a failed bump never fails a serve.
- An entry whose object was reclaimed between ledger read and stream open serves as a miss, not an error.
- Misses are 404s carrying the same short max-age, never a synchronous wait: an `` pointing at a not-yet-captured name picks the image up on a later revalidation.

**Addressing is a seam, not yet a resolver.** `name=` (declared-screenshot manifests) and capture-spec canonicalization plug into `resolveScreenshotEntry`; with neither resolver present, every request is an uncaptured miss. `findMediaCacheEntry` (exact-generation or newest-generation ledger lookup) is the query those resolvers will use. The ledger's `source_url` convention is pinned to the extensionless card-id form, and the GC tombstone join matches either `boxel_index.url` or `file_alias` so both URL shapes reclaim correctly.

**Threading**: `MediaCacheAdapter` is now a Realm constructor option, wired from `MEDIA_CACHE_BUCKET`/`MEDIA_CACHE_DIR` once per server and shared by every realm it mounts; without one, the route serves every request as a miss.

## Test plan

- `media-cache-serving-test` (real Postgres): hit streaming with content-hash validators, public/private cache-control, 304 across all `If-None-Match` shapes, the wrapped-iterable path, reclaimed-object miss, and the last-accessed bump policy (bumps on 200/304 past the throttle window, skipped within it, skipped entirely for declared-lane entries).
- `realm-endpoints/screenshot-test` (booted realm server with a fake-backed MediaCache store, supertest): private realm 401/403/404-with-private-cache-control, HEAD falls through to the generic handlers, `name=` miss, public realm unauthenticated miss with public cache-control, and the write reservation (PUT/PATCH/POST 400, `/_atomic` 422 with a `Reserved path` error). Requests traverse the full serving path — liveness probe, resolver seam — and miss because nothing is captured; endpoint-level hit assertions (live instance serves 200, tombstoned instance misses) land with the capture resolver, which is what makes a hit reachable through this route.
- Existing media-cache GC/adapter and atomic-endpoint suites pass; typechecks clean across runtime-common and realm-server.


lukemelia (commented):
[Claude Code 🤖] This review focused on the two contracts this PR actually stakes its correctness on — the response bytes that reach the wire through the realm-server's Koa bridge, and the realm-read auth the route claims to inherit — since the addressing/resolver logic is deliberately a no-op seam here and gets its real test when the resolver lands.

**Bottom line: no blocking issues. The serving core, cache-header policy, content-hash validators, GC-join widening, and the honest all-miss framing all check out. Two items are latent — both harmless in today's all-miss state, both primed to bite the moment the resolver PR turns the hit path on — and want a decision now rather than a surprise then.**

Verifications that came back clean and aren't worth a thread: the `public`/`private` derivation matches `serveLocalFile` exactly; `ifNoneMatchMatches` handles `*`, lists, and `W/` on both sides; the 304-before-object-existence ordering is correct (a matching validator means the client's cached bytes are still valid even if the object was reclaimed); the GC tombstone join's `(i.url = r.source_url OR i.file_alias = r.source_url)` widening matches the same idiom `getInstance` already uses and can't cross-match instances within a realm; and the `_screenshot/` dispatch sits correctly after both `checkPermission` and the archived-realm seal.

Recommendations:
1. The buffered async-iterable branch in `serveMediaCacheEntry` returns a `Uint8Array` body that the Koa bridge drains through `webStreamToText`, corrupting image bytes — dead with today's Readable-returning adapters, but a trap the "safe because screenshot-sized" comment misdirects away from. Route it through `nodeStream` like the streaming branch. (thread on `media-cache-serving.ts`)
2. `HEAD` is exempt from `checkPermission`, so unauthenticated `HEAD /_screenshot/` on a private realm isn't a 401 — a no-op today, an existence/size/content-hash oracle once the resolver returns `200`s. Decide intended-or-not and pin it with a `HEAD` test before the hit path lights up. (thread on the dispatch in `realm.ts`)

Adjacent, out of scope: `mediaCacheVisibility` is a third copy of the `permissions['*']?.includes('read') ? 'public' : 'private'` rule (`serveLocalFile` inlines its own); whoever consolidates the visibility derivation could fold `serveLocalFile` onto the shared helper. Not this PR.

packages/runtime-common/media-cache-serving.ts
```diff
@@ -0,0 +1,176 @@
...
+      body: null,
+      init,
+      requestContext,
+    });
+    response.nodeStream = stream;
+    return response;
+  }
+  // A bare async iterable (the interface's minimum) is buffered whole. Safe
+  // because captures are screenshot-sized, and the entry carries the exact
+  // size; an adapter serving anything large should return a node Readable.
```
lukemelia: [Claude Code 🤖] **The buffered-iterable branch returns a body shape the Koa bridge corrupts — the "safe" comment reasons about size, not the hazard the streaming branch just warned about.** This path builds a `Uint8Array` body and hands it to `createResponse`, which does `new Response(body, …)`. When the realm-server middleware unwraps that response (`packages/realm-server/middleware/index.ts`, the `handlerResponseToContext` block), it takes exactly the branch the streaming path above documents as corrupting: a native `Response` with a `Uint8Array` body exposes `response.body` as a **web `ReadableStream`**, so it misses `if (nodeStream)` and falls into `else if (body instanceof ReadableStream) { ctxt.body = await webStreamToText(body) }` — a UTF-8 decode over the image bytes. Same failure mode the `isNodeReadable` branch exists to avoid; this branch just re-introduces it for the non-Readable stream shape.

**Verified.** `createResponse` (`create-response.ts:14`) returns `new Response(body, …)`; middleware `index.ts:488-501` streams `nodeStream` verbatim and drains any other body (including the `ReadableStream` that a `Uint8Array` body becomes) through `webStreamToText`. The two production adapters both return node `Readable`s (`media-cache/s3-adapter.ts:105` — `response.Body as Readable`; `media-cache/local-disk-adapter.ts:67` — `createReadStream`), so this branch is **dead in production** and the direct unit test (`media-cache-serving-test.ts`, `streamShape='iterable'`) reads `response.arrayBuffer()` without going through the middleware — which is why it stays green while the real serving path would corrupt.

**Scope: regression, non-blocking** (unreachable with today's adapters). But it's a latent trap primed to fire the day any adapter returns a bare async iterable, and the comment's "Safe because captures are screenshot-sized" points at memory cost, not the body-shape mismatch that actually bites.

**Way out:** buffer, then wrap and ride `nodeStream` like the streaming branch, so both stream shapes leave through the one path the bridge streams verbatim:

```ts
  let body = new Uint8Array(entry.sizeBytes);
  let offset = 0;
  for (let chunk of chunks) {
    body.set(chunk, offset);
    offset += chunk.length;
  }
  let response: ResponseWithNodeStream = createResponse({ body: null, init, requestContext });
  response.nodeStream = Readable.from(Buffer.from(body));
  return response;
```

At minimum, correct the comment so it names the real constraint (must exit via `nodeStream`, not "small enough to buffer") — otherwise the next adapter author reads this as blessed.

packages/runtime-common/realm.ts
```diff
@@ -3183,6 +3195,22 @@ export class Realm {
...
         });
       }
+      // Screenshot serving dispatches on the path prefix, not the router
+      // table: the router keys routes on the Accept header, and the browser
+      // requests this route must serve (`<img>` loads, og:image fetches)
+      // send `image/*`-shaped Accept values that match no supported mime
+      // type. Placed after checkPermission so the route inherits realm-read
+      // auth exactly like any realm resource.
+      if (
+        (request.method === 'GET' || request.method === 'HEAD') &&
```
lukemelia: [Claude Code 🤖] **Admitting `HEAD` here means the route does *not* uniformly inherit realm-read auth — `checkPermission` waves `HEAD` through without a JWT, so the PR description's "a private realm 401s without a JWT" holds for `GET` but not `HEAD`.** `checkPermission` (`realm.ts:4237-4248`) returns early — no token required — when `requiredPermission !== 'realm-owner'` and `request.method === 'HEAD'`. A screenshot request is `read`, so an unauthenticated `HEAD /_screenshot/{card}` on a *private* realm sails past auth and into `serveScreenshot`.

**Why it's latent, not live today.** `resolveScreenshotEntry` always returns `undefined`, so every request — `GET` or `HEAD` — resolves to `mediaCacheMissResponse` (404), and even the `instance()` existence check can't be distinguished (both outcomes 404). No leak yet.

**Why it's load-bearing.** The moment the resolver lands (the next PR in this stack), `serveMediaCacheEntry`'s `HEAD` branch returns `200` with `content-length` = the capture's size and `etag` = the object key, i.e. the content hash. That turns an unauthenticated `HEAD` into an existence + size + content-hash oracle over a private realm's captures, and the `instance()` gate turns into an unauthenticated instance-enumeration probe (`200` vs `404` per card id). The `HEAD` auth exemption is a pre-existing, realm-wide `checkPermission` behavior — not introduced here — but this route is the first to pair it with a per-object content-hash validator and an existence signal, so it's this stack that makes it matter.

**Scope: pre-existing, now load-bearing. Non-blocking for this PR** (all-miss today), but it needs a decision before the resolver ships:
- If unauthenticated `HEAD` on a private realm is *not* intended: re-verify `read` inside `serveScreenshot` for `HEAD` (or drop `HEAD` from the dispatch and let the miss/hit path answer only authenticated `GET`s), and pin it with a `HEAD`-on-private-realm test — the current `screenshot-test.ts` only exercises `GET` for the 401/403 cases.
- If it *is* intended: scope the description's auth claim to `GET`, and add the `HEAD` case to the test so the exemption is asserted rather than incidental.

Either way the ask is: which is it, and pin the answer with a test before the hit path can return `200`.

packages/runtime-common/media-cache-serving.ts
lukemelia: [Claude Code 🤖] Fixed in `dde68a64e1` — the buffered branch is gone entirely: both interface-legal stream shapes now exit via `nodeStream` (an adapter's node Readable passes through; a bare async iterable is wrapped by a `#media-cache-stream` node/browser conditional import, since `Readable.from` can't live in browser-safe code). The comment now names the real constraint — the bridge streams only `nodeStream` verbatim — and the iterable-shape test asserts the bytes ride `nodeStream` rather than a Response body.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Fixed in `dde68a64e1` — unauthenticated HEAD is not intended: the dispatch is GET-only now (no consumer of the route sends HEAD — image loads and crawlers GET), `serveMediaCacheEntry`'s HEAD branch is removed as dead, and a private-realm test pins that HEAD never receives the route's response shape. The resolver-bearing PR stacked on this keeps the same posture.

backspace (approved):

packages/runtime-common/realm.ts
```diff
@@ -3975,6 +4004,64 @@ export class Realm {
...
+    requestContext: RequestContext,
+    instanceLocalPath: string,
+  ): Promise<ResponseWithNodeStream> {
+    if (!this.#mediaCacheAdapter) {
+      return mediaCacheMissResponse({ requestContext });
+    }
+    let instanceURL = this.paths.fileURL(
+      instanceLocalPath.replace(/\.json$/, ''),
+    );
+    let instance = await this.#realmIndexQueryEngine.instance(instanceURL);
```
habdelra: The liveness gate hydrates a whole indexed instance to read one field.

`realmIndexQueryEngine.instance()` selects `i.*` plus the prerendered-HTML join — isolated/head/atom/embedded/fitted HTML, markdown, `pristine_doc`, `search_doc`, `deps` — and the only thing used from it is `.type` on the next line. Everything else is fetched out of Postgres, deserialized, and dropped.

This is on the hot path for every screenshot GET, including 304 revalidations: a 40-image grid re-pays it for all 40 images every 60s (`MEDIA_CACHE_MAX_AGE_SECONDS`), which is exactly the traffic shape this route is built for. The prerendered HTML columns are the largest ones in `boxel_index`, so this is likely the dominant cost of serving a cached screenshot — well above the object-store read it's guarding.

A narrow existence/type probe (`SELECT type, is_deleted` against `url`/`file_alias`, in the spirit of `findMediaCacheEntry` right below) would preserve the tombstone semantics the comment above describes at a fraction of the I/O.

---

packages/runtime-common/media-cache-serving.ts
```diff
@@ -0,0 +1,146 @@
...
+      init: { status: 304, headers },
+      requestContext,
+    });
+  }
+
+  let stream = await mediaCacheAdapter.getStream(entry.objectKey);
+  if (!stream) {
+    return mediaCacheMissResponse({ requestContext });
+  }
+  await touch(dbAdapter, entry);
```
habdelra, line 121: `await touch(...)` puts a blocking, unthrottled `UPDATE` in front of every response body — here on the 200 path and again at line 102 on the 304 path.

Two problems:

1. **It contradicts `touchMediaCacheEntry`'s own contract.** That function's comment notes callers may throttle the bump; nothing here does. Every single serve writes a row. Under the grid traffic this route targets (N images, 60s revalidation) that's N writes per client per minute against one table, all serialized ahead of the bytes. The `try/catch` makes failure non-fatal, but not the latency — a slow write still delays the body.

2. **For `declared`-lane rows it buys nothing at all.** Only the `on-demand` arm of `findMediaCacheGcCandidates` reads `last_accessed_at`; declared captures age out on generation, not access. So the entire cost is pure overhead for that lane.

Given the function's own stated tolerance — "the worst case is an in-use on-demand capture looking idle to the GC one sweep early" — this doesn't need to be awaited or unconditional. Options, roughly in order of preference: skip it for `entry.lane !== 'on-demand'`; throttle on `lastAccessedAt` (the entry is already in hand, so `now - entry.lastAccessedAt < MEDIA_CACHE_MAX_AGE_SECONDS` → skip); or fire-and-forget it after the response is constructed. The first two are free and get most of the win.

---

packages/runtime-common/realm.ts
```diff
@@ -3183,6 +3195,23 @@ export class Realm {
...
+      // table: the router keys routes on the Accept header, and the browser
+      // requests this route must serve (`<img>` loads, og:image fetches)
+      // send `image/*`-shaped Accept values that match no supported mime
+      // type. Placed after checkPermission so the route inherits realm-read
+      // auth exactly like any realm resource. GET only — checkPermission
+      // exempts HEAD from auth realm-wide, so admitting HEAD here would
+      // hand unauthenticated callers an existence/size/content-hash oracle
+      // over a private realm's captures; no consumer of this route (image
+      // loads, crawlers) sends HEAD.
+      if (request.method === 'GET' && localPath.startsWith('_screenshot/')) {
```
habdelra, line 3220: This claims the entire `_screenshot/` subtree by prefix, for GET only — which makes real realm files under that path write-only.

A file at `<realm>/_screenshot/logo.png` can be created by PUT, appears in directory listings, and is indexed like any other realm file, but every GET of it is intercepted here and answered as an uncaptured 404 miss. It never becomes readable, and nothing surfaces the conflict to whoever wrote it.

The GET-only narrowing is correct for the HEAD-oracle reason the comment gives (and I confirmed `checkPermission` really does exempt HEAD realm-wide), but it's also what creates the asymmetry: the write side of the path is unguarded precisely because only GET is claimed.

Worth deciding explicitly rather than leaving implicit. Either reserve the name — reject writes under `_screenshot/` the way other reserved paths are handled, so the collision surfaces at PUT time — or fall through to normal file serving when no capture resolves, so a real file still wins. Silent-404-forever is the one outcome with no recovery path for the user.

---

packages/runtime-common/media-cache-serving.ts
```diff
@@ -0,0 +1,146 @@
...
+// The HTTP face of a MediaCache capture, shared by every route that serves
+// one. The URL is the durable reference — what rendered HTML and
+// `meta.screenshots` embed — and the content hash surfaces only as the
+// validator: a re-capture changes what the URL serves (the ETag rotates),
+// never the URL itself. So the cache policy is a short freshness window
+// with cheap revalidation: an unchanged capture revalidates as a bodyless
+// 304, a changed one arrives in the same response, and no client holds
+// stale bytes for longer than the window.
+export const MEDIA_CACHE_MAX_AGE_SECONDS = 60;
+export const MEDIA_CACHE_STALE_WHILE_REVALIDATE_SECONDS = 3600;
```
habdelra, line 31: `stale-while-revalidate=3600` contradicts the contract stated in the comment directly above it.

That comment promises "no client holds stale bytes for longer than the window," and the window is `max-age=60`. But SWR *is* a second, longer window: for the 59 minutes after freshness lapses, a cache is explicitly licensed to serve the stale bytes and revalidate in the background. So a re-captured screenshot can stay invisible for up to an hour — and on a public realm, where `mediaCacheVisibility` returns `public` and a shared CDN cache holds the entry, invisible to *everyone*, not just one client.

It also diverges from the policy this module says it mirrors: `serveLocalFile`'s image policy is `max-age=60, must-revalidate` — no SWR at all.

That's a real tension in the design, not just a wrong number: the ETag is the content hash, so revalidation is already cheap (a bodyless 304), which is the argument *against* needing SWR here. If the 3600 is deliberate — trading re-capture latency for CDN offload — the comment above should say so instead of claiming the opposite. If it isn't, dropping SWR to match `serveLocalFile` makes the code match its documentation.

Either way the constant and the paragraph above it should agree; right now a reader can't tell which one is the intent.

---

packages/realm-server/tests/realm-endpoints/screenshot-test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    test('a reader without the read grant is refused', async function (assert) {
+      let response = await request
+        .get('/_screenshot/some-card')
+        .set('Accept', 'image/png')
+        .set('Authorization', `Bearer ${createJWT(testRealm, 'not-mary')}`);
+      assert.strictEqual(response.status, 403);
+    });
+
+    test('a reader gets an uncaptured miss with private cache-control', async function (assert) {
```
habdelra, line 57: All five tests in this file are satisfied by the `if (!this.#mediaCacheAdapter)` early return, so they don't exercise the route.

`setupPermissionedRealmCached` never passes a `mediaCacheAdapter`, and no env var in the test setup makes `createMediaCacheAdapterFromEnv()` return one — so `serveScreenshot` bails on its first line in every case here. The `mediaCacheAdapter?` option added to `createRealm` in `tests/helpers/index.ts` has no caller anywhere in the diff.

The concrete consequence: you can delete everything in `serveScreenshot` after that early return — the `paths.fileURL` normalization, the instance lookup, the liveness check, the `resolveScreenshotEntry` call, the `serveMediaCacheEntry` call — and this suite still passes green. The module header says requests here "resolve to no capture, since nothing in these realms has been captured," but they don't get that far; they resolve to no *store*, which is a different branch and the one before all the logic.

That matters most for the PR's headline authorization boundary. The comment on `serveScreenshot` describes the live-parent-instance requirement as where per-instance ACLs land and as what stops captures of deleted instances from serving — and that gate currently has zero coverage. The auth tests above (401/403) pass on `checkPermission` alone and would pass identically with the whole method stubbed out.

`media-cache-serving-test.ts` does good work on the serving core with a real `FakeMediaCacheAdapter`, so the pieces are here. What's missing is wiring one through `createRealm` — the option you already added — and asserting the two cases that distinguish the gate from the early return: a live instance with a capture serves 200, and a tombstoned instance with a capture still in the ledger serves the 404 miss. The HEAD test is also weaker than it reads: `notStrictEqual(status, 200)` passes for any failure reason, including the adapter being absent.

---

habdelra (commented):
Code review of the screenshot-serving internals. Five findings, inline — roughly in severity order:

1. **`realm.ts`** — the liveness gate hydrates a full indexed instance (all prerendered HTML, `search_doc`, `pristine_doc`, deps) to read `.type`, on every screenshot GET including 304s.
2. **`media-cache-serving.ts`** — `await touch()` is an unthrottled blocking `UPDATE` in front of every response body, and is a no-op for GC purposes on the `declared` lane.
3. **`realm.ts`** — `_screenshot/` is claimed as a whole subtree, GET-only, so a real realm file under that path is writable and listable but permanently unreadable.
4. **`media-cache-serving.ts`** — `stale-while-revalidate=3600` contradicts the "no client holds stale bytes for longer than the window" contract in the comment above it, and diverges from the `serveLocalFile` policy it claims to mirror.
5. **`screenshot-test.ts`** — no test passes a `mediaCacheAdapter`, so all five endpoint tests are satisfied by the `if (!this.#mediaCacheAdapter)` early return; the per-instance liveness boundary has no coverage.

Several non-obvious things I checked and found **correct**, noting them so they don't get re-litigated:

- The GET-only dispatch is genuinely load-bearing — `checkPermission` really does exempt `HEAD` realm-wide, so admitting HEAD would hand unauthenticated callers an existence/size/content-hash oracle over a private realm.
- The extensionless `source_url` convention holds up for the widened GC join: `getInstance` matches `i.url OR i.file_alias`, tombstone rows do write the extensionless `file_alias`, and `boxel_index.file_alias` is indexed — so the new `OR` is both correct and index-usable.
- `content-length` from the ledger survives Koa's stream body setter, and both interface-legal adapter stream shapes do reach `nodeStream`.
- `realm.ts` ↔ `media-cache-serving.ts` is a value-level ESM cycle, but `ifNoneMatchMatches` is a hoisted function declaration and nothing invokes it at module-eval time, so there's no TDZ hazard today.
- Only one `new Realm(` call site exists, so the `mediaCacheAdapter` threading is complete.

---

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Fixed in `8699ceac11` — the gate is now `hasLiveInstance`, a `SELECT 1 … LIMIT 1` existence probe with `getInstance`'s exact predicate: the url/file_alias match, instance type, tombstone exclusion, and the effective-error channel (a row `getInstance` would map to `instance-error` probes as not-live, preserving the miss for errored instances). No wide columns are selected, so a serve — 304 revalidations included — no longer pays for a hydration it discards.

packages/runtime-common/media-cache-serving.ts
lukemelia: [Claude Code 🤖] Fixed in `8699ceac11` — the bump is now lane-scoped and throttled, taking both of your first two options: a serve of a non-on-demand entry skips the write entirely (only the GC's on-demand arm reads `last_accessed_at`), and an on-demand bump gates on the in-hand entry's own stamp, writing at most once per `MEDIA_CACHE_TOUCH_THROTTLE_MS` (an hour) per entry. The write stays awaited — once it fires at most hourly it no longer sits ahead of every response body, and keeping it awaited avoids fire-and-forget lifecycle edges. All three behaviors are pinned in media-cache-serving-test: bump past the window on 200 and 304, skip within it, declared-lane never writes.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Fixed in `8699ceac11` — the name is reserved, per your first option. Creation writes are refused up front at both seams: a direct PUT/PATCH/POST under `_screenshot/` gets a 400 from `internalHandle` (the guard sits right after the GET dispatch that creates the shadowing), and an `/_atomic` operation whose href resolves under the subtree gets a 422 `Reserved path` error from the atomic precheck, so the batch surface can't sneak a write past the path check. DELETE stays admitted as the recovery path for anything already stored there. Both seams are pinned with endpoint tests.

packages/runtime-common/media-cache-serving.ts
lukemelia: [Claude Code 🤖] The 3600 is deliberate, and `8699ceac11` makes the comment say so instead of contradicting it. The SWR window is what lets an image grid render from cache while revalidating in the background — without it, every grid render after a 60s freshness lapse blocks on a revalidation round-trip per image, which the cheap 304 makes fast but not free. The trade the comment now states: a re-capture can take up to max-age plus the SWR window to reach every viewer (shared caches included, on public realms), acceptable for derived, eventually-consistent captures whose URL can lag in freshness but never lies about identity. The divergence from `serveLocalFile` is intentional for the same reason — realm icons are user-edited content where must-revalidate freshness is worth the round-trips; captures are not. The PR description carries the same framing.

packages/realm-server/tests/realm-endpoints/screenshot-test.ts
lukemelia: [Claude Code 🤖] Addressed in `8699ceac11`, with one part deliberately deferred. Every realm in this suite now carries a fake-backed MediaCache store — `mediaCacheAdapter` threads through `setupPermissionedRealmCached` → `createRealm` (the previously-caller-less option) — so these requests traverse the full serving path: past the early return, through the liveness probe, into the resolver seam. The module header now says exactly that. The HEAD test pins the exact fall-through status (404 from the generic handlers) alongside the cache-control-shape check, and the write reservation is pinned at both seams.

The two hit-path assertions you asked for — live instance with a capture serves 200, tombstoned instance with a ledger row misses — are not expressible through this route yet: `resolveScreenshotEntry` unconditionally returns undefined, so no endpoint request can reach `serveMediaCacheEntry` regardless of the store. They become expressible the moment the capture resolver lands, and they are the acceptance tests for that change — noting them there so the liveness boundary gets its endpoint-level coverage the first time a hit is reachable.

PR #5843, opened 2026-08-20
GET screenshot URL DSL: canonicalized specs, capture gate, persisted-else-on-demand

lukemelia:
Stacked on #5840 (serving internals), which stacks on #5838 (the MediaCache store). This PR makes the `_screenshot/` route real: hybrid persisted-else-on-demand serving with today's capture capabilities.

## What this adds

**Capture-spec canonicalization** (`packages/runtime-common/capture-spec.ts`, shared with POST): flat unprefixed params parse into a spec whose canonical form (sorted keys, defaults elided) hashes into the ledger key — `?format=isolated`, `?format=isolated&`-noise, and the bare URL are one cache key. Today's parameter surface is `format` (`isolated`/`embedded`, matching what the capture engine supports); the engine-reserved params (`envelope`, `viewport`, `dsf`, `fullPage`, `clip`, `target`) are refused by name, never ignored — ignoring would fold different capture intents onto one cache key. All parameter errors are 400s naming the offending field, with format validation shared verbatim with `POST /_screenshot-card`.

**Cache key = (instance URL, canonical spec hash, the instance's own index generation).** An edit bumps the generation, so an edited card can never serve a stale capture; an unchanged card is a pure ledger hit with zero Chrome work — and hits serve regardless of the gate, including captures a write-holder published via POST on a gated realm.

**The `allowArbitraryScreenshots` gate**: a new BooleanField on the base `RealmConfig` card (absent ⇒ false), read from the realm's *indexed* config on every gate check — editing `realm.json` takes effect with its own index update, no restart, no cache to invalidate (pinned by a test that flips the indexed value between requests). A gated miss is a 403 whose body names the flag; the gate blocks Chrome work only, never serving.

**The miss path on an opted-in realm**:
- *Fail-fast congestion pre-check*: pending jobs in the realm's serialized `screenshot:` lane × recent average capture duration (from the jobs table, 1h lookback, reservation-claim→finish); if that already exceeds the sync budget, immediate 503 + Retry-After computed from the same math — no doomed connection held against the queue.
- *Bounded sync wait*: the capture runs through the existing `screenshot-card` job (rendered as the realm's owner, the same identity an index pass renders under), waited on for up to 25s. The budget is cost posture, not transport: the realm-server ALB idle timeout is configured at 4000s, so the ticket's stay-well-under-the-ALB question is settled with margin.
- *The job persists its own capture* (new `persist` args on `screenshot-card`; the worker writes to the MediaCache before resolving): a wait that times out answers 503 + Retry-After while the render completes and lands durably, so the client's retry is a pure ledger hit — never a second render. The route also persists from the job response itself as a fallback for workers without a store configured.

`name=` (declared screenshots) still misses — manifests are indexing-time artifacts that don't exist yet — and mixing `name=` with spec params is a 400. HEAD keeps the GET-only posture from the serving layer: even a captured spec answers HEAD from the generic handlers, pinned by test.

## Test plan

`media-cache-dsl-test` (real Postgres + real queue runner + real `screenshot-card` task over a stub prerenderer — no Chrome, no HTTP server): canonicalization/hash identities and per-field 400s; gated-realm ledger hit with zero capture work; gated miss 403 naming the flag; indexed-config flip opening the gate with no restart; capture→persist→serve with hit dedupe on the second request; generation-bump re-capture; sync-wait timeout → 503 + Retry-After with the capture landing anyway and the retry re-rendering nothing; congested-lane fail-fast 503; HEAD exclusion; missing-instance miss. The POST handler suite pins `persist: null` on its published jobs; all media-cache and screenshot suites pass (44 tests); typechecks clean across runtime-common, realm-server, base (via host), host, ai-bot, billing.

### DIFF
--- a/packages/base/realm-config.gts
+++ b/packages/base/realm-config.gts
@@ -688,10 +688,10 @@ export class RealmConfig extends CardDef {
   // for arbitrary capture specs. Full captureSpec power on a GET is an
   // unbounded spec space reachable with only realm read, so it is off
   // unless the realm turns it on; the gate blocks Chrome work only, never
-  // serving — declared screenshots and already-captured specs (including
-  // ones a write-holder produced via POST) stream regardless. Read from the
-  // realm's indexed config at request time, so editing this takes effect
-  // with the index update, no restart.
+  // serving — any capture whose canonical spec already has a MediaCache
+  // ledger entry streams regardless. Read from the realm's indexed config
+  // at request time, so editing this takes effect with the index update,
+  // no restart.
   @field allowArbitraryScreenshots = contains(BooleanField);
 
   @field cardTitle = contains(StringField, {
--- a/packages/base/realm-config.gts
+++ b/packages/base/realm-config.gts
@@ -684,6 +684,15 @@ export class RealmConfig extends CardDef {
   // automatically in that case) or when an operator otherwise needs
   // the full isolated render present in the index.
   @field includePrerenderedDefaultRealmIndex = contains(BooleanField);
+  // Opt-in for the realm's GET `_screenshot/` route to trigger NEW captures
+  // for arbitrary capture specs. Full captureSpec power on a GET is an
+  // unbounded spec space reachable with only realm read, so it is off
+  // unless the realm turns it on; the gate blocks Chrome work only, never
+  // serving — declared screenshots and already-captured specs (including
+  // ones a write-holder produced via POST) stream regardless. Read from the
+  // realm's indexed config at request time, so editing this takes effect
+  // with the index update, no restart.
+  @field allowArbitraryScreenshots = contains(BooleanField);
 
   @field cardTitle = contains(StringField, {
     computeVia: function (this: RealmConfig) {
--- a/packages/realm-server/tests/helpers/fake-media-cache-adapter.ts
+++ b/packages/realm-server/tests/helpers/fake-media-cache-adapter.ts
@@ -24,9 +24,15 @@ export class FakeMediaCacheAdapter implements MediaCacheAdapter {
     if (this.streamShape === 'readable') {
       return Readable.from(Buffer.from(bytes));
     }
+    // Yield in two chunks: a single-chunk generator can't distinguish an
+    // implementation that wraps the whole iterable from one that reads only
+    // its head, which is the bug the bare-async-iterable serving test exists
+    // to catch. Split at the midpoint (min 1) so a payload shorter than a
+    // fixed offset still crosses a real chunk boundary.
     return (async function* () {
-      yield bytes.slice(0, 3);
-      yield bytes.slice(3);
+      let mid = Math.max(1, Math.floor(bytes.length / 2));
+      yield bytes.slice(0, mid);
+      yield bytes.slice(mid);
     })();
   }
   async delete(key: string) {
--- a/packages/realm-server/tests/helpers/fake-media-cache-adapter.ts
+++ b/packages/realm-server/tests/helpers/fake-media-cache-adapter.ts
@@ -1,16 +1,21 @@
 import { Readable } from 'node:stream';
 import type { MediaCacheAdapter } from '@cardstack/runtime-common';
 
-// Minimal in-memory store: real bytes behind the MediaCacheAdapter
-// interface, with a switch between the two stream shapes the serving layer
-// handles (a node Readable, which passes through `toNodeStream`, and a bare
-// async iterable, which it wraps).
+// Shared in-memory MediaCacheAdapter for tests: real bytes behind the
+// interface, observable deletes, scriptable per-key delete failures, and a
+// switch between the two stream shapes the serving layer handles (a node
+// Readable, which passes through `toNodeStream`, and a bare async iterable,
+// which it wraps).
 export class FakeMediaCacheAdapter implements MediaCacheAdapter {
   objects = new Map<string, Uint8Array>();
+  deleted: string[] = [];
+  failDeletesFor = new Set<string>();
   streamShape: 'readable' | 'iterable' = 'readable';
 
   async put(key: string, bytes: Uint8Array, _opts: { contentType: string }) {
-    this.objects.set(key, bytes);
+    if (!this.objects.has(key)) {
+      this.objects.set(key, bytes);
+    }
   }
   async head(key: string) {
     let bytes = this.objects.get(key);
@@ -36,6 +41,10 @@ export class FakeMediaCacheAdapter implements MediaCacheAdapter {
     })();
   }
   async delete(key: string) {
+    if (this.failDeletesFor.has(key)) {
+      throw new Error(`simulated delete failure for ${key}`);
+    }
+    this.deleted.push(key);
     this.objects.delete(key);
   }
 }
--- a/packages/runtime-common/jobs/screenshot-card.ts
+++ b/packages/runtime-common/jobs/screenshot-card.ts
@@ -21,12 +21,19 @@ export const SCREENSHOT_CARD_JOB_TIMEOUT_SEC = 60;
 // render identity, and persist target — since joining hands the incoming
 // caller the twin's result verbatim. Queued and in-flight twins both join;
 // an in-flight join just registers a late waiter on the running job.
+//
+// Only the ledger-backed GET lane coalesces. Its persist target pins the
+// source generation, so a joined caller can never be handed a capture of a
+// different revision. `POST /_screenshot-card` publishes with `persist: null`
+// — a render-now request whose identity carries no freshness axis, so an
+// in-flight twin could be up to a reservation-lease old and of a pre-edit
+// card; those always insert.
 function chooseScreenshotCardCoalesceDecision(
   context: QueueCoalesceContext,
 ): QueueCoalesceDecision {
   let { incoming, candidates, inFlightCandidates } = context;
   let incomingArgs = parseScreenshotCardArgs(incoming.args);
-  if (!incomingArgs) {
+  if (!incomingArgs || !incomingArgs.persist) {
     return { type: 'insert' };
   }
   let twin = [...candidates, ...inFlightCandidates].find((candidate) => {
@@ -116,6 +123,11 @@ export interface ScreenshotQueueEstimate {
   // behind the realm's serialized screenshot lane before its own capture
   // even starts.
   estimatedWaitMs: number;
+  // True when a queued or in-flight job already carries this exact capture
+  // identity: the incoming request would coalesce onto it and cost no new
+  // Chrome work, so the caller skips the congestion pre-check rather than
+  // 503-ing a request the lane is about to satisfy for free.
+  hasTwin: boolean;
 }
 
 // The fail-fast congestion pre-check for a capture-triggering request:
@@ -128,26 +140,67 @@ export interface ScreenshotQueueEstimate {
 export async function estimateScreenshotQueueWait(
   dbAdapter: DBAdapter,
   concurrencyGroup: string,
+  // The persist identity of the capture about to be requested. When a queued
+  // or in-flight job already matches it, the request coalesces rather than
+  // rendering, so `hasTwin` lets the caller bypass the congestion gate.
+  twinOf?: {
+    sourceURL: string;
+    captureSpecHash: string;
+    sourceGeneration: number;
+  },
 ): Promise<ScreenshotQueueEstimate> {
   if (dbAdapter.kind !== 'pg') {
-    return { pending: 0, avgCaptureMs: 0, estimatedWaitMs: 0 };
+    return { pending: 0, avgCaptureMs: 0, estimatedWaitMs: 0, hasTwin: false };
   }
-  let [pendingRows, durationRows] = await Promise.all([
+  let [pendingRows, durationRows, twinRows] = await Promise.all([
     query(dbAdapter, [
-      `SELECT COUNT(*) AS pending FROM jobs WHERE status = 'unfulfilled' AND concurrency_group =`,
+      `SELECT COUNT(*) AS pending FROM jobs
+        WHERE status = 'unfulfilled'
+          AND job_type = 'screenshot-card'
+          AND concurrency_group =`,
       param(concurrencyGroup),
     ] as Expression) as Promise<{ pending: number | string }[]>,
     // Execution time is reservation-claim → job-finish (there is no
-    // started_at column); resolved captures only, bounded lookback so the
-    // estimate tracks current conditions.
+    // started_at column). Aggregate per job and measure from the last
+    // (winning) claim: a lease-lapsed retry leaves an earlier abandoned
+    // reservation whose completed_at is also set, and its row spans the
+    // whole expired lease — averaging that in would drag the estimate toward
+    // the lease duration and 503 a healthy queue. Realm-scoped to match the
+    // per-realm `pending` count, resolved captures only, bounded lookback.
     query(dbAdapter, [
-      `SELECT AVG(EXTRACT(EPOCH FROM (j.finished_at - jr.created_at)) * 1000) AS avg_ms
-       FROM jobs j
-       JOIN job_reservations jr ON jr.job_id = j.id AND jr.completed_at IS NOT NULL
-       WHERE j.job_type = 'screenshot-card'
-         AND j.status = 'resolved'
-         AND j.finished_at > NOW() - INTERVAL '${CAPTURE_DURATION_LOOKBACK_HOURS} hours'`,
+      `SELECT AVG(ms) AS avg_ms FROM (
+         SELECT EXTRACT(EPOCH FROM (j.finished_at - MAX(jr.created_at))) * 1000 AS ms
+           FROM jobs j
+           JOIN job_reservations jr ON jr.job_id = j.id AND jr.completed_at IS NOT NULL
+          WHERE j.job_type = 'screenshot-card'
+            AND j.status = 'resolved'
+            AND j.finished_at > NOW() - INTERVAL '${CAPTURE_DURATION_LOOKBACK_HOURS} hours'
+            AND j.concurrency_group =`,
+      param(concurrencyGroup),
+      `GROUP BY j.id, j.finished_at
+       ) t`,
     ] as Expression) as Promise<{ avg_ms: number | string | null }[]>,
+    twinOf
+      ? (query(dbAdapter, [
+          // A pending/in-flight job carrying the same persist target (jsonb
+          // extraction, so the compares are text — sourceGeneration binds as
+          // a string to match). Confined to the GET lane's identity fields;
+          // POST jobs have no persist and never appear here.
+          `SELECT EXISTS (
+             SELECT 1 FROM jobs
+              WHERE status = 'unfulfilled'
+                AND job_type = 'screenshot-card'
+                AND concurrency_group =`,
+          param(concurrencyGroup),
+          `AND args->'persist'->>'sourceURL' =`,
+          param(twinOf.sourceURL),
+          `AND args->'persist'->>'captureSpecHash' =`,
+          param(twinOf.captureSpecHash),
+          `AND args->'persist'->>'sourceGeneration' =`,
+          param(String(twinOf.sourceGeneration)),
+          `) AS has_twin`,
+        ] as Expression) as Promise<{ has_twin: boolean }[]>)
+      : Promise.resolve([{ has_twin: false }]),
   ]);
   let pending = Number(pendingRows[0]?.pending ?? 0);
   let avgRaw = durationRows[0]?.avg_ms;
@@ -157,6 +210,7 @@ export async function estimateScreenshotQueueWait(
     pending,
     avgCaptureMs,
     estimatedWaitMs: pending * avgCaptureMs,
+    hasTwin: Boolean(twinRows[0]?.has_twin),
   };
 }
 
--- a/packages/runtime-common/jobs/screenshot-card.ts
+++ b/packages/runtime-common/jobs/screenshot-card.ts
@@ -1,8 +1,73 @@
+import { param, query, type Expression } from '../expression.ts';
 import type { QueuePublisher } from '../queue.ts';
 import type { ScreenshotPrerenderResponse, DBAdapter } from '../index.ts';
 import type { ScreenshotCardArgs } from '../tasks/screenshot-card.ts';
 
 export const SCREENSHOT_CARD_JOB_TIMEOUT_SEC = 60;
+
+// How long a GET `_screenshot/` request holds its connection waiting for an
+// on-demand capture before answering 503 + Retry-After. A cost-posture
+// bound, not a transport one (the realm-server ALB's idle timeout is far
+// above this): sync-wait callers are humans, agents, and scripts that honor
+// Retry-After — HTML and crawlers live on the never-waiting `name=` path —
+// so the hold is a courtesy and the Retry-After is the contract.
+export const SCREENSHOT_SYNC_WAIT_BUDGET_MS = 25_000;
+
+// The stand-in capture estimate while the jobs table holds no recent
+// screenshot history to average.
+const DEFAULT_CAPTURE_ESTIMATE_MS = 10_000;
+const CAPTURE_DURATION_LOOKBACK_HOURS = 1;
+
+export interface ScreenshotQueueEstimate {
+  pending: number;
+  avgCaptureMs: number;
+  // queue depth × average capture time: what a new arrival would wait
+  // behind the realm's serialized screenshot lane before its own capture
+  // even starts.
+  estimatedWaitMs: number;
+}
+
+// The fail-fast congestion pre-check for a capture-triggering request:
+// when the realm's serialized screenshot lane is already deep enough that a
+// new arrival's wait would blow the sync budget, the caller answers 503 +
+// Retry-After immediately instead of holding a doomed connection against
+// the queue. The jobs-table SQL is Postgres-shaped; on any other adapter
+// (an in-memory test realm) the estimate is zero, which simply skips the
+// pre-check.
+export async function estimateScreenshotQueueWait(
+  dbAdapter: DBAdapter,
+  concurrencyGroup: string,
+): Promise<ScreenshotQueueEstimate> {
+  if (dbAdapter.kind !== 'pg') {
+    return { pending: 0, avgCaptureMs: 0, estimatedWaitMs: 0 };
+  }
+  let [pendingRows, durationRows] = await Promise.all([
+    query(dbAdapter, [
+      `SELECT COUNT(*) AS pending FROM jobs WHERE status = 'unfulfilled' AND concurrency_group =`,
+      param(concurrencyGroup),
+    ] as Expression) as Promise<{ pending: number | string }[]>,
+    // Execution time is reservation-claim → job-finish (there is no
+    // started_at column); resolved captures only, bounded lookback so the
+    // estimate tracks current conditions.
+    query(dbAdapter, [
+      `SELECT AVG(EXTRACT(EPOCH FROM (j.finished_at - jr.created_at)) * 1000) AS avg_ms
+       FROM jobs j
+       JOIN job_reservations jr ON jr.job_id = j.id AND jr.completed_at IS NOT NULL
+       WHERE j.job_type = 'screenshot-card'
+         AND j.status = 'resolved'
+         AND j.finished_at > NOW() - INTERVAL '${CAPTURE_DURATION_LOOKBACK_HOURS} hours'`,
+    ] as Expression) as Promise<{ avg_ms: number | string | null }[]>,
+  ]);
+  let pending = Number(pendingRows[0]?.pending ?? 0);
+  let avgRaw = durationRows[0]?.avg_ms;
+  let avgCaptureMs =
+    avgRaw == null ? DEFAULT_CAPTURE_ESTIMATE_MS : Number(avgRaw);
+  return {
+    pending,
+    avgCaptureMs,
+    estimatedWaitMs: pending * avgCaptureMs,
+  };
+}
 
 export async function enqueueScreenshotCardJob(
   args: ScreenshotCardArgs,
--- a/packages/runtime-common/jobs/screenshot-card.ts
+++ b/packages/runtime-common/jobs/screenshot-card.ts
@@ -1,9 +1,100 @@
 import { param, query, type Expression } from '../expression.ts';
-import type { QueuePublisher } from '../queue.ts';
+import {
+  registerQueueJobDefinition,
+  type QueueCoalesceContext,
+  type QueueCoalesceDecision,
+  type QueuePublisher,
+} from '../queue.ts';
 import type { ScreenshotPrerenderResponse, DBAdapter } from '../index.ts';
-import type { ScreenshotCardArgs } from '../tasks/screenshot-card.ts';
+import type {
+  ScreenshotCardArgs,
+  ScreenshotPersistArgs,
+} from '../tasks/screenshot-card.ts';
 
 export const SCREENSHOT_CARD_JOB_TIMEOUT_SEC = 60;
+
+// Concurrent requests for one capture fold onto one job: the per-realm
+// concurrency group serializes execution but does not dedupe, so without
+// this two simultaneous misses for the same spec would each run a full
+// render (the store's dedupe-on-write only saves the second upload, not the
+// Chrome work). A twin must match the whole capture identity — card, format,
+// render identity, and persist target — since joining hands the incoming
+// caller the twin's result verbatim. Queued and in-flight twins both join;
+// an in-flight join just registers a late waiter on the running job.
+function chooseScreenshotCardCoalesceDecision(
+  context: QueueCoalesceContext,
+): QueueCoalesceDecision {
+  let { incoming, candidates, inFlightCandidates } = context;
+  let incomingArgs = parseScreenshotCardArgs(incoming.args);
+  if (!incomingArgs) {
+    return { type: 'insert' };
+  }
+  let twin = [...candidates, ...inFlightCandidates].find((candidate) => {
+    if (candidate.jobType !== incoming.jobType) {
+      return false;
+    }
+    let candidateArgs = parseScreenshotCardArgs(candidate.args);
+    return (
+      candidateArgs !== undefined &&
+      candidateArgs.cardId === incomingArgs.cardId &&
+      candidateArgs.format === incomingArgs.format &&
+      candidateArgs.runAs === incomingArgs.runAs &&
+      samePersist(candidateArgs.persist, incomingArgs.persist)
+    );
+  });
+  if (!twin) {
+    return { type: 'insert' };
+  }
+  return { type: 'join', jobId: twin.id };
+}
+
+function parseScreenshotCardArgs(
+  args: unknown,
+): ScreenshotCardArgs | undefined {
+  let obj: unknown = args;
+  if (typeof args === 'string') {
+    try {
+      obj = JSON.parse(args);
+    } catch {
+      return undefined;
+    }
+  }
+  if (!obj || typeof obj !== 'object' || Array.isArray(obj)) {
+    return undefined;
+  }
+  let { cardId, format, runAs } = obj as Record<string, unknown>;
+  if (
+    typeof cardId !== 'string' ||
+    typeof format !== 'string' ||
+    typeof runAs !== 'string'
+  ) {
+    return undefined;
+  }
+  return obj as ScreenshotCardArgs;
+}
+
+// Field-by-field rather than JSON.stringify: one side round-trips through
+// jsonb, which does not preserve key order.
+function samePersist(
+  a: ScreenshotPersistArgs | null | undefined,
+  b: ScreenshotPersistArgs | null | undefined,
+): boolean {
+  if (!a || !b) {
+    return !a && !b;
+  }
+  return (
+    a.realmURL === b.realmURL &&
+    a.sourceURL === b.sourceURL &&
+    a.captureSpecHash === b.captureSpecHash &&
+    a.sourceGeneration === b.sourceGeneration &&
+    a.lane === b.lane
+  );
+}
+
+registerQueueJobDefinition({
+  jobType: 'screenshot-card',
+  coalesce: chooseScreenshotCardCoalesceDecision,
+});
 
 // How long a GET `_screenshot/` request holds its connection waiting for an
 // on-demand capture before answering 503 + Retry-After. A cost-posture
--- a/packages/runtime-common/realm.ts
+++ b/packages/runtime-common/realm.ts
@@ -4068,9 +4068,10 @@ export class Realm {
   // parent instance must be live: this gives per-instance ACLs a place to
   // land, and captures of a deleted instance stop serving the moment its
   // index tombstone appears, ahead of GC reclaiming their artifacts. The
-  // gate is a narrow existence probe (`hasLiveInstance`), never an instance
-  // hydration — it runs on every request, 304 revalidations included. Any
-  // request that resolves to no capture — instance missing or errored,
+  // gate is a narrow index read (`liveInstanceGeneration`) that returns the
+  // live instance's generation for the cache key, never a full hydration — it
+  // runs on every request, 304 revalidations included. Any request that
+  // resolves to no capture — instance missing or errored,
   // store unconfigured, addressing unresolvable — is an uncaptured miss:
   // 404 with a short max-age so an `<img>` picks up a later capture on
   // revalidation, never a synchronous wait inside an image load.
@@ -4085,7 +4086,13 @@ export class Realm {
     let instanceURL = this.paths.fileURL(
       instanceLocalPath.replace(/\.json$/, ''),
     );
-    if (!(await this.#realmIndexQueryEngine.hasLiveInstance(instanceURL))) {
+    // One narrow read is both the liveness gate and the cache key's
+    // generation: undefined means the instance is missing, deleted, or
+    // errored — an uncaptured miss — and otherwise it pins the generation an
+    // edit bumps, without hydrating the row.
+    let sourceGeneration =
+      await this.#realmIndexQueryEngine.liveInstanceGeneration(instanceURL);
+    if (sourceGeneration === undefined) {
       return mediaCacheMissResponse({ requestContext });
     }
     let searchParams = new URL(request.url).searchParams;
@@ -4121,7 +4128,7 @@ export class Realm {
       realmURL: this.url,
       sourceURL: instanceURL.href,
       captureSpecHash: await captureSpecHash(parsed.spec),
-      sourceGeneration: Number(instance.generation),
+      sourceGeneration,
     };
     let entry = await findMediaCacheEntry(this.#dbAdapter, entryKey);
     if (entry) {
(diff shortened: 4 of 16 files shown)


packages/base/realm-config.gts
```diff
@@ -684,6 +684,15 @@ export class RealmConfig extends CardDef {
   // automatically in that case) or when an operator otherwise needs
   // the full isolated render present in the index.
   @field includePrerenderedDefaultRealmIndex = contains(BooleanField);
+  // Opt-in for the realm's GET `_screenshot/` route to trigger NEW captures
+  // for arbitrary capture specs. Full captureSpec power on a GET is an
+  // unbounded spec space reachable with only realm read, so it is off
+  // unless the realm turns it on; the gate blocks Chrome work only, never
+  // serving — declared screenshots and already-captured specs (including
+  // ones a write-holder produced via POST) stream regardless. Read from the
```
lukemelia: [Claude Code 🤖] **The "via POST" hit claim contradicts the code — no POST path writes the ledger.** This new field doc (and its twin comment in `realm.ts` `serveScreenshot`: "including captures a write-holder published via POST on a gated realm") tells an operator that already-captured specs "a write-holder produced via POST" stream regardless of the gate. But the only POST surface, `handleScreenshotCard` (`POST /_screenshot-card`), enqueues its job with `persist: null`, and the `screenshot-card` task calls `putMedia` only when `persist` is non-null — so a POST capture is returned in the response body and is never recorded in `media_cache_ledger`.

I traced every `putMedia`/`persist` caller in the tree: the sole ledger writers are the GET on-demand path (`captureScreenshotOnDemand`'s `persist: { ...entryKey, lane: 'on-demand' }` and its response-body fallback `putMedia`). There is no POST→ledger writer, and declared-lane writers don't exist yet (the `name=` branch comment says the manifests "don't exist yet"). So today a POST-produced capture can never become a later GET hit.

Please reconcile one way — your call which:
- If POST should pre-warm the cache (a write-holder seeding captures for later GET hits is the useful reading), pass a real `persist` from `handleScreenshotCard` instead of `null`.
- If POST is deliberately body-only — which `screenshot-card-test`'s new comment asserts — drop the "via POST" clause from both this field doc and the `serveScreenshot` comment, so shipped base-card documentation doesn't promise a capability that isn't wired.

Class: doc/contract drift introduced by this PR (all three comments are new here). Non-blocking, but this is base-card documentation an operator reads to reason about the gate.

packages/runtime-common/realm.ts
```diff
@@ -4034,32 +4057,214 @@ export class Realm {
...
+        requestContext,
+        estimate.estimatedWaitMs,
+      );
+    }
+
+    // Render as the realm's owner — the same identity an index pass renders
+    // under. The requester already proved realm read; the capture is a
+    // realm-derived artifact, not a per-user view.
+    let owner = await this.getRealmOwnerUserId();
+    let job = await enqueueScreenshotCardJob(
```
lukemelia, line 4215: [Claude Code 🤖] **Concurrent misses for the same spec each render — `screenshot-card` has no coalesce handler.** `enqueueScreenshotCardJob` → `publish` calls `getQueueJobCoalesceHandler('screenshot-card')`, which returns undefined: nothing calls `registerQueueJobDefinition` for this job type (contrast `tasks/copy.ts`, `tasks/full-reindex.ts`, `tasks/prerender-html-reconcile.ts`, which each register a `coalesce`). So two GETs that miss on the same `entryKey` before either persists insert two distinct jobs; the `screenshot:{realmURL}` concurrency group serializes their *execution* but doesn't dedupe them, so the second still runs a full Chrome capture after the first and upserts the identical row.

The persist-then-hit design and the congestion pre-check kill the *sequential* retry case (a 503'd caller retries into a hit), but not the *simultaneous* first-request case — which is the traffic this PR newly invites (agents/scripts hitting a fresh spec). A coalesce handler keyed on `(realmURL, cardId, format, sourceGeneration)` that folds concurrent duplicates onto one canonical job — the pattern `full-reindex` already uses — would close it.

Class: follow-up (efficiency). Not a correctness bug, not blocking this PR.

lukemelia (commented):
[Claude Code 🤖] Reviewed as the next owner of the `_screenshot/` route, tracing the miss path end to end (canonicalization → gate → congestion pre-check → sync-wait → persist) and checking it against the queue, media-cache, and index-query-engine machinery it leans on. Read against the real jobs/`job_reservations` schema, `putMedia`/`findMediaCacheEntry`, and both prerenderer implementations.

**Bottom line: no blocking issues — the design holds and the tests pin the important edges (gate flip via indexed config, generation-bump re-capture, timeout-still-persists, congestion fail-fast, HEAD exclusion). Two things to reconcile, both non-blocking: a doc-vs-code contradiction about POST-published captures, and a missing job-coalesce that lets concurrent first-requests double-render.**

1. The `allowArbitraryScreenshots` field doc (and its `serveScreenshot` twin) claim POST-published captures serve as hits, but the POST handler passes `persist: null` and never writes the ledger — see the inline thread on `realm-config.gts`. Decide whether to wire POST persist or drop the clause.
2. `screenshot-card` registers no coalesce handler, so simultaneous misses for one spec double-render — see the inline thread on `realm.ts`. Follow-up.

**Checked and clean (no action):** the gate reads the *indexed* config and defaults closed when the config card is absent/unindexed; hits bypass the gate correctly (`findMediaCacheEntry` precedes the gate); the generation TOCTOU (card edited mid-capture) stores fresher-not-staler bytes under the read generation; the estimate SQL's columns/statuses all match `job-tables.ts`.

**Adjacent, out of scope:** the task's new `(result as any)?.response ?? result` unwrap changes the `POST /_screenshot-card` response body from `{response, timings, pool}` to the bare `ScreenshotPrerenderResponse` for deployments whose worker uses the *local* in-process prerenderer. Production workers use `createRemotePrerenderer` (already bare), so prod is unaffected and this aligns runtime with the declared `Task<…, ScreenshotPrerenderResponse>` type — flagging only in case a dev/test client reads `attributes.response.*`.

packages/base/realm-config.gts
lukemelia: [Claude Code 🤖] Fixed in `6290ec99f7` — the field doc and the hit-path comment now describe exactly what the code does: any capture whose canonical spec already has a ledger entry streams regardless of the gate, with no claim about which surface wrote it. (The POST persist itself is real work, delivered by the next PR stacked on this one — its handler doc carries the publish-by-POST claim there, where it is wired.)

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Fixed in `6290ec99f7` — `screenshot-card` now registers a coalesce handler: a queued or in-flight twin whose capture identity matches (cardId, format, runAs, and the full persist target, compared field-by-field since jsonb does not preserve key order) absorbs the incoming request as a join/late waiter. Pinned by a new test: two concurrent GETs for one uncaptured spec both 200 off a single render.

packages/runtime-common/jobs/screenshot-card.ts
```diff
@@ -1,9 +1,165 @@
...
+      candidateArgs.cardId === incomingArgs.cardId &&
+      candidateArgs.format === incomingArgs.format &&
+      candidateArgs.runAs === incomingArgs.runAs &&
+      samePersist(candidateArgs.persist, incomingArgs.persist)
+    );
+  });
+  if (!twin) {
+    return { type: 'insert' };
+  }
+  return { type: 'join', jobId: twin.id };
```
habdelra, line 55: [Claude Code 🤖] **The coalesce handler is registered for the job type, so it silently changes `POST /_screenshot-card` too — and the POST identity has no freshness axis.** This needs a decision in this PR, because the handler is new here.

**Mechanism.** `registerQueueJobDefinition({ jobType: 'screenshot-card', ... })` installs one handler for every publisher of that job type. `PgQueuePublisher.publish` looks it up by `jobType` alone (`getQueueJobCoalesceHandler(spec.jobType)` in `packages/postgres/pg-queue.ts`), so both `Realm.captureScreenshotOnDemand` and `handleScreenshotCard` route through it. For two POST jobs, `samePersist(null, null)` takes the `if (!a || !b) return !a && !b` branch and returns `true`, so a POST coalesces onto a POST whenever `cardId`, `format`, and `runAs` match.

That matters because the two surfaces carry different identities:

- The GET path's `persist` includes `sourceGeneration`, so an edit between the two requests makes them non-twins and the second renders fresh. That is exactly the property the field doc on `ScreenshotCardArgs.persist` is describing.
- The POST args carry no generation and no timestamp. Its identity is `(cardId, format, runAs)` — stable across edits.

**The window is the reservation lease, not milliseconds.** `findCoalesceCandidates` admits an in-flight candidate while `r.locked_until > NOW() AND r.completed_at IS NULL`, and the lease is set to `min(job.timeout, maxTimeoutSec)` at claim time (`packages/postgres/pg-queue.ts`, the `INSERT INTO job_reservations` in the claim loop). `enqueueScreenshotCardJob` publishes with `timeout: SCREENSHOT_CARD_JOB_TIMEOUT_SEC` = 60, so a POST arriving 59s into an identical render joins it and is handed that render's bytes verbatim.

**The realistic trigger is a retry, not two callers.** Both POST callers I traced await their response — `render-gate.ts`'s `captureOne` loops sequentially and `packages/host/app/tools/screenshot-card.ts` awaits — so a resolved prior job is no longer a candidate (`findCoalesceCandidates` filters `j.status='unfulfilled'`). What does hit it: a client that gives up on a slow POST and retries, or a request that lands while an abandoned job is still leased. In that case the caller who asked for "screenshot this card now" gets a capture that started up to a minute earlier, possibly of the pre-edit card. Before this PR they got a fresh render.

**Suggested resolution.** Scope the handler to the lane it was added for — the GET on-demand path, where the DoS pressure lives and where `sourceGeneration` already supplies the freshness axis:

```js
let incomingArgs = parseScreenshotCardArgs(incoming.args);
// Only the ledger-backed lane coalesces: its identity pins the source
// generation, so a joined caller can never receive a capture of a
// different revision. `persist: null` jobs (POST /_screenshot-card) are
// render-now requests with no freshness component to compare, so they
// always insert.
if (!incomingArgs || !incomingArgs.persist) {
  return { type: 'insert' };
}
```

If POST should coalesce instead, that is a fine call — but then the args need a freshness component (the instance's index generation, which the GET path already computes) and the handler comment should say that a POST caller can be handed an in-flight twin's bytes. Either way, the POST suite currently pins only `persist: null` on the published job; a test for whichever behavior you pick would keep the next change to this handler from flipping it again.

Class: regression — this PR changes an existing endpoint's semantics as a side effect. Non-blocking if you'd rather scope it in a follow-up, but the decision belongs here since the handler ships in this PR.

---

packages/runtime-common/jobs/screenshot-card.ts
```diff
@@ -1,9 +1,165 @@
...
+    // estimate tracks current conditions.
+    query(dbAdapter, [
+      `SELECT AVG(EXTRACT(EPOCH FROM (j.finished_at - jr.created_at)) * 1000) AS avg_ms
+       FROM jobs j
+       JOIN job_reservations jr ON jr.job_id = j.id AND jr.completed_at IS NOT NULL
+       WHERE j.job_type = 'screenshot-card'
+         AND j.status = 'resolved'
+         AND j.finished_at > NOW() - INTERVAL '${CAPTURE_DURATION_LOOKBACK_HOURS} hours'`,
+    ] as Expression) as Promise<{ avg_ms: number | string | null }[]>,
+  ]);
```
habdelra, line 204: [Claude Code 🤖] **The average this query produces is inflated by abandoned reservations, and it can be inflated a lot.** `avgCaptureMs` drives both the fail-fast 503 gate and every `Retry-After` this route emits, so an over-estimate turns into refused requests on a healthy queue.

**What the join actually enumerates.** `JOIN job_reservations jr ON jr.job_id = j.id AND jr.completed_at IS NOT NULL` yields one row per *closed reservation*, not one per job — and a reservation that never produced the result is closed too. `releaseJobReservation` in `packages/postgres/job-finalize.ts` sets `completed_at = NOW()` for both outcomes it handles:

```sql
UPDATE job_reservations
   SET completed_at = NOW(), completion_reason = $reason   -- 'completed' | 'interrupted'
 WHERE completed_at IS NULL AND id = $jobReservationId
```

So a screenshot job whose first worker lost its lease and was re-reserved contributes **two** rows, both measured against the same `j.finished_at`. The abandoned attempt's row is `finished_at - first_attempt_claim`, which spans the entire expired lease — up to `SCREENSHOT_CARD_JOB_TIMEOUT_SEC` (60s) plus the successful attempt — for a render that may have taken 3s. One such job an hour is enough to drag the fleet average toward the lease duration, and with `estimatedWaitMs = pending * avgCaptureMs` against a 25s budget, that flips the pre-check from "admit" to "immediate 503" at a queue depth of one.

**Fix — aggregate per job, and take the winning attempt's start:**

```sql
SELECT AVG(ms) AS avg_ms FROM (
  SELECT EXTRACT(EPOCH FROM (j.finished_at - MAX(jr.created_at))) * 1000 AS ms
    FROM jobs j
    JOIN job_reservations jr ON jr.job_id = j.id AND jr.completed_at IS NOT NULL
   WHERE j.job_type = 'screenshot-card'
     AND j.status = 'resolved'
     AND j.finished_at > NOW() - INTERVAL '1 hours'
   GROUP BY j.id, j.finished_at
) t
```

`MAX(jr.created_at)` is the last (successful) claim, and the `GROUP BY` collapses the retries so one job counts once.

**Second half — a decision, not a bug.** `pending` is scoped to `concurrency_group` (per realm) but `avg_ms` is fleet-wide across every realm. `estimatedWaitMs` therefore multiplies this realm's depth by other realms' render costs. A realm of heavy cards (20s renders) sitting behind a fleet average of 2s under-estimates its own wait by 10× and admits requests that are certain to time out; the mirror case over-rejects. If per-realm is what you want, the predicate is `AND j.concurrency_group = $concurrencyGroup` on the same query — at the cost of a thinner sample on a quiet realm, which the `DEFAULT_CAPTURE_ESTIMATE_MS` fallback already covers. Fleet-wide is a defensible choice; it just isn't what the `ScreenshotQueueEstimate.estimatedWaitMs` doc ("behind the realm's serialized screenshot lane") says, so please pick one and make the comment match.

One-line hardening while you're here: the `pending` count filters on `concurrency_group` but not `job_type`. `screenshot:{realmURL}` is exclusive to `screenshot-card` today (I grepped every `concurrencyGroup` producer), so it's correct — but adding `AND job_type = 'screenshot-card'` costs nothing and stops the count from silently absorbing a future job type that shares the lane.

Class: regression (introduced here). Non-blocking — the failure mode is refusing service, not serving wrong bytes — but it's a two-line SQL fix for something that will look like a mysterious 503 in production.

---

packages/runtime-common/realm.ts
```diff
@@ -4034,32 +4057,214 @@ export class Realm {
...
+    let estimate = await estimateScreenshotQueueWait(
+      this.#dbAdapter,
+      concurrencyGroup,
+    );
+    if (estimate.estimatedWaitMs > this.#screenshotSyncWaitMs) {
+      return this.screenshotRetryLater(
+        requestContext,
+        estimate.estimatedWaitMs,
+      );
+    }
```
habdelra, line 4209: [Claude Code 🤖] **The congestion pre-check runs before `publish`, so it can 503 the exact requests the coalesce handler exists to serve — and it does so at plausible render costs.** These two mechanisms landed in this PR (the pre-check in the first commit, coalescing in the second) and they work against each other on the miss path.

**Mechanism.** `estimateScreenshotQueueWait` counts every `status='unfulfilled'` job in `screenshot:{realmURL}` and multiplies by the average capture time. That count is blind to *which* capture each pending job is: a request whose spec is a twin of an in-flight job would join it via `chooseScreenshotCardCoalesceDecision` and be satisfied by the render already running — but the pre-check rejects it before `enqueueScreenshotCardJob` is ever called, so coalescing never gets a chance.

**Where the line falls.** `estimatedWaitMs = pending * avgCaptureMs` against `SCREENSHOT_SYNC_WAIT_BUDGET_MS` = 25s. The in-flight job itself counts toward `pending` (it is still `unfulfilled` while leased), so with `avgCaptureMs > 12.5s` the *second* concurrent request for the very same card is refused: `2 × 13s = 26s > 25s`. A 13s isolated render of a card with several linked instances is not a pathological number. Beyond that threshold, the second viewer of a card that is mid-render always gets a 503 instead of joining a render that is seconds from finishing.

**Evidence from the new suite.** `media-cache-dsl-test.ts`'s `'concurrent misses for one spec coalesce onto one capture'` has to seed a fake 200ms capture history to reach the queue at all, and its own comment says why: *"Recent capture history keeps the congestion pre-check's estimate under the budget while the first capture is in flight, so the second request reaches the queue and can coalesce instead of failing fast."* With the `DEFAULT_CAPTURE_ESTIMATE_MS` of 10s and that test's 2s budget, `1 × 10_000 > 2_000` and the second request never coalesces. The test is right to seed — but it means the coalescing path is only ever exercised under an estimate low enough to make the pre-check moot, so nothing pins the interaction.

**Way out.** Let the twin bypass the depth check. The cheapest shape is to ask the question the handler already answers before running the pre-check:

```js
// A request whose capture is already queued or rendering costs no new
// Chrome work — the coalesce handler folds it onto that job — so the
// lane's depth is not its wait. Only genuinely new captures face the
// congestion pre-check.
let estimate = await estimateScreenshotQueueWait(this.#dbAdapter, concurrencyGroup, {
  excludeTwinOf: { cardId: entryKey.sourceURL, format: spec.format, persist: { ...entryKey, lane: 'on-demand' } },
});
```

…implemented as an `args`-matching predicate in the `pending` count (`args->>'cardId'`, `args->>'format'`, `args->'persist'->>'captureSpecHash'`, `args->'persist'->>'sourceGeneration'`), returning both the depth and whether a twin exists. If a twin exists, skip the pre-check entirely and go straight to `publish`. Either way, a test that drives two concurrent misses under the *default* estimate and asserts both get 200 would pin it.

Class: regression — a new mechanism defeating another new mechanism in the same PR. Non-blocking (the failure is a 503 the client can retry into a hit), but it undercuts the fix that closed the earlier coalescing thread.

**Separately, on the 503 itself** (`screenshotRetryLater`): it answers with `body: null`. Every other refusal on this route carries a message an operator can grep — the 400s name the offending field, the 403 names the config flag. Given the framing that sync-wait callers are "agents and scripts that honor Retry-After", a one-line body (`screenshot capture is queued; retry after N seconds`) is worth the two characters. Non-blocking.

---

packages/runtime-common/realm.ts
```diff
@@ -4034,32 +4057,214 @@ export class Realm {
...
+          bytes[i] = binary.charCodeAt(i);
+        }
+        await putMedia(this.#dbAdapter, this.#mediaCacheAdapter!, {
+          ...entryKey,
+          bytes,
+          contentType: outcome.contentType ?? 'image/png',
+          lane: 'on-demand',
+        });
+        entry = await findMediaCacheEntry(this.#dbAdapter, entryKey);
+      }
```
habdelra, line 4269: [Claude Code 🤖] **This fallback branch never executes in the new suite, and it is the branch that covers a real production shape — a worker with no store configured.** Two asks here, both small.

**1. The branch is unreachable in tests.** `media-cache-dsl-test.ts` has exactly one worker factory, `startWorker()`, and it always registers the task with `mediaCacheAdapter: adapter`. So the job's own persist always succeeds and `findMediaCacheEntry` at the top of this block always returns an entry — `!entry` is never true, so lines 4201–4214 are dead in the suite. That matters because the branch exists specifically for the configuration this comment names: `TaskArgs.mediaCacheAdapter` is optional (*"a worker process without one configured still registers media-cache jobs, whose tasks then no-op"*), and in that configuration the job logs `screenshot-card asked to persist but this worker has no media cache adapter configured; skipping` and this fallback is the only thing that makes the request succeed.

One test closes it — register the task *without* `mediaCacheAdapter`, then assert the GET still answers 200, that a ledger entry now exists, and that the second request is a pure hit (`captureCalls` unchanged). That last assertion is the interesting one: it proves the web-tier fallback writes an entry the *next* request can find, not just bytes for this response.

**2. The base64 → bytes decode now has two homes.** These five lines:

```js
let binary = atob(outcome.base64);
let bytes = new Uint8Array(binary.length);
for (let i = 0; i < binary.length; i++) {
  bytes[i] = binary.charCodeAt(i);
}
```

are byte-for-byte the same as the decode in `packages/runtime-common/tasks/screenshot-card.ts`'s persist block, and both feed the same `putMedia`. Two copies of one decision drift the first time one of them learns something (a size guard, a different padding tolerance, a switch off `atob`). Worth one exported helper next to `putMedia` in `media-cache.ts` — `base64ToBytes(base64: string): Uint8Array` — with both sites calling it. Keep `atob` rather than `Buffer.from(b64, 'base64')`: `media-cache.ts` is reachable from the browser build (`media-cache-stream-browser.ts` exists for exactly that reason), and `atob` is the portable half.

Class: (1) test-coverage gap on new code; (2) drift-by-duplication introduced here. Both non-blocking.

---

packages/runtime-common/realm.ts
```diff
@@ -4034,32 +4057,214 @@ export class Realm {
...
+  ): Promise<ResponseWithNodeStream> {
+    if (!(await this.allowsArbitraryScreenshots())) {
+      return responseWithError(
+        new CardError(
+          `This realm does not allow arbitrary screenshot captures: set "allowArbitraryScreenshots" to true on the realm's config card to enable them. Captures that already exist still serve.`,
+          { status: 403 },
+        ),
+        requestContext,
+      );
+    }
```
habdelra, line 4188: [Claude Code 🤖] **The gated 403 carries no `cache-control`, so on a gated realm every single `` load pays the full origin round-trip plus two index queries.** Since the flag is absent-⇒-false, that is the default state of every realm in the fleet.

**Mechanism.** The miss path deliberately makes refusal cheap — `mediaCacheMissResponse` sets `cache-control: {private|public}, max-age=MEDIA_CACHE_MAX_AGE_SECONDS`, and `media-cache-serving-test.ts` pins it with the reasoning *"the miss is briefly cacheable, so image retries are cheap"*. This 403 goes out through `responseWithError(new CardError(...))`, which sets no `cache-control` at all. Per RFC 9111 §4.2.2, 403 is not a heuristically-cacheable status, so with no explicit freshness a browser will not reuse it: every page render re-requests.

**What each of those requests costs.** `serveScreenshot` does a `#realmIndexQueryEngine.instance()` for the parent instance, then `findMediaCacheEntry` against the ledger, then `captureScreenshotOnDemand` does a *second* `#realmIndexQueryEngine.instance()` for the config card inside `allowsArbitraryScreenshots()`. Three DB round-trips per refused image, on a realm that has opted out of the feature entirely.

**Fix.** Give the refusal the same short freshness the miss has, so a browser stops asking for a while:

```js
return createResponse({
  body: `This realm does not allow arbitrary screenshot captures: set "allowArbitraryScreenshots" to true on the realm's config card to enable them. Captures that already exist still serve.`,
  init: {
    status: 403,
    headers: {
      'cache-control': `${mediaCacheVisibility(requestContext)}, max-age=${MEDIA_CACHE_MAX_AGE_SECONDS}`,
    },
  },
  requestContext,
});
```

`mediaCacheVisibility` is what keeps a private realm's refusal out of shared caches, same as the miss uses. The short max-age is also what makes the "flip the flag, no restart" story work on the client side: with a `max-age` matching the miss, an operator who opts the realm in sees images appear within the same window an uncaptured card's image would.

Class: regression (new code path). Non-blocking, but it is the default path for every realm that never opts in.

---

packages/runtime-common/tasks/screenshot-card.ts
```diff
@@ -85,7 +106,40 @@ const screenshotCard: Task<ScreenshotCardArgs, ScreenshotPrerenderResponse> = ({
       format,
       priority: jobInfo?.priority,
     });
+    // The local (in-process) prerenderer resolves to `{response, timings,
+    // pool}` while the remote one resolves to the bare response; unwrap so
+    // the job result is one shape either way.
+    let response: ScreenshotPrerenderResponse =
+      (result as any)?.response ?? result;
```
habdelra, line 113: [Claude Code 🤖] **This unwrap describes a wiring I can't find in the tree, and the `as any` is what stops the compiler from saying so.** I'd like to know which path motivated it, because if the answer is "none", the cast is a permanent hole in the one type check that guards this boundary.

**What I traced.** `TaskArgs.prerenderer` is typed `Prerenderer` (the interface in `runtime-common/index.ts`), which declares `prerenderScreenshot?(args): Promise<ScreenshotPrerenderResponse>` — the bare shape. Every construction site of that field:

- `packages/realm-server/worker.ts:135` → `createRemotePrerenderer(prerendererUrl)`
- `packages/realm-server/main.ts:497` → `createRemotePrerenderer(prerendererUrl)`
- `packages/realm-server/tests/helpers/index.ts:1082` `getTestPrerenderer()` → `createRemotePrerenderer(url)` — which is what every `new Worker({...})` in the test helpers resolves to via `providedPrerenderer ?? await getTestPrerenderer()`
- `media-cache-dsl-test.ts:162` → a hand-written stub typed `Promise<ScreenshotPrerenderResponse>`

`createRemotePrerenderer`'s `prerenderScreenshot` returns `requestWithRetry<ScreenshotPrerenderResponse>(...)` — bare. The `Prerenderer` *class* (`packages/realm-server/prerender/prerenderer.ts:621`) is the only thing that returns `{response, timings, pool}`, and it is consumed inside the prerender server process, where `registerPrerenderRoute` destructures it before it ever crosses a task boundary:

```js
let { response, timings, pool } = raceResult.result;   // prerender-app.ts
```

The only other holder of a live class instance is `getPrerendererForTesting`, used exclusively by `prerendering-test.ts` — which never calls `prerenderScreenshot` (grep for `screenshot` in that file returns nothing). So no task has ever seen the wrapped shape.

**Why the cast is the problem, not the unwrap.** `ScreenshotPrerenderResponse.status` is required, so `{response, timings, pool}` is *not* assignable to it — a genuine wiring that produced the wrapped shape would be a compile error at the construction site, which is the check you want. `(result as any)?.response ?? result` erases the return type at the one place it is load-bearing: if `prerenderScreenshot`'s contract changes shape later, this line takes `undefined ?? result` and carries on, and `response.status === 'ready'` quietly goes false rather than failing the build.

**Ask, either way:**
- If a real path produces the wrapped shape, name it in the comment (I'd like to fix my mental model) and narrow the cast to a guard: `let response = isWrapped(result) ? result.response : result;` with a two-line `isWrapped` predicate, so the `any` doesn't cover the whole expression.
- If it was defensive, drop it and let `let response = result;` hold the interface's contract. The interface is the thing that should be corrected if it's wrong about the class.

**Adjacent, same file:** `ScreenshotCardArgs.format` (line 31) still spells the union as `'isolated' | 'embedded'` while `capture-spec.ts` now owns `CAPTURE_FORMATS`/`CaptureFormat` as the single home. TypeScript does catch a divergence today — adding a third format to `CAPTURE_FORMATS` breaks the build where `isCaptureFormat`-narrowed `format` is assigned in `handle-screenshot-card.ts` — but the error will point at the handler rather than at the stale union. Typing this field `CaptureFormat` makes the two agree by construction.

Class: (1) a question plus a scoping ask on new code; (2) drift-by-duplication. Both non-blocking.

---

packages/realm-server/tests/helpers/fake-media-cache-adapter.ts
```diff
@@ -0,0 +1,43 @@
...
+    let bytes = this.objects.get(key);
+    if (!bytes) {
+      return undefined;
+    }
+    if (this.streamShape === 'readable') {
+      return Readable.from(Buffer.from(bytes));
+    }
+    return (async function* () {
+      yield bytes;
+    })();
```
habdelra, line 41: [Claude Code 🤖] **Merging the two fakes silently weakened the one test that exists to exercise the async-iterable path — it now yields a single chunk where it used to yield two.** The extraction is right; this one detail should come with it.

**What changed.** The copy this file replaced in `media-cache-serving-test.ts` deliberately split the payload:

```js
return (async function* () {
  yield bytes.slice(0, 3);
  yield bytes.slice(3);
})();
```

This version yields `bytes` whole.

**Why the split was load-bearing.** `media-cache-serving-test.ts:146` (`'a bare async-iterable stream still exits via nodeStream'`) is the only test that sets `streamShape = 'iterable'`, and its assertion is `deepEqual([...await nodeStreamToBuffer(response.nodeStream!)], [...BYTES])`. The implementation under test is `toNodeStream` in `media-cache-stream-node.ts`, which for a non-`Readable` input returns `Readable.from(stream)`. With a single-chunk generator, a hypothetical implementation that read only the first chunk — `Readable.from([(await it.next()).value])` — passes that assertion identically. With two chunks it fails. The multi-chunk shape is what makes the test able to distinguish "wraps the iterable" from "reads the head of the iterable", which is precisely the bug class an adapter returning a chunked S3 body would hit in production.

**Fix** — restore the split, and let the caller choose a shape that survives a payload shorter than the split point:

```js
    return (async function* () {
      let mid = Math.max(1, Math.floor(bytes.length / 2));
      yield bytes.slice(0, mid);
      yield bytes.slice(mid);
    })();
```

That keeps the new suite working (its `PNG_BYTES` payload is longer than one byte) and puts a real chunk boundary back in front of `toNodeStream`.

**One note, not an ask:** `put` also changed semantics in the merge — the gc-test copy's write-once `if (!this.objects.has(key))` won over the serving-test copy's unconditional `set`. I checked every `putMedia`/`adapter.put` call in the three suites and none re-puts different bytes under one key, and `putMedia` is content-addressed anyway, so nothing is affected today. Worth a one-line comment on `put` saying the write-once is intentional (it mirrors the real store's dedupe-on-write) so a future test author doesn't spend an afternoon on it.

Class: test-coverage regression introduced by this PR's refactor. Non-blocking, and a three-line fix.

---

packages/runtime-common/realm.ts
```diff
@@ -4034,32 +4057,214 @@ export class Realm {
...
+  private async allowsArbitraryScreenshots(): Promise<boolean> {
+    let realmConfigCardURL = new URL(
+      this.paths.fileURL('realm.json').href.replace(/\.json$/, ''),
+    );
+    let entry = await this.#realmIndexQueryEngine.instance(realmConfigCardURL);
+    if (entry?.type !== 'instance') {
+      return false;
+    }
+    return entry.instance.attributes?.allowArbitraryScreenshots === true;
   }
```
habdelra, line 4327: [Claude Code 🤖] **Two small things, one factual and one structural.**

**The comment names a file that doesn't exist.** It says *"a `.realm.json` edit takes effect with its own index update"* — leading dot. The file this method actually reads is `realm.json` (no dot): `this.paths.fileURL('realm.json')`, matching `serveScreenshot`'s sibling readers at `realmConfigHostRoutingMap` and the realm-info overlay, both of which say `/realm.json`. An operator following this comment goes looking for a dotfile. (The PR description carries the same typo, worth fixing there too since it is what people will read first.)

**The card-URL derivation is now in three places.** This exact expression:

```js
new URL(this.paths.fileURL('realm.json').href.replace(/\.json$/, ''))
```

appears here, in the host-routing-map reader, and in the realm-info overlay reader. Three copies of one decision about where the config card lives — the kind that drifts the first time one of them learns something (a rename, an `.override.json` variant, a cache). A private getter is the whole fix:

```js
private get realmConfigCardURL(): URL {
  // The config card's id form: `realm.json` on disk, extensionless as an
  // instance id, which is what the index and every reader speak.
  return new URL(this.paths.fileURL('realm.json').href.replace(/\.json$/, ''));
}
```

…with all three call sites reading it. That also gives the "where does the gate live" question one answer to grep for.

Class: (1) doc error introduced here — please fix in this PR; (2) pre-existing duplication that this PR makes a third copy of, so the consolidation is fair to ask here but fine as a follow-up if you'd rather keep the diff tight.

For what it's worth, the read itself is right and the `entry?.type !== 'instance'` / `=== true` shape gives you the fail-closed behavior the comment claims (absent, unindexed, non-boolean truthy all read as gated), and `'flipping the indexed config opens the gate with no restart'` pins the no-cache property properly.

---

habdelra (commented):
[Claude Code 🤖] Reviewed with the miss path as the lens: everything that decides *whether Chrome runs* and *what a caller gets when it doesn't* — the gate, the congestion pre-check, the sync-wait budget, and the coalesce handler. That is the right lens because the hit path is a pure ledger read the serving layer already owns, while the miss path is four new mechanisms that have to compose correctly under concurrency, and two of them landed in separate commits here without a test that exercises them together.

**Bottom line: no blocking issues. The cache-key design is sound — pinning the instance's own index generation makes a stale capture structurally impossible, and the canonicalize-then-hash step gives it one key per intent. Two findings I'd want settled in this PR (the coalesce handler's scope, and the capture-duration SQL); the rest are cheap and could ride a follow-up.**

**Dispositions on the two open threads.** Both fixes landed and both threads are closed as far as I can verify:

- The base-card field doc and `serveScreenshot`'s hit comment now say only that a capture with a ledger entry streams regardless of the gate, with no claim about which surface wrote it. That matches the code: the sole ledger writers remain the GET on-demand path's `persist` and its response-body fallback.
- `screenshot-card` now registers a coalesce handler and `'concurrent misses for one spec coalesce onto one capture'` pins it. The mechanism is right — see recommendations 1 and 3 for two ways it doesn't yet reach as far as the comment implies.

**Recommendations, most consequential first:**

1. **Scope the coalesce handler to the ledger-backed lane.** Registering it by `jobType` applies it to `POST /_screenshot-card` too, where the identity is `(cardId, format, runAs)` with no freshness component — so a POST can join an in-flight twin up to the 60s reservation lease old and be handed a pre-edit render. A `!incomingArgs.persist → insert` guard keeps POST's render-now contract. Detail and the trigger I could actually reproduce on paper: thread on `chooseScreenshotCardCoalesceDecision`.
2. **Fix the capture-duration average.** `JOIN job_reservations … completed_at IS NOT NULL` counts abandoned reservations as attempts, so one lease-lapsed retry contributes a row spanning the whole expired lease and drags the average toward 60s — which then drives the 503 gate and every `Retry-After`. Aggregate per job with `MAX(jr.created_at)`; decide whether the average should be realm-scoped or fleet-wide and make the doc match. Thread on the estimate SQL.
3. **Let a twin bypass the congestion pre-check.** The pre-check runs before `publish`, so it can refuse the very requests coalescing exists to absorb — at `avgCaptureMs > 12.5s` the second concurrent request for one card is 503'd against the 25s budget. The new coalescing test has to seed a 200ms history to get past it, which is the tell. Thread on the pre-check.
4. **Test the web-tier persist fallback.** `startWorker()` always configures a `mediaCacheAdapter`, so the `!entry` branch in `captureScreenshotOnDemand` — the branch that exists for a worker without a store — never runs in the suite. One test registering the task without the adapter closes it. Same thread flags the base64→bytes decode now living in two files.
5. **Give the gated 403 a `cache-control`.** The miss path deliberately sets a short max-age so image retries are cheap; the 403 sets none, and 403 isn't heuristically cacheable — so on a gated realm (the default for every realm) every `` load pays three DB round-trips. Thread on the gate check.
6. **Restore the two-chunk yield in the extracted fake's iterable shape.** The merged `FakeMediaCacheAdapter` yields one chunk where the serving-test copy yielded two, which is what made `'a bare async-iterable stream still exits via nodeStream'` able to catch a read-only-the-head bug. Thread on the fake.
7. **Two nits on the gate reader**: the comment says `.realm.json` but the file is `realm.json`, and the config-card URL derivation is now in three places in this file. Thread on `allowsArbitraryScreenshots`.

**Verified clean, so you don't have to re-derive it:**

- *Owner-identity render is not a new exposure.* `captureScreenshotOnDemand` renders as the realm owner for a requester holding only read, and the owner's cross-realm JWTs go into the prerender auth — so a card linking to a private cross-realm instance can bake that content into a PNG served to any realm reader. That is the same identity model the index already uses (`indexer.ts` and `prerender-html.ts` both render as `realmUsername`, and the realm's own node-side fetcher sets `X-Boxel-Assume-User: owner` for every request), and prerendered HTML reaches realm readers the same way. The code comment states this correctly; no new hole.
- *The spec space is bounded, so the gate's blast radius is small.* With `format` the only axis and `sourceGeneration` in the key, a card has at most two reachable captures per revision — so an anonymous caller on an opted-in public realm can force at most one render per card per edit, serialized behind the per-realm lane and fail-fast. That is "warm the cache slowly", which is what opting in means.
- *No unhandled rejection on the abandoned job.* `Promise.race` attaches handlers to `job.done`, and `Job`'s constructor already installs a `.catch(() => {})` for exactly this case, so a job that rejects after the sync wait gives up is swallowed rather than aborting the process.
- *The coalesce registration reaches the publisher.* The side-effect import in `pg-queue.ts` is what the docs there promise, and `realm.ts` imports the module as a value anyway, so the handler is registered before any `publish()` on both the web tier and the worker.

**Adjacent, out of scope — flagged for whoever touches it next, not asked of this PR:** repeated `name=` params aren't rejected the way repeated `format=` is (`new Set(searchParams.keys())` collapses them before the mixing check sees a duplicate), so `?name=a&name=b` reads as one name; harmless while nothing consumes `name`, worth a `getAll('name').length > 1` guard when the declared-screenshot resolver lands. And `SCREENSHOT_SYNC_WAIT_BUDGET_MS` has no env override — the `Options` field exists for tests only — so an operator facing a slow prerender fleet can't lower the hold without a deploy.

---

packages/runtime-common/realm.ts
habdelra: [Claude Code 🤖] Confirmed the fix landed and works as described: `chooseScreenshotCardCoalesceDecision` compares the full capture identity field-by-field, `findCoalesceCandidates` supplies both queued and leased twins, and the new test drives two concurrent GETs for one uncaptured spec to a single render. The field-by-field `samePersist` is the right call — jsonb round-trips don't preserve key order, so `JSON.stringify` comparison would have missed twins nondeterministically.

Two things the fix reaches past what it intended, both raised as their own threads on the new code rather than continued here:

1. Registration is by `jobType`, so `POST /_screenshot-card` coalesces too — and its identity carries no generation, so a POST can join an in-flight twin up to the 60s reservation lease old. See the thread on `chooseScreenshotCardCoalesceDecision` for the guard.
2. The congestion pre-check runs before `publish`, so it can 503 a request that would have coalesced. That is why the new test has to seed a 200ms capture history to reach the queue at all. See the thread on the pre-check in `captureScreenshotOnDemand`.

Neither undoes this fix; both are about how far it reaches. Closing this thread.

---

packages/runtime-common/tasks/screenshot-card.ts
lukemelia: [Claude Code 🤖] Verified the trace, and it holds. The worker wires `createRemotePrerenderer` (`worker.ts:135` → `216`), whose `prerenderScreenshot` returns a bare `ScreenshotPrerenderResponse` (`requestWithRetry<ScreenshotPrerenderResponse>`). The only wrapped-shape producer is the `Prerenderer` class (`prerenderer.ts:621`), and it's consumed inside the prerender server via `registerPrerenderRoute`'s destructure — it's never handed to a task as `TaskArgs.prerenderer`. So `result` here is always the bare shape, `(result as any)?.response` is always `undefined`, and the expression falls through to `?? result` every time. The comment describing a local in-process prerenderer that reaches this boundary is aspirational — no construction site wires one.

So you're right that the `as any` hides dead code. Resolution I'd take: drop the unwrap to `let response = result;` and correct the comment, since the bare shape is the only one that reaches here. If we'd rather keep a hedge for a future in-process wiring, a two-line `isWrapped()` guard replaces the `as any` without erasing the return type — either is fine, but not the blanket cast. And I'll take the adjacent `ScreenshotCardArgs.format: CaptureFormat` so the union has one home in `capture-spec.ts`.

Holding this one out of the review-response commit (`aa7743fc`) since it's a call between drop-vs-guard rather than a mechanical fix — happy to fold whichever you prefer into a follow-up on this branch.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Fixed in `aa7743fc`. `estimateScreenshotQueueWait` now takes the capture's persist identity and returns `hasTwin` — an `EXISTS` over pending/in-flight jobs matching `args->'persist'` on sourceURL + captureSpecHash + sourceGeneration. The miss path skips the congestion gate when a twin is already queued or rendering, so the second viewer of a mid-render card coalesces onto it instead of being 503'd against a wait it would never incur. Separately, the sync-wait 503 (`screenshotRetryLater`) now carries a message body, matching the named 400s and 403.

packages/runtime-common/jobs/screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in `aa7743fc`. `chooseScreenshotCardCoalesceDecision` now returns `insert` for any job with no `persist` target, so `POST /_screenshot-card` (`persist: null`) always renders fresh — only the ledger-backed GET lane, whose persist pins the source generation, coalesces. The function header states the scoping.

packages/runtime-common/jobs/screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in `aa7743fc`. The average aggregates per job and measures from the winning claim (`MAX(jr.created_at)` with `GROUP BY j.id`), so a lease-lapsed retry's abandoned reservation no longer contributes a lease-length row. Scoped per-realm (`AND concurrency_group = …`) to match the `pending` count and the `estimatedWaitMs` doc, and added the `job_type = 'screenshot-card'` guard on the pending count.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Typo fixed in `aa7743fc` — the comment now reads `realm.json` (and the PR description too). Left the three-copy `realmConfigCardURL` derivation as a follow-up to keep this commit scoped to the review's correctness items; flagging so the consolidation isn't lost.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Fixed in `aa7743fc`. The gated 403 now carries `cache-control: {public|private}, max-age=MEDIA_CACHE_MAX_AGE_SECONDS` via `mediaCacheVisibility` — the same short window the miss uses — so a gated realm's `` loads stop re-hitting the origin every render, and opting in surfaces images within that window. Body still names the flag (the gated-403 test's `.includes('allowArbitraryScreenshots')` assertion still holds).

packages/realm-server/tests/helpers/fake-media-cache-adapter.ts
lukemelia: [Claude Code 🤖] Fixed in `aa7743fc`. The extracted fake yields two chunks again (split at `floor(len/2)`, min 1), so the bare-async-iterable serving test can still distinguish "wraps the iterable" from "reads the head". Added the one-line note that `put`'s write-once is intentional — it mirrors the store's dedupe-on-write.

packages/runtime-common/realm.ts
lukemelia: [Claude Code 🤖] Taking both as follow-ups rather than in this commit: the web-tier persist-fallback test (register the task without a `mediaCacheAdapter`, then assert the GET still lands a findable ledger entry the next request hits) and the shared `base64ToBytes` helper for the decode now duplicated here and in the task. Flagging so neither is lost — the coverage gap is real; I'm just keeping `aa7743fc` to the correctness items.

habdelra (approved):

PR #5844, opened 2026-08-20
POST /_screenshot-card persists captures to MediaCache and returns served URLs

lukemelia:
Stacked on #5843 (GET DSL) → #5840 (serving) → #5838 (store). This PR completes M1's write side: the POST endpoint stops being ephemeral-only.

## What this adds

**Persistence under the DSL's canonical identity.** A capture of an indexed card enqueues its `screenshot-card` job with a real `persist` target — instance URL (extensionless), canonical spec hash, and the instance's *current* index generation — the exact key the GET `_screenshot/` route resolves. A POST-published capture therefore serves on the durable URL immediately, including on realms whose `allowArbitraryScreenshots` gate is closed: publish-by-POST, now actually wired (the base-card gate doc regains that claim here, where it's true). The endpoint itself stays deliberately ungated — an authenticated surface under realm-read trust.

**Response shape, byte-compatible.** `data.attributes` keeps the raw `status`/`base64`/`width`/`height`/`contentType` fields exactly as before, and gains `captures: [{name, url, width, height, deviceScaleFactor, base64?}]` when the capture persisted — `url` is the durable `_screenshot/` form with the spec's canonical query (`?format=embedded` appears; the default spec elides). `name`/`deviceScaleFactor` are null for ad-hoc captures (they populate with declared batches and the M2 engine). `includeBase64: false` omits the bytes everywhere, ahead of the default flipping once callers migrate to URLs. Cards the index doesn't know (or a server without a store) still capture with the legacy shape — no `captures`, no persist.

**Store-first reads and bounded waits, mirroring the GET DSL.** A request whose canonical identity already has a ledger entry answers with zero render work — the ledger now records the capture's `width`/`height` (additive migration) so the mirror never decodes image bytes, and the object streams back as `base64` only when requested. An edited card misses its stale entry and re-captures at the current generation. The HTTP wait is bounded (same 25s budget, injectable for tests); over budget answers `503` + `Retry-After` while the job persists its capture regardless, so the client's retry is a pure ledger hit — never a second render. A new `cardId must be within realmURL` 400 protects the persist identity.

**Threading**: the `MediaCacheAdapter` now reaches the realm-server's route layer (`RealmServer` → `CreateRoutesArgs`), from the same env-configured instance the realms and workers share.

## Test plan

Seven new handler tests against real Postgres with the in-memory store fake and a scriptable job queue: the persist identity published for the DSL to hit (hash-equal to the GET's key), the served-URL shapes for default and non-default specs, `includeBase64: false`, ledger hits with zero jobs enqueued, generation-bump re-capture, 503 + Retry-After on a stalled job, and the unindexed-card legacy shape. The pre-existing byte-compat test passes untouched, and the full media-cache/DSL/serving/GC sweep is green (66 tests). Typechecks clean across runtime-common, realm-server, host.

### DIFF
--- a/packages/realm-server/handlers/handle-screenshot-card.ts
+++ b/packages/realm-server/handlers/handle-screenshot-card.ts
@@ -3,15 +3,19 @@ import type Koa from 'koa';
 import {
   captureSpecHash,
   ensureTrailingSlash,
+  fetchRealmPermissions,
   findLiveInstanceGeneration,
   findMediaCacheEntry,
   isCaptureFormat,
   screenshotURLFor,
+  touchMediaCacheEntryOnHit,
   type CaptureSpec,
+  type DBAdapter,
   type MediaCacheEntry,
   type MediaCacheEntryKey,
   type ScreenshotPrerenderResponse,
 } from '@cardstack/runtime-common';
+import RealmPermissionChecker from '@cardstack/runtime-common/realm-permission-checker';
 import {
   enqueueScreenshotCardJob,
   estimateScreenshotQueueWait,
@@ -52,8 +56,11 @@ interface CaptureResult {
  * `_screenshot/` DSL resolves, so a capture published here serves on that
  * route immediately, even on realms whose `allowArbitraryScreenshots` gate
  * is closed (the gate blocks new GET-triggered captures, never serving).
- * This endpoint itself is deliberately ungated: it is an authenticated
- * surface with full captureSpec power under realm-read trust.
+ * This endpoint skips that gate deliberately: it is an authenticated
+ * surface with full captureSpec power under realm-read trust. Realm read is
+ * enforced in two places: the ledger fast path (and the generation probe
+ * feeding it) checks it here, since it answers before any job exists; the
+ * render path relies on the worker task's permission check.
  *
  * A request whose canonical identity already has a ledger entry answers
  * from the store with zero render work — which is also what lets a
@@ -89,6 +96,7 @@ interface CaptureResult {
 export default function handleScreenshotCard({
   dbAdapter,
   queue,
+  matrixClient,
   mediaCacheAdapter,
   screenshotSyncWaitMs = SCREENSHOT_SYNC_WAIT_BUDGET_MS,
 }: CreateRoutesArgs): (ctxt: Koa.Context, next: Koa.Next) => Promise<void> {
@@ -128,14 +136,30 @@ export default function handleScreenshotCard({
       return sendResponseForBadRequest(ctxt, 'includeBase64 must be a boolean');
     }
     let withBase64 = includeBase64 !== false;
-    let normalizedRealmURL = ensureTrailingSlash(realmURL);
+    // Both URLs go through `new URL` resolution (dot segments,
+    // percent-encoding, default ports) before anything derives from them:
+    // the containment check below must mean real containment (a dotted
+    // `cardId` prefixed with the realm URL escapes a plain string-prefix
+    // test), and the persist identity must key an instance exactly the way
+    // the GET route's `paths.fileURL` derivation does.
+    let normalizedRealmURL: string;
+    let normalizedCardId: string;
+    try {
+      normalizedRealmURL = ensureTrailingSlash(new URL(realmURL).href);
+      normalizedCardId = new URL(cardId).href;
+    } catch {
+      return sendResponseForBadRequest(
+        ctxt,
+        'realmURL and cardId must be valid absolute URLs',
+      );
+    }
     // The persist identity (and the served URL) hang off the instance's
     // location within its realm.
-    if (!cardId.startsWith(normalizedRealmURL)) {
+    if (!normalizedCardId.startsWith(normalizedRealmURL)) {
       return sendResponseForBadRequest(ctxt, 'cardId must be within realmURL');
     }
     let spec: CaptureSpec = { format };
-    let sourceURL = cardId.replace(/\.json$/, '');
+    let sourceURL = normalizedCardId.replace(/\.json$/, '');
     let instanceLocalPath = sourceURL.slice(normalizedRealmURL.length);
 
     let token = ctxt.state.token as RealmServerTokenClaim;
@@ -154,17 +178,35 @@ export default function handleScreenshotCard({
       // persisted and the response carries no served URL.
       let entryKey: MediaCacheEntryKey | undefined;
       if (mediaCacheAdapter) {
-        let generation = await findLiveInstanceGeneration(dbAdapter, {
-          realmURL: normalizedRealmURL,
-          instanceURL: sourceURL,
-        });
-        if (generation !== undefined) {
-          entryKey = {
+        // The ledger fast path and the generation probe feeding it answer
+        // from the store before any job exists, so the worker task's
+        // permission check never covers them — realm read is enforced here
+        // instead (ahead of the probe, which alone would leak instance
+        // existence on a private realm). Read is checked the way the realm
+        // itself checks it — exact rows plus the `*` and `users` grants. A
+        // caller without read goes straight to the render path, whose
+        // permissions the worker enforces, and never persists.
+        let permissions = await fetchRealmPermissions(
+          dbAdapter,
+          new URL(normalizedRealmURL),
+        );
+        let mayRead = await new RealmPermissionChecker(
+          permissions,
+          matrixClient,
+        ).can(userId, 'read');
+        if (mayRead) {
+          let generation = await findLiveInstanceGeneration(dbAdapter, {
             realmURL: normalizedRealmURL,
-            sourceURL,
-            captureSpecHash: await captureSpecHash(spec),
-            sourceGeneration: generation,
-          };
+            instanceURL: sourceURL,
+          });
+          if (generation !== undefined) {
+            entryKey = {
+              realmURL: normalizedRealmURL,
+              sourceURL,
+              captureSpecHash: await captureSpecHash(spec),
+              sourceGeneration: generation,
+            };
+          }
         }
       }
 
@@ -178,6 +220,7 @@ export default function handleScreenshotCard({
             instanceLocalPath,
             spec,
             mediaCacheAdapter: mediaCacheAdapter!,
+            dbAdapter,
           });
           if (response) {
             return await setContextResponse(ctxt, response);
@@ -187,12 +230,15 @@ export default function handleScreenshotCard({
         }
       }
 
+      // The canonical realm URL keys the per-realm serialization lane (the
+      // job's default concurrency group) so this surface and the GET lane —
+      // which keys off the realm's own URL — share one lane per realm.
       let job = await enqueueScreenshotCardJob(
         {
-          realmURL,
+          realmURL: normalizedRealmURL,
           realmUsername: userId,
           runAs: userId,
-          cardId,
+          cardId: normalizedCardId,
           format,
           persist: entryKey ? { ...entryKey, lane: 'on-demand' } : null,
         },
@@ -226,7 +272,7 @@ export default function handleScreenshotCard({
         // with no second render. The retry hint is one average capture.
         let estimate = await estimateScreenshotQueueWait(
           dbAdapter,
-          `screenshot:${realmURL}`,
+          `screenshot:${normalizedRealmURL}`,
         );
         return await setContextResponse(
           ctxt,
@@ -325,6 +371,7 @@ async function respondFromLedger({
   instanceLocalPath,
   spec,
   mediaCacheAdapter,
+  dbAdapter,
 }: {
   entry: MediaCacheEntry;
   withBase64: boolean;
@@ -332,6 +379,7 @@ async function respondFromLedger({
   instanceLocalPath: string;
   spec: CaptureSpec;
   mediaCacheAdapter: NonNullable<CreateRoutesArgs['mediaCacheAdapter']>;
+  dbAdapter: DBAdapter;
 }): Promise<Response | undefined> {
   let base64: string | undefined;
   if (withBase64) {
@@ -347,6 +395,11 @@ async function respondFromLedger({
   } else if (!(await mediaCacheAdapter.head(entry.objectKey))) {
     return undefined;
   }
+  // A hit consumed through this endpoint is a use like any GET serve: the
+  // same guarded bump (on-demand lane only, hourly-throttled) keeps a
+  // capture refreshed exclusively via POST from aging out of the GC's
+  // idle TTL while in active use.
+  await touchMediaCacheEntryOnHit(dbAdapter, entry);
   let attributes: Record<string, unknown> = {
     status: 'ready',
     width: entry.width,
--- a/packages/realm-server/handlers/handle-screenshot-card.ts
+++ b/packages/realm-server/handlers/handle-screenshot-card.ts
@@ -1,7 +1,22 @@
 import type Koa from 'koa';
 
-import { isCaptureFormat } from '@cardstack/runtime-common';
-import { enqueueScreenshotCardJob } from '@cardstack/runtime-common/jobs/screenshot-card';
+import {
+  captureSpecHash,
+  ensureTrailingSlash,
+  findLiveInstanceGeneration,
+  findMediaCacheEntry,
+  isCaptureFormat,
+  screenshotURLFor,
+  type CaptureSpec,
+  type MediaCacheEntry,
+  type MediaCacheEntryKey,
+  type ScreenshotPrerenderResponse,
+} from '@cardstack/runtime-common';
+import {
+  enqueueScreenshotCardJob,
+  estimateScreenshotQueueWait,
+  SCREENSHOT_SYNC_WAIT_BUDGET_MS,
+} from '@cardstack/runtime-common/jobs/screenshot-card';
 import { userInitiatedPriority } from '@cardstack/runtime-common/queue';
 
 import {
@@ -13,13 +28,46 @@ import {
 import type { CreateRoutesArgs } from '../routes.ts';
 import type { RealmServerTokenClaim } from '../utils/jwt.ts';
 
+// One entry of the response's `captures` array: the durable served URL is
+// the reference callers should embed (a re-capture rotates its bytes, never
+// the URL); `base64` rides along by default until callers migrate to URLs
+// (`includeBase64: false` opts out). `name` and `deviceScaleFactor` are
+// null for ad-hoc captures — they populate for declared-screenshot batches
+// and once the capture engine reports a scale factor.
+interface CaptureResult {
+  name: string | null;
+  url: string;
+  width: number | null;
+  height: number | null;
+  deviceScaleFactor: number | null;
+  base64?: string;
+}
+
 /**
  * Handler for `POST /_screenshot-card`.
  *
- * Enqueues a screenshot-card job via the queue system, waits for the result,
- * and returns it. The job runs in a worker which calls
- * `prerenderer.prerenderScreenshot(...)` after fetching the caller's
- * realm permissions.
+ * Captures one card and persists the capture to the MediaCache under its
+ * canonical identity (instance URL × canonical capture spec × the
+ * instance's current index generation) — the same key the GET
+ * `_screenshot/` DSL resolves, so a capture published here serves on that
+ * route immediately, even on realms whose `allowArbitraryScreenshots` gate
+ * is closed (the gate blocks new GET-triggered captures, never serving).
+ * This endpoint itself is deliberately ungated: it is an authenticated
+ * surface with full captureSpec power under realm-read trust.
+ *
+ * A request whose canonical identity already has a ledger entry answers
+ * from the store with zero render work — which is also what lets a
+ * timed-out request resume: the job persists its capture even after the
+ * HTTP wait gives up (503 + Retry-After, bounded well under the ALB idle
+ * timeout), so the retry is a pure ledger hit.
+ *
+ * Response (201): `data.attributes` carries the raw capture fields —
+ * `status`, `base64`, `width`, `height`, `contentType` — for
+ * byte-compatibility with current-shape callers, plus `captures:
+ * [{name, url, width, height, deviceScaleFactor, base64?}]` when the
+ * capture persisted. A card the index doesn't know (or a server without a
+ * MediaCache store) still captures and returns the raw fields, just
+ * without `captures`.
  *
  * Request body (JSON:API):
  * ```json
@@ -29,7 +77,8 @@ import type { RealmServerTokenClaim } from '../utils/jwt.ts';
  *     "attributes": {
  *       "realmURL": "https://realm.example/user/workspace/",
  *       "cardId": "https://realm.example/user/workspace/Person/fadhlan",
- *       "format": "isolated"
+ *       "format": "isolated",
+ *       "includeBase64": true
  *     }
  *   }
  * }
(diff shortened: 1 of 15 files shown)


packages/realm-server/handlers/handle-screenshot-card.ts
```diff
@@ -84,29 +148,127 @@ export default function handleScreenshotCard({
...
+            realmURL: normalizedRealmURL,
+            sourceURL,
+            captureSpecHash: await captureSpecHash(spec),
+            sourceGeneration: generation,
+          };
+        }
+      }
+
+      if (entryKey) {
+        let entry = await findMediaCacheEntry(dbAdapter, entryKey);
```
lukemelia, line 214: [Claude Code 🤖] **The ledger fast path returns capture bytes with no realm permission check — any authenticated realm-server user can read any persisted capture, private realms included.** This is the one finding I'd block on.

**The mechanism.** The route is mounted behind `jwtMiddleware` alone (`packages/realm-server/routes.ts`), which authenticates *any* realm-server session — it proves who the caller is, not what they may read. The endpoint's only permission enforcement lives in the worker: `tasks/screenshot-card.ts` calls `fetchRealmPermissions` and refuses when `runAs` has no permissions in the target realm — so the render path is safe because an unauthorized user's job errors before Chrome runs. This fast path answers *before* any job exists: `findLiveInstanceGeneration` → `findMediaCacheEntry` → `respondFromLedger` streams the stored object and returns 201 with the base64. Nothing in that chain consults permissions.

**Why the bytes are worth protecting.** The GET `_screenshot/` twin serves the same ledger entries only after `checkPermission` enforces realm read in `internalHandle` — and on private realms the GET lane persists captures rendered as the realm owner. Those owner-rendered private captures are exactly what this path now hands to any account holder who can guess (or enumerate) a card URL. `findLiveInstanceGeneration` also runs before any check, so a private realm's instance existence and index generation leak the same way.

**The way out.** Fetch the realm's permissions in the handler and require read for `token.user` before entering the fast path (before `findLiveInstanceGeneration`, so the existence probe is covered too) — the same `fetchRealmPermissions` the task uses, requiring `read` rather than the task's any-permission bar. The render path can keep relying on the worker's check. A test pinning "user without read on the realm gets no ledger bytes" is the negative-space coverage the new suite is missing — today no test exercises a second identity at all.

**Scope: regression, blocking** — the store-first read is new here; the pre-existing endpoint always went through the permission-checked render.

packages/runtime-common/media-cache.ts
```diff
@@ -283,11 +296,37 @@ export async function findMediaCacheEntry(
...
     createdAt: Number(row.created_at),
     lastAccessedAt: Number(row.last_accessed_at),
   };
 }
 
+// The generation of a live indexed instance, addressable by either its
+// extensionless card-id URL or its `.json` file URL — the capture-identity
+// resolution a caller needs before it can compute a MediaCache key. Returns
+// undefined when the instance is absent or tombstoned.
+export async function findLiveInstanceGeneration(
```
lukemelia, line 319: [Claude Code 🤖] **This is a second implementation of "live instance generation" that disagrees with the one the GET route uses — an errored instance persists here but can never serve there, so the endpoint returns a `captures[].url` that 404s.**

**The divergence.** `IndexQueryEngine.liveInstanceGeneration` (`packages/runtime-common/index-query-engine.ts`) — what `serveScreenshot` gates on — excludes rows in effective-error state: `NOT effectiveHasError()`, which spans both the index error channel (`i.has_error`) and a current render error on the `prerendered_html` join. This raw-SQL version checks only `type = 'instance'` and not-deleted. For an instance whose latest index or render pass errored: the POST resolves a generation, persists the capture, and returns its served URL; the GET's gate returns undefined for the same instance and answers an uncaptured 404 miss. The ledger row sits unservable until the error clears — and since the instance isn't deleted, only the on-demand idle TTL ever reclaims it.

**Why this is also a maintenance trap.** One contract ("which instances are live for capture purposes") now has two homes that will drift again the next time the predicate learns something — exactly how the effective-error channel was missed here.

**The way out.** Share the predicate rather than re-deriving it: either export the effective-error condition from the query engine module and use it in this query (it needs the `prerendered_html` join), or give this helper the engine's exact WHERE clause with a comment binding the two. The realm-scoping (`realm_url =`) this version adds is fine to keep — it's stricter, not different.

**Scope: regression, non-blocking** — but worth fixing in this PR while both copies are one diff apart.

packages/realm-server/handlers/handle-screenshot-card.ts
```diff
@@ -121,3 +283,95 @@ export default function handleScreenshotCard({
...
+    height,
+    deviceScaleFactor: null,
+    ...(withBase64 && base64 !== undefined ? { base64 } : {}),
+  };
+}
+
+// A ledger hit answers with zero render work, in the same envelope a fresh
+// capture produces. Returns undefined when the entry's object is gone from
+// the store (reclaimed under a live row) so the caller re-captures.
+async function respondFromLedger({
```
lukemelia, line 367: [Claude Code 🤖] **A ledger hit here never bumps `last_accessed_at`, so a capture consumed only through this endpoint looks idle to the GC and ages out while actively in use.**

**The mechanism.** The GET route's `serveMediaCacheEntry` bumps the entry's last-accessed stamp on every hit (lane-scoped to on-demand, throttled to hourly — see `touch` in `packages/runtime-common/media-cache-serving.ts`), because `findMediaCacheGcCandidates`' on-demand arm reclaims entries idle past `MEDIA_CACHE_ON_DEMAND_TTL_MS`. This function streams the same objects and updates nothing. A capture that clients refresh exclusively via POST (an agent polling for base64, say) hits the TTL, gets reclaimed, and the next POST silently pays a full re-render — the exact cost the fast path exists to avoid.

**The way out.** Apply the same guarded bump on the hit: the lane/throttle logic already exists in `media-cache-serving.ts` — export that helper (or its guard) rather than duplicating the lane check and throttle window here, for the same one-home reason as the generation-probe comment on `media-cache.ts`.

**Scope: regression, non-blocking** — the store-first read is what makes the bump's absence matter.

packages/realm-server/handlers/handle-screenshot-card.ts
```diff
@@ -84,29 +148,127 @@ export default function handleScreenshotCard({
...
+
       let job = await enqueueScreenshotCardJob(
         {
           realmURL,
           realmUsername: userId,
           runAs: userId,
           cardId,
           format,
-          persist: null,
+          persist: entryKey ? { ...entryKey, lane: 'on-demand' } : null,
```
lukemelia, line 243: [Claude Code 🤖] **This `persist` value falsifies two documented queue contracts, and one of them now has a behavioral edge that needs a decision.**

**The stale docs.** `chooseScreenshotCardCoalesceDecision` (`packages/runtime-common/jobs/screenshot-card.ts`) documents that only the GET lane coalesces because "`POST /_screenshot-card` publishes with `persist: null` … those always insert" — false once this line ships. Same file, `estimateScreenshotQueueWait`'s twin SQL is annotated "POST jobs have no persist and never appear here" — also false. Docs move with contracts; both paragraphs need rewriting to whatever is decided below.

**The edge with teeth.** The coalesce decision requires equal `runAs`, but the twin SQL matches only `sourceURL`/`captureSpecHash`/`sourceGeneration`. The GET lane runs as the realm owner; this endpoint runs as the requester. So a pending POST job now flips `hasTwin` for a GET request with the same capture identity → the GET skips its congestion gate expecting to join → the join fails on `runAs` → it inserts its own render into a lane it was just told was safe to enter. (POST-persisted coalescing itself is sound by the comment's own generation-pinning argument — same-user POST twins joining is a win.)

**The decision.** Either (a) match `runAs` in the twin SQL so `hasTwin` only reports jobs the caller can actually join, or (b) accept POST twins as gate-skippers on the grounds that the twin persists under the same identity, so even a GET that times out retries into a pure ledger hit. (b) is defensible and cheaper; whichever you pick, the two comments should state it.

**Scope: regression of the documented contract, non-blocking.**

packages/realm-server/handlers/handle-screenshot-card.ts
```diff
@@ -84,29 +148,127 @@ export default function handleScreenshotCard({
...
+          });
+          if (response) {
+            return await setContextResponse(ctxt, response);
+          }
+          // The entry's object was reclaimed between the ledger read and
+          // the stream open — fall through and re-capture.
+        }
+      }
+
       let job = await enqueueScreenshotCardJob(
```
lukemelia, line 236: [Claude Code 🤖] **The enqueue and the retry estimate key the per-realm lane off the raw body `realmURL`, while the GET lane keys off the realm's canonical URL — a slash-less `realmURL` splits the serialization lane in two.**

**The mechanism.** `enqueueScreenshotCardJob` defaults `concurrencyGroup` to `` `screenshot:${args.realmURL}` `` and this call passes the body's `realmURL` verbatim; the 503 path's `estimateScreenshotQueueWait` call below gets the same raw string. The GET lane (`captureScreenshotOnDemand` in `packages/runtime-common/realm.ts`) uses `` `screenshot:${this.url}` ``, which always carries the trailing slash. A client posting `"http://example.test"` lands its jobs in `screenshot:http://example.test` while every GET-lane job for that realm sits in `screenshot:http://example.test/` — the per-realm serialization the lane exists for stops covering both, and each side's queue-depth estimate goes blind to the other's jobs. The handler already computes `normalizedRealmURL` two screens up (the persist key uses it); only the enqueue args and the estimate string missed it.

**The way out.** Pass `normalizedRealmURL` as the job's `realmURL` and into the estimate. The task re-normalizes internally for permissions, so nothing else changes.

**Scope: regression, non-blocking.**

packages/realm-server/handlers/handle-screenshot-card.ts
```diff
@@ -73,6 +124,19 @@ export default function handleScreenshotCard({
...
       );
     }
+    if (includeBase64 !== undefined && typeof includeBase64 !== 'boolean') {
+      return sendResponseForBadRequest(ctxt, 'includeBase64 must be a boolean');
+    }
+    let withBase64 = includeBase64 !== false;
+    let normalizedRealmURL = ensureTrailingSlash(realmURL);
+    // The persist identity (and the served URL) hang off the instance's
+    // location within its realm.
+    if (!cardId.startsWith(normalizedRealmURL)) {
```
lukemelia: [Claude Code 🤖] **The containment check is a string prefix over an unnormalized URL, so it passes shapes it means to refuse and derives a `sourceURL` that can disagree with the GET route's for the same instance.**

**Two failure shapes.** (1) `cardId: "${realmURL}../other-realm/card"` starts with `normalizedRealmURL` and passes; `instanceLocalPath` becomes `../other-realm/card`. (2) Percent-encoding or dot-segment spelling differences: this path derives `sourceURL` by string-slicing `cardId`, while the GET derives it via `this.paths.fileURL(...)` — a `new URL` resolution that normalizes. Two spellings of one instance can therefore key differently across the two surfaces.

**Why it's contained today.** A dotted or oddly-encoded `sourceURL` matches no `boxel_index.url`/`file_alias` row (the index stores normalized URLs), so `findLiveInstanceGeneration` returns undefined and no persist or ledger read happens — the request degrades to the legacy render path, where Chrome's own URL resolution normalizes. Verified by reading `findLiveInstanceGeneration`'s exact-match WHERE clause. So this is a correctness-of-validation finding, not an exploitable one.

**The way out.** Normalize before checking: `let normalized = new URL(cardId).href`, refuse unless `normalized.startsWith(normalizedRealmURL)`, and derive `sourceURL`/`instanceLocalPath` from `normalized`. That makes the check mean what it says and pins POST/GET key parity for exotic ids in one move.

**Scope: regression (the check is new here), non-blocking.**

lukemelia (commented):
[Claude Code 🤖] This review focused on the write side this PR turns on: whether the persist identity really is the GET `_screenshot/` route's key, and what the new store-first fast path changes about the auth and queue contracts the render-only endpoint used to satisfy implicitly.

**Bottom line: one blocking finding — the ledger fast path streams persisted capture bytes (private realms included) to any authenticated realm-server user, because the only permission enforcement in this endpoint's flow lives in the worker task the fast path skips. The rest is non-blocking: identity parity with the GET DSL checks out for the normal path (same extensionless `sourceURL`, same `captureSpecHash`, generation pinned from the same table; the round-trip of `canonicalCaptureSpecQuery` through `parseCaptureSpecParams` is exact), the migration is correctly additive with the schema snapshot regenerated, and the response stays byte-compatible.**

Recommendations, in severity order:

1. Gate the ledger fast path (and the generation probe) on realm read for the JWT user — thread on the `findMediaCacheEntry` call in `handle-screenshot-card.ts`. Blocking.
2. `findLiveInstanceGeneration` lacks the effective-error exclusion its GET twin enforces — errored instances persist captures that can never serve, and the served URL it returns 404s. Thread on `media-cache.ts`.
3. A POST ledger hit never bumps `last_accessed_at`, so POST-only consumers age out of the GC's on-demand lane while in active use. Thread on `respondFromLedger`.
4. The queue module's coalesce/twin comments still say POST publishes `persist: null`; one of them guards a `hasTwin`-vs-`runAs` edge that needs a decision. Thread on the `persist:` line in the handler.
5. The enqueue and retry estimate use the raw body `realmURL` where the GET lane uses the canonical realm URL — a slash-less spelling splits the per-realm serialization lane. Thread on the enqueue call.
6. The `cardId` containment check is a string prefix over an unnormalized URL. Contained today; thread on the check.

Adjacent, out of scope: any authenticated user can still enqueue render jobs against arbitrary `realmURL`s — the worker refuses before Chrome runs, but each refusal costs a queue slot in that realm's serialized lane. That predates this PR and reads as follow-up material for whoever next touches the endpoint's admission control.

packages/realm-server/handlers/handle-screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in 6111acb352. The handler now fetches the realm permissions and requires read for the JWT user (via RealmPermissionChecker, so `*` and `users` grants count) before the generation probe, covering the existence leak as well as the ledger bytes. A caller without read goes straight to the render path, whose permissions the worker enforces, and never persists. New test: "a caller without realm read never touches the ledger" seeds distinct ledger bytes and asserts a second identity never receives them.

packages/runtime-common/media-cache.ts
lukemelia: [Claude Code 🤖] Fixed in 6111acb352. `effectiveHasError` and `prerenderedJoin` are now exported from the query engine, and `findLiveInstanceGeneration` builds its WHERE from them (keeping the realm scope), with a comment binding it to `IndexQueryEngine.liveInstanceGeneration`. Instantiating the engine itself was not practical here — its constructor needs a definition lookup and virtual network the handler does not have. New test: "an errored instance captures without persisting".

packages/realm-server/handlers/handle-screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in 6111acb352. The guarded bump is exported from media-cache-serving.ts as `touchMediaCacheEntryOnHit` (same lane check and hourly throttle) and `respondFromLedger` calls it on every hit. New test: "a ledger hit refreshes a stale last-accessed stamp" backdates the entry past the throttle window and asserts the POST hit bumps it.

packages/realm-server/handlers/handle-screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in 6111acb352, taking option (a): the twin SQL now matches `args->>'runAs'` and `twinOf` carries the `runAs` the caller would render under, so `hasTwin` only reports jobs the caller can actually join. The GET lane resolves the realm owner ahead of the congestion pre-check to supply it. Both stale comment paragraphs are rewritten to the new contract. New test: "hasTwin requires the runAs the caller would render under".

packages/realm-server/handlers/handle-screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in 6111acb352. The enqueue passes `normalizedRealmURL` (so the default concurrency group matches the GET lane) and the retry estimate keys off the same string.

packages/realm-server/handlers/handle-screenshot-card.ts
lukemelia: [Claude Code 🤖] Fixed in 6111acb352. Both `realmURL` and `cardId` now go through `new URL(...).href` (invalid URLs answer 400) before the containment check, and `sourceURL`/`instanceLocalPath` derive from the normalized form — so the check means real containment and the POST/GET persist keys agree for exotic spellings.

packages/realm-server/handlers/handle-screenshot-card.ts
habdelra: [Claude Code 🤖] Resolved for the fast path — the gate sits ahead of `findLiveInstanceGeneration`, so the existence probe is covered too, and `RealmPermissionChecker` matches what `Realm` itself accepts (same idiom as `handle-delegate-session.ts`).

One thing the fix surfaces rather than causes, worth a decision: the two lanes now enforce read under **different models**. The handler admits the `*` and `users` grants; the worker (`tasks/screenshot-card.ts`) admits only an exact `realmPermissions[runAsUserId]` row. On a realm carrying a `*: ['read']` grant, a caller with no row of their own gets a ledger hit from the fast path and a refused job from the worker for the same card — one endpoint, two answers to "may this user capture here", and which one they meet depends on whether the capture happens to be cached.

Pre-existing on the worker side, but this PR is what makes it load-bearing: before the fast path, that caller met the worker's answer every time. Either the worker moves onto `RealmPermissionChecker` too, or the handler doc comment should say plainly that the render path is stricter than the gate above it — right now it reads as though the worker enforces the same thing. Follow-up, non-blocking.

---

packages/runtime-common/media-cache.ts
habdelra: [Claude Code 🤖] Resolved — `findLiveInstanceGeneration` and `IndexQueryEngine.liveInstanceGeneration` now compile the same row predicate from the same two exported fragments, so the contract has one home and can't drift silently again. Not instantiating the engine here is the right call for the reason you give.

The coverage is half of it, though. `seedInstanceRow(1, { hasError: true })` sets `boxel_index.has_error`, which exercises only the index channel — the channel that was already excluded by nothing. The channel this fix actually adds is `RENDER_ERROR_IS_CURRENT` over the `prerendered_html` join, and nothing pins it: delete `prerenderedJoin()` and the `ph.*` half of `effectiveHasError`, and "an errored instance captures without persisting" still passes. A second case seeding a `prerendered_html` row with `error_doc` set and `generation >= i.generation` would pin the join and the generation comparison — which is where a future edit to this predicate would actually go wrong. Non-blocking.

---

packages/realm-server/handlers/handle-screenshot-card.ts
habdelra: [Claude Code 🤖] Option (a) resolves the case the comment named — a GET no longer sees a POST job under a different `runAs` as joinable, and the two stale paragraphs now describe what the code does.

The key still isn't the coalesce key, though, and one spelling gets through. `chooseScreenshotCardCoalesceDecision` requires `candidateArgs.cardId === incomingArgs.cardId`; the twin SQL matches persist target + `runAs` and never looks at `cardId`. Those differ in exactly one place: this handler enqueues `cardId: normalizedCardId` with the `.json` suffix intact, while `persist.sourceURL` has it stripped — and the GET lane enqueues `cardId: entryKey.sourceURL`, which is always extensionless (`realm.ts` strips before `paths.fileURL`).

So a realm owner POSTing `…/Person/x.json` leaves a pending job whose persist target *and* `runAs` both match a subsequent GET for that card: `hasTwin` reports true, the GET skips the congestion gate, the coalesce join then fails on `cardId`, and it inserts a render into the lane it was just told was safe. Narrower than the original (it needs the POST caller to be the realm owner and to spell the id with `.json`), but it's the same failure, and the new comment above the SQL asserts the key mirrors the coalesce join exactly — which is what makes it worth closing rather than documenting.

Cheaper on the enqueue side than in the SQL: pass `cardId: sourceURL` here instead of `normalizedCardId`. The render is unaffected — `render-runner.ts` already does `url.replace(/\.json$/i, '')` for `expectedId` — the two lanes then agree on the whole key, and two spellings of one card stop each paying a full Chrome render. Non-blocking.

---

packages/realm-server/handlers/handle-screenshot-card.ts
habdelra: [Claude Code 🤖] Resolved. The guard has one home, the internal call sites keep the short name, and backdating past the throttle window means the test pins the bump rather than the call. `touch` swallows its own failures, so awaiting it in `respondFromLedger` can't turn a serve into a 500.

---

packages/realm-server/handlers/handle-screenshot-card.ts
habdelra: [Claude Code 🤖] Resolved. Both the enqueue and the retry estimate key off `normalizedRealmURL` now, so the default concurrency group lands in the same `screenshot:<realm>/` lane the GET side builds from `this.url`.

---

packages/realm-server/handlers/handle-screenshot-card.ts
habdelra: [Claude Code 🤖] Resolved. Dot segments now collapse before the prefix test, so the check means containment, and a relative or malformed input answers 400 instead of falling through. `sourceURL`/`instanceLocalPath` deriving from the normalized form is what pins POST/GET key parity.

Percent-encoded dot segments (`%2E%2E`) still survive `new URL`, so they pass containment — but as before they match no `boxel_index` row, so they degrade to the render path rather than keying anything. Nothing to do; noting it so the next reader doesn't have to re-derive it.

---

habdelra (commented):
[Claude Code 🤖] Re-review of the fix commit only — I read the actual diff for each of the six threads rather than the replies, and re-ran the twin/drift checks over what changed; the first commit's ledger schema, response shape, and byte-compat coverage stay settled from the previous pass.

**All six findings are addressed and the blocking one is closed — the ledger fast path is gated on realm read ahead of the generation probe. Two threads stay open on details of their own fixes, both non-blocking.**

Dispositions:

1. **Ledger fast path unauthenticated** — resolved. Open note in-thread: the handler now admits `*`/`users` grants while the worker still requires an exact permissions row, so the endpoint's two lanes answer "may this user capture here" differently. Pre-existing, newly load-bearing; needs a decision, not a fix in this PR.
2. **Liveness predicate divergence** — resolved; one home for the predicate now. The new test covers only the `boxel_index.has_error` channel — the `prerendered_html` join this fix adds is untested and would survive being deleted. See thread.
3. **Ledger hit didn't bump `last_accessed_at`** — resolved, and the test backdates past the throttle so it pins the bump.
4. **`hasTwin` vs. the coalesce key** — resolved for the cross-surface case. Still reachable when a realm owner POSTs a `.json`-suffixed `cardId`: the twin SQL ignores `cardId`, the coalesce join requires it, so `hasTwin` can still wave a GET past the congestion gate into a render. One-line fix on the enqueue side; see thread.
5. **Split concurrency lane** — resolved.
6. **String-prefix containment check** — resolved.

CI: nothing red — remaining host and realm-server shards were still running at review time.

Adjacent, out of scope: `media-cache.ts` now imports `index-query-engine.ts`, which pulls the engine module earlier into the `index.ts` cycle than before. Safe today — both exported fragments are hoisted function declarations and the engine's module-level bindings are all string/array literals — but it's a cycle that would bite the first time either module gains initialization-time work.

---

habdelra (approved):

PR #5846, opened 2026-08-21
host: Fix `editTemplate` for prefix RRIs

backspace:
It was assuming the template would always have a URL base, not a prefix RI.

<details><summary>Claude explanation</summary>`identifyCard` reports a module in canonical form, so for a realm reached through a registered prefix it is an identifier rather than a URL. Edit Template parsed it as a URL, which throws — selecting any card whose type lives in the base, catalog or skills realm and asking to edit its template did nothing.

Pass the identifier through instead. `updateCodePath` already accepts one and resolves it, so nothing else has to change; the parse was the whole bug.

An audit of the other consumers found no second instance: of the nineteen `identify` / `identifyCard` call sites in the host, this was the only one feeding a module into `new URL`. The one other place that parses a code ref's module, `serializers/code-ref.ts`, guards on the identifier being URL-shaped first.</details>

### DIFF
--- a/packages/host/app/components/operator-mode/preview-panel/index.gts
+++ b/packages/host/app/components/operator-mode/preview-panel/index.gts
@@ -26,6 +26,7 @@ import {
   isCardInstance,
   isFileDefInstance,
   isResolvedCodeRef,
+  rri,
   cardDefFormats,
   fileDefFormats,
   fieldDefFormats,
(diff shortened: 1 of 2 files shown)


lukemelia (approved):
Good catch. Can we add something to our pr-review skill or elsewhere that will cause AI to catch this in code review?

backspace:
> Good catch. Can we add something to our pr-review skill or elsewhere that will cause AI to catch this in code review?

👍🏻 #5847 covers some of that, and now #5878 addresses it in the skill.

PR #5847, opened 2026-08-21
Add linting for RRI URL construction

backspace:
`new URL(rri)` is valid typewise because an RRI is a branded string, but there‘s no way to know whether it’s a prefix or URL RRI, and only URLs will work. So this adds a lint rule to prevent its use:

```
[lint:js]   12:10  error  `new URL()` on a RealmResourceIdentifier does not survive a prefix-form identifier: it throws with no base, and with one it resolves the prefix as a relative path into a URL that points nowhere. Use `new RealmPaths(ri(x))` for path work, `virtualNetwork.toURL(x)` at a genuine network boundary, or ask the realm server which realm a URL belongs to. If this really is the boundary, disable this rule on the line with a reason  @cardstack/boxel/no-url-from-realm-identifier
```

If it’s truly needed, the lint rule can be [ignored](https://github.com/cardstack/boxel/pull/5847/changes#diff-8059519ed65b0c5f571c9096c895a938d91a4623c4c2ff4a994026b86ebef7c0).

<details><summary>Claude explanation</summary>

The realm-identifier types are branded strings:

```ts
type RealmResourceIdentifier = string & { __rriBrand: unknown };
```

so `new URL(identifier)` typechecks. It fails only at runtime and only once a realm is reached through a registered prefix — with no base it throws, and with one it resolves the prefix as a relative path into a URL that points nowhere. The compiler cannot see the difference, which is why the same defect keeps recurring. The brand is a type, though, so a rule with type information can ask what the compiler cannot.

**The rule.** Asks whether a `new URL` argument carries either brand. It reads parser services directly rather than depending on `@typescript-eslint/utils`, and returns no visitors when type information is absent — so it is inert rather than wrong wherever a config sets no `parserOptions.project`, and can be adopted a package at a time.

**Seeing through brand erasure.** A branded string is a subtype of `string`, so TypeScript reduces `RealmResourceIdentifier | string` to plain `string`. A ternary between an identifier and a template literal therefore carries no brand at all — which is the shape the Edit Template defect took. So the argument is walked for a branded part rather than only asked for its own type: ternary branches, `??`/`||` operands, `+` operands, template interpolations, casts, and `const` bindings resolved to their initializer. Only what fixes the *leading* spelling propagates, so `` `${id}.gts` `` is reported and `` `https://example.test/x/${id}` `` is not.

**Enabled as an error, with no standing violations.** `runtime-common` and host app code both build a TypeScript program for it. Three sites were fixed rather than suppressed:

- `file-serializer.ts` parsed `meta.realmURL` — a `RealmIdentifier` that arrives from the wire — and now resolves it through the VirtualNetwork it was already handed.
- `code-editor.gts` parsed the open file's identifier before passing it to `codeRefWithAbsoluteIdentifier`, which accepts an identifier directly. Same defect as the Edit Template button.
- `trimExecutableExtension` took and returned an identifier, so callers holding a URL href had to brand it to call at all. An overload keeps the caller's form, and `ModuleSyntax` drops a wrapper it only had to satisfy the signature.

The remaining sites were each traced and carry an inline disable naming the guard that makes the parse safe: a prefix form that returns earlier, one that takes the other arm of a ternary, one handled by a catch that resolves it against the realm and reads the same segment, and values that are URL-form by construction. `routes/index.gts` carries the one that is safe rather than correct — the realm list holds URL forms today, and its comment says what would have to be decided if prefix forms ever reach it, since there is no path to take apart in `@scope/name/`.

**Scope.** Host's program covers `app/**` only. The build and config scripts at the package root are outside `tsconfig.json`, where asking for a program is a parse error rather than a lint result. Tests are out for signal: they build URLs by interpolating the test realm's identifier, which is always a URL there, so the rule has 342 things to say about them. Type information costs host's lint about 13 seconds and runtime-common's about 5.

</details>


packages/eslint-plugin-boxel/lib/rules/no-url-from-realm-identifier.js
```diff
@@ -0,0 +1,186 @@
...
+            brandInExpression(node.right, depth + 1)
+          );
+        case 'BinaryExpression':
+          return node.operator === '+'
+            ? brandInExpression(node.left, depth + 1) ||
+                brandInExpression(node.right, depth + 1)
+            : undefined;
+        case 'TemplateLiteral':
+          for (const part of node.expressions) {
+            const found = brandInExpression(part, depth + 1);
```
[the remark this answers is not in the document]
backspace: [Claude Code 🤖] Agreed and fixed in 907cdf67fa.

The underlying principle is narrower than "exclude absolute prefixes": what decides whether the result is an identifier is whatever fixes the *leading* spelling. So a template literal whose first quasi is non-empty is left alone entirely, and only `expressions[0]` is consulted when it is empty. That keeps `` `${id}.gts` `` reported and drops `` `https://example.test/lookup/${id}` ``, and it also handles the case your example doesn't cover — `` `${scheme}://x/${id}` `` — where the leading interpolation is unbranded and the trailing one is irrelevant.

I applied the same reasoning to `+`, which had the same flaw and which you didn't flag: it now propagates from the left operand rather than either, with an empty-string literal falling through to the right.

On scale: ten sites in the repo build a URL by interpolating behind an absolute prefix (ports, host headers, query strings), and none interpolates an identifier today, so this changes no current report — host stays at 8 and runtime-common at 6. Your point about needing unnecessary disables once the rule is an error is the reason to do it now rather than when `realm-server` gets enabled and the pattern becomes live.

packages/eslint-plugin-boxel/lib/rules/no-url-from-realm-identifier.js
```diff
@@ -0,0 +1,186 @@
...
+        case 'Identifier': {
+          const init = constInitializerOf(node);
+          return init ? brandInExpression(init, depth + 1) : undefined;
+        }
+        case 'TSAsExpression':
+        case 'TSNonNullExpression':
+        case 'TSSatisfiesExpression':
+          return brandInExpression(node.expression, depth + 1);
+        default:
+          return undefined;
```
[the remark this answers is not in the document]
backspace: [Claude Code 🤖] I checked this one against the tree rather than in the abstract, and I'm not taking it — but the reasoning is worth recording, because the mechanism you describe is real.

Your example is not hypothetical: `new URL(href.replace(/\.json$/, ''))` appears verbatim at `runtime-common/realm-index-updater.ts:296` and `:359`. I probed the checker for the receiver type at each site, and `href` is plain `string` at both, so the suggestion would not report them. Across the repo there are six `new URL(x.<string method>(…))` sites, and exactly one has a branded receiver: `virtual-network.ts:512`, `new URL(relativeTo.slice(prefix.length), target)` — which strips the prefix and resolves against the mapped target, i.e. the sanctioned de-VN boundary and precisely what the rule's message tells you to do instead. So the one report this would add is a false positive.

The deeper problem is that a receiver-based check can't tell the defect from the fix. `.replace` is both how a prefix survives a transformation and how a prefix is *removed*: `id.replace('@cardstack/base/', 'https://cardstack.com/base/')` is the correction, and it has the same shape as your example. Distinguishing them means modelling the arguments, and a rule that inspects regex and replacement strings to guess at the resulting spelling is guessing.

So this stays a known false negative, and the doc now says which forms do propagate the brand. If a branded `.replace` that keeps the prefix shows up in real code, that's the point to revisit — with the argument-modelling problem in view.

One genuine defect your comment led me to, though the mechanism is unrelated: the rule's message claimed `new URL()` "throws", which is wrong when a base is passed. `new URL(identifier, base)` doesn't throw — the prefix form is a valid relative reference, so it resolves to `base` with `@cardstack/base/card-api` appended, pointing nowhere. Fixed in 907cdf67fa, since a message that promises an exception sends someone looking for one that never arrives.

packages/eslint-plugin-boxel/lib/rules/no-url-from-realm-identifier.js
```diff
@@ -0,0 +1,194 @@
...
+          return;
+        }
+        const brand = brandInExpression(arg);
+        if (brand) {
+          context.report({
+            node,
+            messageId: 'urlFromIdentifier',
+            data: { brand: BRANDS[brand] },
+          });
+        }
```
lukemelia: [Claude Code 🤖] The visitor only walks `node.arguments[0]`, so a branded *base* slips through: `new URL(path, identifierBase)` throws for a prefix-form base exactly like a branded first argument does (`new URL('x', '@scope/name/')` → `Invalid base URL`), and resolving a relative reference against a realm base is a realistic shape. The `new URL(rri(...), 'https://…/')` test pins the first-arg side but leaves the base unguarded, so this is the rule's own defect one argument over.

Extend the check to the base:
```js
const branded = (a) =>
  a && a.type !== 'SpreadElement' ? brandInExpression(a) : undefined;
const brand = branded(node.arguments[0]) || branded(node.arguments[1]);
```
Non-blocking — a completeness follow-up, not a regression; the primary `new URL(identifier)` shape is caught. Small enough to fold in here or take separately.

lukemelia (commented):
[Claude Code 🤖] This review went after the rule's detection logic (the brand walk, the brand-erasure cases, `const` resolution), its test coverage, whether the two package configs actually supply the TS program the rule needs to be active rather than silently inert, and each call-site fix and suppression. It did not re-open the two settled Codex threads.

**No blocking issues** — one non-blocking completeness gap in the rule itself, inline below.

1. The visitor inspects only `new URL`'s first argument, so a branded *base* (`new URL(path, identifierBase)`) slips through the same way — see the thread on `NewExpression` in `no-url-from-realm-identifier.js`.

packages/eslint-plugin-boxel/lib/rules/no-url-from-realm-identifier.js
```diff
@@ -0,0 +1,194 @@
...
+
+    return {
+      NewExpression(node) {
+        if (node.callee.type !== 'Identifier' || node.callee.name !== 'URL') {
+          return;
+        }
+        const arg = node.arguments[0];
+        if (!arg || arg.type === 'SpreadElement') {
+          return;
+        }
```
richardhjtan: [Claude Code 🤖] This matches `new URL` only, and it is not the only prefix-unsafe URL API — `URL.canParse` and `URL.parse` have the same blindness and fail *more* quietly, returning `false` / `null` instead of throwing.

That is not hypothetical. `runtime-common/code-ref.ts`'s `isRelativePath` does `return !URL.canParse(moduleId)`, and its caller `resolveAdoptsFrom` passes `adoptsFrom.module` — typed `RealmResourceIdentifier` on `ResolvedCodeRef`. So for the canonical prefix spelling this repo uses everywhere (`@cardstack/base/card-api` and friends from `constants.ts`), `URL.canParse` answers `false` and the identifier is classified as a *relative path*: the rule's own message, one API over, in a package where the rule is now an error.

I traced whether it bites, and it does not — both downstream paths recognize prefix forms, `resolveRRIReference` passing them through unchanged and `resolveModuleHref` admitting them via `virtualNetwork.isRegisteredPrefix` before resolving. The misclassification is absorbed. But that is a downstream guard doing the work, not a correct classification, and it is precisely the kind of accident this rule exists to stop relying on.

Two asks:

- Add `URL.canParse` / `URL.parse` to the callee match (a `MemberExpression` callee on `URL`). The silent-return failure mode makes them a higher priority than the base argument, not a lower one — a throw at least gets someone's attention.
- With them covered, `isRelativePath` becomes a report, and it belongs with the traced sites: either a disable naming the downstream prefix handling that makes it safe, or the honest fix, which is to ask whether the value is a registered prefix before asking whether it parses.

Together with the base-argument gap in the thread below, this is the same completeness question from the other side: the rule currently guards one of the three ways this codebase asks "is this a URL".

Class: follow-up, not this PR's bug. Non-blocking.

richardhjtan (commented):
[Claude Code 🤖]

This review took the angle the existing threads had not: given the rule's premise that this defect keeps recurring because the compiler cannot see it, which *other* URL APIs share the property, and is any of them already reached with a branded value in the packages where the rule is now an error. It did not re-review the brand-propagation walk or the per-site disables, both already covered above.

**No blocking issues. The rule catches the shape it names; the ask is that `new URL` is not the only prefix-unsafe URL API in play, and the other one is already called with a `RealmResourceIdentifier` inside `runtime-common`.** Detail in the thread on the `NewExpression` visitor.

Also verified rather than left to inference: `RealmPaths`' `RealmIdentifier` overload does handle prefix forms (`ensureTrailingSlash(decodeUriIfNeeded(id))`, no parse), so the remedy the message recommends is sound.

1. Extend the callee match to `URL.canParse` / `URL.parse`, which fail silently rather than throwing — thread on the visitor.

packages/eslint-plugin-boxel/lib/rules/no-url-from-realm-identifier.js
backspace: [Claude Code 🤖] Valid, and folded in — a5c00a0a6c. `new URL('x', '@scope/name/')` throws `TypeError: Invalid URL`, confirmed against node rather than assumed.

I took the structure but not quite the snippet: the two arguments get separate messages, because they fail for different reasons and the advice differs. A branded first argument throws with no base and mis-resolves with one; a branded base just throws. The first argument reports in preference when both are branded, since that is the value the author is usually reasoning about.

The interesting part is what it surfaced. Seven sites, none of which existed as reports before:

- **Two in `VirtualNetwork`** — `virtual-network.ts:378` and `:505`. Both sit inside a branch that has already established the base is a URL (`base.startsWith('http')`, and the `isUrlRelativeTo` branch respectively), so both take a disable naming that branch.
- **Five in host** — `room.gts`, `routes/index.gts`, `migrate-skill.ts`, `write-binary-file.ts`, `write-text-file.ts`. All the same shape: joining a relative path against a realm identifier.

Those five are the reason this wasn't quite as small as it looked. `new RealmPaths(id).fileRRI(path)` is the form-agnostic join they want — it concatenates for a prefix realm and joins for a URL realm — and `getSource` already accepts an identifier. But each consumes the result as a URL object downstream (`finalUrl.href`, `new URL(nonConflictingUrl)`), so converting them is a change to those five call sites rather than to the rule.

So they carry a disable that says the guard is absent and reachability by a prefix form is unverified — the same honest phrasing the earlier `routes/index.gts` and `patch-code.ts` sites use in this PR, rather than claiming a guard that isn't there. That keeps the rule complete and the five sites greppable instead of invisible. Worth triaging as a follow-up; I did not want to fold five call-site refactors into a lint-rule PR without asking.

Four new test cases cover the base shape, including a valid case pinning that a URL base is still fine.

packages/eslint-plugin-boxel/lib/rules/no-url-from-realm-identifier.js
backspace: [Claude Code 🤖] The API gap is real and it's in — a5c00a0a6c. Confirmed both failure modes directly: `URL.canParse('@cardstack/base/card-api')` is `false`, `URL.parse(...)` is `null`, against `true` for the URL spelling.

I agree the silent return makes them the higher priority, and they get their own message for exactly that reason — it names the falsy answer rather than saying "it throws", since telling someone hunting a silent misclassification that an exception is thrown sends them looking for the wrong thing.

**Your second ask doesn't happen, though, and the reason is worth recording.** `isRelativePath` is not reported, even with `canParse` covered. Its signature is `isRelativePath(moduleId: unknown)`, and the `typeof moduleId !== 'string'` check narrows it to plain `string` — so at the `URL.canParse(moduleId)` call site there is no brand left to find. The brand died at the function boundary, one frame above where the rule is looking.

So the trace in your comment is right about the mechanism and right that it is absorbed downstream, but the conclusion that it becomes a report is wrong: the rule is structurally blind to it. That is a genuine limitation of a type-directed check — an `unknown` or `string` parameter erases the brand as effectively as a ternary does, and no amount of walking the argument expression recovers it. Only the caller's site carries the type.

Which means the defect you found stands entirely on your reading rather than on the rule, and is worth its own issue: `isRelativePath('@cardstack/base/card-api')` returns `true`, classifying a canonical prefix identifier as a relative path, and it only works because `resolveRRIReference` and `resolveModuleHref` both happen to tolerate prefix forms. The honest fix is the one you name — ask whether the value is a registered prefix before asking whether it parses — and I would rather do that with test coverage than slip it into this PR.

Thanks for checking the `RealmPaths` overload independently; that the remedy the message recommends actually handles prefix forms is load-bearing for the whole rule, and it wasn't pinned by anything before.

What the extension did surface is seven base-argument sites, detailed in the thread above.

habdelra (approved):

PR #5848, opened 2026-08-21
Update `ember-eslint-parser` and `eslint-plugin-ember`

backspace:
This is a step to make #5847 more useful.

<details><summary>Claude explanation</summary>

`ember-eslint-parser` 0.13.0 to 0.14.6, and `eslint-plugin-ember` 13.3.2 to 13.5.0. The two move together: 13.3.2 depends on the 0.13 line, so bumping only the parser leaves two copies in the tree, and the parser's own duplicate-version guard then fails every `.gts` file. 13.5.0 depends on `^0.14.2`, so the tree collapses to one parser again. The local patch applies to 0.14.6 unchanged.

Beyond keeping current, 0.14.6 fixes how `.gts` files enter a TypeScript program. That fix is not exercised by this branch: it only shows up under `parserOptions.project`, which no config here sets. Measured on a branch that does add it, linting host went from 974 files reporting that the tsconfig does not include them to zero, and from 337s to 49s. This PR is what makes that possible; expect no change to the Lint job's own timing.

The parser moves into the catalog rather than repeating `^0.14.6` across four manifests. Four hand-maintained ranges are how a tree acquires the two copies the guard trips on, and every neighbour — eslint, the `@typescript-eslint` pair, every other `eslint-plugin-*` — is catalogued already.

`packages/host`, `packages/boxel-icons` and `packages/boxel-ui/docs-app` now declare the parser their eslintrc already names. Without it, which of the lockfile's two peer variants they load is decided by hoisting; `noop.js` keeps `parsedFiles` in a module-level Set, so a parser and processor from different copies make every `.gts` file report the setup-guide error, which this repo's patch makes fatal.

</details>

### DIFF
--- a/packages/runtime-common/package.json
+++ b/packages/runtime-common/package.json
@@ -60,7 +60,7 @@
     "decorator-transforms": "catalog:",
     "diff": "catalog:",
     "ember-concurrency-async-plugin": "workspace:*",
-    "ember-eslint-parser": "^0.13.0",
+    "ember-eslint-parser": "^0.14.6",
     "ember-source": "catalog:",
     "eslint": "catalog:",
     "ethers": "catalog:",
--- a/packages/runtime-common/package.json
+++ b/packages/runtime-common/package.json
@@ -60,7 +60,7 @@
     "decorator-transforms": "catalog:",
     "diff": "catalog:",
     "ember-concurrency-async-plugin": "workspace:*",
-    "ember-eslint-parser": "^0.14.6",
+    "ember-eslint-parser": "catalog:",
     "ember-source": "catalog:",
     "eslint": "catalog:",
     "ethers": "catalog:",
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -146,7 +146,7 @@ catalog:
   eslint: ^8.57.1
   eslint-config-prettier: ^9.1.2
   eslint-doc-generator: ^1.7.1
-  eslint-plugin-ember: ^13.3.2
+  eslint-plugin-ember: ^13.5.0
   eslint-plugin-eslint-comments: ^3.2.0
   eslint-plugin-eslint-plugin: ^5.5.1
   eslint-plugin-filenames: ^1.3.2
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -133,6 +133,7 @@ catalog:
   ember-concurrency: ^5.2.0
   ember-concurrency-ts: ^0.3.1
   ember-elsewhere: ^2.0.1
+  ember-eslint-parser: ^0.14.6
   ember-freestyle: ^0.24.0
   ember-keyboard: ^9.0.4
   ember-modify-based-class-resource: ^1.1.0
--- a/packages/boxel-icons/package.json
+++ b/packages/boxel-icons/package.json
@@ -57,6 +57,7 @@
     "@typescript-eslint/parser": "catalog:",
     "babel-plugin-ember-template-compilation": "catalog:",
     "concurrently": "catalog:",
+    "ember-eslint-parser": "catalog:",
     "ember-source": "catalog:",
     "ember-template-imports": "^4.1.1",
     "ember-template-lint": "catalog:",
--- a/packages/boxel-ui/docs-app/package.json
+++ b/packages/boxel-ui/docs-app/package.json
@@ -71,6 +71,7 @@
     "ember-cli-htmlbars": "^6.3.0",
     "ember-concurrency": "catalog:",
     "ember-css-url": "^1.0.0",
+    "ember-eslint-parser": "catalog:",
     "ember-freestyle": "catalog:",
     "ember-load-initializers": "^3.0.0",
     "ember-modifier": "^4.1.0",
--- a/packages/catalog/package.json
+++ b/packages/catalog/package.json
@@ -18,7 +18,7 @@
     "@universal-ember/test-support": "catalog:",
     "concurrently": "catalog:",
     "ember-concurrency": "catalog:",
-    "ember-eslint-parser": "^0.13.0",
+    "ember-eslint-parser": "^0.14.6",
     "ember-modifier": "^4.1.0",
     "ember-resources": "catalog:",
     "ember-template-lint": "catalog:",
--- a/packages/catalog/package.json
+++ b/packages/catalog/package.json
@@ -18,7 +18,7 @@
     "@universal-ember/test-support": "catalog:",
     "concurrently": "catalog:",
     "ember-concurrency": "catalog:",
-    "ember-eslint-parser": "^0.14.6",
+    "ember-eslint-parser": "catalog:",
     "ember-modifier": "^4.1.0",
     "ember-resources": "catalog:",
     "ember-template-lint": "catalog:",
--- a/packages/eslint-plugin-boxel/package.json
+++ b/packages/eslint-plugin-boxel/package.json
@@ -37,7 +37,7 @@
     "eslint": "catalog:"
   },
   "dependencies": {
-    "ember-eslint-parser": "^0.13.0",
+    "ember-eslint-parser": "^0.14.6",
     "requireindex": "catalog:"
   },
   "devDependencies": {
--- a/packages/eslint-plugin-boxel/package.json
+++ b/packages/eslint-plugin-boxel/package.json
@@ -37,7 +37,7 @@
     "eslint": "catalog:"
   },
   "dependencies": {
-    "ember-eslint-parser": "^0.14.6",
+    "ember-eslint-parser": "catalog:",
     "requireindex": "catalog:"
   },
   "devDependencies": {
--- a/packages/host/package.json
+++ b/packages/host/package.json
@@ -123,6 +123,7 @@
     "ember-concurrency": "catalog:",
     "ember-css-url": "^1.0.0",
     "ember-elsewhere": "catalog:",
+    "ember-eslint-parser": "catalog:",
     "ember-exam": "^10.1.0",
     "ember-focus-trap": "^1.0.1",
     "ember-freestyle": "catalog:",
--- a/packages/realm-server/package.json
+++ b/packages/realm-server/package.json
@@ -51,7 +51,7 @@
     "cron": "catalog:",
     "date-fns": "catalog:",
     "decorator-transforms": "catalog:",
-    "ember-eslint-parser": "^0.14.6",
+    "ember-eslint-parser": "catalog:",
     "ember-template-recast": "^6.1.5",
     "eslint": "catalog:",
     "eslint-plugin-qunit": "catalog:",
--- a/packages/realm-server/package.json
+++ b/packages/realm-server/package.json
@@ -51,7 +51,7 @@
     "cron": "catalog:",
     "date-fns": "catalog:",
     "decorator-transforms": "catalog:",
-    "ember-eslint-parser": "^0.13.0",
+    "ember-eslint-parser": "^0.14.6",
     "ember-template-recast": "^6.1.5",
     "eslint": "catalog:",
     "eslint-plugin-qunit": "catalog:",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -363,6 +363,9 @@ catalogs:
     ember-elsewhere:
       specifier: ^2.0.1
       version: 2.0.1
+    ember-eslint-parser:
+      specifier: ^0.14.6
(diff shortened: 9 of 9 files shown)


pnpm-workspace.yaml
```diff
@@ -146,7 +146,7 @@ catalog:
   eslint: ^8.57.1
   eslint-config-prettier: ^9.1.2
   eslint-doc-generator: ^1.7.1
-  eslint-plugin-ember: ^13.3.2
+  eslint-plugin-ember: ^13.5.0
```
habdelra, line 150: [Claude Code 🤖] **`ember-eslint-parser` is the one member of this family that isn't in the catalog, and this diff is the argument for putting it there.** One edit on this line moved `eslint-plugin-ember` for all five of its consumers (`packages/catalog`, `packages/boxel-icons`, `packages/boxel-ui`, `packages/boxel-ui/docs-app`, `packages/host`). The parser needed the identical `^0.14.6` literal typed into four separate manifests — `packages/catalog`, `packages/eslint-plugin-boxel`, `packages/realm-server`, `packages/runtime-common`.

**Why this is more than tidiness here.** The commit message names the failure mode precisely: two parser copies in the tree make the patched `noop.js` guard fatal on every `.gts` file. Four hand-maintained ranges are exactly how a tree acquires two copies — the next bump only has to miss one of them, and the symptom is every `.gts` file failing to lint rather than anything that reads as a version problem.

**The catalog already covers everything around it.** `pnpm-workspace.yaml` carries `eslint`, `eslint-config-prettier`, `eslint-doc-generator`, every `eslint-plugin-*`, the `@typescript-eslint/*` pair, and `babel-eslint`. The catalog protocol is also already exercised on a published manifest in this repo: `@cardstack/eslint-plugin-boxel` is not `private`, and its `dependencies` already carry `"requireindex": "catalog:"` — so the fact that the parser is a runtime `dependency` there rather than a devDependency isn't a blocker.

**The fix.** Add it to the catalog in its alphabetical slot, between `ember-elsewhere` and `ember-freestyle`:

```yaml
  ember-eslint-parser: ^0.14.6
```

then set all four manifests to `"ember-eslint-parser": "catalog:"`. That shrinks this diff rather than growing it.

Class: pre-existing, now load-bearing — the duplication predates this change, but this is the PR that pays for it. Non-blocking.

---

packages/runtime-common/package.json
```diff
@@ -60,7 +60,7 @@
     "decorator-transforms": "catalog:",
     "diff": "catalog:",
     "ember-concurrency-async-plugin": "workspace:*",
-    "ember-eslint-parser": "^0.13.0",
+    "ember-eslint-parser": "^0.14.6",
```
habdelra: [Claude Code 🤖] **This declaration is not what decides which parser copy the realm-server's runtime linter loads — and three packages that name the parser don't declare it at all.** After this bump the copy those packages get is still chosen by pnpm's hoisting rather than by any manifest: the same "relied on finding it hoisted from elsewhere" property the description calls out for the eslint binary.

**How the runtime linter actually loads it.** `packages/runtime-common/tasks/lint.ts` builds ESLint with `cwd: HOST_PKG` and `overrideConfigFile: HOST_ESLINTRC`, takes `eslint` itself through `Module.createRequire(packages/host/package.json)`, then calls `ensurePnpmNodePath()`, which prepends `node_modules/.pnpm/node_modules` to `NODE_PATH`. Its own comment says why: *"pnpm doesn't hoist transitive deps, so parsers/plugins host references by string (e.g. `parser: 'ember-eslint-parser'`) won't resolve from here."* So the parser that lints card source comes out of the hoist directory, not out of runtime-common's tree — this line pins a copy that path never reaches.

**Who names it without declaring it.** `packages/host/.eslintrc.js` (`files: ['**/*.gts'], parser: 'ember-eslint-parser'`), `packages/boxel-icons/.eslintrc.cjs`, and `packages/boxel-ui/docs-app/.eslintrc.js`. None of the three lists it in `package.json`. `packages/eslint-plugin-boxel/lib/config/base.js` passes the bare string on to consumers too, with the comment "Don't directly require the parser here", so eslintrc resolves it relative to the *consuming* package rather than the plugin.

**Why this can bite rather than just being untidy.** `ember-eslint-parser/noop` keeps `parsedFiles` in a module-level `Set`: the parser calls `registerParsedFile`, and `postprocess` checks membership from whichever copy the *processor* was required from. host extends `plugin:ember/recommended-gts`, so its processor comes from `packages/host/node_modules/eslint-plugin-ember` → the eslint-8 peer variant of the parser, while the `parser:` string resolves through the hoist. `pnpm-lock.yaml` on this branch carries two peer variants of `ember-eslint-parser@0.14.6` — one suffixed `(...eslint@8.57.1)`, one `(...eslint@9.39.5)` — and the second exists only because `packages/boxel-ui` pins `eslint: ^9.17.0` instead of the catalog's 8. Two suffixes means two directories in the virtual store, so two module instances with two `parsedFiles` sets. If hoisting lands on the eslint-9 one, every `.gts` file gets the "please follow the setup guide" message, and `patches/ember-eslint-parser.patch` sets `msgs[0].fatal = true`, so it fails rather than warns.

The Lint job is green on this head, so hoisting is currently landing on the matching copy. Nothing in a manifest makes it do so.

**The contrast that shows the fix.** `packages/boxel-ui/eslint.config.mjs` is flat config and writes `parser: ember.parser` — the object taken off the same `eslint-plugin-ember` instance it took `ember.configs.gts` from — so parser and processor are one module by construction, and boxel-ui needs no declaration. The eslintrc packages can't express that; declaring the dependency is the equivalent, because pnpm then links the copy resolved in that importer's own peer context.

**The fix.** Add `"ember-eslint-parser": "catalog:"` (see the comment on `pnpm-workspace.yaml`) to the `devDependencies` of `packages/host`, `packages/boxel-icons` and `packages/boxel-ui/docs-app`. All three resolve eslint 8 through the catalog, so each gets the same peer-suffixed directory `eslint-plugin-ember` already uses there.

Class: pre-existing, now load-bearing. This PR doesn't introduce the hoist dependence, but it is the change that moves the package whose duplication trips the guard — and the headline measurement in the description (974 → 0 on host) was taken on the one package that names the parser without declaring it. Non-blocking.

---

habdelra (commented):
[Claude Code 🤖] Reviewed this as a dependency-coherence change: whether the tree really collapses to one parser, whether the local patch still lands on the code it claims to, and whether the packages that *name* the parser are the ones that decide which copy they get.

**No blocking issues — the bump is coherent and the mechanical claims in the description check out. Two asks, both about how the version is spelled rather than which version it is, plus a description fix before this squashes.**

**What I verified**

- **One copy of each, post-bump.** `pnpm-lock.yaml` now carries exactly one `ember-eslint-parser` resolution version (`0.14.6`) and one `eslint-plugin-ember` (`13.5.0`); no `0.13.0` / `13.3.2` entries survive. Registry metadata backs the ordering argument: `eslint-plugin-ember@13.3.2` depends on `ember-eslint-parser@^0.13.0` and `13.5.0` on `^0.14.2`, so the two genuinely have to move together. Both are at `latest`.
- **The patch still lands on live code.** `src/preprocessor/noop.js` is byte-identical between 0.13.0 and 0.14.6, and `patch -p1 --dry-run` of `patches/ember-eslint-parser.patch` against the 0.14.6 tarball applies clean — the `fatal = true` line still sits inside the `postprocess` that the guard runs through.
- **The `@typescript-eslint/tsconfig-utils` dedupe is real.** 8.61.1 drops out and the parser rides on 8.67.0, which is what the rest of the tree already resolves.
- **CI.** The Lint job — the check that actually exercises the new parser — is green. The red `Realm Server Test Results` is the aggregate report, not a job: all six `Realm Server Tests (n, 6)` shards passed, and its 4 628 runs against 2 231 tests is a full retry, so that is a flake that passed on the second attempt rather than anything this diff caused.

**Recommendations**

1. Put `ember-eslint-parser` in the catalog instead of repeating `^0.14.6` across four manifests — detail in the comment on `pnpm-workspace.yaml`.
2. Declare the parser in the three packages whose eslintrc names it but whose manifest doesn't (`host`, `boxel-icons`, `boxel-ui/docs-app`) — detail in the comment on `packages/runtime-common/package.json`.
3. **The description no longer matches the diff, and it lands in history on squash.** Three things:
   - "Declare eslint in realm-server and software-factory" describes work that is already on `main` at this branch's merge base — both manifests carry `"eslint": "catalog:"` there. The commit message gets this right ("Depends on eslint being declared in…"); the body reads as if the declaration is part of this change.
   - The opening line is garbled: `Claude explanation` is glued onto the first sentence, and an apostrophe is HTML-escaped.
   - The 974 → 0 and 337s → 49s numbers were measured with `parserOptions.project` in play, and `packages/host/.eslintrc.js` sets no `project` on any override on this branch. As written, a reader will expect the Lint job to get ~7× faster here and it won't. Worth saying the numbers come from the follow-on change that adds `parserOptions.project`, which is what this unblocks.

**Adjacent, out of scope**

`packages/boxel-ui` pins `eslint: ^9.17.0` and `typescript-eslint: 8.19.1` outside the catalog while everything else is on the catalog's `^8.57.1`. That is the sole reason the tree carries two peer variants of `eslint-plugin-ember` and `ember-eslint-parser` at all, which is what gives recommendation 2 its teeth. Not this PR's to fix, but whoever unifies eslint versions should know it is the last holdout.

---

pnpm-workspace.yaml
backspace: [Claude Code 🤖] Done in 134d4a95a4 — catalogued at `^0.14.6` in the alphabetical slot, and all four manifests now say `catalog:`.

The argument that persuaded me is the asymmetry you point at: one line moved `eslint-plugin-ember` for five consumers while the parser needed the identical literal typed four times. Since the failure mode is two copies making the patched guard fatal on every `.gts` file, the duplication isn't cosmetic — it is the specific way this tree would reacquire the state the bump exists to fix, and it would present as broken linting rather than as a version problem.

The `requireindex` precedent settled the one thing I would otherwise have hesitated over, since the parser is a runtime `dependency` in `@cardstack/eslint-plugin-boxel` rather than a devDependency.

packages/runtime-common/package.json
backspace: [Claude Code 🤖] Done in 6490409762 — `ember-eslint-parser: catalog:` added to the devDependencies of `host`, `boxel-icons` and `boxel-ui/docs-app`.

I verified the mechanism rather than taking it on trust, and it holds. The lockfile does carry two peer variants of `ember-eslint-parser@0.14.6`, one resolved through `eslint@8.57.1` and one through `eslint@9.39.5`. After declaring, all three packages link the same directory, and — the check that actually matters for the `parsedFiles` Set — host's own parser and the copy its `eslint-plugin-ember` uses now resolve to one path. So parser and processor are one module instance by resolution rather than by luck.

Worth recording that I got this wrong once mid-check: my first comparison reported the two as differing, which was my own path handling mangling a relative symlink rather than a real split. Re-resolving with `realpath` gave the match.

The contrast with `boxel-ui`'s flat config is the part I found most useful — `parser: ember.parser` taken off the same plugin instance makes the question unaskable, and declaring the dependency is the nearest eslintrc equivalent. Lint is green across all seven affected packages after the change.

backspace:
[Claude Code 🤖] All three addressed. Recommendations 1 and 2 are 134d4a95a4 and 6490409762, with detail on their own threads; recommendation 3 is a rewritten description, since it lands in history on squash.

On the description, two of your three points reproduced and one didn't:

- **The eslint-declaration paragraph was describing work already on `main`.** It came from the prerequisite that merged as #5851, and the body read as though it were part of this diff. Removed.
- **The measurement needed context.** The 974 → 0 and 337s → 49s figures were taken on a branch that sets `parserOptions.project`; nothing here does. The description now says so explicitly and states that the Lint job's own timing will not change — you're right that as written it promised a speedup this PR doesn't deliver.
- **The HTML-escaped apostrophe didn't reproduce.** `gh api ... --jq '.body'` on the raw body returns no HTML entities, so that may have been a rendering artifact on your side. The run-on after `</summary>` was real and is fixed.

On the adjacent finding: `boxel-ui` pinning `eslint: ^9.17.0` and `typescript-eslint` outside the catalog is exactly the source of the second peer variant, and I agree it is out of scope here. It is tracked in CS-12586 for the flat-config and eslint-major unification, and the `typescript-eslint` half of it is already fixed on a separate branch. Your point that this is what gives recommendation 2 its teeth is the reason I treated that one as worth doing now rather than deferring with it.

habdelra (approved):

PR #5850, opened 2026-08-21
ci: Change host shard weight measurement to use representative runs

backspace:
`test-module-timings.json` is used to balance host tests across shards, but it’s regenerated by runs on `main`, which differ timing-wise from PR runs because of Percy and index caching. This changes the balancing to use non-Percy PR runs when available.

<details><summary>Claude explanation</summary>`test-module-timings.json` is regenerated on every push to main and consumed by pull_request runs, and those two shapes differ by about 1.5x per test: Percy is always on for a non-PR event (~1.23x) and the index cache is deliberately skipped, so every realm indexes from scratch (the remaining ~1.21x, visible as `Import cached realm index` taking 0s on all 20 main shards against 27s on a PR run).

A uniform factor would not matter, since greedy packing is scale- invariant. This one is uneven — fixture-heavy acceptance modules pay far more of it than unit tests, per-module ratios running 0.86x to 3.40x — so the packer mis-assigns rather than mis-scales. On run 32406440696 it predicted a 1s spread across 20 shards and got 239s, with the slowest shard 19% over prediction. Regenerating from that run's own report drops the predicted slowest shard from 362s to 306s.

Take the report from the run of the PR that produced the merge, and only when that run was Percy-free, index-cached and fully green. Anything else leaves the file alone: stale-but-consistent weights balance better than fresh ones measured in the wrong environment, which is also why there is no fallback to this run's own report.

Over the last 25 PR-derived merges, 3 would have qualified — Percy is the binding filter, since most merged PRs touch UI. Weights move slowly and the existing drift gate already suppresses balance-equivalent rewrites, so roughly daily updates are enough; if that proves too sparse, the same check could scan back over recent merges for the newest qualifying run instead of only considering this one.</details>


.github/workflows/ci-host.yaml
```diff
@@ -1193,15 +1193,66 @@ jobs:
...
+          head_sha=$(gh api "repos/$REPO/pulls/$pr" --jq '.head.sha // empty')
+          [ -n "$head_sha" ] || leave "could not resolve the head commit of PR #$pr"
+
+          # Newest successful CI Host run for that head. A run with failed
+          # shards reports only the modules that ran, and the generator keeps
+          # existing values for the rest — which is how weights from different
+          # eras end up mixed in one file.
+          run_id=$(gh api "repos/$REPO/actions/runs?head_sha=$head_sha&event=pull_request&per_page=50" \
+            --jq '[.workflow_runs[] | select(.name == "CI Host" and .conclusion == "success")]
+                  | sort_by(.created_at) | last | .id // empty')
```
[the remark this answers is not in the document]
backspace: [Claude Code 🤖] Correct, and it would have made the feature a no-op rather than merely degraded — thanks. Fixed in b4e1d08f.

Added `pull-requests: read`, `actions: read`, and one the report doesn't name: `statuses: read`. The Percy check reads combined commit statuses, which is its own permission, so granting only the two named ones would have left that call still 404ing.

Two things fixed beyond the permissions, both pointed at by the "silently leaves shard timings unchanged" half of the report:

- A failed call and an absent result were the same `leave`, which is what would have made a missing permission invisible rather than merely wrong. A call that fails outright now emits a `::warning::` annotation instead, so the next instance of this class shows up in the run summary.
- The Percy check had a worse version of the same problem. It piped `gh api` straight into `grep`, and a pipeline takes grep's exit status — so a failed status read looked like "no percy context found" and took the *permissive* branch, treating a Percy run as Percy-free rather than skipping it. It now captures the statuses, checks the call succeeded, and greps after.

The artifact download deliberately keeps `leave`: with `actions: read` granted and the run listing already succeeding, a download that fails is far more likely to be a merged report the publish job never produced than a permissions problem.

.github/workflows/ci-host.yaml
```diff
@@ -1193,15 +1206,83 @@ jobs:
...
+          head_sha=$(gh api "repos/$REPO/pulls/$pr" --jq '.head.sha // empty') ||
+            broke "could not read PR #$pr"
+          [ -n "$head_sha" ] || leave "could not resolve the head commit of PR #$pr"
+
+          # Newest successful CI Host run for that head. A run with failed
+          # shards reports only the modules that ran, and the generator keeps
+          # existing values for the rest — which is how weights from different
+          # eras end up mixed in one file.
+          run_id=$(gh api "repos/$REPO/actions/runs?head_sha=$head_sha&event=pull_request&per_page=50" \
+            --jq '[.workflow_runs[] | select(.name == "CI Host" and .conclusion == "success")]
```
burieberry: [Claude Code 🤖] `conclusion == "success"` doesn't imply the merged report is complete, so a re-run PR can silently rewrite the weights from a partial suite.

The merge job downloads only `host-test-report-${{ github.run_attempt }}-*`, so on "re-run failed jobs" the merged artifact holds only the re-run shards — but the run's conclusion still flips to `success` and passes this filter. Run `33021863111` has **two** artifacts named `host-test-report-merged` (attempt 1: 121,401 B; attempt 2: 8,259 B), and `gh run download -R … --name host-test-report-merged` resolves to the attempt-2 one — 273 testcases instead of the full suite.

That matters because `MIN_COVERAGE` is computed against *the report's own* total time: a 273-case report is ~100% attributable and sails through, while every module absent from it keeps its old value. That's exactly the "weights from different eras mixed in one file" outcome the comment just above says this filter prevents.

Could this also require `run_attempt == 1`, or assert the report covers all 20 shards before handing it to the generator?

.github/workflows/ci-host.yaml
```diff
@@ -1193,15 +1206,83 @@ jobs:
...
+          # volume. `leave` is the ordinary one: this merge has no run worth
+          # measuring, which is most merges. `broke` is a call that failed
+          # outright — a missing permission, a transport error — which would
+          # otherwise be indistinguishable from "nothing qualifying" and would
+          # disable this for as long as it took someone to notice the file had
+          # stopped moving.
+          leave() { echo "$1 — leaving shard timings unchanged"; exit 0; }
+          broke() { echo "::warning title=Shard timings not refreshed::$1"; exit 0; }
+
+          pr=$(gh api "repos/$REPO/commits/$MERGE_SHA/pulls" --jq '.[0].number // empty') ||
```
burieberry: [Claude Code 🤖] `.[0].number` can resolve to an unrelated open PR.

`GET /commits/{sha}/pulls` returns the merged PR *and* any open PR whose history contains that commit — which happens as soon as someone merges main into their feature branch. The ordering is undocumented, so `.[0]` can be that open PR, and the job would then measure its head run and commit those weights as if they came from this merge.

`select(.merge_commit_sha == $MERGE_SHA)` pins it (that field is populated and matches on real merges), or `.state == "closed"` if you'd rather filter on state.

.github/workflows/ci-host.yaml
```diff
@@ -1193,15 +1206,83 @@ jobs:
...
+            broke "could not list workflow runs for $head_sha"
+          [ -n "$run_id" ] || leave "PR #$pr has no fully successful CI Host run"
+
+          # Percy roughly doubles some tests and leaves others alone, so a
+          # Percy-enabled PR run is no better a model of the median PR run
+          # than this push run is.
+          # Read the statuses first rather than piping straight into grep:
+          # a pipeline takes grep's exit status, so a failed call would look
+          # like "no percy context found" and take the permissive branch.
+          statuses=$(gh api "repos/$REPO/commits/$head_sha/status" --jq '.statuses[].context') ||
```
burieberry: [Claude Code 🤖] This gate reads a third-party commit status, page 1 only.

`GET /commits/{sha}/status` paginates `.statuses` at 30 by default and isn't paginated here; it also depends on Percy having posted by the time this job runs. Either miss silently takes the permissive branch and commits weights inflated by Percy — the exact failure this change exists to prevent, and one that leaves no trace in the log.

The run's own signal is already in hand: the jobs query below exposes `Finalise Percy` (`skipped` vs `success`), which is authoritative for that run, needs no extra API call, and would let the `statuses: read` permission go. Any reason to prefer the commit status over that?

(The `^percy/` grep itself is correct — the context is `percy/-cardstack-host`.)

burieberry (commented):
[Claude Code 🤖] Review of the shard-weight source change. Three inline findings below; two more that fall outside the diff:

**The coverage floor can now hard-fail this job** (`generate-test-module-timings.mjs` invocation, line 1326). The generator attributes module titles by scanning test sources **on disk** — post `git reset --hard origin/main` — while the junit report now comes from the PR head. If main deleted or renamed test modules after that head, their time is unattributable, and dropping below the 0.95 floor exits nonzero. That fails the step, which also blocks the memory-baseline commit that used to be independent of the timings refresh. Given the refresh is explicitly best-effort everywhere else in this step, should the generator invocation be non-fatal here too?

**`needs: host-merge-reports-and-publish` looks dead now** (line 1175). This job no longer consumes anything that job produces — the report comes from the PR's run instead. The dependency only delays the baseline update behind report merging and publishing. Worth dropping?

Checked and clean: `set -uo pipefail` with `x=$(…) || broke` (the assignment carries the substitution's status, and the handlers run in the parent shell so `exit 0` works); all four `// empty` jq guards against empty/missing results; `sort_by(.created_at) | last` on an empty array; `gh run download` with a single `--name` extracting to the `-D` root, so `merged-junit-report/host.xml` resolves; `per_page=100` on jobs and `per_page=50` on runs; the added `pull-requests`/`actions`/`statuses` scopes matching the endpoints used; and `imported > 0` being safe because all shards share one `if` condition.

.github/workflows/ci-host.yaml
backspace: [Claude Code 🤖] Confirmed and fixed in 86c9b704 — the run you cite does carry both artifacts (121,401 B and 8,259 B), and `--name` resolves to the 8 KB one while the run's conclusion reads success.

The filter now also requires `run_attempt == 1`, which is the version of "assert the report is whole" that needs no extra call: a run that has never been re-run has exactly one artifact of that name, so the download cannot pick up a partial. Asserting shard coverage inside the report would be the more direct check, but the merged junit carries no shard identity to count, and the coverage number the generator already computes is measured against the report's own total — which is exactly why it sails through, as you say.

The cost is qualification rate: a PR that only went green on a re-run no longer qualifies. That was already narrow (~12% of merges at the time I measured, mostly filtered by Percy), so this is worth watching — if the file stops moving for weeks, this filter is the first thing to loosen.

.github/workflows/ci-host.yaml
backspace: [Claude Code 🤖] Fixed in 86c9b704 with your first suggestion: `map(select(.merge_commit_sha == env.MERGE_SHA)) | .[0].number`.

Went with merge_commit_sha over `.state == "closed"` because closed-without-merge would still match on state, and a squash or rebase merge populates merge_commit_sha the same way — so this pins the PR that produced *this* commit rather than one that merely happens to be finished.

Checked against #5857's merge (03dcaa3d): resolves to 5857, and that endpoint returned only the one PR in that instance, so the bug was latent rather than firing — it needed an open branch carrying the commit, which is the case you describe.

.github/workflows/ci-host.yaml
backspace: [Claude Code 🤖] No reason to prefer it, and the commit status turns out to be demonstrably wrong sometimes rather than only theoretically. Switched in 86c9b704.

Looking for a case to test the two signals against each other, I found one where they disagree. Run 33083638586 (this repo, #5857's head) ran `percy exec --parallel -- pnpm ember-test-pre-built` with `PERCY_ENABLED: true` — Percy genuinely ran — and its head carried **no** `percy/` commit status at all when I queried it. So the old check would have classified that run Percy-free and let a Percy-inflated report set the weights. Not pagination in that instance, but the same silent-permissive outcome you predicted.

`Finalise Percy` reports `skipped` when Percy did not run and `success` when it did — verified on both a Percy-off run (32406440696) and a Percy-on one (32324478105), plus the disagreeing case above. It now comes out of the jobs fetch the index-cache check was already making, so it costs no extra call, and `statuses: read` is gone from the job's permissions.

One deliberate asymmetry: an answer that is neither `skipped` nor `success` — job absent, renamed, unreadable — is treated as "ran" and skips the refresh, since not measuring is the safe direction and measuring a Percy-loaded run is the failure this filter exists to prevent.

backspace:
[Claude Code 🤖] Both of the out-of-diff findings taken as well, in 86c9b704 — so all five are addressed. Threads above cover the three inline ones.

**Coverage floor hard-failing the job.** Yes, and worse than "the timings don't refresh": it would take the memory-baseline commit down with it, which has nothing to do with shard timings and used to be independent of them. The invocation is now `|| echo "::warning…"`, matching how the rest of the step already treats this refresh. The underlying mismatch stays — sources on disk are main's after the reset, the report is the PR head's — but it now degrades instead of failing, and says so in the run summary.

**`needs: host-merge-reports-and-publish`.** Dropped; `needs: [host-test]` is what remains, since the memory reports still come from its shards. Nothing else in the job reads that artifact any more.

Also merged main in, which brought `--shard-count 16` — the branch predated #5857 and would otherwise have reintroduced 20 into the drift gate.

Thanks for the "checked and clean" list — the `x=$(…) || broke` status question in particular was the thing I was least sure of when I wrote it.

habdelra (approved):

PR #5855, opened 2026-08-21
ci: Move Synapse start to beginning of workflow

backspace:
This is an iteration of #5854, stacked on #5850. Synapse takes long enough that starting in first might help.

<details><summary>Claude explanation</summary>Synapse takes about 67s to answer its healthcheck, and `create realm users` does nothing but wait for it: in a shard log, all twelve realm users queue at 18:34:33 and the first registration request goes out at 18:35:03, a 30s gap that is entirely Synapse coming up.

Starting it just before the index import (measured on #5854) hid only the ~24s the import takes, so the gate still waited 38s and setup did not move. Start it as early as the image allows instead — right after Traefik, before the asset download, skills-realm population, the remaining image warm and the import — so roughly the whole boot happens behind work the shard has to do anyway.

The warm action grows an `images` input so this can pull just the Synapse image and leave the other four to the existing step; `all` remains the default for every other caller.

Expected: `create realm users` falls toward the ~2s the registrations themselves take, without `Wait for realms` growing to absorb it. If the wait does grow, realm-server's own boot is the next wall rather than Synapse's.</details>

### DIFF
--- a/.github/actions/warm-test-images/action.yml
+++ b/.github/actions/warm-test-images/action.yml
@@ -13,6 +13,12 @@ inputs:
   registry-prefix:
     description: GHCR namespace holding the mirrored images.
     default: ghcr.io/cardstack/boxel
+  images:
+    description: >-
+      Which images to warm: "all", or a space-separated subset of synapse,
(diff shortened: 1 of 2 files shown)


backspace:
it didn’t help!

PR #5857, opened 2026-08-21
ci: Change host shard count to 16

backspace:
Twenty concurrent jobs per host run is a large share of the runner pool. During one congested window, 35 boxel workflow runs overlapped — five of them CI Host, so 100 shard jobs demanded at once, from four branches — and shards started staggered over eleven minutes.

Queueing is bimodal. Across eight recent CI Host runs, seven waited under a minute in total; the eighth had every shard waiting a median of 252s and burned 1.7 machine-hours idle. That tail is the thing worth cutting.

Measured on the same suite, no failures on either side:

| | 20 shards | 16 shards | 12 shards |
| -- | -- | -- | -- |
| slowest shard (gates the run) | 13.1m | 13.8m | 18.0m |
| machine time | 3.62h | 3.32h | 3.08h |

Sixteen takes the middle: most of the wall clock kept, a fifth of the slots returned to the pool. Twelve was measured too and costs nearly five minutes a run, which is too much for the same kind of relief.

Setup is ~195s per shard and largely irreducible — it is bound by the runner's cores rather than by latency, so parallelising or pre-computing it moves the cost rather than removing it. Each added shard therefore buys a shrinking slice of test time for a fixed ~3 minutes of machine time, and holds one more slot.

Two caveats worth carrying into the merge:

- If the concurrency cap is in the 40–60 range, five concurrent runs at 16 shards still exceed it. This reduces queueing rather than eliminating it.
- There is no same-window A/B. The 16-shard run happened to land in a congested window while the 20-shard runs did not, so the queue improvement is inferred from demand arithmetic, not measured. What would settle it is queue-delay distributions across a few weeks, split before and after.

Most of the benefit lands on whatever else is queued rather than on the run that gives up the slots, so it does not show up in this PR's own numbers.

Follow-up, not in this PR: the weights are still packed for 20, which shows as a wider spread across 16 shards (216s vs 182s). Rebalancing touches the same file as the shard-weight work in #5850.

### DIFF
--- a/.github/workflows/ci-host.yaml
+++ b/.github/workflows/ci-host.yaml
@@ -459,31 +459,25 @@ jobs:
     needs: [test-web-assets, check-percy, check-index-cache]
     strategy:
       fail-fast: false
+      # Three places carry the shard count and must move together: this list,
+      # `shardTotal` below, and `--shard-count` on the timings generator, which
+      # decides whether a rebalance is worth committing by predicting the
+      # slowest shard.
+      #
+      # Fewer, longer shards rather than more, shorter ones. Setup is ~195s per
+      # shard and largely irreducible (it is bound by the runner's cores, not
+      # by latency — see the four experiments on CS-12583), so each added shard
+      # buys a shrinking slice of test time for a fixed ~3 minutes of machine
+      # time and one more slot held. With ~8,665s of test work: 20 shards is
+      # ~10.5 minutes and 3.49 machine-hours, 12 is ~15.3 minutes and 3.06.
+      #
+      # The modelled wall-clock cost overstates the real one, because it
+      # assumes every shard starts immediately. They do not: shards have been
+      # observed queueing up to 242s waiting for a runner, and 20 concurrent
+      # jobs per host run is a large share of the pool.
       matrix:
-        shardIndex:
-          [
-            1,
-            2,
-            3,
-            4,
-            5,
-            6,
-            7,
-            8,
-            9,
-            10,
-            11,
-            12,
-            13,
-            14,
-            15,
-            16,
-            17,
-            18,
-            19,
-            20,
-          ]
-        shardTotal: [20]
+        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+        shardTotal: [12]
     concurrency:
       group: boxel-host-test${{ github.head_ref || github.run_id }}-shard${{ matrix.shardIndex }}
       cancel-in-progress: true
@@ -1243,7 +1237,7 @@ jobs:
             # job loudly rather than committing a degraded weights file.
             if [ -f merged-junit-report/host.xml ]; then
               node packages/host/scripts/generate-test-module-timings.mjs merged-junit-report/host.xml \
-                --min-drift-seconds 60 --shard-count 20
+                --min-drift-seconds 60 --shard-count 12
             else
               echo "No merged junit report available — leaving shard timings unchanged."
             fi
--- a/.github/workflows/ci-host.yaml
+++ b/.github/workflows/ci-host.yaml
@@ -468,16 +468,18 @@ jobs:
       # shard and largely irreducible (it is bound by the runner's cores, not
       # by latency — see the four experiments on CS-12583), so each added shard
       # buys a shrinking slice of test time for a fixed ~3 minutes of machine
-      # time and one more slot held. With ~8,665s of test work: 20 shards is
-      # ~10.5 minutes and 3.49 machine-hours, 12 is ~15.3 minutes and 3.06.
+      # time and one more slot held. Measured on ~8,500s of test work: at 20
+      # shards the slowest shard took 13.1 minutes for 3.62 machine-hours, at
+      # 12 it took 18.0 for 3.08. Sixteen sits between them, keeping most of
+      # the wall clock while returning a fifth of the slots to the pool.
       #
       # The modelled wall-clock cost overstates the real one, because it
       # assumes every shard starts immediately. They do not: shards have been
       # observed queueing up to 242s waiting for a runner, and 20 concurrent
       # jobs per host run is a large share of the pool.
       matrix:
-        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-host.yaml
```diff
@@ -459,31 +459,27 @@ jobs:
     needs: [test-web-assets, check-percy, check-index-cache]
     strategy:
       fail-fast: false
+      # Three places carry the shard count and must move together: this list,
+      # `shardTotal` below, and `--shard-count` on the timings generator, which
+      # decides whether a rebalance is worth committing by predicting the
+      # slowest shard.
```
richardhjtan, line 465: [Claude Code 🤖] Four places, not three: `packages/host/scripts/generate-test-module-timings.mjs` still has `let shardCount = 20`, and its option doc reads "default 20, matching ci-host.yaml", which this change makes false.

Worth stating the blast radius so the fix isn't over-scoped: `shardCount` only reaches `pack()` and from there `slowestShardSeconds`, so it decides whether the weights file is rewritten, never its contents, and `minDriftSeconds` defaults to 0 which skips that gate entirely. It bites a local run that passes `--min-drift-seconds` without `--shard-count` — reachable because the script's own header example omits the flag.

Cheapest fix that can't drift again: drop the default and exit non-zero when `--shard-count` is absent, leaving this workflow the only place that names the number. Failing that, move both the default and the doc line to 16 and make this comment say four.

Regression, non-blocking.

richardhjtan (commented):
[Claude Code 🤖]

This review went after the places the shard count is duplicated into, the downstream consumers of shard identity, and whether the longer shards show up as memory pressure. It did not re-derive the time-versus-slots tradeoff itself.

**No blocking code issues. The thing to fix before merge is the title: the branch ships 16 shards, and the title and description still say 12.** `squash_merge_commit_title` on this repo is `COMMIT_OR_PR_TITLE`, so squashing these two commits records "change host shard count to 12" against a file that says 16. The description argues 12 against 20 and never reaches 16, so the reasoning for landing between them exists only in the workflow comment.

Recommendations:

1. Retitle to 16, and put enough of the 16-versus-12 reasoning in the description that history explains itself.
2. The timings generator's `shardCount` default and its option doc are a fourth site carrying the count — see the thread on the `strategy:` comment.
3. Drop the tracker reference from the comment — see the thread on the setup-cost paragraph.
4. `.claude/skills/host-test-memory-leak-hunting/SKILL.md` carries the count too and this PR doesn't move it. Line 129 pins the leak-validation recipe to "the original `shardTotal: 20` config", naming shard 3 of 20 as the memory canary — a shard that no longer exists — and line 131 tells a leak-fixer to revert shard-count bumps, which now reads as pointing back at 20. It is the only remaining copy that instructs someone to reproduce against the old count, so it belongs in this PR. Re-picking the canary needs a measurement, not a renumber: from this run's `memory-report-*` artifacts the worst shard peaked at 463 MB used heap, against 664 MB on a 20-shard main run today, so peak heap tracks which modules land together rather than how long the shard is.

Adjacent, out of scope: nothing in CI gates per-shard peak heap — `check-memory-baseline.mjs` compares per-module deltas only, merged across shards — which is why the question above can only be answered by pulling artifacts by hand.

Red checks: none, host CI is green on the head commit.

.github/workflows/ci-host.yaml
```diff
@@ -459,31 +459,27 @@ jobs:
...
+      #
+      # Fewer, longer shards rather than more, shorter ones. Setup is ~195s per
+      # shard and largely irreducible (it is bound by the runner's cores, not
+      # by latency — see the four experiments on CS-12583), so each added shard
```
richardhjtan: [Claude Code 🤖] Please drop the tracker reference on this line. Per the repo's evergreen-comments contract a code comment carries the fact rather than the ticket that produced it, and a reader outside the tracker can't follow it. "bound by the runner's cores, not by latency" is the load-bearing half and stands on its own.

Regression, non-blocking.

.github/workflows/ci-host.yaml
backspace: [Claude Code 🤖] Took the drop-the-default option. `shardCount` is only read by the drift prediction, so it is now required with `--min-drift-seconds` and absent-by-default otherwise — a plain regeneration needs no count, and a gate run without one exits non-zero rather than predicting against a shard count nobody runs. That leaves `ci-host.yaml` as the only place naming the number, so the comment still says three.

Verified all three paths against a real report: gate without the flag errors, plain regeneration writes, gate with `--shard-count 16` gates. In 5ebf702d.

.github/workflows/ci-host.yaml
backspace: [Claude Code 🤖] Dropped, in 5ebf702d. The paragraph now states the mechanism on its own — setup is bound by the runner’s cores rather than by latency, so parallelising or pre-computing it moves the cost rather than removing it — with no ticket reference.

backspace:
[Claude Code 🤖] All four addressed — two in the threads above, two here.

**1. Title and description.** Retitled to 16, and the description now argues 16 rather than 12: the congested-window evidence (35 overlapping runs, five of them CI Host, so 100 shard jobs demanded at once from four branches), the three-way measurement, and why 16 over 12. Both caveats are stated in it too — that five runs at 16 shards likely still exceed the cap, so this reduces queueing rather than removing it, and that there is no same-window A/B, so the queue improvement is inferred from demand arithmetic rather than measured.

**4. The memory-leak skill.** Updated in 5ebf702d, and not by renumbering — your measurement is the reason. 463 MB worst shard at 16 against 664 MB on a 20-shard main run says peak heap tracks which modules a shard is dealt rather than how long it runs, so a fixed canary is the wrong shape of instruction at any count. The recipe now says to take the highest peak from that run's own `memory-report-*` artifacts, and records the gap you noted as adjacent: nothing in CI gates per-shard peak heap, since `check-memory-baseline.mjs` compares per-module deltas merged across shards, which is why this has to be pulled by hand.

I left that gap as a note rather than closing it here — it is a new capability rather than a fix to this change. Worth its own issue if it should be tracked somewhere more durable than a skill file.

lukemelia (approved):

richardhjtan (approved):

PR #5858, opened 2026-08-23
Export content-only FileDef preview components for markdown, image, and audio

lukemelia:
## What this does

The \`file-formats/index\` barrel now exports **MarkdownPreview**, **ImagePreview**, and **AudioPreview** so card authors can render a file's content with no shell chrome — no file bar, no inspector metadata, no Download/Copy-link bar. The default wrapped FileDef templates are unchanged; authors wanting full control import a content-only component instead of fighting the shell with CSS.

## The settled component signature

The three components share \`ContentPreviewSignature\` (exported from the barrel alongside \`FilePreviewSignature\`):

- **\`@model\`** — the FileDef instance itself, the common case: the component projects it through the new \`ensureFileViewModel\` helper, which also picks up the profile axes the instance's class pins (\`static fileFamily\` and friends). A prebuilt \`FileViewModel\` (what the shells pass) is also accepted; \`isFileViewModel\` tells the two apart by the projection's plain-object shape.
- **\`@mode\`** — optional, defaulting to \`embedded\` (complete content). \`fitted\` selects the budgeted snippet rendition a collection cell draws.
- **\`@fields\`** — optional, only for renderers that delegate to a declared field's own component.

Loading/failure/staleness panes are deliberately *not* drawn by these components — inside the default templates that is \`FilePreviewStage\`'s job, and a content-only embedding author owns those states.

\`filePreviewComponentFor\` is also exported from the barrel, so a kind-dispatching consumer can resolve whichever renderer a file's own class declares.

## Where MarkdownPreview lives

Rather than exporting it from \`markdown-file-def.gts\`, the component moves to \`file-formats/markdown-preview.gts\`, matching the image/audio/html renderers. This keeps the barrel from pulling MarkdownDef's whole extraction graph (frontmatter parsing, file-api, card-api statics) into every barrel consumer, and it keeps card-api's guarded dependency-graph fixture in \`realm-indexing-test\` untouched — nothing new becomes reachable from \`card-api\`.

## Tests

New \`Integration | content-only file preview components\` module covers: importing each component (and the helpers) from the barrel via the loader, markdown/image/audio rendering from a bare FileDef instance with explicit no-shell-chrome assertions, the prebuilt-view-model path, the \`embedded\` mode default, and \`filePreviewComponentFor\` dispatch.

Ran locally via the dev test page: the new module plus \`FileDef format templates\` (68), \`markdown-preview\` (19), \`json/csv file def preview\` (39), \`markdown skill *\` (53), and \`realm indexing\` (145) — all green. Typechecks pass in host, ai-bot, bot-runner, and billing.

Linear: [CS-12587](https://linear.app/cardstack/issue/CS-12587)

### DIFF
--- a/packages/base/file-formats/file-preview-stage.gts
+++ b/packages/base/file-formats/file-preview-stage.gts
@@ -45,7 +45,12 @@ export type FilePreviewComponent = ComponentLike<FilePreviewSignature>;
 // itself in the common case (the component projects it through
 // `fileViewModel`), or a prebuilt FileViewModel (what the shells pass). `mode`
 // defaults to 'embedded', the complete-content rendition; 'fitted' selects the
-// budgeted snippet a collection cell draws. These components render content
+// budgeted snippet a collection cell draws. A prebuilt view model must have
+// been projected at the same format passed as `mode`: the fitted budgets are
+// applied at projection time, so a complete-content projection rendered at
+// 'fitted' would feed the whole file to the snippet branch. A consumer that
+// varies `mode` should pass the instance and let the component re-project.
+// These components render content
 // only — the loading/failure/staleness panes are `FilePreviewStage`'s job
 // inside the shells, so an embedding author owns those states.
 export interface ContentPreviewSignature {
--- a/packages/base/file-formats/file-preview-stage.gts
+++ b/packages/base/file-formats/file-preview-stage.gts
@@ -14,7 +14,11 @@ import { eq } from '@cardstack/boxel-ui/helpers';
 import { Warning } from '@cardstack/boxel-ui/icons';
 
 import { downloadNameFor, fileIconFor, humanSize } from './file-presentation';
-import type { FileFormat, FileViewModel } from './file-view-model';
+import type {
+  FileFormat,
+  FileModelLike,
+  FileViewModel,
+} from './file-view-model';
 
 import type { ComponentLike } from '@glint/template';
 
@@ -33,6 +37,25 @@ export interface FilePreviewSignature {
 }
 
 export type FilePreviewComponent = ComponentLike<FilePreviewSignature>;
+
+// The content-only face of a family renderer, for the components the
+// `file-formats/index` barrel exports (MarkdownPreview, ImagePreview,
+// AudioPreview): just the file's content, none of the shell chrome — no file
+// bar, no inspector, no Download/Copy-link. `model` is the FileDef instance
+// itself in the common case (the component projects it through
+// `fileViewModel`), or a prebuilt FileViewModel (what the shells pass). `mode`
+// defaults to 'embedded', the complete-content rendition; 'fitted' selects the
+// budgeted snippet a collection cell draws. These components render content
+// only — the loading/failure/staleness panes are `FilePreviewStage`'s job
+// inside the shells, so an embedding author owns those states.
+export interface ContentPreviewSignature {
+  Args: {
+    model: FileViewModel | FileModelLike | undefined | null;
+    mode?: FileFormat;
+    fields?: Record<string, any>;
+  };
+  Element: HTMLElement;
+}
 
 // The renderer a file's own class declares, if its family has landed one. Read
 // from the instance's constructor rather than the declared field class, since a
--- a/packages/base/file-formats/index.ts
+++ b/packages/base/file-formats/index.ts
@@ -26,8 +26,10 @@ export {
   FITTED_TEXT_CHARACTER_BUDGET,
   FITTED_TEXT_LINE_BUDGET,
   FITTED_WAVEFORM_BAR_BUDGET,
+  ensureFileViewModel,
   fileProfileSource,
   fileViewModel,
+  isFileViewModel,
   type FileFormat,
   type FileModelLike,
   type FileProfileSource,
@@ -48,11 +50,23 @@ export {
 
 export {
   FilePreviewStage,
+  filePreviewComponentFor,
+  type ContentPreviewSignature,
   type FilePreviewComponent,
   type FilePreviewSignature,
 } from './file-preview-stage';
 
+// The content-only family renderers: just the file's content, none of the
+// shell chrome (no file bar, no inspector, no Download/Copy-link). Pass the
+// FileDef instance as `@model` — or a prebuilt view model — and optionally a
+// `@mode`, which defaults to 'embedded' (see `ContentPreviewSignature`).
+// Loading/failure/staleness treatment belongs to the embedding author; inside
+// the default templates it is `FilePreviewStage`'s job. For kind-dispatching
+// consumers, `filePreviewComponentFor(file)` resolves the renderer the file's
+// own class declares.
+export { AudioPreview } from './audio-preview';
 export { ImagePreview } from './image-preview';
+export { MarkdownPreview } from './markdown-preview';
 
 // The compound metadata shapes an extractor writes into. Shared per metadata
 // family rather than per file extension, so a camera or a color profile reads
--- a/packages/base/file-formats/index.ts
+++ b/packages/base/file-formats/index.ts
@@ -63,7 +63,10 @@ export {
 // Loading/failure/staleness treatment belongs to the embedding author; inside
 // the default templates it is `FilePreviewStage`'s job. For kind-dispatching
 // consumers, `filePreviewComponentFor(file)` resolves the renderer the file's
-// own class declares.
+// own class declares — pass it `ensureFileViewModel(file, mode)` as `@model`,
+// since only the content-only components exported here project a bare
+// instance themselves; the other family renderers read projection-only fields
+// (the fitted budgets among them) straight off `@model`.
 export { AudioPreview } from './audio-preview';
 export { ImagePreview } from './image-preview';
 export { MarkdownPreview } from './markdown-preview';
--- a/packages/base/file-formats/markdown-preview.gts
+++ b/packages/base/file-formats/markdown-preview.gts
@@ -99,8 +99,8 @@ export class MarkdownPreview extends GlimmerComponent<ContentPreviewSignature> {
         height: 100%;
         min-height: 0;
         overflow: auto;
-        background: var(--card);
-        color: var(--foreground);
+        background: var(--md-preview-background, var(--card));
+        color: var(--md-preview-foreground, var(--card-foreground));
         text-align: left;
       }
       .md-preview--full {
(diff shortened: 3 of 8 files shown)


packages/base/file-formats/index.ts
```diff
@@ -48,11 +50,23 @@ export {
...
 } from './file-preview-stage';
 
+// The content-only family renderers: just the file's content, none of the
+// shell chrome (no file bar, no inspector, no Download/Copy-link). Pass the
+// FileDef instance as `@model` — or a prebuilt view model — and optionally a
+// `@mode`, which defaults to 'embedded' (see `ContentPreviewSignature`).
+// Loading/failure/staleness treatment belongs to the embedding author; inside
+// the default templates it is `FilePreviewStage`'s job. For kind-dispatching
+// consumers, `filePreviewComponentFor(file)` resolves the renderer the file's
+// own class declares.
```
lukemelia: [Claude Code 🤖] **The kind-dispatch sentence invites passing a bare FileDef instance to renderers that still require the projection — which silently bypasses the fitted work budget.** Non-blocking; the fix is a comment edit, but it belongs in this change since the guidance is introduced here.

**The mechanism.** `filePreviewComponentFor` returns whatever `static previewComponent` the file's class declares, and only the three components exported above (`MarkdownPreview`, `ImagePreview`, `AudioPreview`) project their `@model` through `ensureFileViewModel`. The other family renderers — `TextPreview` (`text-file-def.gts`), `JsonPreview`, `CsvPreview`, `CodePreview`, `ArchivePreview`, `HtmlPreview`, `PdfViewer`, `OfficePreview`, `FontSpecimen`, `Model3DPreview`, `MidiPreview` — still have `FilePreviewSignature` and read projection fields off `@model` directly.

**What actually goes wrong.** `TextPreview` reads `this.args.model?.previewTruncated` (a field that exists only on the projection) and `this.args.model?.content`. Handed a bare instance, the truncation signal is silently absent — and in fitted mode `content` is the complete file, so the invariant `file-view-model.ts`'s header states ("a collection cell must never turn an unbounded payload into HTML") is bypassed with no error, just an unbounded render. Glint catches this for typed consumers (`FilePreviewComponent` requires `model: FileViewModel`), but card code is routinely loosely typed.

**The fix.** Make the dispatch sentence carry the projection requirement, e.g.: "For kind-dispatching consumers, `filePreviewComponentFor(file)` resolves the renderer the file's own class declares — pass it `ensureFileViewModel(file, mode)` as `@model`, since only the content-only components exported here project a bare instance themselves."

Scope: regression (new API guidance), non-blocking.

packages/base/file-formats/file-preview-stage.gts
```diff
@@ -34,6 +38,25 @@ export interface FilePreviewSignature {
...
+// only — the loading/failure/staleness panes are `FilePreviewStage`'s job
+// inside the shells, so an embedding author owns those states.
+export interface ContentPreviewSignature {
+  Args: {
+    model: FileViewModel | FileModelLike | undefined | null;
+    mode?: FileFormat;
+    fields?: Record<string, any>;
+  };
+  Element: HTMLElement;
+}
```
lukemelia, line 63: [Claude Code 🤖] **The prebuilt-view-model arm accepts a projection whose format disagrees with `@mode`, and in fitted mode that re-opens the unbounded-render hole the projection-time budgets exist to close.** Non-blocking — unreachable from the shells — but worth pinning while the surface is new.

**The mechanism.** `ensureFileViewModel` returns a prebuilt `FileViewModel` unchanged; its `format` parameter applies only to the bare-instance arm. A projection built at a complete-content format carries the whole file in `contentPreview` — `fileViewModel` sets `contentPreview: content`, and `content` is only budget-truncated when the projection's own format is `'fitted'`. So `<MarkdownPreview @model={{viewModel}} @mode='fitted' />` with an embedded- or isolated-built view model takes the fitted branch and feeds the complete file to `markdownToHtml`.

**Why the shells are safe.** Each format template projects at its own mode and passes the matching `@mode` (`default-templates/file-def-fitted.gts` builds `fileViewModel(model, 'fitted', …)`), so the mismatch is reachable only through this content-only surface — whose signature comment doesn't yet state the must-match invariant.

**The fix, two options.**
- *(a) Minimal:* add one sentence to this comment — a prebuilt view model must have been projected at the same format passed as `@mode`; a consumer that varies mode should pass the instance (or re-project via `ensureFileViewModel(file, mode)`).
- *(b) Structural:* stamp the projection with the format it was built at (a `format` field on `FileViewModel`) and have `ensureFileViewModel` re-project from `model.source` on mismatch. That closes the hole mechanically, at the cost of discarding any hand-tuned fields on the passed projection.

I'd take (a) here; (b) is a follow-up if mismatches show up in real embeddings.

Scope: regression (new surface), non-blocking. Related: the comment on the kind-dispatch guidance in `file-formats/index.ts` covers the other sharp edge of the same contract.

packages/host/tests/integration/components/content-preview-components-test.gts
```diff
@@ -0,0 +1,214 @@
...
+    let file = makeMarkdownFile();
+    await renderComponent(
+      <template><MarkdownPreview @model={{file}} @mode='fitted' /></template>,
+    );
+    assert
+      .dom('[data-test-markdown-preview]')
+      .hasClass('md-preview--fitted', 'the fitted rendition is selected');
+    assert.dom('[data-test-markdown-preview] h1').hasText('Field Notes');
+  });
+
```
lukemelia, line 181: [Claude Code 🤖] **This test passes even if the `@mode` → projection plumbing were broken, so it doesn't pin the one behavior unique to the bare-instance fitted path.** Non-blocking, test-coverage.

**Why it can't fail for the right reason.** The fixture's content is three lines — far inside `FITTED_TEXT_CHARACTER_BUDGET`/`FITTED_TEXT_LINE_BUDGET` — so `contentPreview` equals the complete content whether `ensureFileViewModel` projects at `'fitted'` or at its `'embedded'` default. And the `md-preview--fitted` class assertion only pins branch selection, which `isFitted` derives from `@mode` directly, upstream of the projection. If `ensureFileViewModel` dropped its `format` argument entirely, every assertion here would still pass.

**The fix.** Give the fixture more lines than `FITTED_TEXT_LINE_BUDGET` (14) and assert that a line beyond the budget is absent from the snippet render. That makes this test the executable guard that `@mode` flows through `ensureFileViewModel` into the fitted budget — which is exactly what the test's name claims.

lukemelia (commented):
[Claude Code 🤖] This review focused on the contracts of the new public surface — what each arm of the content-only `@model` accepts, whether the projection-time fitted budget survives the new entry points, and whether the barrel's dependency-weight rationale holds against the guarded dependency fixture.

**Bottom line: no blocking issues.** The shell-path behavior of the three converted renderers is unchanged, the dependency claims check out against the guarded card-dependency fixture (`card-api`'s graph gains nothing; the barrel already reached `card-api` via `metadata-fields`, and the new value imports land on modules already in the guard list), and the findings below are comment-level sharpenings of the new surface plus one test-coverage gap.

Recommendations:
1. Sharpen the barrel's kind-dispatch guidance — a dispatched renderer needs a prebuilt projection, and the adjacent prose currently suggests otherwise (see the comment on `file-formats/index.ts`).
2. State — or mechanically enforce — the invariant that a prebuilt view model was projected at the same format passed as `@mode` (see the comment on `ContentPreviewSignature` in `file-preview-stage.gts`).
3. Make the fitted test able to detect the `@mode` → projection-budget plumbing breaking (see the comment in `content-preview-components-test.gts`).

Adjacent nit, take or leave: the reflowed header comment in `audio-preview.gts` leaves the word `visual` alone on its own line mid-sentence — worth a rewrap while in the file.

packages/base/file-formats/index.ts
lukemelia: [Claude Code 🤖] Fixed in a52b25f9b6. The dispatch guidance now tells consumers to pass `ensureFileViewModel(file, mode)` as `@model`, and states why: only the content-only components exported from the barrel project a bare instance themselves — the other family renderers read projection-only fields (the fitted budgets among them) straight off `@model`.

packages/base/file-formats/file-preview-stage.gts
lukemelia: [Claude Code 🤖] Fixed in a52b25f9b6, taking option (a). The `ContentPreviewSignature` comment now states the invariant: a prebuilt view model must have been projected at the same format passed as `mode`, since the fitted budgets are applied at projection time — and a consumer that varies `mode` should pass the instance and let the component re-project. The structural fix (stamping the projection with its format and re-projecting on mismatch) remains a follow-up if mismatches show up in real embeddings.

packages/host/tests/integration/components/content-preview-components-test.gts
lukemelia: [Claude Code 🤖] Fixed in a52b25f9b6. The fitted test now builds a fixture with `FITTED_TEXT_LINE_BUDGET + 6` list items under a heading and asserts both that the last line inside the budget renders and that the first line beyond it is absent — so the test fails if `@mode` stops flowing through `ensureFileViewModel` into the projection-time budget. Ran the module on the dev test page: 51/51 assertions pass.

jurgenwerk (approved):

packages/base/file-formats/markdown-preview.gts
```diff
@@ -0,0 +1,171 @@
...
+      </div>
+    {{/if}}
+    <style scoped>
+      .md-preview {
+        width: 100%;
+        height: 100%;
+        min-height: 0;
+        overflow: auto;
+        background: var(--card);
+        color: var(--foreground);
```
burieberry: if we must provide these, then:
```
background-color: var(--md-preview-background, var(--card));
color: var(--md-preview-foreground, var(--card-foreground));
```
If this component is always rendered inside a certain styled container, then we wouldn't need to set these and it would be inherited from the styled container.

burieberry (approved):

packages/base/file-formats/markdown-preview.gts
lukemelia: [Claude Code 🤖] Done in a3c54c8f97 — took the overridable form:

```css
background-color: var(--md-preview-background, var(--card));
color: var(--md-preview-foreground, var(--card-foreground));
```

Kept a default rather than relying on inheritance: this is a content-only, no-shell component meant to be dropped into arbitrary author-controlled containers, so there is no guaranteed styled ancestor to inherit from — the fallback gives it a sane surface on its own while the two hooks let an author retint it without fighting the rule.

Also took the token pairing you implied: switched the foreground fallback from `--foreground` to `--card-foreground`, the semantic partner of the `--card` surface.

PR #5861, opened 2026-08-24
Auto-remove unused imports in the realm lint pipeline

jurgenwerk:
Generated card code routinely arrives with unused imports: the import block streams before the template, so plans drift and leftovers like `FieldDef` or `containsMany` are common. The realm lint pipeline reported them via `@typescript-eslint/no-unused-vars` — an error with no autofix — so the assistant spent a full patch round deleting a few lines it had just written. The pipeline already rewrites imports in the other direction (injecting missing ones, deduping duplicates); this closes the asymmetry.

- Enables `unused-imports/no-unused-imports` (autofixable, via `eslint-plugin-unused-imports@^3`, compatible with ESLint 8) in the realm override block of `runtime-common/tasks/lint.ts` only — host source is untouched, so its authors keep seeing the error in their editors.
- Unused local variables stay reported as before; only imports are auto-removed. If a usage appears later, `missing-card-api-import` re-injects the import.
- Endpoint tests pin the three behaviors: unused imports (named, type-only, default) are removed with zero residual messages; an unused local variable still reports; an import used only inside `<template>` survives the autofix.


packages/runtime-common/tasks/lint.ts
```diff
@@ -135,6 +135,13 @@ async function initESLint(): Promise<any> {
           { importMappings },
         ],
         '@cardstack/boxel/no-duplicate-imports': 'error',
+        // Generated card code routinely carries unused imports (the import
+        // block streams before the template, so plans drift). Removing them
+        // is safe to automate: if a usage appears later, the
+        // missing-card-api-import rule re-injects the import. Unused
+        // *variables* stay reported via @typescript-eslint/no-unused-vars —
+        // deleting those would change behavior.
+        'unused-imports/no-unused-imports': 'error',
```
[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Fixed in 2391b9efc6. The third-party plugin is replaced by `@cardstack/boxel/no-unused-imports`, whose fix never discards module evaluation: a fully-unused value import of an unknown module is rewritten to a bare side-effect import (`import './register';`), and outright deletion is reserved for type-only imports and a safelist of side-effect-free platform module namespaces (`@cardstack/base/*` etc.), where the inject rule can re-create the import if a usage appears. Covered by plugin unit tests and a lint-endpoint test.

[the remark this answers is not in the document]
jurgenwerk: [Claude Code 🤖] Fixed in 2391b9efc6. Automatic deletion is now restricted exactly as suggested: unused specifiers are removed individually while the declaration keeps used ones; a declaration whose bindings are all unused is deleted only when type-only or when its module matches a side-effect-free safelist of platform namespaces; every other module is rewritten to a bindingless side-effect import so its top-level code still runs. The default-import test now pins the side-effect-preserving rewrite instead of deletion.

lukemelia (approved):

packages/eslint-plugin-boxel/lib/rules/no-unused-imports.js
```diff
@@ -0,0 +1,206 @@
...
+          named.push(sourceCode.getText(spec));
+        }
+      }
+      if (named.length > 0) {
+        parts.push(`{ ${named.join(', ')} }`);
+      }
+      const typePrefix = node.importKind === 'type' ? ' type' : '';
+      return `import${typePrefix} ${parts.join(', ')} from ${sourceCode.getText(
+        node.source,
+      )};`;
```
habdelra: `rebuildDeclaration` reconstructs the statement from specifiers + source text only, so anything else attached to the declaration is silently dropped — most consequentially **import attributes**:

```js
// before
import data, { thing } from './x.json' with { type: 'json' };
// after autofix (`data` unused)
import { thing } from './x.json';
```

That's not a cosmetic loss — the resulting import throws at runtime in engines that require the attribute for non-JS module types. The bare-import rewrite at line 196 has the same gap.

Fix is to carry `node.attributes` (ESTree `ImportAttribute[]`) through both rewrite paths, or — simpler and more robust — build the fix by removing only the unused specifier ranges rather than re-emitting the whole declaration. The removal approach also preserves comments inside the specifier list, which the current code erases:

```js
import {
  used,
  // keep this around, we'll need it for X
  unused,
} from 'm';
```

---

packages/eslint-plugin-boxel/lib/rules/no-unused-imports.js
```diff
@@ -0,0 +1,206 @@
...
+    function removeWithNewline(fixer, node) {
+      const text = sourceCode.getText();
+      let end = node.range[1];
+      if (text[end] === '\r' && text[end + 1] === '\n') {
+        end += 2;
+      } else if (text[end] === '\n') {
+        end += 1;
+      }
+      return fixer.removeRange([node.range[0], end]);
+    }
```
habdelra, line 262: `removeWithNewline` extends the removal range forward through the trailing newline, but not backward over any leading comment. When the deleted import had a directive comment above it, that comment survives and silently re-targets the next statement:

```js
// eslint-disable-next-line some-rule
import { Unused } from 'm';
import { Used } from 'n';       // <- now the suppression lands here
```

Same hazard with `// @ts-ignore` and `// @ts-expect-error` (the latter flips to an error when it stops matching, which at least fails loudly — `@ts-ignore` and `eslint-disable-next-line` fail silently).

This isn't hypothetical for this repo — `packages/base/default-templates/missing-template.gts:3-4` is exactly this shape, and it's the one file in `packages/base` + `packages/experiments-realm` the rule currently reports (unused `cssUrl`).

Suggest extending the range back over comments that are directly attached (`sourceCode.getCommentsBefore(node)`, taking only those separated from the import by a single newline), so the comment goes with the statement it annotates.

---

packages/runtime-common/tasks/lint.ts
```diff
@@ -135,6 +135,26 @@ async function initESLint(): Promise<any> {
...
+        '@cardstack/boxel/no-unused-imports': [
+          'error',
+          {
+            sideEffectFreeModules: [
+              '@cardstack/base/*',
+              'https://cardstack.com/base/*',
+              '@cardstack/boxel-ui/*',
+              '@cardstack/boxel-icons/*',
+            ],
+          },
```
habdelra, line 156: This deleting autofix can deadlock against `missing-card-api-import`'s inserting autofix, and the safelist below is what makes it reachable.

`missing-card-api-import` keys off bare identifiers, so a **class member** named after a card-API export is enough to trigger it — `contains`, `field`, `Component`, `DateField`, and friends are all plausible member names on a card:

```gts
export class Foo extends CardDef {
  // no import of `contains` anywhere
  contains(other: Foo) { ... }
}
```

Round trip per fix pass:
1. `missing-card-api-import` injects `import { contains } from '@cardstack/base/card-api';`
2. the injected binding has no *references* (the member name isn't one), so `no-unused-imports` sees it as unused
3. `@cardstack/base/*` is in `sideEffectFreeModules`, so the whole declaration is deleted rather than reduced to a bare import
4. back to step 1

ESLint burns all 10 `--fix` passes and returns the file with the error still present, so `/_lint` responds `passed: false` carrying a diagnostic that no amount of re-linting can clear — the exact failure mode this PR is trying to remove for generated code.

Cheapest guard: have `no-unused-imports` skip bindings whose declaration was inserted by `missing-card-api-import` in the same run, or exclude the card-API module from `sideEffectFreeModules` so step 3 degrades to a bare side-effect import and the loop terminates. Either way, worth an endpoint test pinning "unused-import fix + missing-import fix converge" alongside the three existing ones.

---

habdelra (commented):
Reviewed at `2391b9ef` against merge base `f7e3fda4f`. I built a sandbox with eslint 8.57 + `ember-eslint-parser` and executed the rule, so the three inline findings are reproduced rather than inferred.

**What came back clean**

- The rule's own unit tests pass standalone.
- Scanning all 329 `.gts`/`.gjs` files in `packages/base` + `packages/experiments-realm` produced exactly **one** report, and it's a true positive (`cssUrl` in `packages/base/default-templates/missing-template.gts`). No false-positive removals on templates, type positions, decorators, re-exports, or nested `<template>` bodies — the `Program:exit` timing does its job.
- Auto-registration via `requireindex` works, and `recommended: false` correctly means `recommended-rules.js` needs no update.

**Findings** (three inline)

1. Import attributes (`with { type: 'json' }`) are silently dropped by both rewrite paths — produces a runtime-breaking import.
2. `removeWithNewline` orphans a leading `// @ts-ignore` / `// eslint-disable-next-line` onto the following statement. Live example in `missing-template.gts:3-4`.
3. Autofix loop against `missing-card-api-import`, reachable via the `@cardstack/base/*` safelist entry — exhausts ESLint's 10 fix passes and returns an unclearable `passed: false`.

(1) and (3) look blocking to me; (2) is a silent-correctness issue worth folding into the same pass.

**Minor, not worth inline comments**

- `context.getDeclaredVariables()` is deprecated (emits a DeprecationWarning per worker) and is removed in ESLint 9. Fine on the pinned 8.57.1 — worth a `sourceCode.getDeclaredVariables()` swap whenever the ESLint 9 bump lands.
- A trailing same-line comment (`import X from 'm'; // note`) is left as a stray indented line, but the prettier pass after ESLint cleans it up.
- The PR description still describes enabling `unused-imports/no-unused-imports` from `eslint-plugin-unused-imports@^3`, but the diff implements a hand-written `@cardstack/boxel/no-unused-imports` rule instead. Looks like the body predates the second commit — worth refreshing so the rationale matches what landed.

---

packages/eslint-plugin-boxel/lib/rules/no-unused-imports.js
jurgenwerk: [Claude Code 🤖] Fixed in cf1f76b926, taking the removal-range approach: `rebuildDeclaration` is gone. Unused specifiers are deleted as source ranges (runs of consecutive unused specifiers share one range so each separating comma goes exactly once), and the bare side-effect rewrite replaces only the text up to the module string — so import attributes, the `type` keyword, inline `type` markers, and comments between kept specifiers all survive untouched. When a one-specifier-per-line removal would leave a whitespace-only line, the range expands to the whole line.

New unit tests pin both attribute paths (`with { type: 'json' }` survives a partial removal and the bare rewrite — ember-eslint-parser parses attributes, verified) and the specifier-list comment case from your example.

packages/eslint-plugin-boxel/lib/rules/no-unused-imports.js
jurgenwerk: [Claude Code 🤖] Fixed in cf1f76b926 as suggested: `removeWithNewline` walks `getCommentsBefore(node)` backward, pulling each comment that ends exactly one line above and starts on its own line into the removal range, so a chain of attached directives goes with the statement it annotates. A comment sharing a line with earlier code (a trailing comment on the previous statement) breaks the chain and stays — both shapes are pinned by unit tests. I left `missing-template.gts` alone; with this fix, whenever that file does go through the pipeline, its `@ts-ignore` now travels with the deleted import instead of re-targeting the next statement.

packages/runtime-common/tasks/lint.ts
jurgenwerk: [Claude Code 🤖] Fixed in cf1f76b926, one step upstream of both suggested guards: `missing-card-api-import` no longer fires on class-member keys at all. A member key (`MethodDefinition` / `PropertyDefinition` / accessor, non-computed) is not a scope reference, so an injected import could never satisfy it — it would only enter the loop you traced. I verified the premise first: before the fix, a method named `contains` did trigger the injector in a unit test. With the skip, step 1 never happens, so the safelist stays as is and delete-then-reinject remains symmetric for genuine references (where the injected binding immediately has a reference and is stable).

The bare-import degradation alone would not actually have terminated the loop, for what it is worth — the injector adds specifiers to an existing bindingless import of the same module, so the two rules would still have ping-ponged between `import '…';` and `import { contains } from '…';`. Root-causing the non-reference trigger is what converges.

The convergence endpoint test is in: a card whose class defines `contains(other)` round-trips `/_lint` byte-identical, with no injected import, no messages, and `passed: true`. A plugin unit test pins the member-key skip directly.

PR #5862, opened 2026-08-24
prerenderer: Add `/heap-snapshot` endpoint

backspace:
This is what Claude used to determine the fix in #5863.

### DIFF
--- a/packages/realm-server/prerender/artifact-sink.ts
+++ b/packages/realm-server/prerender/artifact-sink.ts
@@ -52,12 +52,25 @@ const DEFAULT_REGION = 'us-east-1';
 // usual tools recognise: `.cpuprofile` (Chrome DevTools / speedscope),
 // `.trace.json` (Chrome tracing / Perfetto), `.heapprofile` (DevTools
 // allocation-sampling view).
-export type ArtifactKind = 'cpuprofile' | 'trace' | 'heap' | 'v8log';
+export type ArtifactKind =
+  | 'cpuprofile'
+  | 'trace'
+  | 'heap'
+  | 'heapsnapshot'
+  | 'v8log';
 
 const SUFFIX_BY_KIND: Record<ArtifactKind, string> = {
   cpuprofile: 'cpuprofile',
   trace: 'trace.json',
   heap: 'heapprofile',
+  // A full retention snapshot of the *prerender server's own* Node heap,
+  // kept distinct from `heap` above: that one is an allocation sampling
+  // profile of a browser page, taken over CDP, and answers "where is this
+  // render allocating". This one answers "what is still holding memory in
+  // the server process", which is the question a heap that grows across
+  // renders and never comes back down poses. Different producer, different
+  // format, different tooling — so a different suffix.
+  heapsnapshot: 'heapsnapshot',
   // Raw V8 `--prof` tick log (the renderer's `isolate-…-prerender-v8-prof`
   // file), uploaded as-is and symbolized offline with `node --prof-process`.
   // This is the one capture that survives a hard synchronous CPU peg: the
@@ -158,6 +171,15 @@ export function shouldCaptureTrace(): boolean {
 
 export function shouldCaptureHeap(): boolean {
   return flagEnabled('PRERENDER_PROFILE_HEAP');
+}
+
+// Gates the on-demand Node heap snapshot. Off by default and deliberately
+// separate from `PRERENDER_PROFILE_HEAP`: writing this snapshot stops the
+// world for as long as it takes to serialise the whole heap, which is a cost
+// an operator should opt into for one instance rather than something a
+// per-render flag can turn on across the fleet.
+export function shouldAllowHeapSnapshot(): boolean {
+  return flagEnabled('PRERENDER_HEAP_SNAPSHOT');
 }
 
 // True when the sink is configured AND at least one heavyweight capture is
--- a/packages/realm-server/prerender/artifact-sink.ts
+++ b/packages/realm-server/prerender/artifact-sink.ts
@@ -150,6 +150,17 @@ export function artifactSinkEnabled(): boolean {
   return artifactBucket() !== undefined;
 }
 
+// True once this process has spent its upload budget, so a caller can skip
+// work whose only purpose is to produce an upload. `getMaxSessionBytes` is
+// exported for the same reason, but the running total it is compared against
+// is private to this module, which left the answer unreachable from outside.
+// It matters most to the heap snapshot, whose expensive part — stopping the
+// world and writing gigabytes to the container's disk — happens before
+// `uploadArtifact` is ever called.
+export function artifactSinkBudgetSpent(): boolean {
+  return sessionBytesUsed >= getMaxSessionBytes();
+}
+
 // ---------------------------------------------------------------------------
 // Per-mode capture flags. Each heavyweight capture is gated by both the
 // affinity trigger (one deliberately-targeted realm) AND its own flag, so a
@@ -173,11 +184,15 @@ export function shouldCaptureHeap(): boolean {
   return flagEnabled('PRERENDER_PROFILE_HEAP');
 }
 
-// Gates the on-demand Node heap snapshot. Off by default and deliberately
-// separate from `PRERENDER_PROFILE_HEAP`: writing this snapshot stops the
-// world for as long as it takes to serialise the whole heap, which is a cost
-// an operator should opt into for one instance rather than something a
-// per-render flag can turn on across the fleet.
+// Gates the Node heap snapshot, both the route and the automatic capture, and
+// kept separate from `PRERENDER_PROFILE_HEAP` because that one profiles a
+// browser page rather than this process.
+//
+// Read from the environment, which comes from the task definition, so this is
+// per-service: turning it on arms every task in the service, and each one
+// stops the world once on its own when its heap crosses the threshold —
+// staggered across the fleet rather than simultaneous, but fleet-wide. The
+// route is the part that is one instance at a moment somebody chose.
 export function shouldAllowHeapSnapshot(): boolean {
   return flagEnabled('PRERENDER_HEAP_SNAPSHOT');
 }
--- a/packages/realm-server/prerender/heap-snapshot.ts
+++ b/packages/realm-server/prerender/heap-snapshot.ts
@@ -26,6 +26,72 @@ export type HeapSnapshotOutcome =
 // disruptive way to fail than refusing the second request.
 let inFlight = false;
 
+// Heap size at which an instance captures itself without being asked. Chosen
+// as an absolute size rather than a fraction of the limit, because what makes
+// a snapshot expensive is how big the heap is when it is taken, not how close
+// to the ceiling that was. A leak that shows up at 12 GB shows the same
+// retention at 1.5, for a fraction of the pause.
+const DEFAULT_AUTO_CAPTURE_MB = 1536;
+
+// Fires once per process. A leak refills after a restart, so a second
+// snapshot from the same instance describes the same thing at greater cost;
+// and the failure this guards against — a threshold crossed every 30s,
+// pausing the world each time — is worse than missing a capture.
+let autoCaptureArmed = true;
+
+function autoCaptureThresholdMB(): number {
+  let raw = process.env.PRERENDER_HEAP_SNAPSHOT_MB;
+  if (typeof raw === 'string') {
+    let parsed = Number.parseInt(raw.trim(), 10);
+    if (Number.isFinite(parsed) && parsed > 0) {
+      return parsed;
+    }
+  }
+  return DEFAULT_AUTO_CAPTURE_MB;
+}
+
+// Called on each telemetry tick with the heap reading already taken there.
+// Deliberately not awaited by the caller: the write blocks the event loop
+// whether or not anyone waits on it, so awaiting would only delay the tick
+// that follows.
+export function maybeAutoCaptureHeapSnapshot(heapUsedMB: number): void {
+  if (!autoCaptureArmed || !shouldAllowHeapSnapshot()) {
+    return;
+  }
+  let threshold = autoCaptureThresholdMB();
+  if (heapUsedMB < threshold) {
+    return;
+  }
+
+  // Disarm before starting rather than after finishing, so the ticks that
+  // land during a slow capture don't queue up behind it.
+  autoCaptureArmed = false;
+  log.info(
+    `heap reached ${heapUsedMB}MB (auto-capture threshold ${threshold}MB); ` +
+      `capturing one snapshot for this process`,
+  );
+  captureHeapSnapshot()
+    .then((outcome) => {
+      // Re-arm only when nothing was actually written — those outcomes mean
+      // the capture never happened, so this instance still owes one.
+      if (
+        outcome.status === 'disabled' ||
+        outcome.status === 'no-sink' ||
+        outcome.status === 'busy'
+      ) {
+        autoCaptureArmed = true;
+      }
+    })
+    .catch(() => {
+      autoCaptureArmed = true;
+    });
+}
+
+// Test-only: put the once-per-process arm back.
+export function __rearmAutoCaptureForTests(): void {
+  autoCaptureArmed = true;
+}
+
 // Writes the server's own heap to disk and streams it to the artifact sink.
 //
 // `writeHeapSnapshot` serialises straight to a file descriptor rather than
--- a/packages/realm-server/prerender/heap-snapshot.ts
+++ b/packages/realm-server/prerender/heap-snapshot.ts
@@ -1,0 +1,97 @@
+import { writeHeapSnapshot } from 'node:v8';
+import { createReadStream } from 'node:fs';
+import { stat, unlink } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { logger } from '@cardstack/runtime-common';
+import {
+  artifactSinkEnabled,
+  shouldAllowHeapSnapshot,
+  uploadArtifact,
+} from './artifact-sink.ts';
+import { heapTelemetry } from './heap-telemetry.ts';
+
+let log = logger('prerender-heap-snapshot');
+
+export type HeapSnapshotOutcome =
+  | { status: 'captured'; bytes: number; writeMs: number; uploadMs: number }
+  | { status: 'disabled' }
+  | { status: 'no-sink' }
+  | { status: 'busy' }
+  | { status: 'upload-declined'; bytes: number; writeMs: number }
+  | { status: 'failed'; message: string };
+
+// One at a time. Two concurrent snapshots would stop the world twice over and
+// write two multi-gigabyte files to the same disk, which is a far more
+// disruptive way to fail than refusing the second request.
+let inFlight = false;
+
+// Writes the server's own heap to disk and streams it to the artifact sink.
+//
+// `writeHeapSnapshot` serialises straight to a file descriptor rather than
+// building the snapshot in JS memory, so it stays usable on the process it is
+// diagnosing — a heap that is nearly full could not have buffered its own
+// snapshot. The trade is that it stops the world for the whole write: seconds
+// on a small heap, and long enough on a large one that the manager's
+// heartbeat (30s by default) can lapse and evict this instance until it
+// re-registers. Capture at a couple of gigabytes rather than near the limit —
+// the retention pattern is the same and everything about it is cheaper.
+export async function captureHeapSnapshot(): Promise<HeapSnapshotOutcome> {
+  if (!shouldAllowHeapSnapshot()) {
+    return { status: 'disabled' };
+  }
+  if (!artifactSinkEnabled()) {
+    return { status: 'no-sink' };
+  }
+  if (inFlight) {
+    return { status: 'busy' };
+  }
+  inFlight = true;
+
+  let before = heapTelemetry();
+  let path = join(
+    tmpdir(),
+    `prerender-${process.pid}-${Date.now()}.heapsnapshot`,
+  );
+  let bytes = 0;
+  let writeMs = 0;
+
+  try {
+    log.info(
+      `heap snapshot starting heapUsedMB=${before.heapUsedMB} ` +
+        `rssMB=${before.rssMB} path=${path} (the process is paused until this completes)`,
+    );
+    let startedAt = Date.now();
+    writeHeapSnapshot(path);
+    writeMs = Date.now() - startedAt;
+    bytes = (await stat(path)).size;
+    log.info(`heap snapshot written bytes=${bytes} writeMs=${writeMs}`);
+
+    let uploadStartedAt = Date.now();
+    let uploaded = await uploadArtifact({
+      kind: 'heapsnapshot',
+      step: `node-heap-${before.heapUsedMB}mb`,
+      body: createReadStream(path),
+      contentType: 'application/octet-stream',
+    });
+    let uploadMs = Date.now() - uploadStartedAt;
+    if (!uploaded) {
+      // The sink declines rather than throws when its byte budget is spent or
+      // no bucket is configured; the snapshot is still on disk at this point,
+      // so say so rather than reporting success.
+      log.warn(`heap snapshot upload declined bytes=${bytes}`);
+      return { status: 'upload-declined', bytes, writeMs };
+    }
+    log.info(`heap snapshot uploaded bytes=${bytes} uploadMs=${uploadMs}`);
+    return { status: 'captured', bytes, writeMs, uploadMs };
+  } catch (e) {
+    let message = e instanceof Error ? e.message : String(e);
+    log.warn(`heap snapshot failed: ${message}`);
+    return { status: 'failed', message };
+  } finally {
+    // The file is multi-gigabyte and the container's disk is shared with
+    // every render, so it goes whether or not the upload worked.
+    await unlink(path).catch(() => undefined);
+    inFlight = false;
+  }
+}
--- a/packages/realm-server/prerender/heap-snapshot.ts
+++ b/packages/realm-server/prerender/heap-snapshot.ts
@@ -5,6 +5,7 @@ import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { logger } from '@cardstack/runtime-common';
 import {
+  artifactSinkBudgetSpent,
   artifactSinkEnabled,
   shouldAllowHeapSnapshot,
   uploadArtifact,
@@ -17,6 +18,7 @@ export type HeapSnapshotOutcome =
   | { status: 'captured'; bytes: number; writeMs: number; uploadMs: number }
   | { status: 'disabled' }
   | { status: 'no-sink' }
+  | { status: 'budget-spent' }
   | { status: 'busy' }
   | { status: 'upload-declined'; bytes: number; writeMs: number }
   | { status: 'failed'; message: string };
@@ -50,16 +52,47 @@ function autoCaptureThresholdMB(): number {
   return DEFAULT_AUTO_CAPTURE_MB;
 }
 
-// Called on each telemetry tick with the heap reading already taken there.
-// Deliberately not awaited by the caller: the write blocks the event loop
-// whether or not anyone waits on it, so awaiting would only delay the tick
-// that follows.
-export function maybeAutoCaptureHeapSnapshot(heapUsedMB: number): void {
-  if (!autoCaptureArmed || !shouldAllowHeapSnapshot()) {
+// How many consecutive over-threshold ticks to spend waiting for an idle one
+// before capturing anyway. An instance under continuous load would otherwise
+// never capture, and its heap keeps climbing while it waits — so the pause
+// only gets more expensive the longer patience lasts. At the tick interval
+// this is a handful of minutes.
+const IDLE_WAIT_TICKS = 10;
+let ticksWaitingForIdle = 0;
+
+// Called on each telemetry tick with the heap reading and pool state already
+// taken there. Deliberately not awaited by the caller: the write blocks the
+// event loop whether or not anyone waits on it, so awaiting would only delay
+// the tick that follows.
+export function maybeAutoCaptureHeapSnapshot(
+  heapUsedMB: number,
+  poolIdle: boolean,
+): void {
+  // `artifactSinkEnabled` is checked here as well as inside the capture,
+  // because flag-on/bucket-absent is an ordinary state rather than a
+  // misconfiguration: without it this announces a capture, gets `no-sink`,
+  // re-arms, and does the same 30 seconds later for the life of the process.
+  if (
+    !autoCaptureArmed ||
+    !shouldAllowHeapSnapshot() ||
+    !artifactSinkEnabled()
+  ) {
     return;
   }
   let threshold = autoCaptureThresholdMB();
   if (heapUsedMB < threshold) {
+    return;
+  }
+
+  // Prefer an idle tick. Timers do not run during the write, so a pause that
+  // lands mid-render is charged to every render in flight, and one far enough
+  // into its budget crosses the render timeout the moment the loop resumes —
+  // turning a diagnostic into a source of the errors it exists to explain. An
+  // idle sample is also the better answer to the question being asked: what
+  // this process still holds once the work is done, rather than the live
+  // state of whatever is running.
+  if (!poolIdle && ticksWaitingForIdle < IDLE_WAIT_TICKS) {
+    ticksWaitingForIdle++;
     return;
   }
 
@@ -90,6 +123,7 @@ export function maybeAutoCaptureHeapSnapshot(heapUsedMB: number): void {
 // Test-only: put the once-per-process arm back.
 export function __rearmAutoCaptureForTests(): void {
   autoCaptureArmed = true;
+  ticksWaitingForIdle = 0;
 }
 
 // Writes the server's own heap to disk and streams it to the artifact sink.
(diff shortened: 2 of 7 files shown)


packages/realm-server/prerender/heap-snapshot.ts
```diff
@@ -0,0 +1,163 @@
...
+    writeHeapSnapshot(path);
+    writeMs = Date.now() - startedAt;
+    bytes = (await stat(path)).size;
+    log.info(`heap snapshot written bytes=${bytes} writeMs=${writeMs}`);
+
+    let uploadStartedAt = Date.now();
+    let uploaded = await uploadArtifact({
+      kind: 'heapsnapshot',
+      step: `node-heap-${before.heapUsedMB}mb`,
+      body: createReadStream(path),
```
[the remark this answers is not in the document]
backspace: [Claude Code 🤖] Valid, and correctly rated. Fixed in f410346faf.

Both halves are real. `createReadStream(path)` was an argument expression, so the stream existed whether or not `uploadArtifact` went on to read it, and the `finally` unlinks the file immediately afterwards. A `ReadStream` with no `error` listener that fails to open raises an unhandled `error`, which ends the process — a crash in the code whose reason for existing is diagnosing a service that crashes. When the open won the race instead, the descriptor stayed open, and on POSIX an unlinked file with an open descriptor keeps its blocks: several hundred megabytes, at these sizes.

One detail worth adding on reachability, since it changes how likely this is. The `!bucket` early return can't fire here — `captureHeapSnapshot` checks `artifactSinkEnabled()` before it writes anything — so only the byte-budget path can, which needs ~5 GiB of prior successful uploads in one process. That sounds remote for a capture that fires once per process, except the once-per-process guard lives on the *auto*-capture; `POST /heap-snapshot` has only the in-flight guard. Around eleven manual captures at ~500 MB each gets there, and the next one could take the process down.

The fix holds the stream in a variable, attaches an `error` listener, and destroys it in `finally`. Destroying a stream the upload already drained is a no-op, so the working paths are unchanged.

Verified by forcing the decline path with a sink that refuses: three consecutive captures each returned `upload-declined`, the listener logged `heap snapshot read stream failed: The operation was aborted`, and the process stayed up. Before the change that same sequence had no listener to catch it.

packages/realm-server/prerender/heap-snapshot.ts
```diff
@@ -0,0 +1,181 @@
...
+// write two multi-gigabyte files to the same disk, which is a far more
+// disruptive way to fail than refusing the second request.
+let inFlight = false;
+
+// Heap size at which an instance captures itself without being asked. Chosen
+// as an absolute size rather than a fraction of the limit, because what makes
+// a snapshot expensive is how big the heap is when it is taken, not how close
+// to the ceiling that was. A leak that shows up at 12 GB shows the same
+// retention at 1.5, for a fraction of the pause.
+const DEFAULT_AUTO_CAPTURE_MB = 1536;
```
habdelra, line 36: [Claude Code 🤖] **At this default the pause is minutes long, and the artifact does not exist until after the pause ends — so the most likely outcome of an automatic capture is that it destroys its own output.**

**What I measured.** Node v22.22.2, retaining many small objects (the shape a retention leak has, and the shape that drives snapshot cost — the serializer walks the object graph, so cost follows object count rather than bytes):

| `heapUsedMB` | `writeHeapSnapshot` | file written |
| --- | --- | --- |
| 257 | 28.2 s | 513 MB |
| 515 | 47.8 s | 1047 MB |
| 1024 | 119.5 s | 2115 MB |

Doubling the heap from 515 to 1024 cost 2.5× the time, so 1536 extrapolates to roughly three minutes of blocked event loop and a ~3 GB file. Caveat: that's my machine's CPU and one object shape — a heap dominated by large buffers would serialize faster — but the order of magnitude is the point, and the small end is measured, not extrapolated.

**Why the duration matters more than the doc comment allows.** `writeHeapSnapshot` is synchronous: no HTTP responses, no timers, no heartbeat run while it works. Three things key off that.

- **The manager heartbeat.** `manager-app.ts` resolves `heartbeatTimeoutMs` to 30 s and sweeps every 5 s, so the instance goes `stale` and stops being routed to. `captureHeapSnapshot`'s doc comment already names this, but frames it as what happens "on a large one" — 28.2 s at a 257 MB heap is already at that line, and every size above it is past it. Recoverable, as the comment says: the next heartbeat re-registers.
- **The container liveness probe, which is not recoverable.** `prerender.Dockerfile` declares `HEALTHCHECK --interval=30s --timeout=5s --retries=3` against `GET /`, and the `/` handler's own comment in `prerender-app.ts` treats that probe as live ("The Docker HEALTHCHECK discards this body"). Three consecutive misses is on the order of 100 s of blockage. Past that the container is replaced — and because the upload only starts after `writeHeapSnapshot` returns, a replacement during the write or the upload takes the snapshot with it: `tmpdir()` is container-ephemeral. At 1536 MB the pause is comfortably past that threshold, so this is the expected path, not the edge case. I can't see the task definition, so I can't tell whether ECS enforces this image healthcheck, a task-definition copy of it, or a target-group check — but something on that route decides task liveness, and a three-minute block fails all of them.
- **The route's own response.** `POST /heap-snapshot` awaits the capture, so the operator's client can't be answered for the whole pause either. A client that gives up gets no outcome even though the capture ran and uploaded.

**Also a comment-accuracy point**, on `captureHeapSnapshot`'s doc comment: "seconds on a small heap" undersells the floor. 28 s is the smallest heap I could get a number for.

**The way out.** Pick against the real task definition, which I can't read:

1. Measure the pause on an actual prerender task at a candidate threshold and set `DEFAULT_AUTO_CAPTURE_MB` so it fits inside whatever probe decides liveness there. My numbers point at a few hundred MB rather than 1536 — and per this comment's own reasoning ("A leak that shows up at 12 GB shows the same retention at 1.5"), the same argument runs all the way down: the retention pattern at 384 MB is the same one, at a quarter of the pause.
2. If task replacement is an acceptable price, say so here and in the flag's docs, so nobody reads a lost artifact as a bug.

Either way the comment should carry the measured duration at the chosen threshold — this is the number an operator needs before turning the flag on, and it isn't derivable from anything else in the file.

**Scope:** introduced by this PR, and the one finding here I'd call blocking — not because the code is wrong, but because the shipped default is likely to make the feature not work.

---

packages/realm-server/prerender/heap-snapshot.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (!shouldAllowHeapSnapshot()) {
+    return { status: 'disabled' };
+  }
+  if (!artifactSinkEnabled()) {
+    return { status: 'no-sink' };
+  }
+  if (inFlight) {
+    return { status: 'busy' };
+  }
+  inFlight = true;
```
habdelra, line 163: [Claude Code 🤖] **The world stops before anything checks whether the upload can land.**

These guards cover two of the sink's three decline conditions — the flag and a configured bucket. The third, the per-process byte budget, isn't consulted until `uploadArtifact` runs, which is after the pause and after several GB have hit the container's disk. So on a process whose budget is spent, a capture blocks the event loop for minutes, writes a multi-GB file into the same `tmpdir()` that Chrome's user-data dirs and the V8 `--prof` logs live in (`browser-manager.ts` sweeps that root), and then deletes it, having learned nothing.

`artifact-sink.ts` already anticipates this caller. `getMaxSessionBytes` is exported with the comment "for callers that want to short-circuit a capture before doing CDP work" — the budget-spent state just isn't reachable from outside the module, since `sessionBytesUsed` is private. A `sinkBudgetExhausted()` predicate next to `artifactSinkEnabled()`, checked in the same guard block here, turns a three-minute pause into a cheap `upload-declined` (or a clearer `budget-spent`).

**The budget runs the other way too, and that one is worth a line in the operator docs.** At the 1536 MB default the file measures ~3 GB (see the thread on `DEFAULT_AUTO_CAPTURE_MB`), against `DEFAULT_MAX_SESSION_BYTES` of 5 GiB. A single automatic capture spends ~60% of the process's entire artifact budget; the check is `sessionBytesUsed >= max` *before* the upload, so it can also overshoot the ceiling by the whole 3 GB in one object. Either way, the next capture the operator asks for is declined. The case that actually bites: an operator running a targeted `.cpuprofile`/trace session on a task — which is exactly when these flags are on together — finds the sink silently dead for the rest of the process life, one `warn` line explaining it, and nothing tying that to the render they were profiling. Worth considering whether the snapshot should carry its own budget rather than share the profiler's.

**Related, and what makes the budget path reachable at all:** the auto-capture is once-per-process, but `POST /heap-snapshot` has only the `inFlight` guard, so the manual path can repeat a full-cost capture indefinitely. That's the reachability argument already made in the read-stream thread; it applies here unchanged.

**Scope:** introduced by this PR. Non-blocking, and the guard itself is a couple of lines.

---

packages/realm-server/prerender/heap-snapshot.ts
```diff
@@ -0,0 +1,181 @@
...
+        outcome.status === 'no-sink' ||
+        outcome.status === 'busy'
+      ) {
+        autoCaptureArmed = true;
+      }
+    })
+    .catch(() => {
+      autoCaptureArmed = true;
+    });
+}
```
habdelra: [Claude Code 🤖] **With the flag on and no bucket configured, this announces a capture it never takes, every 30 s, for the life of the process.**

The gate here is `shouldAllowHeapSnapshot()` only; `artifactSinkEnabled()` is checked one level down, inside `captureHeapSnapshot`. Those two knobs are independent by design — the runbook has `PRERENDER_ARTIFACTS_BUCKET` wired by Terraform and the mode flags set by an operator through SSM — so "flag on, bucket absent" is an ordinary state, not a misconfiguration you have to construct.

Walk the loop: `#startQueueSnapshotLoop` in `prerenderer.ts` ticks every 30 s. Above the threshold, each tick sets `autoCaptureArmed = false`, logs `heap reached NMB (auto-capture threshold NMB); capturing one snapshot for this process`, calls `captureHeapSnapshot`, gets `no-sink` back, and re-arms. Heap only goes up, so it never falls back below the threshold and stops. The operator reads a log announcing a capture every 30 s with no artifact behind any of them — in a module whose whole job is being legible to someone reading logs at 3am.

The fix is the same cheap gate, moved ahead of the disarm and the log line:

```suggestion
export function maybeAutoCaptureHeapSnapshot(heapUsedMB: number): void {
  if (
    !autoCaptureArmed ||
    !shouldAllowHeapSnapshot() ||
    !artifactSinkEnabled()
  ) {
    return;
  }
```

(needs `artifactSinkEnabled` added to the import from `./artifact-sink.ts`, which already imports its siblings.)

**Same function, smaller point:** `'disabled'` in the re-arm set is unreachable. `captureHeapSnapshot` is an async function, so its body runs synchronously up to the first `await` — its `shouldAllowHeapSnapshot()` call happens in the same tick as the one at the top of this function, and `process.env` can't change in between. Harmless, but it reads as a live case to the next person deciding whether the re-arm set is complete. With the suggestion above, `'no-sink'` becomes similarly unreachable and the set collapses to `'busy'`.

**Scope:** introduced by this PR. Non-blocking — log noise, not lost data.

---

packages/realm-server/prerender/heap-snapshot.ts
```diff
@@ -0,0 +1,181 @@
...
+      contentType: 'application/octet-stream',
+    });
+    let uploadMs = Date.now() - uploadStartedAt;
+    if (!uploaded) {
+      // The sink declines rather than throws when its byte budget is spent or
+      // no bucket is configured; the snapshot is still on disk at this point,
+      // so say so rather than reporting success.
+      log.warn(`heap snapshot upload declined bytes=${bytes}`);
+      return { status: 'upload-declined', bytes, writeMs };
+    }
```
habdelra, line 217: [Claude Code 🤖] **`upload-declined` also covers a genuine S3 failure, so this comment claims a narrower cause than the code has.**

`uploadArtifact` returns `false` on three paths, not two: no bucket, budget spent, and its own `catch` — `artifact-sink failed to upload ${kind} key=${key}` — which swallows every S3 error (credentials, bucket policy, network, a part upload that never completes). All three arrive here as the same `false`, so a capture that stopped the world for minutes and then couldn't reach S3 at all is reported to the operator as "declined", at `warn`, indistinguishable from the budget case. The route compounds it: `upload-declined` falls through `default:` to a 500, the same code as `failed`, so the HTTP status doesn't separate them either.

The distinction is the operator's next move. "Declined" means raise the budget or come back on a fresh task — the capability works. A failure means the task role or the bucket name is wrong and no capture on this task will ever land, so retrying at a lower threshold is wasted pause.

Two ways out, and I'd take the second here:

1. Have `uploadArtifact` return a discriminated result rather than a boolean. Cleanest, but it has six call sites across `utils.ts`, `v8-prof.ts` and this module, and `v8-prof.ts` gates a local-file delete on the boolean — a bigger change than this PR should carry. Follow-up at most.
2. Stop asserting the cause. Reword the comment and the log line to say the upload didn't land, and point at the sink's own `warn` line for why — it already prints the error object, keyed by the same S3 key. One line of prose, no behaviour change.

**Scope:** introduced by this PR (comment vs. code). Non-blocking.

---

packages/realm-server/prerender/heap-snapshot.ts
```diff
@@ -0,0 +1,181 @@
...
+    })
+    .catch(() => {
+      autoCaptureArmed = true;
+    });
+}
+
+// Test-only: put the once-per-process arm back.
+export function __rearmAutoCaptureForTests(): void {
+  autoCaptureArmed = true;
+}
```
habdelra, line 127: [Claude Code 🤖] **This export has no callers — nothing in the suite covers this module at all.**

`grep -rn '__rearmAutoCaptureForTests'` across the repo returns exactly one hit: this definition. There's no `heap-snapshot` test file, and no existing test imports anything from this module. The only executable statement about any of it right now is that it type-checks.

That's out of step with its two neighbours, and both of their harnesses take this module without new infrastructure:

- **`tests/prerender-artifact-sink-test.ts`** has `'each per-mode flag is on only for the exact string "true"'` driven off a table of `[envVar, reader]` pairs — `['PRERENDER_HEAP_SNAPSHOT', shouldAllowHeapSnapshot]` is one line in it. It also has `'each artifact kind maps to its conventional file suffix'`, which enumerates the kinds by hand and now misses `heapsnapshot`. `SUFFIX_BY_KIND` is a `Record<ArtifactKind, string>`, so the compiler catches a *missing* entry — but nothing catches a *wrong* suffix, which is the thing that test exists to pin, and the new kind is the one whose suffix a reader is most likely to second-guess (`.heapsnapshot` vs. the neighbouring `.heapprofile`).
- **`tests/prerender-server-test.ts`** drives the app through supertest against `buildPrerenderApp`. `POST /heap-snapshot` with the flag unset asserts the 404 resting state and the JSON body's `status: 'disabled'`, and it stops nothing: the guard returns before any write, so the test costs microseconds.
- **The arm/disarm logic this export exists to support**, as a plain unit test with the flag on and no bucket: below threshold → no capture; at threshold → armed goes false; a second tick → nothing; `no-sink` → re-armed. That last case is the loop in the thread on `maybeAutoCaptureHeapSnapshot` — a test for it would have caught that before review.

None of these needs S3, Chrome, or a real snapshot. What would have to break for them to fail is exactly what the module promises.

**Scope:** introduced by this PR (new code, no coverage). Non-blocking, but cheap enough that I'd rather see it here than filed.

---

packages/realm-server/prerender/artifact-sink.ts
```diff
@@ -160,6 +173,15 @@ export function shouldCaptureHeap(): boolean {
...
 }
 
+// Gates the on-demand Node heap snapshot. Off by default and deliberately
+// separate from `PRERENDER_PROFILE_HEAP`: writing this snapshot stops the
+// world for as long as it takes to serialise the whole heap, which is a cost
+// an operator should opt into for one instance rather than something a
+// per-render flag can turn on across the fleet.
+export function shouldAllowHeapSnapshot(): boolean {
+  return flagEnabled('PRERENDER_HEAP_SNAPSHOT');
+}
```
habdelra: [Claude Code 🤖] **"for one instance" isn't what this flag does any more.**

The comment draws the line between "a cost an operator should opt into for one instance" and "something a per-render flag can turn on across the fleet", and puts `PRERENDER_HEAP_SNAPSHOT` on the first side. That was true of the route alone. It stopped being true when the auto-capture landed on the same flag: `maybeAutoCaptureHeapSnapshot` gates on `shouldAllowHeapSnapshot()`, the flag is read from `process.env`, and per the operator runbook these values come from SSM through the task definition — which is per-service, not per-task. Setting it arms every instance in the service, each of which then takes its own multi-minute pause the moment its heap crosses the threshold. That is fleet-wide; it's just staggered rather than simultaneous.

The route is the "one instance, chosen moment" part. The flag that enables it isn't, and this is the comment an operator reads before deciding what turning it on costs.

Two things worth doing:

1. Reword so the cost is stated as what it is — every task in the service will pause itself once, unattended, at the threshold — and keep the "chosen moment" framing for the route.
2. If a genuinely single-instance opt-in is wanted, the auto-capture needs to be separable from the route. `PRERENDER_HEAP_SNAPSHOT_MB=0` currently falls through `parsed > 0` to the default rather than meaning anything, so it's free to give it the meaning "route only, never automatic" — one branch in `autoCaptureThresholdMB`, and the knob already exists.

**Two enumerations in this file also missed the new kind**, both of them comments the diff sits inside: the module header ("full `.cpuprofile`s, CDP trace streams, and heap-allocation profiles") and the doc comment above `ArtifactKind` ("`.cpuprofile` (Chrome DevTools / speedscope), `.trace.json` (Chrome tracing / Perfetto), `.heapprofile` (DevTools allocation-sampling view)"). The second one is where a reader goes to find out what tool opens a given suffix, which for `.heapsnapshot` is DevTools → Memory → *Load profile…*, a different panel from `.heapprofile`'s. (`v8log` is missing from both too — that's pre-existing, not this PR's, but you're editing the list.)

**Scope:** introduced by this PR (comment vs. code). Non-blocking.

---

packages/realm-server/prerender/prerenderer.ts
```diff
@@ -997,7 +998,11 @@ export class Prerenderer {
       // has no queue behind it to explain it. Kept separate from the
       // snapshot line so existing greps of that line are unaffected.
       try {
-        log.info('prerender-heap %s', formatHeapTelemetry(heapTelemetry()));
+        let telemetry = heapTelemetry();
+        log.info('prerender-heap %s', formatHeapTelemetry(telemetry));
+        // Reuses the reading just taken, so the instance that crosses the
+        // threshold captures itself off the same tick that reports it.
+        maybeAutoCaptureHeapSnapshot(telemetry.heapUsedMB);
```
habdelra: [Claude Code 🤖] **The capture fires regardless of what the pool is doing, and the pause is charged to whatever happens to be rendering.**

Timers don't fire while `writeHeapSnapshot` holds the loop; they fire late, all at once, when it lets go. So a render 40 s into its budget when a two-minute pause starts is past `RENDER_TIMEOUT_MS` (`prerender-constants.ts`, default 60 s) the instant the loop resumes, and fails — on an indexer visit that's an error row on a card that had nothing wrong with it. Every render in flight on that instance, up to the pool's capacity, takes the same hit together.

An idle signal is one line away: the very next `try` block computes `this.#pagePool.getQueueDepthSnapshot()`, which carries `totalPending` and the per-affinity `busy` counts. Hoisting that call above the heap block and passing an idle flag into `maybeAutoCaptureHeapSnapshot` costs nothing per tick.

And idle is the better *sample*, not just the cheaper moment. The comment directly above this one argues exactly that — "The heap holds memory from work already finished and keeps growing with the pool idle, so gating this on current load would hide exactly the growth that has no queue behind it to explain it". A snapshot taken mid-render is full of live render state that isn't the leak; a snapshot on an idle tick is post-render retention, which is the question being asked.

The trade is an instance that never goes idle never captures. If that's a real concern under sustained indexing, capture anyway after N consecutive above-threshold ticks — at 30 s a tick, "prefer idle, give up after five minutes of waiting" is still far better targeted than capturing on the first tick over the line.

**Scope:** introduced by this PR. Non-blocking, and it's a design call rather than a defect — but the current behaviour turns a diagnostic into a source of the same indexing errors the diagnostic is meant to explain, which is worth deciding deliberately.

---

habdelra (commented):
[Claude Code 🤖] I reviewed this as an operational change rather than a code change: the code is short and the module boundaries are clean, so what decides whether this works is what the capture *costs* on a live task and whether an operator can find and use it. That meant measuring `writeHeapSnapshot` rather than reading about it, and tracing the pause against everything on this process that has a timeout.

**Bottom line: the design is right and the failure modes are thought about, but the shipped default threshold is very likely to destroy the artifact it captures.** I measured 28.2 s of blocked event loop at a 257 MB heap and 119.5 s at 1024 MB; `DEFAULT_AUTO_CAPTURE_MB` is 1536, which extrapolates to roughly three minutes. The upload only begins after the write returns, `tmpdir()` is container-ephemeral, and the image's `HEALTHCHECK` is `--interval=30s --timeout=5s --retries=3` — so a task replaced mid-capture loses the snapshot entirely. That's the one blocking item; everything else is small.

**On the existing thread** (owning the read stream rather than assuming it's consumed): I traced the fix and it holds. The stream is bound before `uploadArtifact`, the `error` listener is attached at construction so a losing open-vs-unlink race can't reach the unhandled path, and `body?.destroy()` runs in `finally` ahead of the `unlink` on every exit. Nothing further needed there. The reachability argument in that reply — that the manual route has no once-per-process guard — is load-bearing for one of the findings below, so it's worth keeping in mind rather than treating as settled trivia.

**Recommendations, most to least important:**

1. **Set `DEFAULT_AUTO_CAPTURE_MB` from a measured pause on a real task**, or state explicitly that task replacement and artifact loss are accepted. Measurements, mechanism and options in the thread on `DEFAULT_AUTO_CAPTURE_MB` in `heap-snapshot.ts`. *Blocking.*
2. **Check the sink's byte budget before the world-stopping write**, not after — `getMaxSessionBytes` is exported for exactly this. Also worth deciding whether a ~3 GB snapshot should share the profiler's 5 GiB session budget at all. See the thread on `captureHeapSnapshot`'s guard block.
3. **Gate `maybeAutoCaptureHeapSnapshot` on `artifactSinkEnabled()`.** Flag-on/bucket-absent is an ordinary state, and it currently logs an announced capture every 30 s forever without taking one. Suggestion block in that thread.
4. **Decide whether the auto-capture should wait for an idle tick.** As written it can push in-flight renders past `RENDER_TIMEOUT_MS`, turning the diagnostic into a source of the indexing errors it's meant to explain — and an idle sample answers the retention question better anyway. Thread on the call site in `prerenderer.ts`.
5. **Add the tests.** `__rearmAutoCaptureForTests` currently has zero callers and no test file touches this module; the artifact-sink flag/suffix tables and the supertest route harness both take the new surface without new infrastructure — including the missing `heapsnapshot` row in `'each artifact kind maps to its conventional file suffix'`. Thread on that export.
6. **Update the operator runbook.** `.claude/skills/indexing-diagnostics/SKILL.md` Mode H is where someone goes to run one of these captures, and it doesn't know this exists: the two-tier table doesn't list the snapshot, the SSM knob table has no `PRERENDER_HEAP_SNAPSHOT` / `PRERENDER_HEAP_SNAPSHOT_MB` row, "Running a capture session" has no path for it, and "Reading each artifact" doesn't say that `.heapsnapshot` opens in DevTools → Memory → *Load profile…* (a different panel from `.heapprofile`'s allocation-sampling view). The route's own comment says to reach it by SSM port-forward, which is precisely the kind of instruction that has to live in the runbook rather than in the handler. This is also where the auto-capture threshold, its measured pause, and the budget interaction from (2) belong — the flag now does something unattended, and nothing outside this file says so.
7. **Fix the two comment/code divergences:** the "for one instance" rationale on `shouldAllowHeapSnapshot` (the flag is per-service, so the auto-capture arms the whole fleet), and `upload-declined`'s claim that the sink only declines for budget/bucket reasons when `uploadArtifact` also returns `false` on any S3 error. Both in their own threads.

**Adjacent, out of scope.** The prerender server has no endpoint authentication of any kind — the `auth` field on the render routes is a realm session handed *into* the render, not a check on the caller — so `/heap-snapshot` matches the existing posture and I'm not asking for anything here. Worth noting only because this is the first route whose cost to an unauthenticated caller is measured in minutes of process downtime rather than one render; if a reachability assumption about that port ever changes, this is the endpoint that changes with it. Separately, `v8log` is absent from both kind-enumerating comments in `artifact-sink.ts` — pre-existing, and only worth mentioning because this PR edits that list.

---

packages/realm-server/prerender/heap-snapshot.ts
backspace: [Claude Code 🤖] Pushing back on the blocking item, with data from four real captures — but taking the mechanism, which is sound.

This threshold has already run in production-shaped conditions. Four automatic captures on a staging task, at exactly the default:

| heapUsedMB | writeMs | file | uploadMs |
| --- | --- | --- | --- |
| 1565 | **15.9 s** | 512 MB | 13.1 s |
| 1559 | **12.4 s** | 407 MB | 9.0 s |
| 1588 | **7.3 s** | 226 MB | 4.7 s |
| 1564 | **6.9 s** | 198 MB | 3.6 s |

All four wrote, uploaded, and survived. No task was replaced, no instance was swept. So the extrapolation is out by roughly an order of magnitude for this workload — ~3 minutes predicted against 6.9–15.9 s observed, and ~3 GB predicted against 198–512 MB.

Your caveat is exactly the reason. Cost follows object count, and a real prerender heap is unusually sparse: the 1564 MB snapshot I analysed has 2.03 M nodes and 7.67 M edges, and its file is 0.13× the heap. Yours at 1024 MB produced a file **2×** the heap, so roughly fifteen times the object density. Both numbers are right about their own heap; the prerender one is the one this default has to survive.

Two corrections to the mechanism, one in each direction:

- **The `HEALTHCHECK` isn't enforced.** `healthCheck` is `null` on the task definition, so ECS never runs the Dockerfile's — that's Docker-local and ECS only acts on what the task definition declares. `healthStatus` reads `UNKNOWN` on every running task. So task replacement mid-capture isn't the failure mode here.
- **The heartbeat is.** 30 s, and the worst observed pause is 15.9 s — half of it, with no margin worth relying on. That's the real bound, and it's why I've kept the default low rather than raised it.

Changed in 15ec5f8c7f: the doc comment now carries these measured figures instead of "seconds on a small heap", names the 30 s heartbeat as the number that bounds everything, and says plainly that object count rather than size drives the cost — so the next person to consider raising this sees what it actually cost here, and why their heap might differ.

Left the default at 1536. Four for four at that value, with the observed pause at half the binding limit.

packages/realm-server/prerender/heap-snapshot.ts
backspace: [Claude Code 🤖] Taken in 15ec5f8c7f. `artifactSinkBudgetSpent()` sits next to `artifactSinkEnabled()` and is checked in the same guard block, before `inFlight` and well before the write — the guards now read `disabled`, `no-sink`, `budget-spent`, `busy`, and only then does anything expensive happen. New `budget-spent` outcome so the operator sees why nothing was captured.

One note on how reachable that state is, since it changes the priority rather than the correctness. `sessionBytesUsed` only advances on a **successful** upload — the `catch` returns `false` without counting anything — so reaching the 5 GiB budget takes something like eleven successful captures on one process. The automatic capture fires once per process, so that means the manual route, which is the reachability point from the earlier stream thread. Worth guarding because the guard is free; rarer than "a process whose budget is spent" suggests.

I tried to prove the guard by forcing the spent state and couldn't, for that same reason — the running total is private and only a real upload moves it. So this one is verified by construction and position rather than by observation, which I'd rather say than imply otherwise.

On the budget running the other way: with the files this workload actually produces (198–512 MB, measured — see the threshold thread) one capture spends 4–10% of the 5 GiB, not ~60%. Still worth an operator knowing, so the runbook now says the snapshot shares that budget with the profiler artifacts and that the number moves with the threshold.

packages/realm-server/prerender/heap-snapshot.ts
backspace: [Claude Code 🤖] Taken as suggested in 15ec5f8c7f — `artifactSinkEnabled()` added to the gate, ahead of the disarm and the log line.

You're right that it's an ordinary state rather than one you have to construct, and right about the loop: re-arming on `no-sink` was deliberate — an instance that couldn't capture still owes one — but combined with a heap that only goes up it meant the announcement repeated forever. An operator reading logs at 3am would see a capture claimed every thirty seconds with nothing behind any of them, which is precisely the opposite of what this module is for.

Verified: five consecutive over-threshold ticks with the flag on and no bucket now return in 51 ms total, silently, and the instance stays armed for when a bucket appears.

packages/realm-server/prerender/prerenderer.ts
backspace: [Claude Code 🤖] Taken in 15ec5f8c7f. `getQueueDepthSnapshot()` is hoisted above the heap block and an idle flag passed in — idle meaning `totalPending === 0` and no tabs busy on file, module or command work.

The argument that persuaded me is the second one, not the first. A pause charged to in-flight renders is bad; a snapshot full of live render state that isn't the leak is worse, because it's the diagnostic being wrong rather than expensive. And you're right that the comment three lines up was already making this case for the telemetry line and I hadn't carried it into the capture.

On the trade you name — an instance that never goes idle never captures — I bounded the patience rather than leaving it open: after `IDLE_WAIT_TICKS` (10, so about five minutes at the tick interval) it captures regardless. Waiting indefinitely gets worse rather than safer, since the heap climbs the whole time and the pause grows with it, so patience has to expire somewhere. Verified: with every tick reporting busy, the capture fires on tick 10.

packages/realm-server/prerender/heap-snapshot.ts
backspace: [Claude Code 🤖] Taken in 15ec5f8c7f, using both harnesses exactly as you describe.

`tests/prerender-artifact-sink-test.ts` gains the `['PRERENDER_HEAP_SNAPSHOT', shouldAllowHeapSnapshot]` row in the flag table, and `assert.strictEqual(suffix('heapsnapshot'), 'heapsnapshot')` in the suffix table — with a comment saying why it differs from the `heapprofile` line above it, since that's the pair a reader second-guesses and the compiler only catches a missing row, not a wrong one.

`tests/prerender-server-test.ts` gains the disabled-route case: `POST /heap-snapshot` with the flag unset asserts the 404 and `status: 'disabled'`. As you say it stops nothing — the guard returns before any write.

`__rearmAutoCaptureForTests` still has no caller in the suite. It's now genuinely used, but by the standalone harness I run these behaviours through rather than by QUnit, because the arm/disarm cycle needs a real capture to advance and a real capture stops the world for seconds. I'd rather say that than quietly leave it looking covered: the auto-capture path is verified by measurement (thresholds, the sink gate, the idle fallback firing on tick 10) and not by the suite.

packages/realm-server/prerender/artifact-sink.ts
backspace: [Claude Code 🤖] Right, and reworded in 15ec5f8c7f. The comment described the route and then got attached to a flag that now also arms an unattended fleet-wide capture — written before the auto-capture existed and not revisited when it landed.

It now says the flag gates both the route and the automatic capture, that it is read from the environment and therefore per-service, and that turning it on means every task in the service stops the world once on its own when its heap crosses the threshold — staggered rather than simultaneous, but fleet-wide either way. The "one instance at a moment somebody chose" framing stays, attached to the route where it's true.

Didn't take the `PRERENDER_HEAP_SNAPSHOT_MB=0` idea. It's a clean use of a value that currently means nothing, but `0` reading as "route only, never automatic" isn't guessable from the name, and the same knob would then carry a threshold and a mode. If a genuinely single-instance opt-in is wanted I'd rather it be its own flag, and I haven't seen a case for one yet — the operator path is staging, where fleet-wide is what you want.

The runbook now carries the same distinction, since that's where someone decides what turning it on costs: the tier table marks it as the row that ignores affinity, and the knob table says per-service and unattended in as many words.

packages/realm-server/prerender/heap-snapshot.ts
backspace: [Claude Code 🤖] Right on all three counts, and taken as your option 2 in 15ec5f8c7f.

The comment claimed two causes where the code has three, and the third is the one whose next move differs: a budget needs a fresh task, a permissions or naming fault means no capture on any task will ever land and retrying at a lower threshold just buys more pause for nothing.

The comment now says plainly that `uploadArtifact` returns the same `false` for all three and deliberately doesn't name one, and the log line changed from `heap snapshot upload declined` to `heap snapshot upload did not land … see the artifact-sink log line above for why` — pointing at the line that does know, which the sink already emits at `warn` with the underlying error.

Not taking option 1 here, for the reason you give: six call sites and `v8-prof.ts` gating a local delete on the boolean is more than this PR should carry. Worth a follow-up if a second caller ever needs to tell those apart.

Left the HTTP status alone. `upload-declined` and `failed` both landing on 500 is defensible — from the caller's side both mean "no artifact, and not because you asked wrongly" — and splitting them would encode a distinction the body already carries in `status`. Happy to be argued out of that.

backspace:
[Claude Code 🤖] All seven addressed in 15ec5f8c7f — six taken, one pushed back on with data. Thank you; the operational framing found things a code read wouldn't have.

**Taken:** the budget checked before the write rather than after (new `budget-spent` outcome); the sink gate on the auto-capture, which was announcing a capture every 30 s forever with flag-on/bucket-absent; idle gating, with patience bounded at ten ticks so a permanently-busy instance still captures before its heap grows further; both comment/code divergences; the artifact-sink flag and suffix rows plus the route's disabled-state test; and the runbook — tier row, both SSM knobs, a capture procedure with the port-forward, the measured cost, and the DevTools panel, which is not the one `.heapprofile` opens in.

**Pushed back on the blocking item.** This threshold has already run: four automatic captures on a staging task at the default produced writes of 6.9–15.9 s and files of 198–512 MB, all uploaded, none replaced. The extrapolation is out by about an order of magnitude for this workload, and your own caveat says why — a prerender heap is unusually sparse (2.03 M nodes at 1564 MB, file 0.13× the heap, against 2× in your harness). Also, the `HEALTHCHECK` isn't enforced: `healthCheck` is `null` on the task definition, so ECS never runs it. The manager's 30 s heartbeat is the real bound, worst observed pause is half of it, and the doc comment now carries the measured figures and names the heartbeat as the number to respect. Detail in that thread.

**Two things I left alone**, both flagged rather than hidden: `__rearmAutoCaptureForTests` still has no QUnit caller — the arm cycle needs a real capture to advance, so that path is verified by measurement rather than by the suite — and `upload-declined` still shares a 500 with `failed`, since the body's `status` already carries the distinction.

On the adjacent note about endpoint authentication: agreed it matches the existing posture, and agreed it's the first route whose cost to a caller is measured in seconds of process downtime. Worth recording somewhere that isn't a PR thread if the reachability of that port ever comes up.

habdelra (approved):

